PR #200, opened 2015-01-09
Removed data table views from Predictions widget.

thocevar:


astaric:
The body of the predict widget was a complicated reimplementation of the data table widget. This pull request removes the presentation part of the widget, data table widget can be used to view/sort results.

PR #211, opened 2015-01-16
Show units when setting the relative train size in Test Learners.

marinkaz:


markotoplak:
Please merge with master and update the pull request.

PR #213, opened 2015-01-16
Fixed view undo so that it returns to the exact previous point. 

markotoplak:
Also allow view redo (with ViewBox default key "+").


markotoplak:
This also solves the bug with axes (when changing variables shown) when the graph is zoomed-in.

PR #223, opened 2015-01-23
Testlearners multi-class selection.

acopar:
Evaluate prediction performance for selected class. To refresh results, apply button in learner has to be pressed.


lanzagar:
Superseded by #265

PR #246, opened 2015-02-05
Table.Y is 1D for single class data.

astaric:


janezd:
@astaric, you were not sure about this change. What are the arguments against it? Having to use _Y in Table?

I also hate having self.X and self._Y. On the other hand, I would hate it even more to make the code consistent by defining X as a property that returns _X.

I checked the pull request and I think it's (technically) OK, we just have to decide whether we like it. I haven't implemented any classifier, so I'm not the right person to make the decision.

PR #248, opened 2015-02-05
Add domain __eq__.

astaric:


Orange/data/variable.py
```diff
@@ -476,6 +484,14 @@ def ordered_values(values):
                 return presorted
         return sorted(values)
 
+    def __eq__(self, other):
+        if not isinstance(other, DiscreteVariable):
+            return False
+        return self.name == other.name and self.values == other.values
+
+    def __hash__(self):
+        return hash(self.name) ^ hash(self.name)
```
lanzagar: hash(self.name) ^ hash(self.values)

PR #249, opened 2015-02-05
Loading data table from URL

nikicc:


Orange/data/table.py
```diff
@@ -20,6 +20,9 @@
 
 from threading import Lock
 
+import tempfile
+import urllib
```
lanzagar: Imports from the python standard library should be in the first group of imports (on top).

Orange/tests/test_table.py
```diff
@@ -1080,6 +1080,15 @@ def test_calling_new_with_keyword_argument_filename_calls_read_data(
         read_data.assert_called_with(self.filename)
 
 
+class CreateTableWithUrl(TableTests):
+    def load_from_url(self):
+        d1 = data.Table('iris')
+        d2 = data.Table('https://raw.githubusercontent.com/biolab/orange3/'
+                        '04e48a6b11794fd300ad13579e8903e6026f077d/Orange/datasets/iris.tab')
```
lanzagar: Wouldn't a link to the latest version of the data set be better than to a version from a specific commit?
i.e. https://raw.githubusercontent.com/biolab/orange3/master/Orange/datasets/iris.tab

PR #252, opened 2015-02-09
Add Sklearn wrappers for matrix decomposition and projection.

marinkaz:
Included is base SklProjection and matrix decompositions / projections / manifold estimators: 
* PCA, 
* Sparse PCA,
* Randomized PCA,
* MDS, 
* Isomap,
* Locally linear embedding.


lanzagar:
assertEqual should be used instead of assertEquals.

> test_pca (test_pca.TestPCA) ... Orange/tests/test_pca.py:21: DeprecationWarning: Please use assertEqual instead.

(also in other tests)

PR #260, opened 2015-02-09
Port simple tree learner and random forest from Orange 2.

jzbontar:


jzbontar:
The test that fails is not very important. pkgutil.walk_packages thinks _simple_tree.so is a Python package (and fails trying to import it), when in fact _simple_tree.so is a c shared library, which I load in simple_tree.py.

All important tests pass (test_simple_tree and test_simple_random_forest).

PR #285, opened 2015-02-11
Reshuffle classes for discretization and continuization; add documentation

janezd:


Orange/preprocess/discretize.py
```diff
@@ -6,13 +6,11 @@
 from .transformation import Transformation
 from . import _discretize
 
-__all__ = ["Discretizer", "EqualFreq", "EqualWidth", "EntropyMDL",
-           "DiscretizeTable"]
+__all__ = ["EqualFreq", "EqualWidth", "EntropyMDL", "DomainDiscretizer"]
```
janezd: What about keeping `EqualFreq`, `EqualWidget` and `EntropyMDL` in `Orange.preprocess.discretize` instead of pulling them to `Orange.preprocess`?Their intended use (now) is

    disc.method = Orange.preprocess.EqualFreq(n=2)

or, more probably,

    disc.method = preprocess.EqualFreq(n=2)

which would change to

    disc.method = Orange.preprocess.discretize.EqualFreq(n=2)

or 

    disc.method = preprocess.discretize.EqualFreq(n=2)

It is a bit longer, but much cleaner -- `preprocess,EqualFreq` looks weird. It is also more consistent with the flags for continuization, which are inside the `Continuizer` class and not in the `preprocess` module.

PR #316, opened 2015-02-16
Silhouette plot generator.

mstrazar:
Silhouette plot with image save.


mstrazar:
Added interactive plot.

PR #320, opened 2015-02-16
Variable unpickler uses make; anonymous variables are no longer stored

janezd:
@astaric Almost there. The tests that fails is test_SimpleTree_classification_pickle, yet the problem is general. The test uses data with domain with variables that have no names (e.g. empty strings) and have values [0, 1]. Variable unpickler cannot distinguish between them and thus cannot unpickle them correctly. We can't solve this: this test will never pass.

Solutions:
- anonymous variables cannot be pickled (`Variable.__reduce__` raises an exception if `not self.name`),
- anonymous variables are not allowed at all (`Variable.__init__` raises an exception if the name is empty).

I vote for the former. Let us allows "quick variables", but they are temporary in nature, so they cannot be pickled.

Perhaps we should go further: anonymous variables should not be stored in `all_vars` and `Variable.make` should raise en exception if the name is empty. Anonymous variables are "unmakable".

### DIFF
--- a/Orange/tests/test_variable.py
+++ b/Orange/tests/test_variable.py
@@ -6,10 +6,30 @@ from Orange.testing import create_pickling_tests
 from Orange.data import ContinuousVariable, DiscreteVariable, StringVariable
 
 
-class DiscreteVariableTest(unittest.TestCase):
+# noinspection PyPep8Naming,PyUnresolvedReferences
+class VariableTest:
     def setUp(self):
-        DiscreteVariable._clear_cache()
+        self.varcls._clear_all_caches()
 
+    def test_dont_pickle_anonymous_variables(self):
+        self.assertRaises(pickle.PickleError, pickle.dumps, self.varcls())
+
+    def test_dont_store_anonymous_variables(self):
+        self.varcls()
+        self.assertEqual(len(self.varcls._all_vars), 0)
+
+    def test_dont_make_anonymous_variables(self):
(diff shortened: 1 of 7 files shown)


janezd:
**Edit:** I named the attributes in the test, so it now passes.

janezd:
@astaric Done, you can merge if you like it

Orange/tests/test_variable.py
```diff
@@ -99,16 +121,23 @@ def test_adjust_decimals(self):
         a.val_from_str_add("5.1234")
         self.assertEqual(a.str_val(4.654321), "4.6543")
 
+    def dont_pickle_anonymous_variables(self):
```
astaric: What is this methods supposed to do?

janezd: Which one?

If you try to pickle variables without names, it raises an exception.

If you refer to test_adjust_decimals, it val_from_str_add adjusts the number of decimals to the largest number it sees in the string data. Since the `add` was given four decimals, `str_val` should give four decimals.

astaric: def dont_pickle_anonymous_variables(self)

As is does not start with test_, it is not considered a test method. And there is already

VariableTest. test_dont_pickle_anonymous_variables

which does that.

janezd: Sorry, I overlooked it. Fixed.

PR #321, opened 2015-02-19
Widgets fix

ales-erjavec:
Fix various issues in widgets.

### DIFF
--- a/Orange/preprocess/discretize.py
+++ b/Orange/preprocess/discretize.py
@@ -43,7 +43,7 @@ class Discretizer(Transformation):
     @classmethod
     def create_discretized_var(cls, var, points):
         lpoints = list(points)
-        if points:
+        if lpoints:
             values = [
                 cls._fmt_interval(low, high, var.number_of_decimals)
                 for low, high in zip([-np.inf] + lpoints, lpoints + [np.inf])]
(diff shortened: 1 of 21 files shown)


Orange/classification/random_forest.py
```diff
@@ -2,7 +2,7 @@
 
 from Orange.classification import SklLearner, SklModel
 
-__all__ = ["RandomForestLearner"]
+__all__ = ["RandomForestLearner", "RandomForestClassifier"]
```
lanzagar: It was decided that __all__ should contain only classes that a user should see (and would use).
If other classes are needed elsewhere in the code they should be imported explicitly with the full path.

Orange/preprocess/preprocess.py
```diff
@@ -10,7 +10,7 @@
 from . import impute, discretize
 from ..misc.enum import Enum
 
-__all__ = ["Continuize", "Discretize", "Impute", "SklImpute"]
+__all__ = ["Preprocess", "Continuize", "Discretize", "Impute", "SklImpute"]
```
lanzagar: It was decided that __all__ should contain only classes that a user should see (and would use).
If other classes are needed elsewhere in the code they should be imported explicitly with the full path.

PR #345, opened 2015-03-06
Change the info/warning/error status widget implementation.

ales-erjavec:
Now uses an overlay widget.

### DIFF
--- a/Orange/widgets/widget.py
+++ b/Orange/widgets/widget.py
@@ -5,10 +5,9 @@ import os
 import warnings
 
 from PyQt4.QtCore import QByteArray, Qt, pyqtSignal as Signal, pyqtProperty,\
-    QDir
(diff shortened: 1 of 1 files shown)


ales-erjavec:
However, I am not sure this is the right approach, it 'hides' too much of the widget behind it (the overlay can also be placed at the bottom, but that is not much better).

janezd:
I am not sure, either. It is OK for errors and warnings which you have to read and "accept" (the widget could not do something because of this and that), but not for less important warning and informational messages (like you supplied test data to Test learners, but you're not using it).

It looks very nice, though. ;) Even if we do not use them right now, I would keep the overlays for future use.

We have to decide. Perhaps we can show errors with overlays, but incorporate informative messages in particular widgets to some placeholder in the place that is reserved, available and does not change the widget size.

PR #370, opened 2015-03-27
Replace mock with unittest.mock

janezd:


lanzagar:
Mock should also be removed from requirements.txt if it is no longer needed.

PR #375, opened 2015-03-27
.pickle format for Table, reorganization of file writers and readers

janezd:
I don't know how to modify the mock in one of the tests, so I disabled the test. @astaric , is there an elegant way to solve this?

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -651,3 +651,4 @@ FILE_FORMATS = (TabDelimFormat, TxtFormat, BasketFormat, ExcelFormat,
 FILE_WRITERS = {c.EXT: c for c in FILE_FORMATS if hasattr(c, "write_file")}
 FILE_READERS = {c.EXT: c for c in FILE_FORMATS if hasattr(c, "read_file")}
 FORMAT_NAMES = {c.EXT: c.NAME for c in FILE_FORMATS}
+
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -31,7 +31,10 @@ class FileReader:
         return values, decimals
 
 
-class TabDelimReader:
+class TabDelimFormat:
+    NAME = "Tab-delimited file"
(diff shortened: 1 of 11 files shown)


astaric:
Something like this?

    @patch("os.path.exists", Mock(return_value=True))
    def test_read_data_calls_reader(self):
        table_mock = Mock(data.Table)
        reader_instance = Mock(read_file=Mock(return_value=table_mock))

        with patch.dict(io.FILE_READERS, {'.xlsx': lambda: reader_instance}):
            table = data.Table.from_file("test.xlsx")

        reader_instance.read_file.assert_called_with("test.xlsx", data.Table)
        self.assertEqual(table, table_mock)

Orange/data/table.py
```diff
@@ -469,18 +477,16 @@ def from_file(cls, filename):
...
-
+        reader = io.FILE_READERS.get(ext)
+        if not reader:
+            desc = io.FORMAT_NAMES.get(ext)
+            if desc:
+                raise IOError("Reading {}s is not supported".
+                              format(desc.lower()))
+            else:
+                raise IOError("Unknown file name extension.")
+        print(absolute_filename)
```
astaric: Please remove.

Orange/data/io.py
```diff
@@ -617,3 +631,24 @@ def read_file(self, file, cls=None):
...
+    @staticmethod
+    def read_file(file, _=None):
+        return pickle.load(open(file, "rb"))
+
+    @staticmethod
+    def write_file(filename, table):
+        pickle.dump(table, open(filename, "wb"))
+
+
+FILE_FORMATS = (TabDelimFormat, TxtFormat, BasketFormat, ExcelFormat,
```
astaric: I would prefer a class decorator for this, so the registration of the format is closer  to the class definition.

janezd: @astaric Brilliant. Done.

PR #400, opened 2015-05-11
Fix softmax regression and add unittest.

VesnaT:


Orange/classification/softmax_regression.py
```diff
@@ -10,26 +11,40 @@ class SoftmaxRegressionLearner(Learner):
...
+        Parameters for L-BFGS algorithm.
     """
 
     name = 'softmax'
 
-    def __init__(self, lambda_=1.0, normalize=True, preprocessors=None,
-                 **fmin_args):
+    preprocessors = [Continuize(normalize_continuous=Continuize.NormalizeBySD),
+                     Impute(),
+                     RemoveNaNColumns()]
```
kernc: Do these `SoftmaxRegressionLearner.preprocessors` ever get used?

janezd: See the base class, https://github.com/biolab/orange3/blob/master/Orange/classification/base.py, in particular lines 33 in 51.

kernc: It was late. :P Thanks.

Orange/classification/softmax_regression.py
```diff
@@ -10,26 +11,40 @@ class SoftmaxRegressionLearner(Learner):
...
     - consider using many logistic regression models (one for each
       value of the class variable) instead of softmax regression.
 
-    lambda\_ : float, optional (default=1.0)
-       the regularization parameter. Higher values of lambda\_
-       force the coefficients to be smaller
+    Parameters
+    ----------
+
+    lambda_ : float, optional (default=1.0)
```
lanzagar: Underscores should be escaped ("lambda\_") otherwise sphinx has problems while compiling the documentation.

Orange/tests/test_softmax_regression.py
```diff
@@ -0,0 +1,34 @@
...
+
+import Orange
+from Orange.classification import SoftmaxRegressionLearner
+
+
+class SoftmaxRegressionTest(unittest.TestCase):
+    def test_SoftmaxRegression(self):
+        table = Orange.data.Table('iris')
+        learner = SoftmaxRegressionLearner()
+        results = Orange.evaluation.CrossValidation(table, [learner], k=10)
```
lanzagar: When using cross-validation in unit tests it is usually enough to use k=3 (or even k=2) which makes it about 3x faster.

PR #410, opened 2015-05-15
Fix assert in data table widget.

ales-erjavec:
The assert would be falsely be triggered if input data table has
lenght 0.


ales-erjavec:
Superseded by #412

PR #418, opened 2015-05-25
Linear projection

ales-erjavec:
Add a Linear Projection widget.


astaric:
Since 876811046, dicts are no longer valid signal declarations. Signals can be declared using tuples (short form), or instances of InputSignal/OutputSignal (where keyword arguments can be used for clarity).

PR #421, opened 2015-05-26
Add Variable.is_{continuous,discrete,string} properties

kernc:
~~Merges at one's own peril.~~

### DIFF
--- a/Orange/widgets/data/owcontinuize.py
+++ b/Orange/widgets/data/owcontinuize.py
@@ -111,7 +111,7 @@ class OWContinuize(widget.OWWidget):
     #     self.send("Preprocessor", PreprocessedLearner(
     #         lambda data, weightId=0, tc=(self.targetValue if self.classTreatment else -1):
     #             Table(continuizer(data, weightId, tc)
-    #                 if data.domain.class_var and isinstance(self.data.domain.class_var, DiscreteVariable)
+    #                 if data.domain.class_var and self.data.domain.class_var.is_discrete
     #                 else continuizer(data, weightId), data)))
 
     def commit(self):
@@ -138,17 +138,7 @@ class OWContinuize(widget.OWWidget):
 from Orange.preprocess.transformation import \
     Identity, Indicator, Indicator1, Normalizer
 
-from functools import partial, wraps, reduce
-
-
-# flip:: (a * b -> c) -> (b * a -> c)
-def flip(func):
-    """Flip parameter order"""
-    return wraps(func)(lambda a, b: func(b, a))
-
-
-is_discrete = partial(flip(isinstance), Orange.data.DiscreteVariable)
-is_continuous = partial(flip(isinstance), Orange.data.ContinuousVariable)
+from functools import reduce
 
 
 class WeightedIndicator(Indicator):
@@ -219,9 +209,9 @@ def continuize_domain(data_or_domain,
 
     def needs_dist(var, mtreat, ctreat):
         "Does the `var` need a distribution given specified flags"
-        if isinstance(var, Orange.data.DiscreteVariable):
+        if var.is_discrete:
             return mtreat == Continuize.FrequentAsBase
-        elif isinstance(var, Orange.data.ContinuousVariable):
+        elif var.is_continuous:
             return ctreat != Continuize.Leave
         else:
             raise ValueError
@@ -267,7 +257,7 @@ def continuize_var(var,
                    continuous_treatment=Continuize.Leave,
                    zero_based=True):
 
-    if isinstance(var, Orange.data.ContinuousVariable):
+    if var.is_continuous:
         if continuous_treatment == Normalize.NormalizeBySpan:
             return [normalize_by_span(var, data_or_dist, zero_based)]
         elif continuous_treatment == Normalize.NormalizeBySD:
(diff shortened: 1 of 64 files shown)


kernc:
Please use this new provided interface.

It mostly* works.

__*__ Mostly as far as it was tested. Not all tests pass cleanly so there is no way to know. :grinning:

Orange/widgets/data/owcontinuize.py
```diff
@@ -138,17 +138,7 @@ def sendReport(self):
...
-from functools import partial, wraps, reduce
-
-
-# flip:: (a * b -> c) -> (b * a -> c)
-def flip(func):
-    """Flip parameter order"""
-    return wraps(func)(lambda a, b: func(b, a))
-
-
-is_discrete = partial(flip(isinstance), Orange.data.DiscreteVariable)
```
kernc: Hard to see such fun stuff go. :crying_cat_face:

Orange/widgets/visualize/owdistributions.py
```diff
@@ -16,13 +15,7 @@
...
 from Orange.widgets.widget import InputSignal
 
-
-def is_discrete(var):
-    return isinstance(var, Orange.data.DiscreteVariable)
-
-
-def is_continuous(var):
-    return isinstance(var, Orange.data.ContinuousVariable)
+import pyqtgraph as pg
```
astaric: Why move pyqtgraph imports bellow Orange ones? Pyqtgraph is external library, same as numpy (and PyQt).

kernc: Ah, yes, because the modules didn't exec for me with that import order. In fact, I get this inside virtualenv:
```python
>>> import pyqtgraph
>>> import Orange.widgets
  File "/home/jk/Downloads/orange3/Orange/widgets/__init__.py", line 4, in <module>
    import pkg_resources
  File "/usr/lib/python3/dist-packages/pkg_resources.py", line 50, in <module>
    from cStringIO import StringIO as BytesIO
ImportError: No module named 'cStringIO'
```
The culprits are [pyqtgraph setting `basestring` as builtin in Python3](https://github.com/pyqtgraph/pyqtgraph/blob/1ecceeaa9248c180133b8a52edf7ada29b43931c/pyqtgraph/python2_3.py#L44) and pkg_resources relying on `basestring` existing only in Python 2.

Upgrading to the latest setuptools resolved it for me. I shall revert those import changes.

kernc:
Rebased and PEP8-ified. AFAIK prime for merging ASAP. :+1:

nikicc:
I am afraid this migh cause some problems. For example, load a file, remove the class variable with select columns (e.g. move it to meta attributes) and then connect it to scatterplot. The code breaks at:
`  File "..../Orange/widgets/utils/scaling.py", line 119, in set_data
    self.data_has_continuous_class = full_data.domain.class_var.is_continuous
AttributeError: 'NoneType' object has no attribute 'is_continuous'`
since the `class_var` is None and therefore of course does not have `is_continuous` attribute.

janezd:
This is serious: `isinstance` could have been used for checking for `None`, too. @kernc, could you go through the changes and check in which of them we should also add checking for `None`? Or @VesnaT, if you ran out of bug reports?

kernc:
```py
self.data_has_continuous_class = full_data.domain.class_var and full_data.domain.class_var.is_continuous
```
or
```py
self.data_has_continuous_class = isinstance(full_data.domain.class_var, ContinuousVariable)
```
or
```py
self.data_has_continuous_class = full_data.domain.has_continuous_class   # TBD
```
Seems like one of these is needed wherever `class_var` is assumed to exist. I'd go with no. 3 for aesthetic reasons. Any preference?

janezd:
I prefer no. 3, too.

Otherwise, when we don't have a domain but a single object, I prefer

    has_continuous_class = class_var and class_var.is_continuous

over

    has_continuous_class =  isinstance(class_var, ContinuousVariable)

precisely to prevent the problem that @nikicc discovered: the second case obscures the fact that `class_var` may also be `None`.

kernc:
But what about `Domain.class_vars`? If class variables are several, [`class_var` gets set to `None` anyway](https://github.com/biolab/orange3/blob/683df404847c5afe5c2186093d054349ee04b65f/Orange/data/domain.py#L107), and something like
```py
def has_continuous_class(self):
    return (self.class_var and self.class_var.is_continuous or
            any(var.is_continuous for var in self.class_vars))
```
or even
```py
def has_continuous_class(self):
    return any(var.is_continuous for var in self.class_vars)
```
doesn't resolve anything; `Domain.has_continuous_class` might be `True`, but `class_var` is still `None` and will fail elsewhere.

Why are widgets and algorithms allowed to assume there is a single class variable only. And if so, why is there `Domain.class_vars` ?

janezd:
Widgets assume a single class. Multiple classes are uncommon and have been added only later. Just a few algorithms actually handle them; it was even questionable whether to include this support or not.

Since this is an exception, widgets only check `class_var`. If there are multiple classes, this is equivalent (for those widgets) to having no class, hence `class_var` is `None` when `class_vars` contains more than one element.

`has_continuous_class` should only check `self.class_var and self.class_var.is_continuous`. Functions that will call it will expect this behaviour and would fail with multiple classes.

I don't think we need a function that will check `any` or `all` `var.is_continuous`.

PR #438, opened 2015-06-10
Add Elliptic Envelope learner and widget Outliers

VesnaT:


lanzagar:
Rebase on latest master and the tests should pass.

Orange/classification/elliptic_envelope.py
```diff
@@ -0,0 +1,27 @@
...
+        self.params = vars()
+        self.clf = None
+
+    def fit(self, X, Y=None, W=None):
+        self.clf = self.__wraps__(**self.params)
+        if W is not None:
+            return self.__returns__(self.clf.fit(X, W.reshape(-1)))
+        return self.__returns__(self.clf.fit(X))
+
+    def mahalanobis(self, observations):
```
astaric: Missing docstring.

Orange/classification/elliptic_envelope.py
```diff
@@ -0,0 +1,39 @@
...
+        """Return Mahalanobis distances of the training set observations.
```
lanzagar: Is there a reason why we would not use the same as scikit-learn:
Computes the squared Mahalanobis distances of given observations.

(also below)

Orange/classification/elliptic_envelope.py
lanzagar: If we wrap the original method so that the Orange learner also provides it,
we should probably allow observations to be an Orange.data.Table?

Orange/classification/elliptic_envelope.py
```diff
@@ -0,0 +1,39 @@
...
+class EllipticEnvelopeLearner(SklLearner):
+    __wraps__ = skl_covariance.EllipticEnvelope
+    name = 'elliptic envelope'
+
+    def __init__(self, store_precision=True, assume_centered=False,
+                 support_fraction=None, contamination=0.1,
+                 random_state=None, preprocessors=None):
+        super().__init__(preprocessors=preprocessors)
+        self.params = vars()
+        self.clf = None
```
lanzagar: self.supports_weights = True
and then the fit method is not even needed

astaric:
Can you rebase on the current master, so we can merge this pull request?

lanzagar:
The tests failed for a reason and did not need fixing like in the last commit.
Elliptic envelope is not a classifier and does not fit a discrete class. It is an unsupervised method that should work on data with any kind of class or even without a class.
However, scikit-learn does partly reuse the interface for supervised methods (e.g. predict()) so we should decide how to include it into Orange.

But the class should probably be in Orange/unsupervised and not force any conditions on the class variable.

PR #449, opened 2015-06-18
Widgets I get to use repeatedly

kernc:
~~Not for merging.~~ RFC


kernc:
One of the high overseers would transitively like to see this merged. So with no reservations, please go ahead. :smile:

PR #451, opened 2015-06-18
New module Orange.util

kernc:


kernc:
Please review at convenience.

Orange/util.py
```diff
@@ -0,0 +1,68 @@
...
+                raise NotImplementedError('Can\'t instantiate abstract class ' + obj.__name__)
+            return old__new__(cls, *args, **kwargs)
+
+        obj.__new__ = _refuse__new__
+        return obj
+    else:
+        return abc.abstractmethod(obj)
+
+
+def cached(func):
```
astaric: Is this decorator really suitable for functions that receive arguments? If I use it to decorate a function sum, and then sum different lists, the result will always be the same (the sum of the first list).

Even when fixed, why is it better than itertools.lru_cache(maxsize=1)?

Orange/util.py
```diff
@@ -0,0 +1,68 @@
...
+import sys
+import abc
+import numpy as np
+
+
+def scale(values, min=0, max=1):
+    """Return values scaled to [min, max]"""
+    ptp = np.nanmax(values) - np.nanmin(values)
+    if ptp == 0:
+        return np.clip(values, min, max)
```
astaric: Did you consider using (min+max) / 2? This would produce the same scaled value for all constant arrays.

astaric:
Coveralls would like to see some test :)

Orange/util.py
kernc: I didn't yet. My implementation also returns the same value for all constant array, but, obviously, by just clipping it to the selected range, keeps more information, however "worthless" it may be. Would centering be better? If I have an attribute of values 1–5, and a random subset is selected consisting only of 5's, then a range of 3's is returned (provided the scaling is done to [1, 5], for whatever reason), whereas 5's are kept in my case.

Orange/util.py
kernc: Absolutely correct again. Throwing it out.

PR #469, opened 2015-06-26
Majority classifier modified such that it returns the same random cla…

mstrazar:
…ss for datasets with uniform class distributions. This is based on SHA1 hash function and should be consistent across systems.

### DIFF
--- a/Orange/classification/majority.py
+++ b/Orange/classification/majority.py
@@ -1,4 +1,5 @@
-from numpy import tile, array
+from numpy import tile, array, zeros
+from hashlib import sha1
 
 from Orange import data
 from Orange.classification import Learner, Model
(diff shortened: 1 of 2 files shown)


Orange/classification/majority.py
```diff
@@ -56,7 +71,11 @@ def predict(self, X):
         :return: predicted value
         :rtype: vector of majority values
         """
-        return tile(self.dist, (len(X), 1))
+        probs = tile(self.dist, (len(X), 1))
+        if self.unif_maj:
```
lanzagar: The if condition fails when self.unif_maj == 0.
Should be:

    if self.unif_maj is not None:

Orange/tests/test_majority.py
```diff
@@ -55,10 +55,10 @@ def test_missing(self):
         iris = Table('iris')
         learn = MajorityLearner()
         for e in iris:
-            e.set_class("?")
+             e.set_class("?")
```
lanzagar: Was this intentional?

mstrazar: No, fixed.

Orange/classification/majority.py
mstrazar: Fixed.

astaric:
Manually merged in bf41cfc.

PR #471, opened 2015-06-26
OWWidget: test input signal handlers after __init__ ...

kernc:
... catch connection removing bugs early.

RFC

### DIFF
--- a/Orange/widgets/widget.py
+++ b/Orange/widgets/widget.py
@@ -145,6 +145,18 @@ class OWWidget(QDialog, metaclass=WidgetMetaClass):
         if self.want_basic_layout:
             self.insertLayout()
 
+        def patched__init__(*args, _old__init__=cls.__init__, **kwargs):
+            """
+            After initialization, test input signal handlers with None as input
+            => catch connection deleting bugs early.
+            """
+            _old__init__(*args, **kwargs)
+            for input in self.inputs:
+                args = (None,) if input.single else (None, 0)
+                getattr(self, input.handler)(*args)
+
+        cls.__init__ = patched__init__
+
         return self
 
     def __init__(self, *args, **kwargs):


Orange/widgets/widget.py
```diff
@@ -151,6 +151,11 @@ def __init__(self, *args, **kwargs):
         """QDialog __init__ was already called in __new__,
         please do not call it here."""
 
+        # Test input signal handlers with None == catch deleting connection bugs early
+        for input in self.inputs:
+            args = (None,) if input.single else (None, 0)
+            getattr(self, input.handler)(*args)
+
     @classmethod
```
ales-erjavec: No. This is most definitely not the right place to check this.
At the point when most widgets call the base `__init__` they are not even fully initialized.
e.g.

    inputs = [("Data", int, "set_number")]
    def __init__(self):
        super().__init__()  # will call set_number(None)
        self.number  = None  # member instances are usually defined after
    
    def set_number(self, data):
        if self.number == number:  # BAM: Attribute error
             ...

kernc: Indeed, you are correct, sir. Please have another look.

Orange/widgets/widget.py
```diff
@@ -145,6 +145,18 @@ def __new__(cls, parent=None, *args, **kwargs):
...
+            => catch connection deleting bugs early.
+            """
+            _old__init__(*args, **kwargs)
+            for input in self.inputs:
+                args = (None,) if input.single else (None, 0)
+                getattr(self, input.handler)(*args)
+
+        cls.__init__ = patched__init__
+
         return self
```
ales-erjavec: No.

You are capturing the `cls.__init__` on every `__new__` call, but  the old `__init__` will (except for the first call) be a `patched__init__` of the penultimate widget instance. Note that the `pathched__init__` captures the `self` instance.

In essence this creates a chain of all class instances which have ever been created in a process and clears all of their input whenever a new instance of the class is created.

PR #473, opened 2015-06-29
Box plot bar annotate

ales-erjavec:


astaric:
I have encountered the following problems:
* Label falls out of window on iris dataset.
* Error is raised when grouping is turned off.

PR #474, opened 2015-06-29
New Orange.data.filter.FilterRegex

kernc:


astaric:
Some tests would be nice.

kernc:
> Some tests would be nice.

Can haz.

PR #476, opened 2015-06-29
Test Learners - report an error instead of crashing.

thocevar:
Test Learners reports an error instead of crashing when one of the learners crashes.


lanzagar:
This can show very cryptic error messages above the widget (e.g. for linear regression on iris "negative value in bincount") and hides what caused them, so I don't like it except as a temporary solution.

Once classification and regression learners can be easily differentiated, we should rethink how to handle other errors.

kernc:
Temporary solutions are not the way to go. :-1:

PR #482, opened 2015-07-02
Add separator to dropdown - alternative approach

VesnaT:


Orange/widgets/utils/itemmodels.py
```diff
@@ -257,9 +258,13 @@ def __setitem__(self, s, value):
 
             if not isinstance(value, list):
                 value = list(value)
+            separators = [start + i for i, v in enumerate(value) if v is self.Separator]
```
kernc: Is `self.Separator` ever `False`?

PR #485, opened 2015-07-03
Replace NaN values in distance matrix with 0 for Pearson and Spearman metrics

VesnaT:
… metrics.


lanzagar:
When is the distance NaN?
If only when data contains NaNs, would it not be better to impute the data before computing the correlation?

VesnaT:
No. The distance is NaN, when all features of a sample have the same value.

VesnaT:
It has nothing to do with NaNs in data.

PR #486, opened 2015-07-06
Enable discretization which does not remove constant features

VesnaT:


lanzagar:
The if condition checks if there are no attributes.
How is this related to the commit message and checking if a feature has constant value (and was removed)?

lanzagar:
As discussed, discretization needs to be changed so that it does not remove features (also because of problems elsewhere).

After that this fix should become unnecessary.

PR #495, opened 2015-07-10
Make copies of variables for Random and Model imputers.

nikicc:
Random and Model imputers used to irreversably changed the input data, since there was no copy made for the codomain. Now we make copies so the input data is not changed.


lanzagar:
Random imputation randomizes all values instead of just the missing.

astaric:
My take on fixing Model based imputation:

https://github.com/astaric/orange3/commit/36d616ba2b43586ac6263b47cecba156070d17aa

It does not modify original data and works when used through imputer or compute_value. I am not sure how it fits into the current imputer widget infrastructure.

ales-erjavec:
I think the primary error is in the ImputerModel__call__, specifically the 

    if Xp is X:
            Xp = Orange.data.Table(Xp)

which should create a copy of the table, while in reality it is a view.

PR #499, opened 2015-07-14
OWSave is used to save images and graphs within imageViewer and TreeV…

VesnaT:
…iewer

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -5,13 +5,14 @@ import pickle
 from itertools import chain
 
 import bottlechest as bn
-import numpy as np
+from sklearn import tree
 from scipy import sparse
 # We are not loading openpyxl here since it takes some time
 
 from Orange.data import _io
 from Orange.data import Domain
 from Orange.data.variable import *
+from PyQt4 import QtGui, QtCore
 
 
 # A singleton simulated with a class
@@ -20,6 +21,8 @@ class FileFormats:
     names = {}
     writers = {}
     readers = {}
+    img_writers = {}
+    graph_writers = {}
 
     @classmethod
     def register(cls, name, extension):
@@ -31,6 +34,10 @@ class FileFormats:
                 cls.writers[extension] = format
             if hasattr(format, "read_file"):
                 cls.readers[extension] = format
+            if hasattr(format, "write_image"):
+                cls.img_writers[extension] = format
+            if hasattr(format, "write_graph"):
+                cls.graph_writers[extension] = format
             return format
         return f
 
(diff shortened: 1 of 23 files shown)


Orange/data/io.py
```diff
@@ -656,3 +669,51 @@ def read_file(cls, file, _=None):
     def write_file(cls, filename, table):
         pickle.dump(table, open(filename, "wb"))
 
+    def write(self, filename, table):
+        self.write_file(filename, table)
+
+
+@FileFormats.register("Portable Network Graphics", ".png")
+class PngFormat:
```
kernc: Consider also [QSvgGenerator](http://doc.qt.io/qt-4.8/qsvggenerator.html#details)?

Orange/data/io.py
```diff
@@ -5,13 +5,14 @@
 from itertools import chain
 
 import bottlechest as bn
-import numpy as np
```
kernc: Isn't `np` used further below in the file?

Orange/data/io.py
```diff
@@ -5,13 +5,14 @@
...
+from sklearn import tree
 from scipy import sparse
 # We are not loading openpyxl here since it takes some time
 
 from Orange.data import _io
```
kernc: Seeing where `_io` is used, wouldn't it make sense to just move the exotic import into that function? For a faster boot time at the least! :smiley:

Orange/data/io.py
```diff
@@ -6,12 +6,13 @@
 
 import bottlechest as bn
 import numpy as np
+from sklearn import tree
```
kernc: Try moving this import below all the others, or where it's only ever used?

kernc:
... When you get it working, some tests would be in order. :]

astaric:
The tests fails, because tests for pull requests did not install pyqt on travis (installation used to be slow). If you rebase on current master, the test will probably pass.

Do we really want to have PyQt imports inside the Orange.**data**.io? As FileFormats is a registry, readers and writers can be implemented anywhere. If ImgFormat and classes that extend it can only save scenes, that are usually related to widgets, perhaps a more suitable place for this code would be inside Orange.widgets package?

lanzagar:
Try to save a png from the Calibration Plot widget.
I get a 'module' not callable error
Problematic lines:
https://github.com/biolab/orange3/pull/499/files#diff-8c9565c33ed8859fe616fb9774235f27R33
https://github.com/biolab/orange3/pull/499/files#diff-8c9565c33ed8859fe616fb9774235f27R74

Check for similar problem in SVG...

kernc:
Pull requests, as contained wholes, should (imho) not be merged until [properly squashed](https://www.google.com/search?q=github+squash).

PR #502, opened 2015-07-15
Hide unused attribute values in Mosaic Display.

thocevar:


lanzagar:
1) This only removes the label, which is not the same as removing a value - its "zero width/height" mosaic box area is still there.

2) It removes the labels inconsistently:
- on x-axis only if the whole column is empty
- on y-axis if the leftmost box is empty

PR #505, opened 2015-07-16
Remove/deprecate Orange.canvas.utils.environ

ales-erjavec:


ales-erjavec:
NOTE: I changed the directory structure for canvas/widget settings (described in Orange.misc.environ docstring).

PR #507, opened 2015-07-16
Coveralls might have been unable to find .coverage file?

kernc:
:+1: with no better ideas.


kernc:
Yup. Perhaps needs some more tuning, but it works. :smiley:

PR #514, opened 2015-07-18
Fix fallbacks

astaric:


Orange/statistics/contingency.py
```diff
@@ -80,8 +80,9 @@ def from_data(cls, data, col_variable, row_variable=None):
             self[...] = dist
             self.unknowns = unknowns
         except NotImplementedError:
-            self = np.zeros(
-                (len(row_variable.values), len(col_variable.values)))
+            shape = len(row_variable.values), len(col_variable.values)
+            self = super().__new__(cls, shape)
+            self[...] = np.zeros(shape)
```
kernc: Couldn't this simply be `self[...] = 0` to avoid generation of an intermediary array?

PR #519, opened 2015-07-20
PCA - Change attribute names

VesnaT:


Orange/projection/pca.py
```diff
@@ -71,7 +71,7 @@ class PCAModel(Projection, metaclass=WrapperMeta):
     def __init__(self, proj, domain):
         def pca_variable(i):
             v = Orange.data.ContinuousVariable(
-                'PC %d' % i, compute_value=Projector(self, i))
```
lanzagar: Why manual casting to str instead of
'PC%d' % (i + 1)

Orange/projection/pca.py
```diff
@@ -71,7 +71,7 @@ class PCAModel(Projection, metaclass=WrapperMeta):
...
+                'PC%s' % str(i + 1), compute_value=Projector(self, i))
```
lanzagar: Why manual casting to str instead of
'PC%d' % (i + 1)

PR #522, opened 2015-07-21
Build PyQt on travis (instead of using wheelhouse).

astaric:


kernc:
Why are we building PyQt again? Seems like [an appropriate version](http://packages.ubuntu.com/precise/python/python3-pyqt4) is available.

astaric:
Which python? We are testing on 3.4. The last time i checked, only python
3.2 was provided using apt-get.

kernc:
Aye.

PR #524, opened 2015-07-21
Consistent Orange.data.io.FileFormat

kernc:
More consistent FileFormat readers/writers. Blocked on https://github.com/biolab/orange3/pull/451

The code is ugly as shit to accommodate all the peculiarities of the incorrectly (under)documented format, but at least it's a consistent single point of failure.

Documentation should be revised.


Orange/tests/test_domain.py
```diff
@@ -368,7 +368,7 @@ def test_conversion(self):
         x, y, metas = domain.convert([42, 13, "White"])
         assert_array_equal(x, np.array([42, 13]))
         assert_array_equal(y, np.array([0]))
-        assert_array_equal(metas, np.array([Unknown, Unknown, None]))
```
kernc: Not exactly sure what's the deal here. The previous implementation seems to have returned `None` for unknown `StringVariable`s?

Orange/data/variable.py
```diff
@@ -17,15 +22,173 @@ def make_variable(cls, compute_value, *args):
...
-class VariableMeta(type):
-    # noinspection PyMethodParameters
-    def __new__(mcs, name, *args):
-        cls = type.__new__(mcs, name, *args)
-        if not hasattr(cls, '_all_vars') or cls._all_vars is Variable._all_vars:
-            cls._all_vars = {}
-        if name != "Variable":
-            Variable._variable_types.append(cls)
-        return cls
+def is_discrete_values(values):
```
kernc: What makes a certain attribute discrete? The current implementation says it is discrete when:
* it has no more than `max_values = max(sqrt(n_instances), 26)` distinct values (26 is arbitrary; adapted to fit one particular test case);
* if numeric, all values are integers (specified without a floating point, i.e. `4`, not `4.0`) between 0 and `max_values`.

I'd wish to simplify this to:
* has no more than `max_values`;
* if numeric, all of whole numbers (e.g. `4.0`, `42` ...);

but in that case, some more test cases would need to be adapted.

@janezd, @astaric, @lanzagar: comments?

Orange/data/variable.py
```diff
@@ -266,15 +389,16 @@ def val_from_str_add(self, s):
         Convert a value from a string and adjust the number of decimals if
         `adjust_decimals` is non-zero.
         """
-        return _variable.val_from_str_add_cont(self, s)
+        if s in MISSING_VALUES:
+            return Unknown
+        self.number_of_decimals = len(s.strip().split('.', 1)[1]) if '.' in s else 0
```
janezd: This function does not work correctly. The mechanism for adjusting the number of decimals is more complicated. Please see the original.

Also, Marko implemented the Cython code since he discovered it speeds up the file parsing considerably. After reimplementing this function correctly, please check how it affects the speed of reading large files with continuous values.

Orange/data/variable.py
kernc: Amended in 'FIXUP' commit.

kernc:
Ready for another review. ^^

kernc:
This has seen another revision. :smile: 

Re the new chardet dependency and file encoding auto-detection in general: chardet seems to [test viability of multiple encodings in parallel](http://chardet.readthedocs.org/en/latest/how-it-works.html), and we can assume it works better than anything we could come up with. Its algorithm is used by Mozilla, python-requests, and many other players. If chardet says the encoding is X, the file can (likely) at least be interpreted in encoding X (as opposed to program crash). Particularly with Google Sheets widget and support for discrete attributes with values `כן` and `לא` in Hebrew, I see no reason against it.

Orange/tests/test_txt_reader.py
```diff
@@ -48,29 +37,25 @@ def read_easy(self, s, name):
         try:
             file.write(s)
             file.close()
-            table = TxtFormat().read_file(filename)
+            table = CSVFormat().read_file(filename)
 
             f1, f2, f3 = table.domain.variables
-            self.assertIsInstance(f1, ContinuousVariable)
-            self.assertEqual(f1.name, name + "1")
+            self.assertIsInstance(f1, DiscreteVariable)
```
astaric: All features are written as *.0 in the input file, aren't they supposed to be continuous?

Was the rule from
https://github.com/kernc/orange3/commit/f6e5ad69918427d78ebe20cc8deb8b224e6dd55a#diff-8fc32ad6a8ac706482fe82637339979aR46
droped?

kernc: Yes, the new rule, in https://github.com/kernc/orange3/commit/76303e934b393b111b941844c135014bd6b5069a, is `all(int(i) == i) → discrete`. @janezd gave verbally his hesitant but unconditional agreement on that.

Orange/data/io.py
```diff
@@ -1,384 +1,537 @@
...
+                        copy=False, dtype=object, order='F')
+
+        # Data may actually be longer than headers were
+        try: rowlen = data.shape[1]
+        except IndexError: pass
+        else:
+            for lst in (names, types, flags):
+                _equal_length(lst)
+
+        NAMEGEN = namegen('_')
```
kernc: Won't these name-generated variables (`_0`, `_1`, `_2`, ...) clash with one another if multiple sources without headers are loaded and these variables try to be unified with the existing in the `coltype.make()` call below? Wouldn't a global `NAMEGEN` make more sense?

kernc:
Sadly, Sphinx's Napoleon extension seems to only support Google/NumPy style documentation and a _very_ limited subset of Markdown (Heading 1, Heading 2, **bold** and `kbd`). So the doc is in reST for the references to work.

-----

This is :+1:.

kernc:
Due to a slight error in encoding auto-detection mechanism (file utility returns encoding and a _newline_), chardet ran every time. I only noticed it because it was particularly slow with a 2MB ISO-8859-2 CSV file. Now, in addition to the aforementioned error being fixed, text reader tries to open the file as US-ASCII before considering any auto-detection whatsoever. Some further speed-up should be apparent. :smile:

PR #526, opened 2015-07-22
New preprocessor remover used for purging in widgets

VesnaT:


Orange/preprocess/preprocess.py
```diff
@@ -249,6 +249,71 @@ def __call__(self, data):
         return normalizer(data)
 
 
+class Remove(Preprocess):
+    """
+    Construct a preprocessor for removing constant features/classes
+    and unused values.
+    Given a data table, preprocessor returns a new table and list of
```
kernc: ... and **a** list ...

Orange/preprocess/preprocess.py
```diff
@@ -249,6 +249,71 @@ def __call__(self, data):
...
+    results. In the new table the constant features/classes and unused
```
kernc: In the new table**,** ...

Orange/preprocess/preprocess.py
```diff
@@ -249,6 +249,71 @@ def __call__(self, data):
...
+    values are removed. The list of results consist of two dictionaries.
```
kernc: ... list ... consist**s** ...

Orange/preprocess/remove.py
```diff
@@ -0,0 +1,279 @@
...
+        attrs_state = [purge_var_M(var, data, self.attr_flags)
+                       for var in domain.attributes]
+        class_state = [purge_var_M(var, data, self.class_flags)
+                       for var in domain.class_vars]
+
+        att_vars, att_res = self.get_vars_and_results(attrs_state)
+        cls_vars, cls_res = self.get_vars_and_results(class_state)
+
+        domain = Domain(att_vars, cls_vars, domain.metas)
+        return data.from_table(domain, data), [att_res, cls_res]
```
lanzagar: Preprocessors need to return a new data table (e.g. `data2 = Preprocessor(data)`)
so they can be chained etc.

If the results dict is really needed, we can save the (last) results on the Remover instance:
```
remover = Remover(7, 7)
data2 = remover(data)
ar, cr = remover.attr_results, remover.cls_results
```

Another option is to add a parameter to `__call__(self, data, info=False)`
and only return the dicts when `info=True`

Orange/preprocess/remove.py
```diff
@@ -0,0 +1,279 @@
+import numpy as np
+from collections import namedtuple
```
lanzagar: https://www.python.org/dev/peps/pep-0008/#imports

lanzagar:
Checkboxes in both Select Rows and Purge Domain should be corrected (discussion)

PR #527, opened 2015-07-22
OWPreprocess - Handle exception when preprocessing

VesnaT:


Orange/preprocess/fss.py
```diff
@@ -66,7 +66,7 @@ def __call__(self, data):
 
         if not isinstance(data.domain.class_var, method.class_type):
             raise ValueError(("Scoring method {} requires a class variable " +
-                              "of type {}.").format(type(method), method.class_type))
+                              "of type {}.").format(method, method.class_type))
```
lanzagar: Maybe we can make it even shorter:
```method.__name__, method.class_type.__name__```

File: Orange/preprocess/fss.py [hunk 2]
```diff
@@ -66,7 +66,10 @@ def __call__(self, data):
 
         if not isinstance(data.domain.class_var, method.class_type):
             raise ValueError(("Scoring method {} requires a class variable " +
-                              "of type {}.").format(type(method), method.class_type))
+                              "of type {}.").format(
+                method.__name__ if hasattr(method, '__name__') else method,
```
lanzagar: Ugh, I missed that method can be a class or an instance. Nice catch.
In that case:
```(method if type(method) == type else type(method)).__name__```

PR #540, opened 2015-07-24
Improve mds visualization

janezd:


lanzagar:
Some observations (not necessarily results of changes, some were present before):
* ```_initialize()``` should probably be called at the end of ```init()```, to avoid errors when changing sliders before any data is connected.
* The default value of ```connected``` should either be 0 or much lower (2 is enough imho). Currently the default settings make the widget too slow even on small data such as wine.
Maybe a square-root slider scale would be more appropriate (i.e. the value is squared before use).
* I think the edges connecting similar pairs could be de-emphasized.
E.g. pen=pg.mkPen(0.8, width=1.5) reduces the cluttered feeling with lots of edges.
* The zoom/select box looks bad with the margin set to 0 (maybe it looks better on mac?).
The icons are also smaller than in Scatter Plot - not that important, but a uniform look of standard controls would be a bonus.
* Selecting points also resets the view.
* Output option "Without coordinates" is quite useless for the ```Data``` output, where outputting "Only coordinates" would be more useful (to use MDS as dim. reduction - big data in, 2D data out).
But this is used also for the ```Data subset``` output, where "Without coordinates" might be desired? We can probably remove or leave the first option, and add "Only coordinates" at the end.

janezd:
@lanzagar , thanks. 

> Selecting points also resets the view.

This one is kind of major. Slider also reset the view. Graph is always redrawn from scratch.

We have OWScatterplotGraph, which is a widget, similar to Scatter plot except for a few missing controls. Actual OWScatterplot adds these controls. We should extract the plot object, so we could reuse it for MDS and also for FreeViz, RadViz. Reusing the scatter plot will also fix this bug, so I haven't fixed it.

> `_initialize()` should probably be called at the end of init()

Fixed.

> The default value of connected should either be 0 or much lower (2 is enough imho).

I played a bit and set it to 5 for now. ;)

> Maybe a square-root slider scale would be more appropriate

It is essentially squared; this is the proportion of all pairs, so it rises with square of points.

> I think the edges connecting similar pairs could be de-emphasized. E.g. pen=pg.mkPen(0.8, width=1.5)

I liked them ... but OK. :)

> The zoom/select box looks bad with the margin set to 0 (maybe it looks better on mac?).

On Mac, it looks bad with (default) margins. I set them to 2 now.

> Output option "Without coordinates" is quite useless for the Data output

I added a new option. I think this one is useful with selection.

If you agree and if everything else is alright, you can merge this and we add a ticket about extracting the scatter plot graph.

lanzagar:
I agree, merged this and added a commit to prevent a minor error (9dd9da8d)

PR #541, opened 2015-07-27
Preprocessors argument in linear regression.

acopar:
Add missing preprocessors argument in linear regression learners.


Orange/regression/linear.py
```diff
@@ -14,6 +14,10 @@ class LinearRegressionLearner(SklLearner):
     __wraps__ = skl_linear_model.LinearRegression
     name = 'linreg'
 
+    def __init__(self, preprocessors=None):
+        super().__init__(preprocessors=preprocessors)
+        self.params = vars()
```
lanzagar: This learner does not have any scikit-learn parameters so this line is unnecessary.

PR #542, opened 2015-07-27
Widget settings fix

ales-erjavec:
* Fix erroneous initial widget geometry due to shared OWWidget.savedWidgetGeometry setting.
* Store/retrieve the settings from the widget when removing from the workflow (ensures that settings
  are restored between undo/redo operations.


ales-erjavec:
@astaric Can you please check if this change is ok.

astaric:
It looked ok, so I have merged it :)

PR #545, opened 2015-07-28
More scattermap improvements

janezd:


lanzagar:
Yes! Simplifications, I like them.

PR #546, opened 2015-07-28
Sensible values for C in LogReg widget

janezd:


lanzagar:
I still think it would be simpler to just let the user input any number :)
(less code and faster to set a desired number)

janezd:
I don't disagree. :) I just don't have a good explanation for C that could be put into user interface and be understandable to non-experts. That's why I use slider with "weak" and "strong".

And, well, it's easier to drag slider than to type numbers. :)

PR #552, opened 2015-07-31
Add preprocessor input to widget Test Learners.

lanzagar:


Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -164,7 +165,19 @@ def set_test_data(self, data):
         if self.resampling == OWTestLearners.TestOnTest:
             self._invalidate()
 
+    def set_preprocessor(self, preproc):
```
astaric: Should we allow multiple preprocessors?

lanzagar: I vote no (for now).
Multiple preprocessors can be set up in the Preprocess widget and sent in a single signal. If we wanted multiple preprocessor input signals, we have to deal with the order of signals and the complexity (for the user) increases.

For now, I suggest using a single input (like in all other widgets with preprocessor input). And if, after using this feature for some time, users find they need the complexity of multiple signals, I think the place to solve this is in the Preprocess widget.
E.g. it is the only widget with multiple preprocessor inputs, then the user joins them in the desired order and sends them forward in a single signal.

PR #554, opened 2015-07-31
Apply preprocessors to MDS widget it gets a sample matrix at its input

marinkaz:


Orange/widgets/unsupervised/owmds.py
```diff
@@ -20,6 +20,7 @@
 from Orange.widgets.utils import itemmodels
 
 import Orange.data
+import Orange.preprocess
```
lanzagar: This import is not needed.

PR #559, opened 2015-08-04
Only call autoRange at the most opportune time.

ales-erjavec:
The MDS plot should preserve the axis range on color, shape, size, ...
change. The autoRange should only be called when the mds embedding
coordinates change.


lanzagar:
Clicking the small A in the bottom left corner of the plot is broken. It did not work correctly before either, but since clicking it makes an empty selection this now does nothing (before it also reset the zoom).

I see now that some visualizations do not have this button at all (e.g. Scatter Plot does not have it, while Scatter Map has it). Here we do not need it since we have a button to reset the view among the Zoom/Select controls anyway. And it is probably easier to remove it than to fix it?

PR #564, opened 2015-08-05
Remove dependency on custom SQL extension "binning"

lanzagar:
The function it provided (bin) is now available in core PostgreSQL (as width_bucket).


kernc:
Nice! How come you didn't contribute it? Have you raised the idea for the other one on their hackers list?

lanzagar:
The nice guys from 2ndQuadrant who wrote the extensions also got it accepted into core postgres.
Another pull request should follow soon for the other extension.

PR #565, opened 2015-08-05
OWDistributions - Fix condition

VesnaT:
Fix weights condition to be consistent with curve estimation condition


lanzagar:
Now weights and curves have the same length (and can be zipped),
but colors still do not!

PR #566, opened 2015-08-06
Intersphinx help fix

ales-erjavec:


lanzagar:
Thanks, this fixes the error.

PR #567, opened 2015-08-06
OWDistributions - Fix color and legend conditions

VesnaT:


lanzagar:
If you are fixing this widget, could you check why regular histograms don't work with grouping (e.g. on iris)

PR #570, opened 2015-08-06
Simplified the distributions widget

markotoplak:
I have chosen average shifted histograms (with a gaussian kernel) as the only visualization for continuous attributes, because it returns similar results as gaussian KDE estimation from scikits but it was much (more that 10x) faster.

All visualizations are changed.

I have not implemented class probabilities yet.


kernc:
The last commit (https://github.com/markotoplak/orange3-1/commit/85fab179282adf93c60e6a3a1e27e252bb3fbff9) probably shouldn't be.

PR #575, opened 2015-08-07
SklLearner: change default value of _params

lanzagar:
SklLearner.params are a dict when set, so the default value (when not set) should be an empty dict.
This enables them to always be used in constructors as (**learner.params)


astaric:
Merged in fd25632

PR #578, opened 2015-08-10
Multiple inheritance for SklLearnerClassification and SklLearnerRegression

VesnaT:


lanzagar:
Why do we need ModelClassification and ModelRegression?
If their purpose is to be able to check the type of model instances, then SklModel{Classification,Regression} should also be subclasses of them.

PR #583, opened 2015-08-12
OWDataTable: Add 'Deselect all' button

VesnaT:


lanzagar:
Rows can be deselected by holding Ctrl while clicking them.

PR #584, opened 2015-08-12
C++ implementation for density image in Scatterplot.

thocevar:


lanzagar:
Is the change from checkbox to button still planned?
I am adding the [WIP] tag just in case, remove when ready.

PR #585, opened 2015-08-13
Table: Add copy method

astaric:


Orange/tests/test_table.py
```diff
@@ -598,6 +598,18 @@ def test_extend(self):
         with self.assertRaises(ValueError):
             d.extend(x)
 
+    def test_copy(self):
+        t = data.Table(np.zeros((5, 3)), np.arange(5), np.zeros((5, 3)))
+
+        copy = t.copy()
+        self.assertTrue(np.all(t.X == copy.X))
```
kernc: For direct ndarray comparison, I believe it's better (more elaborate) to use `np.testing.assert_equal`.

astaric: I had that, but as there is not assert_not_equal, I opted for the same kind of assert before and after the modification of the copied array.

PR #588, opened 2015-08-14
Test_regression: Handle TypeError while testing all regression learners

VesnaT:


Orange/tests/test_regression.py
```diff
@@ -25,6 +26,10 @@ def all_learners(self):
...
     def test_adequacy_all_learners(self):
         for learner in self.all_learners():
-            learner = learner()
-            table = Table("iris")
-            self.assertRaises(ValueError, learner, table)
+            try:
+                learner = learner()
+                table = Table("iris")
+                self.assertRaises(ValueError, learner, table)
+            except TypeError as err:
```
kernc: Sorry, what's the explanation for this? When does TypeError get raised?

lanzagar: Almost all learners can be constructed with default parameters (i.e. `learner()`) and tested.
Some (currently just PolynomialLearner, which needs a base learner) require positional arguments and raise an error. This skips them.

PR #589, opened 2015-08-14
Add ReliefF feature scoring

kernc:
Good to go.


Orange/widgets/data/owrank.py
```diff
@@ -71,8 +78,7 @@ def __new__(cls, name, shortname, score,
 
 class OWRank(widget.OWWidget):
     name = "Rank"
-    description = "Rank data features by their correlation to " \
-                  "the class variable."
+    description = "Ranks and filters data features by their relevance."
```
lanzagar: Imperative is used in other descriptions and sounds better imho:
"Rank and filter data...

lanzagar:
Coveralls want a test for ReliefF :)

kernc:
Some tests provided.

Orange/preprocess/score.py
```diff
@@ -196,3 +203,70 @@ class Gini(ClassificationScorer):
...
+        self.k_nearest = k_nearest
+
+    def score_data(self, data, feature):
+        assert len(data.domain.class_vars) == 1, 'TODO: multi-class?'
+        assert data.domain.class_var.is_discrete
+        if len(data.domain.class_var.values) == 1:  # Single-class non-problem
+            return 0 if feature else np.zeros(data.X.shape[1])
+
+        K_NEAREST = self.k_nearest or 10
+        N_ITER = self.n_iterations or 50
```
lanzagar: Why aren't 10 and 50 default values for these parameters in __init__?

Orange/preprocess/score.py
```diff
@@ -196,3 +203,70 @@ class Gini(ClassificationScorer):
...
+    feature_type = Variable
+    class_type = DiscreteVariable
+
+    def __init__(self, n_iterations=None, k_nearest=None):
+        self.n_iterations = n_iterations
+        self.k_nearest = k_nearest
+
+    def score_data(self, data, feature):
+        assert len(data.domain.class_vars) == 1, 'TODO: multi-class?'
+        assert data.domain.class_var.is_discrete
```
lanzagar: It is better to raise an appropriate error (with an informative message) if conditions for the method are not met.

Orange/preprocess/score.py
```diff
@@ -196,3 +203,70 @@ class Gini(ClassificationScorer):
...
+    class_type = ContinuousVariable
+    # TODO: preprocessors
+
+    def __init__(self, n_iterations=None, k_nearest=None):
+        self.n_iterations = n_iterations
+        self.k_nearest = k_nearest
+
+    def score_data(self, data, feature):
+        assert len(data.domain.class_vars) == 1, 'TODO: multi-class?'
+        assert data.domain.class_var.is_continuous
```
lanzagar: Same as above.

Orange/preprocess/score.py
```diff
@@ -196,3 +203,70 @@ class Gini(ClassificationScorer):
...
+
+        K_NEAREST = self.k_nearest or 50
+        N_ITER = self.n_iterations or 50
```
lanzagar: Same as above.

Orange/preprocess/score.py
kernc: Sure it. I left them in because those conditions weren't set in stone. If there are multiple classes, ReliefF could score the features according to all of them (returning a 2D array), or to the first one, ... I wasn't sure what the best interface was. Did fix it now, though.

Orange/preprocess/score.py
kernc: That's a good question. But it bears a simple answer.

`__init__` never gets called.

PR #600, opened 2015-08-17
OWVennDiagram: Show intersection of multiple datasets

VesnaT:
When using outputs from OWConfusionMatrix as an input to OWVennDiagram and its output as an input to OWDataTable, the intersection of datasets had missing values, eventhough it was possible to obtain them from the input. The corrected method now replaces missing values with suitable data according to 'item_id' and 'source'.


Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -769,6 +769,19 @@ def expanded(feat):
                 newf = expanded_features[source_var][group]
                 instance[newf] = source_inst[source_var]
 
+    group_metas = [(group_name, meta) for group_name in group_names
+                   for meta in newmetas if group_name in meta.name
+                   and meta.is_discrete]
```
lanzagar: Why is this fix applied only to discrete metas?

lanzagar:
Superseded by #685

PR #602, opened 2015-08-18
Contingency for ordinary tables with unknowns

markotoplak:
Therefore owdistributions does not crash anymore. Sparse tables with
unknown values of row variable (usually class) are still not supported.
The implementations could probably be faster but we would need
to modify bottleneck for it.

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1271,6 +1271,7 @@ class Table(MutableSequence, Storage):
             row_data = self._Y[:, row_indi - n_atts]
 
         W = self.W if self.has_weights() else None
+        nan_inds = None
 
         col_desc = [self.domain[var] for var in col_vars]
         col_indi = [self.domain.index(var) for var in col_vars]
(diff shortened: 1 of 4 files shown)


Orange/data/table.py
```diff
@@ -1270,7 +1270,9 @@ def _compute_contingency(self, col_vars=None, row_var=None):
         else:
             row_data = self._Y[:, row_indi - n_atts]
 
+        X = self.X
```
lanzagar: Is this ever used?

Orange/data/table.py
```diff
@@ -1280,10 +1282,13 @@ def _compute_contingency(self, col_vars=None, row_var=None):
             raise ValueError("contingency can be computed only for discrete "
                              "and continuous values")
 
-        if any(var.is_continuous for var in col_desc):
-            if bn.countnans(row_data):
-                raise ValueError("cannot compute contigencies with missing "
-                                 "row data")
+        unknown_rows = bn.countnans(row_data)
+        if unknown_rows:
+            nan_inds = np.isnan(row_data)
```
lanzagar: `np.isnan` will not work if `row_data` is a column from metas that also contained a StringVariable because its dtype will be np.object (isnan needs floats)

PR #604, opened 2015-08-19
DistMatrix: Fix use of DistMatrix

ales-erjavec:


kernc:
Sorry. I, too, hate NumPy breaking the valid-object-evaluates-to-True-in-boolean-contexts contract. But sadly, there's no alternative. :disappointed: 

While you're at it, owmds.py:465 has `self._effective_matrix = Orange.distance.Euclidean(preprocessed_data)` which is on 492 ... used as `self._effective_matrix.X`, which warns of deprecation.

Orange/clustering/hierarchical.py
```diff
@@ -355,13 +355,14 @@ def optimal_leaf_ordering(tree, distances, progress_callback=None):
...
-        DistMatrix that was used to compute the clustering.
+    :param numpy.ndarray distances:
+        A (N, N) numpy.ndarray of distances that were used to compute
+        the clustering.
     :param function progress_callback:
         Function used to report on progress.
 
     """
-    distances = distances.X
+    distances = numpy.asarray(distances)
```
lanzagar: Why is this line needed?
Since distances already are a numpy.ndarray it does nothing, right?

Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -944,8 +944,8 @@ def _update(self):
 
         if distances is not None:
             # Convert to flat upper triangular distances
-            i, j = numpy.triu_indices(distances.X.shape[0], k=1)
-            distances = distances.X[i, j]
+            i, j = numpy.triu_indices(distances.shape[0], k=1)
+            distances = numpy.asarray(distances[i, j])
```
lanzagar: Also here...
(we called distances.shape above so distances really shouldn't be something other than a np.ndarray)

Orange/clustering/hierarchical.py
ales-erjavec: `distances` might be a ndarray  'subclass' (for instance a numpy.matrix). If so then asarray creates an array that is a view of it's contents.

Orange/widgets/unsupervised/owhierarchicalclustering.py
ales-erjavec: Here distances will always be a DistMatrix.  But there is something unnerving about a 'symmetric matrix' with shape (N, ).

Orange/clustering/hierarchical.py
lanzagar: Before the change it was a DistMatrix. Now, let's say we are more general, but even if an exotic subclass is provided, it only needs to support indexing (line 370), then it is again converted to an (contiguous) array.

But the added check does not hurt since it does not do much in normal circumstances.

Orange/widgets/unsupervised/owhierarchicalclustering.py
kernc: It has a shape (N, N). And I foresaw the frequent need of upper triangular, so `DistMatrix.flat()` is also provided. `triu()` might be better, now that I think about it.

ales-erjavec:
```python
D = Orange.misc.DistMatrix(numpy.eye(4))
i, j = numpy.triu_indices(D.shape[0], 1)
print("Shape:", D[i, j].shape)
print(repr(D[i, j]))
```
outputs
```
Shape: (6,)
DistMatrix([ 0.,  0.,  0.,  0.,  0.,  0.])
```

kernc: Sure, but that's because `D[i, j]` is no longer a symmetric matrix; it's a 1D array constucted anew from `i` and `j` indices. It's the same with
```py
>>> A = np.arange(D.size).reshape(D.shape)
>>> A[i, j]
array([ 1,  2,  3,  6,  7, 11])
```
There is no problem, right?

PR #605, opened 2015-08-20
[WIP] owdbscan: the first version of the DBSCAN widget

BlazZupan:
The first version of the new clustering widget that implements DBSCAN algorithm. The widget has data on the input and on the output augments this data set with column for a cluster and a column indicating if the data instance is an essential member of the cluster. Only three distance functions are included, as others, for some unknown reason, proved to be very slow (something to be yet investigated why). The output should be nicely rendered in scatterplot, once this is corrected such that it also shows data with unknown class and the size of the point could be related to a meta variable (a task I asked for Aleš's help yesterday).

### DIFF
--- a/Orange/widgets/unsupervised/owdbscan.py
+++ b/Orange/widgets/unsupervised/owdbscan.py
@@ -8,17 +8,6 @@ from Orange.data import Table, Domain, DiscreteVariable, ContinuousVariable
 from Orange.clustering import DBSCAN
 from Orange import distance
 
-_METRICS = [
-    ("Euclidean", "euclidean"),
-    ("Manhattan", "manhattan"),
-    # ("Cosine", distance.Cosine),
-    ("Jaccard", distance.Jaccard),
-    # ("Spearman", distance.SpearmanR),
-    # ("Spearman absolute", distance.SpearmanRAbsolute),
-    # ("Pearson", distance.PearsonR),
-    # ("Pearson absolute", distance.PearsonRAbsolute),
-]
-
 
 class OWDBSCAN(widget.OWWidget):
     name = "DBSCAN"
@@ -32,6 +21,16 @@ class OWDBSCAN(widget.OWWidget):
 
     OUTPUT_CLASS, OUTPUT_ATTRIBUTE, OUTPUT_META = range(3)
     OUTPUT_METHODS = ("Class", "Feature", "Meta")
+    METRICS = [
+        ("Euclidean", "euclidean"),
+        ("Manhattan", "manhattan"),
+        # ("Cosine", distance.Cosine),
+        ("Jaccard", distance.Jaccard),
+        # ("Spearman", distance.SpearmanR),
+        # ("Spearman absolute", distance.SpearmanRAbsolute),
+        # ("Pearson", distance.PearsonR),
+        # ("Pearson absolute", distance.PearsonRAbsolute),
+    ]
 
     min_samples = Setting(5)
     eps = Setting(0.5)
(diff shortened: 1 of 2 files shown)


Orange/widgets/unsupervised/owdbscan.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def check_data_size(self):
+        if len(self.data) < 2:
+            self.error("Not enough unique data instances" +
+                       "({}).".format(len(self.data)))
+            return False
+        return True
+
+    def commit(self):
+        print("RUN")
```
kernc: This line probably needs to go.

Orange/widgets/unsupervised/owdbscan.py
```diff
@@ -0,0 +1,171 @@
...
+        self.layout().setSizeConstraint(QtGui.QLayout.SetFixedSize)
+
+    def adjustSize(self):
+        self.ensurePolished()
+        self.resize(self.leftWidgetPart.sizeHint())
+
+    def check_data_size(self):
+        if len(self.data) < 2:
+            self.error("Not enough unique data instances" +
+                       "({}).".format(len(self.data)))
```
kernc: You needn't a `+` to concatenate strings like that. It works as in C where neighbouring strings are combined.
```py
self.error("Not enough unique data instances "
           "({})".format(len(self.data)))
```

Orange/widgets/unsupervised/owdbscan.py
```diff
@@ -0,0 +1,171 @@
...
+        if not self.check_data_size():
+            return
+        self.model = DBSCAN(
+            eps=self.eps,
+            min_samples=self.min_samples
+        )(self.data)
+        self.send_data()
+
+    def send_data(self, row=None):
+        print("SEND")
```
kernc: And this.

Orange/widgets/unsupervised/owdbscan.py
lanzagar: Remove debug line.

Orange/widgets/unsupervised/owdbscan.py
lanzagar: Remove debug line.

Orange/widgets/unsupervised/owdbscan.py
```diff
@@ -0,0 +1,171 @@
...
+        box = gui.widgetBox(self.controlArea, "Parameters")
+        gui.spin(box, self, "min_samples", 1, 100, 1, callback=self._invalidate,
+                 label="Core point neighbors")
+        gui.doubleSpin(box, self, "eps", 0.1, 0.9, 0.1,
+                       callback=self._invalidate,
+                       label="Neighborhood distance")
+
+        box = gui.widgetBox(self.controlArea, self.tr("Distance Metric"))
+        gui.comboBox(box, self, "metric_idx",
+                     items=list(zip(*_METRICS))[0],
```
lanzagar: The selected metric is never used

Orange/widgets/unsupervised/owdbscan.py
```diff
@@ -0,0 +1,169 @@
...
+
+        self.send("Annotated Data", new_table)
+
+    def set_data(self, data):
+        self.data = data
+        if self.data is None:
+            self.send("Annotated Data", None)
+        self.commit()
+
+    def _invalidate(self):
```
lanzagar: Why is this method needed?
Calling commit directly instead of calling _invalidate would be the same...

astaric: auto_commit button replaces commit method with its own (one that checks the autocommit flag). If components reference self.commit as a callback, they would perform unconditional commits.

PR #612, opened 2015-08-20
Scatter Plot: Add Features output

lanzagar:


Orange/widgets/visualize/owscatterplot.py
```diff
@@ -348,6 +351,17 @@ def commit(self):
         self.send("Selected Data", selected)
         self.send("Other Data", unselected)
 
+    def send_features(self):
+        features = None
+        if self.attr_x or self.attr_y:
+            dom = Domain([], metas=(StringVariable(name="name"),))
```
janezd: @lanzagar Please rename "name" to "feature". Otherwise OK.

PR #613, opened 2015-08-20
Model based imputer

mstrazar:
Extended model-based imputation to multiple Learners.
- User can specify separate learners for continuous and discrete variables. If not specified, models default to Mean Learner (average / most frequent value). 
- Models now impute only unknown values and not the whole column.
- Fixed a bug for Model based and Random imputation where the *data on input* was overriden by imputation results.


lanzagar:
When making pull requests, please check the diff and commits before actually clicking "Create  pull request"!

lanzagar:
Replaced by #636

PR #616, opened 2015-08-20
Fix rstripped 0 in number of decimals

janezd:
Rstriping was intended for numbers like 4.0, but it also stripped the right zero from 60 (changing it into 6).

### DIFF
--- a/Orange/preprocess/discretize.py
+++ b/Orange/preprocess/discretize.py
@@ -27,8 +27,8 @@ class Discretizer(Transformation):
         assert low is None or high is None or low < high
         assert decimals >= 0
 
-        def fmt_value(value, decimals):
-            return (("%%.%if" % decimals) % value).rstrip("0").rstrip(".")
+        def fmt_value(value, decimals=4):
+            return str(round(value, decimals)).rstrip('.0')
 
         if (low is None or np.isinf(low)) and \
                 not (high is None or np.isinf(high)):


Orange/preprocess/discretize.py
```diff
@@ -28,7 +28,11 @@ def _fmt_interval(low, high, decimals):
         assert decimals >= 0
 
         def fmt_value(value, decimals):
-            return (("%%.%if" % decimals) % value).rstrip("0").rstrip(".")
+            val = ("%%.%if" % decimals) % value
+            if "." in val:
+                val = val.rstrip("0").rstrip(".")
+            return val
```
kernc: what about
```py
def fmt_value(value, decimals):
    return ("%%.%ig" % decimals) % value

fmt_value(5.123123, 3)  # 5.123
fmt_value(5.1, 3)       # 5.1
fmt_value(5.0, 3)       # 5
fmt_value(5, 3)         # 5
```

lanzagar: This is more readable and should produce the same results:
```py
def fmt_value(value, decimals=4):
    return str(round(value, decimals)).rstrip('.0')
```

janezd: @lanzagar You're right. I modified the code.

janezd:
@lanzagar You're too fast. I just discovered that your version doesn't fix the problem that the patch is supposed to fix: str(round(60, 3)).rstrip(".0") returns "6" instead of "60". I'll re-prepare the original patch.

lanzagar:
Argh. It looks like a string, it smells like a string, but it is a set of characters not an array... My bad.
Then I guess we need the if (because nobody wants to bring in the heavy guns and import re).

PR #619, opened 2015-08-20
owselectcolumns: renamed Class to Target Variable in GUI

BlazZupan:
With this we match the notation in Data Table widget, where "target variable" is used. Class in some way implies that this feature is discrete, not continuous, and hence we (Janez and myself) voted for renaming.


lanzagar:
I don't mind either here, but achieving consistency across Orange will be much harder.
A bigger task for another day, and until then I think we will have to accept seeing both terms in various places.

PR #624, opened 2015-08-21
owdistributions: continuous relative frequencies

markotoplak:


markotoplak:
Lan, could you check it now? Thanks!

PR #636, opened 2015-08-21
[WIP] Model based imputer 2

mstrazar:
Extended model-based imputation to multiple Learners.
- User can specify separate learners for continuous and discrete variables. If not specified, models default to Mean Learner (average / most frequent value). 
- Models now impute only unknown values and not the whole column.
- Fixed a bug for Model based and Random imputation where the *data on input* was overriden by imputation results.

### DIFF
--- a/Orange/widgets/data/owimpute.py
+++ b/Orange/widgets/data/owimpute.py
@@ -29,8 +29,6 @@ from Orange.preprocess.transformation import \
 
 from functools import reduce
 import numpy as np
-import warnings
-
 
 
 def _margins(margins, container):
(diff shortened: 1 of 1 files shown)


lanzagar:
Methods for imputing have been moved from the widget to their correct place in the Orange library.
Can you check the state of model imputation there and make sure it can be used in scripts as well as in the widget. This will probably require some refactoring of your changes.

lanzagar:
Also, there are some print calls in the code - they should be avoided. Widgets should show appropriate warnings in the GUI and library calls should generally be silent.

kernc:
And while at it, please put replacement of `numpy` with `np` into a separate commit. You can fix this with:
```bash
git reset HEAD^^  # revert the two commits, obtain changes unstaged and in-tact
git fetch upstream master   # update the biolab's master here called upstream by convention*
git rebase --autostash -i upstream/master  # rebase your branch on this master; resolve conflicts
git add -p  # selectively add changes to a changeset
git commit  # commit this changeset
# repeat git add -p ...
```
* https://help.github.com/articles/syncing-a-fork/

janezd:
@mstrazar Check whether this is still needed or obsolete.

lanzagar:
As discussed things have changed so much that it would make more sense to prepare a new pull request.

PR #638, opened 2015-08-21
Confusion matrix: Cell width adjustments

janezd:


lanzagar:
@janezd please look at #657 and comment (or merge if you agree)

PR #642, opened 2015-08-24
OWClassificationTreeGraph: Fix labels for discrete attributes

VesnaT:


lanzagar:
This code assumes in more than one place that if a variable has compute_value (and compute_variable.variable) it is a continuized discrete variable.

Please test (at least) on:
File [iris] -> Preprocess [Impute, Center and Scale, ..., and combinations] -> Classification Tree -> Classification Tree Viewer

We can discuss how best to approach this before you start fixing the problems...

Orange/widgets/classify/owclassificationtreegraph.py
```diff
@@ -363,10 +366,21 @@ def split_condition(self):
...
         if self.i > 0:
-            sign = [">", "<="][self.tree.children_left[self.parent.i] == self.i]
-            thresh = self.tree.threshold[self.parent.i]
-            return "%s %s" % (
-                sign, self.domain.attributes[self.attribute()].str_val(thresh))
+            attribute = self.domain.attributes[self.attribute()]
+            parent_attr = self.domain.attributes[self.parent.attribute()]
+            parent_attr_cv = parent_attr.compute_value
+            is_left_child = self.tree.children_left[self.parent.i] == self.i
+            if parent_attr_cv and isinstance(parent_attr_cv, Indicator) and \
```
lanzagar: The first check in `if parent_attr_cv and isinstance(parent_attr_cv, Indicator)` is unnecessary and included in the second part (None is not an instance of Indicator)

(similarly below)

Orange/widgets/classify/owclassificationtreegraph.py
```diff
@@ -363,10 +366,21 @@ def split_condition(self):
...
+                    hasattr(parent_attr_cv.variable, "values"):
+                values = parent_attr_cv.variable.values
+                return values[abs(parent_attr_cv.value - is_left_child)] \
+                    if len(values) == 2 \
+                    else "%s %s" % ("≠" * is_left_child,
```
lanzagar: This results in a space before the value when not is_left_child (probably unintended?)
`else "≠ " * is_left_child + values[parent_attr_cv.value]`
is also shorter if you want to multiply strings and booleans instead of using if

PR #643, opened 2015-08-24
Various fixups

kernc:
Mergeable at will.

### DIFF
--- a/Orange/widgets/utils/colorpalette.py
+++ b/Orange/widgets/utils/colorpalette.py
@@ -2,6 +2,7 @@ import os
 import sys
 import math
 from numbers import Number
+from collections import Iterable
 
 import numpy as np
 from PyQt4 import QtCore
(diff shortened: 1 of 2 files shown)


lanzagar:
This is not _utter_ shit, but the pull request title and description are not very informative.
Short instructions on how to reproduce the bugs and test these fixes would be welcome.

kernc:
Thanks. Have amended the commits with some explanation.

Orange/widgets/utils/colorpalette.py
```diff
@@ -597,66 +597,52 @@ def __getitem__(self, val):
 
 
 class ColorPaletteGenerator:
-    maxHueVal = 260
 
-    def __init__(self, number_of_colors=None, rgb_colors=DefaultRGBColors):
+    def __init__(self, number_of_colors=0, colors=DefaultRGBColors):
```
lanzagar: Changing the name of a constructor parameter for a class that is (in)directly used in most graphical widgets usually requires some refactoring.
I am guessing you didn't use the Find Usages feature of PyCharm (or grep) or tried some standard Orange widgets after the changes? At least the Scatter Plot and Data Table widgets crash and probably some others as well...

kernc: Aye, I didn't. :flushed: Sorry to have put it on you like that. Thanks for reminding me so gently. Will surely strive to be exhaustive in the future.

Have reverted the parameter's name change and the mentioned widgets now work. Have also reconsidered the changed default value of the other parameter, but it doesn't seem to have an ill effect anywhere.

Would you recommend I revert all but the minimal required* changes (i.e. handling NaNs) or is the patch, in general, ok?

\* It's not really required. I could handle NaNs manually in my widget.

lanzagar: I am very much in favour of improving the code beyond the minimum required changes in general. If possible/applicable the change and other improvements are better in separate commits.
Just test both, the change and other improvements :)
In our cases the author usually knows the implications of his pull request better than the one merging it, so don't count on somebody else to find errors.

The patch looks OK at first glance, finish with testing and changes if any more are planned and I'll take a closer look.

Orange/widgets/utils/colorpalette.py
```diff
@@ -597,66 +597,52 @@ def __getitem__(self, val):
...
-        else:
-            return self.rgb_colors[int(index)]
-
-    # noinspection PyPep8Naming
-    def getRGBA(self, index):
-        if isinstance(index, np.ndarray):
-            return self.rgba_array[index.astype(int)]
+        if isinstance(index, Number):
+            return QColor(*self.getRGB(index))
+        return [QColor(*c) for c in self.getRGB(index)]
```
lanzagar: The old version always returned a single QColor for a given index.
When extending functionality I find it better to leave the previous way the default one.

In this case that means check for the new special case first
`if isinstance(index, Iterable): return [...]`
and in all other cases retain the previous functionality.

In your case when index is not a Number and not iterable it crashes in a new way...

kernc: That's a very good advice indeed!

PR #645, opened 2015-08-25
[WIP] owdistributions: probabilities

markotoplak:


lanzagar:
The second commit (from Janez) changes the symbol to "x" for probabilities and then the third commit ignores it and changes it back to "o"...
Intentional? -> remove second commit
Or not? -> fix

janezd:
@lanzagar I added confidence intervals, and "o" looks better. Consider this WIP. Right now I'm trying some other changes.

lanzagar:
OK.
If you are changing things - I was saving this for a discussion tomorrow, but I definitely do not like (probability) confidence intervals that extend below 0...
And maybe the whole idea of having confidence intervals (and probabilities?) is unnecessarily complicating the widget?

janezd:
Probabilities are useful. We can discuss confidence intervals, though. :)

PR #650, opened 2015-08-26
OWMds: Change selection tool shape to rectangle

VesnaT:


lanzagar:
We should probably make sure (in a separate pull request) that all visualization which allow selections do it in the same way:
- rectangular selection (at least by default)
- holding Ctrl adds to selection (currently does not work in scatterplot)
- consistent output signal names (I prefer "Selected Data" to "Data Subset")
- do we also output unselected "Other Data"?

PR #655, opened 2015-08-27
[WIP] OWTable: remove color palette selection

kernc:
The user doesn't select their color palette. If they do, they may have to re-do it for every widget that supports it. They probably don't.

If _visualize continuous values_ and _color by instance classes_ were true, the widget crashed on data tables with `len(class_var.values) > len(colors in palette)`. An example is audiology.tab dataset with a new continuous feature constructed (e.g. `sqrt(speech) + sqrt(air) + sqrt(ar_u)`).

This patch prevents user's choice and solves the crash by setting the perfect color palette for each dataset.

Allowing to opt for a particular color palette, as a pure convenience to the user, should be done globally and elsewhere.

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -25,8 +25,8 @@ from Orange.statistics import basic_stats
 
 from Orange.widgets import widget, gui
 from Orange.widgets.settings import Setting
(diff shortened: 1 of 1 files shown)


janezd:
Depends on #753

janezd:
Included in PR #753

PR #659, opened 2015-08-28
OWScatterplotgraph: Change select/deselect keys

VesnaT:
Change select/deselect keys to be consistent with other widgets.


lanzagar:
Are other widgets really all consistent with this behaviour?
Please check that the same 3 modifiers are used for same effects in others as well.

VesnaT:
No, they are not. I'm changing the commit message.

PR #669, opened 2015-08-28
Test learners error report

ales-erjavec:


Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -222,16 +222,17 @@ def set_learner(self, learner, key):
...
-
-        self._update_stats_model()
+            self._invalidate([key])
 
     def set_train_data(self, data):
         """
         Set the input training dataset.
         """
-        self.error(0)
+        self.error([0, 1, 2, 3])
```
lanzagar: This clears all errors and warnings even if they are still relevant.
E.g. test data has no class so error 1 is shown, then train data is updated and the error disappears (but should still be shown)

lanzagar:
So we are not making new pull requests:
- could you also change Model to Method in line 208 (so it is the same as in line 378)
- handleNewSignals calls _update_header() twice: once explicitly and again in the first line of apply

ales-erjavec:
Ready for another review.

PR #675, opened 2015-08-31
Tests for TestOnTestData

janezd:


janezd:
@lanzagar @astaric I though I'd add something to this one so I changed it to WIP. I forgot what I wanted to add, so I removed WIP. Then I repeated the cycle again. Please merge, so I don't repeat it again. :)

PR #676, opened 2015-08-31
LinearProjection: Density image improvements.

thocevar:


kernc:
You sure this belongs in `widgets.utils`? Is it generally useful when constructing visualization widgets? If so, the usage seems severely underdocumented.

thocevar:
Class density image is used as a background in ScatterPlot and LinearProjection - `widgets.utils` seemed like an appropriate place. There probably won't be a third use case for this.

PR #678, opened 2015-08-31
OWGoogleSheets: new data source

kernc:
@BlazZupan and @marinkaz might also want to review this. Should eventually work with https://github.com/biolab/orange3/pull/524.

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1,4 +1,5 @@
 import os
+import re
 import zlib
 from collections import MutableSequence, Iterable, Sequence, Sized
 from itertools import chain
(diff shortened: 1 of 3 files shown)


marinkaz:
I tested the functionality on a few google sheets. Somehow the first two data rows are always lost and never appear in the downloaded data. @kernc, can you take a look into the issue?

kernc:
Well, that's improperly parsed TSV, something fixed in #524. Do have a look at that if you will. :smile_cat: 

Google just "refreshed" their logo and the 'G'. Anything about it?

kernc:
With the newest changes, this is now blocked on #524.

PR #679, opened 2015-08-31
Rank: Add ReliefF, refactoring

thocevar:


Orange/widgets/data/owrank.py
```diff
@@ -23,47 +23,19 @@ def table(shape, fill=None):
...
-        handles_discrete=True),
-    score_meta(
-        "Gini Gain", "Gini", score.Gini,
-        supports_regression=False,
-        supports_classification=True,
-        handles_continuous=False),
+    score_meta("Information Gain", "Inf. gain", score.InfoGain),
+    score_meta("Gain Ratio", "Gain Ratio", score.GainRatio),
+    score_meta("Gini Gain", "Gini", score.Gini),
+    score_meta("ReliefF", "ReliefF", score.ReliefF),
```
kernc: This looks great! But `RReliefF` is missing, I think.

thocevar: There are also some scoring methods from sklearn (chi2, anova, ...) missing. I'm not sure which do we want in the widget, probably not all of them.

kernc: I agree, only the real-time ones.

PR #684, opened 2015-09-02
Testing: ShuffleSplit replaces Bootstrap

VesnaT:
ShuffleSplit replaces Bootstrap, because Bootstrap will be removed in 0.17 version of Scikit-learn.

### DIFF
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -2,11 +2,10 @@ import numpy as np
 
 import sklearn.cross_validation as skl_cross_validation
 
-import Orange.data
-from Orange.data import Domain, Table
+from Orange.data import Table
 
 __all__ = ["Results", "CrossValidation", "LeaveOneOut", "TestOnTrainingData",
-           "Bootstrap", "TestOnTestData", "sample"]
+           "ShuffleSplit", "TestOnTestData", "sample"]
 
 
 class Results:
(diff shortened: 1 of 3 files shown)


Orange/evaluation/testing.py
```diff
@@ -366,21 +365,21 @@ def __init__(self, data, learners, store_data=False, store_models=False,
         self.call_callback(1)
 
 
-class Bootstrap(Results):
```
kernc: Would you consider wrapping the new interface in the old, deprecated? Something like
```py
@deprecated
class Bootstrap(Results):
    def __init__(...):
        ... ShuffleSplit(...) ...
```

Orange/tests/test_evaluation_testing.py
```diff
@@ -460,3 +460,14 @@ def test_fixed_training_size(self):
...
+
+class TestShuffleSplit(unittest.TestCase):
+    def test_results(self):
+        nrows, ncols = 100, 10
+        data = random_data(nrows, ncols)
+        train_size, n_resamples = 0.6, 10
+        res = ShuffleSplit(data, [NaiveBayesLearner()], p=train_size,
+                           n_resamples=n_resamples)
+        self.assertEqual(len(res.predicted[0]),
+                         n_resamples * (nrows * (1 - train_size)))
```
lanzagar: a * b * (1 - c) is more readable than a * (b * (1 - c))

Orange/evaluation/testing.py
lanzagar: I wouldn't since this is not a change of the interface, but actually removal of Bootstrap and introduction of ShuffleSplit. They are similar enough that most of the code can be reused, but the method is different.

And since the class was used only in one place (and even there incorrectly), I don't think we need to bother with deprecating it first...

Orange/evaluation/testing.py
```diff
@@ -366,21 +365,21 @@ def __init__(self, data, learners, store_data=False, store_models=False,
...
-    def __init__(self, data, learners, n_resamples=10, p=0.75, random_state=0,
-                 store_data=False, store_models=False, preprocessor=None,
-                 callback=None):
+class ShuffleSplit(Results):
+    def __init__(self, data, learners, n_resamples=10, p=0.75, test_size=None,
+                 random_state=0, store_data=False, store_models=False,
```
lanzagar: The original scikit ShuffleSplit has train_size and test_size which can be fractions or int sizes, and either one of them or both can be provided. Our wrapper makes a mess of this by having parameters p and test_size with different defaults (providing only test_size uses p=0.75 instead of setting it to the complement).

I vote for sticking to scikit parameter names and defaults.

VesnaT: Test_size is set to None to be the complement of train_size value. If we stick to scikit default, we should probabla add test_size slider to the widget OWTestLearners? Or calculate the complement there and pass it to the wrapper?

lanzagar: I think we can leave the widget as it is.
I was referring to the class constructor.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -326,11 +326,12 @@ def update_progress(finished):
...
                 self.data, learners, **common_args)
-        elif self.resampling == OWTestLearners.Bootstrap:
-            results = Orange.evaluation.Bootstrap(
+        elif self.resampling == OWTestLearners.ShuffleSplit:
+            train_size = self.sample_p / 100
+            results = Orange.evaluation.ShuffleSplit(
                 self.data, learners, n_resamples=self.n_repeat,
-                p=self.sample_p / 100, random_state=rstate,
-                **common_args)
+                train_size=train_size, test_size=1 - train_size,
```
lanzagar: This sometimes produces an error saying that test and train data sizes sum to num_instances + 1. Which should probably be fixed in scikit-learn, but in the meantime using the following works and is nicer anyway:
`test_size=None`

PR #685, opened 2015-09-02
OWVennDiagram: Show intersection of multiple datasets (2)

lanzagar:
Alternative approach to #600

@VesnaT please check

### DIFF
--- a/Orange/widgets/visualize/owvenndiagram.py
+++ b/Orange/widgets/visualize/owvenndiagram.py
@@ -52,7 +52,7 @@ class OWVennDiagram(widget.OWWidget):
     inputhints = settings.Setting({})
     #: Use identifier columns for instance matching
     useidentifiers = settings.Setting(True)
-    autocommit = settings.Setting(False)
+    autocommit = settings.Setting(True)
 
     want_graph = True
 
--- a/Orange/widgets/visualize/owvenndiagram.py
+++ b/Orange/widgets/visualize/owvenndiagram.py
@@ -529,8 +529,10 @@ class OWVennDiagram(widget.OWWidget):
 
             # add columns with source table id and set id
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -762,6 +764,14 @@ def expanded(feat):
         indices = inst_by_id[inst_id]
         instance = newtable[i]
 
+        for var in domain.variables + domain.metas:
+            if var in idvarlist:
+                continue
+            if numpy.isnan(instance[var]):
+                for ind in indices:
+                    if not numpy.isnan(table[ind, var]):
```
VesnaT: I'm getting an error on this line when changing from Use identifiers to Use instance equality.
ValueError: 'name (Naive Bayes)' is not in domain

Reproduce the error: Three datasets on the input of Venn diagram. Select "Use identifiers". Select a subset that is intersection of only two datasets. Select "Use instance equality". Change selected subset.

Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -762,6 +764,14 @@ def expanded(feat):
         indices = inst_by_id[inst_id]
         instance = newtable[i]
 
+        for var in domain.variables + domain.metas:
+            if var in idvarlist:
+                continue
+            if numpy.isnan(instance[var]):
+                for ind in indices:
```
kernc: Why loop so explicitly? Doesn't table handle advanced indexing and broadcasting?

Orange/widgets/visualize/owvenndiagram.py
lanzagar: Should be fixed in the updated version.

Expanded features were already handled below the new for loop, so they should be skipped here.

PR #686, opened 2015-09-03
Score: Remove zero columns

VesnaT:
When the data domain is not fully represented (i.e. no appearances of some value of Discrete attribute), there are zero columns in contingency array. These columns are now removed to be able to devide by np.sum(D, axis=0) when calculating entropy.

### DIFF
--- a/Orange/preprocess/score.py
+++ b/Orange/preprocess/score.py
@@ -194,6 +194,8 @@ class GainRatio(ClassificationScorer):
         h_class = _entropy(np.sum(cont, axis=1))
         h_residual = _entropy(np.compress(np.sum(cont, axis=0), cont, axis=1))
         h_attribute = _entropy(np.sum(cont, axis=0))
+        if h_attribute == 0:
+            h_attribute = 1
         return nan_adjustment * (h_class - h_residual) / h_attribute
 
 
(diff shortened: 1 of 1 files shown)


kernc:
Wouldn't you like to use this fine example to practice [squashing](https://www.google.com/search?q=squash+commits). :)

VesnaT:
Why squashing? Those three were intentionally made :P

kernc:
But they're all a single line and very related? Even the commit message is outright copied over??? :confused:

janezd:
@kernc @VesnaT @lanzagar @astaric @markotoplak @BlazZupan  Having the rules and following them is important to reduce the mess that would otherwise cause us to lose track of what we're doing, decrease our productivity and introduce bugs out of negligence. I'm all for rules.

On the other hand, I've recently got way too many comments about commit messages, about how pull requests should be organized, and so forth. I am not expressing my opinion about this particular case, but in general we are either totally incompetent w.r.t. the use of git, or we are totally nitpicking. Or perhaps both. :)

I appreciate that we became so better organized in comparison with the mess we used to have. I just wouldn't want us to go over the edge, so that we spend more time debating pull requests than making them.

kernc:
Pull requests are a mechanism of code review process and in healthy group projects are meant, encouraged even, to be debated on.

PR #689, opened 2015-09-04
OWWidget geometry restore fix

ales-erjavec:
Restore the widget geometry only on first show event. This should fix a problem on windows where restoring a widget from a minimized state would resize it to a less then perfect shape.


ales-erjavec:
@thocevar can you double check that this works

thocevar:
It works!

PR #690, opened 2015-09-04
Impute: Fix imputation of discrete features with most frequent values.

thocevar:


Orange/widgets/data/owimpute.py
```diff
@@ -563,8 +563,13 @@ def column_imputer_minimal(variable, table):
 
 
 def column_imputer_average(variable, table):
-    stats = basic_stats.BasicStats(table, variable)
-    return column_imputer_defaults(variable, table, stats.mean)
+    if variable.is_continuous:
```
lanzagar: The function is called `column_imputer_average` so I think it should do that and not switch to something else for discrete variables.

The function below was obviously intended for the functionality of the else clause, but is currently broken (no return) and not used.
I think this should be fixed instead.
E.g. have the if in `imputer_for_column` and call either `column_imputer_average` (for cont) or `column_imputer_modus` (for disc).

Orange/preprocess/preprocess.py
```diff
@@ -154,13 +155,21 @@ def __call__(self, data):
         if not self.force and not np.isnan(data.X).any():
             return data
         self.imputer = skl_preprocessing.Imputer(strategy=self.strategy)
-        X = self.imputer.fit_transform(data.X)
+        self.imputer.fit(data.X)
+        self.fix_statistics_for_discrete_vars(data)
```
lanzagar: This class is only a wrapper for the scikit imputer and I don't like "correcting" the imputed values based on var types. The `strategy` parameter is passed to scikit and the result should be the same as in scikit.

If we want something more useful for Orange, I suggest we add an additional 'smart' strategy (can even be default) which uses 'mean' on continuous vars and 'most_frequent' on discrete.

PR #692, opened 2015-09-07
MDS Widget: Add a missing call to set_number_of_colors for discrete p…

janezd:
…alette

MDS did not work when the number of classes was too large since the palette was too small.


Orange/widgets/unsupervised/owmds.py
```diff
@@ -1081,6 +1081,7 @@ def color_data(table, var=None, mask=None, plotstyle=None):
             col = mdsplotutils.column_data(table, var, mask)
             if var.is_discrete:
                 palette = plotstyle.discrete_palette
+                palette.set_number_of_colors(len(var.values))
```
ales-erjavec: This palette is a global shared instance. It should not be modified.

ales-erjavec:
An alternative approach:  #695

PR #693, opened 2015-09-07
official addons: Add Orange3-Network

astaric:
Add Orange3-Network to the list of add-ons that are shown in the add-ons dialog, even if pypi does not return them.

This list will not be needed once

https://bitbucket.org/pypa/pypi/issues/325/searching-by-keywords-stopped-working

is resolved.


astaric:
Related to https://github.com/biolab/orange3-prototypes/pull/10

PR #694, opened 2015-09-07
OWMds, OWLinearProjection, OWHeatMap: Selection

VesnaT:


lanzagar:
MDS and Linear Projection work fine.
In Heat Map shift and alt can be dragged over multiple rows, while control selections can not (works only on the row where mouse button was pressed down). Can you make control behave the same as the other two?

Also please change the name of MDS output from Data Subset to Selected Data so it is consistent with others.

PR #698, opened 2015-09-08
OWMosaic: Selection

VesnaT:


lanzagar:
Selections are now OK but they still do not do anything - Selected Data output is always empty.
Is that functionality planned for a different pull request?

VesnaT:
Yes. It will be done in another pull request.

PR #699, opened 2015-09-08
OWHierarchicalClustering: Selection

VesnaT:

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -387,6 +387,9 @@ class DendrogramWidget(QGraphicsWidget):
     def is_selected(self, item):
         return item in self._selection
 
+    def is_included(self, item):
+        return self._selected_super_item(item) is not None
+
     def select_item(self, item, state):
         """Set the `item`s selection state to `select_state`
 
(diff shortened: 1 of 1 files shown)


lanzagar:
Shift selecting should never result in less instances being selected (shift only adds). Here, when a subset of a larger selection is shift-clicked the rest is deselected.

janezd:
@lanzagar Not true, at least not in general. If you click within selection, Shift reduces it.

lanzagar:
@janezd Which selection are you referring to? And how does it reduce it:
a) Deselects the clicked object - then it would be the same as Ctrl
b) Deselects everything else and only keeps the clicked object - same as no modifier
c) The current implementation in this PR actually does neither of the above. It leaves the other clusters as they are, deselects the parent cluster of the newly clicked subset and adds only the subset. I find it weird and not consistent with Scatter Plot and other point selection widgets.

Of course what I said in the previous comment is not true in every software. But for the new selection shortcuts that we (@VesnaT) are trying to consistently use in all Orange widgets Shift is used to add to the selection (and Alt to remove from, and Ctrl to add/remove = change).

janezd:
OK, I agree, hierarchical clustering is really different. On the other hand, for hierarchical clustering it does not really matter since we usually select a single cluster - I don't see many cases where you'd like to select different parts of dendrograms. So anything that you find consistent is OK for me.

PR #701, opened 2015-09-08
settings: Simplify ClassValuesContextHandler

astaric:
Pull methods that can be used in other context handlers to
ContextHandler base class.

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -4,6 +4,7 @@ import copy
 import itertools
 import pickle
 import warnings
+from Orange.util import abstract
 
 from Orange.misc.environ import widget_settings_dir
 from Orange.data import Domain, Variable
@@ -395,7 +396,10 @@ class Context:
 
 
 class ContextHandler(SettingsHandler):
-    """Base class for setting handlers that can handle contexts."""
+    """Base class for setting handlers that can handle contexts.
+
+    Classes deriving from it need to implement method `match`.
+    """
 
     MAX_SAVED_CONTEXTS = 50
 
@@ -466,6 +470,7 @@ class ContextHandler(SettingsHandler):
         else:
             self.settings_to_widget(widget)
 
+    @abstract
     def match(self, context, *args):
         """Return the degree to which the stored `context` matches the data
         passed in additional arguments). A match of 0 zero indicates that
@@ -476,7 +481,7 @@ class ContextHandler(SettingsHandler):
         return 0 and 2.
 
         Derived classes must overload this method."""
-        raise TypeError(self.__class__.__name__ + " does not overload match")
+        pass
 
     def find_or_create_context(self, widget, *args):
         """Find the best matching context or create a new one if nothing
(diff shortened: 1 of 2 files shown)


Orange/widgets/settings.py
```diff
@@ -547,6 +547,11 @@ def close_context(self, widget):
     def settings_to_widget(self, widget):
         widget.retrieveSpecificSettings()
 
+        context = widget.current_context
+        if context is None:
+            return
+        self.provider.unpack(widget, context.values)
```
lanzagar:
```py
if context is not None:
    self.provider.unpack(widget, context.values)
```

kernc:
Would it make sense to add some more docstrings and usage instructions?

Orange/widgets/settings.py
```diff
@@ -565,6 +570,18 @@ def packer(setting, instance):
...
 
+    def fast_save(self, widget, name, value):
+        super().fast_save(widget, name, value)
+
+        context = widget.current_context
+        if context is None:
+            return
+
+        if name in self.known_settings:
+            setting = self.known_settings(name)
```
lanzagar: `self.known_settings[name]`

astaric:
I have added docstrings for methods that did not have them.

As for the usage instructions, this is an abstract class with one abstract method. You need to create a subclass and provide the implementation of the match method and you are good to go.

kernc:
Please excuse my seeming brain-rot (:flushed:), but I don't understand it, nor do I see code reflecting your statement. Which class is abstract? Which method is abstract? How do the things fit together? And why explain it to me over and over again and not also to countless Indians eager to have their contributions merged? If you dislike lengthy class or module docstrings, `doc/development/settings.rst` seems appropriate, though former can easily build the latter.

janezd:
I don't really care about usage instructions here.

This PR has a different problem.

Open File -> Scatterplot. If you already have some saved settings, Scatterplot will fail with something like

    unable to set <PyQt4.QtGui.QComboBox object at 0x1154fff78> to value '('aquatic', 1)'.
    Possible values are hair, feathers, eggs, milk, airborne, aquatic, predator, toothed,
    backbone, breathes, venomous, fins, legs, tail, domestic, catsize, type.

If you change the values of the combo boxes or at least touch them, scatter plot starts working.

Now change the file. Every time you do so, scatter plot fail and the tuple accumulates. In the end, you can get something like

    File "/Users/janezdemsar/Dropbox/orange3/Orange/widgets/visualize/owscatterplotgraph.py",
    line 515, in update_data
        index_y = self.attribute_name_index[attr_y]
    KeyError: (((('aquatic', 1), -2), -2), -2)
    Pending nodes: set()
    Blocking nodes: set()
    unable to set <PyQt4.QtGui.QComboBox object at 0x1154fff78> to value '(((('aquatic', 1), -2), -2), -2)'.
    Possible values are hair, feathers, eggs, milk, airborne, aquatic, predator, toothed, backbone,
    breathes, venomous, fins, legs, tail, domestic, catsize, type.

I don't have time to dig into it, but apparently somewhere two values are packed in a tuple and not unpacked at retrieval.

janezd:
I haven't really explored the code. Is it problematic in principle or is it just a bug and you're going to do it differently?

PR #707, opened 2015-09-10
(R)ReliefF fix

kernc:
* Fix two major bugs with (R)ReliefF:
  1. prepare() modified original data,
  2. contingencies `std::map` didn't work correctly
     (contingency table's memoryview wasn't copied into the map container correctly. Do prefer `std::map::insert()` to `std::map::operator[]`, which doesn't seem to work as well with cython),
  3. better handle missing data.
* ~~Increase number of iterations to 100 for more stable results.~~


astaric:
Can you add tests that confirm that mentioned bugs are really fixed?

kernc:
Added.

PR #708, opened 2015-09-10
k-Means widget: Fix the signal name in send

janezd:
There was a call self.send("Data", None) instead of self.send("Annotated Data", None)

### DIFF
--- a/Orange/widgets/unsupervised/owkmeans.py
+++ b/Orange/widgets/unsupervised/owkmeans.py
@@ -315,7 +315,7 @@ class OWKMeans(widget.OWWidget):
         else:
             km = self.km
         if not self.data or not km:
-            self.send("Data", None)
+            self.send("Annotated Data", None)
             self.send("Centroids", None)
             return
 


Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -315,7 +315,7 @@ def send_data(self, row=None):
         else:
             km = self.km
         if not self.data or not km:
-            self.send("Data", None)
```
kernc: Avoid errors like that by not [stringly typing](http://c2.com/cgi/wiki?StringlyTyped) and instead using [simple `enum`s](https://github.com/biolab/orange3-text/blob/master/orangecontrib/text/widgets/owlda.py#L14). Python doesn't catch that invalid class member at its compile time, but some advanced IDEs might.

janezd: In general I support putting string into constants when their content in several places must match exactly. But here they would decrease readability, and we would only get some kind of testing in some IDEs.

I prefer testing. Simple automated testing of widgets will only find superficial errors (well, including this one). Manually testing each widgets after implementing them will catch more. It is not as good as unit tests since you can re-run them, but until we know how to make them, we need to thoroughly test each widget manually!

Anyway, instead of using constants, I created a test (#709). Ironically, this test would be more difficult to implement if we had constants.

PR #709, opened 2015-09-11
Add tests for declarations of widget outputs

janezd:
A tests that parses all widget files for `self.send("<signal name>"` and checks whether the <signal name> is defined as an output.

The test uses regular expression which requires that this is the first non-white thing in the line; this excludes sends that are commented out. There may be false positives (e.g. commented within triple quotes) or false negatives (if sends are written in a more contrived way). The former are easy to fix and the latter do not and should not appear. The test is thus sufficient.

The test takes some time since it requires the widget registry to scan for all widget files (and even import them, if run on Travis).

### DIFF
--- a/Orange/tests/__init__.py
+++ b/Orange/tests/__init__.py
@@ -4,7 +4,7 @@ import unittest
 from Orange.widgets.tests import test_setting_provider, \
     test_settings_handler, test_context_handler, \
     test_class_values_context_handler, test_domain_context_handler, \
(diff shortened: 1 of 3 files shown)


janezd:
Blocked by #708 , which fixes a bug that this test detects.

kernc:
> and the latter do not and should not appear.

Do you propose this "best practice" be extended to and enforced in "official" add-ons?

PR #710, opened 2015-09-11
Chdir when loading schema; OWFile checks for recent files in cwd

janezd:
If we save a schema and some files, reloading schema on another machine does not work since absolute paths to the files change. This cannot be solved in general, but file widget can check whether the missing files are in the same directory as the schema.

Instead of having a global variable with the schema directory, I suggest we change the current directory when loading schema. @ales-erjavec , is this OK?

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -46,9 +46,8 @@ class OWFile(widget.OWWidget):
     def __init__(self):
         super().__init__()
         self.domain = None
-        self.recent_files = [fn for fn in self.recent_files
-                             if os.path.exists(fn)]
         self.loaded_file = ""
+        self._relocate_recent_files()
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owfile.py
```diff
@@ -86,6 +85,17 @@ def __init__(self):
         if len(self.recent_files) > 0:
             self.open_file(self.recent_files[0])
 
+    def _relocate_recent_files(self):
+        rec = []
+        for fn in self.recent_files:
+            if os.path.exists(fn):
+                rec.append(fn)
+            else:
+                name = os.path.split(fn)[1]
```
astaric: Should we also check datasets included with Orange?

https://github.com/biolab/orange3/blob/master/Orange/data/table.py#L31

janezd:
The file widget already tries the current working directory while loading. The problem is that the directory is not changed when loading schema, and that `__init__` filters out non-existing files by absolute paths, without considering current directory.

PR #711, opened 2015-09-11
Change working dir on schema load and fix file widget

janezd:
A better alternative for #710 because it keeps absolute paths.

The problem is: If we save a schema and some files, reloading schema on another machine does not work since absolute paths to the files change. The problem was that non-existing files are filtered out, although when it opens the file it also looks for them in the working directory.

I suggest we change the current directory when loading schema. @ales-erjavec , is this OK?

### DIFF
--- a/Orange/canvas/application/canvasmain.py
+++ b/Orange/canvas/application/canvasmain.py
@@ -938,6 +938,7 @@ class CanvasMainWindow(QMainWindow):
         dirname = os.path.dirname(filename)
 
         self.last_scheme_dir = dirname
+        os.chdir(dirname)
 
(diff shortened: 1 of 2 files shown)


ales-erjavec:
No. We really can/should not change the working directory.

I propose to pass the workflow base dir explicitly through the workflow model to the widget instance.
See ales-erjavec/orange3@0f578ef4621fea2e28190eb08679c890189e924a for a preliminary implementation

janezd:
@ales-erjavec, thanks for looking into this. If you have time, please finish gh-716 soon since we need this for an experiment with students on BCM.

PR #712, opened 2015-09-11
Distributions: set maximal width of control area

janezd:
Distributions widget's combo has the width of the largest class (try, for instance, GDS4182). It can expand controlArea by too much. I now set a reasonable maximum.

@ales-erjavec, this probably happens in other widgets, too. What about a class attribute `maximal_control_area_width` (or something shorter), and then `OWWidget` would do something like

    if self.maximal_control_area_width is not None:
        self.controlArea.setMaximumWidth(self.maximal_control_area_width)

We could also have a class attribute `control_area_width` for setting a fixed size.

### DIFF
--- a/Orange/widgets/visualize/owdistributions.py
+++ b/Orange/widgets/visualize/owdistributions.py
@@ -191,6 +191,7 @@ class OWDistributions(widget.OWWidget):
         self._legend.hide()
         self._legend.anchor((1, 0), (1, 0))
         self.graphButton.clicked.connect(self.save_graph)
+        self.controlArea.setMaximumWidth(260)
 
     def update_views(self):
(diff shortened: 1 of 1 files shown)


janezd:
See #714 for a more general solution.

ales-erjavec:
All combo boxes which do not have a known set of items should use 
```python
cb.setSizeAdjustPolicy(QComboBox.AdjustToMinimumContentsLengthWithIcon)
cb.setMinimumContentsLength(12)
```
Setting a fixed size on the control area will (more often then not) lead to various labels, check boxes, ... to be clipped when using styles other then the one the developer used.

kernc:
> All combo boxes which do not have a known set of items should use 

Can this be set default in `gui.comboBox()`?

janezd:
#717 provides a better general solution.

PR #714, opened 2015-09-12
[WIP] Control the control area width in OWWidget

janezd:
A proposal for limiting the size of the control area.

We need something like this in general, to prevent variables with long names or values from expanding the control area. If we apply this patch, we would need to check all widgets. For instance, this had problems with k-means that have mainArea, but hide when it is not needed.

This approach is good since in cases when we don't want to limit the size, the widget will always have a bad layout and we can spot is just by opening it - so we can fix it. With what we have now (that is, no control), widgets usually look good and the problem appears only in weird data sets, so the problem went unnoticed so far for most widgets.


janezd:
May truncate other components on machines with different visual settings. #717 is safer.

PR #716, opened 2015-09-14
OWFile relocatable recent file paths

ales-erjavec:
Store/restore recent file paths based on a set of predefined known/fixed
search paths (including loaded workflow scheme relative path if available).

### DIFF
--- a/Orange/canvas/application/canvasmain.py
+++ b/Orange/canvas/application/canvasmain.py
@@ -953,7 +953,8 @@ class CanvasMainWindow(QMainWindow):
         from a saved `filename`. Return `None` if an error occurs.
 
         """
-        new_scheme = widgetsscheme.WidgetsScheme(parent=self)
+        new_scheme = widgetsscheme.WidgetsScheme(
+            parent=self, basedir=os.path.dirname(filename))
         errors = []
         try:
             scheme_load(new_scheme, open(filename, "rb"),
--- a/Orange/canvas/application/canvasmain.py
+++ b/Orange/canvas/application/canvasmain.py
@@ -954,7 +954,7 @@ class CanvasMainWindow(QMainWindow):
 
         """
         new_scheme = widgetsscheme.WidgetsScheme(
-            parent=self, basedir=os.path.dirname(filename))
+            parent=self, env={"basedir": os.path.dirname(filename)})
         errors = []
         try:
             scheme_load(new_scheme, open(filename, "rb"),
@@ -1175,6 +1175,7 @@ class CanvasMainWindow(QMainWindow):
         try:
             with open(filename, "wb") as f:
                 f.write(buffer.getvalue())
+            scheme.set_runtime_env("basedir", os.path.dirname(filename))
             return True
         except (IOError, OSError) as ex:
             log.error("%s saving '%s'", type(ex).__name__, filename,
--- a/Orange/canvas/scheme/scheme.py
+++ b/Orange/canvas/scheme/scheme.py
@@ -6,7 +6,7 @@ Scheme Workflow
 The :class:`Scheme` class defines a DAG (Directed Acyclic Graph) workflow.
 
 """
-
+import types
 from operator import itemgetter
 from collections import deque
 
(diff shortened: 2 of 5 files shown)


ales-erjavec:
@janezd: Can you test this by downloading and extracting: https://www.dropbox.com/s/bujhb0yfbkqg8xw/test.zip?dl=0
It contains a .ows file and a datasets directory. When opening the workflow (using this branch) it should resolve the recent paths to include the files in the datasets subdirectory.

Other wise if there are no complaints, this is ready for merging.

ales-erjavec:
One problem: creating a workflow on windows and opening it on UNIX-like system does not restore the items, although the reverse does work.

janezd:
@ales-erjavec, tested on OsX: works like Charm. Great. Could you take care of Windows, too? As I understand, students will exchange .ows files, and we have a mixture of Windows and Os Xs.

ales-erjavec:
Now also relocates between windows/posix systems.
Note: in order to create a relocatable workflow it is first necessary to save a .ows file and then reload it.

janezd:
@ales-erjavec I did not get " it is first necessary to save a .ows file and then reload it.", and I can't try it (I don't have any Windows ready). If somebody creates a schema on Windows, saves it and sends it to somebody with Mac, will it work?

ales-erjavec:
The save/reload comment was general and not related to windows. The file widget (and in more general terms the whole workflow model) is only made aware of the base directory location when loaded from an existing .ows file. Therefore it is necessary to first save a new scheme (as it does not yet have an associated file) and load it again (I am working on a solution that would notify/propagate the changed base directory to the widgets when the workflow is saved).

Otherwise the .ows files created on Windows can now be opened on OSX, Linux, ...

kernc:
So this change would need to be applied in any data-loading widget? We have quite a couple ... with an outlook on more.

Would it not perhaps be easier on the widget developers (potentially many) to force their `recent_files` `Setting` (or equivalent (... `FileSetting`?)) be of particular `[dirname+'/'+basename, ...]` format (e.g. list of strings), and then at the ows-saving stage we rewrite all those dirnames to be ows-relative, with exception of the _datasets_ dir, which gets saved separately and is reconstructed correctly for each platform? Thus no extensive changes or code complexity is introduced in the data-loading widgets; only the use of `FileSetting`. Wouldn't ~this work?

I'm only asking because I see 146 lines of this PR added to owfile.py which is about 60% of its current size. :fearful:

janezd:
@kernc, weren't you in favour of simplifying widget settings handlers? :) Analysing settings and rewriting them would be insanely complex, dangerous and unmaintainable.

You're right, though, that we should use class `RecentFiles` (and perhaps also what is currently in `_relocate_recent_files`?) should be used in other widgets as well. @ales-erjavec, you probably planned this, right? This would not add too much overhead, and it would be simple and safe.

kernc:
> weren't you in favour of simplifying widget settings handlers?

Still am.

Obviously, you have to mark state-preserving attributes somehow. A simple, appropriately-named `Setting` sub-class could override its `__reduce__` and `__setstate__` (or whatever `Setting` objects use to pickle) to transparently relativize† the filepath strings of its list. Straightforward, SLOC-wise pretty miniscule (=maintainable), and surely not dangerous* if documented? Be it a single line change, I don't see it get much more simple, or how it counters my disposition.
```py
    recent_files = RecentFilesSetting([])
```

__*__ The crudeness of method that may safely work for `RecentFilesSetting` does not necessarily scale to other, potential new types of settings.

__†__ Only some paths are relativized. The paths that share a common prefix with the widget or an installed python package, respectively, aren't relativized but are instead resolved in a platform-dependent way.

janezd:
Settings do not work like this.

ales-erjavec:
The save/reload is no longer necessary. The path items get subsumed under the basedir on first (or subsequent) save.

janezd:
I like the solution with signals and events.

I tested it: I cannot make it fail -- and it's not for lack of trying. :) I guess it's done and I can merge it?

ales-erjavec:
>  ... I can merge it?

Yes

PR #717, opened 2015-09-14
Combo box size adjust policy

ales-erjavec:
Fix widget layout by explicitly setting/fixing combo box's contents sizeHint. Specifically when the combo box is used to display variables or variable values (from the widget inputs) and the name/values are long strings the comboBox's size hint could force the controlArea to grow horizontally to display the whole string.

This is solved by setting an appropriate sizeAdjustPolicy and minimumContentsLength on the QComboBox.

### DIFF
--- a/Orange/widgets/classify/owclassificationtreegraph.py
+++ b/Orange/widgets/classify/owclassificationtreegraph.py
@@ -46,7 +46,8 @@ class OWClassificationTreeGraph(OWTreeViewer2D):
         box = gui.widgetBox(self.controlArea, "Nodes", addSpace=True)
         self.target_combo = gui.comboBox(
             box, self, "target_class_index", orientation=0, items=[],
-            label="Target class", callback=self.toggle_target_class)
+            label="Target class", callback=self.toggle_target_class,
+            contentsLengthHint=8)
         gui.separator(box)
         gui.button(box, self, "Set Colors", callback=self.set_colors)
         dlg = self.create_color_dialog()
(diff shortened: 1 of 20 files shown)


ales-erjavec:
An alternative to gh-712 and gh-714

kernc:
So now, basically, every `gui.comboBox` call needs to also pass `contentsLengthHint` and possibly `sizeAdjustPolicy`? Something to the like of #714 seems more reasonable, despite its caveat.

janezd:
@ales-erjavec , I didn't think about different visual settings. You're right, #712 and #714 are not a good idea.

I would like avoid the additional argument, if possible. Especially in List boxes, which are (almost?) always filled with names or values, thus we would always have to specify the width.

Couldn't we have a default, say 12 (I saw that you used this in most cases). If all widgets that have dynamic content have to specify the argument, this seems reasonable. If the content is static and we forget to change the default to something appropriate, we would notice it, I guess. For a contrast, if we don't have a default and we forget to pass this argument, we usually won't notice the problem. Do you think this would work?

ales-erjavec:
I agree we would notice a problem sooner, but from an API point I would prefer to keep the default as is. The comboBox function 'as is' specifically does support/encourage use with fixed items/content, and without being told what it should do, having it sized to contents is a better default.

Maybe a better approach would be to separate the fixed/variable contents comboBox 'constructors'?

janezd:
@ales-erjavec I implemented your suggestion in #720. Review (and merge, if you like it) at your convenience.

janezd:
Fixed in #724

PR #720, opened 2015-09-18
Combo box size adjust policy with comboBox and varComboBox

janezd:
Implementation of changes suggested by @ales-erjavec in the discussion following #717.

Function `comboBox` works like before (any arguments that were passed as keyword arguments are still used, too), so these changes should not break anything.

Function `varComboBox` sets the default `contentsLengthHint` to 12; calls of `comboBox` are replaced by `varComboBox` where necessary, and the length hint is set to 8 if the `comboBox` contains values, not variable names.


ales-erjavec:
@janezd see #724 for another approach

janezd:
Fixed in #724

PR #724, opened 2015-09-22
gui.comboBox: Limit the combo box width

ales-erjavec:
Add a maximumContentsLenghtHint parameter to prevent the combo box from
growing uncontrollably depending on its contents.

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -1701,6 +1701,60 @@ def valueSlider(widget, master, value, box=None, label=None,
     return slider
 
 
+class OrangeComboBox(QtGui.QComboBox):
+    """
+    A QtGui.QComboBox subclass extened to support bounded contents width hint.
+    """
+    def __init__(self, parent=None, maximumContentsLength=-1, **kwargs):
+        super().__init__(parent, **kwargs)
+        self.__maximumContentsLength = maximumContentsLength
+
+    def setMaximumContentsLength(self, length):
+        """
+        Set the maximum contents length hint.
+
+        The hint specifies the upper bound on the `sizeHint` and
+        `minimumSizeHint` width specified in character length.
+        Set to 0 or negative value to disable.
+
+        .. note::
+             This property does not affect the widget's `maximumSize`.
+             The widget can still grow depending in it's sizePolicy.
+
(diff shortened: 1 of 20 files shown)


ales-erjavec:
Yet another alternative to #717, #720.

This is sort of a safety net. A ``comboBox(..., items=["A", "B", "C"])`` would still be size hinted appropriately to the contents but ``comboBox(..., items=["A", "B", "C" * 42])`` would have it's size hint bounded.

I would still like to add the contentsLengthHint from #717 (with a default None), and specify it where appropriate.

Orange/widgets/gui.py
```diff
@@ -1701,6 +1701,60 @@ def valueSlider(widget, master, value, box=None, label=None,
     return slider
 
 
+class OrangeComboBox(QtGui.QComboBox):
+    """
+    A QtGui.QComboBox subclass extened to support bounded contents width hint.
+    """
+    def __init__(self, parent=None, maximumContentsLengthHint=-1, **kwargs):
+        super().__init__(parent, **kwargs)
```
kernc: I don't think `QComboBox` supports any arguments besides `parent`.

ales-erjavec: > I don't think QComboBox supports any arguments besides parent.

PyQt allows (with some caveats) for any QObject's properties to be set with ``__init__`` keyword parameters.

janezd:
@ales-erjavec, I'm OK with this one, too. But we still need to limit the length, possibly by just applying your commit https://github.com/ales-erjavec/orange3/commit/cd06d6565351b87608f5e46777c42054c21cf88c?

ales-erjavec:
> But we still need to limit the length, possibly by just applying your commit ales-erjavec@cd06d65?

@janezd The `gui.comboBox` constructed widgets are by default already limited in width (albeit more loosely then in the earlier attempts at this).

Never the less, this train of thought is continued in ales-erjavec/orange3@219e1b66db3ef7d41fff62b905f4d4c93094d925 and ales-erjavec/orange3@00613d2ece1487fd3c5ef74e0f0136524171e8ec

janezd:
Is this finished now, so we can merge it and remove the alternatives?

PR #735, opened 2015-10-02
PaintData: Add Table name.

thocevar:
Add a field to name the painted dataset to distinguish between them in the Hub widget.


Orange/widgets/data/owpaintdata.py
```diff
@@ -803,6 +803,7 @@ class OWPaintData(widget.OWWidget):
     outputs = [("Data", Orange.data.Table)]
 
     autocommit = Setting(False)
+    table_name = Setting("data")
```
kernc: Why not "Painted data" ?

PR #736, opened 2015-10-02
DistanceMap: Deselect on right-click.

thocevar:


janezd:
Better alternative: if only a single cell is chosen, clear selection.

Pop-up on right click should still be disabled, though.

thocevar:
Clicking a cell (i, j) actually selects two instances - i and j (unless i=j).

PR #742, opened 2015-10-07
owhierarchicalclustering: Commit on new input

ales-erjavec:


lanzagar:
Why was there a `handleNewSignals` method at all? The widget has only one input (Distances) and `set_distances` already calls `_invalidate_clustering` which calls `_update_labels`. So the second call to `_update_labels` in `handleNewSignals` was unnecessary repetition.

Would it be better to add the `unconditional_commit` after `_invalidate_clustering` in `set_distances` and remove the `handleNewSignals` method?

ales-erjavec:
@lanzagar It's a remnant from days long past, when there was no DistMatrix class and the widget had a separate 'Items' input. I have removed it.

PR #743, opened 2015-10-08
[WIP] OWFile: Enable open file from url

VesnaT:
Accordingly, data file combo box is now editable.

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -1,6 +1,6 @@
 import os
 import sys
-
+import urllib
 from collections import namedtuple
 
 from PyQt4 import QtGui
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owfile.py
```diff
@@ -21,6 +21,14 @@ def add_origin(examples, filename):
             var.attributes["origin"] = dir_name
 
 
+def is_url_valid(url):
+    try:
+        urllib.request.urlopen(url)
+        return True
+    except:
```
kernc: The [documentation](https://docs.python.org/3.4/library/urllib.request.html#module-urllib.request) says it raises `URLError` on errors?

VesnaT: Should handle urllib.error.HTTPError as well.

kernc: https://docs.python.org/3.4/library/urllib.error.html#urllib.error.HTTPError

Orange/widgets/data/owfile.py
```diff
@@ -291,8 +325,19 @@ def browse_file(self, in_demos=0):
             self.recent_paths.remove(recent)
 
         self.recent_paths.insert(0, recent)
-        self.set_file_list()
-        self.open_file(self.recent_paths[0].abspath)
+
+    @staticmethod
+    def is_url_valid(url):
+        try:
+            return bool(Table.from_url(url)), ""
```
janezd: This test is too expensive if data is large. I think we should just check whether the link works and close the connection without transferring any data.

kernc: why not just `return True, Table(...)`?

janezd:
Additional changes needed:

- nicer error reporting when the URL is invalid or contains invalid data
- support for Google sheets
- standard icons for files and URLs

VesnaT:
Support for Google sheets is planned for another pull request.

VesnaT:
Support for Google sheets is planned for another pull request.

PR #748, opened 2015-10-09
Default continuous palette

acopar:
colorbrewer: add blue-yellow and red-block-green palettes
colorpalette: continuous pallete from white-black to blue-yellow
owheatmap, owdistancemap: select default blue-yellow palette


Orange/widgets/utils/colorbrewer.py
```diff
@@ -360,6 +360,8 @@
                   (188, 128, 189), (204, 235, 197), (255, 237, 111)]}},
 
     'sequential': {
+        "Blue-Yellow": {2: [(0, 0, 255), (255, 255, 0)]},
+        "Red-Black-Green": {3: [(255, 0, 0), (0, 0, 0), (0, 255, 0)]},
         'Reds': {3: [(254, 224, 210), (252, 146, 114), (222, 45, 38)],
```
ales-erjavec: The color palettes listed here are defined by http://colorbrewer2.org/, it should only contain those. Individual widgets can add their own. Also I don't think Red-Black-Green should be used in 'Distance Map'.

PR #750, opened 2015-10-09
[WIP] score.FCBF: a Fast Correlation-Based Filter for feature selection

kernc:
... is what the title says. The method is pretty popular ([1068+ citations](https://scholar.google.com/scholar?q=fast+correlation-based+filter+fcbf)) for reducing datasets with lots of features by extracting "optimal" feature sets, like [mRMR](https://en.wikipedia.org/wiki/Minimum_redundancy_feature_selection) does.

It's fast compared to ReliefF and can be enabled by default.

### DIFF
--- a/Orange/preprocess/_relieff.cpp
+++ b/Orange/preprocess/_relieff.cpp
@@ -779,7 +779,7 @@ typedef npy_double __pyx_t_5numpy_double_t;
 typedef npy_longdouble __pyx_t_5numpy_longdouble_t;
 
 /* "Orange/preprocess/_relieff.pyx":32
- * from numpy.math cimport INFINITY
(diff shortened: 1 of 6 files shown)


kernc:
Added a test for coverage.

BlazZupan:
Please add to the documentation (evaluation.cd.rst, the file should actually be renamed to evaluation.scoring.rst). Also, split the description of scores to "Classication Scoring" and "Regression Scoring".

kernc:
It was, in fact, preprocess.rst that needed to be added to. I didn't split it in two proposed sections as I found it pretty clear as-is; I made `feature_type` and `class_type` attributes shown instead. Is this ok?

PR #751, opened 2015-10-09
owpreprocess & preprocess: include PCA and CUR as preprocessors

marinkaz:


lanzagar:
As discussed, this functionality should be available through the Orange API (not just in widgets).

kernc:
This looks great! Could you rebase it on latest master so @lanzagar can merge it?

PR #753, opened 2015-10-10
Widget for assigning colors to variables

janezd:
To avoid unnecessary domain conversions, I introduced variable proxies: these are variables that are equivalent to original w.r.t. to data, but can differ with respect to outward appearance, such as name, colors ... The equivalence operator (`__eq__`) treats them as the same variable.

Basic functionality of the widget is implemented (probably with glitches).

What remains is removing the "Set color" functionality from widgets, and providing a simple fall-back for cases when the user does not use this widget.


Orange/data/variable.py
```diff
@@ -243,12 +250,31 @@ def __init__(self, name="", compute_value=None):
...
                 self._all_vars[name] = self
 
+    def make_proxy(self):
+        """
+        Copy the variable and set the master to `self.master` or to `self`.
+
+        :return: copy of self
+        :rtype: Variable
+        """
+        var = copy.copy(self)
```
kernc: Why do variables have a `copy()` method?

janezd: `Variable.copy` is a misnomer. It should be named something like `derive`. I'm not sure I want to use it here, `copy.copy` looks safer since I want to have exactly the same thing. @astaric ?

I actually don't like what `Variable.copy` does. For instance, it does not copy the `colors` attribute, which the widget that I added is assigning to variables. I'm not even sure it does what it is intended for: `ContinuousVariable.copy` returns instances of `ContinuousVariable` even when `compute_value` is a discretizer.

astaric: +1 for rename to derive.

Variable.derive is currently used in imputers, where we want a variable that looks the same, but with additional compute_value. For discretization, you need to define new DiscreteVariable with given values, so derive does not apply.

We can use copy.copy in variable.derive, if widgets are going to add new attributes  to variables.

astaric:
Clicking in an empty list (i.e. when dataset has no continuous/discrete variables) raises an Exception, that is not caught.

janezd:
I can't replicate that. I used Select columns to remove discrete or continuous or both, but didn't get an error clicking inside the empty space. What should I try?

janezd:
Ignore the message about not being able to replicate the exception occurring when clicking in empty space. I fixed it now.

PR #758, opened 2015-10-14
provider: Remove non-existent argument (Py3.5)

lanzagar:
`strict` has been deprecated since Python 3.3 and removed in 3.5
https://docs.python.org/3.4/library/html.parser.html


lanzagar:
@ales-erjavec can you confirm that it is OK to just remove this argument.
It does not exist anymore in py3.5 and produces an error.

ales-erjavec:
Sure, why not.

lanzagar:
Great, then this can be merged I guess.

PR #759, opened 2015-10-16
io: New heuristic for discerning headers from data

kernc:
If the first three lines of a dataset match the three-line header
format => it's a three-line header (regardless of the column names
used in the first line).

Otherwise, if the first and second line don't have all-digits in
exactly the same places and not-all-digits everywhere else => it's
a single-line header.

Otherwise, no header.

Also, (@BlazZupan) string variables are considered discrete only if at most `int(round(len(values)**.7))` values, so for 5 instances, there can be at most 3 distinct values for the variable to be considered discrete, and for 100, 25.

### DIFF
--- a/Orange/tests/test_tab_reader.py
+++ b/Orange/tests/test_tab_reader.py
@@ -74,6 +74,32 @@ class TestTabReader(unittest.TestCase):
         self.assertEqual(c1.name, "Class 1")
         self.assertEqual(c1.attributes, {'x': 'a longer string'})
 
+    def test_read_data_oneline_header(self):
+        samplefile = """\
+        data1\tdata2\tdata3
+        0.1\t0.2\t0.3
+        1.1\t1.2\t1.5
+        """
+        file = io.StringIO(samplefile)
+        table = TabFormat().read_file(file)
+
+        self.assertEqual(len(table), 2)
+        self.assertEqual(len(table.domain), 3)
+        self.assertEqual(table.domain[0].name, 'data1')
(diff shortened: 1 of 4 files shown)


Orange/tests/test_tab_reader.py
```diff
@@ -86,3 +87,17 @@ def test_reuse_variables(self):
...
+        data = Table(path.join(path.dirname(__file__), 'weird.tab'))
+        self.assertEqual(len(data), 6)
+        self.assertEqual(len(data.domain), 7)
+        self.assertEqual(data.domain[0].name, "['5534fab7fad58d5df50061f1', '5534fab8fad58d5de20061f8']")
+        self.assertEqual(data.domain[6].name, "['5534fab8fad58d5de20061f6', '5534fab7fad58d5de20061f4']")
+        attributes = dict(
+            Timepoint='0',
+            id="['5534fab9fad58d5de20061fa', '5534fab9fad58d5de20061f9']",
+            Name="['Gene expressions (dd_AX4_on_Ka_00Hr_bio1_mapped.bam)', 'Gene expressions (dd_AX4_on_Ka_00Hr_bio2_mapped.bam)']",
+            Replicate="['1', '2']",
```
kernc: @BlazZupan, @janezd, @astaric were these supposed to be proper lists?

BlazZupan:
I'm fine with this and have tested it on some data set for Gadi. We need to merge this urgently. Please do so if you agree.

PR #762, opened 2015-10-16
Tests: Fix some Windows related issues.

thocevar:


Orange/widgets/tests/test_widgets_outputs.py
```diff
@@ -14,7 +15,7 @@ def test_outputs(self):
             signal_names = {output.name for output in desc.outputs}
             module_name, class_name = desc.qualified_name.rsplit(".", 1)
             fname = importlib.util.find_spec(module_name).origin
-            with open(fname) as f:
+            with codecs.open(fname, encoding='utf-8') as f:
```
lanzagar: Why change to `codecs.open` from the built-in `open` (which also has an encoding parameter)?

PR #763, opened 2015-10-16
Popup short survey and show feedback link on the Welcome screen

kernc:
The two links are:
* http://orange.biolab.si/short-survey.html should redirect to short survey shown when Orange is first run;
* http://orange.biolab.si/long-survey.html should redirect to longer feedback form.

@markotoplak and @ajdapretnar are on it.

### DIFF
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -16,8 +16,8 @@ import signal
 
 import pkg_resources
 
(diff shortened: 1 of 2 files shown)


markotoplak:
Jernej, please use http://orange.biolab.si/survey/short.html and http://orange.biolab.si/survey/long.html

kernc:
:+1:

PR #765, opened 2015-10-16
Test&Score,Predictions: output predictions and probabilities

nikicc:
Test and score now also returns predicted values and probabilities

### DIFF
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -2,7 +2,7 @@ import numpy as np
 
 import sklearn.cross_validation as skl_cross_validation
 
-from Orange.data import Table
+from Orange.data import Table, Domain, ContinuousVariable, DiscreteVariable
 
 __all__ = ["Results", "CrossValidation", "LeaveOneOut", "TestOnTrainingData",
            "ShuffleSplit", "TestOnTestData", "sample"]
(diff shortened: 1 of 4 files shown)


Orange/evaluation/testing.py
```diff
@@ -202,6 +202,53 @@ def get_fold(self, fold):
 
         return results
 
+    def get_table_with_probabilities(self, model_names):
+        """
+            Return the Orange.data.Table with data, predictions and probabilities (if the task is classification).
```
kernc: The standard is for docstrings to be aligned with opening tri-quotes (cf. e.g. `np.hstack()`).

Orange/evaluation/testing.py
```diff
@@ -202,6 +202,53 @@ def get_fold(self, fold):
...
+
+            :param model_names: A list of strings containing learners' names.
```
kernc: Would you consider writing any new docstrings in [Napoleon style](http://sphinxcontrib-napoleon.readthedocs.org/en/latest/index.html)?

janezd: +1 for Napoleon. Much easier to read.

nikicc: Sure. Did we agree on whether to use Google or NumPy version or can I choose?

PR #774, opened 2015-10-22
OWRank: Get scores from linear regression and random forest

VesnaT:
OWRnk can now get signals from OWLinearRegression and OWRandomForest.
Therefore Liner regression learner and Random forest learner extend Scorer, which is additional input to OWRank.

When attribute is continuized, an average of related scores in given.

### DIFF
--- a/Orange/classification/logistic_regression.py
+++ b/Orange/classification/logistic_regression.py
@@ -1,16 +1,29 @@
+import numpy as np
+
 import sklearn.linear_model as skl_linear_model
 
 from Orange.classification import SklLearner, SklModel
 from Orange.preprocess import Normalize
+from Orange.preprocess.score import LearnerScorer
+from Orange.data import Variable, DiscreteVariable
 
 __all__ = ["LogisticRegressionLearner"]
+
+
+class _FeatureScorerMixin(LearnerScorer):
+    feature_type = Variable
+    class_type = DiscreteVariable
+
+    def score(self, model):
+        # Take the maximum attribute score across all classes
+        return np.max(np.abs(model.skl_model.coef_), axis=0)
 
 
 class LogisticRegressionClassifier(SklModel):
     pass
 
 
-class LogisticRegressionLearner(SklLearner):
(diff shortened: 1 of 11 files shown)


kernc:
What about Logistic Regression?

kernc:
The problem with Linear Regression is that it can assign significant _negative_ weights, which will get important attributes sorted below unimportant ones that have zero weight. Would it make sense to instead take the absolute value of weight?

VesnaT:
Logistic Regression will be added when this PR is merged..

kernc:
Disagreed. :+1: 

Also seems to be blocked on https://github.com/biolab/orange3/pull/825.

janezd:
@kernc @VesnaT I agree, we should take absolute values of coefficients. Alternatives are to compute some statistics (e.g. 1 - p_value) or to sort by absolute values. Note that some for some scores lower values can be better, so each score should have an associated cmp function (e.g. le(x), ge, ge(abs(x))...)

kernc:
> so each score should have an associated cmp function (e.g. le(x), ge, ge(abs(x))...)

Not sure I fully agree there. When the user sorts the TableWidget's column, they expect a consistent ascending/descending behavior and the red visual cue lines to mean what they say, namely that `-2 < 0.1` holds.

Since we never promise to output the raw coefficients (or similar), couldn't transforming the results (per scorer), like `abs(weights)`, `1/abs(weights)`, `1 - scale(weights, 0, 1)` (..., as appropriate), also work?

astaric:
This PR adds Scorer input to the Rank widget and adds a LearnerScorer class, that can fix scores created on continuized dataset.

Adding cmp to scores is out of scope, an so is determining what is the correct way to score features using linear regression.

@janezd @kernc if you think that Linear regression scoring is not ok, I suggest that the related code is moved to a separate pull request and the rest is merged (Rank using Scorers and RF scorer).

kernc:
@astaric With logistic regression and RF regressor missing, this PR seems incomplete.

kernc:
Please rebase on latest master.

PR #778, opened 2015-10-23
OWLinearRegression: output table with coefficients

janezd:


astaric:
Code accesses LinearModel.skl_model, which does not exist. 

It would be better, if widgets called a method on LinearModel, that would return the list of coefficients. This method would then take care of the getting the coefficients from the actual (skl) model and constructing them in a predefined way.

PR #788, opened 2015-10-23
Preprocess: Remove examples with missing values in class variables

VesnaT:


Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -239,6 +240,11 @@ def set_train_data(self, data):
             self.error(0, "Train data input requires a class variable")
             data = None
 
+        self.warning(4)
+        if data and np.isnan(data.Y).any():
+            self.warning(4, "Train data has missing values in class variable(s)")
```
janezd: Instances with unknown target values were removed from train data.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -254,6 +260,11 @@ def set_test_data(self, data):
             self.error(1, "Test data input requires a class variable")
             data = None
 
+        self.warning(5)
+        if data and np.isnan(data.Y).any():
+            self.warning(5, "Test data has missing values in class variable(s)")
```
janezd: Instances with unknown target values were removed from test data.

Also, consider joining the two warnings if it's not too complicated.

Orange/widgets/regression/owlinearregression.py
```diff
@@ -101,7 +103,20 @@ def _alpha_changed(self):
 
     def commit(self):
         alpha = self.alphas[self.alpha_index]
-        args = {"preprocessors": self.preprocessors}
+        preprocessors = self.preprocessors
+        if self.data is not None and np.isnan(self.data.Y).any():
+            self.warning(0, "Missing values in class variable(s)")
```
janezd: Instances with missing values of dependent variable are removed from the data.

We should not say "class" on regression data.

Orange/tests/test_preprocess.py
```diff
@@ -52,3 +54,24 @@ def test_nothing_to_remove(self):
...
         d = Orange.preprocess.preprocess.RemoveConstant(data)
         self.assertEqual(len(d.domain.attributes), 4)
+
+
+class TestRemoveNanClass(unittest.TestCase):
+    def test_remove_nan_classes(self):
+        table = Table("imports-85")
+        self.assertTrue(np.isnan(table.Y).any())
+        table = RemoveNaNClasses(table)
+        self.assertTrue(not np.isnan(table.Y).any())
```
janezd: I'm just curious: is this the same as `self.assertFalse(np.isnan(table.Y).any())` or am I missing something?

Orange/widgets/regression/owlinearregression.py
lanzagar: Why not try to be consistent and use the same as above (owtestlearners):
Instances with unknown target values were removed from the data.

PR #790, opened 2015-10-25
OWFile: Fix Reload for local files

kernc:
Fixes [issue 177](https://docs.google.com/spreadsheets/d/1spQ3zuXUEvoD-hpYEm33-lyS3OzV6cjhE7YlFlBoOw8/edit#gid=0).

@ales-erjavec can you confirm this has no negative consequences?

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -259,7 +259,9 @@ class OWFile(widget.OWWidget):
 
     def reload(self):
         if self.recent_paths:
-            if self.file_combo.currentText() == self.recent_paths[0].relpath:
+            basename = self.file_combo.currentText()
+            if (basename == self.recent_paths[0].relpath or
+                basename == os.path.basename(self.recent_paths[0].abspath)):
                 return self.open_file(self.recent_paths[0].abspath)
         self.select_file(len(self.recent_paths) + 1)
 


ales-erjavec:
I cannot replicate the error. Can you describe the problem in more detail?

kernc:
I could replicate it by saving some random /tmp/test.csv, saving, exiting. Then on restart, clicking Reload produced the said error.

ales-erjavec:
Ah, ok I see. The fix should be on line 264 (as you mentioned in the commit message).
Note that not only can `relpath` be None, it can also be a nested path (i.e. "/" joined components), while the file_combo.currentText() will only contain the last component (basename).

kernc:
Why does `relpath` get to be `None` in the first place? Doesn't _every_ absolute path have a (at least one) prefix and (likewise) a relative path?

ales-erjavec:
The `prefix` and `relpath` when not None record where the path is located relative to some specific location (e.g. the location of the workflow .ows file).

Note that `prefix` is not stored as a actual path but a name identifying that special location (`'basedir'`, `'sample-datasets'`, ...)

An example. Take the following directory structure
```
$HOME/work/
    workflow.ows
    datasets/
        dataset1.tab
        dataset2.tab
```
A path within `$HOME/work/` is recorded as 

```RecentPath("$HOME/work/dataset/dataset1.tab", prefix="basedir", relpath="datasets/dataset1.tab")```

while a path outside should be

```RecentPath("$HOME/Documents/someotherdataset.tab", prefix=None, relpath=None)```

**Comment by kernc** [comment]
So if I understand it correctly, a dir structure like
```
$HOME/work/
    workflows/
        scheme1.ows
        scheme2.ows
        scheme3.ows
    datasets/
        dataset1.tab
        dataset2.tab
        dataset3.tab
```
would not resolve datasets upon relocation of `work/` since they are not within `'basedir'` (`$HOME/work/workflows/`) prefix. Is that correct?

ales-erjavec:
Yes.

PR #793, opened 2015-10-25
OWWidget: allow toggling the visibility of controlArea

kernc:


janezd:
@kernc, this is brilliant. One small comment though: I guess this functionality should be enabled only for widgets that have `mainArea`. It makes no sense to hide `controlArea`, if `controlArea` is all there is.

janezd:
After this change, Combo box in the File widget no longer expands when the widget is resized.

kernc:
How about nao?

janezd:
What about widgets that don't have `controlArea`? In Scatter plot > rank projections, the splitter is at the left and "works" in a sense. This is uncommon, though.

Orange/widgets/widget.py
```diff
@@ -198,32 +197,48 @@ def createPixmapWidget(self, parent, iconName):
         gui.rubber(self.warning_bar)
         self.warning_bar.setVisible(False)
 
-        self.topWidgetPart = gui.widgetBox(self,
-                                           orientation="horizontal", margin=0)
-        self.leftWidgetPart = gui.widgetBox(self.topWidgetPart,
-                                            orientation="vertical", margin=0)
+        self.want_main_area = self.want_graph or self.want_main_area
+
+        class Splitter(QSplitter):
```
janezd: Is there any reason why each widget has a separate definition of the splitter class? This is useful for getting data from the closure, but these two classes don't use it. What about moving this class from the method into OWWidget's namespace? (I know I'm nitpicking.)

kernc: Makes sense.

kernc: Makes sense. But Qt doesn't handle it well in OWWidget's namescape (I suppose because this function is called in widget's constructor and `self` is not yet _fully_ available by then).

kernc:
Widgets that don't have `controlArea` should declare so. Now that VizRank does too, there is no splitter handle there anymore.

Orange/widgets/widget.py
kernc: Nevermind. It works now also as `OWWidget.Splitter`.

PR #794, opened 2015-10-25
widgets: set resizing_enabled=False on relevant small widgets without mainArea

kernc:


Orange/widgets/data/owfile.py
```diff
@@ -154,6 +154,7 @@ class OWFile(widget.OWWidget):
         doc="Attribute-valued data set read from the input file.")]
 
     want_main_area = False
+    resizing_enabled = False
```
janezd: Shouldn't we allow resizing of the file widget -- it can contain long URLs?

kernc: Disregarding how I use a tiling WM that tiles windows that don't provide "Dialog" WM hint, and how the widget looks really scarcely populated in such conditions, and how an odd edge case this is; observe how the combo box doesn't span the available width.

Regardless, why are long URLs an issue? The same argument could be said for other widgets that present information: the information can be wide, the containers will scroll horizontally, etc.
People don't parse URLs. #678 contains `Table.from_url()` that best converts those URLs into titles.

kernc: Said PR also includes a change that stretches the combo's drop-down listbox to accommodate its contents, regardless of combo's size.

PR #795, opened 2015-10-25
gui.auto_commit: allow callback specified other than commit()

kernc:
Not tested beyond it's still working for widgets using it as currently and my own use case.

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -2,6 +2,7 @@ import math
 import os
 import re
 import itertools
+from types import LambdaType
 
 import pkg_resources
 import numpy
@@ -15,6 +16,7 @@ from Orange.widgets.utils import getdeepattr
 from Orange.data import ContinuousVariable, StringVariable, DiscreteVariable, Variable
 from Orange.widgets.utils import vartype
 from Orange.widgets.utils.constants import CONTROLLED_ATTRIBUTES, ATTRIBUTE_CONTROLLERS
+from Orange.util import namegen
 
 YesNo = NoYes = ("No", "Yes")
 _enter_icon = None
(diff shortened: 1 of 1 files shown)


janezd:
Is there another rationale for this than allowing a different name?

Orange/widgets/gui.py
```diff
@@ -2239,7 +2239,7 @@ def paintEvent(self, ev):
 
 
 def auto_commit(widget, master, value, label, auto_label=None, box=True,
-                checkbox_label=None, orientation=None, **misc):
+                checkbox_label=None, orientation=None, callback='commit', **misc):
```
janezd: I would prefer `callback=None` as the default argument. Furthermore, I don't like passing strings like this; we have to do it for attributes, while for methods we can pass the reference, e.g. `callback=self.apply`. The method's name is then retrieved through `callback.__name__`.

Orange/widgets/gui.py
```diff
@@ -2293,12 +2295,12 @@ def commit():
     def do_commit():
         nonlocal dirty
         QApplication.setOverrideCursor(QCursor(Qt.WaitCursor))
-        master.unconditional_commit()
+        getattr(master, 'unconditional_' + callback)()
```
janezd: If the method is given a reference instead of string, this line can be replaced by `callback()`.

kernc:
It's for another button that does something else than "commit." Say I have two such auto-able buttons on a widget.

janezd:
I thought about this, but didn't want to over-design. If you have a use case with two buttons, I agree. If you just want to rename, I hesitatingly agree. :)

kernc:
Renaming wouldn't warrant a pull request. I just cringe and try to live with it.

PR #799, opened 2015-10-26
[WIP] owpredictions: Handle SqlTable input

lanzagar:
- Show an appropriate error with instructions for large SqlTable inputs
- If the table is small enough, download automatically


lanzagar:
A better way to do this is part of #802

PR #806, opened 2015-10-28
owsql: Option to download data to local memory

lanzagar:


Orange/widgets/data/owsql.py
```diff
@@ -221,6 +230,30 @@ def open_table(self):
...
+                confirm.addButton("OK", QMessageBox.YesRole)
+                confirm.exec()
+                self.download = False
+            elif table.approx_len() > AUTO_DL_LIMIT:
+                confirm = QMessageBox(self)
+                confirm.setIcon(QMessageBox.Warning)
+                confirm.setText("Data appears to be big. Do you really want to "
+                                "download it to local memory?")
+                confirm.addButton("Yes", QMessageBox.YesRole)
+                no_button = confirm.addButton("No", QMessageBox.NoRole)
```
kernc: Could reduce these lines into
```py
QMessageBox.question(self, 'Question',
                     'Do you want to...',
                     QMessageBox.Yes | QMessageBox.No)
```
call.

lanzagar: Shortened.

PR #812, opened 2015-10-28
owheatmap: Change the default under/overflow colors

ales-erjavec:
By default, use extremes from the color gradient instead of white/black.

### DIFF
--- a/Orange/widgets/visualize/owheatmap.py
+++ b/Orange/widgets/visualize/owheatmap.py
@@ -174,24 +174,35 @@ def color_palette_model(palettes, iconsize=QSize(64, 16)):
 
 
 def color_palette_table(colors, samples=255,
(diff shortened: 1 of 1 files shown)


Orange/widgets/visualize/owheatmap.py
```diff
@@ -174,24 +174,35 @@ def color_palette_model(palettes, iconsize=QSize(64, 16)):
 
 
 def color_palette_table(colors, samples=255,
-                        threshold_low=0.0, threshold_high=1.0, gamma=None):
+                        threshold_low=0.0, threshold_high=1.0,
+                        underflow=None, overflow=None,
+                        gamma=None):
```
kernc: Isn't there something like this, but apparently in much poorer condition, in `Orange.widget.utils.colorpalette.ExtendedContinuousPaletteGenerator`? Does it do (approximately) the same thing?

PR #814, opened 2015-10-29
IO fixups

kernc:
IO still failed to open some files. This fixes it once and for all with falling back to reading the file as utf-8 and ignoring all errors (skipping any bytes that fail to convert).

More reasonably and storage-aware, NaN values are saved as blanks.


Orange/data/variable.py
```diff
@@ -13,7 +13,7 @@
 # For storing unknowns
 Unknown = ValueUnknown = float("nan")
 # For checking for unknowns
-MISSING_VALUES = {"?", ".", "", "NA", "~", None}
+MISSING_VALUES = {"?", "nan", ".", "", "NA", "~", None}
```
kernc: Without this, `StringVariable`s that had "nan" values considered them as values, not missing.

Orange/data/variable.py
```diff
@@ -54,7 +54,9 @@ def is_discrete_values(values):
             return False
 
     # Strip NaN from unique
-    unique = {i for i in unique if not (isinstance(i, Number) and np.isnan(i))}
+    unique = {i for i in unique
+              if (not i in MISSING_VALUES and
```
kernc: This was not necessary, but only because this check is made in io already. In other situations, it will prove more robust to repeat it here.

PR #815, opened 2015-10-30
[WIP] Handle columns with all NaN values

ales-erjavec:


ales-erjavec:
Putting this on hold as it would conflict with #824

PR #818, opened 2015-10-30
Compatibility with newest Numpy and Scipy.

thocevar:
Various fixes of test errors and warnings:
* test_kmeans_parameters (RuntimeWarning: Mean of empty slice.)
  * Reduce the number of clusters.
* GradientPaletteGeneratorTest (RuntimeWarning: invalid value encountered in rint)
  * Avoid NaNs by temporarily settings them to 0.
* test_all_models_work_after_unpickling (ERROR: pickling a NaiveBayes model crashed)
  * Convert class_freq to numpy array.
* TestDiscretizer (ERROR: np.digitize doesn't accept empty lists of cut points anymore)
  * Wrap np.digitize (changed in numpy 1.10) for use in Discretizer.transform
* TestDomainInit (FutureWarning: numpy equal will not check object identity in the future.)
  * Avoid array comparisson with NaNs.
* test_SoftmaxRegressionPreprocessors (RuntimeWarning: invalid value encountered in multiply)
  * Reduce the values in additional column to avoid (numerical?) problems with softmax regression.
* TestPCA (test_chain returns different results from expected)
  * Limit PCA to 3 components.

### DIFF
--- a/Orange/preprocess/discretize.py
+++ b/Orange/preprocess/discretize.py
@@ -17,9 +17,13 @@ class Discretizer(Transformation):
         super().__init__(variable)
         self.points = points
 
+    @staticmethod
+    def digitize(x, bins):
+        return np.digitize(x, bins) if bins else [0]*len(x)
+
     def transform(self, c):
         if c.size:
-            return np.where(np.isnan(c), np.NaN, np.digitize(c, self.points))
+            return np.where(np.isnan(c), np.NaN, self.digitize(c, self.points))
         else:
             return np.array([], dtype=int)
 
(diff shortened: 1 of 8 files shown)


marinkaz:
It seems okay to me.

Orange/preprocess/discretize.py
```diff
@@ -16,9 +16,12 @@ def __init__(self, variable, points):
         super().__init__(variable)
         self.points = points
 
+    def digitize(self, x, bins):
```
kernc: Could this be made a `staticmethod`?

kernc:
@thocevar Please put human-readable description of changes from your above comment into the relevant commits.

PR #821, opened 2015-10-30
io: load var attributes as python objects (like JSON except not)

kernc:
JSON doesn't support single-quote-quoted strings. So were working with Python object dumps now
> a=b c=['d', 3] e={True: False}

Is this an issue?

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -2,6 +2,7 @@ import re
 import logging
 import subprocess
 from os import path
+from ast import literal_eval
 from itertools import chain, repeat
 from functools import lru_cache
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


janezd:
Is there any reason for not using proper JSON? (I noticed `True` and `False` instead `true` and `false`.)

Otherwise, I'm OK with any format and any type of quotes, for as long as the attributes are returned as normal Python objects.

kernc:
Yes.

> JSON doesn't support single-quote-quoted strings.

And we have single-quotes (or even no quotes) in files as the former implementation serialized attributes with
```py
'{!s}={!s}'.format(*a) for a in attributes.items()
```
And Python prefers single-quotes.

There appears to be no way to force `json` package to accept single-quotes (even though it provides several other overriding facilities).

We could go replacing `'` with `"` but ... I'd rather not.

kernc:
Gee, I was hoping for some more discussion here. Opting out of JSON doesn't makes us such an interoperable player if we ever choose to replace Python or if anyone else chooses to support our formats (both of which is somewhat unlikely, but possible).

PR #822, opened 2015-11-02
OWPreprocess: Rename Center and Scale to Normalize

VesnaT:


Orange/widgets/data/icons/Normalize.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
kernc: What does Normalize do? I don't like the icon.

lanzagar:
I was about to say that the rename and icon are fine, but Normalize should have the option to normalize to [0,1] or [-1,1] intervals.
Currently "scale by span" is not very useful because nobody wants to have s span of 1, but intervals [-0.731, 0.269] or [123.1, 124.1]

Maybe it is better to add this in a new PR anyway.

PR #823, opened 2015-11-03
Widget development guide/tutorial

ales-erjavec:
This is mostly the tutorial ported from Orange 2, but updated to Orange 3, and without the detailed documentation on the context handlers (how to subclass, ...).


Orange/widgets/widget.py
```diff
@@ -86,12 +95,22 @@ class OWWidget(QDialog, metaclass=WidgetMetaClass):
...
 
-    # Default widget layout settings
+    # Default widget GUI layout settings
+    # ----------------------------------
+
+    #: Should the widget have basic layout
+    #: (If this flag is false then all the following flags are also ignored).
     want_basic_layout = True
+    #: Should the widget construct a `mainArea` (this is a resizable
+    #: area left of the controlArea).
```
kernc: This is `controlArea`.

PR #827, opened 2015-11-06
Generate widget index with travis.

markotoplak:


doc/widget_index.py
```diff
@@ -0,0 +1,46 @@
...
+        ret.append((cat, []))
+    if li.name == "li":
+        a = li.find("a")
+        imglink = webdocprefix + get_icon_path(a.get("href"))
+        doclink = webdocprefix + a.get('href')
+        text = a.string
+        ret[-1][1].append({"img": imglink, "doc": doclink, "text": text })
+    
+with open(outfile, 'wt') as f:
+    f.write(json.dumps(ret))
```
kernc: If this file is getting version tracked anywhere, I'd suggest calling `json.dump(ret, f, indent=1)`.

markotoplak: Thanks!

doc/widget_index.py
```diff
@@ -0,0 +1,46 @@
+"""
+Generete widget index for the webpage.
+"""
+
+import os
+from bs4 import BeautifulSoup
```
lanzagar: Can you group imports: https://www.python.org/dev/peps/pep-0008/#imports
It makes it easier to see which external libraries are needed when their imports are not lost between python std library imports.

PR #828, opened 2015-11-06
save: Renamed widget to 'Save Data'

nikicc:


lanzagar:
Changes like this should now be accompanied with updates to the widget documentation as well.

PR #833, opened 2015-11-09
colorpalette: Fix use of ColorPaletteGenerator

ales-erjavec:
Fix IndexErrors encountered when the color palettes were incorrectly
used with a large number of colors.

Add a `ColorPaletteGenerator.resolve` helper method to aid in more
robust/flexible palette use.

### DIFF
--- a/Orange/widgets/utils/colorpalette.py
+++ b/Orange/widgets/utils/colorpalette.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import copy
 import math
 from numbers import Number
 from collections import Iterable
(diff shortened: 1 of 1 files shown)


janezd:
@ales-erjavec @markotoplak 

Blocked by #753: after #753 is merged, changes in owtable.py will need to be removed since colors will be managed by another widget.

janezd:
I guess we keep just the `resolve` method (for future use), but the changes in other two files are obsoleted by #891.

janezd:
@ales-erjavec , #891 has been merged. Can you rebase and check whether we still need any of the changes in tables and trees?

PR #836, opened 2015-11-09
gui.lineEdit: commit button should accept default action

kernc:
If lineEdit and some other default push button are on the same page, pressing Enter after typing into lineEdit activated the default push button as well.


kernc:
I don't understand, however, why it isn't enough to `setAutoDefault(True)`. :worried:

PR #837, opened 2015-11-09
owrank: Fix a TypeError on selection

ales-erjavec:
Due to b8cbd688c82fe5c11d681f897dfc2916a477e832 (gh-795) the `gui.auto_commit`
replacement `commit` function has a changed signature (`*args, **kwargs`)
and as a result has different call semantics when used as a target of a
PyQt4 connection.

In this case it would be called with two arguments (selected and
deselected QItemSelection), throwing an error when calling the real
shadowed `commit`, while before both arguments would be dropped by
Qt/PyQt.

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -5,11 +5,11 @@ Rank
 Rank (score) features for prediction.
 
 """
-
+import sys
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


kernc:
Sorry. I don't know why I introduced `*args` in the first place.

PR #839, opened 2015-11-11
LegendItem: Ensure the restored legend position is inside the viewbox

ales-erjavec:


Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -90,6 +91,24 @@ def __init__(self, size=None, offset=None, pen=None, brush=None):
             brush = QBrush(QColor(232, 232, 232, 100))
         self.__brush = brush
 
+    def storeAnnchor(self):
```
janezd: Typeo: storeAnnchor?

PR #841, opened 2015-11-11
distmatrix: fixed bug in invert methods

nikicc:
The `invert` method was missing and option for `(max - X)` that was described in comments. Also the code was inconsistent with the method documentation.

### DIFF
--- a/Orange/misc/distmatrix.py
+++ b/Orange/misc/distmatrix.py
@@ -101,6 +101,8 @@ class DistMatrix(np.ndarray):
         elif typ == 1:
             return 1.-self
         elif typ == 2:
+            return np.max(self) - self
+        elif typ == 3:
             return 1./self
         else:
             raise ValueError('Unknown option for typ of matrix inversion.')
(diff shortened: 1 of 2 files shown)


Orange/misc/distmatrix.py
```diff
@@ -101,6 +101,8 @@ def invert(self, typ):
         elif typ == 1:
             return 1.-self
         elif typ == 2:
+            return np.max(self) - self
+        elif typ == 3:
             return 1./self
```
kernc: So I looked through the codebase, and this method is used exactly ... nowhere.

[YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

kernc:
So even if there is a widget yet unported that would use this method, it still makes no sense to use it. No one should ever use it. It's remembering a method's name and an arbitrary mapping of four parameters (or looking it all up) versus applying any of the simple element-wise operations when and as needed.

I vote for purging.

PR #844, opened 2015-11-13
owdistributions: tooltips

markotoplak:


Orange/widgets/visualize/owdistributions.py
```diff
@@ -148,54 +152,60 @@ def __init__(self):
...
             plot.setMouseEnabled(False, False)
             plot.setMenuEnabled(False)
 
-        disable_mouse(self.plot.getViewBox())
+        disable_mouse(self.plot)
         disable_mouse(self.plot_prob)
 
+        self.tooltip_items = []
+        self._tooltip_delegate = HelpEventDelegate(self.help_event)
+        self.plot.scene().installEventFilter(self._tooltip_delegate)
```
kernc: Why is `_tooltip_delegate` an instance attribute if not used anywhere else? Seems like these two could as well be a single (logical) line?

markotoplak: It is garbage collected otherwise: http://stackoverflow.com/questions/3308013/creating-an-event-filter

kernc: Thanks. From the source you link, it seems like you can instantiate it without making it a member attribute if you set its parent?
```py
self.plot.scene().installEventFilter(HelpEventDelegate(self.help_event, self))
```

markotoplak: Thanks.

BlazZupan:
Please rebase and resolve conflicts with current head.

PR #849, opened 2015-11-16
Simplify 'Feature Constuctor' widget GUI

ales-erjavec:
* Remove GUI controls for setting the number of decimals, base value, ...
* Replace the QListView controls for editing discrete variable's values
  with a QLineEdit (values are comma separated, escaped with '\' if necessary.
* Move the QListView with defined variables below the 'Editor'.


ales-erjavec:
I have changed the button/editor layout. Request for comments.

janezd:
@ales-erjavec I tried rearranging the controls as you suggested in the last layout, but it looks a bit confusing. So let's stick with the one that you implemented now.

A few minor changes, though: could you set the width of the New and Remove buttons to 120 and put `addSpace(10)` between them? Can you also increase the width of the combo boxes Select feature and Select function? At least on my computer I only get Select fea and Select fun...

PR #857, opened 2015-11-17
Revision of tutorial workflows.

BlazZupan:
Updated to match the current settings of File widget.
The text accompanying the workflows was revised.


rokgomiscek:
Tutorial "Classification tree" doesn't work. Upon loading returns error: ValueError("'Classification Tree' is not a valid output channel name for 'Classification Tree'.",)

BlazZupan:
Rok, thanks, fixed.

PR #863, opened 2015-11-19
OWScatterplot: Rank projection fix

VesnaT:
Use RReliefF instead of ReliefF for regression datasets. 
I'm not sure whether this is the right solution or should I just disable rank projection for regression datasets.


markotoplak:
This is good. We also need to modify the scoring of knn projections to accomodate regression. Blaž and I were considering the R2 score.

PR #868, opened 2015-11-20
scatterplot rank projections: R2 score for regression

markotoplak:


Orange/widgets/visualize/owscatterplot.py
```diff
@@ -578,7 +579,9 @@ def run(self):
                     y = y_full[valid]
                     knn = NearestNeighbors(n_neighbors=self.k).fit(X)
                     ind = knn.kneighbors(return_distance=False)
-                    score = norm * np.sum(y[ind] == y.reshape(-1, 1))
+                    score = norm * np.sum(y[ind] == y.reshape(-1, 1)) \
```
janezd: @markotoplak Any chance of turning this into an ordinary if? I find these expressions difficult to read, and when they stretch across three lines, it's just a matter of swapping them around...

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -578,7 +579,9 @@ def run(self):
...
+                            if isinstance(self.parent_widget.data.domain.class_var, DiscreteVariable) \
```
VesnaT: This line its too long.
https://www.python.org/dev/peps/pep-0008/#id14

VesnaT:
Have you thought about adding std to mean? Something like:
score = r2_score(y, np.mean(y[ind], axis=1) + np.std(y[ind], axis=1))

markotoplak:
@VesnaT, this is an interesting solution. I think that a good solution that also incorporates variance is out of scope of this pull request - it may be a fun research project. :)

lanzagar:
I always had RMSE in mind for this.
It probably does not make a big difference, but do you think there are some (dis)advantages between using R2 or RMSE?
If results are similar, we should probably select the faster method.

markotoplak:
I opted for R2 because higher value == better, which is the same as you get for discrete class.

janezd:
R2 is normalized between 0 and 1, is it? So R2 has some absolute meaning while RMSE is not scale-invariant.

lanzagar:
The range of values for R2 is indeed better. And the speed difference should not be big since R2 is basically just normalized RMSE.

PR #870, opened 2015-11-20
OwSelectColumns: Add "Apply on change". Task 178

rokgomiscek:
OwSelectColumns: Add "Apply on change". Task 178

### DIFF
--- a/Orange/widgets/data/owselectcolumns.py
+++ b/Orange/widgets/data/owselectcolumns.py
@@ -304,6 +304,7 @@ class OWSelectAttributes(widget.OWWidget):
 
     settingsHandler = SelectAttributesDomainContextHandler()
     domain_role_hints = ContextSetting({})
+    auto_commit = Setting(False)
 
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owselectcolumns.py
```diff
@@ -605,7 +609,8 @@ def update_completer_model(self, *_):
                         [list(v.attributes.items()) for v in vars], [])
         items.extend(["%s=%s" % item for item in labels])
         items.extend(reduce(list.__add__, list(map(list, labels)), []))
-
+        if self.auto_commit:
```
janezd: @rokgomiscek This is wrong: you have to call `self.commit`. `gui.auto_commit` replaces your `self.commit` with a function the checks the auto commit flag and calls the original commit or sets the dirty flag.

Therefore, remove `if` and always call.

rokgomiscek: Thank you, fixed

janezd:
@rokgomiscek Arranging Reset and Apply vertically doesn't look nice, it extends the widget vertically and leaves empty space on the right. Please use this layout:

        layout.addWidget(
            gui.button(None, self, "Reset", callback=self.reset),
            3, 0, 1, 1)
        layout.addWidget(
            gui.auto_commit(None, self, "auto_commit", "Apply", "Auto apply"),
            3, 2, 1, 1)

(I'll also create PR on your PR, out of curiousity. I don't know why it apparently has so many commits - is it because I applied you PR to my master which contains the commits you haven't yet applied. Will rebase of your branch somehow magically remove my commits?)

rokgomiscek:
Fixed

PR #872, opened 2015-11-20
Progress bar context handler and decorator

janezd:
Packs `progressBarInit` and `progressBarFinished` into context manager to ensure the progress bar is removed in case of exceptions.

If PR is accepted, this should become the preferred way of handling progress bars.

### DIFF
--- a/Orange/widgets/widget.py
+++ b/Orange/widgets/widget.py
@@ -1,3 +1,4 @@
+import contextlib
 import sys
 import time
 import os
@@ -580,6 +581,39 @@ class OWWidget(QDialog, Report, metaclass=WidgetMetaClass):
 
         if processEvents is not None and processEvents is not False:
             qApp.processEvents(processEvents)
+
+    @contextlib.contextmanager
+    def progressBar(self, iterations=0):
+        """
+        Context manager for progress bar.
+
+        Using it ensures that the progress bar is removed at the end without
+        needing the `finally` blocks.
+
+        Usage:
+
+            with self.progressBar(20) as progress:
+                ...
+                progress.advance()
+
+        or
+
+            with self.progressBar() as progress:
+                ...
+                progress.advance(0.15)
+
+        or
+
+            with self.progressBar():
+                ...
+                self.progressBarSet(50)
+
+        :param iterations: the number of iterations (optional)
+        :type iterations: int
+        """
+        progress_bar = gui.ProgressBar(self, iterations)
+        yield progress_bar
+        progress_bar.finish()  # Let us not rely on garbage collector
 
     #: Widget's status message has changed.
     statusMessageChanged = Signal(str)
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -372,42 +372,41 @@ class OWTestLearners(widget.OWWidget):
             preprocessor=self.preprocessor,
             callback=update_progress)
         self.setStatusMessage("Running")
-        self.progressBarInit()
 
-        try:
-            if self.resampling == OWTestLearners.KFold:
-                if len(self.data) < self.k_folds:
-                    self.error(4, "Number of folds exceeds the data size")
-                    return
+        with self.progressBar():
+            try:
+                if self.resampling == OWTestLearners.KFold:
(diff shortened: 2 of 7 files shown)


Orange/widgets/widget.py
```diff
@@ -539,6 +539,48 @@ def progressBarFinished(self, processEvents=QEventLoop.AllEvents):
         if processEvents is not None and processEvents is not False:
             qApp.processEvents(processEvents)
 
+    def progressBar(self, show_bar=True):
```
kernc: What is `show_bar` and why would it ever be set to `False`?

kernc: The current _preferred_ progress bar interface is to use `gui.progressBar` which provides a nice `advance()` wrapper for when the number of steps is known in advance.

Could this method accept instead a number of steps (like the one in `gui` does) and expose a similar method on the in-`__enter__`-returned `ProgressBarHandler` object?

Orange/widgets/widget.py
```diff
@@ -539,6 +539,48 @@ def progressBarFinished(self, processEvents=QEventLoop.AllEvents):
...
+        Decorator and context manager for progress bar.
+
+        Using this decorator ensures that the progress bar is removed at the
+        end of the with-block or at exit from function even in case of
+        exceptions, therefore removing the need to remove the bar in
+        `finally` blocks.
+
+        Usage as context handler:
+
+            with self.progressBar:
```
kernc: The examples above make a _call_ to `self.progressBar()`.

Orange/widgets/widget.py
```diff
@@ -539,6 +539,48 @@ def progressBarFinished(self, processEvents=QEventLoop.AllEvents):
...
+                do_something()
+                update_progress_bar_while_doing_it()
+
+        Usage as decorator:
+
+            @OWWidget.progressBar
+            def compute(self):
+                do_something()
+                update_progress_bar_while_doing_it()
```
kernc: I don't like this usage. It removes the progress bar initialization logic away from the code and into a decorator. This implies that either the code with a call to `progressBarSet()` (the example in this docstring could be clearer in this regard — what's `update_progress_bar_while_doing_it()`?), or the decorator could both be used on their own, which is obviously not the case.

There should be one obvious way to do it. Nothing wrong with `with self.progressBar()` (other than it increments indent level, but so does `finally`).

kernc:
If this should become the preferred way, would you port existing code in widgets (i.e. make more examples) and mark the `gui.ProgressBar` [deprecated](https://github.com/biolab/orange3/blob/master/Orange/util.py#L14).

janezd:
@kernc Now that I reimplemented it, I totally agree: decorators were a terrible idea. Context managers are nice, and the argument with the number of iterations simplifies the existing code. Please review the PR.

I haven't touched the old interface. There are some cases in which the init, advance and finish appear in different functions, therefore we will keep the old interface as a low-level functions for progress bars. For the same reason, `gui.ProgressBar` still makes sense.

We now have three mechanisms: context handler, widget's methods and `gui.ProgressBar`. They are useful for different cases, though.

Orange/widgets/widget.py
```diff
@@ -570,6 +570,39 @@ def progressBarFinished(self, processEvents=QEventLoop.AllEvents):
...
+
+            with self.progressBar(20) as progress:
+                ...
+                progress.advance()
+
+        or
+
+            with self.progressBar() as progress:
+                ...
+                progress.advance(15)
```
kernc: Calling `advance()` without initializing with total number of iterations results in what? 1500%?

kernc:
If you fix the conflicts #889 introduced, this looks good to me.

Did you get to replace all the uses of `progressBarInit()` (& co.) in the library?

janezd:
I resolved the conflicts.

I also replaced all uses of `progressBarInit()` wherever possible, that is, wherever the init and tear down were in the same block.

kernc:
Would you like to [squash these two commits](https://www.google.com/search?q=github+squash+commits) since the first one includes just about everything and the second one is only a single-line fix-up?

kernc:
Can you please rebase this once again so we can merge?
Thanks. :smile:

janezd:
Rebased.

PR #875, opened 2015-11-23
Simple Tree

VesnaT:
Include simple tree in regression package
Print average instead of sum in regression tree. 
Number of decimals depends on value of root's sum.

### DIFF
--- a/Orange/classification/simple_tree.py
+++ b/Orange/classification/simple_tree.py
@@ -276,11 +276,14 @@ class SimpleTreeModel(Model):
             else:
                 node = self.node
         n = node.contents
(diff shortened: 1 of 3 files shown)


BlazZupan:
Printout of regression trees is now fine. But please correct the precision when printing out the classification trees and number of instances in the nodes. There, one decimal number is enough. Now prints out with many:

physician-fee-freeze ([168.0, 267.0])
: n
   adoption-of-the-budget-resolution ([3.7476415634155273, 249.660400390625])
   : n --> democrat ([2.1769492626190186, 23.48069953918457])

PR #878, opened 2015-11-24
owdatasampler: Use correct sql sampling type constants

lanzagar:


Orange/widgets/data/owdatasampler.py
```diff
@@ -104,13 +104,13 @@ def f():
...
         spin = gui.spin(ibox, self, "sampleSizeSqlTime", minv=1, maxv=3600,
-                        callback=set_sampling_type(self.FixedSize))
+                        callback=set_sampling_type(self.SqlTime))
         spin.setSuffix(" sec")
         gui.appendRadioButton(sampling, "Percentage")
         ibox = gui.indentedBox(sampling)
         spin = gui.spin(ibox, self, "sampleSizeSqlPercentage", spinType=float,
                         minv=0.0001, maxv=100, step=0.1, decimals=4,
-                        callback=set_sampling_type(self.FixedProportion))
+                        callback=set_sampling_type(self.SqlProportion))
```
kernc: The two values are from two different enums?

lanzagar: Yes, the widget presents a different set of choices for standard and sql data.
Hence two sets of constants - they are defined on lines 32 and 33.

PR #879, opened 2015-11-24
OWCalibrationPlot: Handle regression data

VesnaT:
Widget crashed when given regression data. Show error message instead.


tomazc:
@VesnaT can you please change the first error message ("Give me data!!") to something nicer, e.g.: Evaluation results require information on test data.
Similar error message also in owliftcurve.py.

VesnaT:
Changed.

PR #884, opened 2015-11-26
OWDataSampler: Show error message when given small dataset

VesnaT:
Widget crashed when given dataset smaller than desired output. Now it shows error message insted.

### DIFF
--- a/Orange/widgets/data/owdatasampler.py
+++ b/Orange/widgets/data/owdatasampler.py
@@ -185,6 +185,8 @@ class OWDataSampler(widget.OWWidget):
         else:
             if self.indices is None or not self.use_seed:
                 self.updateindices()
+                if self.indices is None:
+                    return
             if self.sampling_type in [self.FixedProportion, self.FixedSize]:
                 remaining, sample = self.indices
                 self.outputInfoLabel.setText(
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owdatasampler.py
```diff
@@ -202,6 +204,31 @@ def commit(self):
...
+            if self.data.domain.has_discrete_class else 0
+
+        if self.sampling_type == self.FixedSize:
+            size = self.sampleSizeNumber
+            repl = self.replacement
+        elif self.sampling_type == self.FixedProportion:
+            size = np.ceil(self.sampleSizePercentage / 100 * data_length)
+            repl = False
+        elif data_length < self.number_of_folds:
+            err_msg = "Number of folds should not be bigger than input"
```
janezd: "Number of folds exceeds the data size"

Orange/widgets/data/owdatasampler.py
```diff
@@ -202,6 +204,31 @@ def commit(self):
...
+
+        if not repl and (data_length <= size or (data_length <= num_classes
+                                                 and self.stratify)):
+            err_msg = "Output should be smaller than input"
```
janezd: This message is confusing. It can refer to the first condition, but not to the second. Also, consider that the first condition cannot occur if sampling is `self.FixedProportion`, therefore it can be moved to `self.FixedSize`.

I think it would be better to have two different messages, something like
*"Sample size exceeds the original data size"* and
*"Not enough data for stratified sampling"*

I guess it would be better to split it into two different messages.

VesnaT:
FIxed.

PR #890, opened 2015-12-02
Venn Diagram:  Add option to output unique/all instances.

ales-erjavec:


lanzagar:
When unique option is checked, the number is wrong:
e.g. use two file widgets and load iris, intersection shows 300 items while the output contains 147

When it is not checked it says 300 and actually outputs 300. Is this intended behaviour?
The intersection then effectively becomes union if unique is not checked?

ales-erjavec:
Yes that is the intended behavior. The 'Unique items only' applies to the output only (tool tips over the diagram do show unique counts though).

thocevar:
Selecting an intersection of two datasets should not output more elements than the size of the smaller dataset (regardless of whether "Unique items only" option is chosen).

"Use data identifiers" could be disabled when there are no identifiers in the dataset instead of reporting a problem.

PR #891, opened 2015-12-02
Use Variable.colors for coloring palettes

janezd:


Orange/widgets/visualize/owmosaic.py
```diff
@@ -142,7 +140,6 @@ def __init__(self):
         self.attributeValueOffset = 3
         self.residuals = []  # residual values if the residuals are visualized
         self.aprioriDistributions = []
-        self.colorPalette = None
```
lanzagar: This was removed but is still used in line 357 (of the new file)

lanzagar:
Scatter Map uses default colours instead of the ones that were selected.

PR #900, opened 2015-12-07
OWSelectRows: Split purge attributes

VesnaT:
Split purge attributes to Remove unused features and Remove constant features.
If one attribute in the data set had only 1 defined instance and the other instances empty (NaN), the widget used to remove the attribute, because it considered it a constant. Now removing unused and constant features is done separately.


janezd:
But isn't this attribute indeed constant in the sense that it has not other (defined) values than just a single one? What would be the expected behaviour in this case - from the user's perspective?

lanzagar:
We have a special widget for advanced purging (with specific requirements).
I think a single purge option for features/classes is enough in Select Rows.

The described example actually already did what I would expect most people want.

PR #904, opened 2015-12-07
OWDistanceMap: Use domain context to store selected annotation var

ales-erjavec:


janezd:
@ales-erjavec @astaric Is it OK that attributes in context settings are stored by indices, not names? I guess this should only be done with `PerfectDomainContextHandler`.

One thing I tried is loading glass.tab and selecting the third attribute. Then I loaded iris.tab and the third attribute was selected - obviously retrieving the context from glass.tab.

Furthermore, if I take glass.tab again, but use Select columns to change the order of attributes, Distance map would still annotate by the third attribute. Since some widgets may purge unused attributes, context settings must use attribute names, not indices, I guess.

lanzagar:
Is this WIP now?
The comment from janez looks relevant.

PR #907, opened 2015-12-08
PCA: Use it as feature scorer

VesnaT:
Add an output to PCA widget that can go to Rank widget.

### DIFF
--- a/Orange/projection/pca.py
+++ b/Orange/projection/pca.py
@@ -1,3 +1,4 @@
+import numpy as np
 import sklearn.decomposition as skl_decomposition
 
 try:
@@ -10,14 +11,24 @@ except ImportError:
         pass
 
 import Orange.data
+from Orange.data import Variable
 from Orange.misc.wrapper_meta import WrapperMeta
 from Orange.preprocess import Continuize
 from Orange.projection import SklProjector, Projection
+from Orange.preprocess.score import LearnerScorer
 
 __all__ = ["PCA", "SparsePCA", "RandomizedPCA", "IncrementalPCA"]
 
 
-class PCA(SklProjector):
+class _FeatureScorerMixin(LearnerScorer):
+    feature_type = Variable
+    component = 0
+
+    def score(self, model):
+        return np.abs(model.components_[self.component])
+
+
+class PCA(SklProjector, _FeatureScorerMixin):
     __wraps__ = skl_decomposition.PCA
     name = 'pca'
 
--- a/Orange/tests/test_pca.py
+++ b/Orange/tests/test_pca.py
@@ -3,7 +3,7 @@ import numpy as np
 import pickle
 
 import Orange
-from Orange.preprocess import Continuize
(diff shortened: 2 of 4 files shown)


BlazZupan:
Other scorers (like linear and logistic regression) provide scoring method. PCA scores features only when given data (I've tried this on widgets). There's some redundancy in this: Rank widget is given the data, and so is PCA as well. Was this intentional?

lanzagar:
The scorer that is sent as the output from the PCA widget does not perform normalization. Even if that option was selected in the widget.
Especially for scoring features, normalization should almost always be used!

Orange/widgets/unsupervised/owpca.py
```diff
@@ -147,13 +162,13 @@ def start(self):
         else:
             self.address = self.addresstext.text()
             with remote.server(self.address):
-                from Orange.projection.pca import RemotePCA
```
lanzagar: RemotePCA needs to be imported inside the remote context!

janezd:
@lanzagar, can you check and merge if this is OK now?

Orange/projection/pca.py
```diff
@@ -10,14 +11,35 @@ def save_state(_):
...
 from Orange.misc.wrapper_meta import WrapperMeta
 from Orange.preprocess import Continuize
 from Orange.projection import SklProjector, Projection
+from Orange.preprocess.score import LearnerScorer, LearnerScorerError
 
 __all__ = ["PCA", "SparsePCA", "RandomizedPCA", "IncrementalPCA"]
 
 
-class PCA(SklProjector):
+class PCAComponentOutOfRangeError(LearnerScorerError):
```
lanzagar: This Error in no longer needed?

Orange/tests/test_pca.py
```diff
@@ -106,3 +106,26 @@ def test_chain(self):
...
+        data = Orange.data.Table('iris')
+        pca = PCA()
+        scores = pca.score_data(data)
+        self.assertEqual(len(scores), len(data.domain.attributes))
+        self.assertGreater(sum(scores), 0)
+        self.assertEqual(['petal length', 'sepal length'],
+                         sorted([data.domain.attributes[i].name
+                                 for i in np.argsort(scores)[-2:]]))
+
+    def test_PCA_scorer_component(self):
```
lanzagar: The test should now probably check the first component.

Orange/tests/test_pca.py
```diff
@@ -106,3 +106,26 @@ def test_chain(self):
...
+        data = Orange.data.Table('iris')
+        pca = PCA()
+        pca.component = 1
+        scores = pca.score_data(data)
+        self.assertEqual([round(s, 4) for s in scores],
+                         [0.6565, 0.7297, 0.1758, 0.0747])
+
+    def test_PCA_scorer_component_out_of_range(self):
```
lanzagar: This test in no longer relevant.

PR #909, opened 2015-12-11
setup.py: added docutils as a requirement

nikicc:
`docutils` was missing as a requirement and consequently in the bundle on our website. This caused the description of add-ons not to be properly formatted.
screen shot 2015-12-11 at 10 18 02


kernc:
> `docutils` was missing as a requirement and consequently in the bundle on our website. This caused the description of add-ons not to be properly formatted.

This information, the _why_ of it, is something that, when not evident, always serves well in the extended commit description.

nikicc:
I'll keep this in mind for the next time.

PR #911, opened 2015-12-11
Clean up requirements

lanzagar:
Sort requirements.txt and add some comments to setup.py


kernc:
So do you also pledge to watch over keeping the two places with requirements consistent and up-to-date?

lanzagar:
No.

Somebody should organize this better when he has time.

PR #917, opened 2015-12-11
[WiP] distances: switch axis 0 and 1; numpy, scipy tests compatibility

nikicc:


kernc:
In favor of reducing competing efforts, #818 is already about NumPy 1.10 compatibility.

PR #918, opened 2015-12-11
Color: Fix model reset

janezd:
Color widget failed to update when it received new data because the model did not emit the necessary signals.

### DIFF
--- a/Orange/widgets/data/owcolor.py
+++ b/Orange/widgets/data/owcolor.py
@@ -34,9 +34,9 @@ class ColorTableModel(QAbstractTableModel):
         return Qt.ItemIsEditable | Qt.ItemIsEnabled | Qt.ItemIsSelectable
 
     def set_data(self, variables):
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owcolor.py
```diff
@@ -34,9 +34,9 @@ def flags(self, _):
         return Qt.ItemIsEditable | Qt.ItemIsEnabled | Qt.ItemIsSelectable
 
     def set_data(self, variables):
+        self.emit(SIGNAL("modelAboutToBeReset()"))
         self.variables = variables
-        self.emit(SIGNAL("dataChanged(QModelIndex, QModelIndex)"),
-                  self.index(0, 0), self.index(self.n_columns(), self.n_rows()))
+        self.emit(SIGNAL("modelReset()"))
```
kernc: Consider not using [old-style signals](http://pyqt.sourceforge.net/Docs/PyQt4/new_style_signals_slots.html) in new code.

It's
```py
self.modelReset.emit()
```

PR #924, opened 2015-12-14
LogLoss: Unittest

VesnaT:
Orange.evaluation.scoring.Score.from_predicted uses only classes, but LogLoss needs probabilities. In order to verify correctness of Orange.evaluate.LogLoss, new unit test was written.

### DIFF
--- a/Orange/tests/test_evaluation_scoring.py
+++ b/Orange/tests/test_evaluation_scoring.py
@@ -87,7 +87,8 @@ class Scoring_AUC_Test(unittest.TestCase):
 
     def test_multiclass_auc_multi_learners(self):
         data = Orange.data.Table('iris')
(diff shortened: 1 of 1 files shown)


Orange/tests/test_evaluation_scoring.py
```diff
@@ -153,6 +155,25 @@ def test_log_loss(self):
...
+        return - ll / len(act)
+
+    def test_log_loss_calc(self):
+        data = Orange.data.Table('titanic')
+        learner = Orange.classification.LogisticRegressionLearner()
+        results = Orange.evaluation.TestOnTrainingData(data, [learner])
+
+        actual = np.copy(results.actual)
+        actual = actual.reshape(actual.shape[0], 1)
+        actual = np.hstack((np.abs(actual - 1), actual))
```
lanzagar: Why `np.abs(actual - 1)` instead of `1 - actual` ?

PR #598, opened 2015-08-17
Ability to learn without and predict wih missing values

markotoplak:
All data sets need preprocessors for testing as we can always have unknown values in testing data even when there are none in learning data.

PR #925, opened 2015-12-14
Preprocess test data when evaluating learners

kernc:


lanzagar:
This uses preprocessors on test data and applies them on training data.
Learners should not get information from the test data before fitting the model as it can lead to (misleadingly) higher evaluation scores.

kernc:
Closed in favor of https://github.com/biolab/orange3/pull/598.

PR #940, opened 2015-12-18
owsave: display filename and add save auto commit

tomazc:


lanzagar:
The OWSave class is reused for widgets that open when clicking the Save Graph button in various visualizations.
Do these changes also make sense there? I don't know what auto_commit would do in those cases...

janezd:
@lanzagar You're right, but let us solve this differently. I never liked this reuse of `OWSave`: we are treating widget as an ordinary dialog. I suggest we extract the common functionality into a super class and use that class in other widgets, while OWSave adds widgety stuff, like the "manifest", auto commit etc.

Orange/widgets/data/owsave.py
```diff
@@ -69,7 +76,7 @@ def save_file_as(self):
             filename += format_extensions[0]
         self.filename = filename
         self.last_dir, file_name = os.path.split(self.filename)
-        self.save.setText("Save as '%s'" % file_name)
+        self.info.setText("Saving as '%s'" %  file_name)
         self.save.setDisabled(False)
         self.save_file()
```
janezd: `gui.auto_commit` assigns a wrapper to `self.save_file`. In this line, you must call `self.unconditional_save_file`, otherwise "Save as..." will only save if auto save is on. The other call (in function `self.dataset`) is OK, because there you want conditional saving.

janezd:
This PR is superseded by #946.

PR #943, opened 2015-12-21
Rank projection: Consider missing values

VesnaT:
Pairs of attributes where data has no missing values should have better ranking.


Orange/widgets/visualize/owscatterplot.py
```diff
@@ -562,11 +562,12 @@ def run(self):
                     knn = NearestNeighbors(n_neighbors=n_neighbors).fit(X)
                     ind = knn.kneighbors(return_distance=False)
                     if isinstance(self.parent_widget.data.domain.class_var,
-                            DiscreteVariable):
+                                  DiscreteVariable):
```
lanzagar: If you are also improving this line, you could change it to:
`if self.parent_widget.data.domain.has_discrete_class:`

PR #946, opened 2015-12-21
Add auto save to OWSave; Move saving of graphs to separate function

janezd:
Alternative to #940, w.r.t. auto save: the file name is here shown as button text. This design is cleaner than an extra label.

Abusing this widget for saving graphs was a bad idea and did not really work. When reused for this purpose the widget showed Save As button that doesn't make sense, the widget didn't close but remained as a modal dialog after saving.

Save graph button in all widgets now opens the standard dialog for saving files, without detour to the modal `OWSave` with format selection and two buttons with the same function. This works on Linux, too, since it uses `getSaveFilenameAndFilter` function.

- `want_graph` is replaced with `graph_name` that can contain the name of the attribute with the graph (e.g. `plot.plotItem`). This attribute is used for both, saving and reports.
- If `graph_name` is specified, `OWWidget` adds the button to the layout (as before), but now also connects the (new) method `OWWidget.save_graph`. Widget therefore no longer need to connect the button.
- `save_graph` is eliminated in all widgets, except for tree viewer; the implementation in `OWWidget` reads `want_graph` to obtain the graph objects.
- `report_plot` no longer requires the graph object since it is read from `graph_name`
- the layout of Save Graph and Report buttons is improved: now they are both in the same box
- the function for saving graphs remembers the last used directory and file format.

If the extension does not match the format (not possible on Os X), user is prompted to change the extension or the format.

The same dialog is now also used in OWSave.

In summary, this removes all code for saving graphs from widgets; widgets only need to specify the name of the attribute with the graph.

### DIFF
--- a/Orange/widgets/utils/saveplot.py
+++ b/Orange/widgets/utils/saveplot.py
@@ -2,76 +2,32 @@ import os.path
 
 from PyQt4 import QtGui, QtCore
 
-
-CHANGE_EXT, CHANGE_FORMAT, KEEP = range(3)
-
-
-def fix_extension(ext, format, suggested_ext, suggested_format):
-    dlg = QtGui.QMessageBox(
-        QtGui.QMessageBox.Warning,
-        "Mismatching extension",
-        "Extension '{}' does not match the chosen file format, {}.\n\n"
-        "Would you like to fix this?".format(ext, format))
-    role = QtGui.QMessageBox.AcceptRole
-    keep_settings = dlg.addButton("Save as it is", role)
-    change_ext = \
-        suggested_ext and dlg.addButton("Use extension " + suggested_ext, role)
-    change_format = \
-        suggested_format and dlg.addButton("Save as " + suggested_format, role)
-    dlg.exec()
-    if dlg.clickedButton() == keep_settings:
-        return KEEP
-    elif dlg.clickedButton() == change_ext:
-        return CHANGE_EXT
-    elif dlg.clickedButton() == change_format:
-        return CHANGE_FORMAT
+from Orange.widgets.utils import filedialogs
 
 
 # noinspection PyBroadException
 def save_plot(data, file_formats, filename=""):
-    writers = sorted(set(file_formats.values()), key=lambda w: w.PRIORITY)
-    filters = ['{} (*{})'.format(w.DESCRIPTION, ' *'.join(w.EXTENSIONS))
-               for w in writers]
-
     _LAST_DIR_KEY = "directories/last_graph_directory"
-    _LAST_FMT_KEY = "directories/last_graph_format"
+    _LAST_FILTER_KEY = "directories/last_graph_filter"
     settings = QtCore.QSettings()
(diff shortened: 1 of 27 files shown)


Orange/widgets/utils/saveplot.py
```diff
@@ -0,0 +1,39 @@
...
+    filename, filter = QtGui.QFileDialog.getSaveFileNameAndFilter(
+        None, 'Save as ...', start_dir, ';;'.join(filters), last_ext)
+    if not filename:
+        return
+
+    writer = writers[filters.index(filter)]
+    if not os.path.splitext(filename)[1] and writer.EXTENSIONS:
+        filename += writer.EXTENSIONS[0]
+    try:
+        writer.write(filename, data)
```
kernc: So if the extension _is_ provided by the user, the file is nevertheless saved in the format of the chosen filter? I think that's worse than before. The correct writer in that case should be:
```python
filename, ext = os.path.splitext(filename)
writer = next((w for w in writers if ext in w.EXTENSIONS), None)
if not writer:
    error('Unsupported filename extension: {ext}. Supported are: {supported}'.format(...)
```

janezd: I agree that you may sometimes want to specify the format by typing an extension. But what if I'm intentionally specifying some particular non-standard extension? Say that the user sets the filter to svg, but then modifies the extension. If the extension is known, the filter is just (silently) overriden? This may be unexpected. And if the extension is unrecognized, the user could still have had a reason for this. When saving data, I may want to save in tab format, but as .txt because Excel recognizes it.

Both approaches make sense, so perhaps we should let the user make the decision (with another dialog) whether to (a) fix the extension, (b) use the format corresponding to the extension (your suggestion), (c) use the "wrong" combination (current solution) or (d) cancel?

kernc: Or simpler, if the extension is valid, the correct writer for that extension should be used. If the extension is invalid (.txt), the writer for the chosen filter is the right one.
```python
writer = next(...)
if not writer:
   writer = writers[filters.index(filter)]
```

janezd: I still think it's better to ask than to second-guess the users intention. I added a function that shows a dialog which offers changing the format, the extension or keeping both.

janezd:
- If the extension is unrecognized, Qt automatically adds the one from filter. This is undesired, but I can't prevent it. (I believe that this is in Qt as it also happens on Linux, not just Os X).
- On Os X, if the extension belongs to another filter, Os X requests the user to fix it. This is OK, I guess.
- In the same case on Linux, extension is not fixed automatically; Orange asks the user whether to change the extension or file format, or go back to the dialog.

kernc:
> * In the same case on Linux, extension is not fixed automatically; Orange asks the user whether to change the extension or file format, or go back to the dialog.

Yet again, only the Linux users are proven competent. :laughing:

janezd:
@kernc Can you check and merge? I'm sick of rebasing (and solving conflicts) every Friday. :)

PR #947, opened 2015-12-22
FilterString: Fix is_defined filtering

VesnaT:
IsDefined filtering did not work because String missing values were represented as np.nan.
Now String missing values are represented as None.

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -369,7 +369,8 @@ class FileFormat(metaclass=FileFormatMeta):
 
             if type_flag in StringVariable.TYPE_HEADERS:
                 coltype = StringVariable
-
+                values = [None if isinstance(i, float)
+                          else i for i in orig_values]
             elif type_flag in ContinuousVariable.TYPE_HEADERS:
                 coltype = ContinuousVariable
                 try:
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -352,9 +352,12 @@ class FileFormat(metaclass=FileFormatMeta):
             flag = Flags(Flags.split(flags[col]))
             if flag.i: continue
 
+            type_flag = types and types[col].strip()
             try:
-                orig_values = [np.nan if i.strip() in MISSING_VALUES else i.strip()
-                               for i in data[:, col]]
+                orig_values = [np.nan if i.strip() in MISSING_VALUES and
+                                         type_flag not in
+                                         StringVariable.TYPE_HEADERS
+                               else i.strip() for i in data[:, col]]
             except IndexError:
                 # No data instances leads here
                 orig_values = []
@@ -362,15 +365,12 @@ class FileFormat(metaclass=FileFormatMeta):
                 # only to satisfy test_table.TableTestCase.test_append
                 coltype = DiscreteVariable
 
-            type_flag = types and types[col].strip()
             coltype_kwargs = {}
             valuemap = []
             values = orig_values
 
             if type_flag in StringVariable.TYPE_HEADERS:
                 coltype = StringVariable
-                values = [None if isinstance(i, float)
-                          else i for i in orig_values]
             elif type_flag in ContinuousVariable.TYPE_HEADERS:
                 coltype = ContinuousVariable
                 try:
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1144,9 +1144,9 @@ class Table(MutableSequence, Storage):
             elif isinstance(f, data_filter.FilterString) and \
                             f.oper == f.IsDefined:
                 if conjunction:
-                    sel *= (col != "")
+                    sel *= col.astype(bool)
                 else:
-                    sel += (col != "")
+                    sel += col.astype(bool)
             elif isinstance(f, data_filter.FilterDiscrete):
                 if conjunction:
                     s2 = np.zeros(len(self), dtype=bool)
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -775,11 +775,11 @@ class StringVariable(Variable):
     @staticmethod
     def str_val(val):
         """Return a string representation of the value."""
-        if isinstance(val, Real) and isnan(val):
+        if val is None:
             return "?"
         if isinstance(val, Value):
             if val.value is None:
-                return "None"
+                return "?"
             val = val.value
         return str(val)
 
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -775,10 +775,10 @@ class StringVariable(Variable):
     @staticmethod
     def str_val(val):
         """Return a string representation of the value."""
-        if val is None:
+        if val is "":
             return "?"
         if isinstance(val, Value):
-            if val.value is None:
+            if val.value is "":
                 return "?"
             val = val.value
         return str(val)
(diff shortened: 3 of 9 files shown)


janezd:
io.data.FileFormat.write_data: change `isinstance(val, Number) and isnan(val)` to `isinstance(val, Number) and isnan(val)  or var.is_string and val is None`.

Check also: `widgets.data.owmergeddata.group_table_indices`.

`OWVennDiagram._itemsForInput`, `OWVennDiagram.commit`, `owvenndiagram.reshape_wide` (in two places) and `owvenndiagram.unique_non_nan` probably failed on string values before (`numpy.isnan(val)` fails if `val` is string, right?)

Orange/data/io.py
```diff
@@ -369,7 +369,8 @@ def _equal_length(lst):
 
             if type_flag in StringVariable.TYPE_HEADERS:
                 coltype = StringVariable
-
+                values = [None if isinstance(i, float)
```
kernc: Why not `''` (an empty string) instead of `None`? Since `''` in data stands for a missing value, there's no way to discern missing strings from non-missing empty strings anyway.

Could also avoid another pass through the column if `type_flag` was set before `orig_values` a couple lines above.

VesnaT:
It does not fail, since 'val' is usually Orange.data.variable.Value and it can be passed to numpy.isnan().
So I just added some tests...

Orange/data/io.py
janezd: @VesnaT, I was also thinking about empty strings as an alternative to `None`. Empty strings are also false, and they have the same type, so it may be more practical.

Also, consider the attribute contains gene names. Empty name is then the same as missing name. It is probably uncommon to have a "known, but empty" value.

Orange/data/io.py
```diff
@@ -92,13 +94,14 @@ class Flags:
...
-        ('class',     'c'),
-        ('ignore',    'i'),
-        ('meta',      'm'),
-        ('weight',    'w'),
-        ('.+?=.*?',    ''),  # general key=value attributes
+        ('class', 'c'),
+        ('ignore', 'i'),
+        ('meta', 'm'),
+        ('weight', 'w'),
+        ('.+?=.*?', ''),  # general key=value attributes
```
kernc: Yes, because this is now much better! Such improve! Wow!

Orange/data/io.py
```diff
@@ -312,22 +328,25 @@ def data_table(self, data, headers=None):
...
+                    for i in headers[0]])
                 names = list(names)
             elif 2 == len(headers):
                 names, _flags = map(list, headers)
             else:
                 # Use heuristics for everything
                 names, _flags = [], []
-            types = [''.join(filter(str.isupper, flag)).lower() for flag in _flags]
+            types = [''.join(filter(str.isupper, flag)).lower() for flag in
+                     _flags]
```
kernc: Could you at least break on `for`? When I was initiated, @astaric said we keep `--max-line-length` at about 120.

Orange/data/io.py
```diff
@@ -440,15 +471,19 @@ def valuemap_index(val):
 
                 # Reorder discrete values to match existing variable
                 if var.is_discrete and not var.ordered:
-                    new_order, old_order = var.values, coltype_kwargs.get('values', var.values)
+                    new_order, old_order = var.values, coltype_kwargs.get(
+                        'values', var.values)
```
kernc: You call this better, more clear? IMHO, if you wish to break on 80 so badly, at least do it properly. The example here being no pearl, but code that is worked on by lots of people is foremost meant to be readable; breaking lines at any cost does not constitute code any more readable than does breaking them poorly.

Orange/data/io.py
```diff
@@ -482,8 +517,10 @@ def _vartype(var):
             if var.is_continuous or var.is_string:
                 return var.TYPE_HEADERS[0]
             elif var.is_discrete:
-                return Flags.join(var.values) if var.ordered else var.TYPE_HEADERS[0]
+                return Flags.join(var.values) if var.ordered else \
+                var.TYPE_HEADERS[0]
```
kernc: There: http://stackoverflow.com/a/17319775/1090455

Orange/data/io.py
```diff
@@ -109,9 +112,10 @@ def __init__(self, flags):
             if self._RE_ALL.match(flag):
                 if '=' in flag:
                     k, v = flag.split('=', 1)
-                    self.attributes[k] = (v if Flags._RE_ATTR_UNQUOTED_STR(v) else
-                                          literal_eval(v) if v else
-                                          '')
+                    self.attributes[k] = (
+                    v if Flags._RE_ATTR_UNQUOTED_STR(v) else
+                    literal_eval(v) if v else
+                    '')
```
kernc: I can't even.

Orange/data/io.py
```diff
@@ -362,28 +386,29 @@ def _equal_length(lst):
...
+                        try:
+                            float(num)
+                        except ValueError:
+                            break
                     raise ValueError('Non-continuous value in (1-based) '
-                                     'line {}, column {}'.format(row + len(headers) + 1,
-                                                                 col + 1))
+                                     'line {}, column {}'.format(
+                        row + len(headers) + 1,
+                        col + 1))
```
kernc: What the??

Orange/data/io.py
nikicc: Have we reached a verdict on the `''` vs `None` dilemma?

Orange/data/io.py
```diff
@@ -352,9 +373,12 @@ def _equal_length(lst):
             flag = Flags(Flags.split(flags[col]))
             if flag.i: continue
 
+            type_flag = types and types[col].strip()
             try:
-                orig_values = [np.nan if i.strip() in MISSING_VALUES else i.strip()
-                               for i in data[:, col]]
+                orig_values = [np.nan if i.strip() in MISSING_VALUES and
+                                         type_flag not in
+                                         StringVariable.TYPE_HEADERS
```
kernc: How about avoiding evaluating this condition for each value. Python doesn't (as of yet) simplify comprehensions even when the condition is known to be static.

More importantly, what if `type_flag` is unspecified and the values will only later in the lines below be found to, in fact, constitute a `StringVariable`?

It might be better to replace `nan` values there and then.

kernc:
Quite obviously, I seriously oppose accepting "PEP8" commit (the title is not descriptive enough; PEP8 what?! First draft?). Would require much more sound justification to change my view than "the idiotic tool told me to."

Orange/data/io.py
kernc: Since the tab format is specified to count `\t\t` values as missing (and since Python's `csv` module doesn't handle `\t\t` and `\t""\t` differently), we can't and never have distinguished between missing values and true empty strings anyway.

janezd:
I like PEP8 formatting rules and don't mind about the commit name (we can rename it to "code formatting", if you prefer). However, its purpose is to make code more readable and automated tools are terrible at that. A typical example is https://github.com/VesnaT/orange3/commit/5fe75b7af57612bd25c7505c9cb661b23b1d6b9a#diff-1805316caf2ecdb375af0afa32d0fe69R140. If the line is broken like this

     _RE_TYPES = re.compile(r'^\s*({}|{}|)\s*$'.format(
         _RE_DISCRETE_LIST.pattern,
         '|'.join(flatten(getattr(vartype, 'TYPE_HEADERS')
                          for vartype in Variable.registry.values()))))

it is both PEP8-compliant and also more readable than the original (for my taste, that is).

Let us adhere to PEP8 where it makes sense, esp. in the code we are changing anyway, but not with automated tools.

kernc:
The PEP8 commit is pertinent to the `io.py` module. It makes sense to tag it so.

What do I set max line length to? :smiley:

PR #949, opened 2015-12-23
Linear regression: Include elastic net regression

VesnaT:


janezd:
@BlazZupan: @VesnaT and I discussed consistency vs. aesthetics. The alternative (see below) is not as inconsistent as the original layout is ugly -- there is no good position for the sliders.

Several widgets already have "non-standard" interfaces (Impute and Discretize, and also Feature constructor and Select Rows). I applied the same layout idea to SVMs (see PR #952)

screen shot 2015-12-24 at 09 51 12

screen shot 2015-12-24 at 09 52 26

Apply can be put in a single line, of course. Slider that don't apply in particular context are disabled. They could also be hidden, though I'm not sure it would look nice. In contrast with what I did for SVMs, I prefer sliders over spin boxes in this case; I guess the situation is similar to logistic regression where the slider solves the ambiguity of C parameter.

BlazZupan:
I like the last layout, with the parameters on the right. "Elastic net mixing parameter" could be simplified to "Elastic net mixing". Shoud "L1 ratio" be "L1/L2 ratio:"? Else, great that we finally have elastic net.

BlazZupan:
Should "Apply on every change" be just "Apply on change". Then, it would be even easier to fin within a single line.

janezd:
- Yes, let's remove the word "parameter".
- L1 / L2 ratio could be misleading. The number is the proportion of L1 (@VesnaT, right?)
- As for auto apply, I would remove the label altogether, like in most other widgets; so it's just the checkbox and the button, in the same line of course

VesnaT:
The number is combination of L1 and L2. For ratio = 1, it is L1. For ratio = 0, it is L2. And for 1 > ratio > 0, it is L1 * ratio and L2 * (1-ratio).

lanzagar:
Maybe it could just say L2 on the left and L1 on the right (below the slider). And in the middle is the number with the current value of the slider position.

Also - is the slider disabled when Elastic net is not selected? I think the settings would be clearer for the user if it would be.

VesnaT:
The slider is disabled..

janezd:
Like this?
screen shot 2015-12-24 at 17 50 58

lanzagar:
Looks good!

lanzagar:
When moving one of the two sliders - the model is relearned for every intermediate change of the value, right?
Could this be changed so that learning is done only once per slider-change (when releasing the mouse button)? I think there are some controls like this in other widgets, but can't remember where?

Orange/widgets/regression/owlinearregression.py
```diff
@@ -124,6 +157,9 @@ def commit(self):
             learner = RidgeRegressionLearner(alpha=alpha, **args)
         elif self.reg_type == OWLinearRegression.Lasso:
             learner = LassoRegressionLearner(alpha=alpha, **args)
+        elif self.reg_type == OWLinearRegression.Elastic:
```
lanzagar: This if-elif construct is missing an else statement and is potentially problematic.
I just tested this pull request and tried the Elastic option. Then on the master branch, Orange crashed because the learner was not set (stored widget settings contained reg_type=3).
The problem is of course because of incorrect settings, but it would not fail so badly if this had a proper else clause.

Orange/widgets/regression/owlinearregression.py
```diff
@@ -27,12 +27,13 @@ class OWLinearRegression(OWProvidesLearner, widget.OWWidget):
...
 
     #: Types
-    OLS, Ridge, Lasso = 0, 1, 2
+    OLS, Ridge, Lasso, Elastic = 0, 1, 2, 3
 
     learner_name = settings.Setting("Linear Regression")
     ridge = settings.Setting(False)
     reg_type = settings.Setting(OLS)
     alpha_index = settings.Setting(0)
+    l1_ratio = settings.Setting(0.5)
```
lanzagar: This looks like a sensible default.
Could you also change the above line to something other than the smallest possible value (e.g. the index of 0.5 or 1).

PR #951, opened 2015-12-23
Use separate function for saving graphs instead of OWSave

janezd:
Save graph button in all widgets now opens the standard dialog for saving files, without detour to te modal `OWSave` with format selection and two buttons with the same function. This works on Linux, too, since it uses `getSaveFilenameAndFilter` function.

- `want_graph` is replaced with `graph_name` that can contain the name of the attribute with the graph (e.g. `plot.plotItem`). This attribute is used for both, saving and reports.
- If `graph_name` is specified, `OWWidget` adds the button to the layout (as before), but now also connects the (new) method `OWWidget.save_graph`. Widget therefore no longer need to connect the button.
- `save_graph` is eliminated in all widgets, except for tree viewer; the implementation in `OWWidget` reads `want_graph` to obtain the graph objects.
- `report_plot` no longer requires the graph object since it is read from `graph_name`
- the layout of Save Graph and Report buttons is improved: now they are both in the same box
- the function for saving graphs remembers the last used directory and file format.

In summary, this removes all code for saving graphs from widgets; widgets only need to specify the name of the attribute with the graph.

Blocks #946.

### DIFF
--- a/Orange/widgets/utils/saveplot.py
+++ b/Orange/widgets/utils/saveplot.py
@@ -1,0 +1,42 @@
+import os.path
+from operator import attrgetter
+
+from PyQt4 import QtGui, QtCore
+
+
+# noinspection PyBroadException
+def save_plot(data, file_formats, filename=""):
+    formats = [(f.DESCRIPTION, f.EXTENSIONS)
+               for f in sorted(set(file_formats.values()),
+                               key=attrgetter("OWSAVE_PRIORITY"))]
+    filters = ['{} (*{})'.format(desc, ' *'.join(exts))
+               for desc, exts in formats]
+
+    _LAST_DIR_KEY = "directories/last_graph_directory"
+    _LAST_EXT_KEY = "directories/last_graph_extension"
+    settings = QtCore.QSettings()
+    start_dir = settings.value(_LAST_DIR_KEY, filename)
+    if not start_dir or not os.path.exists(start_dir):
+        start_dir = os.path.expanduser("~")
+    last_ext = settings.value(_LAST_EXT_KEY, "")
+    if last_ext not in filters:
+        last_ext = filters[0]
+
+    filename, filter = QtGui.QFileDialog.getSaveFileNameAndFilter(
+        None, 'Save as ...', start_dir, ';;'.join(filters), last_ext)
+    if not filename:
+        return
+
+    ext = os.path.splitext(filename)[1]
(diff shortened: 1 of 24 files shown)


janezd:
@astaric This PR accidentally includes a fix for fonts in El Capitan (in Orange/canvas/__main__.py). Could you please check that it does not break the layout in the earlier versions of Os X?

kernc:
The accident is most easily resolved:
```
git reset HEAD^
git commit Orange/canvas/__main__.py -m "Adjust fonts on Macintosh"
git commit -am "Stop abusing OWSave for saving graphs"
git push -f
```

Orange/widgets/utils/saveplot.py
```diff
@@ -0,0 +1,42 @@
...
+    if last_ext not in filters:
+        last_ext = filters[0]
+
+    filename, filter = QtGui.QFileDialog.getSaveFileNameAndFilter(
+        None, 'Save as ...', start_dir, ';;'.join(filters), last_ext)
+    if not filename:
+        return
+
+    ext = os.path.splitext(filename)[1]
+    exts = formats[filters.index(filter)][1]
```
kernc: So a chosen filter overrides my explicitly specified extension? Isn't normally the other way around?

janezd: This is what would happen on Windows unless you put the name in quotes. But I agree, this is annoying. Also, on Os X the system dialog warns you that you're specifying a wrong extension and asks you to confirm -- and after confirmation this function adds the unwanted extension anyway. :)

What do you suggest? I guess we add an extension only if `filename` has none, that is, if there is no dot after slash?

janezd: I now changed that in PR#946. Extension is added only if there is none. It can't happen on Os X, though -- the dialog always returns correct extension. See https://github.com/biolab/orange3/pull/946/files#diff-0a256bbfcab13a3efd498498377c9fa9R29.

These to two PRs are related and should be "squashed" together. (Should I? How? Just cherry pick commits from both branches into one?)

kernc: If they are largely similar or about the same thing, cherry picking them into one probably does make sense. You can `merge --ff-only` one into the other (or merge and `rebase -i` on master to flatten it), resolve the conflicts, push.

janezd: The other was already rebased on this one, so I'm closing this one.

PR #954, opened 2015-12-24
Fix fonts on El Capitan

janezd:


kernc:
The docstring says this is a temporary fix for [QTBUG-32789](https://bugreports.qt.io/browse/QTBUG-32789), which is marked as resolved as of Qt 4.8.6. Is this patch still necessary?

janezd:
For older Os X, we may want to keep it for those with older versions of Qt. The fix for El Capitan is needed and is not just temporary: Qt developers said they will not port the fix back from 5.0 to 4.8. I removed the misleading comment.

kernc:
So this fix applies to Qt 4? I think the comment was in order. It points someone like me to more information.

janezd:
I updated the comment to include new bug id's.

PR #957, opened 2015-12-28
OWSaveClassifier: append .pickle extension if not already provided

kernc:
Fixes:
http://stackoverflow.com/questions/34489451/orange3-save-classifier-widget

### DIFF
--- a/Orange/widgets/classify/owsaveclassifier.py
+++ b/Orange/widgets/classify/owsaveclassifier.py
@@ -22,7 +22,9 @@ class OWSaveClassifier(widget.OWWidget):
     #: A list of recent filenames.
     history = Setting([])
 
-    FILTER = "Pickle files (*.pickle *.pck)\nAll files (*.*)"
+
+    FILE_EXT = '.pkcls'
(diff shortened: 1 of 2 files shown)


kernc:
An interesting, ambiguity-preventing alternative would be to introduce a new extension for pickled classifiers, e.g. `.pkcls`.

Orange/widgets/classify/owsaveclassifier.py
```diff
@@ -22,7 +22,9 @@ class OWSaveClassifier(widget.OWWidget):
     #: A list of recent filenames.
     history = Setting([])
 
-    FILTER = "Pickle files (*.pickle *.pck)\nAll files (*.*)"
+
+    SUPPORTED_EXTENSIONS = ('.pkcls', '.pickle', '.pkl')
```
janezd: We're beginning to have too many extensions. Can we decide for just one? I like pkcls. Same for data, where we would use pkdata.

kernc: There's a thing called backward compatibility. Otherwise, I'm perfectly happy with preventing users from loading their once reliably stored trained classifiers, so .pkcls it is ...

PR #958, opened 2015-12-29
AdaBoost: Add to the scripting part for cls and reg

VesnaT:


BlazZupan:
AdaBoost and similar ensemble-based algorithms (like stacking, bagging, etc.) should probably be placed under a different folder. A proposal during Orange Friday was that that should be under Orange.ensembles. Could you please relocate? Ensemble algorithms would benefit from a callback function that would receive info on progress. If that is not available in sklearn, we may need to reimplement.

Orange/tests/test_ada_boost.py
```diff
@@ -0,0 +1,100 @@
...
+from Orange.data import Table
+from Orange.classification import TreeLearner
+from Orange.regression import TreeRegressionLearner
+from Orange.ensembles import SklAdaBoostLearner, SklAdaBoostRegressionLearner
+from Orange.evaluation import CrossValidation, CA, RMSE
+
+
+class SklAdaBoostTest(unittest.TestCase):
+    def test_adaboost(self):
+        table = Table("iris")
```
kernc: Could you load those tables in a `setUp()` method once and then use that? I know it's not uniform practice across our tests, but with new ones, we really can _make better_.

lanzagar:
Orange/__init__.py needs to be updated as well

ales-erjavec:
The new `Orange.ensembles` package must also be added to the setup.py's PACKAGES list

PR #959, opened 2015-12-29
Bolder colors for both Distributions graph and legend

ajdapretnar:
See task #231. @BlazZupan Check if this is what you had in mind. Legend now takes argument from overall color palette, but we might want to separate args for legend and plot.


lanzagar:
When making pull requests, make sure to start the new branch from an updated master if possible.
Your changes are applied to an old version of the code (Janez changed colors in the meantime) so there are conflicts that can't be merged automatically.

For the change itself, it is better if @BlazZupan says if this is what he wanted.

PR #963, opened 2016-01-04
SQL log: Improvements

lanzagar:


Orange/data/sql/table.py
```diff
@@ -704,10 +705,12 @@ def _execute_sql_query(self, query, param=None):
         connection = self.connection_pool.getconn()
         cur = connection.cursor()
         try:
+            utfquery = cur.mogrify(query, param).decode('utf-8')
+            sql_log.debug("Executing: {}".format(utfquery))
```
kernc: You don't need a timestamp with that?

Issue #948, opened 2015-12-22
Python 3.5 Windows 10


jamartinh:
Hi, this, line is causing me an error in python 3.5 windows 10 

lib = ctypes.pydll.LoadLibrary(os.path.join(path, "_grid_density" + sysconfig.get_config_var("SO")))

It seems this will happen in every import using this method.

The problem is that python 3.5 Anaconda has empty SO (".pyd") 

I can import the library just by importing _grid_density since the python import recognize the correct extension.

kernc:
How did you manage to install Orange with Anaconda Python 3.5?

jamartinh:
After installing requirements...

python setup.py develop

I have visual studio 2015 community.
64 bits python.

jamartinh:
Thanks !!!

kernc:
@jamartinh are you confirming it now works?

jamartinh:
Yes, it seems to work now!

2016-01-17 5:01 GMT+01:00 Kernc <notifications@github.com>:

> @jamartinh <https://github.com/jamartinh> are you confirming it now works?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/biolab/orange3/issues/948#issuecomment-172290068>.
>



-- 
/ .- .-.. .-.. / -.-- --- ..- / -. . . -.. / .. ... / .-.. --- ...- .
José Antonio Martín H. (PhD)           E-Mail: jamartinh@fdi.ucm.es
Computer Science Faculty               Phone: (+34) 91 3947650
Complutense University of Madrid       Fax: (+34) 91 3947527
C/ Prof. José García Santesmases,s/n   28040 Madrid, Spain
web: http://www.dacya.ucm.es/jam/
LinkedIn: http://www.linkedin.com/in/jamartinh (Let's connect)
.-.. --- ...- . / .. ... / .- .-.. .-.. / .-- . / -. . . -..

PR #967, opened 2016-01-11
extensions: Change how ctypes 'extensions' are located/loaded

ales-erjavec:
In Python 3.5 windows extension modules (.pyd) are ABI tagged, but
`sysconfig.get_config_var("EXT_SUFFIX"), contains the untaged suffix.

Import the actual stub modules and use the module's `__file__`
attribute to locate the library.

### DIFF
--- a/Orange/classification/simple_tree.py
+++ b/Orange/classification/simple_tree.py
@@ -1,5 +1,3 @@
-import os
-import sysconfig
 import ctypes as ct
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


kernc:
I belive this fixes #948.

Orange/classification/simple_tree.py
```diff
@@ -8,9 +7,8 @@
 __all__ = ['SimpleTreeLearner']
 
 path = os.path.dirname(os.path.abspath(__file__))
-
-_tree = ct.pydll.LoadLibrary(
-    os.path.join(path, "_simple_tree" + sysconfig.get_config_var("SO")))
+from . import _simple_tree
```
lanzagar: Looks like `import os` and `path = ...` could be removed in this file as well.

kernc:
Please rebase.

PR #971, opened 2016-01-11
Install add-ons in user's local site-packages

kernc:


lanzagar:
Orange still check if I have root privileges, shows an error and does not let me manage addons.
See [line 418](https://github.com/biolab/orange3/pull/971/files#diff-55a6a7f617395a75195203c03b442985R418).

kernc:
Gah, I meant to delete that! Thanks. How about now?

lanzagar:
Works for me, but not for people running it in virtualenv:

pip.exceptions.InstallationError: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.

I wanted to suggest this anyway and it might solve the problem as well:
Should we check if we have permission to install addons as previously (works in virtualenv and if run as root). That way an addon can be installed system-wide (useful if lots of users want to use it on a shared computer).
And if not, only then switch to the `--user` installation.

Orange/canvas/application/addons.py
```diff
@@ -565,6 +557,7 @@ def __on_installer_error(self, command, pkg, retcode, output):
 
     def __on_installer_finished(self):
         message_information(
+            "Changes successfully applied in <i>{}</i>.<br>".format(USER_SITE) +
```
lanzagar: This now prints that the changes have been made in the local user dir even if `user_install` is False and they were not.

lanzagar:
Although this pull request is probably not the most appropriate one to add Orange3-Associate the bigger question is - is this addon now finished and production ready enough that we want to have it explicitly listed?
All addons on pypi show up in the list anyway, right?

Orange/canvas/application/addons.py
kernc: Thanks, please see again.

kernc:
It's my opinion that Orange3-Associate is as good as ever. Anyone is welcome to try it out. :stuck_out_tongue_winking_eye:

PR #978, opened 2016-01-15
[WIP] Orange.data.Domain: reworked iterating, indexing, member access

kernc:
Summary of changes:
* `for var in data.domain` (`iter(domain)`) iterates over _all_ Variables in the domain, including meta variables. The are several arguments for this change:
  * String variables are also Variables. There are tickets about extending widgets' functionality to include meta features. Lots of widgets already use metas by `data.domain.variables + data.domain.metas` (40+ results in the codebase, excluding any `chain()` calls).
  * It is easy enough to filter out only the primitive variables with `... for var in data.domain if var.is_primitive` or apply just about any other condition.
  * `data.domain.variables` is still accessible and works as before (although it is deprecated as its utility is questionable and is yet to be assessed). 
  * This is consistent with `str()` of `Table` and `RowInstance` instances.
* Indexing and slicing is reworked. Most importantly, negative indexes no longer return metas but are just plain true Pythonic negative indexes, i.e. from the back. It works almost the same, except that the order of returned metas is reversed, e.g. in domain with three metas `domain[-1]` now is what `domain[-3]` was before. I doubt much code is relying on this.
* `Domain.anonymous` attribute is removed. It is used nowhere besides tests.
* `Domain.checksum()` is depreacted in favor of the equality operator.
* Some more basic `is_*`/`has_*` boolean members are converted to properties for consistency.
* Domain has a new constructor which acts mostly the same. The more significant of changes is that any non-primitive variables passed as `attributes` or `class_vars` are put into `metas` instead.

Please see individual commits.

It is expected for some tests to fail. I didn't want to touch tests or auxiliaries until the interface is agreed on.

Some relevant discussion: https://github.com/biolab/orange3-text/pull/23#discussion_r49006804

### DIFF
--- a/Orange/data/domain.py
+++ b/Orange/data/domain.py
@@ -7,6 +7,8 @@ import weakref
 from .variable import *
 import numpy as np
 
+from Orange.util import deprecated
+
 
 class DomainConversion:
     """
@@ -445,14 +447,13 @@ class Domain:
             col_idx = self.index(attr)
         return [attr], np.array([col_idx])
 
+    @deprecated
     def checksum(self):
         return hash(self)
 
     def __eq__(self, other):
-        if not isinstance(other, Domain):
-            return False
-
-        return (self.attributes == other.attributes and
+        return (isinstance(other, Domain) and
+                self.attributes == other.attributes and
                 self.class_vars == other.class_vars and
                 self.metas == other.metas)
 
--- a/Orange/data/domain.py
+++ b/Orange/data/domain.py
@@ -349,10 +349,22 @@ class Domain:
         """
         Return the index of the given variable (represented as
         `int`, `str`, or instance of :class:`Variable`) in the domain.
+
+        If `var` is an iterable of above values, then return three tuples:
+        indices in attributes, indices in class_vars, and indices in metas.
         """
         if isinstance(var, (int, np.integer)):
             return int(var)
         try:
+            if isinstance(var, Iterable):
+                vars = [self[val] for val in var]
+                Xind = tuple(self.attributes.index(v)
+                             for v in vars if v in self.attributes)
+                Yind = tuple(self.class_vars.index(v)
+                             for v in vars if v in self.class_vars)
+                Mind = tuple(self.metas.index(v)
+                             for v in vars if v in self.metas)
+                return Xind, Yind, Mind
             return self._indices[var]
         except KeyError:
             raise ValueError("Variable '{}' is not in domain".format(var))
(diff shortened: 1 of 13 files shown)


Orange/data/domain.py
```diff
@@ -7,6 +8,11 @@
 from .variable import *
 import numpy as np
 
+from Orange.util import deprecated
+
+
+log = logging.getLogger()
```
astaric: This shadows
from math import log
above (causing many failing tests)

Orange/data/domain.py
```diff
@@ -58,69 +69,101 @@ def __init__(self, source, destination):
 
 
 class Domain:
-    def __init__(self, attributes, class_vars=None, metas=None, source=None):
+    def __init__(self, attributes=[], class_vars=[], metas=[], source=None):
```
astaric: Please avoid using lists as default parameters. Even if it works in this case, it is a bad practice. If the value would be stored, all instances would share the **same** instance of the list.

kernc: [This](https://github.com/biolab/orange3/pull/978/files#diff-698efb0bca09486f1b8892fac56a8661R72) better? :smiley:

Orange/data/domain.py
```diff
@@ -58,69 +69,101 @@ def __init__(self, source, destination):
 
 
 class Domain:
-    def __init__(self, attributes, class_vars=None, metas=None, source=None):
+    def __init__(self, attributes=(), class_vars=(), metas=(), source=None):
```
astaric: Should we move the indices/source logic to a separate class method, which converts indices to variables and constructs a new domain (by calling its init)?

kernc: The beauty here is that one can construct superset domains by specifying only some `attributes` as `str`, for example, while leaving the others as `Variable`. Thus only the strings are sourced.

A separate `from_domain()` method would more strictly imply a domain subset, I guess, but I sure am not against it.

janezd: Specifying some variables as strings and others as variables does not make much sense in, say, widgets. Let `from_domain` be strict. However, this form of constructor is useful for lecturing since it allows you to refer to attributes by names when constructing new domains (and to also add new, computed attributes to the existing domain) in the command line, all without too much typing. So please leave this one as it is, but I sure am not against a stricter `from_domain`, either.

kernc: For strict domain subsets one can already do, IMHO preferred:
```py
domain[list_of_vars]
```
where
```py
features = ['Feature 1', 'Target']
new = domain[features]
```
retains correct `attributes`, `class_vars`, and `metas`, i.e.:
```py
assert len(new.attributes) == len(new.class_vars) == 1
```

astaric:
We might consider incorporating ideas from this pull request into the code base, but as it is, the PR touches too many places and breaks too many tests.

PR #981, opened 2016-01-18
Add RemoveNanClasses preprocessor

VesnaT:
Preprocessor checks for missing values in class variable and removes associated instances.


janezd:
@VesnaT Replace adult_sample with a smaller data set

nikicc:
Also could you rebase, so the big file gets out of history? Otherwise looks ok to me.

PR #983, opened 2016-01-19
Small fixups from here and there

kernc:


Orange/util.py
```diff
@@ -3,20 +3,25 @@
...
 from itertools import chain, count
 from collections import OrderedDict
-import logging
+import warnings
 
 import numpy as np
 
 
-log = logging.getLogger()
+warnings.simplefilter('default', DeprecationWarning)
```
kernc: IIUC, this should make e.g. below `DeprecationWarning` fire only once per each line where the deprecated function is used. Perhaps there is a more canonical place to install this filter instead of this module?

ales-erjavec: I think this would set the filter for all DeprecationWarnings not just the ones raised by `deprecated` (or was this the intention). I think you can just subclass a DeprecationWarning to be raised by the decorator and set the default filter on that.

kernc: Would filtering all DeprecationWarnings (once per line) actually be undesirable?

ales-erjavec: Yes. Someone might setup an elaborate warning filter rules (e.g. on subclasses of DeprecationWarning, only raised in some specific module, ...), which would be overridden simply by importing Orange.

kernc: Aye, thanks.

PR #985, opened 2016-01-21
Distance matrix widget

janezd:


lanzagar:
Selection behaves strangely:
- It remains in effect when the input changes. It should probably be reset.
- The first row/column is reserved for the labels and is unselectable. When Labels is set to None this causes problems: 1st row/column with distances is unselectable.
- Deselecting a cell (with Ctrl-click) removes the row and column completely instead of just that combination.
E.g. select (1,3), then select (1,5), then deselect (1,5) => only (5,5) remains instead of the expected state from before the select&deselect

lanzagar:
Also, could selections be marked in the first row/column with labels as well?
It would make it much easier to read the names of selected items, especially for submatrices somewhere in the middle.

Orange/widgets/unsupervised/owdistancematrix.py
```diff
@@ -0,0 +1,370 @@
...
+        if self.distances is not None:
+            self._update_labels()
+
+    def _update_labels(self):
+        var = column = None
+        if self.annotation_idx == 0:
+            labels = None
+        elif self.annotation_idx == 1:
+            labels = [str(i + 1) for i in range(self.distances.shape[0])]
+        elif self.annot_combo.model()[self.annotation_idx] == "Attribute names":
```
lanzagar: This can produce an
`IndexError: list index out of range`
when `self.annotation_idx` is loaded from settings and greater than the size of the model

To reproduce: load iris, connect to distances (rows), connect to dist matrix (label=iris), delete widget, change to column distances and connect again.

kernc:
> The first row/column is reserved for the labels and is unselectable.

Shouldn't these rather be horizontal and vertical [header](http://doc.qt.io/qt-4.8/qtableview.html#horizontalHeader) row and column?

janezd:
>> The first row/column is reserved for the labels and is unselectable.
> Shouldn't these rather be horizontal and vertical header row and column?

Conceptually yes, but `QHeaderView`'s cell can't be individually styled because `QHeaderView` does not respect `BackgroundRole` or `ForegroundRole` on all systems. It works, for instance, if I set the (whole) application style to plastique. The way it's done now involves lots of code, but I see no way to use `QHeaderView` (short of subclassing it and reimplementing `paint`).

janezd:
On the other hand, the first row and column are not frozen, so they scroll away. Background colors are nice, but do they really matter if the user doesn't see them? You may be right about using headers instead.

kernc:
One begins to favor semantics and function whenever possible.

janezd:
All fixed (including switching to headers), except:

> Deselecting a cell (with Ctrl-click) removes the row and column completely instead of just that combination. E.g. select (1,3), then select (1,5), then deselect (1,5) => only (3,3) remains instead of the expected state from before the select&deselect

The state of selection is what you see. If you select (1, 3) and (1, 5), items 1, 3 and 5 are selected. History about how this selection was constructed is not recorded (and not necessarily remembered by the user -- or even done by another user and sent to this one in saved schema). Therefore, deselecting (1, 5) when items {1, 3, 5} are selected must deselect both 1 and 3.

Also, what would be the expected behaviour in this case: select (1, 3), select(1, 5), deselect (3, 5)?

> Also, could selections be marked in the first row/column with labels as well?

Good idea, but now that the table uses standard headers this can't be done. `setHighlightSection(True)` highlights only columns/rows in which all cells are selected. `QHeaderView` also does not respect `BackgroundRole` and/or `ForegroundRole`.

@kernc: Function is favored over visual appearance, unless visualization *is* the function. It didn't work in this case, though.

markotoplak:
For me, this widget always sends selections as if it is always in "Auto send" mode. 

Empty selections (or just putting the widget on canvas) produce an Index Error in 

  File "/home/marko/orange3/Orange/misc/distmatrix.py", line 97, in submatrix
    obj = self[np.ix_(row_items, col_items)]
IndexError: arrays used as indices must be of integer (or boolean) type

janezd:
@markotoplak, the cause of the first bug is interesting. Signals should not be connected to `commit` function before the call to `gui.auto_commit`.

I couldn't replicate the second bug, but I guess I fixed it. Please check.

PR #987, opened 2016-01-22
Time variable

kernc:
RFC. Docs missing.

### DIFF
--- a/Orange/widgets/utils/__init__.py
+++ b/Orange/widgets/utils/__init__.py
@@ -1,10 +1,13 @@
 from functools import reduce
+from Orange.data.variable import TimeVariable
 
 
 def vartype(var):
     if var.is_discrete:
         return 1
     elif var.is_continuous:
+        if isinstance(var, TimeVariable):
+            return 4
         return 2
     elif var.is_string:
         return 3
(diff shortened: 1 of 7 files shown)


Orange/widgets/utils/__init__.py
```diff
@@ -1,13 +1,16 @@
...
 def vartype(var):
     if var.is_discrete:
         return 1
-    elif var.is_continuous:
+    elif var.is_continuous and not isinstance(var, TimeVariable):
         return 2
     elif var.is_string:
         return 3
+    elif isinstance(var, TimeVariable):
+        return 4
```
lanzagar: Not really important, but:

If we want this if-else flat and ordered, then I am in favour of changing it to go from 4 to 0.
TimeVariable (and future new types) can be added to the top and any unhandled variables are caught at the bottom as generic variable types.

Another option would be a nested if-else with all continuous types added inside the elif is_continuous. This shows the structure better, but loses the ordering and becomes more complex.

kernc: This abomination of a mechanism shall not retain its current form for much longer. But mending it now is out of scope in this PR.

lanzagar:
Looks good.
Will documentation be added as well?

kernc:
Some docs have been added.

PR #988, opened 2016-01-22
IO fixups

kernc:


Orange/tests/test_xlsx_reader.py
```diff
@@ -49,10 +47,10 @@ def test_no_flags(self):
...
         self.assertEqual(domain[0].values, ["green", "red"])
         np.testing.assert_almost_equal(table.X,
-                                       np.array([[1, 0.5, 0],
-                                                 [1, 0.1, 0],
-                                                 [0, 0, float("nan")]]))
-        np.testing.assert_almost_equal(table.Y, np.array([21, 123, 0]))
+                                       np.array([[1, 0.5, 0, 21],
+                                                 [1, 0.1, 0, 123],
+                                                 [0, 0, np.nan, 0]]))
+        np.testing.assert_almost_equal(table.Y, np.array([]).reshape(3, 0))
```
lanzagar: An empty array will probably be _exactly_ equal?

kernc: Thanks, gosh, I almost missed that! Great catch!

PR #989, opened 2016-01-25
OWColor: Add report

VesnaT:


janezd:
This report is not very useful. I made a PR to your PR, which puts more info in the report. See https://github.com/VesnaT/orange3/pull/10

PR #993, opened 2016-01-27
Unknown str value fix

ales-erjavec:
* Change `Value` constructor to better handle missing values for StringVariables
* Update the StringVariable.Unknown constant (now a `""`)
* Change tests which which assumed a single unknown value representation
* Fix `Value.__hash__` and `Value.__contains__`


kernc:
Great, thanks for catching all that. :smiley: We could avoid these types of bugs with a simpler, understandable, explainable data model.

But one works with what one has.

PR #995, opened 2016-01-28
[WIP] RFC: New Table base interface

kernc:
Brace yourself.

**TLDR**: This is the new, proposed `Table` class. There is a proof-of-concept code at the bottom. Obviously, it only works with `Domain` from https://github.com/biolab/orange3/pull/978 if you'd care to test it.

The idea is to have fairly simple and uniform interfaces to all supported data backends (herein called drivers; currenly NumpyDriver, SqlDriver, SparseDriver; neither yet implemented in full), and have the new `Table` (and its child classes) only wrap one of these drivers. This makes the Table interface dependable and useful regardless of which driver is used (i.e. whether `data` is an instance of `Table`, `Corpus(Table)`, or `Relation(Table)`, it is independent of the underlying format/medium the data is stored in). The goal is to make sparse tables really compatible as well as most of the multiple (No)SQL backends through [PyDAL](http://pydal.readthedocs.org/en/latest/), [SQLAlchemy Core](http://docs.sqlalchemy.org/en/latest/core/), or [Blaze](http://blaze.pydata.org/).

Think of a future where Bag of Words widget returns a sparse array and nobody cares and _nothing_ breaks. That's the general idea. Also ease of use.

All of this, design-wise, is RFC. Please gently let me know where I needlessly complicated stuff and if something crucial is missing. :seedling: 

`Ellipsis` is sometimes used as a TODO marker, but not always.

It's all mostly compatible. As a major incompatible difference that comes to mind, the new table objects are read-only.

There is a `NumpyDriver`-based proof-of-concept runnable code at the bottom.

### DIFF
--- a/Orange/data/base.py
+++ b/Orange/data/base.py
@@ -1,0 +1,625 @@
+from collections.abc import Sequence, Iterable
+
+import numpy as np
+from scipy.sparse import issparse
+
+from Orange.data import Domain, Variable
+from Orange.util import deprecated, Registry, abstract
+
+
+class Expression:
+    """
+    Roughly replaces Orange.data.filter.Filter.
+    """
+    def __init__(self, e1, op=None, e2=None):
+        self.e1 = e1
+        self.op = op
+        self.e2 = e2
+
+    def __str__(self):
+        return '({} {} {})'.format(self.e1, self.op, self.e2)
+
+    def __and__(self, other):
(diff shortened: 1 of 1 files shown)


astaric:
Implementing yet another file representation from scratch is likely not a good idea.. If we were to migrate to a new implementation of a DataTable, an existing library (such as Pandas) would probably be a better choice.

PR #997, opened 2016-01-31
OWDistances, SklDistance: support discrete variables

kernc:
Distances widget traditionally stripped discrete variables, requiring an additional Continuize step, which finally would make Distances favor fewer-valued features. Now, it counts distance between two values of a discrete variable as binary, either 1 (different) or 0 (same).

### DIFF
--- a/Orange/distance/__init__.py
+++ b/Orange/distance/__init__.py
@@ -19,11 +19,34 @@ def _preprocess(table):
     return new_data
 
 
-def _orange_to_numpy(x):
+def one_hot_encode(domain, X):
+    """
+    Encode discrete Orange.data.Variables into one binary column for each value.
+    Return a dummy table.
+    """
+    if not domain.has_discrete_attributes():
+        return X
+    cols = []
+    for i, (var, x) in enumerate(zip(domain.attributes, X.T)):
+        if var.is_discrete:
+            n_values = len(var.values)
+            for val in range(n_values):
+                # Make the diffs across cols of same variable sum to 1
+                cols.append((x == val) / n_values)
+        else:
+            cols.append(x)
+    return np.column_stack(cols)
+
+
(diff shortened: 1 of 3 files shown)


janezd:
@kernc, thanks for this PR, Distances are now much better. (I also like the trick you used.) Can you rebase, please? It seems there should be no conflicts.

lanzagar:
This definitely makes distances much better / more correct.

We probably want to have an option in Continuize that converts a discrete variable into [0, 1/k] variables instead of just [0, 1] and [-1, 1] options currently available.
So the `one_hot_encode` function could be moved/merged into the continuize module in a future pull request.

lanzagar:
Actually, upon closer inspection this looks like it does not work as desired.

If 7-valued categorical var gets replaced with 7 variables containing 0 or 1/7, the distance between two different values is not 1 (but e.g. 2/7 for Manhattan).
I am in favour of using discrete variables instead of ignoring them, but we need to reconsider how. Maybe the simplest choice of just automatically applying the Continuize preprocessor is best?

janezd:
I'm in favour of writing our own classes for computation of distances. The code is trivial to write in Cython, and it will allow us to treat missing and discrete values correctly, and to normalize continuous values, all on the fly without having to rewrite the data three times. At the same time, it would be cleaner and require less work than the current patches.

janezd:
@lanzagar `Continuize` (PR #1019) does not solve the problem: attributes with multiple values are still treated incorrectly. This has to be built in the computation of distances.

lanzagar:
Well smart continuization (distance specific) would solve the problem, but I agree that is a lot of work and we might as well just reimplement distances and have a cleaner solution.

E.g. for Manattan continuize with 0.5 instead of 1 in the appropriate column would be correct, and for Euclidean sqrt(2)/2. But for others it becomes trickier.

PR #1002, opened 2016-02-03
Added UCI dermatology dataset

doblak:


kernc:
UCI site says:

> Every other feature (clinical and histopathological) was given a degree in the range of 0 to 3. Here, 0 indicates that the feature was not present, 3 indicates the largest amount possible, and 1, 2 indicate the relative intermediate values. 

Thus the feature set sounds more of an ordinal (our continuous) than a categorical (our discrete) one?

Also, you can remove the second commit. Newlines at the end of files don't really hurt anyone.

doblak:
@kernc 

Thanks for reviewing the pull request, you are right about the attributes. I pushed a new commit, there are 33 continuous attributes (1 for age and 32 for 0-3 degree ratings) and 1 discrete (whether any of diseases was observed in family history).

Will be more careful in the future as I plan to add more datasets.

PR #1003, opened 2016-02-03
OWTreeGraph: Prune node hover tooltip

VesnaT:
Text was redundant when split by the same attribute more than once.

### DIFF
--- a/Orange/widgets/classify/owclassificationtreegraph.py
+++ b/Orange/widgets/classify/owclassificationtreegraph.py
@@ -1,5 +1,5 @@
 import sys
-
+from collections import OrderedDict
 import numpy
 
 from sklearn.tree._tree import TREE_LEAF
@@ -286,7 +286,7 @@ class TreeNode(GraphicsNode):
                     else "≠ " * is_left_child + values[parent_attr_cv.value]
             else:
                 thresh = self.tree.threshold[self.parent.i]
-                return "%s %s" % ([">", "<="][is_left_child],
+                return "%s %s" % ([">", "≤"][is_left_child],
                                   attribute.str_val(thresh))
         else:
             return ""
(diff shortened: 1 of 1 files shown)


Orange/widgets/classify/owclassificationtreegraph.py
```diff
@@ -323,10 +323,19 @@ def rulew(self, i=0):
...
                 sign = [">", "<="][is_left_child]
                 value = "%.3f" % self.tree.threshold[self.parent.i]
-            pr.append((attr_name, sign, value))
+            if (attr_name, sign) in pr:
+                old_val = pr[(attr_name, sign)]
+                if sign == ">":
+                    pr[(attr_name, sign)] = max(float(value), float(old_val))
+                elif sign == "≠":
+                    pr[(attr_name, sign)] = "{}, {}".format(old_val, value)
+                elif sign == "<=":
```
kernc: So we use Unicode "not equal to" symbol but ASCII "less-than or equal to" construct?

:-1: for consistency.

Use `≤`?

janezd:
Can you sign the CLA, so that the PR will pass all checks?

Orange/widgets/classify/owclassificationtreegraph.py
```diff
@@ -321,12 +321,21 @@ def rulew(self, i=0):
...
+                elif sign == "≠":
+                    pr[(attr_name, sign)] = "{}, {}".format(old_val, value)
+                elif sign == "≤":
+                    pr[(attr_name, sign)] = min(float(value), float(old_val))
+            else:
+                pr[(attr_name, sign)] = value
             return pr
         else:
-            return []
+            return {}
```
janezd: Should this be `OrderedDict`?

PR #1006, opened 2016-02-04
loader hack for python3

vatula:
Old hack for python 3 seems not to work any more.

A hack seems not to work for `python 3.5` and orange-canvas fails to start (throws) when `pkg_resources.__init__.py` calls resource_exists with orange.qss when loader is commented out it all works

### DIFF
--- a/Orange/canvas/__init__.py
+++ b/Orange/canvas/__init__.py
@@ -1,6 +1,3 @@
-# pkg_resources selects a suitable resource loader based on the modules's
-# __loader__ attribute. In python2 it defaults to None, which maps to the
-# default resource loader, but it python3 it does not. As a result,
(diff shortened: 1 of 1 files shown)


kernc:
@astaric I have absolutely no idea about implications of this change or why it was there in the first place. I tested it on Linux and Windos, and the icons (e.g. in Preprocess widget) and add-ons (both use `pkg_resources`) still work. If this works on Mac too, you are welcome to do with it whatever. :+1:

astaric:
This fix was added when setuptools still had a single digit version, and might have been unnecessary for a long time. Everything works on OSX as well after applying this, so I am merging it.

kernc:
Thanks @vatula for finding this too!

PR #1009, opened 2016-02-05
Confusion matrix: Add shading of cells

janezd:


nikicc:
Currently, most entries are white. Maybe we could color the diagonal elements in green and off-diagonal in red (or something similar) and then scale each color separately?

janezd:
Diagonal is now green and colors correspond to values normalized by dividing with the highest number on the diagonal.

Non-diagonal cells are red and normalized by dividing with the maximum of all non-diagonal elements, maximum within rows, or maximum within columns, depending upon the settings.

kernc:
> ... green ... red ...

Did you account for [red-green color blindness](https://www.google.com/search?num=20&q=red-green+color+blindness+percent)?

nikicc:
Looks nicer to me. However, there is still a minor bug in calculation of lightness. For example, for Majority on Iris when looking at proportion of actual we set the lightness to `-1245` which causes cells to be completely black:
screen shot 2016-02-07 at 18 00 55

janezd:
The second commit (report, https://github.com/janezd/orange3/commit/a3295a7a62ceb993dbdd937c549cf0a119387193) is blocked by PR #1016. 

@kernc, I would personally prefer red vs. green, but I changed green to blue.

@nikicc Thanks, I fixed the normalization.

nikicc:
I also prefer red-green, but if this friendlier to a larger audience, we can keep it as it is.
@janezd can you please reabse?

PR #1010, opened 2016-02-05
[FEATURE] Rank outputs scores

astaric:
Output scores from Rank widget in a Table and allow them to be used as annotations in MDS widget.

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -56,7 +56,7 @@ class OWRank(widget.OWWidget):
 
     inputs = [("Data", Orange.data.Table, "setData"),
               ("Scorer", score.Scorer, "set_learner", widget.Multiple)]
-    outputs = [("Reduced Data", Orange.data.Table),
+    outputs = [("Reduced Data", Orange.data.Table, widget.Default),
                ("Scores", Orange.data.Table)]
 
     SelectNone, SelectAll, SelectManual, SelectNBest = range(4)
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owrank.py
```diff
@@ -488,6 +494,17 @@ def selectedAttrs(self):
...
 
+    def create_scores_table(self):
+        features = [ContinuousVariable(s[0]) for s in self.measures]
+        metas = [StringVariable("Feature names")]
+        domain = Domain(features, metas=metas)
+
+        scores = np.array(self.measure_scores).T
+        feature_names = np.array([a.name for a in self.data.domain.attributes])
+        # Reshape to 2d array as Table does not like 1d arrays
+        feature_names = feature_names[:, None]
```
kernc:
```py
feature_names = np.atleast_2d([a.name for a in self.data.domain.attributes])
```

astaric: Creates a row instead of a column, and I dislike the .T. The optimal way would be for Table to do this transformation instead.

BlazZupan:
Should output scores for all features (independently of selection of attributes). Now outputs none if under some circumstances. Try: put select attributes on manual, build schema File->Rank->Table (Scores-Data). The Table is empty. Now select one feature in Rank and the table fills up. Same goes if now select None and then Manual and then select None.

janezd:
@BlazZupan Should "Reduced data" be the default output for the Rank widget? I guess most users will need this signal and not the table with ranked features, which is not that useful. So, we may want to avoid showing the dialog with signal connections.

janezd:
Output the table at recompute, not on commit.

PR #1011, opened 2016-02-05
KNNLearner returns KNNClassifier

doblak:
Running for example:

`Orange.evaluation.testing.CrossValidation(Orange.data.Table(primary-tumor), [Orange.classification.knn.KNNLearner()], k=5)`

throws an error in Orange.evaluation.testing.py:338: 
`ValueError: could not broadcast input array from shape (71,20) into shape (71,21)`

The fix in pull request fixes it (the result gets extended to proper dimensions according to domain classes if testing set doesn't cover all of them). Hopefully that doesn't break anything else. Please review.

### DIFF
--- a/Orange/classification/knn.py
+++ b/Orange/classification/knn.py
@@ -1,11 +1,14 @@
 import sklearn.neighbors as skl_neighbors
-from Orange.classification import SklLearner
+from Orange.classification import SklLearner, SklModel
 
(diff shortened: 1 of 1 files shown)


lanzagar:
Thank you for bringing attention to this problem.
While your change does fix KNNLearner, a more general fix is available in #1018.

doblak:
Even better, thank you.

PR #1012, opened 2016-02-05
[WIP] File Widget: Add table with names and types

janezd:
@astaric This is basically done; it constructs the output domain -- it just does not apply it yet (see the comment). You can implement the settings.

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -1,5 +1,5 @@
 import os
-from itertools import chain
+from itertools import chain, count
 from warnings import catch_warnings
 
 from PyQt4 import QtGui, QtCore
@@ -659,9 +659,11 @@ class OWFile(widget.OWWidget):
         attributes = []
         class_vars = []
         metas = []
-        places = [attributes, class_vars, metas]
-        for (name, tpe, place, vals), orig_var in \
-                zip(self.tablemodel.variables,
+        weight =[]
+        places = [attributes, class_vars, metas, weight]
+        cols = [[], [], [], []]  # Xcols, Ycols, Mcols, Wcols
+        for column, (name, tpe, place, vals), orig_var in \
+                zip(count(), self.tablemodel.variables,
                     chain(self.data.domain.variables, self.data.domain.metas)):
             if place == 3:
                 continue
@@ -670,7 +672,9 @@ class OWFile(widget.OWWidget):
             else:
                 var = tpe(name)
             places[place].append(var)
+            cols[place].append(column)  # This assumes that no columns were skipped in the original file!
         domain = Domain(attributes, class_vars, metas)
+        data = Table.from_file(self.loaded_file, domain=domain, columns=cols)
         self.apply_button.hide()
 
 
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -8,7 +8,7 @@ from PyQt4.QtGui import QSizePolicy, QTableView
 from Orange.widgets import widget, gui
 from Orange.widgets.data.owcolor import HorizontalGridDelegate
 from Orange.widgets.settings import Setting
-from Orange.widgets.utils.itemmodels import PyListModel
+from Orange.widgets.utils.itemmodels import PyListModel, TableModel
 from Orange.data.table import Table, get_sample_datasets_dir
 from Orange.data.io import FileFormat
 from Orange.data import \
@@ -159,15 +159,27 @@ class VarTableModel(QtCore.QAbstractTableModel):
         self.variables = []
 
     def set_domain(self, domain):
-        self.emit(QtCore.SIGNAL("modelAboutToBeReset()"))
+        def may_be_numeric(var):
+            if var.is_continuous:
+                return True
+            if var.is_discrete:
+                try:
+                    [float(x) for x in var.values]
+                    return True
+                except ValueError:
+                    return False
+            return False
+
+        self.modelAboutToBeReset.emit()
         self.variables = self.original = [
             [var.name, type(var), place,
-             ", ".join(var.values) if var.is_discrete else ""]
+             ", ".join(var.values) if var.is_discrete else "",
+             may_be_numeric(var)]
             for place, vars in enumerate(
                 (domain.attributes, domain.class_vars, domain.metas))
             for var in vars
         ]
-        self.emit(QtCore.SIGNAL("modelReset()"))
+        self.modelReset.emit()
 
     def rowCount(self, parent):
         return 0 if parent.isValid() else len(self.variables)
@@ -189,8 +201,11 @@ class VarTableModel(QtCore.QAbstractTableModel):
             if col == 1:
                 return gui.attributeIconDict[self.vartypes.index(val) + 1]
         if role == QtCore.Qt.ForegroundRole:
-            if self.variables[row][3] == 3:
-                return QtGui.QColor(208, 208, 208)
+            if self.variables[row][2] == 3 and col != 2:
+                return QtGui.QColor(160, 160, 160)
+        if role == QtCore.Qt.BackgroundRole:
+            place = self.variables[row][2]
+            return TableModel.ColorForRole.get(place, None)
 
     def setData(self, index, value, role):
         row, col = index.row(), index.column()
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owfile.py
```diff
@@ -142,6 +146,126 @@ def __repr__(self):
...
+    name2type = dict(zip(typenames, vartypes))
+    type2name = dict(zip(vartypes, typenames))
+
+    def __init__(self, widget):
+        super().__init__()
+        self.widget = widget
+        self.variables = []
+
+    def set_domain(self, domain):
+        self.emit(QtCore.SIGNAL("modelAboutToBeReset()"))
```
kernc: Please use new-style signals: `self.modelAboutToBeReset.emit()`

Orange/widgets/data/owfile.py
```diff
@@ -142,6 +146,126 @@ def __repr__(self):
...
+        self.variables = self.original = [
+            [var.name, type(var), place,
+             ", ".join(var.values) if var.is_discrete else ""]
+            for place, vars in enumerate(
+                (domain.attributes, domain.class_vars, domain.metas))
+            for var in vars
+        ]
+        self.emit(QtCore.SIGNAL("modelReset()"))
```
kernc: `self.modelReset.emit()`

Orange/widgets/data/owfile.py
```diff
@@ -142,6 +146,142 @@ def __repr__(self):
...
+    name2type = dict(zip(typenames, vartypes))
+    type2name = dict(zip(vartypes, typenames))
+
+    def __init__(self, widget):
+        super().__init__()
+        self.widget = widget
+        self.variables = []
+
+    def set_domain(self, domain):
+        def may_be_numeric(var):
```
janezd: This function assumes that string variables cannot be numeric: the recognition process is such that if all values can be interpreted as numbers the attribute is numeric. Is this OK or should we scan the actual attribute values in the table?

Orange/widgets/data/owfile.py
```diff
@@ -476,6 +641,24 @@ def workflowEnvChanged(self, key, value, oldvalue):
...
+                zip(self.tablemodel.variables,
+                    chain(self.data.domain.variables, self.data.domain.metas)):
+            if place == 3:
+                continue
+            if name == orig_var.name and tpe == type(orig_var):
+                var = orig_var
+            else:
+                var = tpe(name)
+            places[place].append(var)
+        domain = Domain(attributes, class_vars, metas)
```
janezd: @kernc @astaric Now what? Read the file with faked headers? This may not work with all readers, and it is undesired if the data is downloaded from the web. The other option is to reinterpret the existing data in numpy. Do we already have something to help with this?

Orange/widgets/data/owfile.py
kernc: Sorry, I don't understand. Where do you use the result of `may_be_numeric()`?

Orange/widgets/data/owfile.py
```diff
@@ -142,6 +146,142 @@ def __repr__(self):
...
+        self.widget = widget
+        self.variables = []
+
+    def set_domain(self, domain):
+        def may_be_numeric(var):
+            if var.is_continuous:
+                return True
+            if var.is_discrete:
+                try:
+                    [float(x) for x in var.values]
```
kernc: It may not matter much in this case, since `len(var.values)` is usually small, but in general, always prefer avoiding building unnecessary intermediary lists, i.e., `sum(float(x) for x in var.values)`.

Orange/widgets/data/owfile.py
```diff
@@ -142,6 +146,142 @@ def __repr__(self):
...
+                    return True
+                except ValueError:
+                    return False
+            return False
+
+        self.modelAboutToBeReset.emit()
+        self.variables = self.original = [
```
kernc: Where is `self.original` used?

Orange/widgets/data/owfile.py
kernc: [Sure](http://goo.gl/9lN72Q).

Orange/widgets/data/owfile.py
janezd: `self.original` was intended to (and still can) be used to determine whether we need to construct a new domain (if `self.variables != self.original`).

Orange/widgets/data/owfile.py
janezd: I expected this. :) I wrote "Do we **already** have ...", though.

kernc: If you just supply the new domain in `Table.from_table()`, it doesn't work?

janezd: I don't see how it could: if I change a name and type of variable and move it from attributes to meta attributes ... it cannot guess what to move where.

kernc: Ahh, you change the name too. Tough.

astaric: Table.from_numpy, called with new domain and original data?

janezd:
@astaric had this nice idea to construct transformations for variables (`StringVariable` -> `DiscreteVariable`) etc, convert the read data and then switch the domain to a new one (without transformations). I started implementing that but ... Consider the transformation from continuous to string: the result won't be the same string as in the original file. The file can contain some encoded categories (e.g. 1.1, 1.2, 1.3, 1.4, 2.1, 2.2, 3.1, 3.2, 3.3 ..., 3.9, 3.10, 3.11). We would interpret this as floats with two decimals. When converting this data to strings, we would get 1.10, 1.20 ... Hence, the data cannot be converted after it's read because the information is already lost at that point.

I thought about fake headers. This doesn't work either since the File widget doesn't know the file format. And it's a hack anyway.

The only clean and viable solution seems to be for the file readers to accept two additional arguments: `Domain` and mapping from column indices to domain variables (e.g., the third column becomes the second meta attribute). Current file readers already include some of this. @kernc, how difficult would it be to allow the caller to skip parsing of headers and use the pre-constructed domain and column-to-variable mappings?

janezd:
More specifically, @kernc, how difficult would it be to adapt the readers to support the last line in

```
        attributes = []
        class_vars = []
        metas = []
        weight =[]
        places = [attributes, class_vars, metas, weight]
        cols = [[], [], [], []]  # Xcols, Ycols, Mcols, Wcols
        for column, (name, tpe, place, vals), orig_var in \
                zip(count(), self.tablemodel.variables,
                    chain(self.data.domain.variables, self.data.domain.metas)):
            if place == 3:
                continue
            if name == orig_var.name and tpe == type(orig_var):
                var = orig_var
            else:
                var = tpe(name)
            places[place].append(var)
            cols[place].append(column)  # This assumes that no columns were skipped in the original file!
        domain = Domain(attributes, class_vars, metas)
        data = Table.from_file(self.loaded_file, domain=domain, columns=cols)
```

astaric:
As I see it, we have three options for domain override.

1. We can pass it as domain and list of columns
2. We can pass it as a list of tuples, containing values that would usually be found in the three header rows (i.e. [(name, type, flags)]
3. Same as 2., but using a class (VariableDescription) to store the values. We could use this class in variables list in the model. Furthermore, reader could (should?) be extended with a method that reads the header and returns a list of variable descriptions (including ignored/skipped variables), which can be displayed in the model.

I prefer 3.
Thoughts?

janezd:
I vote for 3.

The reader currently

1. reads data as strings,
2. reads headers,
3. scans this raw data to determine the variable types, when needed,
4. transforms raw data into `Table`.

Step 1 is separate, 2-4 are packed in `FileFormat.data_table`.

If the File widget (or another caller) provides VariableDescriptions to `Table.from_file`, it would skip 2. and 3., but still needs to read the data from file again since it's not stored anywhere. Th. We can avoid that by refactoring the reader into separate functions.

Another issue are columns that are skipped by the reader due to the ignore flag. The File widget would not see these columns, since it only has the `Domain`. Hence when the File widget calls the reader again, the reader would have to skip those columns that it skipped initially, plus the columns that the user decided to skip in the File widget. To do this, the reader would need to parse the headers again.

It would be better to have separate functions for

1. reading headers and raw data + a function that selects the proper reader based on the file extension
2. parsing headers and returning `VariableDescriptors`; `VariableDescriptor` includes the variable type (e.g. `StringVariable`) or `None` if unknown yet.
3. scanning the data; the function gets the raw data and a list of `VariableDescriptors`, and augments the descriptors in-place
4. transforming the raw data into data table using the given `VariableDescriptors`.

All of this already exists (except for `VariableDescriptors`) so it just needs some refactoring. I could do this, but I think @kernc will do it better. If we agree about this, of course.

janezd:
Rebased/refactored in #1199.

PR #1014, opened 2016-02-07
gui.py: Add verticalBox, horizontalBox

janezd:
Most calls of `gui.widgetBox` have argument `orientation="horizontal"` or `orientation="vertical"`, which is annoyingly long and uses strings instead of constants.

Functions `gui.verticalBox` and `gui.horizontalBox` call `gui.widgetBox` and supply the `orientation` argument.

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -552,6 +552,14 @@ def widgetBox(widget, box=None, orientation='vertical', margin=None, spacing=4,
     return b
 
 
+def hBox(*args, **kwargs):
+    return widgetBox(orientation="horizontal", *args, **kwargs)
+
+
+def vBox(*args, **kwargs):
+    return widgetBox(orientation="vertical", *args, **kwargs)
+
+
 def indentedBox(widget, sep=20, orientation="vertical", **misc):
     """
     Creates an indented box. The function can also be used "on the fly"::


Orange/widgets/gui.py
```diff
@@ -552,6 +552,14 @@ def widgetBox(widget, box=None, orientation='vertical', margin=None, spacing=4,
     return b
 
 
+def horizontalBox(*args, **kwargs):
+    return widgetBox(orientation="horizontal", *args, **kwargs)
+
+
+def verticalBox(*args, **kwargs):
```
kernc: Congratulations!

Please make these [`hBox` and `vBox`](https://www.google.com/search?q=hbox+vbox).

janezd:
This may be sarcasm (priors are high), but I'll bite. Changed to `vBox`, `hBox`.

Orange/widgets/gui.py
```diff
@@ -552,6 +552,14 @@ def widgetBox(widget, box=None, orientation='vertical', margin=None, spacing=4,
     return b
 
 
+def hBox(*args, **kwargs):
+    return widgetBox(orientation="horizontal", *args, **kwargs)
```
kernc: Sorry, didn't see this before. I don't think this syntax is correct. Positional arguments can't follow keyword arguments.
Should be `return widgetBox(*args, orientation='horizontal', **kwargs)`.

janezd: Both are OK.
```
>>> def f(*args, **kwargs):
...     print(args, kwargs)
...
>>> def g(*args, **kwargs):
...     f(*args, x=42, **kwargs)
...
>>> g(1, 2, u=13)
(1, 2) {'u': 13, 'x': 42}
>>>
>>>
>>> def g(*args, **kwargs):
...     f(x=42, *args, **kwargs)
...
>>> g(1, 2, u=13)
(1, 2) {'u': 13, 'x': 42}
```

I prefer giving the remaining arguments at the end.

PR #1015, opened 2016-02-07
Change the URL line edit in the File widget back to combo box

janezd:
URL Combo is again editable.

For Google sheets, the combo shows the last known sheet name for this URL (the underlying model maps the URL through dictionary of known names). In the line edit, the URL is shown as placeholder. The user is not expected to edit Google sheet URLs.

Other URLs are editable since they may be human friendly. Also, other URLs don't have any associated names so they're not mapped.

Alternative is that Google sheets get a separate radio button in which the user can either enter a new URL or select one from history. In this case, we could also show additional data, such as time stamps, last editors, previous versions... if available.

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -552,6 +552,14 @@ def widgetBox(widget, box=None, orientation='vertical', margin=None, spacing=4,
     return b
 
 
+def hBox(*args, **kwargs):
+    return widgetBox(orientation="horizontal", *args, **kwargs)
+
+
+def vBox(*args, **kwargs):
+    return widgetBox(orientation="vertical", *args, **kwargs)
+
+
 def indentedBox(widget, sep=20, orientation="vertical", **misc):
     """
     Creates an indented box. The function can also be used "on the fly"::
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -2,7 +2,7 @@ import os
 from warnings import catch_warnings
 
 from PyQt4 import QtGui, QtCore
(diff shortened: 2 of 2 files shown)


Orange/widgets/gui.py
```diff
@@ -552,6 +552,14 @@ def widgetBox(widget, box=None, orientation='vertical', margin=None, spacing=4,
     return b
 
 
+def horizontalBox(*args, **kwargs):
```
kernc: Already in https://github.com/biolab/orange3/pull/1014.

kernc:
I think I may, thanks. But if there is a use case for such a QWidget outside our File widget (e.g. somewhere else where files/URLs are loaded, or even elsewhere), I'd much prefer to see the former reusably decoupled from the latter.

Orange/widgets/gui.py
janezd: This branch was not based on master but on the one which I pushed into PR #1014. It thus contained the same commit that's already there (it no longer does, since I amended that commit).

I thought that merging this PR into master would skip the commit that's already in the master. Would it? Is this an appropriate way of basing your work on your other branches that were not merged yet?

kernc: I think so. ¯\_(ツ)_/¯

lanzagar:
The (minimum) size of the the URL box seems not to be set? It stretches based on the contents, but this can be too small and it does not resize later. Also with cleared settings (first use) it is completely too small.
Since the whole widget is small anyway it could probably just stretch the entire available area.

janezd:
Fixed.

I don't remember it looking like this. Perhaps I've never seen an empty combo.

PR #1016, opened 2016-02-07
Report: Fix crash when table item's DisplayRole data is None

janezd:


Orange/canvas/report/report.py
```diff
@@ -189,15 +189,18 @@ def report_table(self, name, table=None, header_rows=0, header_columns=0,
         name, table = self._fix_args(name, table)
         join = "".join
 
+        def data(item):
+            return item and item.data(Qt.DisplayRole) or ""
```
janezd: This condition assumes that `item.data` returns a string or `None`. http://doc.qt.io/qt-4.8/qt.html#ItemDataRole-enum

PR #1019, opened 2016-02-12
Revert "OWDistances, SklDistance: support discrete variables"

kernc:
See also:
https://github.com/biolab/orange3/pull/997#issuecomment-183279100

### DIFF
--- a/Orange/distance/__init__.py
+++ b/Orange/distance/__init__.py
@@ -19,34 +19,11 @@ def _preprocess(table):
     return new_data
 
 
-def one_hot_encode(domain, X):
-    """
-    Encode discrete Orange.data.Variables into one binary column for each value.
-    Return a dummy table.
-    """
-    if not domain.has_discrete_attributes():
-        return X
-    cols = []
-    for i, (var, x) in enumerate(zip(domain.attributes, X.T)):
-        if var.is_discrete:
-            n_values = len(var.values)
-            for val in range(n_values):
-                # Make the diffs across cols of same variable sum to 1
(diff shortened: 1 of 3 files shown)


janezd:
This is wronger :) than before: it will treat a variable with n values as n variables, so its maximal contribution to the distance will be n.

We will have to write our own classes instead of using the SKL's. Distances are just one example where SKL is useless. Trees are another. SKL is too narrow minded. There's no point in trying to patch it; we couldn't properly patch it for trees and neither can we patch it for distances - without having more work than writing everything from scratch.

kernc:
Please find revised.

lanzagar:
I agree we need to rethink this and probably reimplement some methods.
So let's revert this for now (before updating the stable branch), and we will tackle it in the coming cycles.

lanzagar:
> This is wronger :) than before: it will treat a variable with n values as n variables, so its maximal contribution to the distance will be n.

This however is not true (at least not for manhattan and euclidean). The n variables will always be 0 for both data instances in at least n-2 cases so those do not contribute anything.

Default continuization would result in doubled Manhattan distances regardless of the number of values in a discrete var. Which is better than using 1/k which is dependent on the number of values and leads to smaller contributions from vars with more values.
Similar for Euclidean (with squares and a root).

janezd:
You're right, it's 2, and it's not a function of the number of values. It's still 2, not 1. Use 1/2 instead of 1/k for Manhattan and sqrt(1/2) for Euclidean? :)

PR #1023, opened 2016-02-16
Sieve enhancements

janezd:
Sieve is ancient. It contains, for instance, its own simple implementation of context settings. So far, I:

- removed checkboxes that were never used,
- removed conditions combos haven't been used since Gregor defended his thesis,
- removed code for "optimization" that was never used, not ported to Orange 3 and commented out,
- reorganized controls to save space (the layout is non-standard now, but the control area would be almost empty),
- refactored controls to use models and context settings,
- properly implemented handling of input signal Features (other widgets should follow this model, IMHO; set the values in combos and disable them for as long as the signal is present and useful)
- let the user select areas; the widget outputs them without apply button, since this is more intuitive, expected and useful in this case

The drawing part needs some refactoring, but I leave this for another PR.

### DIFF
--- a/Orange/widgets/visualize/owsieve.py
+++ b/Orange/widgets/visualize/owsieve.py
@@ -2,6 +2,8 @@ from collections import defaultdict
 from itertools import product
 from math import sqrt, floor, ceil
 
+import numpy as np
+
 from PyQt4.QtCore import Qt, QSize
 from PyQt4.QtGui import (QGraphicsScene, QGraphicsView, QColor, QPen, QBrush,
                          QDialog, QApplication, QSizePolicy)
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owsieve.py
```diff
@@ -210,23 +150,55 @@ def showEvent(self, ev):
...
+        else:
+            self.selection = {index}
+        self.update_selection()
+
+    def update_selection(self):
+        if self.areas is None:
+            self.send("Selection", None)
+            return
+
+        selected = np.zeros(len(self.data), dtype=bool)
```
janezd: @lanzagar I couldn't use the standard filters (at least I don't know of any that can do this), so this doesn't work on SQL. It should be trivial to add it to the query, though.

lanzagar:
And I like the other changes A LOT!
Though simplifying code and the UI in legacy complexity is always great in my opinion :)

Orange/widgets/visualize/owsieve.py
```diff
@@ -1,26 +1,36 @@
 from collections import defaultdict
 from itertools import product
 from math import sqrt, floor, ceil
-import random
-import sys
 
-from PyQt4.QtCore import Qt
-from PyQt4.QtGui import (QGraphicsScene, QGraphicsView, QColor, QPen, QBrush,
-                         QDialog, QApplication)
+import numpy as np
```
lanzagar: This line is a leftover from numpy-based selections and is no longer needed.

PR #1026, opened 2016-02-19
Widgets for saving and loading distances

janezd:
The PR contains

- methods for loading and saving to `DistMatrix` from/to file
- the corresponding widgets
- some additional tests
- removal of `DistMatrix.get_KNN`

It also pulls the code for recent paths from the file widget into a separate module, so it can be reused in multiple widgets.

Widgets for file distances intentionally support just a single data format. Let's try to keep it simple this time.

### DIFF
--- a/Orange/tests/test_distances.py
+++ b/Orange/tests/test_distances.py
@@ -1,4 +1,3 @@
-from contextlib import contextmanager
 from unittest import TestCase
 import pickle
 
--- a/Orange/tests/test_distances.py
+++ b/Orange/tests/test_distances.py
@@ -1,3 +1,4 @@
+from contextlib import contextmanager
 from unittest import TestCase
 import pickle
 
@@ -9,6 +10,8 @@ from Orange.data import (Table, Domain, ContinuousVariable,
 from Orange.distance import (Euclidean, SpearmanR, SpearmanRAbsolute,
                              PearsonR, PearsonRAbsolute, Manhattan, Cosine,
                              Jaccard, _preprocess)
+from Orange.misc import DistMatrix
+from Orange.util import OrangeDeprecationWarning
 
 
 def tables_equal(tab1, tab2):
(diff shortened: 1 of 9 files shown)


Orange/tests/test_distances.py
```diff
@@ -34,6 +37,157 @@ def test_pickling(self):
...
+    def named_file(self, content, encoding=None):
+        import tempfile
+        import os
+        file = tempfile.NamedTemporaryFile(
+                "wt", delete=False, encoding=encoding)
+        file.write(content)
+        name = file.name
+        file.close()
+        yield name
+        os.remove(name)
```
ales-erjavec: this should be
```python
try:
    yield name
finally:
    os.remove(name)
```

Orange/widgets/unsupervised/owdistancefile.py
```diff
@@ -0,0 +1,158 @@
...
+                start_file = ""
+            if not start_file or not os.path.exists(start_file):
+                widgets_dir = os.path.dirname(gui.__file__)
+                orange_dir = os.path.dirname(widgets_dir)
+                start_file = os.path.join(orange_dir, "doc", "datasets")
+            if not start_file or not os.path.exists(start_file):
+                d = os.getcwd()
+                if os.path.basename(d) == "canvas":
+                    d = os.path.dirname(d)
+                start_file = os.path.join(os.path.dirname(d), "doc", "datasets")
```
ales-erjavec: All of this (after `get_sample_datasets_dir()`) is unnecessary (the same applies to the OWFile widget also).
From memory this is how Orange was organized in v2 (possibly even v1).

Orange/tests/test_distances.py
lanzagar: Should we have a contextmanager like this implemented properly and available in some utils module?
We appear to be using something similar in multiple places in Orange (search "NamedTemporaryFile")

Orange/widgets/utils/filedialogs.py
```diff
@@ -85,3 +87,269 @@ def get_file_name(start_dir, start_filter, file_formats):
...
+        """Move the n-th file to the top of the list"""
+        recent = self.recent_paths[n]
+        del self.recent_paths[n]
+        self.recent_paths.insert(0, recent)
+
+    def last_path(self):
+        """Return the most recent absolute path or `None` if there is none"""
+        abspath = self.recent_paths and self.recent_paths[0].abspath
+        if abspath != "(none)":
+            return abspath
```
ales-erjavec: if `self.recent_paths` is `[]` (the default value) this will return an `[]` not a `None`. Also the abspath of the recent item should never be a string "(none)".

Orange/widgets/data/owfile.py
```diff
@@ -21,128 +22,7 @@ def add_origin(examples, filename):
             var.attributes["origin"] = dir_name
 
 
-class RecentPath:
```
ales-erjavec: Please import RecentPath item from the new location into this module's namespace so that settings saved before the change can be loaded.

Orange/widgets/utils/filedialogs.py
```diff
@@ -85,3 +87,269 @@ def get_file_name(start_dir, start_filter, file_formats):
...
+        return None
+
+    @property
+    def value(self):
+        return os.path.basename(self.abspath)
+
+    @property
+    def icon(self):
+        provider = QFileIconProvider()
+        return provider.icon(QFileIconProvider.Drive)
```
ales-erjavec: `return provider.icon(QFileInfo(self.abspath))`?

Orange/widgets/utils/filedialogs.py
```diff
@@ -85,3 +87,269 @@ def get_file_name(start_dir, start_filter, file_formats):
...
+                if self.prefix == prefix:
+                    path = os.path.join(base, self.relpath)
+                    if os.path.exists(path):
+                        return RecentPath(
+                            os.path.normpath(path), self.prefix, self.relpath)
+        return None
+
+    @property
+    def value(self):
+        return os.path.basename(self.abspath)
```
ales-erjavec: Is `value` being used anywhere? I would prefer it it was renamed to basename.

Orange/tests/test_distances.py
janezd: I moved this context manager to `Orange.tests.named_file`. I haven't looked for all other uses; this belongs to another PR.

Orange/widgets/unsupervised/owdistancefile.py
janezd: OK, I also removed it from the File Widget. Please check whether I understood it correctly.

Orange/widgets/utils/filedialogs.py
janezd: Renamed to `basename`. It was used in two places.

janezd:
@ales-erjavec Is this OK now?

PR #1028, opened 2016-02-21
RF without pruning by default

doblak:
Brieman: "Each tree is grown to the largest extent possible. There is no pruning."

We should probably use that as default. It yields much better performance.

More info: http://stats.stackexchange.com/questions/36298/why-is-pruning-not-needed-for-random-forest-trees

### DIFF
--- a/Orange/classification/random_forest.py
+++ b/Orange/classification/random_forest.py
@@ -25,7 +25,7 @@ class RandomForestLearner(SklLearner, _FeatureScorerMixin):
     name = 'random forest'
 
     def __init__(self, n_estimators=10, max_features="auto",
-                 random_state=None, max_depth=3, max_leaf_nodes=5,
+                 random_state=None, max_depth=None, max_leaf_nodes=None,
                  preprocessors=None):
         super().__init__(preprocessors=preprocessors)
         self.params = vars()


lanzagar:
I agree with the change, but for different reasons.
The RandomForest learner in Orange uses the scikit-learn implementation and should in my opinion use the same defaults unless there are very good reasons for different ones.
The default parameters might have changed in recent versions of scikit-learn, but currently they do not have pruning by default (same as you suggest). There are also many parameters that they (now) support which are missing from Orange. So if I could suggest you add those in this pull request as well, and in the future, we should think of a better way to track the changes to default parameters.

The signature from sklearn is:
`RandomForestClassifier(n_estimators=10, criterion='gini', max_depth=None, min_samples_split=2, min_samples_leaf=1, min_weight_fraction_leaf=0.0, max_features='auto', max_leaf_nodes=None, bootstrap=True, oob_score=False, n_jobs=1, random_state=None, verbose=0, warm_start=False, class_weight=None)`
I think we can use the same in Orange, but drop the `warm_start` parameter as that needs to be handled differently in Orange.

lanzagar:
I am merging this pull request to avoid delays, and we will update other parameters (and methods) separately.

Thank you for your contribution.

PR #1033, opened 2016-02-25
canvas: Create a placeholder widget in case of an error during initia…

ales-erjavec:
…lization


janezd:
I discussed this with @BlazZupan. The mock widget is a good idea, but we don't want to show the traceback to the user. Instead, we would like to show a message that something is seriously wrong with the widget and ask the user to push a button to send us a report.

This requires a more general framework for error reporting; we have it as one of GSoC tasks. So I guess we wait with this PR for now (unless this PR is urgent for any other reason).

PR #1034, opened 2016-02-26
[FIX] Exporting graphs to dot format

kernc:


Orange/widgets/io.py
```diff
@@ -27,27 +27,32 @@ def _get_exporter():
...
             painter.begin(buffer)
             painter.setRenderHint(QtGui.QPainter.Antialiasing)
 
-            target = cls._get_target(scene, painter, buffer, source)
-            scene.render(painter, target, source)
+            target = cls._get_target(scene, painter, buffer, rect)
+            try:
+                scene.render(painter, target, rect)
+            except TypeError:
+                scene.render(painter)  # PyQt4 QWidget.render() takes different params
```
VesnaT: Maybe you could replace try/except block with if statement.

kernc: I don't find [it](https://github.com/biolab/orange3/pull/1034/files) much better now. Other blocks were left for generality.

kernc:
@lanzagar Please also consider merging this. :)

PR #1037, opened 2016-03-01
Scikit RF params sync

doblak:
#1028 was merged as a partial fix, opening a pull request with proper method signature (as in sci-kit).


lanzagar:
Could you make this with a single commit (the last one, without the merge commits) and rebase it on latest master.

doblak:
Here you go.

PR #1038, opened 2016-03-01
added UCI spambase dataset

doblak:
added UCI spambase dataset (https://archive.ics.uci.edu/ml/datasets/Spambase)


lanzagar:
I suggest we enhance all pull requests that want to add data sets included in Orange with at least the following information (in the PR description):
- size of added file(s)
- a sentence or two about why we want this data in our default installation and repositories
- included in the above or separate: distinguishing features of this data - does it have multiple targets with missing values that no other data has. Or is it like 10 other data sets (which ones) - it can still be included based on relevance of the data.

I am not saying this data set should not be included, but it would be good if the reasons are written somewhere for posterity, because there are thousands of data sets out there.

If there will be interest in (too) many data sets, we should consider an add-on for them, or on-demand downloads or something like that. (@astaric, @janezd, @kernc)

doblak:
There is no special reason for this data set being added, except that it quite often appears in papers.

In general, it is cumbersome for the user to transpile such datasets from raw format to Orange readable format, therefore it makes sense to only do it once and with caution to avoid errors.

However you points are valid as well. I already opened #1040 to address the ambiguity of data sets included in Orange. In addition there may be permission/license issues with including data sets like that (I wrote a mail to UCI repository to clarify if such inclusion is allowed for their data sets with no special licensing/citation requests, but received no response yet).

Considering all this, your suggestion for an add-on seems to be a good solution to the problem. I vote for closing the PR, but maybe we should wait for others to express their opinions.

kernc:
We agree with @lanzagar 100%. As discussed, we should probably have a dataset loading widget (with credits, meta info, filtering & all) such as [GEO Data Sets](https://github.com/biolab/orange-bio/blob/master/orangecontrib/bio/widgets3/OWGEODatasets.py) from Bioinformatics add-on.

doblak:
Agreed, closing.

PR #1039, opened 2016-03-01
added UCI sonar dataset

doblak:
added UCI sonar dataset 

https://archive.ics.uci.edu/ml/datasets/Connectionist+Bench+(Sonar,+Mines+vs.+Rocks)


kernc:
Why are learn and test separate?

doblak:
The original data is divided into two groups, train (*CM, *.CR) and test group (CM, CR). There is also a joined dataset.

https://archive.ics.uci.edu/ml/machine-learning-databases/undocumented/connectionist-bench/sonar/

Maybe we should keep thins simple and omit the train/test datasets and only include the joined one?

lanzagar:
I suggest we add just the joined data set. Sometimes it is beneficial to have the original split, to replicate results and make fair comparisons. But this data appeared in a paper from 1988, so I do not think it is that relevant anymore.
If you think using the original split is still important for many users, I suggest we have the joined data set with a meta variable containing that information (e.g. 0=train, 1=test).

doblak:
Fixed, only the joined set remains.

But the same question arrises in this PR as in #1038. We should probably finish the discussion there first.

doblak:
Closed for the same reasons as #1038

Issue #1051, opened 2016-03-03
Dragging of the threshold (# of dimensions) in the PCA Widget


janezd:
PCA Widget allows the user to manually drag the threshold in the graph. The threshold (the number of dimensions) is an integer, so it has to snap to the grid. Currently the number is rounded down, but the line can be at an arbitrary point.

RachitKansal:
@janezd  I would like to take up this task.

RachitKansal:
I have recreated the bug on my local machine, will send a PR soon.

RachitKansal:
I have just made a PR https://github.com/biolab/orange3/pull/1053

kernc:
Should the graph also color (paint) the area between? The threshold line is absolutely not visible enough.

RachitKansal:
I think coloring can make the graph clumsy, I was thinking thickening of the line is a better option. @janezd please take a look at this screen shot, If this is better.
![snap](https://cloud.githubusercontent.com/assets/9735009/13503318/d495db02-e194-11e5-97db-646db06976fa.png)

janezd:
It is better, in my opinion. It's something one must play with, though, to see what looks good. If I was trying to aesthetically improve the widget, I would make a dashed line instead of a solid, I would make the graph curves (the red and the greenish line) a bit thicker on the part to the left of the threshold and a bit thinner on the right. It's not really a part of this issue - the issue is fixed, I guess (this was cheap :)). But if you want you can also try to make it nicer.

@BlazZupan, how much of the control area do we actually need? I'm much in favour of getting rid of as much as we can. (See how the new Sieve looks like.) Does one ever need more than 100 components? What about changing the first box into info, and putting it below the graph. So the only way to change the number of components is to drag the line. The second box would be eliminated altogether: the data would always be normalized (does non-normal PCA make sense?) and the graph would always show up to 100 components. The buttons go at the bottom, to the right of the info. Hence, we would keep only the graph and a single-line box with info and buttons below it.

lanzagar:
@janezd @BlazZupan 

> Does one ever need more than 100 components?

Yes, often you want all components (e.g. for comparison). If we want the graph to cover the whole range it might be better to have it in logarithmic scale (x-axis). I think it would also be acceptable to have it limited to 100 components if there is a separate option to use all components instead.

> The second box would be eliminated altogether: the data would always be normalized (does non-normal PCA make sense?) and the graph would always show up to 100 components.

[...]

PR #1053, opened 2016-03-03
Fixed  Dragging of the threshold (# of dimensions) in the PCA Widget #1051 

RachitKansal:
The movement of threshold is now restricted to only snapping the grid.

### DIFF
--- a/Orange/widgets/unsupervised/owpca.py
+++ b/Orange/widgets/unsupervised/owpca.py
@@ -238,7 +238,7 @@ class OWPCA(widget.OWWidget):
             bounds=(0, p - 1)
         )
         self._line.setCursor(Qt.SizeHorCursor)
-        self._line.setPen(pg.mkPen(QColor(Qt.darkGray), width=1.5))
+        self._line.setPen(pg.mkPen(QColor(Qt.darkGray), width=5))
         self._line.sigPositionChanged.connect(self._on_cut_changed)
 
         self.plot.addItem(self._line)
@@ -248,6 +248,7 @@ class OWPCA(widget.OWWidget):
     def _on_cut_changed(self, line):
         # cut changed by means of a cut line over the scree plot.
         value = line.value()
+        self._line.setValue(round(value))
         current = self._nselected_components()
         components = int(numpy.floor(value)) + 1
 


janezd:
The fix for the threshold is OK, AFAIK.

Discretization is not OK: the widget needs to show the discretized attributes. The only effect of the code is that it discretizes what the widget's output. The output needs to stay continuous.

RachitKansal:
@janezd  please have a look at the mosaic, I'm still working on Sieve.

RachitKansal:
ok, yeah that is much better. Will do it.

RachitKansal:
@janezd  I have worked on the issue, but I believe it requires thorough checking. Now the combo boxes allow Continuous attributes and give a warning to the user about default discretization which is being carried out.
It wasn't earlier making the mosaic because the function DrawData was actually using the self.data, which was saved on setData and hence was considering those continuous variables. Hence I had to change the function DrawData and DrawText consequently to work on the discretized dataset.
I have tested it on the iris dataset.

janezd:
It looks much better now. I haven't really tested it yet, I just tried it on Iris - yes, it works as intended. You've managed the first step.

The reason I haven't really explored it is because you're not out of the woods yet, though, and you'll probably add some further changes to your code. Attach a Table widget to the scatter plot and select a box by clicking on it. If the attributes are discrete, the widget outputs a subset. If they are continuous, nothing happens. You now need to find a way to relate the data belonging to the box back to the original data instances.

RachitKansal:
I'm sorry I couldn't understand the task regarding scatter plot..can you please clarify it further.

janezd:
I gave scatter plot just as an example. The point is that the File widget should output just a single `Table`, containing the data from one of the Excel's sheets. With this, all other widgets (like, for instance Scatter plot), would work without changes.

janezd:
@RachitKansal, I'm really sorry, I didn't pay attention and thought we're in a different "thread". As for the scatter plot, I think it's being solved in #1070.

To avoid chaos in the future: let one PR solve one issue and let us discuss issues (e.g. *can you give me a better explanation about this issue*) within issues, not to PRs.

janezd:
I can't merge this PR since it fixes multiple issues. It is also in conflict with the current master since some of these fixes were already merged via your other PR.

Please clean up this PR to include only the commits related to PCA. Also, squash them into single commit. Then I can merge it.

janezd:
@RachitKansal, can you clean this up? Remove the commits that do not belong to this PR (keep only the PCA-related stuff) and then rebase it to master.

RachitKansal:
@janezd I have removed the commits not related to PCA. Please have a look now.

Issue #1045, opened 2016-03-03
Multiple sheets in Excel Files


janezd:
Orange does not support multiple sheets in Excel files. Add this functionality to the file widget. Don't ruin the GUI layout.

rahulroxx:
I did a quick fix for selecting multiple xls in the file widget . Should i update the comboBox too ? 

`import os
from warnings import catch_warnings

from PyQt4 import QtGui, QtCore
from PyQt4.QtGui import QSizePolicy as Policy

from Orange.widgets import widget, gui
from Orange.widgets.settings import Setting
from Orange.widgets.utils.itemmodels import PyListModel
from Orange.data.table import Table, get_sample_datasets_dir
from Orange.data.io import FileFormat


def add_origin(examples, filename):
    """Adds attribute with file location to each variable"""
    vars = examples.domain.variables + examples.domain.metas
    strings = [var for var in vars if var.is_string]
    dir_name, basename = os.path.split(filename)
    for var in strings:
        if "type" in var.attributes and "origin" not in var.attributes:
            var.attributes["origin"] = dir_name


class RecentPath:
    abspath = ''
    prefix = None   #: Option[str]  # BASEDIR | SAMPLE-DATASETS | ...
    relpath = ''  #: Option[str]  # path relative to `prefix`
    title = ''    #: Option[str]  # title of filename (e.g. from URL)

    def __init__(self, abspath, prefix, relpath, title=''):
        if os.name == "nt":
            # always use a cross-platform pathname component separator
            abspath = abspath.replace(os.path.sep, "/")
            if relpath is not None:
                relpath = relpath.replace(os.path.sep, "/")
        self.abspath = abspath
        self.prefix = prefix
        self.relpath = relpath
        self.title = title

    def __eq__(self, other):
        return (self.abspath == other.abspath or
                (self.prefix is not None and self.relpath is not None and
                 self.prefix == other.prefix and
                 self.relpath == other.relpath))

    @staticmethod
    def create(path, searchpaths):
        """
        Create a RecentPath item inferring a suitable prefix name and relpath.
        Parameters
        ----------
        path : str
            File system path.
        searchpaths : List[Tuple[str, str]]
            A sequence of (NAME, prefix) pairs. The sequence is searched
            for a item such that prefix/relpath == abspath. The NAME is
            recorded in the `prefix` and relpath in `relpath`.
            (note: the first matching prefixed path is chosen).
        """
        def isprefixed(prefix, path):
            """
[...]

PR #1058, opened 2016-03-04
#1045 Issue Multiple Excel file in file widget 

rahulroxx:
@janezd i was able to select multiple files with file widget with the auto save and remember file name . But i am facing  problem to create a new tab within the table widget > 
The problem is I am sending the output signals from filewidget(multiple files paths) to tablewidget that contains  but input of table widget is not handling that . I saw the testing of the widget it is succesfully creating multiple tabs with multiple Table() instances and save_dataset() Handling the est to create tabs. So How can i Use that in this case ?


rahulroxx:
@janezd  only check the latest pr i had some name conflicts in previous pr's

PR #1059, opened 2016-03-04
[FEATURE] Silhouette Plot widget

ales-erjavec:
Moved here from biolab/orange3-prototypes


lanzagar:
The widget crashes on SqlTable inputs.
Since I don't think we want to make any special cases for SQL data, you could just add the decorator that checks the SQL data and downloads/samples based on size.

```
@check_sql_input
def set_data(self, data):
```

ales-erjavec:
>  you could just add the decorator that checks the SQL data and downloads/samples based on size.

Done.

PR #1060, opened 2016-03-05
Allow user to select meta attributes

### LINKED ISSUE #1049: Meta attributes in Sieve and Mosaic
Sieve and Mosaic do not let the user choose meta attributes. Add them to the combo.

kwikadi:
In sieve and mosaic widgets. Fixes #1049

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -2,7 +2,7 @@ import os
 import sys
 from collections import defaultdict
 from functools import reduce
-from itertools import product
+from itertools import product, chain
 from math import sqrt
 
 import numpy
@@ -294,7 +294,7 @@ class OWMosaicDisplay(OWWidget):
         self.attr3Combo.addItem("(None)")
         self.attr4Combo.addItem("(None)")
 
-        for attr in data.domain:
+        for attr in chain(data.domain.attributes, data.domain.metas):
             if attr.is_discrete:
                 for combo in [self.attr1Combo, self.attr2Combo, self.attr3Combo, self.attr4Combo]:
                     combo.addItem(self.icons[attr], attr.name)
--- a/Orange/widgets/visualize/owsieve.py
+++ b/Orange/widgets/visualize/owsieve.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from itertools import product
+from itertools import product, chain
 from math import sqrt, floor, ceil
 
 from PyQt4.QtCore import Qt, QSize
@@ -97,7 +97,11 @@ class OWSieveDiagram(OWWidget):
         if self.data is None:
             self.attrs[:] = []
         else:
-            self.attrs[:] = [var for var in self.data.domain if var.is_discrete]
+            self.attrs[:] = [
+                var for var in chain(self.data.domain.attributes,
+                                     self.data.domain.metas)
+                if var.is_discrete
+            ]
         if self.attrs:
             self.attrX = self.attrs[0].name
             self.attrY = self.attrs[len(self.attrs) > 1].name


kernc:
It looks like this solution would work, but there is quite a bit redundancy there. Since you are iterating over domain variables (`domain`) and domain meta variables (`domain.metas`), you could just chain them before iterating over both?

kwikadi:
Unfortunately my laptop's charger stopped working. I'll send in a fix in a
few hours, though.

kwikadi:
I've removed the redundancy.

kernc:
You could use [`itertools.chain`](https://docs.python.org/3/library/itertools.html#itertools.chain) to "join" the two [iterables](http://stackoverflow.com/questions/9884132/what-exactly-are-pythons-iterator-iterable-and-iteration-protocols) instead of concatenating (`+`), which generates an intermediate list that is thrown away right afterwards.

You can avoid the same kind of iteration redundancy in the other file.

kwikadi:
Fixed, I think.

kwikadi:
Oh, wait. Do you want me to assign it to another variable before iterating over it?

kernc:
Looks fine to me.

Orange/widgets/visualize/owsieve.py
```diff
@@ -97,7 +97,7 @@ def set_data(self, data):
         if self.data is None:
             self.attrs[:] = []
         else:
-            self.attrs[:] = [var for var in self.data.domain if var.is_discrete]
+            self.attrs[:] = [var for var in chain(self.data.domain.attributes, self.data.domain.metas) if var.is_discrete]
```
kernc: Could you break this in two or three lines? Seem a bit long as-is.

kwikadi: Sure

kwikadi:
Done

kernc:
Looks good, but could you avoid the last "Fix" commit by merging it into the first ("Allow ... in Sieve") commit? You can do so with:
```bash
git rebase --interactive HEAD^^^
```
and then mark the last commit as _fixup_ on the first one (will require rearranging the commit lines a bit). See https://www.google.com/search?q=github+squash+pull+request. This makes for a much nicer commit history.

kwikadi:
Yeah, no problem. And I knew about interactive rebase (although this is recent) :)

kwikadi:
Um, done.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -294,7 +294,12 @@ def initCombos(self, data):
         self.attr3Combo.addItem("(None)")
         self.attr4Combo.addItem("(None)")
 
-        for attr in data.domain:
+        for attr in chain(data.domain.attributes, data.domain.metas):
+            if attr.is_discrete:
+                for combo in [self.attr1Combo, self.attr2Combo, self.attr3Combo, self.attr4Combo]:
+                    combo.addItem(self.icons[attr], attr.name)
+
+        for attr in data.domain.metas:
```
kernc: Are you sure this is ok? Won't `domain.metas` be included twice now? Something must have changed since the rebase.

kwikadi:
Due to commit reordering and a merge conflict that I overlooked, removing the erroneous code was creating problems (thanks to the 'Fix Indentation' commit, removing the code from the 'Mosaic' commit would have made it empty), so I squashed all commits into a single commit.

kernc:
Looks good. Thanks for your contribution!

Orange/widgets/visualize/owmosaic.py
```diff
@@ -294,7 +294,7 @@ def initCombos(self, data):
         self.attr3Combo.addItem("(None)")
         self.attr4Combo.addItem("(None)")
 
-        for attr in data.domain:
+        for attr in chain(data.domain.attributes, data.domain.metas):
```
lanzagar: Wasn't the pull request supposed to just add meta variables?
Why have we switched from `data.domain` to `data.domain.attributes` which does not have the class variables?
Probably not intended... @kernc @janezd ?

Orange/widgets/visualize/owsieve.py
```diff
@@ -97,7 +97,11 @@ def set_data(self, data):
         if self.data is None:
             self.attrs[:] = []
         else:
-            self.attrs[:] = [var for var in self.data.domain if var.is_discrete]
+            self.attrs[:] = [
+                var for var in chain(self.data.domain.attributes,
```
lanzagar:
```
chain(self.data.domain,
      self.data.domain.metas)
```
?

Issue #1044, opened 2016-03-03
Classification tree outputs learner when given new data


janezd:
Construct a workflow in which you have a File widget and Classification tree widget, both connected to Test learners (note that File is *not* connected to Classification tree). Test Learners runs cross validation and shows the results.

Now connect File to Tree. The tree widget should output only the learner, but it also outputs the tree learner again (although it does not change when the data is changed), causing the Test Learners to unnecessarily recompute the cross validation.

s-alexey:
I've suggested my solution. 

But it isn't only `OWClassificationTree` problem. I think it would be better to add some logic into `OWProvidesLearner` class. 

something like:

```python
    def apply(self):
        self.update_learner()
        self.update_model()
        
    def update_learner(self):
        raise NotImplementedError()
    
    def update_model(self):
        raise NotImplementedError()
```

kernc:
Code consolidation and refactoring are an integral part of a healthy development process. As you say, the more one can solve problems cleanly, the better.

s-alexey:
Ok, I can perform `OWProvidersLearner` improvement. Here is a drawft. I would like to know you suggestions.
```python 
class OWProvidesLearner(_OWAcceptsPreprocessor):
    """
    Base class for all classification / regression learner-providing widgets
    that extend it.
    """

    LEARNER = None
    OUTPUT_MODEL_NAME = None
    OUTPUT_MODEL_CLASS = None

    inputs = [("Data", Table, "set_data")] + _OWAcceptsPreprocessor.inputs

    @property
    def outputs(self):
        return [
            ("Learner", self.LEARNER),
            (self.OUTPUT_MODEL_NAME, self.OUTPUT_MODEL_CLASS)
        ]

    model_name = Setting(OUTPUT_MODEL_NAME)

    def __init__(self):
        super().__init__()

        self.data = None
        self.learner = None
        self.model = None
        # here can be gui configuration (e.g. gui.lineEdit for model_name)
        # we can add method get_setting_box() 
        # 
        # gui.lineEdit
        # box = get_setting_box()
        # apply and report buttons 

    @check_sql_input
    def set_data(self, data):
        """Set the input train data set."""
        self.error(0)
        self.data = data
        if data is not None and data.domain.class_var is None:
            self.error(0, "Data has no target variable")
            self.data = None
        self.update_model()

    def apply(self):
        self.update_learner()
        self.update_model()

    def update_learner(self):
        # ...
        # self.learner.name = self.model_name
[...]

PR #1063, opened 2016-03-05
OWClassificationTree: Fix set_data behaviour.

s-alexey:
This commit solves #1044 but every other learner still has the same problem.

### DIFF
--- a/Orange/widgets/classify/owclassificationtree.py
+++ b/Orange/widgets/classify/owclassificationtree.py
@@ -94,8 +94,11 @@ class OWClassificationTree(OWProvidesLearner, widget.OWWidget):
             preprocessors=self.preprocessors
         )
         self.learner.name = self.model_name
-        self.model = None
+        self.send("Learner", self.learner)
+        self.update_model()
 
+    def update_model(self):
+        self.model = None
         if self.data is not None:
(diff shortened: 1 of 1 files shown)


kernc:
WRT lines of code, I have no objections here, except what's stated in https://github.com/biolab/orange3/issues/1044#issuecomment-192754237. Thanks for setting a reasonable commit message title! The commit message's description is directly the opposite, however: setting new data _should_ update and send the new model; the _learner_ (unfitted model) is what doesn't change until its parameters do.

lanzagar:
A problem appears when the user changes settings, but does not yet push Apply. The Learner is not yet updated.
Now if the input data changes, it does not update the learner anymore, so the existing learner with old settings is used, which produces very confusing (wrong) results.

This happens because there is no "auto apply" option and no callbacks on individual settings controls.

@kernc can you coordinate this, possibly in connection with #1072.

s-alexey:
> A problem appears when the user changes settings, but does not yet push Apply. 

@lanzagar, I think it's better to only notify the user when he updates settings but doesn't press apply button. It means we should add a callback to every setting. I will add it to #1072.

astaric:
Fixed in #1072

Issue #1047, opened 2016-03-03
Select Columns does not trigger auto apply


janezd:
Place Select Columns within some schema and check the auto apply checkbox. Change the order of attributes by dragging within the list box. The widget does not emit the new data table although the auto apply is enabled.

rahulroxx:
@ales-erjavec 	Fixed Issue #1047

rahulroxx:
@ales-erjavec Fixed issue with PR #1064

ajdapretnar:
Fixed via #1353.

PR #1064, opened 2016-03-06
[WIP] Fixed auto apply in Select Columns (Issue #1047)

rahulroxx:
@ales-erjavec This fixes Issue #1047 .

### DIFF
--- a/Orange/widgets/data/owselectcolumns.py
+++ b/Orange/widgets/data/owselectcolumns.py
@@ -361,6 +361,12 @@ class OWSelectAttributes(widget.OWWidget):
                           Orange.data.ContinuousVariable))
 
         self.used_attrs_view.setModel(self.used_attrs)
+
+        ua = self.used_attrs
+        ua.dataChanged.connect(self.update_completer_model)
+        ua.rowsInserted.connect(self.update_completer_model)
+        ua.rowsRemoved.connect(self.update_completer_model)
+
         self.used_attrs_view.selectionModel().selectionChanged.connect(
             partial(self.update_interface_state, self.used_attrs_view))
         box.layout().addWidget(self.used_attrs_view)


Orange/widgets/data/owselectcolumns.py
```diff
@@ -361,6 +361,12 @@ def __init__(self):
                           Orange.data.ContinuousVariable))
 
         self.used_attrs_view.setModel(self.used_attrs)
+
+        ua = self.used_attrs
+        ua.dataChanged.connect(self.update_completer_model)
+        ua.rowsInserted.connect(self.update_completer_model)
+        ua.rowsRemoved.connect(self.update_completer_model)
```
ales-erjavec: The update_completer_model is probably not what you want to call here. It only updates the completion model for the filter line edit in the left "Available Variables" box. It does not trigger an update of the widget's output.
Note also that changes in the `class_attrs` and `meta_attrs` should also trigger an update. Note also that a drag/drop between the used_attrs/class/meta_attrs_views will trigger possible multiple rowsRemoved signals from drag source model followed by a rowsInserted signal in the destination model. Ideally the widget output should only be updated/computed once, after the drag/drop operation has been completed.

rahulroxx:
@ales-erjavec I think `update_completer_model` has `self.commit()` which auto updates and send output signals to update the other widgets input .I tested it and works fine ( Auto apply triggered which updates the   Data Table widget ) . I didn't updated meta and class attrs that's what i need to do now . 
Please correct me if i am wrong !  Thanks for helping me ! :)

rahulroxx:
@ales-erjavec What would you suggest  ?

ales-erjavec:
> I think update_completer_model has self.commit() which auto updates and send output signals to update the other widgets input.

You are right, I missed that. However it should not be there.
I think the best solution would be to introduce a new method to invalidate the output. Call this method on any change in the attrs/class/metas models and call commit if here is no drag/drop operation in progress. Then augment the `VariableListItemView` with two new signals that are emitted in startDrag, one immediately before the call to `drag.exec_` and the other the the end (after the view's model is updated). Use these signals to track drag/drop operation state, and delay the call to commit.

rahulroxx:
@ales-erjavec Yes !

rahulroxx:
@ales-erjavec 

> Then augment the VariableListItemView with two new signals that are emitted in startDrag, one immediately before the call to drag.exec_ and the other the the end (after the view's model is updated). Use these signals to track drag/drop operation state, and delay the call to commit.

I should check for default_action and move_action right ?

ales-erjavec:
Sorry for a late response. 

> `default_action` and `move_action` signals to check before commit?

I am not sure I understand the question exactly.
But yes, which action [(Qt.MoveAction, Qt.InternalMove, Qt.IgnoreAction, ...)](http://doc.qt.io/qt-4.8/qt.html#DropAction-enum) took place is important and needs to be checked.

rahulroxx:
@ales-erjavec I meant two signals to check drag and drop operations .

rahulroxx:
@ales-erjavec Need assistance on invalidating the output .  rowsInsert, rowsRemoved will surely be depreceated. as changes in attrs/metas/class should be computed once. Exac

ajdapretnar:
I am closing this PR, since we're preparing a wider refactoring and reworking of several widgets. Thank you @rahulroxx for your help, we appreciate it nevertheless.

Issue #1065, opened 2016-03-06
Table not loading breast-cancer.tab


wadhwasahil:
When I try to load the file breast-cancer.tab in a data table format, the column under tumor size is showing missing values whereas they exist in the original file.
Why is it happening as all the other files tested by me are working/opening correctly?

s-alexey:
It happens because of the `tumor size` column type. Just change the second line in the file 
from
d	d	d	**0-4 5-9**	0-2 3-5 6-8 9-11 12-14 15-17	d	d	d	d	d
to 
d	d	d	**d**	0-2 3-5 6-8 9-11 12-14 15-17	d	d	d	d	d

*note:* do not remove tab spaces around 0-4 5-9

wadhwasahil:
oh. Thanks (y) .
But how does system knows which intervals are for which column?
I mean, **0-4 5-9** are for tumor-size and **0-2 3-5 ..... 15-17**  are for inv-nodes. They are only seperated by spaces while **d** are seperated by a tab. So, how does system knows which intervals are for which column?

s-alexey:
Look carefully, there is a tab between 5-9 and 0-2.

wadhwasahil:
Okay. got it!

lanzagar:
Nice catch!
However, I would not close this issue yet - the fix has not been applied in the Orange repository yet.
Maybe one of you wants to create a pull request with it?

lanzagar:
Thank you both for noticing the issue and fixing it.

wadhwasahil:
I have also generated PR for this. LOL

PR #1067, opened 2016-03-07
fixed issue #1065

shark-S:
as comments in issue #1065 it has been updated


lanzagar:
You are a bit too late - this has been merged in another pull request (#1066) half an hour ago.
But to learn something from this pull request - there are 5 commits in yours with only the first one needed. Before this could be merged, you would need to remove the last 4 - it is not only about the final diff, commits matter as well!

shark-S:
thanks for notification  . I have made a mistake so that's why it create 4 commits next time i will take care of it.

Issue #1048, opened 2016-03-03
Scatter plot should hide attributes with no defined values


janezd:
If an attribute does not have any defined values (all nan) and the user selects it, the scatter plot should show a warning.

themistoklik:
Hello @janezd ,

A warning can be displayed by adding a data.has_missing() check to owscatterplot in the set_data function.

Such a warning would be displayed not only when all values of an attribute are NaN, but also if any attribute value is NaN (which I think is useful information).

What would be more useful would be to pinpoint the feature names where the NaNs exist so the user is informed. How do i go about getting each attribute's values to run a has_missing() there and warn about NaNs specifically in that attribute?

I do understand that for large datasets such a warning could prove dysfunctional, so if you disagree , please answer me for future reference.

ajdapretnar:
The warning can be: Columns with missing values have been removed.

I still believe Scatter Plot should not offer empty attributes (NaN) in the 'Axis data' dropdowns as there is absolutely no point in having them there.

janezd:
I agree with both: attributes with no defined values should be removed since they are not useful. If any such attributes were removed, we need a warning. Second, as @themistoklik suggested, we can warn if any data points were removed due to missing values -- that is, if an attribute has some missing values, but at least one value is defined.

I don't thing that marking the attributes with missing values in advance (say in the combo) would be practical. If the data set has missing values, a large number of attributes would have them and marking those wouldn't be so useful.

For large data, read directly from database, the widget would stay as it is. We can't check whether all data is missing, and for plotting we use sampling anyway.

wadhwasahil:
@janezd . If I have the attribute name "age", so how can I print all the values of "age" in the dataset?
Also, how to check if there is a missing value? Do you have any pre defined function for it?

wadhwasahil:
@janezd Please can you see my PR and suggest changes.
1. How to remove warnings from the console if I had posted it. I tried graph.update().
2. I have considered only those attributes with Nan or "?" values and generated warnings.

Waiting for your reply.
Thanls

kernc:
> If I have the attribute name "age", so how can I print all the values of "age" in the dataset?

The following is standard in numpy, pandas, and works with Orange as well:
```python
print(data[:, 'age'])   # all rows, 'age' column
```

wadhwasahil:
Please see my PR and suggest  changes @kernc

wadhwasahil:
@kernc Have a look at the new PR :)

[...]

PR #1070, opened 2016-03-07
Scatter Plot Warnings for NaN attributes

wadhwasahil:
Some work towards #1048


Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,10 +380,24 @@ def update_graph(self, attributes=None, reset_view=True, **_):
         self.graph.zoomStack = []
         if attributes and len(attributes) == 2:
             self.attr_x, self.attr_y = attributes
+        # print (self.attr_x)
+        # print (self.attr_y)
+        # print (self.data[0])
+        idx = self.data.domain.index(self.attr_x)
+        idy = self.data.domain.index(self.attr_y)
+        flag = 0
+        for x in self.data[:,idx]:
```
kernc: Could use `self.data[:, self.attr_x]` directly here.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,10 +380,24 @@ def update_graph(self, attributes=None, reset_view=True, **_):
...
+            if x == '?' or not x:
+                flag = 1
+        for y in self.data[:,idy]:
+            if y == '?' or not y:
+                flag = 1
+        if flag == 1:
+            self.warning("Some of the attributes contain missing values")
```
kernc: Could use a check like:
```py
np.isnan(self.data[:, idy]).any()
```
This only works with regular `Table` and not `SqlTable`, though. @astaric may propose a better alternative.

astaric: BasicStats can be used to compute number of unknowns in a backend independent way.

https://github.com/biolab/orange3/blob/55283da471c68efee41727698903c0efbe4536b3/Orange/statistics/basic_stats.py#L18

Orange/widgets/visualize/owscatterplot.py
wadhwasahil: where to use the above code which you mentioned above? I mean which line to be removed

Orange/widgets/visualize/owscatterplot.py
wadhwasahil: finally which method should I use for checking of NaN?

wadhwasahil:
@astaric what is `dat and variable` in the `__init__` function?

wadhwasahil:
`X = BasicStats(self.data[:, self.attr_x], self.attr_x)
        Y = BasicStats(self.data[:, self.attr_y], self.attr_y)        
        if X or Y:
            self.warning("Some of the attributes contain missing values")  `
Is it fine?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,10 +380,16 @@ def update_graph(self, attributes=None, reset_view=True, **_):
         self.graph.zoomStack = []
         if attributes and len(attributes) == 2:
             self.attr_x, self.attr_y = attributes
+        X = BasicStats(self.data[:, self.attr_x], self.attr_x)
+        Y = BasicStats(self.data[:, self.attr_y], self.attr_y)
+        if X.nans or Y.nans:
+            self.warning("Some of the attributes contain missing values") 
+        else:
+             self.warning()
```
kernc: One space too many here.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,10 +380,16 @@ def update_graph(self, attributes=None, reset_view=True, **_):
         self.graph.zoomStack = []
         if attributes and len(attributes) == 2:
             self.attr_x, self.attr_y = attributes
+        X = BasicStats(self.data[:, self.attr_x], self.attr_x)
+        Y = BasicStats(self.data[:, self.attr_y], self.attr_y)
+        if X.nans or Y.nans:
+            self.warning("Some of the attributes contain missing values")
```
kernc: "Some attributes ..."

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,10 +380,16 @@ def update_graph(self, attributes=None, reset_view=True, **_):
...
+        Y = BasicStats(self.data[:, self.attr_y], self.attr_y)
+        if X.nans or Y.nans:
+            self.warning("Some of the attributes contain missing values") 
+        else:
+             self.warning()      
         if not self.graph.have_data:
             return
         self.graph.update_data(self.attr_x, self.attr_y, reset_view)
-
+
```
kernc: Don't need trailing spaces, especially on empty lines. :smiley: 
Besides configuring your editor, most of which support this setting, you can also add chunks of changes to the commit with [`git add --patch`](https://www.google.com/search?q=git+add+patch).

wadhwasahil: The  spaces  are present as they were present before. I haven't changed any of that.
So, can you merge the PR or some more changes do yo expect?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,10 +380,16 @@ def update_graph(self, attributes=None, reset_view=True, **_):
         self.graph.zoomStack = []
         if attributes and len(attributes) == 2:
             self.attr_x, self.attr_y = attributes
+        X = BasicStats(self.data[:, self.attr_x], self.attr_x)
+        Y = BasicStats(self.data[:, self.attr_y], self.attr_y)
```
lanzagar: A call to BasicStats should be fast, but I am wondering if we want this recomputed on every update_graph call.
Would it make more sense to compute BasicStats for all variables only once, when new data is set? @astaric @kernc

Orange/widgets/visualize/owscatterplot.py
lanzagar: Actually, here we know exactly which attributes are problematic, so explicit names are probably better than "Some attributes"
Or if att. names are not of interest (the user sees which ones are selected anyway), maybe something along the lines of:
"Some data instances are not shown due to missing values."

Orange/widgets/visualize/owscatterplot.py
wadhwasahil: But it makes sense to see that the variables contains NaN for each graph update. However, I tried it and it isn't slow at all. The complexity remains `O(m)` where m = no of tuples in the table.

lanzagar: I was not suggesting to move this from `update_graph`, just that the BasicStats can be computed once (e.g. in `set_data`) and the results stored for subsequent use.
But now that I think about it - maybe the warnings should be moved somewhere else as well - they depend only on the selection of attributes right? So warnings should be raised when attribute selection changes. Not when somebody selects/deselects "Show class density".

The complexity is of course O(m) for a single call, the question is how many calls are made.

wadhwasahil: Thanks for your reply.
Where should I add warnings in the code?

wadhwasahil:
#1077 also done a little bit.

Orange/widgets/visualize/owscatterplot.py
lanzagar: I have not really looked in depth at this problem, so I can't say off the top of my head. But I think it is good exercise for you to try and work out a solution yourself.

lanzagar:
Please keep the contents of one pull request limited to a single topic.

I suggest the master branch in your fork just tracks the upstream master, and you add all your commits to separate branches (each problem its own branch and pull request).

wadhwasahil:
@lanzagar Will do it.

wadhwasahil:
Do I have to close this PR to be able to send the other two PR's?

Orange/widgets/visualize/owscatterplot.py
kernc: The line we're commenting on has been changed (four space characters have been added). If it weren't, it would not be [marked green and the one above it red](https://github.com/biolab/orange3/pull/1070/files?diff=unified#diff-0f1cb6d8751836f98c5293bec4004169L385)?

Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -139,7 +139,6 @@ def setData(self, data, key=None):
             # Store hints only on the first setData call.
             self._storeHints()
             self._inputUpdate = True
-
```
kernc: I think here is [another whitespace change](https://github.com/biolab/orange3/pull/1070/files?diff=unified#r55426507) that has nothing to do with this PR.

Orange/widgets/visualize/owscatterplot.py
kernc: @wadhwasahil The `warning` call is fine where it is; it shows the warning depending on the attribute selection. But the _`BasicStats` call_ is the one that better be moved somewhere where it is computed only once per data, like @lanzagar said.

kernc:
@wadhwasahil One pull request tracks one branch, so you should open one PR for each git branch you introduce some change/fix in.

kernc:
Ah, I see you've already mostly figured it out. Closing in favor of #1096 and #1097.

PR #1071, opened 2016-03-08
Fix for Mosaic and Sieve attributes

kwikadi:
Fixes bug introduced in the previous PR, pointed out by @lanzagar.  It was introduced when I was trying to iterate through both `data.domain` and `data.domain.metas` using `+` and a type mismatch occurred. `chain` does not have any such issues.


lanzagar:
OK, I see now why the class variables got lost in the previous PR (#1060).
For future reference:
`data.domain.{variables,attributes,class_vars,metas}` are tuples
`data.domain` is an iterator over `data.domain.variables` (which are attributes + class_vars)

So if you wanted to use + the correct choice would be `data.domain.variables + data.domain.metas`.
But chain is better anyway.

PR #1072, opened 2016-03-08
OWProvidesLearner: add base methods. Refs #1044.

s-alexey:
Add some methods to OWProvidesLearner.
optional:
Create OWProvidesLearnerI and ProviderMetaClass to extend OWProvidesLearner functinality.
example:
Refactor OWClassificationTree widget wrt new base class functionality.


Orange/widgets/classify/owclassificationtree.py
```diff
@@ -1,24 +1,21 @@
...
 
-class OWClassificationTree(OWProvidesLearner, widget.OWWidget):
+class OWClassificationTree(OWProvidesLearnerI):
     name = "Classification Tree"
     icon = "icons/ClassificationTree.svg"
     description = "Classification tree algorithm with forward pruning."
     priority = 30
 
     LEARNER = TreeLearner
+    OUTPUT_MODEL_CLASS = TreeClassifier
```
kernc: Learners should define `__returns__`, which is a type of the returned model, so e.g. `TreeClassifier == TreeLearner.__returns__`.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -24,3 +28,99 @@ class OWProvidesLearner(_OWAcceptsPreprocessor):
...
+    """ Metaclass that add inputs, outputs and model_name setting.
+    """
+    def __new__(mcls, name, bases, attrib):
+
+        # check whether it is abstract class
+        if attrib.get('name', False):
+
+            for attr in ['LEARNER', 'OUTPUT_MODEL_NAME', 'OUTPUT_MODEL_CLASS']:
+                if attr not in attrib:
+                    raise AttributeError("'{name}' must have  ")
```
kernc: I wish there was some generic method to check that all required attributes are defined/overriden that would seamlessly work with other base/meta classes, i.e. without hardcoded names like that.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -24,3 +28,99 @@ class OWProvidesLearner(_OWAcceptsPreprocessor):
...
+                    ("Learner", attrib['LEARNER']),
+                    (attrib['OUTPUT_MODEL_NAME'], attrib['OUTPUT_MODEL_CLASS'])
+                ]
+            # adds extra outputs
+            if attrib.get('extra_outputs', None):
+                attrib['outputs'] += attrib.pop('extra_outputs')
+
+            if attrib.get('inputs', None) is None:
+                attrib['inputs'] = [("Data", Table, "set_data")] + \
+                                   _OWAcceptsPreprocessor.inputs
```
kernc: You could just prepend `inputs` of all the `bases`? Likewise for outputs above instead of `extra_outputs`?

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -24,3 +28,99 @@ class OWProvidesLearner(_OWAcceptsPreprocessor):
...
+
+            for attr in ['LEARNER', 'OUTPUT_MODEL_NAME', 'OUTPUT_MODEL_CLASS']:
+                if attr not in attrib:
+                    raise AttributeError("'{name}' must have  ")
+
+            # allows have outputs = []
+            if attrib.get('outputs', None) is None:
+                attrib['outputs'] = [
+                    ("Learner", attrib['LEARNER']),
+                    (attrib['OUTPUT_MODEL_NAME'], attrib['OUTPUT_MODEL_CLASS'])
```
kernc: This could be devised from `LEARNER`, as above.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -24,3 +28,99 @@ class OWProvidesLearner(_OWAcceptsPreprocessor):
...
+                attrib['outputs'] += attrib.pop('extra_outputs')
+
+            if attrib.get('inputs', None) is None:
+                attrib['inputs'] = [("Data", Table, "set_data")] + \
+                                   _OWAcceptsPreprocessor.inputs
+
+            if attrib.get('model_name', None) is None:
+                attrib['model_name'] = Setting(attrib['OUTPUT_MODEL_CLASS'])
+
+        return super(ProviderMetaClass, mcls).__new__(mcls, name, bases, attrib)
```
kernc: I guess it's ok to make code Py3-only if it looks better, like `super().__new__(...)`.

kernc:
In general, I am pro such a change.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -24,3 +28,99 @@ class OWProvidesLearner(_OWAcceptsPreprocessor):
...
+                self.model.name = self.model_name
+                self.model.instances = self.data
+
+        self.send(self.OUTPUT_MODEL_NAME, self.model)
+
+    def send_report(self):
+        raise NotImplementedError()
+
+
+class ProviderMetaClass(WidgetMetaClass):
```
kernc: `Provider` sounds awfully broad.

s-alexey:
I have made some changes. @kernc can you review it?

s-alexey:
I have added AdaBoost classification and regression widgets. Please, take a look at it. Can somebody  provide an appropriate icon?

Orange/classification/majority.py
```diff
@@ -82,3 +82,5 @@ def predict(self, X):
 
     def __str__(self):
         return 'ConstantModel {}'.format(self.dist)
+
+MajorityLearner.__returns__ = ConstantModel
```
kernc: Aww, you are so considerate! Why didn't you use the same approach in naive_bayes.py below? :smiley:

Orange/widgets/regression/owlinearregression.py
```diff
@@ -45,30 +42,22 @@ class OWLinearRegression(OWProvidesLearner, widget.OWWidget):
...
         gui.separator(box, 20, 20)
         self.alpha_box = box2 = gui.widgetBox(box, margin=0)
         gui.widgetLabel(box2, "Regularization strength")
         self.alpha_slider = gui.hSlider(
-            box2, self, "alpha_index",
-            minValue=0, maxValue=len(self.alphas) - 1,
-            callback=self._alpha_changed, createLabel=False)
+                box2, self, "alpha_index",
+                minValue=0, maxValue=len(self.alphas) - 1,
+                callback=self._alpha_changed, createLabel=False)
```
kernc: I don't think I like this PyCharm's default double indentation on line continuation (can be changed). Neither [PEP-8](https://www.python.org/dev/peps/pep-0008/#indentation) nor [Google's Style Guide](https://google.github.io/styleguide/pyguide.html#Indentation) recommend it.

janezd: I agree. I hate it, too, when PyCharm does that, but haven't investigated whether it can be changed, so I always formatted it manually. (Thanks, @kernc.)

Double indentation is sometimes appropriate. Not here, though.

Orange/widgets/regression/owlinearregression.py
```diff
@@ -145,52 +132,38 @@ def commit(self):
...
-                    [attr.name for attr in predictor.domain.attributes]
-                coef_table = Table(domain, list(zip(coefs, names)))
-                coef_table.name = "coefficients"
-
-        self.send("Linear Regression", learner)
-        self.send("Model", predictor)
+            coefs = [self.model.intercept] + list(self.model.coefficients)
+            names = ["intercept"] + \
+                    [attr.name for attr in self.model.domain.attributes]
+            coef_table = Table(domain, list(zip(coefs, names)))
```
kernc: `domain` above has a single variable named 'coef' and one meta, 'name'. Does this line work?

Orange/widgets/regression/owrandomforestregression.py
```diff
@@ -1,35 +1,24 @@
...
+class OWRandomForestRegression(owrandomforest.OWRandomForest):
     name = "Random Forest Regression"
     description = "Random forest regression algorithm."
+    icon = "icons/RandomForest.svg"
+    priority = 40
 
     LEARNER = RandomForestRegressionLearner
-    outputs = [("Learner", LEARNER),
-               ("Model", RandomForestRegressor)]
+    OUTPUT_MODEL_NAME = "Model"
```
kernc: For other regressors above, it's called 'Predictor', here a 'Model'. We could unify these a bit more too. How about introducing `OWBaseClassifier` and `OWBaseRegressor` that set appropriate output model names?

kernc: For other regressors above, it's called 'Predictor', here a 'Model'. We could unify these a bit more too. How about introducing `OWBaseClassifier` and `OWBaseRegressor` that set appropriate output model names?

Compatible with established terminology, we could also just go with 'Model' everywhere. @astaric, @janezd what do you think?

Orange/widgets/regression/owlinearregression.py
kernc: Settings > Editor > Code Style > Python > Tabs and Indents > Continuation Indent.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -1,26 +1,187 @@
...
+        if self.data:
+            self.report_data("Data", self.data)
+
+    # GUI
+    def add_learner_name_widget(self):
+        gui.lineEdit(self.controlArea, self, 'learner_name', box='Name',
+                     tooltip='The name will identify this model in other widgets')
+
+    def add_main_layout(self):
+        pass
```
kernc: All overridable methods should be decorated `Orange.util.abstract`, put at the top of the class, and docstringed _extensively_.

Orange/widgets/tests/test_providers.py
```diff
@@ -0,0 +1,53 @@
...
+from Orange.preprocess.preprocess import Preprocess
+from Orange.widgets.utils.owlearnerwidget import OWBaseLearner
+
+
+class TestProviderMetaClass(unittest.TestCase):
+
+    def assertChannelsEqual(self, first, second, msg=None):
+        self.assertEqual(len(first), len(second), msg)
+        self.assertEqual(set(i.name for i in first),
+                         set(i[0] for i in second), msg)
```
kernc: The first test is contained in the second.

Orange/widgets/regression/owrandomforestregression.py
janezd: I can give you my current preference, but I'm known to occasionally change my mind, especially w.r.t. terminology. :)

*Model* can (often) be used as *predictor*. Model is a result of modelling; like in *supervised and unsupervised modelling*. For instance, classifiers and regressors are called *predictive models*. Clustering is also a model (of the data), although typically not a predictor.

*Model* refers to content ("model describes the data"), while *Predictor* describes the use ("predictor classifies the data instances").

From this point on, I leave the decision to others. :) I don't care so much.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -1,26 +1,187 @@
...
+
+    def update_model(self):
+        self.model = None
+        self.good_data = False
+        if self.data is not None:
+            self.error(self.DATA_ERROR_ID)
+            if not self.learner.check_learner_adequacy(self.data.domain):
+                self.error(self.DATA_ERROR_ID, self.learner.learner_adequacy_err_msg)
+            elif len(np.unique(self.data.Y)) < 2:
+                self.error(self.DATA_ERROR_ID, "Data contains only one target value.")
```
kernc: "Data contains one single target value. There is nothing to learn."

kernc:
I like this PR very much! It solves a lot of bugs we had with particular learners, and it's a model of how OOP should be. If @janezd or @BlazZupan confirm everything still works, I'd sure like to see it merged ASAP. :smiley:

Orange/widgets/regression/owlinearregression.py
s-alexey: Yes, it's work. zip transposes coefs and names into two columns matrix.

Orange/widgets/regression/owrandomforestregression.py
s-alexey: I have thought about it too =)

We often inherit regression widgets from classification ones (KNNRegression, TreeRegression). That complicates creating separated base classes. But we can add a new attribute (`type = CLASIFFICATION` or `type = REGRESSION`) to base class.

Orange/widgets/tests/test_providers.py
s-alexey: The first assert is used to check whether a new item can overwrite default one with the same name.

Orange/widgets/classify/owsvmclassification.py
```diff
@@ -1,17 +1,15 @@
...
-from Orange.widgets import widget, settings, gui
-from Orange.widgets.utils.owlearnerwidget import OWProvidesLearner
-from Orange.widgets.utils.sql import check_sql_input
+from Orange.classification.svm import SVMLearner, NuSVMLearner
+from Orange.widgets import settings, gui
+from Orange.widgets.utils.owlearnerwidget import OWBaseLearner
 
 
-class SVMBaseMixin(OWProvidesLearner):
+class SVMBaseMixin(OWBaseLearner):
```
astaric: SVMBaseMixin is now used as a sole parent of SVM widgets, so Mixin could be dropped from the name.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -1,26 +1,209 @@
...
+
+        if self.data:
+            self.report_data("Data", self.data)
+
+    # GUI
+    def add_learner_name_widget(self):
+        gui.lineEdit(self.controlArea, self, 'learner_name', box='Name',
+                     tooltip='The name will identify this model in other widgets')
+
+    @abstract
```
astaric: Is this method really abstract?

Currently, Majority and Naive Bayes widgets do not work, as they do not overload this method (although they do not provide any additional options).

Orange/widgets/classify/owadaboost.py
```diff
@@ -0,0 +1,76 @@
...
+class OWAdaBoostClassification(OWBaseLearner):
+    name = "AdaBoost"
+    description = "An AdaBoost classifier."
+    icon = "icons/AdaBoost.svg"
+    priority = 80
+
+    LEARNER = SklAdaBoostLearner
+    OUTPUT_MODEL_NAME = "Classifier"
+
+    inputs = [("Classifier", LearnerClassification, "set_base_model")]
```
astaric: Classifier is name of the outputs that contain trained *model*, but here, a learner is expected. Should this be renamed to learner?

Orange/widgets/regression/owadaboostregression.py
```diff
@@ -0,0 +1,66 @@
...
+    losses = ["linear", "square", "exponential"]
+
+    learner_name = Setting("AdaBoost Regression")
+    n_estimators = Setting(50)
+    learning_rate = Setting(1.)
+    loss = Setting(0)
+
+    def add_main_layout(self):
+        self.base_estimator = TreeRegressionLearner()
+        box = gui.widgetBox(self.controlArea, "Parameters")
```
astaric: OWAdaBoostRegression is missing attribute self.base_label, which is used in parent's set_base_model method.

s-alexey:
@astaric, thanks for comments. I have fixed everything.

Orange/widgets/regression/owrandomforestregression.py
kernc: Having separate base classes could help us limit what connects to what (i.e. some widget accepts input 'Learner', but only of `OWBaseClassifier` type). if we ever decide for/need it.

How about simple overriding:
```py
class OWClsX(OWBaseClassifier):
    ...

class OWRegX(OWClsX):
    OUTPUT_MODEL_NAME = OWBaseRegressor.OUTPUT_MODEL_NAME
    ...
```
Seems the same number of symbols will need to be imported anyway. If you'd rather go the metaclass way, you're welcome too!

Orange/widgets/tests/test_providers.py
kernc: Not sure I understand. The second assert will fail if the two sets are not of equal length – exactly what the first assert tests.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -1,26 +1,207 @@
...
+                     "Press Apply to submit changes.")
+
+    def get_model_parameters(self):
+        """Creates an `OrderedDict` or a sequence of pairs with current model
+        configuration.
+
+        Returns:
+            OrderedDict or List: (option, value) pairs or dict
+        """
+        return None
```
kernc: Can also return `None`? I guess let's keep the code more invariant by returning an empty list-like iterable by default, e.g. `[]`.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -1,26 +1,207 @@
...
+            self.report_data("Data", self.data)
+
+    # GUI
+    def add_learner_name_widget(self):
+        gui.lineEdit(self.controlArea, self, 'learner_name', box='Name',
+                     tooltip='The name will identify this model in other widgets')
+
+    def add_main_layout(self):
+        """Creates layout with the learner configuration widgets.
+        """
```
kernc: Please move these meant-to-be-overridden methods to the top. Perhaps also be more verbose:
> Creates layout with the learner configuration widgets.

> Override this method for laying out any learner-specific parameter controls.
> See ``setup_layout()`` for execution order.

Orange/widgets/regression/owrandomforestregression.py
astaric: I am not sure that separate classes are the way to go, as I would rather see a single widget that is used for both regression and classification (but this is out of scope for this PR).

Regression widgets extending classification widgets is IMHO good enough, so I would just make sure that all regression widgets use the same name for model output (I vote for Predictor).

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -1,26 +1,229 @@
...
+    def default_inputs(cls, attrib):
+        return [("Data", Table, "set_data"), ("Preprocessor", Preprocess, "set_preprocessor")]
+
+    @classmethod
+    def default_outputs(cls, attrib):
+        learner_class = attrib['LEARNER']
+        if issubclass(learner_class, LearnerClassification):
+            model_name = 'Classifier'
+        else:
+            model_name = 'Predictor'
```
s-alexey: We can choose output model name here.

Orange/widgets/regression/owrandomforestregression.py
s-alexey: We can just check `LEARNER` superclass. If `LEARNER` is a subclass of LearnerClassifier, `OUTPUT_MODEL_NAME = "Classifier"` and "Predictor" in the other case.

Orange/widgets/utils/owlearnerwidget.py
kernc: Looks reasonable to me.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -1,26 +1,229 @@
...
+                channel.append(item)
+
+        return channel
+
+    @classmethod
+    def add_extra_attributes(cls, name, attrib):
+        return attrib
+
+
+class DefaultLearnerWidgetChannelsMetaClass(DefaultWidgetChannelsMetaClass):
```
kernc: Can you just rename this to something more up-to-date, like `OWBaseLearnerMeta`.

astaric:
This PR looks complete to me and I would like to merge it today.

Is this ok with you or were you planning to add anything else to it?

s-alexey:
@astaric I have made some changes right now. 

It looks complete to me too.

PR #1074, opened 2016-03-08
[WIP] Selection of a single sheet from Excel 

rahulroxx:
@janezd  I didn't added recentFile support due to recent merge in owfile.py , i therefore made a mechanism to get multiple sheets if any from a single file and update in combobox 
Please help in context switching and maintaining and triggering recent paths for sheets ! 
Thanks !

screenshot from 20160308 05-20-05


Orange/widgets/data/owfile.py
```diff
@@ -93,7 +94,12 @@ def __init__(self):
         box.setSizePolicy(Policy.MinimumExpanding, Policy.Fixed)
         self.file_combo.setSizePolicy(Policy.MinimumExpanding, Policy.Fixed)
         self.file_combo.activated[int].connect(self.select_file)
+        self.sheet_combo = sheet_combo = QtGui.QComboBox(box) 
+        sheet_combo.setSizePolicy(Policy.MinimumExpanding, Policy.Fixed) 
+        sheet_combo.setVisible(False)
```
kernc: According to [contrib guidelines](https://github.com/biolab/orange3/blob/master/.github/CONTRIBUTING.md#coding-style), prefer Qt properties specified as kwargs to the constructor.

Orange/widgets/data/owfile.py
```diff
@@ -179,6 +185,30 @@ def _url_set(self):
...
 
+    #Filling comboBox With Sheet Names
+    def fill_sheet_combo(self, filename):
+        _, filename = os.path.split(filename)
+        _,sheetname = filename.split(':')
+        self.sheet_combo.addItem(sheetname)
+
+    #Checking For multiple sheets otherwise default legacy code        
+    def multiple_excelsheets(self, filename):
+        book = xlrd.open_workbook(filename)
```
kernc: Does this work if `filename` is not an Excel file?

Orange/widgets/data/owfile.py
```diff
@@ -179,6 +185,30 @@ def _url_set(self):
         self.source = self.URL
         self.load_data()
 
+    #Filling comboBox With Sheet Names
```
kernc: These types of comments are wrong. Wherever you've seen them, forget it. If anything, use docstrings.

Orange/widgets/data/owfile.py
```diff
@@ -194,11 +224,7 @@ def browse_file(self, in_demos=False):
             self, 'Open Orange Data File', start_file, self.dlg_formats)
         if not filename:
             return
-
-        self.add_path(filename)
-        self.source = self.LOCAL_FILE
-        self.load_data()
-
+        self.multiple_excelsheets(filename)
```
kernc: This program flow is not so nice. The function is called `multiple_excelsheets` but it really handles all types of files? I'd comparatively prefer an `is_excel(filename)` method and the rest of the lines remaining here.

rahulroxx:
@kernc Thanks For pointing out the mistakes 
I will make a class that handles any type of file  , remove comments 
I will improve the above in next PR !

kernc:
@rahulroxx, you don't need to (read: _don't_) open a new PR. You can amend _this one_ by adding commits to your `rahulroxx/owfile` branch. You can also replace commits by [force-pushing](http://stackoverflow.com/questions/10510462/force-git-push-to-overwrite-remote-files) to it.

rahulroxx:
@kernc refactored code

Orange/widgets/data/owfile.py
```diff
@@ -197,8 +200,33 @@ def browse_file(self, in_demos=False):
...
+            book = open_workbook(fn)
+            has_sheets = book.nsheets
+            if has_sheets > 1:
+                return True
+
+        except XLRDError:
+            return False
+
+        else:
+            return False
```
kernc: These whole function could read:
```py
try:
    return xlrd.open_workbook(fn).nsheets > 1
except xlrd.XLRDError:
    return False
```
Which do you prefer?

Orange/widgets/data/owfile.py
```diff
@@ -219,6 +247,14 @@ def load_from_file():
                     fn = os.path.join(".", basename)
                     self.information("Loading '{}' from the current directory."
                                      .format(basename))
+            if self.excel_has_sheets(fn):
+                book = open_workbook(fn)
+                if book.nsheets > 1:
```
kernc: If `excel_has_sheets` returned `True`, then `book.nsheets` is already known to be >1?

rahulroxx:
@kernc
```
 try:
    return xlrd.open_workbook(fn).nsheets > 1
except xlrd.XLRDError:
    return False
``` 
select excel sheets Only otherwise return False ,
yes that was a redundant check `book.nsheets > 1` .  I have changed function name to is_multsheet_excel() and made it static

rahulroxx:
@kernc Thanks ! I have made is_multisheet_excel() staticmethod !

janezd:
@janezd i didn't yet used ExcelFormat.read(excelfile:sheetname) , i have passed the filename which
   and ExcelFormat.read(excelfile:sheetname) takes sheet[0] for now ! 
  i'll fix this with context settings

Settings come last. It's better to finish the widget first. Can you make the combo fully functional first? The way the setting will work may depend on that.

Can you put the new combo below the file name, in an extra line below the first line? Also, the combo should be hidden if the file is not multisheet xls. Currently the widget calls `self.sheet_combo.setVisible(True)`, but not `self.sheet_combo.setVisible(False)`.

rahulroxx:
@janez i made the whole widget working  

> Can you put the new combo below the file name, in an extra line below the first line?

below first comboBox ?

janezd:
Yes. I guess (but I haven't tried) it would look nice if you put the sheet selection combo below the first combo, so that both also have the same width.

rahulroxx:
@janezd  i have tried to `self.sheet_combo.move(xpos , ypos) ` but it seems to be not working in the current layout .

kernc:
@rahulroxx We use [Qt's vertical and horizontal boxes](http://doc.qt.io/qt-4.8/layout.html) for most of layouting. Perhaps you can introduce one of those?

rahulroxx:
@kernc       
        ```
** layout = QtGui.QGridLayout()**
        gui.widgetBox(self.controlArea, margin=0, orientation=layout)
        vbox = gui.radioButtons(None, self, "source", box=True, addSpace=True,
                                callback=self.load_data, addToLayout=False)

        rb_button = gui.appendRadioButton(vbox, "File", addToLayout=False)
        layout.addWidget(rb_button, 0, 0, QtCore.Qt.AlignVCenter)

        box = gui.hBox(None, addToLayout=False, margin=0)
        box.setSizePolicy(Policy.MinimumExpanding, Policy.Fixed)
        self.file_combo.setSizePolicy(Policy.MinimumExpanding, Policy.Fixed)
        self.file_combo.activated[int].connect(self.select_file)
        self.sheet_combo = QtGui.QComboBox(box)
        self.sheet_combo.setSizePolicy(Policy.MinimumExpanding, Policy.Fixed)
        self.sheet_combo.activated[str].connect(self.select_sheet)
        self.sheet_combo.setVisible(False)
**        box.layout().addWidget(self.file_combo)
        box.layout().addWidget(self.sheet_combo)**
        _self.sheet_combo.move(30,30)_
```
File widget already has GridLayout and i have added combox named sheet_combo but self.sheet_combo,move(**args) doesn't shifted the comboBox

**Comment by kernc** [comment]
Yes, because you can't `move` laid-out items by pixels. You put them in a different layout. In the case above, both combos are added to the layout of widget box `box`, which is, if you look at its definition, an `hBox`, so a horizontal box.

Now if you make it vertical ...

**Comment by rahulroxx** [comment]
@kernc thanks . i made a child vBoxLayout and added it to parent QGridLayout . @janezd  now Sheet_combo appears below file_combo :)

**Comment by rahulroxx** [comment]
@janezd now the widget is fully functional , please tell me futher steps for context settings , if there are no remainders . Thanks  !
rr

**Comment by ajdapretnar** [comment]
Can I dislike the GUI as presented? :( I'd much rather have file combobox across the entire window and have sheet, browse... and reload in one line below. Url entry should also extend across the entire window. Now this just looks haphazard.... @rahulroxx Not your fault though, we're just being picky. @janezd What do you think?

**Comment by janezd** [comment]
Sure, this is not OK. The widget now has a lot of empty space and short combos. (I fixed URL two weeks ago - hasn't this been merged?)

I imagined that the second combo would appear below the first, but reload buttons would stay in the first line -- not vertically centered, as they are now.

I like @ajdapretnar's suggestion, too. The problem is that when the widget loads non-excel or single-sheet excel file, there is no second combo, so the buttons are in the first line.

This is one of the things that has to be tried and seen before we can decide. The current layout is not OK, though.

**Comment by rahulroxx** [comment]
@janezd @ajdapretnar yes my repo is up to date . This screenshot is only when excel has multiple sheets otherwise the default code runs and the layout is not messed up .  I agree with you both and i also didn't liked this layout i suggest the first screenshot i posted in this PR . what do you think ?

**Comment by ajdapretnar** [comment]
Ah yes, I forgot the 'sheet' button is hidden when the file contains only one sheet.

However, I have another idea. While I like the first suggestion from @rahulroxx, the problem is all comboboxes might get squished too much. I prefer the widget layout to breathe a bit more.
How about putting the reload button somewhere else? Urls don't get updated upon clicking 'reload', but I have a need to do this often. Reload should in my opinion be a 'stand-alone' button that works for both options. That would also allow us to put 'sheet' button next to the file combobox.

What do you think?

**Comment by rahulroxx** [comment]
@ajdapretnar Nice idea ! . I too faced problem for filling url all the time while testing . 

> How about putting the reload button somewhere else?

Below file icon ?

**Comment by ajdapretnar** [comment]
I would much rather (and I think @janezd will agree) have the report button totally at the bottom across the entire window (as is in other widgets) and have 'browse documentation sets' and 'reload' together in one line.
TBD.
Also @rahulroxx sorry for being so nitpicky about the GUI, but while we're at it, we can make it look great, right? :)

**Comment by rahulroxx** [comment]
@ajdapretnar Yes ! we will do it :+1: 
 Reload button at bottom with both url and file combo functionality according to second screenshot layout ?

**Comment by ajdapretnar** [comment]
@rahulroxx Ummm, it was 'Report' button at the bottom and 'Reload' next to 'Browse documentation sets'. =) But yes, Reload should reload the selected data type, either a file or url.
Just wait please for @janezd to confirm the layout.

**Comment by rahulroxx** [comment]
@ajdapretnar OK :)

**Comment by janezd** [comment]
Just wait please for @janezd to confirm the layout.

I can't. I have to see it before confirming it. :)

Seriously, when I program widgets I'm playing with layouts a lot before finding one that I like. It's impossible to know in advance.

**Comment by rahulroxx** [comment]
@janezd @ajdapretnar reload button at bottom , better ?
fin

**Comment by rahulroxx** [comment]
@janezd @ajdapretnar  Left Aligned the browse_file 
fial

**Comment by janezd** [comment]
The widget still has too much empty space. The "..." should be in the first or second line, not in between.

As for Reload ... it's lost down there. After #1012 (the screenshot below) is finished and merged, this will be even worse.

@ajdapretnar, why would you like to separate Reload from the combos? To reload from URL, do the same as in the browser: go to URL and press return. What about having the Reload and "..." in the first line if there is no sheet combo, and by the sheet combo, when there is? Just an idea, I can't guarantee it'll look OK and feel useful.

screen shot 2016-03-16 at 17 02 49

**Comment by ajdapretnar** [comment]
@janezd This is a nice layout! But I still don't think pressing Enter to reload url data is fun. :P (read: I don't like two different methods to reload data sets) I vote for Reload button to work for both, but it's definitely up to you (and whoever is going to submit a PR).

**Comment by rahulroxx** [comment]
@ajdapretnar  in the file widget it is optional to load data either from file or from url . 
but having a common reload function would send the instance data ( file + url ) to another widget . Wouldn't it be conflicting situation ? or the latter will override the data . Please correct me if i am wrong .

**Comment by janezd** [comment]
Reload button would reload the currently selected source. But I don't know where to put it for this function. If it's in the same line as combo, it applies only to that combo. If it's at the bottom, it's lost.

**Comment by rahulroxx** [comment]
@janezd why lost ?

**Comment by janezd** [comment]
Lost in the dialog. Not at its obvious place. The reload button in your browser is next to the url, for instance.

**Comment by rahulroxx** [comment]
@janezd just a suggestion !  if we place report between file and url like this 
io
?

**Comment by janezd** [comment]
I like the way it looks now - but it's not obvious that Reload reloads URL's, too. Also, the buttons should be aligned (and of the same height) as the combos. In the existing layout, they are, at least on Os X; could be different on Linux.

How does the layout look like when the sheet combo is hidden?

**Comment by rahulroxx** [comment]
@janezd it doesn't look nice ! 
bke

**Comment by kernc** [comment]
When the window is made [a bit wider](https://github.com/biolab/orange3/pull/1074#issuecomment-197402013), I don't have anything against seeing both combos in the same line. One holds the filename (usually short, <30 characters), the other sheet name (usually short, <30 characters).

**Comment by janezd** [comment]
This will be especially true in the new layout. Perhaps the whole moving of the combos was a bad idea. Sorry.

**Comment by rahulroxx** [comment]
@janezd what would you suggest ?

**Comment by rahulroxx** [comment]
@janezd @kernc @ajdapretnar 
recent fix in spacing now gui looks like this with combobox 
yup






without comboBox 
yipee

**Comment by rahulroxx** [comment]
@janezd @kernc @ajdapretnar 
recent fix in spacing now gui looks like this 
**without sheet combobox** 
yup

**with sheet combobox**
tt

**Comment by rahulroxx** [comment]
i like this one !
**with sheet_combo**
with

**without sheet_combo**

without

what do you think ?

**Comment by ajdapretnar** [comment]
I don't think having an empty space when sheet_combo is not present is a good idea, sorry. :(

Here are some other options we can go with. So sorry for the fugly presentation, my skills in Qt Designer are at a level of a blind monkey, but you get the idea behind it.

file-widget-suggestion

**Comment by rahulroxx** [comment]
@ajdapretnar Thanks a lot ! I liked the first two of your layout . Can i use first two   to start redesining the widget ?

**Comment by ajdapretnar** [comment]
@rahulroxx Sure, if you can make it look actually presentable (unlike mine), so ahead. But obviously BDFL @janezd has to confirm the final layout. :)

**Comment by rahulroxx** [comment]
@ajdapretnar Thanks ! :)

**Comment by janezd** [comment]
Apologies for being destructive. The first two layouts assume that the browse button is rather wide. Also, the browse button should be to the right or below the file combo, not the url combo.

I like the second row, but @ajdapretnar doesn't, I guess. In this layout, reload reloads only the file.

I have one more. Reload is now again at the bottom (which I really don't like - I can imagine that some users will interpret it as OK or Apply, that is, as if they *have* to push it to load the data), but it's not in the middle and it should have an icon to be more visible. (Also, Browse documentation should not be "default", but I don't know how to disable this in Qt Designer.)

The layout will also look reasonable when the second line is hidden.

screen shot 2016-03-17 at 12 11 20

For this layout, the two combos and sheet need to be aligned, and also the rows have to be properly align. I guess QGridLayout (with HBox within the second row) would do the trick.

**Comment by rahulroxx** [comment]
what if we place "Reload" button in between both the buttons ? and i would like to see the layout if the sheet_combo is hidden

**Comment by janezd** [comment]
I don't think Report will be much used, so I prefer to have the more commonly used Reload on the outside. This will probably changed in the future anyway, since we're preparing some additions to the file widget.

If the file is not an excel file or it has just a single sheet, hide the second row (Sheet and combo). If the file combo refers to multisheet excel, the second row is shown even if the user actually loads URL.

If the URL radio button is chosen and the user selects a sheet in the sheet combo, the File radio button has to be checked automatically.

**Comment by rahulroxx** [comment]
@janez is this we were initially expecting ? 
fixwithout

with combo 

fixedwith

**Comment by janezd** [comment]
Either this (and Reload reloads just the file) or what I've drawn above (and Reload reloads both). In either case, I guess it's better if you add Sheet in front of the sheet combo.

**Comment by rahulroxx** [comment]
@janezd 

>  I guess it's better if you add Sheet in front of the sheet combo.

yaa sure ! 

Do everyone agree with this layout ?

**Comment by rahulroxx** [comment]
@janezd can i use this layout for further work ? 
makefinal

**Comment by rahulroxx** [comment]
@janezd The widget is fully functional with layout . would you please help me on context settings ?

**Comment by janezd** [comment]
> If the URL radio button is chosen and the user selects a sheet in the sheet combo, the File radio button has to be checked automatically.

Also, the combos should be aligned to the right, like in your screenshot (except with the label sheet) and in my pictures. This below is not OK.

screen shot 2016-03-18 at 20 00 05

Do not specify widths (you can increase the minimal width to the above combo, though). Instead, use `QGridLayout` and put the two combos into the same column.

**Comment by janezd** [comment]
I cannot guide you through contexts settings. It requires writing a new context handler, which is not documented AFAIK. I'll try to do it myself, and I'll send you the code to include in your PR.

File: Orange/widgets/data/owfile.py [hunk 7]
```diff
@@ -197,11 +213,37 @@ def browse_file(self, in_demos=False):
...
         self.source = self.LOCAL_FILE
+        self.fill_sheet_combo(filename)
         self.load_data()
 
 
+    def fill_sheet_combo(self, path):
+        if os.path.exists(path) and self.is_multisheet_excel(path):
+            book = open_workbook(path)
+            sheets = book.nsheets
+            if sheets > 1:
```
janezd: Is this necessary? Doesn't `self.is_multisheet_excel` check this?

Orange/widgets/data/owfile.py
```diff
@@ -197,11 +213,37 @@ def browse_file(self, in_demos=False):
...
+                self.sheet_label.setVisible(True)
+                self.sheet_combo.setVisible(True)
```
janezd: It would be more elegant to put both controls into an hBox and hide the hBox. If you use `QGridLayout`, you will have to put them into such box anyway.

Orange/widgets/data/owfile.py
```diff
@@ -101,11 +101,26 @@ def __init__(self):
         button = gui.button(
             box, self, "Reload", callback=self.reload, autoDefault=False)
         button.setIcon(self.style().standardIcon(QtGui.QStyle.SP_BrowserReload))
-        button.setSizePolicy(Policy.Fixed, Policy.Fixed)
+        button.setSizePolicy(Policy.Fixed, Policy.Fixed) 
+        hBLayout = QtGui.QHBoxLayout()
+        self.sheet_combo = QtGui.QComboBox()
+        self.sheet_combo.setSizePolicy(Policy.MinimumExpanding, Policy.Fixed)
+        self.sheet_combo.setSizePolicy(Policy.MinimumExpanding, Policy.Fixed)
```
janezd: Elephants and pythons never forget. Setting the size policy once suffices. No need to repeat. :))))

janezd:
`__init__ calls

    QtCore.QTimer.singleShot(0, self.load_data)

which loads the most recent file. If this file was an Excel file, the combo box must be filled and shown -currently it remains hidden. Calling `fill_sheet_combo` before this line would probably suffice. Please try and thoroughly check.

Orange/widgets/data/owfile.py
```diff
@@ -219,6 +261,13 @@ def load_from_file():
                     fn = os.path.join(".", basename)
                     self.information("Loading '{}' from the current directory."
                                      .format(basename))
+            if self.is_multisheet_excel(fn):
+                #book = open_workbook(fn)
+                if sheet:
+                    data = ExcelFormat.read_file(fn+':'+sheet)
```
janezd: Add spaces before and after `+`.

Orange/widgets/data/owfile.py
kernc: Please follow the [coding style for setting non-changing Qt properties](https://github.com/biolab/orange3/blob/master/CONTRIBUTING.md#coding-style). For many calls below too.

janezd: Most of these are mine. I just recently learned about this feature. I promise to behave in the future.

rahulroxx:
@janezd 
without fix width it looks like this 
odpmt

janezd:
That's why I recommended the `QGridLayout`. Put the box with the sheet label and the corresponding combo in the same column as the file combo, and set the size policy to MinimumExpanding (no need for aligning). They will fill the entire column and be right-aligned. Put the ... and reload button in a separate column.

rahulroxx:
@janezd 
i did exactly the following 
file & sheet combo placed in same column in gridLayout 
policy min expanding of both label and sheet 
separate column for ... and reload  done :+1: 
proposedlayout

janezd:
As I see in the snapshot of the code, `file_combo`, `button` and the other `button` are all in the same `box`, hence they occupy the same column.

Moreover, the label Sheet is now under the radio button file instead of under the file combo.

I would try a 3x3 grid and place the elements as follows (... and reload, and the sheet label and combo are in a horizontal children layout).

    File  | file combo               |  ..., reload
          | sheet label, sheet combo |
    Url   | url combo spanning over two columns

rahulroxx:
@janezd Moving Reload button to 0,3 in grid gives me error 

```
  File "owfile.py", line 393, in <module>
    ow = OWFile()
  File "owfile.py", line 162, in __init__
    self.set_file_list()
  File "/home/rahul/Desktop/orange3/Orange/widgets/utils/filedialogs.py", line 342, in set_file_list
    self.file_combo.clear()

```

janezd:
I can't see neither your code nor the error, but an error in `self.file_combo_clear()` does not seem directly related to layouts. Please check your code more carefully.

rahulroxx:
@janezd i missed last line :) 
@janezd Moving Reload button to 0,3 in grid gives me error 

```
  File "owfile.py", line 393, in <module>
    ow = OWFile()
  File "owfile.py", line 162, in __init__
    self.set_file_list()
  File "/home/rahul/Desktop/orange3/Orange/widgets/utils/filedialogs.py", line 342, in set_file_list
    self.file_combo.clear()

```
RuntimeError: wrapped C/C++ object of type QComboBox has been deleted

rahulroxx:
error

janezd:
You must be doing something wrong. Discover what and fix it...

rahulroxx:
@janez Yes !

rahulroxx:
@janezd 

> Is this necessary? Doesn't self.is_multisheet_excel check this?
Yes , it is required Because the fill_combo only fills sheets when there are multiple sheets  

```
        QtCore.QTimer.singleShot(
            0, lambda: self.fill_sheet_combo(self.last_path()))
```
Here , is_multiple_sheet() is not preceeding .

finalwith

finalwithout

janezd:
I'm not sure about this.

You have

    if os.path.exists(path) and self.is_multisheet_excel(path):
        book = open_workbook(path)
        sheets = book.nsheets
        if sheets > 1:

where `is_multisheet_excel` is defined as

    def is_multisheet_excel(fn):
        try:
            return open_workbook(fn).nsheets > 1
        except XLRDError:
            return False

Both if's check the same condition, (n)sheets > 1.

You don't need to call `fill_sheet_combo` through timer since it doesn't involve the progress bar. Call it directly, also to be sure that it is called before loading the data.

rahulroxx:
@janezd Exactly , the latter if in file_combo seems to be redundant

rahulroxx:
@janezd Exactly , the latter if in `sheet_combo` seems to be redundant 

Synced Reload Button with sheet_combo

rahulroxx:
@janezd Do I need to do changes in `RecentWidgetComboMixing` in `../utils/fildialogs.py`, tell me if i missed something ?

janezd:
No need to change the mixin. Just move the combo into its own grid element.

If I try to load Orange with fresh settings the widget doesn't open at all, and it gives me

```
  File "/Users/janezdemsar/Dropbox/orange3/Orange/canvas/document/schemeedit.py", line 1253, in __onNodeActivate
    widget = self.scheme().widget_for_node(node)
  File "/Users/janezdemsar/Dropbox/orange3/Orange/canvas/scheme/widgetsscheme.py", line 63, in widget_for_node
    return self.widget_manager.widget_for_node(node)
  File "/Users/janezdemsar/Dropbox/orange3/Orange/canvas/scheme/widgetsscheme.py", line 191, in widget_for_node
    state = self.__materialize(state)
  File "/Users/janezdemsar/Dropbox/orange3/Orange/canvas/scheme/widgetsscheme.py", line 229, in __materialize
    widget = self.create_widget_instance(node)
  File "/Users/janezdemsar/Dropbox/orange3/Orange/canvas/scheme/widgetsscheme.py", line 316, in create_widget_instance
    widget.__init__()
  File "/Users/janezdemsar/Dropbox/orange3/Orange/widgets/data/owfile.py", line 161, in __init__
    self.fill_sheet_combo(self.last_path())
  File "/Users/janezdemsar/Dropbox/orange3/Orange/widgets/data/owfile.py", line 225, in fill_sheet_combo
    if os.path.exists(path) and self.is_multisheet_excel(path):
  File "/Users/janezdemsar/env/o3/bin/../lib/python3.5/genericpath.py", line 19, in exists
    os.stat(path)
TypeError: stat: can't specify None for path argument
```

rahulroxx:
@janez it seems like os can't find self.last_path() as i used the recent path for combo filling . I should therefore check if self.last_path() exists because the user will not have any recent paths when he uses Orange for First time !

rahulroxx:
Thanks Janez , 
I will fix this  as soon as I complete my  proposal and share as draft  :)

rahulroxx:
@janezd  I have fixed the issue of `--clear-widget-settings` in `owfile.py `

rahulroxx:
@janezd please have a look at it , tell me if it has more issues.

Orange/widgets/data/owfile.py
```diff
@@ -93,21 +93,39 @@ def __init__(self):
...
-        button = gui.button(
-            box, self, '...', callback=self.browse_file, autoDefault=False)
-        button.setIcon(self.style().standardIcon(QtGui.QStyle.SP_DirOpenIcon))
-        button.setSizePolicy(Policy.Maximum, Policy.Fixed)
-        button = gui.button(
-            box, self, "Reload", callback=self.reload, autoDefault=False)
-        button.setIcon(self.style().standardIcon(QtGui.QStyle.SP_BrowserReload))
-        button.setSizePolicy(Policy.Fixed, Policy.Fixed)
-        layout.addWidget(box, 0, 1,  QtCore.Qt.AlignVCenter)
+        layout.addWidget(self.file_combo, 0, 1)
```
janezd: If you do it like this, `box` is not inserted into the layout. Moreover, the sheet combo is not right-aligned to the file combo due to some additional spacing to the right (I've not idea where it comes from). Inserting the file combo into the box and box into the layout will align the two combos. Thus, replace this line with

        box.layout().addWidget(self.file_combo)
        layout.addWidget(box, 0, 1)

janezd:
The layout is almost fixed, except the margin (see the inline comment).

There's a bug in the logic of the widget, though. To recreate, open Excel and create a single-sheet file. Load it into the widget. The sheet combo remains hidden (as it should, since this is a single-sheet excel file). Now go back to Excel, add another sheet and save the file. Go to the widget and hit Reload. Since this is now multiple-sheets excel file, the combo should appear. It doesn't since `reload` just calls `load_data`.

rahulroxx:
@janezd I fixed the bugs . I have a query that a xls file contains 6 sheets , what happens when I am pointing to 'Sheet 6' in sheet_combo and I delete it from xls file  and save xls file ( now xls file contains 5 sheets ) , now where should the sheet_combo point after reloading ? what if sheet_combo is pointing to 'Sheet 3' in above case ?

janezd:
Good question. And the answer is even better: you don't care. Context settings take care of that. That's exactly why we have them.

Before populating the combo box, you need to call `self.closeContext`. Then populate the combo and *always* set it to the default (which is the first sheet). Finally, call `self.openContext` and it will change the selection if possible. This will also work if the user reorders the sheets, that is, the sheet with the same name will be selected after reload.

Speaking about the context settings, I see that they disappeared from your code?

rahulroxx:
@janezd Thanks ! I'm not familiar with context settings yet . I am trying the get the whole workflow with context settings . I will write the documentation for context settings too after I fully understand the workflow . 

> I see that they disappeared from your code?

Yes,

 I will include the them as you said  :+1:

rahulroxx:
@janezd I didn't merged your context settings yet due to continuous changes in GUI at that time . and it seems that branch has conflicts now ( because i updated code ) .  Please would you send it again with updated code ?

rahulroxx:
@janezd  
I faced error while adding context settings 

```
  File "owfile.py", line 212, in select_file
    self.fill_sheet_combo(self.last_path())
  File "owfile.py", line 266, in fill_sheet_combo
    self.openContext(path, sheet_names)
  File "/home/rahul/Desktop/orange3/Orange/widgets/widget.py", line 456, in openContext
  File "/home/rahul/Desktop/orange3/Orange/widgets/settings.py", line 469, in open_context
    self.settings_to_widget(widget)
  File "owfile.py", line 64, in settings_to_widget
    widget.xls_sheet = widget.current_context.xls_sheet
  File "/home/rahul/Desktop/orange3/Orange/widgets/widget.py", line 450, in __setattr__
    notify_changed(obj, field_name, value)
  File "/home/rahul/Desktop/orange3/Orange/widgets/gui.py", line 284, in notify_changed
    callback(value)
  File "/home/rahul/Desktop/orange3/Orange/widgets/gui.py", line 2637, in __call__
    self.action(*args)
  File "/home/rahul/Desktop/orange3/Orange/widgets/gui.py", line 2677, in action
    value = self.attribute2controlDict.get(value, value)
TypeError: unhashable type: 'list'
  File "owfile.py", line 212, in select_file
    self.fill_sheet_combo(self.last_path())
  File "owfile.py", line 266, in fill_sheet_combo
    self.openContext(path, sheet_names)
  File "/home/rahul/Desktop/orange3/Orange/widgets/widget.py", line 456, in openContext
  File "/home/rahul/Desktop/orange3/Orange/widgets/settings.py", line 469, in open_context
    self.settings_to_widget(widget)
  File "owfile.py", line 64, in settings_to_widget
    widget.xls_sheet = widget.current_context.xls_sheet
  File "/home/rahul/Desktop/orange3/Orange/widgets/widget.py", line 450, in __setattr__
    notify_changed(obj, field_name, value)
  File "/home/rahul/Desktop/orange3/Orange/widgets/gui.py", line 284, in notify_changed
    callback(value)
  File "/home/rahul/Desktop/orange3/Orange/widgets/gui.py", line 2637, in __call__
    self.action(*args)
  File "/home/rahul/Desktop/orange3/Orange/widgets/gui.py", line 2677, in action
    value = self.attribute2controlDict.get(value, value)
TypeError: unhashable type: 'list'
```

rahulroxx:
@janezd Added Context settings and cleaned code . 

Updation and Deletion of Excel sheets in ( Excel Works fine )

rahulroxx:
@janezd Please tell me if it has issues !

Orange/widgets/data/owfile.py
```diff
@@ -94,20 +117,40 @@ def __init__(self):
...
+        reloadButton = gui.button(
+            None, self, "Reload", callback=self.reload, autoDefault=False)
+        reloadButton.setIcon(self.style().standardIcon(
+            QtGui.QStyle.SP_BrowserReload))
+        reloadButton.setSizePolicy(Policy.Fixed, Policy.Fixed)
+        layout.addWidget(reloadButton, 0, 3)
+
+        self.hBLayout = gui.hBox(None, addToLayout=False, margin=0)
+        self.sheet_combo = gui.comboBox(None, self, "xls_sheet",
+                                        sendSelectedValue=True)
```
janezd: Context settings don't work.

Open workbook with two sheets. Select Sheet 2. Save schema. Close Orange. Open Orange. Load schema. You'll see that Sheet 1 is selected instead of Sheet 2.

To fix, add argument `callback=self.select_sheet` in construction of `self.sheet_combo` - like it was in my PR to your PR. :)

Please check again if the fix really works. (It works for me.)

Orange/widgets/data/owfile.py
```diff
@@ -94,20 +117,40 @@ def __init__(self):
...
+        self.sheet_combo.setSizePolicy(
+            Policy.MinimumExpanding, Policy.Fixed)
+        self.sheet_combo.activated[str].connect(self.select_sheet)
```
janezd: Remove this line. Attributes must be handled by `gui`'s callback, otherwise context settings do not work out of the box.

rahulroxx:
@janezd gui callback dones't take 1 additional function argument ( in my case self.select_sheet takes sheet_name as addition argument ) ,

rahulroxx:
@janezd  
`add argument callback=self.select_sheet in construction of self.sheet_combo`

But self.select_sheet takes one more argument ( sheet_name ) , and using `callback=self.select_sheet `

show up error that one more additional function parameter is missing . 

Is there any way to pass `sheet_name`

rahulroxx:
@janezd Thanks !

rahulroxx:
Actually the problem was if was using `self.select_sheet(self, sheet_name) ` , callback=self.select_sheet doesn't take any additional parameter i.e sheet_name that's why I used signals out of the box  but now , I fixed this with your pull request . ( with context settings )

self.sheet is replaced by self.xls_sheet , I wasn't familiar with context settings at that time . 
Thanks again :) 

But now this  issue is fixed ! Please have a look now !

rahulroxx:
@janezd @kernc 

Do we need to include specific sheet ( from xls file ) and data in send_report in owfile widget ?

kernc:
Sure. Reports are meant to be reproducible.

rahulroxx:
@kernc Thanks !

rahulroxx:
@kernc Please have a look now ! I have updated changes and probably layout and functionality of add on sheet combo is fixed . Please let me know if there are issues . 
Thanks :)

janezd:
Done!

I have added a small fix for context settings, squashed your commits into one, so they are now in PR #1164.

PR #1090, opened 2016-03-09
setup.py: Remove build date from the public version identifier

ales-erjavec:
This partially reverts commit 36f27c48c2c16a0c4da781e6ff01e1291bcc27db.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -131,9 +131,7 @@ if not release:
         GIT_REVISION = "Unknown"
 
     if not ISRELEASED:
-        from datetime import datetime
-        timestamp = datetime.now().strftime('%y%m%d')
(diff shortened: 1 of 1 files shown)


kernc:
Public release number? It was appended only for versions _not_ "released."

Why not?

ales-erjavec:
> Public release number? It was appended only for versions not "released."

Sorry, that should be 'public version number'.

> Why not?

Because two source/binary distributions build for the exact same code checkout must not have two different versions just because they were build/installed on different days. Whether they are final releases or not is irrelevant.

Also, what exactly will happen when a final release is ready?

ales-erjavec:
I updated the commit message and pull request title (public release number -> public version identifier).

Issue #1080, opened 2016-03-09
Report for Univariate Regression


ajdapretnar:
Implement Report button function in Univariate Regression as is in other widgets (e.g. Scatter Plot widget). Report on learner name, model parameters and data.

iamved:
I have been done with learner name and data but facing issue with adding parameters. Please tell me how to add variable such polynomial expansion and Target. I tried for both but getting some errors.

janezd:
See how it's done in #1092. Also, see comments in #1091.

iamved:
sorry I forget to delete the comment and the issue was solved already at that point of time.

PR #1091, opened 2016-03-09
 Report for Univariate Regression #1080

RachitKansal:
@ajdapretnar please review this code.

### DIFF
--- a/Orange/widgets/regression/owunivariateregression.py
+++ b/Orange/widgets/regression/owunivariateregression.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 from PyQt4.QtGui import QLayout
 from PyQt4 import QtCore
 from PyQt4 import QtGui
@@ -15,6 +16,7 @@ from Orange.widgets import widget, settings, gui
 from Orange.widgets.utils import itemmodels
 from Orange.widgets.utils.owlearnerwidget import OWProvidesLearner
 from Orange.widgets.utils.sql import check_sql_input
+from Orange.canvas import report
 
 
 class OWUnivariateRegression(OWProvidesLearner, widget.OWWidget):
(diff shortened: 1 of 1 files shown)


janezd:
Report crashes if the user clicks the button when the widget has no data.

Report has to include the graph.

If you add the graph, the whole report can be just the graph and all the information can go into the caption. Thus, instead of everything reported now, just output the plot add the caption "Polynomial expansion: n" (where n is the actual number). The data size is not so important to report here. The names of x and y are already on the plot as the axis legend, and the other attributes are irrelevant. Name is also unimportant.

So: just the plot and the expansion as caption.

Orange/widgets/regression/owunivariateregression.py
```diff
@@ -81,6 +82,7 @@ def __init__(self):
         self.comboBoxAttributesY.setModel(self.y_var_model)
 
         gui.rubber(self.controlArea)
+        self.controlArea.layout().addWidget(self.report_button)
```
janezd: The button is already added to the layout. This line is needed only if you want to relocate it.

RachitKansal:
Ok, will make the necessary changes

RachitKansal:
@janezd  made the required changes. Please verify

Orange/widgets/regression/owunivariateregression.py
```diff
@@ -109,6 +111,16 @@ def __init__(self):
 
         self.apply()
 
+    def send_report(self):
+        if(self.data == None):
```
janezd: Remove parentheses.

Use `self.data is None` instead of `self.data == None`. (PEP 8: *Comparisons to singletons like None should always be done with is or is not, never the equality operators.*)

janezd:
Please squash the commits.

PR #1094, opened 2016-03-10
OWImpute: fix model based imputer. Fixes #1082.

### LINKED ISSUE #1082: Impute widget crashes on learner input
Impute widget crashes when provided with a learner input. Should automatically switch to 'Model-based imputer' and use it. Also the widget should check for model compatibility (classification or regression).

s-alexey:
Switch to Model-based imputer in case learner is provided.

Added learner adequacy checker.

### DIFF
--- a/Orange/preprocess/impute.py
+++ b/Orange/preprocess/impute.py
@@ -26,7 +26,43 @@ class ReplaceUnknowns(Transformation):
         return numpy.where(numpy.isnan(c), self.value, c)
 
 
-class Average:
+class BaseImputeMethod:
+    name = "Don't impute"
+    short_name = "leave"
+    description = ""
+    format = "{var.name} -> {short_name}"
+    columns_only = False
+    support_discrete = True
+    support_continuous = True
+
+    def __call__(self, data, variable):
+        return variable
+
+    def str(self, var):
+        return self.format.format(var=var, short_name=self.short_name)
+
+    def __str__(self):
+        return self.name
+
+    def copy(self):
+        return self
+
+
+class DropInstances(BaseImputeMethod):
+    name = "Remove instances with unknown values"
+    short_name = "drop"
+    description = ""
+
+    def __call__(self, data, variable):
+        index = data.domain.index(variable)
+        return numpy.isnan(data[:, index]).reshape(-1)
+
+
+class Average(BaseImputeMethod):
+    name = "Average/Most frequent"
+    short_name = "average"
+    description = "Replace with average/modus for the column"
+
     def __call__(self, data, variable, value=None):
         variable = data.domain[variable]
         if value is None:
@@ -53,15 +89,25 @@ class ImputeSql:
         return 'coalesce(%s, %s)' % (self.var.to_sql(), str(self.default))
 
 
-class Default:
+class Default(BaseImputeMethod):
+    name = "Value"
+    short_name = "value"
+    description = ""
+    columns_only = True
+
     def __init__(self, default=0):
         self.default = default
 
     def __call__(self, data, variable, *, default=None):
         variable = data.domain[variable]
         default = default if default is not None else self.default
-        return variable.copy(
-            compute_value=ReplaceUnknowns(variable, default))
+        return variable.copy(compute_value=ReplaceUnknowns(variable, default))
+
+    def copy(self):
+        return Default(self.default)
+
+    def str(self, var):
+        return self.format.format(var=var, short_name=self.default)
 
 
 class ReplaceUnknownsModel:
@@ -99,18 +145,46 @@ class ReplaceUnknownsModel:
         return column
 
 
-class Model:
-    def __init__(self, learner):
+class Model(BaseImputeMethod):
+    name = "Model-based imputer"
+    short_name = "model"
+    description = ""
+
+    def __init__(self, learner=None):
         self.learner = learner
 
-    def __call__(self, data, variable):
+    def __call__(self, data, variable, copy=False):
         variable = data.domain[variable]
         domain = domain_with_class_var(data.domain, variable)
         data = data.from_table(domain, data)
-        model = self.learner(data)
-        assert model.domain.class_var == variable
-        return variable.copy(
-            compute_value=ReplaceUnknownsModel(variable, model))
+
+        if self.learner.check_learner_adequacy(domain):
+            model = self.learner(data)
+            assert model.domain.class_var == variable
+            return variable.copy(
+                compute_value=ReplaceUnknownsModel(variable, model))
+        else:
+            return variable.copy()
+
+    def str(self, var):
+        return super().str(var) + " ({})".format(self.learner)
+
+    def copy(self):
+        return Model(self.learner)
+
+    @property
+    def support_discrete(self):
+        if self.learner is None:
+            return False
+        from Orange.regression.base_regression import LearnerRegression
+        return not isinstance(self.learner, LearnerRegression)
+
+    @property
+    def support_continuous(self):
+        if self.learner is None:
+            return False
+        from Orange.classification.base_classification import LearnerClassification
+        return not isinstance(self.learner, LearnerClassification)
 
 
 def domain_with_class_var(domain, class_var):
@@ -152,7 +226,13 @@ class Lookup(Lookup):
         return numpy.where(mask, unknown, values)
 
 
-class AsValue:
+class AsValue(BaseImputeMethod):
+    # name = "Value"
+    # short_name = "value"
+    name = "As a distinct value"
+    short_name = "new value"
+    description = ""
+
     def __call__(self, data, variable):
         variable = data.domain[variable]
         if variable.is_discrete:
@@ -231,7 +311,11 @@ class ReplaceUnknownsRandom(Transformation):
         return c
 
 
-class Random:
+class Random(BaseImputeMethod):
+    name = "Random values"
+    short_name = "random"
+    description = "Replace with a random value"
+
     def __call__(self, data, variable):
         variable = data.domain[variable]
         dist = distribution.get_distribution(data, variable)
--- a/Orange/preprocess/impute.py
+++ b/Orange/preprocess/impute.py
@@ -47,6 +47,9 @@ class BaseImputeMethod:
     def copy(self):
         return self
 
+    def __eq__(self, other):
+        return type(self) == type(other)
+
 
 class DropInstances(BaseImputeMethod):
     name = "Remove instances with unknown values"
@@ -156,15 +159,16 @@ class Model(BaseImputeMethod):
     def __call__(self, data, variable, copy=False):
         variable = data.domain[variable]
         domain = domain_with_class_var(data.domain, variable)
-        data = data.from_table(domain, data)
 
         if self.learner.check_learner_adequacy(domain):
+            data = data.from_table(domain, data)
             model = self.learner(data)
             assert model.domain.class_var == variable
             return variable.copy(
                 compute_value=ReplaceUnknownsModel(variable, model))
         else:
-            return variable.copy()
+            raise ValueError("`{}` doesn't support domain type"
+                             .format(self.learner))
 
     def str(self, var):
         return super().str(var) + " ({})".format(self.learner)
--- a/Orange/preprocess/impute.py
+++ b/Orange/preprocess/impute.py
@@ -2,9 +2,11 @@ import numpy
 
 import Orange.data
 from Orange.statistics import distribution, basic_stats
+from Orange.util import abstract
 from .transformation import Transformation, Lookup
 
-__all__ = ["ReplaceUnknowns", "Average"]
+__all__ = ["ReplaceUnknowns", "Average", "DoNotImpute", 'DropInstances',
+           "Model", "AsValue", "Random", "Default"]
 
 
 class ReplaceUnknowns(Transformation):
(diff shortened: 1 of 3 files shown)


Orange/preprocess/impute.py
```diff
@@ -107,11 +107,14 @@ def __call__(self, data, variable):
...
-        return variable.copy(
-            compute_value=ReplaceUnknownsModel(variable, model))
 
+        if self.learner.check_learner_adequacy(domain):
+            model = self.learner(data)
+            assert model.domain.class_var == variable
+            return variable.copy(
+                compute_value=ReplaceUnknownsModel(variable, model))
+        else:
+            return variable.copy()
```
astaric: When learner on input is not compatible with the variable, the variable does not get imputed.

I would rather move the check to the imputation widget. When the learner on input does not support variable type, a Model imputer with default_learner could be constructed.

astaric:
I like the fix as it is simple and straightforward. The problem is that the change on the Model imputer now allows the user to construct a model imputer for a variable of the wrong type, which would not impute anything. This is not desired.

I would rather move the check to the widget, and use some kind of default_learner for the variables with incompatible type (rather than not impute them).

If Model based imputer option was renamed to Model based imputer (<learner>), this would also signal to the user which learner will be used for imputation. When a given learner only supports one kind of variables, text could dynamically change to "Model base imputer (continuous: <learner1>, discrete: <learner2>" or something like this.

s-alexey:
@astaric I have changed my PR. 
It's not quite what you suggested but hope you will like it.

Orange/preprocess/impute.py
```diff
@@ -99,18 +145,46 @@ def __call__(self, data):
...
+
+    def str(self, var):
+        return super().str(var) + " ({})".format(self.learner)
+
+    def copy(self):
+        return Model(self.learner)
+
+    @property
+    def support_discrete(self):
```
astaric: What about the classes that support both (simple tree?)

Orange/preprocess/impute.py
```diff
@@ -99,18 +145,46 @@ def __call__(self, data):
...
-        return variable.copy(
-            compute_value=ReplaceUnknownsModel(variable, model))
+
+        if self.learner.check_learner_adequacy(domain):
+            model = self.learner(data)
+            assert model.domain.class_var == variable
+            return variable.copy(
+                compute_value=ReplaceUnknownsModel(variable, model))
+        else:
+            return variable.copy()
```
astaric: I still do not like this. If I call an instance of a Model imputer with data, I would like to have imputed data. If the learner does not support variable type and variable contains missing values, this should IMO raise an exception.

Or maybe model could accept multiple learners as a parameter, one for continuous and one for discrete?

astaric:
I generally like the new hierarchy, although I cannot decide whether it should have and "abstract" class on top instead of the DontImpute imputer.

BTW, have you considered having a kind of a registry (https://github.com/biolab/orange3/blob/master/Orange/util.py#L105) to collect all available imputers (instead of hardcoding them in the widget)?

Orange/widgets/data/owimpute.py
```diff
@@ -220,192 +104,163 @@ def __init__(self):
...
+
         self.unconditional_commit()
 
     def set_learner(self, learner):
         self.learner = learner
+        self.METHODS[self.MODEL_BASED_IMPUTER].learner = learner
+
+        button = self.default_button_group.button(self.MODEL_BASED_IMPUTER)
+        variable_button = self.variable_button_group.button(self.MODEL_BASED_IMPUTER)
+        if self.learner is None:
```
astaric: Currently, Model based imputer uses a default learner (simple tree), and the input is only used as a way to override the learner.

Orange/preprocess/impute.py
s-alexey: I don't like too, I have been meaning to replace it with an exception.

Orange/widgets/data/owimpute.py
s-alexey: I can add default learner or add a new `SimpleTree` impute method.

Orange/preprocess/impute.py
s-alexey: They are supported. 
It's the second time I have faced this problem (check learner type). Is there a good way to do it?

astaric: What about constructing a dummy domain and using check_learner_adequacy?

Orange/widgets/data/owimpute.py
astaric: I would prefer a default learner. This way, everything "just works", and if user wants to change the learner, he can.

Orange/widgets/data/owimpute.py
```diff
@@ -220,192 +113,169 @@ def __init__(self):
...
     def set_data(self, data):
         self.closeContext()
-        self.clear()
+        self.varmodel[:] = []
+        self.variable_methods = {}
+        self.modified = False
         self.data = data
+
         if data is not None:
             self.varmodel[:] = data.domain.variables
```
s-alexey: Is there a good way to filter variables without missed values?

Orange/preprocess/impute.py
```diff
@@ -26,7 +27,61 @@ def transform(self, c):
...
 
-class Average:
+class BaseImputeMethod:
+    name = ""
+    short_name = ""
+    description = ""
+    format = "{var.name} -> {short_name}"
+    columns_only = False
+
+    def __init__(self, *args, **kwargs):
```
astaric: If __init__ does nothing, you can ommit it.

Orange/preprocess/impute.py
```diff
@@ -26,7 +27,61 @@ def transform(self, c):
...
+        pass
+
+    @abstract
+    def __call__(self, data, variable):
+        raise NotImplementedError()
```
astaric: @abstract decorator will raise NotImplementedError. The body of the method could include a helpful docstring instead (describing what is the purpose of the method and its parameters).

Orange/preprocess/impute.py
```diff
@@ -26,7 +27,61 @@ def transform(self, c):
...
+
+    def str(self, var):
+        return self.format.format(var=var, short_name=self.short_name)
```
astaric: This method could use a docstring describing what it is supposed to return and how does that differ from __str__. Depending on the purpose, a more descriptive method name could be used.

Orange/preprocess/impute.py
```diff
@@ -26,7 +27,61 @@ def transform(self, c):
...
+
+    def __str__(self):
+        return self.name
+
+    def copy(self):
+        return self
+
+    def __eq__(self, other):
+        return type(self) == type(other)
```
astaric: Could use a docstring/comment on why is this a good idea.

Orange/preprocess/impute.py
```diff
@@ -152,7 +240,13 @@ def transform(self, column):
         return numpy.where(mask, unknown, values)
 
 
-class AsValue:
+class AsValue(BaseImputeMethod):
+    # name = "Value"
```
astaric: This comments can be removed.

Orange/widgets/data/owimpute.py
```diff
@@ -1,169 +1,53 @@
...
-def radio_button(text="", checked=False, group=None, group_id=None):
-    button = QRadioButton(text, checked=checked)
-    if group is not None:
-        group.addButton(button, )
-        if group_id is not None:
-            group.setId(button, group_id)
-    return button
+    method = impute.DoNotImpute()
 
+    # def str(self, var):
```
astaric: Commented lines can be removed.

Orange/widgets/data/owimpute.py
astaric: Not that I know of.

astaric:
The titles for sections Default method and Individual attribute settings are not displayed correctly on OSX.

screen shot 2016-05-09 at 09 33 28

Orange/widgets/data/owimpute.py
```diff
@@ -220,192 +113,169 @@ def __init__(self):
...
-        else:
-            data = None
+
+            self.warning(1)
+            with self.progressBar(len(self.varmodel)) as progress:
+                for i, var in enumerate(self.varmodel):
+                    method = self.variable_methods.get(i, self.default_method)
+
+                    if not method.check_supports_variable(var):
+                        self.warning(1, "Default method has ignored some variables.")
```
astaric: Default method could not impute some of the variables?

astaric:
I have made some comments in the code. The only "major" issue I have found is the layout issue above. The rests seems ready for merge to me.

s-alexey:
@astaric I have made some changes, can you review it and check the layout issue?

astaric:
The layout issue remains. @kernc, can you check if it looks ok on your computer?

kernc:
@astaric The layout seems ok here.
screenshot - 05102016 - 01 43 04 pm

Orange/preprocess/impute.py
```diff
@@ -26,7 +28,65 @@ def transform(self, c):
         return numpy.where(numpy.isnan(c), self.value, c)
 
 
-class Average:
+class BaseImputeMethod:
+    name = ""
+    short_name = ""
+    description = ""
+    format = "{var.name} -> {short_name}"
```
kernc: How about specifying this as `"{var.name} → {self.short_name}"`. Then `format_variable()` method below could call format it with `self.format.format(var=var, self=self)` and there would be no reason to override the method if `format` is already overriden?

Orange/preprocess/impute.py
```diff
@@ -26,7 +28,65 @@ def transform(self, c):
...
+        return self.format.format(var=var, short_name=self.short_name)
+
+    def __str__(self):
+        return self.name
+
+    def copy(self):
+        return self
+
+    @classmethod
+    def check_supports_variable(cls, variable):
```
kernc: `supports_variable()` is shorter and grammatically clearer; the method either supports variable or not supports variable.

Orange/preprocess/impute.py
```diff
@@ -26,7 +28,65 @@ def transform(self, c):
...
+
+    def __call__(self, data, variable):
+        index = data.domain.index(variable)
+        return numpy.isnan(data[:, index]).reshape(-1)
+
+
+class Average(BaseImputeMethod):
+    name = "Average/Most frequent"
+    short_name = "average"
+    description = "Replace with average/modus for the column"
```
kernc: Modus is [mode](https://en.wikipedia.org/wiki/Mode_%28statistics%29) in English.

Orange/widgets/data/owimpute.py
```diff
@@ -176,38 +57,47 @@ class OWImpute(OWWidget):
...
+        box = QtGui.QGroupBox(title=self.tr("Default method"), flat=False)
+        main_layout.addWidget(box)
+
+        button_group = QtGui.QButtonGroup()
+        for i, method in enumerate(self.METHODS):
+            if not method.columns_only:
+                button = QtGui.QRadioButton(method.name)
+                button.setChecked(i == self.default_method_index)
+                button_group.addButton(button, i)
+                main_layout.layout().addWidget(button)
```
kernc: `box.layout().addWidget(button)`? That's what was before.

Orange/widgets/data/owimpute.py
```diff
@@ -220,192 +110,170 @@ def __init__(self):
...
-        self.value_stack = value_stack = QStackedLayout()
-        self.value_combo = QComboBox(
+        self.value_combo = QtGui.QComboBox(
             minimumContentsLength=8,
-            sizeAdjustPolicy=QComboBox.AdjustToMinimumContentsLength,
+            sizeAdjustPolicy=QtGui.QComboBox.AdjustToMinimumContentsLength,
             activated=self._on_value_changed)
-        self.value_line = QLineEdit(editingFinished=self._on_value_changed)
-        self.value_line.setValidator(QDoubleValidator())
+        self.value_double = QtGui.QDoubleSpinBox(
```
kernc: Does this value get saved as a Setting?

Orange/widgets/data/owimpute.py
```diff
@@ -220,192 +110,170 @@ def __init__(self):
...
+            button = QtGui.QRadioButton(text=method.name)
+            button_group.addButton(button, i)
+            method_layout.addWidget(button)
 
-        self.value_stack = value_stack = QStackedLayout()
-        self.value_combo = QComboBox(
+        self.value_combo = QtGui.QComboBox(
             minimumContentsLength=8,
-            sizeAdjustPolicy=QComboBox.AdjustToMinimumContentsLength,
+            sizeAdjustPolicy=QtGui.QComboBox.AdjustToMinimumContentsLength,
```
kernc: Forward compatible, it's better to import and refer to `QObjects` directly. In Qt 5, for example, `QComboBox` is in `QtWidgets` namespace (module).

Orange/widgets/data/owimpute.py
```diff
@@ -220,192 +110,170 @@ def __init__(self):
...
-        assert self.METHODS[-1].short == "value"
+        button_group = QtGui.QButtonGroup()
+        for i, method in enumerate(self.METHODS):
+            button = QtGui.QRadioButton(text=method.name)
+            button_group.addButton(button, i)
+            method_layout.addWidget(button)
 
-        self.value_stack = value_stack = QStackedLayout()
-        self.value_combo = QComboBox(
+        self.value_combo = QtGui.QComboBox(
```
kernc: The selected value here should be saved as a Setting as well.

astaric:
The box surrounding default methods/individual attribute settings is now (again) shown on OSX, so the layout issue is gone.

Orange/preprocess/impute.py
```diff
@@ -26,7 +28,65 @@ def transform(self, c):
...
+
+    def __call__(self, data, variable):
+        index = data.domain.index(variable)
+        return numpy.isnan(data[:, index]).reshape(-1)
+
+
+class Average(BaseImputeMethod):
+    name = "Average/Most frequent"
+    short_name = "average"
+    description = "Replace with average/mode for the column"
```
kernc: average value / mode _of_

astaric:
If @kernc has no more comments, I am going to merge this today. If you were planning on changing anything, please let me know.

Orange/widgets/data/owimpute.py
s-alexey: It wasn't saved as a Settings previously and I'm not sure we should save it now. It's an index and actual value depends on selected variable.

s-alexey:
I'm not planning any changes.

PR #1095, opened 2016-03-10
fixes: distance transformation report issue

iamved:


iamved:
fixed issue report for distance transformation

janezd:
Fixed in #1113.

PR #1096, opened 2016-03-10
Scatter Plot Warnings fixed

wadhwasahil:

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -21,6 +21,7 @@ from Orange.widgets.settings import \
 from Orange.widgets.utils.toolbar import ZoomSelectToolbar
 from Orange.widgets.visualize.owscatterplotgraph import OWScatterPlotGraph
 from Orange.widgets.widget import OWWidget, Default, AttributeList
+from Orange.statistics.basic_stats import BasicStats
 
 
 def font_resize(font, factor, minsize=None, maxsize=None):
@@ -60,6 +61,7 @@ class OWScatterPlot(OWWidget):
     attr_y = ContextSetting("")
 
     graph = SettingProvider(OWScatterPlotGraph)
+    zoom_select_toolbar = SettingProvider(ZoomSelectToolbar)
 
     jitter_sizes = [0, 0.1, 0.5, 1, 2, 3, 4, 5, 7, 10]
 
@@ -76,13 +78,15 @@ class OWScatterPlot(OWWidget):
         axispen = QtGui.QPen(self.palette().color(QtGui.QPalette.Text))
         axis = plot.getAxis("bottom")
         axis.setPen(axispen)
-
         axis = plot.getAxis("left")
         axis.setPen(axispen)
 
+        self.attr_nans = dict()
+
         self.data = None  # Orange.data.Table
         self.subset_data = None  # Orange.data.Table
-        self.data_metas_X = None  # self.data, where primitive metas are moved to X
+        # self.data, where primitive metas are moved to X
+        self.data_metas_X = None
         self.sql_data = None  # Orange.data.sql.table.SqlTable
         self.attribute_selection_list = None  # list of Orange.data.Variable
         self.__timer = QTimer(self, interval=1200)
@@ -132,11 +136,11 @@ class OWScatterPlot(OWWidget):
             **common_options)
         self.cb_attr_shape = gui.comboBox(
             box, self, "graph.attr_shape", label="Shape:",
-            emptyString="(Same shape)", callback=self.graph.update_shapes,
+            emptyString="(Same shape)", callback=self.update_shapes,
             **common_options)
         self.cb_attr_size = gui.comboBox(
             box, self, "graph.attr_size", label="Size:",
-            emptyString="(Same size)", callback=self.graph.update_sizes,
+            emptyString="(Same size)", callback=self.update_sizes,
             **common_options)
 
         g = self.graph.gui
@@ -158,7 +162,8 @@ class OWScatterPlot(OWWidget):
         buttons = self.zoom_select_toolbar.buttons
         buttons[g.Zoom].clicked.connect(self.graph.zoom_button_clicked)
         buttons[g.Pan].clicked.connect(self.graph.pan_button_clicked)
-        buttons[g.SimpleSelect].clicked.connect(self.graph.select_button_clicked)
+        buttons[g.SimpleSelect].clicked.connect(
+            self.graph.select_button_clicked)
         buttons[g.ZoomReset].clicked.connect(self.graph.reset_button_clicked)
         self.controlArea.layout().addStretch(100)
         self.icons = gui.attributeIconDict
@@ -218,6 +223,7 @@ class OWScatterPlot(OWWidget):
     def set_data(self, data):
         self.information(1)
         self.__timer.stop()
+        self.attr_nans.clear()
         self.sampling.setVisible(False)
         self.sql_data = None
         if isinstance(data, SqlTable):
@@ -284,7 +290,8 @@ class OWScatterPlot(OWWidget):
             if subset_data.approx_len() < AUTO_DL_LIMIT:
                 subset_data = Table(subset_data)
             else:
-                self.warning(0, "Data subset does not support large Sql tables")
+                self.warning(
+                    0, "Data subset does not support large Sql tables")
                 subset_data = None
         self.subset_data = self.move_primitive_metas_to_X(subset_data)
 
@@ -368,7 +375,27 @@ class OWScatterPlot(OWWidget):
         self.cb_class_density.setEnabled(self.graph.can_draw_density())
         self.send_features()
 
+    def update_shapes(self):
+        if (self.cb_attr_shape.count() > 1 and
+            self.combo_value(self.cb_attr_shape) == None):
+                self.warning("Please select a shape attribute")
+        else:
+            self.warning()
+        self.graph.update_shapes()
+
+    def update_sizes(self):
+        if (self.cb_attr_size.count() > 1 and
+            self.combo_value(self.cb_attr_size) == None):
+                self.warning("Please select a size attribute")
+        else:
+            self.warning()
+        self.graph.update_sizes()
+
     def update_colors(self):
+        if self.combo_value(self.cb_attr_color) == None:
+            self.warning("Please select a color attribute")
+        else:
+            self.warning()
         self.graph.update_colors()
         self.cb_class_density.setEnabled(self.graph.can_draw_density())
 
@@ -379,6 +406,54 @@ class OWScatterPlot(OWWidget):
         self.graph.zoomStack = []
         if attributes and len(attributes) == 2:
             self.attr_x, self.attr_y = attributes
+        data_x = self.data[:, self.attr_x]
+        data_y = self.data[:, self.attr_y]
+        if self.attr_x not in self.attr_nans:
+            self.attr_nans[self.attr_x] = BasicStats(data_x, self.attr_x).nans
+        if self.attr_y not in self.attr_nans:
+            self.attr_nans[self.attr_y] = BasicStats(data_y, self.attr_y).nans
+        self.information()
+        self.warning()
+        flag_x = 0
+        flag_y = 0
+        if self.attr_nans[self.attr_x]:
+            flag_x = 1
+            if self.attr_nans[self.attr_x] == len(data_x):
+                flag_x = 2
+        if self.attr_nans[self.attr_y]:
+            flag_y = 1
+            if self.attr_nans[self.attr_y] == len(data_y):
+                flag_y = 2
+        if flag_x == 2:
+            if flag_y == 2:
+                self.warning("'{}' and '{}' have no defined values".format(
+                    self.attr_x, self.attr_y))
+            else:
+                self.warning("'{}' has no defined values".format(
+                    self.attr_x))
+        else:
+            if flag_y == 2:
+                self.warning("'{}' has no defined values".format(
+                    self.attr_y))
+        if flag_x == 1:
+            if flag_y == 1:
+                self.information("'{}' and '{}' have missing values".format(
+                    self.attr_x, self.attr_y))
+            else:
+                self.information("'{}' has missing values".format(
+                    self.attr_x))
+        else:
+            if flag_y == 1:
+                self.information("'{}' has missing values".format(
+                    self.attr_y))
+            if self.combo_value(self.cb_attr_color) == None:
+                    self.warning("Please select a color attribute")
+            if (self.cb_attr_size.count() > 1 and
+                self.combo_value(self.cb_attr_size) == None):
+                    self.warning("Please select a size attribute")
+            if (self.cb_attr_shape.count() > 1 and
+                self.combo_value(self.cb_attr_shape) == None):
+                    self.warning("Please select a shape attribute")
         if not self.graph.have_data:
             return
         self.graph.update_data(self.attr_x, self.attr_y, reset_view)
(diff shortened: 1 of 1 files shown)


wadhwasahil:
fixes #1048 .

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,8 +380,14 @@ def update_graph(self, attributes=None, reset_view=True, **_):
         self.graph.zoomStack = []
         if attributes and len(attributes) == 2:
             self.attr_x, self.attr_y = attributes
+        X = BasicStats(self.data[:, self.attr_x], self.attr_x)
+        Y = BasicStats(self.data[:, self.attr_y], self.attr_y)
```
kernc: The [discussion from #1070](https://github.com/biolab/orange3/pull/1070#discussion_r55517993) is still relevant, though. You can optimize this by computing basic stats only once, that is, when the data is connected.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,8 +380,14 @@ def update_graph(self, attributes=None, reset_view=True, **_):
...
+        if X.nans or Y.nans:
+            self.warning("Some attributes contain missing values") 
+        else:
+             self.warning()      
         if not self.graph.have_data:
-            return
+            return
```
kernc: You should set your code editor to [strip trailing whitespace](https://www.google.com/search?q=strip+trailing+whitespace&tbm=isch) characters, or, at least, not introduce new ones.

Orange/widgets/visualize/owscatterplot.py
wadhwasahil: @kernc, how to check data is connected?

kernc: I think [`set_data` is the callback](https://github.com/biolab/orange3/pull/1096/files#diff-0f1cb6d8751836f98c5293bec4004169R45) called when something new is sent on the data signal.

wadhwasahil: I have 1 idea:
Create an array of size n(n=no of attributes) where arr[i] = 1 if ith attribute contains NaN values else arr[i] = 0.
But this will take `O(n*m)` m = no of tuples.
This will definitely be optimal solution. What do you think?

kernc: Actually, looking at `BasicStats` implementation, I think you might be right. In that case, code as-is is ok, I guess.
@lanzagar what do you think?

wadhwasahil:
@kernc Is there any issue left with the PR?

kernc:
@lanzagar, what do you think?

wadhwasahil:
@kernc I think what I did with `BasicStats` is fine. This won't take much time for processing.
What do you think about it?

kernc:
I don't have a qualified opinion. Perhaps @lanzagar or @astaric do. There's also `DomainBasicStats` IIRC; can you perhaps use that once in advance?

wadhwasahil:
@kernc Waiting for @lanzagar for his reviews

wadhwasahil:
@kernc Changed the logic. Now BasicStats() is only called once for each dataset. Have a look now.
I think the problem is solved here.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -59,6 +60,8 @@ class OWScatterPlot(OWWidget):
     attr_x = ContextSetting("")
     attr_y = ContextSetting("")
 
+    is_invalid = dict()
```
kernc: This could be named `is_all_nan` and be made a `set()` instead.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -219,6 +224,11 @@ def reset_graph_data(self, *_):
     def set_data(self, data):
         self.information(1)
         self.__timer.stop()
+        for attr in data.domain:
+            if BasicStats(data[:, attr], attr).nans:
```
kernc: I don't know about the efficiency of this, but the condition tests if the attribute contains _any_ nans. The [original issue](https://github.com/biolab/orange3/issues/1048) is about columns of _all_ nans.

wadhwasahil: @kernc : but the warnings should be added for attributes which contain even 1 nan. I don't think it is wrong.

lanzagar:
1) I like that you fixed long lines, indentation, etc.
But please, make code improvements in a separate commit, so it is easier to review relevant changes.

2) If you compute BasicStats for all variables, I think one call to DomainBasicStats should definitely be preferred to a for loop (for better code, not efficiency reasons).
BUT, since you introduced caching of those results, an even better approach might be to not compute everything in advance, but to do it just the first time when needed.
i.e. when attributes are selected, you check in your dict to see if those attributes are already there.
If they are not, you compute BasicStats and add them to the dict, otherwise use the already computed results.
This would avoid computations for thousands of attributes at the beginning, in case such data is used.

3) I think you should handle both cases - some missing values and all missing values.
You could change the dict to store the number of nans instead of just 0/1 (also rename to something like `attr_nans`) and use one warning when the number is `==len(data)` and another if not, but is `>0`.
To be more precise - when an attribute is selected that has all nans, use warning with an appropriate message and do not even continue with plotting.
If only some are nan, show an information() instead and plot the data.

I think @janezd, who started Issue #1048, agreed that 2 types of warnings/information are best.

janezd:
Yes. Warning and information.

wadhwasahil:
Should i use dictionary or set?
And what info() should be called?

kernc:
Use attr→nancount dict. If some vals are nans, show `self.information('...')`.

wadhwasahil:
@lanzagar @kernc If the no of `nans = len(data)`, then stop the plotting of graph. How can I prevent the graph from being plotted?

janezd:
@lanzagar, haven't we said that if an attribute has just nans, we don't show it in the combo and show a warning?

lanzagar:
@janezd that was the second idea (that @ajdapretnar suggested) to not show them at all. And then you said that in that case the user would not know why they are not there and it would be better to have them in the combo box, but when selected show a warning. I like this option.
The new idea, if I understand it correctly, would be to not have them in the dropdown, but always show a warning that they were removed (and why). I like it a bit less, since then you can't get rid of the warning even if you are just looking at visualizations using valid attributes.

janezd:
You're right. Show all; warning if no data is shown and information if only some data is hidden. Right?

wadhwasahil:
What should I do now?

wadhwasahil:
and what warning and information should I write?

kernc:
Just make up something sensible and then revise. Software programming is an iterative process.

wadhwasahil:
@kernc If all nan's are present in an attribute then the graph has to be shown or not?

kernc:
Le me answer with a question. Say you have:
```py
x = np.random.random(10)
y = np.tile(np.nan, 10)
```
How do you draw a scatter plot of that?

wadhwasahil:
can't be drawn. However isn't it already handled by Orange. If some attribute is empty , so plot can't be drawn.

wadhwasahil:
@kernc @lanzagar @janezd have a look now. Waiting for replies.

wadhwasahil:
@kernc @janezd @lanzagar Will the PR's merge today?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,6 +386,20 @@ def update_graph(self, attributes=None, reset_view=True, **_):
...
+        if self.attr_y not in self.attr_nans:
+            self.attr_nans[self.attr_y] = BasicStats(data_y, self.attr_y).nans
+        if (self.attr_nans[self.attr_x] == len(data_x) or
+                self.attr_nans[self.attr_y] == len(data_y)):
+            self.warning("Some attributes are empty")
+        elif (self.attr_nans[self.attr_x] or self.attr_nans[self.attr_y]):
+            self.information("Some attributes contain missing values")
+        else:
+            self.information()
+            self.warning()
```
kernc: If the user selected an attribute that is all-nan, a warning is posted. If the user then selects an attribute that is some-nan, the information is posted, but the warning bar still remains (isn't cleared).

Please revise.

wadhwasahil: wont `self.warning()` remove the warning bar? Also, which dataset has an attribute with all nans?

wadhwasahil:
@janend : please have a look here also.

Orange/widgets/visualize/owscatterplot.py
kernc: It will, but only when the `else` branch is entered, which it isn't if some of the attributes are nan (where the second branch is). You can construct an example dataset if you need it.

wadhwasahil:
@janezd @kernc Final changes done. Waiting for your replies.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,6 +386,22 @@ def update_graph(self, attributes=None, reset_view=True, **_):
...
             self.attr_x, self.attr_y = attributes
+        data_x = self.data[:, self.attr_x]
+        data_y = self.data[:, self.attr_y]
+        if self.attr_x not in self.attr_nans:
+            self.attr_nans[self.attr_x] = BasicStats(data_x, self.attr_x).nans
+        if self.attr_y not in self.attr_nans:
+            self.attr_nans[self.attr_y] = BasicStats(data_y, self.attr_y).nans
+        if (self.attr_nans[self.attr_x] == len(data_x) or
+                self.attr_nans[self.attr_y] == len(data_y)):
+        	self.information()
```
kernc: Yes, this could work (although not indented correctly). You could also clear both, information and warning, unconditionally above this `if` block. Improves readability (imho) and saves three lines. :smiley:

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -431,14 +453,14 @@ def send_report(self):
         if self.data:
             domain = self.data.domain
             disc_attr = domain[self.attr_x].is_discrete or \
-                        domain[self.attr_y].is_discrete
+                domain[self.attr_y].is_discrete
```
kernc: Don't you like rows neatly aligned? :smiley: 
I'd prefer:
```py
disc_attr = (domain[self.attr_x].is_discrete or
             domain[self.attr_y].is_discrete)
```

wadhwasahil:
@kernc Done.

wadhwasahil:
@kernc @janezd Done.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -59,6 +60,8 @@ class OWScatterPlot(OWWidget):
     attr_x = ContextSetting("")
     attr_y = ContextSetting("")
 
+    attr_nans = dict()
```
kernc: This should probably be set in the constructor and re-set in `set_data` method (even if some data uses the same domain as another, it might carry a different payload).

wadhwasahil: first part is clear.
How do I re-set? Should I clear the dict() ?

kernc: Yes, that's what I meant. Sorry.

wadhwasahil: @kernc done. ready for Merge  I guess.

kernc:
@lanzagar What do you think? :smiley:

wadhwasahil:
@kernc As per @lanzagar and @janezd directions :smile:

wadhwasahil:
@janezd Waiting for your reply.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,6 +387,19 @@ def update_graph(self, attributes=None, reset_view=True, **_):
...
+        data_y = self.data[:, self.attr_y]
+        if self.attr_x not in self.attr_nans:
+            self.attr_nans[self.attr_x] = BasicStats(data_x, self.attr_x).nans
+        if self.attr_y not in self.attr_nans:
+            self.attr_nans[self.attr_y] = BasicStats(data_y, self.attr_y).nans
+        self.information()
+        self.warning()
+        if (self.attr_nans[self.attr_x] == len(data_x) or
+            self.attr_nans[self.attr_y] == len(data_y)):
+            self.warning("Some attributes are empty")
```
kernc: Could you tell the user exactly which attribute is empty? And perhaps say not empty but "has no defined values" instead.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,6 +387,19 @@ def update_graph(self, attributes=None, reset_view=True, **_):
...
+        elif (self.attr_nans[self.attr_x] or self.attr_nans[self.attr_y]):
+            self.information("Some attributes contain missing values")
```
kernc: And similar here, _which_ attribute contains missing values.

Orange/widgets/visualize/owscatterplot.py
wadhwasahil: Yeah but that would include many if-else statements for finding which attribute is causing the warning/information.
I think we should keep it `Some attributes have no defined values`. What do you think?

kernc: It can be only `attr_x` or `attr_y` (or both) that is empty. That's not a lot of if statements?

wadhwasahil:
@kernc Have a look now.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -380,6 +387,46 @@ def update_graph(self, attributes=None, reset_view=True, **_):
...
+            if flag_y == 1:
+                self.information("{} and {} have missing values".format(
+                    self.attr_x, self.attr_y))
+            else:
+                self.information("{} has missing values".format(
+                    self.attr_x))
+        else:
+            if flag_y == 1:
+                self.information("{} has missing values".format(
+                    self.attr_y))
```
kernc: That would work. But what about shorter:
```py
for attr, data in ((self.attr_x, data_x),
                   (self.attr_y, data_y)):
    if self.attr_nans[attr] == len(data):
        self.warning(id(attr), "Attribute '{}' has no defined values".format(attr))
    elif self.attr_nans[attr]:
        self.information(id(attr), "Attribute '{}' contains some missing values".format(attr))
```

wadhwasahil: `2 problems during execution` is showing for 2 incomplete attributes.

kernc: It shows the same with your code if one attribute has _all_ missing values (one warning) and some attribute has _some_ missing values (one information). If you hover over it, it shows more details in the tooltip. I don't know, I like the latter code better. @lanzagar, what do you think?

wadhwasahil:
@kernc @janezd what do you think about new commit?

wadhwasahil:
@janezd Would you mind having a look?

janezd:
We discussed it today at the meeting: warnings should indeed be issued also if the attribute for the color, shape or size is missing.

wadhwasahil:
This PR however solves the initial problem. Can't we have another PR for color, shape or size @janezd ?

kernc:
I, too, agree it would be better to have it in this PR. It's all about _the same exact issue_, and it can be pretty easily generalized: create a separate function that checks an attribute. If the attribute has nans, warn. Call this function on change of _any_ combo box.

Easy-peasy. :smile:

wadhwasahil:
@kernc should I include a function in scatterplot or generalize it for every widget available?

kernc:
As you see fit.

wadhwasahil:
@janezd @kernc The attribute for size is fixed (Same size), isn't it? Or does it change anytime?
So, I am including the checks in update graph itself. Is it ok?

kernc:
Sizing the markers only works with continuous variables. Regarding the label, I think it's fine if no label is shown when there is nan. But the user should probably still be warned.

wadhwasahil:
@kernc : I am doing this way
1. For continuous variables: check for nans, color, label and size.
2. For discrete variables: check for nans, color, label and shape.
is it ok?

kernc:
Sounds reasonable.

wadhwasahil:
@kernc @janezd have a look and suggest changes .

wadhwasahil:
@kernc @janezd @lanzagar I have updated the changes. Please have a look.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -368,7 +375,27 @@ def update_attr(self, attributes=None):
...
+            self.combo_value(self.cb_attr_shape) == None):
+                self.warning("Please select a shape attribute")
+        else:
+            self.warning()
+        self.graph.update_shapes()
+
+    def update_sizes(self):
+        if (self.cb_attr_size.count() > 1 and
+            self.combo_value(self.cb_attr_size) == None):
+                self.warning("Please select a size attribute")
```
kernc: This segment looks awfully duplicated (three times). Could we put this check in a single place?

kernc:
@wadhwasahil Thanks for working on this. There appears to have been a slight misunderstanding. The user isn't required to select an attribute for shape/color/size/... Only is the user does select an attribute for any of those things, and _if the values for that attribute are some or all nan_, then the appropriate warning (like for the X, Y attributes) should be raised.

wadhwasahil:
I should have saved the previous commit :cry: However, I will change it soon and update the commit. :+1:

astaric:
Thank you for your contribution. This issues has meanwhile been resolved so I am closing this PR as duplicate.

Issue #1077, opened 2016-03-09
Report for Confusion Matrix


ajdapretnar:
Report for Confusion Matrix widget looks strange. Implement gridlines, color correct and misclassified instances accordingly (as in the widget), variable names and sums in bold.

smkuls:
Is this fine? [owconfusionmatrix.py ](https://github.com/biolab/orange3/blob/master/Orange/widgets/evaluate/owconfusionmatrix.py) is internally calling report_table which is used by other widgets too (such as Test Learners). If I modify it to add colors, it will affect the others (right now I just added an optional border parameter to the call). I can customize it individually for Confusion Matrix but I don't know if that will be a good idea.

![image](https://cloud.githubusercontent.com/assets/5889240/13665553/1ac872a4-e6d3-11e5-93ee-f9acd17d73d7.png)

wadhwasahil:
how did u make the test bold? Can you provide the line of change?

smkuls:
I updated the report_list function in [report.py](https://github.com/biolab/orange3/blob/master/Orange/canvas/report/report.py)

` def report_list(data,
                        header_rows=header_rows, header_columns=header_columns,border=border):
            if border:
                cells = ["<td style=\"border : 1px solid black\">{}</td>", "<th>{}</th>"]
            else:
                cells = ["<td>{}</td>", "<th>{}</th>"]
            return join("  <tr>\n    {}</tr>\n".format(
                join(cells[rowi < header_rows or coli < header_columns]
                     .format(fmtnum(elm)) for coli, elm in enumerate(row))
            ) for rowi, row in enumerate(data))
`

wadhwasahil:
@smkuls `self.report_table(
                "Confusion matrix for {} (showing {})".
                format(self.learners[index],
                       self.quantities[self.selected_quantity].lower()),
                self.tablemodel, header_rows = 2, header_columns =2)`
Is this you mean?

wadhwasahil:
Some One please check my PR and suggest changes. 2 issues are being solved in the above PR of mine. 
Thanks in advance.

smkuls:
@wadhwasahil  If you ask somebody for a fix then you don't include that in "your" pull request. Have some courtesy.

wadhwasahil:
@smkuls . I think we are here to help each other. If you want I will change the (2,2) line and do something different. I don't know what offends you. After all we are a community and here to share our thoughts and ideas. 
[...]

PR #1097, opened 2016-03-10
Report for Confusion Matrix Fixed

wadhwasahil:

### DIFF
--- a/Orange/canvas/report/report.py
+++ b/Orange/canvas/report/report.py
@@ -162,8 +162,7 @@ class Report:
             self.report_html += get_html_img(plot.scene())
 
     # noinspection PyBroadException
-    def report_table(self, name, table=None, header_rows=0, header_columns=0,
-                     num_format=None):
+    def report_table(self, name, table=None, header_rows=0, header_columns=0, tot_rows = 0, num_format=None):
         """
         Add content of a table to the report.
 
@@ -230,12 +229,11 @@ class Report:
         else:
             def fmtnum(s):
                 return s
-
         def report_list(data,
-                        header_rows=header_rows, header_columns=header_columns):
+                        header_rows=header_rows, header_columns=header_columns, tot_rows = tot_rows):
             cells = ["<td>{}</td>", "<th>{}</th>"]
             return join("  <tr>\n    {}</tr>\n".format(
-                join(cells[rowi < header_rows or coli < header_columns]
+                join(cells[rowi < header_rows or coli < header_columns or rowi == 3 + tot_rows]
                      .format(fmtnum(elm)) for coli, elm in enumerate(row))
             ) for rowi, row in enumerate(data))
 
@@ -256,7 +254,7 @@ class Report:
         else:
             body = None
         if body:
-            self.report_html += "<table>\n" + body + "</table>"
+            self.report_html += "<table border = '1px' style = 'border-collapse:collapse'>\n" + body + "</table>"
 
     # noinspection PyBroadException
     def report_list(self, name, data=None, limit=1000):
--- a/Orange/canvas/report/report.py
+++ b/Orange/canvas/report/report.py
@@ -10,6 +10,7 @@ from Orange.widgets.utils import getdeepattr
 
 
 class Report:
+
     """
     A class that adds report-related methods to the widget.
     """
@@ -162,7 +163,7 @@ class Report:
             self.report_html += get_html_img(plot.scene())
 
     # noinspection PyBroadException
-    def report_table(self, name, table=None, header_rows=0, header_columns=0, tot_rows = 0, num_format=None):
+    def report_table(self, name, table=None, header_rows=0, header_columns=0, confusion_rows=0, num_format=None):
         """
         Add content of a table to the report.
 
@@ -229,11 +230,12 @@ class Report:
         else:
             def fmtnum(s):
                 return s
+
         def report_list(data,
-                        header_rows=header_rows, header_columns=header_columns, tot_rows = tot_rows):
+                        header_rows=header_rows, header_columns=header_columns, confusion_rows=confusion_rows):
             cells = ["<td>{}</td>", "<th>{}</th>"]
             return join("  <tr>\n    {}</tr>\n".format(
-                join(cells[rowi < header_rows or coli < header_columns or rowi == 3 + tot_rows]
+                join(cells[rowi < header_rows or coli < header_columns or rowi == 3 + confusion_rows]
                      .format(fmtnum(elm)) for coli, elm in enumerate(row))
             ) for rowi, row in enumerate(data))
 
@@ -254,7 +256,8 @@ class Report:
         else:
             body = None
         if body:
-            self.report_html += "<table border = '1px' style = 'border-collapse:collapse'>\n" + body + "</table>"
+            self.report_html += "<table border = '1px' style = 'border-collapse:collapse'>\n" + \
+                body + "</table>"
 
     # noinspection PyBroadException
     def report_list(self, name, data=None, limit=1000):
@@ -629,9 +632,11 @@ def describe_data_brief(data):
     items.update(describe_domain_brief(data.domain))
     return items
 
+
 def colored_square(r, g, b):
     return '<span class="legend-square" ' \
            'style="background-color: rgb({}, {}, {})"></span>'.format(r, g, b)
+
 
 def list_legend(model, selected=None):
     """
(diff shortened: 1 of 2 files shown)


wadhwasahil:
Fixes #1077

kernc:
This PR includes code meant for #1096. You probably did #1096 first, and then branched this branch off of that one, e.g.
```bash
git checkout -b scatter_plot
...  # make changes for #1096
git add -p && git commit
git checkout -b confusion
... # confusion branch now includes the scatter_plot branch commit
```
Instead, you should probably switch to master branch before `checkout -b` to another (confusion) branch.

You can _update_ this branch by:
```bash
git checkout confusion
git checkout HEAD^ Orange/widgets/visualize/owscatterplot.py  # to revert the irrelevant part
git add Orange/widgets/visualize/owscatterplot.py
git commit --amend  # to add the revert to the head commit
git push --force  # to overwrite this PR, instead of opening a new one
```

wadhwasahil:
@kernc  Done.

Orange/canvas/report/report.py
```diff
@@ -162,8 +162,7 @@ def report_plot(self, name=None, plot=None):
             self.report_html += get_html_img(plot.scene())
 
     # noinspection PyBroadException
-    def report_table(self, name, table=None, header_rows=0, header_columns=0,
-                     num_format=None):
+    def report_table(self, name, table=None, header_rows=0, header_columns=0, tot_rows = 0, num_format=None):
```
kernc: 1. [PEP-8](https://www.python.org/dev/peps/pep-0008/) states the correct way is to specify function parameter defaults as `key=value`, that is, without whitespace around the equals operator.
2. `tot_rows` is not a good identifier name. What does it stand for?

Orange/widgets/evaluate/owconfusionmatrix.py
```diff
@@ -58,7 +58,7 @@ class OWConfusionMatrix(widget.OWWidget):
     quantities = ["Number of instances",
                   "Proportion of predicted",
                   "Proportion of actual"]
-
+    val = 2
```
kernc: Is this `val` used anywhere?

Orange/widgets/evaluate/owconfusionmatrix.py
```diff
@@ -426,11 +426,13 @@ def sum_item(value, border=""):
     def send_report(self):
         if self.results is not None and self.selected_learner:
             index = self.selected_learner[0]
+            cmatrix = confusion_matrix(self.results, index)
+            n= len(cmatrix)
```
kernc: What is the length of the confusion matrix? :smiley:

Orange/canvas/report/report.py
wadhwasahil: total no of distinct classes

Orange/widgets/evaluate/owconfusionmatrix.py
wadhwasahil: no :P

Orange/widgets/evaluate/owconfusionmatrix.py
wadhwasahil: too confusing man.

Orange/canvas/report/report.py
kernc: Incidentally, this is the answer to my other question below.

wadhwasahil: So, is everything good to merge?

kernc: With only a few small issues left, soon! :smile:

You can [re-read the comments](https://github.com/biolab/orange3/pull/1097/files) and read between the lines. In this chunk, I'd prefer if `tot_rows` was named more explicitly. Is the parameter even necessary? Can't it be devised given `table` input?

Orange/widgets/evaluate/owconfusionmatrix.py
kernc: It's likely the number of distinct class values. To avoid computing confusion matrix all over, could the value be devised from `self.tablemodel`?

wadhwasahil: confusion matrix is a very small matrix as it contains the total number of different class values.
The maximum size of it would be 1000*1000 for a dataset of ImageNet and that we won't be using as it contains 1k different classes.
So In my opinion, it won't be a problem to calculate the confusion matrix again as it is very small dimensional and easy to construct.
What do you think?

kernc: I'm not certain about this, but if you have 1e9 examples that belong to three distinct classes, you have to make at least 1e9 operations to compute the confusion matrix ...

Or what did you mean?

kernc: Depending on what `self.tablemodel` is, the cost of computing the number of rows can be O(1) operation.

wadhwasahil: Optimized :smile:

wadhwasahil:
@kernc Please check now. Is it ready to merge?

Orange/canvas/report/report.py
```diff
@@ -10,6 +10,7 @@
 
 
 class Report:
+
```
kernc: Might seem nitpicking, but no need for this, actually.

Orange/canvas/report/report.py
```diff
@@ -162,8 +163,7 @@ def report_plot(self, name=None, plot=None):
             self.report_html += get_html_img(plot.scene())
 
     # noinspection PyBroadException
-    def report_table(self, name, table=None, header_rows=0, header_columns=0,
-                     num_format=None):
+    def report_table(self, name, table=None, header_rows=0, header_columns=0, confusion_rows=0, num_format=None):
```
kernc: I still don't feel this is the right approach. First, you name the parameter `confusion_rows`, which possibly makes sense only when working with confusion matrices; but `report_table()` is meant to report _any_ kind of table. Second, by adding your new parameter somewhere in the middle of the parameters list (as opposed to the end), you break the function's existing API and should, at the very least, ensure no other calls rely on the old parameter order. Finally, even if you think you absolutely must use the number of rows in the table, you can get them from the `table` parameter, which is an instance of table model [or something compatible](https://github.com/biolab/orange3/pull/1097/files#diff-a8517f05dab824c1e209188aaf6661d5L243) (in either case, it knows for certain how many rows it has). So no, I don't think this new parameter is ok.

Orange/canvas/report/report.py
```diff
@@ -232,10 +232,10 @@ def fmtnum(s):
                 return s
 
         def report_list(data,
-                        header_rows=header_rows, header_columns=header_columns):
+                        header_rows=header_rows, header_columns=header_columns, confusion_rows=confusion_rows):
             cells = ["<td>{}</td>", "<th>{}</th>"]
             return join("  <tr>\n    {}</tr>\n".format(
-                join(cells[rowi < header_rows or coli < header_columns]
+                join(cells[rowi < header_rows or coli < header_columns or rowi == 3 + confusion_rows]
```
kernc: What does this do? Where does _3_ come from?

Orange/canvas/report/report.py
```diff
@@ -256,7 +256,8 @@ def report_list(data,
         else:
             body = None
         if body:
-            self.report_html += "<table>\n" + body + "</table>"
+            self.report_html += "<table border = '1px' style = 'border-collapse:collapse'>\n" + \
```
kernc: HTML attributes and their values are, by convention, [not separated with spaces](http://www.w3schools.com/html/html_attributes.asp).

Orange/canvas/report/report.py
wadhwasahil: 1 row = empty
2 row = class names
3 = class names
So we have to add the confusion_rows  + 3.
Hope I am clear .

Orange/canvas/report/report.py
wadhwasahil: I was trying to make the code more clearer.. But I will fix it.

Orange/canvas/report/report.py
wadhwasahil: I just used the extra parameter to bold the last row of the confusion matrix. If I remove it, everything will be fine except the last row will **NOT** be bold.
So it fine in that case?

janezd:
The general methods for reporting should not be patched for the specifics of the confusion matrix widget. There are two ways to do it. Either the widget should not use the general method for reporting the table and instead generate the HTML itself (simpler, but it would be better if all HTML is generated in one place, in general methods), or the method for reporting tables should get the data (e.g. `BackgroundRole`, font properties...) from the model.

Issue #1081, opened 2016-03-09
Report button on Distances widget


ajdapretnar:
Implement Report button function in Distances widget as is in other widgets (see Scatter Plot widget). Report on the used parameters.

wadhwasahil:
There is no report button on Scatter Plot Widget. So, do we have to add that?

janezd:
There is report on scatter plot. You meant scatter map? Scatter map will be removed, so no point in improving it.

wadhwasahil:
i have added it to distance transformation widget.

PR #1100, opened 2016-03-11
Report for Distance Transformation

wadhwasahil:
Report for Distance Transformation 
Fix to #1081


janezd:
Reports have to look nice. With capitals and without underscores. :)

Skip the section name - it's redundant.

Merge the two lines into a single line, as in the following examples:

- Transformation: normalization to interval [0, 1]
- Transformation: inversion (-X) and normalization to [-1, 1]
- Transformation: inversion (max(X) - X)
- Transformation: none

wadhwasahil:
@janezd Done.

janezd:
Please check the output before making the pull request. One combination that I got is, for instance, *Transformation: Inversion (1 - X), Normalization To interval [-1, 1]*.

wadhwasahil:
That corresponds to second bullet point. What's wrong with the output?

janezd:
It's sloppy: Transformation: Inversion (1 - X), **N**ormalization **T**o interval [-1, 1].

wadhwasahil:
Should I keep the word "interval" or remove it?
So, I will `lower()` the text and commit. Is it ok?

janezd:
Yes, use `lower()` or, better, just write all strings in lower case.

Orange/widgets/unsupervised/owdistancetransformation.py
```diff
@@ -69,5 +72,22 @@ def commit(self):
...
+                items['Transformation'] = 'None'
+            else:
+                items['Transformation'] = 'Inversion (' + \
+                    self.inversion_options[self.inversion_method][0] + ')'
+        else:
+            if self.inversion_method == 0:
+                items['Transformation'] = 'Normalization ' + \
+                    self.normalization_options[self.normalization_method][0]
+            else:
+                items['Transformation'] = 'Inversion (' + self.inversion_options[self.inversion_method][0] + '), Normalization ' + \
```
janezd: This line is too long. Break it before column 80.

janezd:
Oops, yes, use `lower`.

Also, you can reorganize the code to avoid duplication: just concatenate the strings for inversion and for normalization and in the end you output `s or "None"`.

janezd:
In `__init__`, it would make sense to move the report button to the same box as apply (otherwise it's wider than apply). Do this:

```
        box = gui.auto_commit(self.controlArea, self, "autocommit", "Apply",
                              checkbox_label="Apply on any change")
        gui.separator(box)
        box.layout().addWidget(self.report_button)
```

wadhwasahil:
@janezd As you can see I have to append "Normalization" and "Inversion". So, if just append both of them together then again I have to check if some field is empty or not.
So, in my opinion if-else can be left as it is while I change everything to `lower()`. What say?

janezd:
I meant something like this

```
        s = []
        if self.normalization_method:
            s.append('Normalization {}'.format(
                 self.normalization_options[self.normalization_method][0]))
        if self.inversion_method:
            s.append('Inversion ({})'.format(
                self.inversion_options[self.inversion_method][0]))
        self.report_items("Model parameters",
                          {"Transformation": " and ".join(s) or "None"})
```

But your code was OK, too. Just add `lower()`.

wadhwasahil:
I am struggling with line breaks after 80. can you provide some assistance?

janezd:
I don't get the question. Just break the line at some appropriate place before the 80th column. :)

wadhwasahil:
@janezd Done

iamved:
Is this issue about distance transformation report button? because I solved this issue two days ago #1095 . Please make me clear, if there some kind of confusion or you just fixed the same issue again?

janezd:
@iamved, you actually did a bit more and also fixed the problem with contexts. And with @wadhwasahil, we discussed more about how the report should look like. We'll use both.

Don't consider this a waste of time. The purpose of this is to get used to the code base. These simple issues are about learning, not so much about contributing. It would be a waste of everybody's time only if you were both working on something really complex, or if you wouldn't learn anything.

iamved:
okay. That's a really good thing, we are learning so much about the code base from these issues. Please add few more new issues.

iamved:
@janezd Also please reply to my query I posted in google group in your free time. Thanks

wadhwasahil:
@janezd Is it ready to merge?

kernc:
@wadhwasahil Please squash the commits together.

wadhwasahil:
@kernc Please see #1113 and I will close.

PR #1004, opened 2016-02-03
[ENH] Parallelized cross validation & other evaluation methods

kernc:
Parallel multiprocess execution of model evaluation methods.

Assumptions:
* there is some `n_cpu * (data.nbytes + C)` extra RAM available,
* passed arguments (learners, data) are picklable, else it fallbacks to one single thread.

PR #1102, opened 2016-03-11
OWTestLearners: Add stratified option, fix report crash. Closes 1085

s-alexey:
Add stritified option for k-fold and shuffle splitters.

Fix report bug: change unknown self.Bootsrap to self.ShuffleSplit
spliter.

### DIFF
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -284,18 +284,19 @@ class CrossValidation(Results):
     .. attribute:: random_state
 
     """
-    def __init__(self, data, learners, k=10, random_state=0, store_data=False,
+    def __init__(self, data, learners, k=10, stratified=True, random_state=0, store_data=False,
                  store_models=False, preprocessor=None, callback=None, warnings=None):
         super().__init__(data, len(learners), store_data=store_data,
                          store_models=store_models, preprocessor=preprocessor,
                          callback=callback)
         self.k = k
+        self.stratified = stratified
         self.random_state = random_state
-        Y = data.Y.copy().flatten()
+
         indices = None
-        if data.domain.has_discrete_class:
+        if self.stratified and data.domain.has_discrete_class:
             indices = skl_cross_validation.StratifiedKFold(
-                Y, self.k, shuffle=True, random_state=self.random_state
+                data.Y, self.k, shuffle=True, random_state=self.random_state
             )
             if any(len(train) == 0 or len(test) == 0 for train, test in indices):
                 if warnings is not None:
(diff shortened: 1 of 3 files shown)


Orange/evaluation/testing.py
```diff
@@ -447,12 +448,20 @@ def __init__(self, data, learners, n_resamples=10, train_size=None,
...
         self.test_size = test_size
+        self.stratified = stratified
         self.random_state = random_state
 
-        indices = skl_cross_validation.ShuffleSplit(
-            len(data), n_iter=self.n_resamples, train_size=self.train_size,
-            test_size=test_size, random_state=self.random_state
-        )
+        if self.stratified and data.domain.has_discrete_class:
+            Y = data.Y.copy().flatten()
```
kernc: Is a copy necessary?

s-alexey: I'm not sure. but in CrossValidation.__init__ copy is used (line 295).

kernc: Ah, yes, no longer the case after https://github.com/biolab/orange3/pull/1004. I don't think a copy is needed. Otherwise, the PR looks great! I suggest you start working on your proposal. :smiley:

s-alexey: Thanks for your review. I will remove copy soon.

kernc:
@janezd, how do you feel about this? Would you prefer a single checkbox below all radios?

janezd:
I guess it's better with separate checkboxes (the way it's now in this PR), so it's clear that they don't apply to Leave one out and random sampling.

I wonder if it would be better to put "Leave one out" after "Random sampling", so it wouldn't be hidden between the Cross validation's and Random sampling's options. (Conceptually it's indeed closer to Cross validation, but perhaps this doesn't matter.)

kernc:
I agree.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -176,6 +180,8 @@ def __init__(self):
         ibox = gui.indentedBox(rbox)
         gui.spin(ibox, self, "k_folds", 2, 50, label="Number of folds:",
                  callback=self.kfold_changed)
+        gui.checkBox(ibox, self, "cs_stratified", "Stratified",
```
kernc: Is "Stratified sampling" as a label ok?

janezd:
I prefer keeping just "Stratified". It is sufficient, and the widget is cluttered enough already.

kernc:
@s-alexey, if you agree, can you perhaps reshuffle the order options [as proposed](https://github.com/biolab/orange3/pull/1102#issuecomment-197989244).

janezd:
CodeCov reports that the non-stratified sampling is no longer covered with unit tests. Can you add the corresponding test?

astaric:
Rebased and merged in  83e871f

PR #1106, opened 2016-03-11
Update README and such

kernc:


kernc:
@astaric, @lanzagar this is pretty important. The OS X instruction from _DEVELOPMENT_ file, if still relevant, probably belong in the wiki.

zidarsk8:
@kernc One thing that would be nice in the readme, is how to completely reset the dev env.

I was trying the sample text widget and played around a bit, then I changed its name and the old name was still there when I opened orange. @markotoplak then helped me find the `~/.cache/Orange` which I had to delete to remove the old plugin, But I am not sure if that is enough or if there's something else being cached anywhere, which can cause problems.

kernc:
>  how to completely reset the dev env.

It's mentioned. It's under `--clear-widget-settings` which is under `./orange.sh --help`. :stuck_out_tongue:

README.md
```diff
@@ -4,47 +4,63 @@ Orange
...
-Orange Canvas requires PyQt, which is not pip-installable in Python 3. You
+Orange GUI requires PyQt, which is not pip-installable in Python 3. You
 have to download and install it system-wide. Make sure that the virtual
 environment for orange is created with `--system-site-packages`, so it will
 have access to the installed PyQt4.
 
-To start Orange Canvas from the command line, run:
+To start Orange GUI from the command line, run:
+
+    ./orange.sh
```
lanzagar: Before, the command to start Orange worked from everywhere (assuming python3 could be run).
Now we say run "./orange.sh" and this depends on where the user is ("." has to be the Orange dir)!

PR #1092, opened 2016-03-09
Report Creation for Linear Regression

kwikadi:
Attempts to fix Issue #1079

PR #1107, opened 2016-03-11
Layout Fixed and Report Generated for SGD Regression

### LINKED ISSUE #1078: Report for Stochastic Gradient Regression
Implement Report function in Stochastic Gradient Regression as in other widgets. Report on learner name, model parameters and data.

wadhwasahil:
Fixes #1078 and layout improved. Now box is adjusted.

### DIFF
--- a/Orange/widgets/regression/owsgdregression.py
+++ b/Orange/widgets/regression/owsgdregression.py
@@ -1,8 +1,7 @@
 # -*- coding: utf-8 -*-
+from collections import OrderedDict
 
 from PyQt4 import QtGui
-from PyQt4.QtGui import QGridLayout, QLabel
-from PyQt4.QtCore import Qt
 
 from Orange.data import Table
 from Orange.regression.linear import SGDRegressionLearner, LinearModel
@@ -25,24 +24,30 @@ class OWSGDRegression(OWProvidesLearner, widget.OWWidget):
     learner_name = settings.Setting("SGD Regression")
 
     alpha = settings.Setting(0.0001)
-    #: epsilon parameter for Epsilon SVR
     epsilon = settings.Setting(0.1)
     eta0 = settings.Setting(0.01)
     l1_ratio = settings.Setting(0.15)
     power_t = settings.Setting(0.25)
     n_iter = settings.Setting(5)
-
-    #: Loss of function to be used
     SqLoss, Huber, Epsilon_i, SqEpsilon_i = 0, 1, 2, 3
     L1, L2, ElasticNet = 0, 1, 2
-    #: Selected loss of function
     loss_function = settings.Setting(SqLoss)
     penalty_type = settings.Setting(L2)
-    InvScaling, Constant = 0, 1
+    Constant, InvScaling = 0, 1
     learning_rate = settings.Setting(InvScaling)
 
     want_main_area = False
     resizing_enabled = False
+
+    LOSS_FUNCTIONS = ["Squared loss",
+                      "Huber",
+                      "Epsilon insensitive",
+                      "Squared epsilon insensitive"]
+    PENALTIES = ["Absolute norm (L1)",
+                 "Euclidean norm (L2)",
+                 "Elastic net (L1 and L2)"]
+    LEARNING_RATES = ["Constant",
+                      "Inverse scaling"]
 
     def __init__(self):
         super().__init__()
@@ -53,82 +58,58 @@ class OWSGDRegression(OWProvidesLearner, widget.OWWidget):
         box = gui.widgetBox(self.controlArea, self.tr("Name"))
         gui.lineEdit(box, self, "learner_name")
 
-        form = QGridLayout()
-        typebox = gui.radioButtonsInBox(
-            self.controlArea, self, "lossfunc", [],
-            orientation=form,
-        )
+        def add_form(box):
+            gui.separator(box)
+            box2 = gui.hBox(box)
+            gui.rubber(box2)
+            form = QtGui.QFormLayout()
+            form.setContentsMargins(0, 0, 0, 0)
+            box2.layout().addLayout(form)
+            return form
 
-        # Loss function control
-        box = gui.widgetBox(self.controlArea, self.tr("Loss function to be used"))
-        buttonbox = gui.radioButtonsInBox(
-            box, self, "loss_function",
-            btnLabels=["Squared loss",
-                       "Huber",
-                       "Epsilon insensitive",
-                       "Squared Epsilon insensitive"],
-            callback=self._on_func_changed
-        )
+        box = gui.radioButtonsInBox(
+            self.controlArea, self, "loss_function", box="Loss function",
+            btnLabels=self.LOSS_FUNCTIONS, callback=self._on_func_changed)
+        form = add_form(box)
+        epsilon = gui.doubleSpin(
+            box, self, "epsilon", 0.0, 10.0, 0.01, controlWidth=70)
+        form.addRow("ε:", epsilon)
 
-        parambox = gui.widgetBox(box, orientation="horizontal")
-
-        box = gui.widgetBox(self.controlArea, self.tr("Penalty"))
-        buttonbox = gui.radioButtonsInBox(
-            box, self, "penalty_type",
-            btnLabels=["Absolute norm (L1)",
-                       "Euclidean norm (L2)",
-                       "Elastic Net (both)"],
-            callback=self._on_penalty_changed
-        )
-
-        parambox = gui.widgetBox(box, orientation="horizontal")
-
-        box = gui.widgetBox(self.controlArea, self.tr("Learning rate"))
-        buttonbox = gui.radioButtonsInBox(
-            box, self, "learning_rate",
-            btnLabels=["Inverse scaling",
-                       "Constant"],
-            callback=self._on_lrate_changed
-        )
-
-
-
-        box = gui.widgetBox(self.controlArea, self.tr("Constants"))
-
-        form = QtGui.QFormLayout()
-        form.setContentsMargins(0, 0, 0, 0)
-
-        box.layout().addLayout(form)
-
-        alpha = gui.doubleSpin(box, self, "alpha", 0.0, 10.0, step=0.0001)
-        form.addRow("Alpha:", alpha)
-
-        spin = gui.doubleSpin(box, self, "eta0", 0.0, 10, step=0.01)
-        form.addRow("Eta0:", spin)
-
-        epsilon = gui.doubleSpin(box, self, "epsilon", 0.0, 10.0, step=0.01)
-        form.addRow("Epsilon:", epsilon)
-
-        l1_ratio = gui.doubleSpin(box, self, "l1_ratio", 0.0, 10.0, step=0.01)
+        box = gui.radioButtons(
+            self.controlArea, self, "penalty_type", box="Penalty",
+            btnLabels=self.PENALTIES, callback=self._on_penalty_changed)
+        form = add_form(box)
+        alpha = gui.doubleSpin(
+            box, self, "alpha", 0.0, 10.0, 0.0001, controlWidth=70)
+        form.addRow("α:", alpha)
+        l1_ratio = gui.doubleSpin(
+            box, self, "l1_ratio", 0.0, 10.0, 0.01, controlWidth=70)
         form.addRow("L1 ratio:", l1_ratio)
 
-        power_t = gui.doubleSpin(box, self, "power_t", 0.0, 10.0, step=0.01)
+        box = gui.radioButtonsInBox(
+            self.controlArea, self, "learning_rate", box="Learning rate",
+            btnLabels=self.LEARNING_RATES, callback=self._on_lrate_changed)
+        form = add_form(box)
+        spin = gui.doubleSpin(
+            box, self, "eta0", 0.0, 10, 0.01, controlWidth=70)
+        form.addRow("η<sub>0</sub>:", spin)
+        power_t = gui.doubleSpin(
+            box, self, "power_t", 0.0, 10.0, 0.01, controlWidth=70)
         form.addRow("Power t:", power_t)
-
-
-        # Number of iterations control
-        box = gui.widgetBox(self.controlArea, "Number of iterations")
-        gui.doubleSpin(box, self, "n_iter", 0, 1e+6, step=1)
+        gui.separator(box, height=8)
+        niterations = gui.doubleSpin(
+            box, self, "n_iter", 0, 1e+6, 1, controlWidth=70)
+        form.addRow("Number of iterations:", niterations)
 
         self._func_params = [epsilon]
         self._penalty_params = [l1_ratio]
         self._lrate_params = [power_t]
 
-        gui.button(self.controlArea, self, "&Apply",
-                   callback=self.apply, default=True)
+        box = gui.hBox(self.controlArea)
+        box.layout().addWidget(self.report_button)
+        gui.button(box, self, "&Apply", callback=self.apply, default=True)
 
         self._on_func_changed()
-
         self.apply()
 
     @check_sql_input
@@ -139,9 +120,10 @@ class OWSGDRegression(OWProvidesLearner, widget.OWWidget):
             self.apply()
 
     def apply(self):
-        loss = ["squared_loss", "huber", "epsilon_insensitive", "squared_epsilon_insensitive"][self.loss_function]
+        loss = ["squared_loss", "huber", "epsilon_insensitive",
+                "squared_epsilon_insensitive"][self.loss_function]
         penalty = ["l1", "l2", "elasticnet"][self.penalty_type]
-        learning_rate = ["invscaling", "constant"][self.learning_rate]
+        learning_rate = ["constant", "invscaling"][self.learning_rate]
         common_args = dict(
             loss=loss,
             alpha=self.alpha,
@@ -180,6 +162,23 @@ class OWSGDRegression(OWProvidesLearner, widget.OWWidget):
         for spin, enabled in zip(self._func_params, mask):
             spin.setEnabled(enabled)
 
+    def send_report(self):
+        items = OrderedDict()
+        items['Loss function'] = self.LOSS_FUNCTIONS[self.loss_function]
+        if self.loss_function != self.SqLoss:
+            items['Loss function'] +=  ", ε={}".format(self.epsilon)
+        items['Penalty'] = self.PENALTIES[self.penalty_type]
+        if self.penalty_type == self.ElasticNet:
+            items['Penalty'] += ": L1 : L2 = {} : {}".format(
+                                self.l1_ratio, 1.0 - self.l1_ratio)
+        items['Penalty'] = items['Penalty'] + ', α={}'.format(self.alpha)
+        items['Learning rate'] = self.LEARNING_RATES[self.learning_rate]
+        items['Learning rate'] += ", η<sub>0</sub>={}".format(self.eta0)
+        if self.learning_rate == self.InvScaling:
+            items['Learning rate'] += ", power_t={}".format(self.power_t)
+        items['Number of iterations'] = self.n_iter
+        self.report_items("Model parameters", items)
+
     def _on_penalty_changed(self):
         enabled = [[False],  # l1
                    [False],  # l2
@@ -198,7 +197,6 @@ class OWSGDRegression(OWProvidesLearner, widget.OWWidget):
             spin.setEnabled(enabled)
 
 
-
 if __name__ == "__main__":
     import sys
     from PyQt4.QtGui import QApplication
@@ -210,4 +208,3 @@ if __name__ == "__main__":
     ow.show()
     a.exec_()
     ow.saveSettings()
-


janezd:
See comments for #1091 and #1092. Reports must look nice, use consistent style, be concise...

wadhwasahil:
@janezd : Have a look now

janezd:
Loss Function -> Loss function
No of Iterations -> Number of iterations

Lists like `["Absolute Norm (L1)","Euclidean Norm (L2)", "Elastic Net"]` now appear in `__init__` and
`sendReport`. Moved them to class attributes so they can be used in both places. This simplifies code
maintenance. Use the cases from `__init__` (e.g. *Epsilon insensitive* not *Epsilon Insensitive*).

While you're at it, you can fix a few existing inconsistencies in the widget: change `"Squared Epsilon insensitive"` to `"Squared epsilon insensitive"` and `"Elastic Net (both)"` to `"Elastic net (L1 and L2)"`. Also, you can change `"Loss function to be used"` to `"Loss function"`.

Could you also move the report button to the same box as apply, like in linear regression?

wadhwasahil:
Yeah. Sure. But why did the check fail?

janezd:
I've no idea. Apparently a problem on the server. Definitely not your fault.

kernc:
Travis failed to fetch Python nightly to test on it. @janezd, it just needed a reload.

wadhwasahil:
@janezd I have everything besides moving the `report` button to the same box as `apply`. Can you guide me how to do that?

Orange/widgets/regression/owsgdregression.py
```diff
@@ -44,6 +45,11 @@ class OWSGDRegression(OWProvidesLearner, widget.OWWidget):
     want_main_area = False
     resizing_enabled = False
 
+    loss_function_options = ["Squared loss","Huber","Epsilon insensitive",
+                            "Squared Epsilon insensitive"]
+    penalty_options = ["Absolute norm (L1)","Euclidean norm (L2)",
+                       "Elastic Net (L1 and L2)"]
+    learning_rate_options = ["Inverse scaling","Constant"]
```
kernc: With multiple values like that, I guess it's best to list the one per line, [like it was originally](https://github.com/biolab/orange3/pull/1107/files#diff-8694470c8acf1a953aef01d51dc21efeL69).
And since those are [class constants](https://google.github.io/styleguide/pyguide.html?showone=Naming#Naming), please make the identifiers uppercase.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -115,7 +114,10 @@ def __init__(self):
         power_t = gui.doubleSpin(box, self, "power_t", 0.0, 10.0, step=0.01)
         form.addRow("Power t:", power_t)
 
-
+        scroll = QtGui.QScrollArea()
+        scroll.setWidget(box)
+        scroll.setWidgetResizable(True)
+        scroll.setFixedHeight(400)
```
kernc: Please see the [Coding style section](https://github.com/biolab/orange3/blob/master/CONTRIBUTING.md#coding-style) of the guidelines for how to set static Qt properties.

Orange/widgets/regression/owsgdregression.py
wadhwasahil: @kernc 
1. By one per line do you mean
L1
L2 
and so on?
2. By Making them uppercase, u mean change loss_function_options to LOSS_FUNCTION_OPTIONS?

Orange/widgets/regression/owsgdregression.py
wadhwasahil: I added this by my side as the windows was not fitting on the screen. Should I remove it completely?

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +186,15 @@ def _on_func_changed(self):
         for spin, enabled in zip(self._func_params, mask):
             spin.setEnabled(enabled)
 
+    def send_report(self):
+        items = OrderedDict()
+        items['Loss Function'] = self.loss_function_options[self.loss_function].capitalize()
+        items['Penalty'] = self.penalty_options[self.penalty_type].capitalize()
+        items['Learning Rate'] = self.learning_rate_options[self.learning_rate].capitalize()
+        items['No of Iterations'] = self.n_iter
+        self.report_items("Model parameters", items)
```
kernc: I think it's "Loss **f**unction", "Learning **r**ate", "No**.** of **i**terations". Unsure.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +186,15 @@ def _on_func_changed(self):
...
+
+
```
kernc: Prefer a single empty line between methods.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -139,7 +144,8 @@ def set_data(self, data):
             self.apply()
 
     def apply(self):
-        loss = ["squared_loss", "huber", "epsilon_insensitive", "squared_epsilon_insensitive"][self.loss_function]
+        loss = ["squared_loss", "huber", "epsilon_insensitive",
+                "squared_epsilon_insensitive"][self.loss_function]
         penalty = ["l1", "l2", "elasticnet"][self.penalty_type]
         learning_rate = ["invscaling", "constant"][self.learning_rate]
```
kernc: This is not really your fault, but I find these strings really far away from [their presentation counterparts](https://github.com/biolab/orange3/pull/1107/files#diff-8694470c8acf1a953aef01d51dc21efeR48). It's not good to have repeated code maintained at multiple places. Especially when it's simple to fix.

wadhwasahil: But in sklearn library the parameters have to be like that. e.g `l1` and `l2` are predefined in sklearn library.
I can't change these in the **apply()** function as these are inbuilt.
What do you suggest?

Orange/widgets/regression/owsgdregression.py
kernc: Yes to both.

Orange/widgets/regression/owsgdregression.py
kernc: Ah, interesting. What you are saying is that sometimes the control area of widgets is too large (too high) and it doesn't fit in the screen's vertical estate. I totally agree and I hate it when it happens. Wrapping control area contents inside a `QScrollArea` should alleviate this, but it should probably be done once on `OWWidget` for all widgets that extend it.

wadhwasahil: That would require another issue :stuck_out_tongue:

Orange/widgets/regression/owsgdregression.py
janezd: Benevolent dictator of reports (= me) prefers: "Loss function", "Learning rate", "Number of iterations". Sentence case, no unnecessary abbr. :)

Orange/widgets/regression/owsgdregression.py
kernc: I meant something like:
```py
PENALTIES = (
    ('Absolute norm (L1)',      'l1'),
    ('Euclidean norm (L2)',     'l2'),
    ('Elastic net (L1 and L2)', 'elasticnet'),
)
```
Now, there's the presentational form in the first column, and the library form in the second. But they are all defined together and in one single place.

Likewise for losses and learning rates.

This is, technically, out of scope of this PR, but the refactoring would still be welcome if in its own commit.

wadhwasahil: @kernc Thanks for your suggestions. Should I change it to this?

wadhwasahil:
@kernc Please check the code now. I have done all the changes you have said. Thanks.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +185,16 @@ def _on_func_changed(self):
         for spin, enabled in zip(self._func_params, mask):
             spin.setEnabled(enabled)
 
+    def send_report(self):
+        items = OrderedDict()
+        items['Loss function'] = self.loss_function_options[
+            self.loss_function].capitalize()
```
kernc: Please find another way to split this line.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +185,16 @@ def _on_func_changed(self):
...
+        items['Penalty'] = self.penalty_options[self.penalty_type].capitalize()
```
kernc: `capitalize()` here will lowercase "L1" and "L2" in the string. All presentational strings are properly capitalized already.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -124,9 +125,12 @@ def __init__(self):
...
 
-        gui.button(self.controlArea, self, "&Apply",
-                   callback=self.apply, default=True)
-
+        box = gui.widgetBox(self.controlArea, True, orientation="horizontal")
+        box.layout().addWidget(self.report_button)
+        gui.separator(box, 20)
+        self.btn_apply = gui.button(box, self, "&Apply",
+                                    callback=self.apply, disabled=0,
+                                    default=True)
```
kernc: Please revert above chunk. It needlessly complicates everything.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -44,6 +44,16 @@ class OWSGDRegression(OWProvidesLearner, widget.OWWidget):
...
 
+    loss_function_options = ["Squared loss",
+                             "Huber",
+                             "Epsilon insensitive",
+                             "Squared epsilon insensitive"]
+    penalty_options = ["Absolute norm (L1)",
+                       "Euclidean norm (L2)",
+                       "Elastic net (L1 and L2)"]
+    learning_rate_options = ["Inverse scaling",
+                             "Constant"]
```
kernc: These are class-level constants, thus should probably be uppercase.

wadhwasahil:
` items['Loss function'] = self.loss_function_options[

                                 self.loss_function]
` 
is it ok?

Orange/widgets/regression/owsgdregression.py
wadhwasahil: @kernc Done (y).

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +181,16 @@ def _on_func_changed(self):
         for spin, enabled in zip(self._func_params, mask):
             spin.setEnabled(enabled)
 
+    def send_report(self):
+        items = OrderedDict()
+        items['Loss function'] = self.LOSS_FUNCTION_OPTIONS[
+                                 self.loss_function]
```
kernc: What about
```py
items['Loss function'] = \
    self.LOSS_FUNCTION_OPTIONS[self.loss_function]
```
If you rename `LOSS_FUNCTION_OPTIONS` to `LOSS_FUNCTIONS` (and `LEARNING_RATES` and `PENALTIES` likewise), the it all even fits on a single line.

wadhwasahil: @kernc Done.

kernc:
@lanzagar what do you think?

wadhwasahil:
@kernc I think you tagged @lanzagar in the wrong PR. That would be this #1077 . :stuck_out_tongue:

kernc:
@lanzagar do you have any reservations with merging this?

lanzagar:
Well the changes made look OK to me.
But the issue asked for the report to include the "learner name, model parameters and data".
This PR adds only model parameters and even these are not complete (see below).
Looking at Linear Regression - it does include all 3.
I think it is really important we try to keep the reports for similar widgets (e.g. regression models) similar!

And going back to the incompleteness of model parameters - the values of constants are not included, but can result in models to be fitted quite differently.
I think one of the very important features of these reports is reproducibility, and for that we need to report on all relevant settings.

But to not be too critical - I think the PR looks good and makes progress toward what we want from the Report. It just does not resolve Issue #1078 completely yet.
I am fine with merging this, and adding to it in another future PR.

To define more precisely what the Reports should include and how to structure them, maybe a discussion with the powers that be is in order (@janezd @BlazZupan @astaric). For example, should we include the fitted model parameters (coefficients), when that is easy to do?

wadhwasahil:
@kernc @lanzagar Can we add all the extra details in the next PR please. And what parameters are missing? I can't understand.
Thanks in advance.

lanzagar:
> And what parameters are missing? I can't understand.

The widget has a box "Constants" with `Alpha`, `Eta0`, `Epsilon`, `L1 ratio`, `Power t`.
None of those values are in the report.

wadhwasahil:
Will add those.

wadhwasahil:
@kernc Changes done. Ready to roll. :)

wadhwasahil:
@janezd Have a look now. I think I have done what was needed. 
Thanks

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +179,18 @@ def _on_func_changed(self):
...
             spin.setEnabled(enabled)
 
+    def send_report(self):
+        items = OrderedDict()
+        items['Loss function'] = self.LOSS_FUNCTIONS[self.loss_function]
+        items['Penalty'] = self.PENALTIES[self.penalty_type]
+        items['Learning rate'] = self.LEARNING_RATES[self.learning_rate]
+        items['Constants'] = "Alpha={}, Eta0={}, Epsilon={},\
+                              L1 ratio={}, Power_t={}".format(self.alpha,
+                              self.eta0,self.epsilon,self.l1_ratio,
```
kernc: Space after commas.

wadhwasahil: Done. Probably. :stuck_out_tongue_closed_eyes:

wadhwasahil:
@kernc Done.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +179,18 @@ def _on_func_changed(self):
...
 
+    def send_report(self):
+        items = OrderedDict()
+        items['Loss function'] = self.LOSS_FUNCTIONS[self.loss_function]
+        items['Penalty'] = self.PENALTIES[self.penalty_type]
+        items['Learning rate'] = self.LEARNING_RATES[self.learning_rate]
+        items['Constants'] = "Alpha={}, Eta0={}, Epsilon={},\
+                              L1 ratio={}, Power_t={}".format(self.alpha,
+                              self.eta0, self.epsilon, self.l1_ratio,
+                              self.power_t)
```
kernc: @janezd would probably have this formatted differently. Not all options are applicable in all cases, i.e. L1 ratio only makes sense with elasticnet penalty.

janezd:
Report follows the widget. The problem here is the widget. I will rearrange the GUI and make a pull request to your pull request.

wadhwasahil:
@janezd What should I do?

janezd:
I created a pull request to your pull request. If you accept it, the "constants" will be where they belong.

Please reorganize the report accordingly.

- Show epsilon in parentheses after the loss function (e.g. "Loss function: Huber; ε=0.1"). Omit it when squared loss is used.
- Show alpha after the penalty type, e.g. "Penalty: absolute norm (L1); α=2.1". Show L1 ratio only if elastic net is chosen, e.g. "Penalty: elastic net (L1 and L2): L1 : L2 = 0.15 : 0.85, α=2.1".
- Similarly, power_t is shown only for inverse scaling.
- Number of iterations is shown in separate row, not together with Learning rate.

Sorry for this additional work; the widget should have been improved before we posted an issue about adding the report.

wadhwasahil:
@janezd @kernc Where does value of L2 comes from? is it 1 - L1?

kernc:
Yes, the parameter is the [ratio of the two norms](https://en.wikipedia.org/wiki/Elastic_net_regularization#Specification).

wadhwasahil:
@kernc Fixed the new look. I think we are good here.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -53,82 +58,58 @@ def __init__(self):
...
         form.addRow("L1 ratio:", l1_ratio)
 
-        power_t = gui.doubleSpin(box, self, "power_t", 0.0, 10.0, step=0.01)
+        box = gui.radioButtonsInBox(
+            self.controlArea, self, "learning_rate", box="Learning rate",
+            btnLabels=self.LEARNING_RATES, callback=self._on_lrate_changed)
+        form = add_form(box)
+        spin = gui.doubleSpin(
+            box, self, "eta0", 0.0, 10, 0.01, controlWidth=70)
+        form.addRow("η0:", spin)
```
kernc: @janezd Subscript zero (₀) is also in unicode: η₀

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +162,25 @@ def _on_func_changed(self):
         for spin, enabled in zip(self._func_params, mask):
             spin.setEnabled(enabled)
 
+    def send_report(self):
+        items = OrderedDict()
+        items['Loss function'] = self.LOSS_FUNCTIONS[self.loss_function]
+        if self.LOSS_FUNCTIONS[self.loss_function] != "Squared loss":
+            items['Loss function'] = items['Loss function'] + ",\
```
janezd: Don't break lines within string literals. Better:

            items['Loss function'] = \
                items['Loss function'] + ", ε={}".format(self.epsilon)

and, even better,

            items['Loss function'] += ", ε={}".format(self.epsilon)

Use incremental operators elsewhere, too.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +162,25 @@ def _on_func_changed(self):
         for spin, enabled in zip(self._func_params, mask):
             spin.setEnabled(enabled)
 
+    def send_report(self):
+        items = OrderedDict()
+        items['Loss function'] = self.LOSS_FUNCTIONS[self.loss_function]
+        if self.LOSS_FUNCTIONS[self.loss_function] != "Squared loss":
```
janezd: Do not compare string literals. Use:

    if self.loss_function == SqLoss:

Also elsewhere.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +162,25 @@ def _on_func_changed(self):
...
+        if self.PENALTIES[self.penalty_type] == "Elastic net (L1 and L2)":
+            items['Penalty'] = items['Penalty'] + (": L1 : L2 = {} : {}".
+                               format(self.l1_ratio, 1.0 - self.l1_ratio))
+        items['Penalty'] = items['Penalty'] + ', α={}'.format(self.alpha)
+        items['Learning rate'] = self.LEARNING_RATES[self.learning_rate] + ",\
+                                 η0={}".format(self.eta0)
+        if self.LEARNING_RATES[self.learning_rate] != "Inverse scaling":
+            items['Learning rate'] = items['Learning rate'] + (", power_t={}".
+                                     format(self.power_t))
+        items['No of iterations'] = self.n_iter
```
janezd: No need for abbreviation here: Change "No of iterations" to "Number of iterations".

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +162,25 @@ def _on_func_changed(self):
...
+        items['Loss function'] = self.LOSS_FUNCTIONS[self.loss_function]
+        if self.LOSS_FUNCTIONS[self.loss_function] != "Squared loss":
+            items['Loss function'] = items['Loss function'] + ",\
+                                     ε={}".format(self.epsilon)
+        items['Penalty'] = self.PENALTIES[self.penalty_type]
+        if self.PENALTIES[self.penalty_type] == "Elastic net (L1 and L2)":
+            items['Penalty'] = items['Penalty'] + (": L1 : L2 = {} : {}".
+                               format(self.l1_ratio, 1.0 - self.l1_ratio))
+        items['Penalty'] = items['Penalty'] + ', α={}'.format(self.alpha)
+        items['Learning rate'] = self.LEARNING_RATES[self.learning_rate] + ",\
```
janezd: Don't break lines within string literals.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +162,25 @@ def _on_func_changed(self):
...
+                                 η0={}".format(self.eta0)
+        if self.LEARNING_RATES[self.learning_rate] != "Inverse scaling":
```
kernc: Imho it's not ok to test for string equality like that. The user may change presentational string in the `LEARNING_RATE` and shouldn't be expected to also have to change strings in other places?

wadhwasahil:
@janezd Have a look now. I hope I did everything.

Orange/widgets/regression/owsgdregression.py
kernc: @janezd Is this correct: η₀ ?

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +162,22 @@ def _on_func_changed(self):
...
+        items['Loss function'] = self.LOSS_FUNCTIONS[self.loss_function]
+        if self.loss_function != self.SqLoss:
+            items['Loss function'] +=  ", ε={}".format(self.epsilon)
+        items['Penalty'] = self.PENALTIES[self.penalty_type]
+        if self.penalty_type == self.ElasticNet:
+            items['Penalty'] += ": L1 : L2 = {} : {}".format(
+                                self.l1_ratio, 1.0 - self.l1_ratio)
+        items['Penalty'] = items['Penalty'] + ', α={}'.format(self.alpha)
+        items['Learning rate'] = ", η0={}".format(self.eta0)
+        if self.learning_rate != self.InvScaling:
```
janezd: I've written this, but can't find it - perhaps I forgot to click "Comment": change `!=` to `==`. `power_t` is valid with `InvScaling`.

(To avoid unnecessary commit, you can amend this to the previous commit.)

wadhwasahil: Changed. Please check now.

Orange/widgets/regression/owsgdregression.py
janezd: Yes. I was lazy. @wadhwasahil, can you enclose the 0 in `<sub>`, </sub>`? And do the same in the report?

Orange/widgets/regression/owsgdregression.py
janezd: You were too fast. I wasn't done commenting yet.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +162,22 @@ def _on_func_changed(self):
...
+        items = OrderedDict()
+        items['Loss function'] = self.LOSS_FUNCTIONS[self.loss_function]
+        if self.loss_function != self.SqLoss:
+            items['Loss function'] +=  ", ε={}".format(self.epsilon)
+        items['Penalty'] = self.PENALTIES[self.penalty_type]
+        if self.penalty_type == self.ElasticNet:
+            items['Penalty'] += ": L1 : L2 = {} : {}".format(
+                                self.l1_ratio, 1.0 - self.l1_ratio)
+        items['Penalty'] = items['Penalty'] + ', α={}'.format(self.alpha)
+        items['Learning rate'] = ", η0={}".format(self.eta0)
```
janezd: The type of learning rate (Constant of Inverse scaling) is missing. This code now produces
```
Learning rate: , η0=0.01
```

(While fixing this, also enclose 0 in `<sub>`.)

Orange/widgets/regression/owsgdregression.py
wadhwasahil: @janezd Done.

wadhwasahil:
@janezd Have a look now.

Orange/widgets/regression/owsgdregression.py
```diff
@@ -180,6 +162,23 @@ def _on_func_changed(self):
         for spin, enabled in zip(self._func_params, mask):
             spin.setEnabled(enabled)
 
+    def send_report(self):
+        items = OrderedDict()
+        items['Loss function'] = self.LOSS_FUNCTIONS[self.loss_function]
+        if self.loss_function != self.SqLoss:
+            items['Loss function'] +=  ", ε={}".format(self.epsilon)
```
janezd: There is a double space after `+=`, but I'm merging anyway. Enough is enough. :)

Hint: use editors that check for PEP 8 compliance.

wadhwasahil: Thanks.

wadhwasahil: @janezd i hope i am not leaving a bad impact for that double space.

janezd: LOL :)

wadhwasahil: :stuck_out_tongue:

kernc: Good job!

PR #1108, opened 2016-03-11
distmatrix: Use with statement to open files

lanzagar:
This makes sure they are closed at the end (they were not before).

### DIFF
--- a/Orange/misc/distmatrix.py
+++ b/Orange/misc/distmatrix.py
@@ -140,63 +140,66 @@ class DistMatrix(np.ndarray):
         Args:
             filename: file name
         """
-        fle = open(filename, encoding=detect_encoding(filename))
-        line = fle.readline()
-        if not line:
-            raise ValueError("empty file")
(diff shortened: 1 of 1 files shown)


Orange/misc/distmatrix.py
```diff
@@ -140,63 +140,63 @@ def from_file(cls, filename):
...
-                    raise ValueError("invalid element at row {}, column {}".
+                    flag_data = flag.split("=")
+                    if len(flag_data) == 2:
+                        name, value = map(str.strip, flag_data)
+                    else:
+                        name, value = "", None
+                    if name == "axis" and value.isdigit():
+                        axis = int(value)
+                    else:
+                        raise ValueError("invalid flag '{}'".format(flag, filename))
```
janezd: Use better editors are learn to count to 80.

lanzagar: If one hurries one forgets to break lines or use proper English.
Corrected.

PR #1109, opened 2016-03-11
[WIP] LearnerScorer: Change signature of score method

lanzagar:
It used to have model as a parameter and now it gets the data directly.
This allows more freedom for individual learners to define how to score
variables.
E.g. Logistic regression and Linear regression models need to normalize
the data first, before the coefficient make sense.
Random forest on the other hand does not require normalization.

### DIFF
--- a/Orange/tests/test_logistic_regression.py
+++ b/Orange/tests/test_logistic_regression.py
@@ -14,7 +14,8 @@ class LogisticRegressionTest(unittest.TestCase):
         ca = CA(results)
         self.assertTrue(0.8 < ca < 1.0)
 
-    def test_LogisticRegressionPreprocessors(self):
+    @unittest.skip("Re-enable when Logistic regression supports normalization.")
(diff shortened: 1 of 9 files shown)


wadhwasahil:
My check also failed. Why does this happen?

astaric:
Coverage decrease is expected as a result of the skipped test.

Orange/tests/test_logistic_regression.py
```diff
@@ -14,7 +14,8 @@ def test_LogisticRegression(self):
         ca = CA(results)
         self.assertTrue(0.8 < ca < 1.0)
 
-    def test_LogisticRegressionPreprocessors(self):
+    @unittest.skip("Re-enable when Logistic regression supports normalization.")
```
kernc: I'd add a TODO or FIXME marker. Easier to spot a TODO than no TODO (e.g. PyCharm colors TODOs differently).

PR #1112, opened 2016-03-12
[ENH] Preprocessor widget: add random feature selection

### LINKED ISSUE #1087: Preprocess: add a random selection of features
Add 'Random feature selection' to Preprocessor widget. It should be similar to 'Select relevant features' , but implemented as a separate preprocessor. It should allow a selection per proportion or a number of features (again, see 'Select relevant features').

matevzkren:
Added a new preprocessor to the Preprocessor widget. Selects
features at random. Both selection per proportion and selection per number of
features implemented. FIxes #1087.

### DIFF
--- a/Orange/preprocess/fss.py
+++ b/Orange/preprocess/fss.py
@@ -121,7 +121,8 @@ class SelectRandomFeatures:
         if type(self.k) == float:
             self.k = int(len(data.domain.attributes) * self.k)
         domain = Orange.data.Domain(
-            random.sample(data.domain.attributes, self.k),
+            random.sample(data.domain.attributes,
+                          min(self.k, len(data.domain.attributes))),
             data.domain.class_vars, data.domain.metas)
         return data.from_table(domain, data)
 
--- a/Orange/widgets/data/icons/SelectColumnsRandom.svg
+++ b/Orange/widgets/data/icons/SelectColumnsRandom.svg
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 19.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="-281 373 48 48" style="enable-background:new -281 373 48 48;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:#FFFFFF;}
+	.st1{fill:#333333;}
+	.st2{fill:#666666;}
+	.st3{fill:#B2B2B2;}
+	.st4{fill:#989898;}
+	.st5{fill:#231F20;stroke:#FFFFFF;stroke-width:1.67;stroke-miterlimit:10;}
+</style>
+<rect x="-251" y="405" class="st0" width="10" height="8"/>
+<g>
+	<path class="st1" d="M-275,379v36h36v-36H-275z M-241,387v8v1v8v9h-11h-10h-1h-10v-8v-1v-8v-1v-8l0,0h10l0,0h1l0,0h10l0,0h1l0,0
+		H-241L-241,387z"/>
+	<polygon class="st2" points="-262,404 -262,396 -262,395 -262,387 -263,387 -263,395 -273,395 -273,396 -263,396 -263,404 
+		-273,404 -273,405 -263,405 -263,413 -262,413 -262,405 	"/>
+	<polygon class="st2" points="-251,387 -252,387 -252,395 -252,396 -252,404 -252,405 -252,413 -241,413 -241,404 -251,404 
+		-251,396 -241,396 -241,395 -251,395 	"/>
+	<rect x="-273" y="405" class="st0" width="10" height="8"/>
+	<rect x="-273" y="396" class="st0" width="10" height="8"/>
+	<polygon class="st0" points="-273,387 -263,387 -263,387 	"/>
+	<rect x="-273" y="387" class="st0" width="10" height="8"/>
+	<rect x="-262" y="405" class="st3" width="10" height="8"/>
+	<rect x="-262" y="396" class="st3" width="10" height="8"/>
+	<polygon class="st3" points="-262,387 -252,387 -252,387 	"/>
+	<rect x="-262" y="387" class="st3" width="10" height="8"/>
+	<rect x="-251" y="396" class="st0" width="10" height="8"/>
+	<rect x="-251" y="405" class="st0" width="10" height="8"/>
+	<polygon class="st0" points="-251,387 -241,387 -241,387 	"/>
+	<rect x="-251" y="387" class="st0" width="10" height="8"/>
+	<rect x="-262" y="395" class="st4" width="10" height="1"/>
+	<rect x="-262" y="404" class="st4" width="10" height="1"/>
+</g>
+<g>
+	<path class="st5" d="M-261,390.6h-5c0-2.4,0.8-4.3,2.5-5.9c1.7-1.6,4-2.3,6.7-2.3c1.8,0,3.3,0.3,4.6,1c1.3,0.6,2.3,1.6,3,2.8
+		c0.8,1.2,1.1,2.5,1.1,3.8c0,0.7-0.1,1.4-0.4,2.2c-0.2,0.7-0.6,1.4-1.1,2.2c-0.3,0.4-1,1.2-2.2,2.2c-1.1,1-1.8,1.7-2.1,2.1
+		c-0.3,0.4-0.5,0.9-0.7,1.6c-0.2,0.7-0.3,1.6-0.4,2.8h-5.2l0-0.8c0-1.7,0.3-3.2,0.9-4.5c0.6-1.4,1.6-2.6,3-3.8
+		c1.3-1.1,2.2-1.9,2.4-2.4c0.3-0.4,0.4-0.9,0.4-1.5c0-0.8-0.3-1.5-1-2c-0.7-0.6-1.6-0.8-2.7-0.8c-1.2,0-2.1,0.3-2.9,1
+		C-260.6,388.7-261,389.5-261,390.6z M-257.4,405c0.9,0,1.7,0.3,2.4,0.9c0.7,0.6,1,1.4,1,2.2s-0.3,1.6-1,2.3
+		c-0.6,0.6-1.4,0.9-2.3,0.9c-0.9,0-1.7-0.3-2.3-0.9c-0.6-0.6-1-1.4-1-2.3c0-0.9,0.3-1.6,1-2.2C-259,405.3-258.3,405-257.4,405z"/>
+</g>
+</svg>
--- a/Orange/widgets/data/owpreprocess.py
+++ b/Orange/widgets/data/owpreprocess.py
@@ -560,6 +560,108 @@ class FeatureSelectEditor(BaseEditor):
 # TODO: Model based FS (random forest variable importance, ...), RFE
 # Unsupervised (min variance, constant, ...)??
 
+class RandomFeatureSelectEditor(BaseEditor):
+    #: Strategy
+    Fixed, Percentage = 1, 2
+
+    def __init__(self, parent=None, **kwargs):
+        super().__init__(parent, **kwargs)
+        self.setLayout(QVBoxLayout())
+
+        self.__strategy = RandomFeatureSelectEditor.Fixed
+        self.__k = 10
+        self.__p = 75.0
+
+        box = QGroupBox(title="Strategy", flat=True)
+        self.__group = group = QButtonGroup(self, exclusive=True)
+        self.__spins = {}
+
+        form = QFormLayout()
+        fixedrb = QRadioButton("Fixed", checked=True)
+        group.addButton(fixedrb, RandomFeatureSelectEditor.Fixed)
+        kspin = QSpinBox(
+            minimum=1, value=self.__k,
+            enabled=self.__strategy == RandomFeatureSelectEditor.Fixed
+        )
+        kspin.valueChanged[int].connect(self.setK)
+        kspin.editingFinished.connect(self.edited)
+        self.__spins[RandomFeatureSelectEditor.Fixed] = kspin
+        form.addRow(fixedrb, kspin)
+
+        percrb = QRadioButton("Percentage")
+        group.addButton(percrb, RandomFeatureSelectEditor.Percentage)
+        pspin = QDoubleSpinBox(
+            minimum=0.0, maximum=100.0, singleStep=0.5,
+            value=self.__p, suffix="%",
+            enabled=self.__strategy == RandomFeatureSelectEditor.Percentage
+        )
+        pspin.valueChanged[float].connect(self.setP)
+        pspin.editingFinished.connect(self.edited)
+        self.__spins[RandomFeatureSelectEditor.Percentage] = pspin
+        form.addRow(percrb, pspin)
+
+        self.__group.buttonClicked.connect(self.__on_buttonClicked)
+        box.setLayout(form)
+        self.layout().addWidget(box)
+
+    def setStrategy(self, strategy):
+        if self.__strategy != strategy:
+            self.__strategy = strategy
+            b = self.__group.button(strategy)
+            b.setChecked(True)
+            for st, rb in self.__spins.items():
+                rb.setEnabled(st == strategy)
+            self.changed.emit()
+
+    def setK(self, k):
+        if self.__k != k:
+            self.__k = k
+            spin = self.__spins[RandomFeatureSelectEditor.Fixed]
+            spin.setValue(k)
+            if self.__strategy == RandomFeatureSelectEditor.Fixed:
+                self.changed.emit()
+
+    def setP(self, p):
+        if self.__p != p:
+            self.__p = p
+            spin = self.__spins[RandomFeatureSelectEditor.Percentage]
+            spin.setValue(p)
+            if self.__strategy == RandomFeatureSelectEditor.Percentage:
+                self.changed.emit()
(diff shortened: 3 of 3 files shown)


Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +535,112 @@ def createinstance(params):
...
+            enabled=self.__strategy == RandomFeatureSelectEditor.Percentile
+        )
+        pspin.valueChanged[float].connect(self.setP)
+        pspin.editingFinished.connect(self.edited)
+        self.__spins[RandomFeatureSelectEditor.Percentile] = pspin
+        form.addRow(percrb, pspin)
+
+        # form.addRow(QRadioButton("FDR"), QDoubleSpinBox())
+        # form.addRow(QRadioButton("FPR"), QDoubleSpinBox())
+        # form.addRow(QRadioButton("FWE"), QDoubleSpinBox())
```
kernc: Commented-out lines in the code are only allowed if preceded by a comment explaining why they are there.

Orange/widgets/data/icons/SelectColumnsRandom.svg
```diff
@@ -0,0 +1,39 @@
...
+	<path fill="#231F20" stroke="#FFFFFF" stroke-miterlimit="10" d="M20.188,16.784h-4.769c-0.024-2.336,0.776-4.267,2.402-5.792
+		s3.768-2.288,6.425-2.288c1.711,0,3.166,0.315,4.367,0.946s2.164,1.541,2.885,2.73c0.723,1.189,1.084,2.439,1.084,3.749
+		c0,0.716-0.115,1.42-0.346,2.111c-0.23,0.692-0.596,1.401-1.092,2.129c-0.291,0.425-0.99,1.141-2.094,2.147
+		c-1.055,0.959-1.717,1.635-1.984,2.029c-0.266,0.395-0.49,0.919-0.672,1.574c-0.184,0.655-0.299,1.565-0.346,2.73h-4.933
+		l-0.018-0.819c0-1.626,0.297-3.103,0.892-4.432c0.594-1.329,1.56-2.575,2.894-3.74c1.287-1.128,2.063-1.904,2.33-2.33
+		c0.266-0.424,0.4-0.922,0.4-1.492c0-0.776-0.324-1.435-0.975-1.975c-0.648-0.54-1.52-0.81-2.611-0.81
+		c-1.116,0-2.032,0.329-2.748,0.987C20.563,14.898,20.2,15.746,20.188,16.784z M23.664,30.925c0.885,0,1.641,0.304,2.266,0.91
+		c0.625,0.607,0.938,1.341,0.938,2.202c0,0.862-0.311,1.599-0.928,2.212c-0.619,0.612-1.365,0.919-2.24,0.919
+		c-0.873,0-1.616-0.307-2.229-0.919c-0.613-0.613-0.919-1.35-0.919-2.212c0-0.861,0.303-1.595,0.91-2.202
+		C22.068,31.229,22.802,30.925,23.664,30.925z"/>
```
kernc: Could the white border around the question mark be made a pixel or two wider?

screenshot - 03122016 - 08 51 51 pm

Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +535,112 @@ def createinstance(params):
...
+        kspin = QSpinBox(
+            minimum=1, value=self.__k,
+            enabled=self.__strategy == RandomFeatureSelectEditor.Fixed
+        )
+        kspin.valueChanged[int].connect(self.setK)
+        kspin.editingFinished.connect(self.edited)
+        self.__spins[RandomFeatureSelectEditor.Fixed] = kspin
+        form.addRow(fixedrb, kspin)
+
+        percrb = QRadioButton("Percentile")
```
kernc: I think ["percentile" should really be "percentage"](https://www.google.com/search?q=percent+vs+percentile) in this context. @janezd will confirm.

matevzkren: On that note, do you prefer 'Random Feature Selection' or 'Select Random Features' for the description title?

kernc: Indeed. The latter one matches the line above it as well as grammatically all the rest that start with a verb.

matevzkren:
I've corrected the icon and implemented the requested changes. Percentile -> percentage swap was also pushed preemptively, but can be quickly reverted at any point if decided against.

lanzagar:
Bug 1:
File (wine) -> Preprocess (Select Random Features)
Then change to iris dataset in File.
It still tries to sample 10 features which results in an error.

Bug 2:
Check the Auto Commit option in Preprocess, change settings for Select Random Features. They are not commited.

matevzkren:
@lanzagar Both preprocessors (Select Random & Select Relevat) auto-commit alike when spinBox = deselected && value = changed. What did you have in mind?

lanzagar:
I didn't notice it updated on editingFinished... I expected to see changes happen immediately when using the spin controls to increase/decrease the value.
We might want to reconsider this in general (for Select Relevant and in other places as well).

As for the number of features - it now does not crash on numbers higher then number of features anymore, but they can still be selected. I think the maximum value in the spin box should be in sync with the data size.

Also @kernc mentioned you might want to use the spin control from Orange.widgets.gui...

Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +535,108 @@ def createinstance(params):
...
+
+        form = QFormLayout()
+        fixedrb = QRadioButton("Fixed", checked=True)
+        group.addButton(fixedrb, RandomFeatureSelectEditor.Fixed)
+        kspin = QSpinBox(
+            minimum=1, value=self.__k,
+            enabled=self.__strategy == RandomFeatureSelectEditor.Fixed
+        )
+        kspin.valueChanged[int].connect(self.setK)
+        kspin.editingFinished.connect(self.edited)
```
kernc: To simplify, can you perhaps use `gui.spin()`?

Orange/preprocess/fss.py
```diff
@@ -121,7 +121,8 @@ def __call__(self, data):
         if type(self.k) == float:
             self.k = int(len(data.domain.attributes) * self.k)
         domain = Orange.data.Domain(
-            random.sample(data.domain.attributes, self.k),
+            random.sample(data.domain.attributes,
+                          min(self.k, len(data.domain.attributes))),
```
kernc: Is it possible to perhaps set this maximum also on the UI spin box?

matevzkren:
> I didn't notice it updated on editingFinished... I expected to see changes happen immediately when using the spin controls to increase/decrease the value.
> We might want to reconsider this in general (for Select Relevant and in other places as well).

@lanzagar Though random feature selection is extremely lightweight, I don't think an update on every spin click _in general_ is preferred. With calculation time usually dependant on data size, the current update practice feels safer. Admittedly, there might be valid reasoning I'm unaware of!

kernc:
@matevzkren, what do you think, could you add checks so that no more than `len(data.domain.attributes)` features can be selected in spin boxes and that these maximums update when new data is connected?

Regarding `gui.spin`, I'm not as certain.

matevzkren:
@kernc Please consider this to be a work in progress. I wanted to play around with gui.py and see what could be created. I'm starting to realise you only intended for me to use `gui.spin` however. Note that everything can be changed (depending on what you think is better, I **can avoid gui.py** altogether or only use `gui.spin`).

Several extras were implemented, including how percentage `gui.hSlider` is presented and handled (`QAbstractSlider.setTracking` simply didn't meet my expectations). Underlying spins are disabled when a different strategy is in play. No more than `len(data.domain.attributes)` features can even be selected.

Overall, the code is somewhat specific to this preprocessor. For example, `k_spin` is inserted into the radio button. If required, the implementation can be simplified and some extras taken away.

Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +536,125 @@ def createinstance(params):
...
+            self.edited.emit()
+        else:
+            self.changed.emit()
+
+    def update_p_spin(self):
+        curr_val = self._p
+        prev_val = self._ticks_to_update_slider[1]
+        if (self._ticks_to_update_slider[0] is not None and (prev_val is None or
+            any([curr_val < tick <= prev_val or curr_val >= tick > prev_val
+                for tick in self._ticks_to_update_slider[0]]))):
```
kernc: [Pylint reports](https://travis-ci.org/biolab/orange3/jobs/120086851) bad-continuation for this condition. Please make it neater.

Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +536,125 @@ def createinstance(params):
...
+        if edited:
+            self.edited.emit()
+        else:
+            self.changed.emit()
+
+    def update_p_spin(self):
+        curr_val = self._p
+        prev_val = self._ticks_to_update_slider[1]
+        if (self._ticks_to_update_slider[0] is not None and (prev_val is None or
+            any([curr_val < tick <= prev_val or curr_val >= tick > prev_val
```
kernc: `any()` is great, but it also accepts a generator and returns as soon as it finds the first truish item. No need to build the whole list.

Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +536,125 @@ def createinstance(params):
...
+            if self._max_num_attributes is None:
+                k_spin.setMaximum(10)
+                p_spin.setTickInterval(100/5)
+                self._ticks_to_update_slider = [None, None]
+            else:
+                k_spin.setMaximum(self._max_num_attributes)
+                p_spin.setTickInterval(p_spin.maximum() /
+                                       self._max_num_attributes)
+                self._ticks_to_update_slider = \
+                    [[math.ceil(i*p_spin.maximum()/self._max_num_attributes)
```
kernc: `numpy` is already imported and it provides `ceil()`.

Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +536,125 @@ def createinstance(params):
 # TODO: Model based FS (random forest variable importance, ...), RFE
 # Unsupervised (min variance, constant, ...)??
 
+class RandomFeatureSelectEditor(BaseEditor):
+    #: Strategy
+    Fixed, Percentage, FDR, FPR, FWE = 0, 1, 2, 3, 4
```
kernc: What are FDR, FPR, and FWE?

Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +536,125 @@ def createinstance(params):
...
+
+        percentage_rb = gui.appendRadioButton(group=strategy_buttons,
+                                              label="Percentage:",
+                                              insertInto=box)
+        ibox = gui.indentedBox(widget=box)
+        p_spin = gui.hSlider(
+            widget=ibox, master=self, value="_p", minValue=0, maxValue=100,
+            step=4, ticks=20, labelFormat="%d %%", callback=self.update_p_spin,
+            disabled=self._strategy != RandomFeatureSelectEditor.Percentage)
+        self._spins[RandomFeatureSelectEditor.Percentage] = p_spin
```
kernc: Percentage editor is not a spin box. You don't even need a reference to "`p_spin`" and if you use gui.py constructs, you don't need `set_p()` method — you simply assign to `self._p` and the slider's value will be updated. Similar goes for `self._k` and its spin box, except you need its reference for setting the max value.

Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +536,125 @@ def createinstance(params):
...
+            k_spin = self._spins[RandomFeatureSelectEditor.Fixed]
+            p_spin = self._spins[RandomFeatureSelectEditor.Percentage]
+            if self._max_num_attributes is None:
+                k_spin.setMaximum(10)
+                p_spin.setTickInterval(100/5)
+                self._ticks_to_update_slider = [None, None]
+            else:
+                k_spin.setMaximum(self._max_num_attributes)
+                p_spin.setTickInterval(p_spin.maximum() /
+                                       self._max_num_attributes)
```
kernc: No need to adjust `p_spin`. It has percentage granularity since initialization, and that's that.

Orange/widgets/data/owpreprocess.py
```diff
@@ -1803,6 +1938,8 @@ def __on_modelchanged(self):
     def set_data(self, data=None):
         """Set the input data set."""
         self.data = data
+        self.controler.set_max_num_attributes(len(data.domain.attributes) if
+                                              self.data else None)
```
kernc: Bad line breaking. Before `if` would be better, or the whole argument in another line. Please don't rely on PyCharm doing too much of your work. :laughing:

Orange/widgets/data/owpreprocess.py
```diff
@@ -535,6 +536,125 @@ def createinstance(params):
...
+        if self._strategy != strategy:
+            self._strategy = strategy
+            self.update_strategy(edited=False)
+
+    def setParameters(self, params):
+        self.set_k(params.get("k", 10))
+        self.set_p(params.get("p", 75))
+        self.set_max_num_attributes(params.get("max_num_attributes", None))
+        self.set_strategy(params.get("strategy",
+                                     RandomFeatureSelectEditor.Fixed))
```
kernc: You can just assign to `self._k`, `self._p`, and `self._strategy` here, and their respective controls will be updated. That's what `master, value` parameters of gui.py controls are for. Then you probably also don't need `set_strategy()` and `update_strategy()` methods. If `parameters()` and `createinstance()` below are the only two entry points for `OWPreprocess` to use, you don't even need callbacks on your gui.py elements.

kernc:
I left you some comments. Looks good and seems to work well. Perhaps the previous commiter (@ales-erjavec) would prefer you use his provided `edited, changed, setParameters()` interface. I'm not so sure mixing the styles is such a good idea (:laughing:), but I like gui.py. As you can see from my comments above, it makes Orange GUI programming somewhat easier, most of the time.

Orange/widgets/data/owpreprocess.py
ales-erjavec: > ... gui.py constructs, you don't need set_p() method — you simply assign to self._p and the slider's value will be updated.

I doubt that would work in this case.

kernc: Oh, right, that only works in classes that extend `OWComponent`. :unamused:

Orange/widgets/data/owpreprocess.py
```diff
@@ -1258,6 +1386,13 @@ def setModelData(self, widget, index):
         assert isinstance(params, dict)
         self._model.setData(index, params, ParametersRole)
 
+    def set_max_num_attributes(self, num_data_attributes=None):
+        self.max_num_attributes = num_data_attributes
+        for widget in self.view.widgets():
+            row = self.view.indexOf(widget)
+            index = self.model().index(row, 0)
+            self.setWidgetData(widget, index)
+
```
ales-erjavec: This should not be here (in this class).

In fact the individual preprocessors (edited by this widget) should not (in fact can't) be dependent on the input data.

matevzkren:
Thanks for the comments,

@kernc I chose to reference "p_spin" for the following reasons. `hSliders` do have percentage granularity enabled by default, but it remains the same throughout, as far as I can tell. I chose to edit it in an attempt to offer the user a visual connection, i.e. portray how many variables there are in total and how many more percent to trigger the next threshold. If the feature isn't needed, I'll delete. Writing it down as a spin was to simplify, but I can change that.

Also, the random selection of features is not ran on every percentage tick. `self.edited.emit()` is only used when the conditions are met and a different integer number of variables is requested. This causes a lot less updates and feels nicer overall (`update_p_spin`).

In regard to the auxiliary functions, I primarily wanted to match the signals from previous implementations of other preprocessors. Imo, `update_strategy()`is still needed to set the correct `setDisabled` marks (that seems to not happen automatically, but I could be missing something). Having removed the auxilliary functions, the preprocessor worked fine, but was definitely not as fluid overall. I will look into it more tomorrow, then report back.

@ales-erjavec I made some changes, but I'm not sure whether it's correct. Either way, some further explanation would be appreciated (if you don't mind).

Thanks to both, again. I'm not sure how I'm letting some of these mistakes through, but hey-yo ^^

kernc:
> I chose to reference "p_spin" for the following reasons. `hSliders` do have percentage granularity enabled by default, but it remains the same throughout, as far as I can tell. I chose to edit it in an attempt to offer the user a visual connection, i.e. portray how many variables there are in total and how many more percent to trigger the next threshold. If the feature isn't needed, I'll delete. Writing it down as a spin was to simplify, but I can change that.

> Also, the random selection of features is not ran on every percentage tick. `self.edited.emit()` is only used when the conditions are met and a different integer number of variables is requested. This causes a lot less updates and feels nicer overall (`update_p_spin`).

Understood. IMHO `data[:, selected_cols]` was never a bottleneck and we should dedicate exactly 0% of code complexity to such optimizations.

Orange/widgets/data/owpreprocess.py
kernc: @matevzkren, with this not working out of the box, I don't think there is much reason to obfuscate code by using gui.py widgets. So either have this class extend `gui.OWComponent` (some adjustment might be necessary) to reap its benefits (i.e. dropping the setter methods as noted in the comment above), or revert to using Qt directly as the other widgets here do (in which case I'm sorry to have misled you :smile:).

ales-erjavec:
> @ales-erjavec I made some changes, but I'm not sure whether I'm in the clear. If you don't mind, some further explanation would be appreciated.

Regarding the max_num_attributes. When the preprocessor will actually get the data it will in general not have the same number of attributes as input data to the widget, but the result of applying the preprocessors (on the input data) before it. In particular that number could be larger.
Example: Imagine a dataset with 2 discrete features each with 5 values/levels and the first preprocessor applied is a "Continuize" which expands the domain into 10 indicator columns. A "Random Feature Selection" after it would still only allow the user to select 2 features. I think that having the GUI impose such arbitrary limits is counter productive.

PR #1114, opened 2016-03-14
Refactor evaluation.testing module

s-alexey:
Working on  #1102 I have found that there are something that can be improved:

- inherit TestOnTrain from TestOnTest.
- Make the main loop body in `__init__` method of each method as a separated method (`fit(data)`).
- Some settings/data raises errors in sklearn functions (for example, when test set is too small). Possibly, we need to add 'check(data)' method.
- Use LeaveOneOut from sklearn.
- `sample` function is used only in tests.
- remove preprocessor per sample calls.

Work in progress, but I want your suggestions.

### DIFF
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -42,7 +42,7 @@ class Results:
         folds (List[Slice or List[int]]): A list of indices (or slice objects)
             corresponding to rows of each fold.
     """
-
+    score_by_folds = True
     # noinspection PyBroadException
     # noinspection PyNoneFunctionAssignment
     def __init__(self, data=None, nmethods=0, *, learners=None, train_data=None,
@@ -350,6 +350,30 @@ class Results:
         """
         raise NotImplementedError()
 
+    def split_by_model(self):
+        """Split evaluation results by models
+        """
+        data = self.data
+        nmethods = len(self.predicted)
+        for i in range(nmethods):
+            res = Results()
+            res.data = data
+            res.domain = self.domain
+            res.row_indices = self.row_indices
+            res.actual = self.actual
+            res.folds = self.folds
+            res.score_by_folds = self.score_by_folds
+
+            res.predicted = self.predicted[(i,), :]
+            if getattr(self, "probabilities", None) is not None:
+                res.probabilities = self.probabilities[(i,), :, :]
+
+            if self.models is not None:
+                res.models = self.models[:, i]
+
+            res.failed = [self.failed[i]]
+            yield res
+
 
 class CrossValidation(Results):
     """
@@ -368,8 +392,6 @@ class CrossValidation(Results):
     """
     def __init__(self, data, learners, k=10, stratified=True, random_state=0, store_data=False,
                  store_models=False, preprocessor=None, callback=None, warnings=None):
-        self.folds = []
-        self.learners = learners
         self.k = k
         self.stratified = stratified
         self.random_state = random_state
@@ -399,6 +421,7 @@ class CrossValidation(Results):
 
 class LeaveOneOut(Results):
     """Leave-one-out testing"""
+    score_by_folds = False
 
     def __init__(self, data, learners, store_data=False, store_models=False,
                  preprocessor=None, callback=None):
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -59,8 +59,8 @@ class Results:
     # noinspection PyBroadException
     # noinspection PyNoneFunctionAssignment
     def __init__(self, data=None, nmethods=0, nrows=None, nclasses=None,
-                 store_data=False, store_models=False, domain=None,
-                 actual=None, row_indices=None,
+                 learners=None, store_data=False, store_models=False,
+                 domain=None, actual=None, row_indices=None,
                  predicted=None, probabilities=None,
                  preprocessor=None, callback=None):
         """
@@ -104,11 +104,13 @@ class Results:
         self.models = None
         self.folds = None
         self.indices = []
-        self.learners = []
         self.row_indices = row_indices
         dtype = np.float32
         self.preprocessor = preprocessor
         self.callback = callback
+        self.learners = learners
+        if learners:
+            nmethods = len(learners)
 
         def set_or_raise(value, exp_values, msg):
             for exp_value in exp_values:
@@ -266,21 +268,24 @@ class Results:
         predictions.name = data.name
         return predictions
 
-    def fit(self, data):
+    def fit(self, train_data, test_data=None):
+        test_data = test_data or train_data
+
+        self.prepare_arrays(test_data)
         n_callbacks = len(self.learners) * len(self.indices)
         if self.store_models:
-            self.models = [[None] * len(self.learners)] * len(self.indices)
+            self.models = np.tile(None, (len(self.indices), len(self.learners)))
 
         ptr = 0
         for idx, (train_indices, test_indices) in enumerate(self.indices):
-            train_data = data[train_indices]
-            test_data = data[test_indices]
+            train_fold = train_data[train_indices]
+            test_fold = test_data[test_indices]
 
             if self.preprocessor is not None:
-                train_data = self.preprocessor(train_data)
+                train_fold = self.preprocessor(train_fold)
 
             for k, learner in enumerate(self.learners):
-                model = self.train_if_succ(k, learner, train_data)
+                model = self.train_if_succ(k, learner, train_fold)
                 self.call_callback((len(self.learners) * idx + k) / n_callbacks)
                 if not model:
                     continue
@@ -288,18 +293,18 @@ class Results:
                 if self.store_models:
                     self.models[idx][k] = model
 
-                result_slice = slice(ptr, ptr + len(test_indices))
+                result_slice = slice(ptr, ptr + len(test_fold))
 
-                if data.domain.has_discrete_class:
-                    values, probs = model(test_data, model.ValueProbs)
+                if train_data.domain.has_discrete_class:
+                    values, probs = model(test_fold, model.ValueProbs)
                     self.predicted[k][result_slice] = values
                     self.probabilities[k][result_slice, :] = probs
-                elif data.domain.has_continuous_class:
-                    values = model(test_data, model.Value)
+                elif train_data.domain.has_continuous_class:
+                    values = model(test_fold, model.Value)
                     self.predicted[k][result_slice] = values
                 else:
                     raise Results.DataError("Unknown target type")
-            ptr += len(test_indices)
+            ptr += len(test_fold)
 
         self.call_callback(1)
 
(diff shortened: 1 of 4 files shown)


kernc:
Maaaan, that's so much conflict with my https://github.com/biolab/orange3/pull/1004. :scream: 

But if you say it works and is compatible with #1004, I'm totally up for resolving.

Provided mine doesn't get merged first. :imp:

kernc:
> sample function is used only in tests.

Yeah, that's an interesting peculiarity of this codebase. Thanks for noticing. :heart:

Orange/evaluation/testing.py
```diff
@@ -289,58 +347,30 @@ def __init__(self, data, learners, k=10, random_state=0, store_data=False,
...
-                    values, probs = model(test_data, model.ValueProbs)
-                    self.predicted[i][fold_slice] = values
-                    self.probabilities[i][fold_slice, :] = probs
-                elif data.domain.has_continuous_class:
-                    values = model(test_data, model.Value)
-                    self.predicted[i][fold_slice] = values
-            ptr += len(test)
-        self.call_callback(1)
+        if self.preprocessor is not None:
+            data = self.preprocessor(data)
```
janezd: Preprocessor has to be called on each training set separately, not on the entire data set. Preprocessing is considered a part of the learning and should not be able to "see" and use the test data.

Orange/evaluation/testing.py
```diff
@@ -351,89 +381,17 @@ def __init__(self, data, learners, store_data=False, store_models=False,
...
-                    self.predicted[i][test_idx] = values
-                    self.probabilities[i][test_idx, :] = probs
-                elif data.domain.has_continuous_class:
-                    values = model(test_data, model.Value)
-                    self.predicted[i][test_idx] = values
-        self.call_callback(1)
+        self.indices = skl_cross_validation.LeaveOneOut(len(data))
 
+        if self.preprocessor is not None:
+            data = self.preprocessor(data)
```
janezd: This preprocessor, too, I guess has to be called for each fold separately. @BlazZupan, what do you think -- how wrong is it to call preprocessor on the entire data set in leave one out? It is not a big difference, but on the other hand, learning is supposed to be much slower than preprocessing, so calling the preprocessor on each training subset should not be such a terrible overhead to justify cheating.

s-alexey: I can change it.
But, actually, what kind of preprocessing this function usually does? If it can do something like subtracting mean value then `preprocess(test)` just turns values into zeros. I was a little confused because of an absence of `fit` and `transform` methods.

janezd: It can do anything, in principle. If it does anything based on values computed from the data, it is wrong to use the entire data set. If it, for instance, subtracts the mean, this mean has to be computed from training data only. It must not see, use or touch the test data in any way.

s-alexey: I agree, that's why it is better to use `preprocess.fit(train)` and `preprocess.transform(test)`. In the case of mean, it's not clear when should preprocess compute mean value and haw can it  differ train data from test data.

janezd: Sure it can differ. It's different data, so it would be strange if means were the same. However, preprocessors can do something more complicated and be susceptible to, say, outliers. In this case seeing the entire data set is even a greater sin. :)

s-alexey:
I agree that `preprocess(train)` should not know about test data. But preprocess(test) can know about train data (for instance  `test - mean(train)`).

Orange/evaluation/testing.py
s-alexey: I agree that `preprocess(train)` should not know about test data. But preprocess(test) can know about train data (for instance  `test - mean(train)`).

kernc: Isn't this how it works now?

s-alexey: Now it runs preprocess only on train data ([cross validation](https://github.com/biolab/orange3/blob/master/Orange/evaluation/testing.py#L319-#L320) and [leave-one-out](https://github.com/biolab/orange3/blob/master/Orange/evaluation/testing.py#L382-#L383)).

kernc: Ahh, yes, I always forget. The way this works (since way before scikit-learn's comprehensible interface) is that the preprocessor modifies the returned `train_data`'s domain in such a way that its variables have `compute_value` callable attribute, which is probably _conceptually_ roughly analogous to scikit's `transform()`. So the preprocessing of test data happens when the `train_data`'s and `test_data`'s domains are matched ([somewhere inside](https://github.com/biolab/orange3/blob/master/Orange/data/table.py#L316) the [model's](https://github.com/biolab/orange3/blob/master/Orange/base.py#L128) [call](https://github.com/biolab/orange3/blob/master/Orange/evaluation/testing.py#L336)).

Who designed this is something I wonder every day.

janezd: > Who designed this is something I wonder every day.

I did. :) It's one of the few things in Orange that I've designed twenty years ago and am still proud of. Split data into train and test (not with CV but with Sample widget), take the training data, discretize some attributes, do something else with others, multiply a few of them in the Feature constructor, put the data through a few other widgets and finally use it for fitting a model. If you now want to test the model on the test data, the chain of transformations need to be stored in the variables.

s-alexey: Thanks, I've got it now. And I have changed commit.

Orange/evaluation/testing.py
```diff
@@ -289,59 +348,26 @@ def __init__(self, data, learners, k=10, stratified=True, random_state=0, store_
...
         if self.stratified and data.domain.has_discrete_class:
-            indices = skl_cross_validation.StratifiedKFold(
+            self.indices = skl_cross_validation.StratifiedKFold(
                 data.Y, self.k, shuffle=True, random_state=self.random_state
             )
-            if any(len(train) == 0 or len(test) == 0 for train, test in indices):
+            if any(len(train) == 0 or len(test) == 0 for train, test in self.indices):
                 if warnings is not None:
                     warnings.append("Using non-stratified sampling.")
-                indices = None
```
kernc: If any of the folds is empty, use non-stratified sampling. This is what the warning says. `self.indices` must be set to `None` for the behavior to match.

Orange/evaluation/testing.py
```diff
@@ -289,59 +348,26 @@ def __init__(self, data, learners, k=10, stratified=True, random_state=0, store_
         super().__init__(data, len(learners), store_data=store_data,
                          store_models=store_models, preprocessor=preprocessor,
                          callback=callback)
+        self.folds = []
+        self.learners = learners
```
kernc: Since this is set everywhere, could it be pulled up into `Results`?

Orange/evaluation/testing.py
```diff
@@ -268,6 +266,67 @@ def get_augmented_data(self, model_names, include_attrs=True, include_prediction
...
+        row_indices = []
+
+        ptr = 0
+        for fold_idx, (train, test) in enumerate(self.indices):
+            self.folds.append(slice(ptr, ptr + len(test)))
+            row_indices.append(test)
+            ptr += len(test)
+
+        self.row_indices = np.concatenate(row_indices, axis=0)
+        self.actual = data[self.row_indices].Y.flatten()
```
kernc: `Y.ravel()` doesn't make another copy.

Orange/evaluation/testing.py
```diff
@@ -268,6 +266,67 @@ def get_augmented_data(self, model_names, include_attrs=True, include_prediction
         predictions.name = data.name
         return predictions
 
+    def fit(self, data):
+        n_callbacks = len(self.learners) * len(self.indices)
+        if self.store_models:
+            self.models = [[None] * len(self.learners)] * len(self.indices)
```
kernc: Could be:
```py
np.tile(None, (len(self.learners), len(self.indices)))
```

Orange/evaluation/testing.py
```diff
@@ -462,65 +415,8 @@ def __init__(self, data, learners, n_resamples=10, train_size=None,
...
-        self.actual = actual
-        self.predicted = predicted.reshape(nmodels, nrows)
-        if data.domain.has_discrete_class:
-            self.probabilities = probabilities
-        self.call_callback(1)
+        self.prepare_arrays(data)
+        self.fit(data)
 
 
 class TestOnTestData(Results):
```
kernc: Could this class also make use of `Results.prepare_arrays()` and `Results.fit()`? Perhaps by storing `self._train_data` and `self._test_data` (both equal to `data` in e.g. `CrossValidation`) and setting:
`self.indices = ((Ellipses, Ellipses),)`?

Orange/evaluation/testing.py
s-alexey: Thanks, I fixed it.

Orange/evaluation/testing.py
s-alexey: Thanks, I have changed `init` and `fit` methods.

Orange/evaluation/testing.py
```diff
@@ -268,6 +269,70 @@ def get_augmented_data(self, model_names, include_attrs=True, include_prediction
...
+            ptr += len(test_fold)
+
+        self.call_callback(1)
+
+    def prepare_arrays(self, data):
+        self.folds = []
+        row_indices = []
+
+        ptr = 0
+        for fold_idx, (train, test) in enumerate(self.indices):
```
kernc: [Pylint says](https://travis-ci.org/biolab/orange3/jobs/124404928#L529) `fold_idx` and `train` are unused. Should we care?

kernc:
When this is considered feature complete, please remove the WIP tag. Thanks!

Orange/evaluation/testing.py
s-alexey: Thanks for noticing, `enumerate` should be dropped.

Orange/evaluation/testing.py
```diff
@@ -268,6 +269,70 @@ def get_augmented_data(self, model_names, include_attrs=True, include_prediction
...
+            for k, learner in enumerate(self.learners):
+                model = self.train_if_succ(k, learner, train_fold)
+                self.call_callback((len(self.learners) * idx + k) / n_callbacks)
+                if not model:
+                    continue
+
+                if self.store_models:
+                    self.models[idx][k] = model
+
+                result_slice = slice(ptr, ptr + len(test_fold))
```
kernc:
```py
result_slice = self.folds[idx]
```
Then you don't need `ptr`.

Orange/evaluation/testing.py
```diff
@@ -268,6 +269,70 @@ def get_augmented_data(self, model_names, include_attrs=True, include_prediction
...
+                elif train_data.domain.has_continuous_class:
+                    values = model(test_fold, model.Value)
+                    self.predicted[k][result_slice] = values
+                else:
+                    raise Results.DataError("Unknown target type")
+            ptr += len(test_fold)
+
+        self.call_callback(1)
+
+    def prepare_arrays(self, data):
```
kernc: Could you also add some docstrings to these new base methods? We are now slowly working towards having more docstrings. :smiley:

Orange/evaluation/testing.py
kernc: But then all `prepare_arrays()` overrides should call their supers?

s-alexey: No, all overrides are speed up versions of the base method.

lanzagar:
I tried Leave One Out on iris with majority and logistic regression.
The widget shows a warning "Some scores could not be computed" and does not have the AUC scores.

Could you check please.

s-alexey:
Thanks, @lanzagar. 
It should work now. Could you check it again, please?

lanzagar:
The problem seems to be fixed now.
Everything else also looks good to me. I have not heard any other objections or considerations so I am merging.

Issue #1041, opened 2016-03-01
Color widget delet exception


fsaina:
When deleting the 'Color' widget an Exception occurs in owcolor.py:

> line 302, in storeSpecificSettings
>     for var in self.disc_colors]
> AttributeError: 'NoneType' object has no attribute 'disc_data'

The issue seems that the 'current_context' variable(with disc_colors and count_data attributes) isn't initialized although it is used in overriding methods: 'storeSpecificSettings' and 'retrieveSpecificSettings'

With enough info on the expected behaviour I could make a PR.

ales-erjavec:
The `current_context` is initialized and valid between calls to `openContext` and `closeContext`. This is done in `set_data` method.
The `storeSpecificSettings` should check if `current_context` is None (this is a perfectly valid state); and do nothing in that case.

ales-erjavec:
The same problem exists in in 'Distance Matrix' widget.

iamved:
Fixed issue with PR #1075

PR #1115, opened 2016-03-14
Check if current_context is not none

ebadawy:
Fix issue #1041


astaric:
Thanks, this should fix the issue. However, if #1104 gets merged, this fix will become unnecessary, as storeSpecificSettings will not be called when no open context exists.

lanzagar:
#1104 has been merged so this should no longer be necessary.

PR #1121, opened 2016-03-16
[ENH] Add pylint checks to travis

zidarsk8:
Add pylint tests to travis to make sure pull requests don't introduce any new lint issues.

If you wish to test the code in this pull request, feel free to create dummy pull requests to
https://github.com/zidarsk8/orange3/

The code has already been merged to my repo and travis runs the tests on PRs so you can test how the script will work here after it's merged.


bin/check_pylint_diff
```diff
@@ -0,0 +1,117 @@
...
+Note: all changes that are not committed will be ignored.
+
+
+Given the commit tree:
+
+       D---E---F---G---H
+            \\         /
+             A---B---C
+
+- Running '$SCRIPT' on C will check the diff between B and C
```
kernc: Shouldn't this check the diff between C and it's merge-base, E? A pull request can be of several commits and one should prefer the whole of it not introduce new errors, not just its last commit?

zidarsk8: It can not dot that, since you don't know if there even is a merge base. Imagine you're on master branch with only commits A B C,  and you run it on the C commit. The script assumes you want to just check the last commit (which is the diff betwenn B and C). And in case of

```
    I---J---K---L---M
     \
       D---E---F---G---H
            \         /
             A---B---C
```

I can not just assume it should take merge-base of C-H or C-M.

Using a default branch (such as maste) is also bad since you can't be sure that the repo even contains that branch. Jenkins just pull the commits by their hashes.

So you see, doing anything special when you're on a normal commit is really hard and not worth the efford (and I'd probably do it wrong)

kernc: >  Imagine you're on master branch with only commits A B C

Does the figure above convey this? Aren't branches usually spun _off_ of master? I might be misinterpreting the figure.

I feel like the script should compare the pull-request HEAD's results to that of:
```bash
git remote add upstream gh:biolab/orange3
git fetch upstream master
THIS_COMMIT=$(git merge-base HEAD upstream/master)
```
I really don't understand how this could not work. All topic branches are (should be) spun off (and rebased on) upstream/master. But in the alternative case, it should be possible to do it explicitly: checkout pull request, evaluate, checkout upstream/master (biolab/orange3/master), evaluate, compare?

I don't see how comparing just the last two commits can be useful at all.

zidarsk8: @kernc Okay, maybe the text is not clear, but it won't check last two commits, but only changes in the last commit. last commit is basically the diff between previous commit and current state.

As for why we don't check anything else:
* If I'm on a merge commit I know which two branches to compare.
* If I'm on a normal commit, I don't know anything about any other branch (unless that info is given as an argument)

Taking `gh:biolab/orange3 master" branch is useless for my fork. Trying to figure out which branch to take is really hard and prone to errors.

kernc: > only changes in the last commit. last commit is basically the diff between previous commit and current state

Yes, and in case of pull requests, that's not enough since PRs usually consist of multiple commits?

The use case I see is this: We have this software. We want to ensure no code from PRs is merged into master if it introduces more lint errors than it fixes.

Does this PR, as-is, satisfy that use case?

zidarsk8: Your travis checks are made so it tests the PR, so it will always work on a merge commit. Which means, it will pass if and only if the amount of lint issues will not increase After the PR is merged.

kernc: Aye, I see. Travis [makes a merge](https://docs.travis-ci.com/user/pull-requests#How-Pull-Requests-are-Tested).

> Rather than test the commits from the branches the pull request is sent from, we test the merge between the origin and the upstream branch.

Great, thanks for explaining and your patience! :smile:

zidarsk8: You can wait for me to make sure that you can only use this on merge commits to be extra safe.

bin/check_pylint_diff
```diff
@@ -0,0 +1,117 @@
...
+mkdir -p $TMP_REPO
+cp -a .git $TMP_REPO/
+cd $TMP_REPO
+git reset --hard HEAD
+
+TEST_COMMIT=$(git rev-parse HEAD)
+if [[ "$ARG1" != "" ]]; then
+  PARENT_1=$ARG1
+else
+  PARENT_1=$(git show --pretty=raw $TEST_COMMIT | grep parent | head -1 | grep -oE "[^ ]*$")
```
kernc: This could be `git show --pretty=raw $TEST_COMMIT | awk '/^parent/{ print $2 }'` for a bit simpler.

.travis.yml
```diff
@@ -41,6 +41,8 @@ script:
     - cd $TRAVIS_BUILD_DIR/dist/Orange-*
     - python -c "from Orange.tests import *"
     - PYTHONUNBUFFERED=x coverage run --source=Orange setup.py test
+    - cd $TRAVIS_BUILD_DIR
+    - ./bin/check_pylint_diff
```
kernc: Can you make it only run once, e.g., in Python 3.4 build, or better still in a separate build task that can fail fast? I believe you can do so by [setting an env variable in the build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Explicity-Including-Builds) and then only conditionally linting if the variable is set.

bin/check_pylint_diff
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
```
kernc: We don't have _bin/_ dir, we have _.travis/_.

.travis.yml
```diff
@@ -35,21 +35,10 @@ before_install:
     - set -e  # fail on any error
 
 install:
-    - source $TRAVIS_BUILD_DIR/.travis/install.sh
-    - source $TRAVIS_BUILD_DIR/.travis/install_postgres.sh
-    - source $TRAVIS_BUILD_DIR/.travis/install_pyqt.sh
-    - source $TRAVIS_BUILD_DIR/.travis/build_doc.sh
-    - pip install codecov
+    - source $TRAVIS_BUILD_DIR/.travis/stage_install.sh
```
astaric: -1 for this change

Before, travis output grouped each install section in its own group,
now, everything is grouped together and you need to scroll to the relevant part.

kernc:
@astaric, @ales-erjavec any idea why this fails?

ales-erjavec:
> ... any idea why this fails?

None. 

I would test that the current directory ends up in `sys.path`, and `pkg_resources.working_set.entries`, before running the tests.
```
python -c "import sys; assert '' in sys.path"
python -c "import pkg_resources; assert '' in pkg_resources.working_set.entries"
```

astaric:
pkg_resources.working_set.entries does not contain current directory.

Adding $ORANGE_DIR to PYTHONPATH env variable fixes the issue.

kernc:
Does any of you two have an opinion on what line change causes this change in behavior?

PR #1122, opened 2016-03-16
[FIX] Impute widget: no longer crashes on learner inputs (issue #1082)

matevzkren:
No longer crashes on learner inputs incompatible with different data domains (each attribute variable as class_var). On learner input signal, model-based imputer is automatically chosen. Any disputes are shown as (!) next to variable names.

### DIFF
--- a/Orange/widgets/data/owimpute.py
+++ b/Orange/widgets/data/owimpute.py
@@ -1,4 +1,5 @@
 import sys
+from functools import wraps
 
 from collections import namedtuple
 
@@ -94,8 +95,13 @@ class DisplayFormatDelegate(QtGui.QStyledItemDelegate):
             fmt = state.method.format
             text = fmt.format(var=var, params=state.params,
                               **state.method._asdict())
+
+            if var in show_error_vars:
+                text += " (!)"
+
             option.text = text
 
+show_error_vars = set()
 
 METHODS = (
     {"name": "Default (above)",
@@ -177,6 +183,8 @@ class OWImpute(OWWidget):
     outputs = [("Data", Orange.data.Table)]
 
     METHODS = METHODS
+    model_mthd_index = [i for i, m in enumerate(METHODS[1:-1], 1) if
+                        m.short == "model"][0]
 
     settingsHandler = settings.DomainContextHandler()
 
@@ -265,6 +273,7 @@ class OWImpute(OWWidget):
         box.layout().insertWidget(0, self.report_button)
         self.data = None
         self.learner = None
+        self.not_compatible_with_learner = None
 
     def set_default_method(self, index):
         """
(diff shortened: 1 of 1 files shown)


kernc:
Might be just me, but I think the commit message title ("Fixed Impute widget.") is [too vague](https://github.com/biolab/orange3/blob/master/CONTRIBUTING.md#commit-messages).

matevzkren:
@kernc I agree and have amended. Thanks!

As said, any code/implementation suggestions are also greatly appreciated.

Orange/widgets/data/owimpute.py
```diff
@@ -94,8 +95,16 @@ def initStyleOption(self, option, index):
             fmt = state.method.format
             text = fmt.format(var=var, params=state.params,
                               **state.method._asdict())
+
+            if error_shown and var in not_compatible_with_learner \
```
kernc: PEP8 says to prefer `(...)` for values spanning multiple lines (i.e. wrapping it all in parentheses).

Orange/widgets/data/owimpute.py
```diff
@@ -94,8 +95,16 @@ def initStyleOption(self, option, index):
...
+                    and (state.method.short == "model"
+                         or state.method.short == ""):
```
kernc: `state.method.short in ('model', '')`

Orange/widgets/data/owimpute.py
```diff
@@ -348,6 +376,20 @@ def imputed_vars_for_column(self, column):
         else:
             assert False
 
+    def _handle_commit(func):
```
kernc: The name is a bit vague. When and why is it needed? The only reference point available to a future explorer is the commit hash which ties things together. I tend to dislike this.

matevzkren: Still somewhat undecided on the decorator name and whether comments are perhaps required..

Orange/widgets/data/owimpute.py
```diff
@@ -348,6 +375,20 @@ def imputed_vars_for_column(self, column):
...
 
+    def _handle_value_errors(func):
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            global error_shown
+            try:
+                func(self, *args, **kwargs)
+                self.error()
+                error_shown = False
+            except ValueError as ve:
```
kernc: When can this error happen? If it's only when a model-based imputer is chosen and it doesn't handle the (target) variables of that type (e.g. linear regression can't impute discrete values), this could be mentioned somewhere, I guess.

matevzkren: The exception is raised on learner inadequacy for the given data domain. Whether it _only_ happens then, I can't say with certainty. Answering this, I'm beginning to think excepting the ValueError might not be the best practice here. I'll rewrite.. thanks!!

kernc: > The exception is raised on learner inadequacy for the given data domain when model-based imputation is selected.

I'd appreciate this as a comment somewhere. :smiley:

matevzkren:
@kernc I've abandoned excepting the ValueError altogether. It feels safer and more elegant.

Could anything go wrong?

astaric:
An alternative fix for the same issue has been merged in #1094 so I am closing this PR.

PR #1123, opened 2016-03-17
Fix core unittest run errors

ales-erjavec:
* Fix core unittests/doctest discovery on Windows
* Allow core unittests to run even if PyQt4 is not installed

### DIFF
--- a/Orange/util.py
+++ b/Orange/util.py
@@ -33,11 +33,10 @@ def deprecated(obj):
 
     Examples
     --------
-    >>> import sys; old_stderr, sys.stderr = sys.stderr, sys.stdout  # So doctest can catch that warning
     >>> @deprecated
     ... def old():
     ...     return 'old behavior'
-    >>> old()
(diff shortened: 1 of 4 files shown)


Orange/util.py
```diff
@@ -33,11 +33,10 @@ def deprecated(obj):
...
     Examples
     --------
-    >>> import sys; old_stderr, sys.stderr = sys.stderr, sys.stdout  # So doctest can catch that warning
     >>> @deprecated
     ... def old():
     ...     return 'old behavior'
-    >>> old()
+    >>> old()  # doctest: +SKIP
     /... OrangeDeprecationWarning: Call to deprecated ... old ...
     'old behavior'
```
kernc: What's the output on Windows?

ales-erjavec: 'C:\\... OrangeDeprecationWarning' which is the problem (doctest ELLIPSIS cannot match at the beginning of the output line)

PR #1125, opened 2016-03-18
OWRank: Add ANOVA, Chi2 and Univariate Linear Regression

thocevar:
- Add existing feature scorers from sklearn to the widget (ANOVA, Chi2, Univariate Linear Regression)
- Add imputation to preprocessors for scores from sklearn (they fail with missing data)
- Attempt to score individual attributes in datasets with mixed attribute types

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -541,10 +541,19 @@ class ScoreValueItem(QtGui.QStandardItem):
 
 
 class MySortProxyModel(QtGui.QSortFilterProxyModel):
+
+    @staticmethod
+    def comparable(val):
+        return val is not None and float('-inf') < val < float('inf')
+
     def lessThan(self, left, right):
         role = self.sortRole()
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owrank.py
```diff
@@ -531,10 +541,19 @@ def __lt__(self, other):
 
 
 class MySortProxyModel(QtGui.QSortFilterProxyModel):
+
+    @staticmethod
+    def comparable(val):
+        return val and float('-inf') < val < float('inf')
```
kernc: Does this sort nans below the rest?

kernc: Nevermind. :laughing:

thocevar: Nice catch.

lanzagar:
Impute() preprocessor handles only missing values in features. If the class variable has them it still fails. Probably the RemoveNaNClasses preprocessor needs to be added as well?

PR #1127, opened 2016-03-18
Fixed no report for test and score widget due to unreferenced Bootstrap

wibrt:
[FIX] Test&Score widget: Don't fail to report due to Bootstrap


lanzagar:
Thanks for pointing this out!
Actually, it looks as if the report function just uses a legacy reference to Bootstrap, which was replaced with ShuffleSplit in f922bd34.

Instead of commenting out the section, could you change `self.Bootstrap` to `self.ShuffleSplit` and make sure that fixes the problem?

wibrt:
commited from a VM with a generic user (jos),  
but authenticated as a valid CLA github/orange user..,
will make a new branch_PR

PR #1132, opened 2016-03-18
Standard buttons box in OWWidget

janezd:
I'm not sure about adding declarations for apply buttons. They may have different names, boxes, auto-commits, checkboxes for it, labels, different layouts. They are so diverse that any general solution will be either too complex or seldom used.

### DIFF
--- a/Orange/widgets/data/owconcatenate.py
+++ b/Orange/widgets/data/owconcatenate.py
@@ -17,6 +17,7 @@ from PyQt4.QtCore import Qt
 
 import Orange.data
 from Orange.widgets import widget, gui, settings
+from Orange.widgets.settings import Setting
 from Orange.widgets.utils.sql import check_sql_input
 
 
@@ -54,6 +55,8 @@ class OWConcatenate(widget.OWWidget):
                    "Intersection of attributes in all tables")
 
     id_roles = ("Class attribute", "Attribute", "Meta attribute")
+
+    auto_commit = Setting(True)
 
     def __init__(self):
         super().__init__()
@@ -115,9 +118,11 @@ class OWConcatenate(widget.OWWidget):
         cb.disables.append(ibox)
         cb.makeConsistent()
 
-        gui.separator(self.buttonsArea, 20)
-        gui.button(self.buttonsArea, self, "Apply Changes",
-                   callback=self.apply, default=True)
+        box = gui.auto_commit(
+            self.controlArea, self, "auto_commit", "Apply Changes",
+            "Apply on change", commit=self.apply)
+        box.layout().insertWidget(0, self.report_button)
+        box.layout().insertSpacing(1, 20)
 
     @check_sql_input
     def set_primary_data(self, data):
--- a/Orange/widgets/data/owconcatenate.py
+++ b/Orange/widgets/data/owconcatenate.py
@@ -115,15 +115,9 @@ class OWConcatenate(widget.OWWidget):
         cb.disables.append(ibox)
         cb.makeConsistent()
 
-        gui.separator(self.controlArea, 8)
-
-        box = gui.hBox(self.controlArea, box=True)
-        box.layout().addWidget(self.report_button)
-        gui.separator(box, 20)
-        gui.button(
-            box, self, self.tr("Apply Changes"),
-            callback=self.apply, default=True
-        )
+        gui.separator(self.buttonsArea, 20)
+        gui.button(self.buttonsArea, self, "Apply Changes",
+                   callback=self.apply, default=True)
 
     @check_sql_input
     def set_primary_data(self, data):
--- a/Orange/widgets/widget.py
+++ b/Orange/widgets/widget.py
@@ -11,7 +11,7 @@ from PyQt4.QtCore import QByteArray, Qt, pyqtSignal as Signal, pyqtProperty,\
 from PyQt4.QtGui import QDialog, QPixmap, QLabel, QVBoxLayout, QSizePolicy, \
     qApp, QFrame, QStatusBar, QHBoxLayout, QStyle, QIcon, QApplication, \
     QShortcut, QKeySequence, QDesktopServices, QSplitter, QSplitterHandle, \
-    QWidget
+    QWidget, QPushButton
 
 from Orange.data import FileFormat
 from Orange.widgets import settings, gui
@@ -19,7 +19,7 @@ from Orange.canvas.registry import description as widget_description
 from Orange.canvas.report import Report
 from Orange.widgets.gui import ControlledAttributesDict, notify_changed
 from Orange.widgets.settings import SettingsHandler
-from Orange.widgets.utils import vartype, saveplot, getdeepattr
+from Orange.widgets.utils import saveplot, getdeepattr
 from .utils.overlay import MessageOverlayWidget
 
 
@@ -115,10 +115,13 @@ class OWWidget(QDialog, Report, metaclass=WidgetMetaClass):
     #: area to the right of the `controlArea`).
     want_main_area = True
     #: Should the widget construct a `controlArea`.
-    want_standard_buttons_box = "vertical"
+    want_buttons_area = "auto"
     #: Should the widget construct a `standardButtons` box; valid only if
     #  `want_control_area` is `True`.
-    # Possible values are "vertical", "horizontal" and None
+    # Possible values are "vertical", "horizontal", "auto" and None.
+    # "auto" uses vertical if there are three default buttons or more,
+    # and horizontal otherwise. If "box" is added (e.g. "vertical box"),
+    # the button will be placed in a visible box.
     want_control_area = True
     #: Widget painted by `Save graph" button
     graph_name = None
@@ -191,7 +194,7 @@ class OWWidget(QDialog, Report, metaclass=WidgetMetaClass):
         self.__msgchoice = 0
 
         if self.want_basic_layout:
-            self.insertLayout()
+            self.set_basic_layout()
 
         sc = QShortcut(QKeySequence(Qt.ShiftModifier | Qt.Key_F1), self)
         sc.activated.connect(self.__quicktip)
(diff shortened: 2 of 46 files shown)


Orange/widgets/widget.py
```diff
@@ -114,6 +114,10 @@ class OWWidget(QDialog, Report, metaclass=WidgetMetaClass):
     #: area to the right of the `controlArea`).
     want_main_area = True
     #: Should the widget construct a `controlArea`.
+    want_standard_buttons_box = "vertical"
```
kernc: How about just setting this to default `'horizontal'`? Report and Save image both fit in one line. :smiley:

Orange/widgets/widget.py
```diff
@@ -251,22 +255,27 @@ def createPixmapWidget(self, parent, iconName):
...
-                self.report_button.setAutoDefault(0)
+            if self.want_standard_buttons_box:
+                self.controlArea = gui.vBox(
+                    left_side, addSpace=0)
+                self.standardButtons = gui.widgetBox(
+                    left_side, addSpace=0, spacing=9,
+                    orientation=self.want_standard_buttons_box)
+                if self.graph_name is not None:
+                    self.graphButton = gui.button(
+                        self.standardButtons, None, "&Save Graph",
```
kernc: `'&Save Image'`, please. It's most generally an image that is saved. It could be argued that "dot" , the third image exporter, is also an image format. :stuck_out_tongue:

Orange/widgets/widget.py
```diff
@@ -251,22 +255,27 @@ def createPixmapWidget(self, parent, iconName):
...
-                self.controlArea = gui.widgetBox(leftSide, margin=0)
-            if self.graph_name is not None:
-                self.graphButton = gui.button(leftSide, None, "&Save Graph")
-                self.graphButton.clicked.connect(self.save_graph)
-                self.graphButton.setAutoDefault(0)
-            if hasattr(self, "send_report"):
-                self.report_button = gui.button(leftSide, None, "&Report",
-                                                callback=self.show_report)
-                self.report_button.setAutoDefault(0)
+            if self.want_standard_buttons_box:
```
kernc: I think I'd have a similar failsafe above this line:
```py
if not self.want_standard_buttons_box:
    if self.graph_name is not None or hasattr(self, 'send_report'):
        self.want_standard_buttons_box = \
            self.__class__.want_standard_buttons_box
```

Orange/widgets/widget.py
```diff
@@ -251,22 +255,27 @@ def createPixmapWidget(self, parent, iconName):
...
+                self.controlArea = gui.vBox(
+                    left_side, addSpace=0)
+                self.standardButtons = gui.widgetBox(
```
kernc: `self.buttonsArea`?

Orange/widgets/widget.py
janezd: `want_standard_buttons_box` is on by default. We shouldn't override overrides.

Orange/widgets/widget.py
janezd: `want_standard_buttons_box` default is now `"auto"` (with `"horizontal"`, `"vertical"` and `None` as other options). `"auto"` will give vertical layout if there are more than two buttons, otherwise horizontal.

Orange/widgets/widget.py
janezd: `self.buttonsArea`, of course. Thanks.

Orange/widgets/widget.py
janezd: YES! Overlooked this now; it'll be in the next commit.

janezd:
As a design rule, I would put the commit/apply button at the bottom right of the widget (or control area), since this is where the user what expect the default, confirmation button -- not somewhere in the middle of the widget, followed by various report and save graphs. I have already changed a few nonconforming widgets.

`buttonsArea` can't be used in widgets for which we would like to have a box around the buttons. This depends on overall design. I was tempted to introduce `want_standard_buttons_box = "vertical box"`, where the flag `box` instructs `OWWidget` to put the buttons in a visible box. I decided against because I have a dream that one day we will stop using `"vertical"` as a string literal and have `gui.Vert` and `gui.Hor` instead. Suggestions? Flags? `want_standard_buttons_box = gui.Vert | OWWidget.box` (with meh for `OWWidget.box`).

I like your idea to have more declarative construction of GUI, and I'm pondering about the apply/commit. However, these declarations are complex and would require named tuples. The class attribute would be something like

   apply_button = Apply(label="Apply", callback="apply")

or

    apply_button = Apply(label="Apply", callback="apply", auto="True", checkbox_label="Apply on change")

where `Apply` is a named tuple. This would in essence just move the `auto_commit` arguments from the function call to class Attribute. In the end, it would again look like a call. I am not sure about this idea.

We could add a class attribute with a list of other buttons. A widget could declare

    want_buttons = [("Reset", "reset"), OWWidget.Spacing, ("Inform Santa", "santa"])

and the buttons would be added and connected to the corresponding methods. (Unfortunately, we have to use string here since the bound methods do not exist before instantiation...)

The names of all those `want_...` constants look ridiculous. Rename to `insert_...`?

Orange/widgets/regression/owsgdregression.py
```diff
@@ -25,25 +24,31 @@ class OWSGDRegression(OWProvidesLearner, widget.OWWidget):
...
 
+    LOSS_FUNCTIONS = ["Squared loss",
+                      "Huber",
+                      "Epsilon insensitive",
+                      "Squared epsilon insensitive"]
+    PENALTIES = ["Absolute norm (L1)",
+                 "Euclidean norm (L2)",
+                 "Elastic net (L1 and L2)"]
+    LEARNING_RATES = ["Constant",
+                      "Inverse scaling"]
```
kernc: Another PR's commits must have creeped in here.

Orange/widgets/data/owconcatenate.py
```diff
@@ -115,15 +115,9 @@ def __init__(self):
...
-
-        box = gui.widgetBox(self.controlArea, True, orientation="horizontal",)
-        box.layout().addWidget(self.report_button)
-        gui.separator(box, 20)
-        gui.button(
-            box, self, self.tr("Apply Changes"),
-            callback=self.apply, default=True
-        )
+        gui.separator(self.buttonsArea, 20)
+        gui.button(self.buttonsArea, self, "Apply Changes",
```
kernc: Why not auto apply (perhaps with default False)? I think we should have, by policy, auto commit option in all widgets.

janezd: I thought about it (haven't we also discussed it), but each widget has its special way of arranging the texts, checkboxes and buttons. If we move auto apply to declarations, we would basically move the arguments of the `auto_commit` method to widget's attributes.

kernc: I just meant, since you touched these lines, why not make it an `auto_commit` button instead of a simple one?

janezd: We don't have reasonable defaults, I guess.

janezd: Oops, I answered the comment without noticing that it's about particular widget, not `OWWidget`. 

I haven't yet checked this particular case, but you're certainly right in general: I'll add auto commits wherever appropriate.

Orange/widgets/classify/ownaivebayes.py
```diff
@@ -24,30 +24,15 @@ class OWNaiveBayes(OWProvidesLearner, widget.OWWidget):
...
+        gui.button(self.buttonsArea, self, self.tr("&Apply"),
                    callback=self.apply)
-        self.controlArea.layout().addWidget(self.report_button)
-
-        self.data = None
-        self.preprocessors = None
-
-        self.initialize()
-
-    def initialize(self):
```
janezd: I removed this because it duplicates `apply` and, besides, it's wrong since it outputs some kind of "default learner" without proper arguments.

Orange/widgets/widget.py
```diff
@@ -114,6 +115,13 @@ class OWWidget(QDialog, Report, metaclass=WidgetMetaClass):
     #: area to the right of the `controlArea`).
     want_main_area = True
     #: Should the widget construct a `controlArea`.
+    want_buttons_area = "auto"
+    #: Should the widget construct a `standardButtons` box; valid only if
+    #  `want_control_area` is `True`.
+    # Possible values are "vertical", "horizontal", "auto" and False.
+    # "auto" uses vertical if there are three default buttons or more,
+    # and horizontal otherwise. If "box" is added (e.g. "vertical box"),
+    # the button will be placed in a visible box.
```
kernc: The comment comes above the line.

Orange/widgets/widget.py
```diff
@@ -114,6 +115,13 @@ class OWWidget(QDialog, Report, metaclass=WidgetMetaClass):
     #: area to the right of the `controlArea`).
     want_main_area = True
     #: Should the widget construct a `controlArea`.
```
kernc: This comment goes below above its line.

Orange/widgets/widget.py
```diff
@@ -197,117 +207,132 @@ def __init__(self, *args, **kwargs):
...
+
+    def _create_default_buttons(self):
+        # These button is inserted in buttons_area, if it exists
+        # Otherwise it is up to the widget to add them to some layout
+        if self.graph_name is not None:
+            self.graphButton = QPushButton("&Save Image", autoDefault=False)
+            self.graphButton.clicked.connect(self.save_graph)
+        if hasattr(self, "send_report"):
+            self.report_button = QPushButton("&Report", autoDefault=False)
+            self.report_button.clicked.connect(self.show_report)
```
kernc: Could this be merged into `_insert_buttons_area()`?

kernc:
The code raises [10 new warnings](https://travis-ci.org/biolab/orange3/jobs/120198800).

Orange/widgets/widget.py
janezd: No. I extracted this from `_insert_buttons_area` so the buttons can be created without also creating `buttonsArea`.

janezd:
Cylcomatic complexity increased from 15.7 to 15.755. However, this is due to the smaller number of functions. PR is being punished for removal of redundant function, while the complexity in fact *decreased* from 785 (in 50 functions) to 772 (in 49 functions). 

Perhaps Radon should measure absolute complexity.

Orange/widgets/widget.py
```diff
@@ -197,117 +215,132 @@ def __init__(self, *args, **kwargs):
...
-            self.mainArea = gui.widgetBox(splitter,
-                                          orientation="vertical",
-                                          margin=4,
-                                          sizePolicy=QSizePolicy(QSizePolicy.Expanding,
-                                                                 QSizePolicy.Expanding))
-            splitter.setCollapsible(1, False)
-            self.mainArea.layout().setContentsMargins(0 if self.want_control_area else 4, 4, 4, 4)
-
+            self._insert_main_area()
         if self.want_status_bar:
```
kernc: Including all add-ons, there are 0 (zero) widgets that want this.

janezd: I suspected this, but I haven't checked. I'll remove it, there are too many whistles and bells here already.  (It will further increase the cyclomatic complexity, though. :)

PR #1135, opened 2016-03-24
Windows installer: update pip on target system if required

ales-erjavec:
Fix an installation error when running windows installers due to older installed pip (requires at least version >=6).


kernc:
This would probably need to be included ASAP, right?

PR #1136, opened 2016-03-25
[FIX] Fixed no report for test and score widget due to legacy Bootstr…

wibrt:
Follow up to ( PR  #1127 ), which had a github automated user checking problem

Reporting should work now using ShuffleSplit

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -599,10 +599,10 @@ class OWTestLearners(widget.OWWidget):
                       format(self.k_folds))]
         elif self.resampling == self.LeaveOneOut:
             items = [("Sampling type", "Leave one out")]
-        elif self.resampling == self.Bootstrap:
-            items = [("Sampling type",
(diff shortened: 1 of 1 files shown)


lanzagar:
There seems to be an indentation problem here (5 spaces instead of 4).
Please fix (and use git commit --amend to keep the changes in a single commit).

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -599,10 +599,10 @@ def send_report(self):
                       format(self.k_folds))]
         elif self.resampling == self.LeaveOneOut:
             items = [("Sampling type", "Leave one out")]
-        elif self.resampling == self.Bootstrap:
-            items = [("Sampling type",
+        elif self.resampling == self.ShuffleSplit:
+             items = [("Sampling type",
                       "{} random samples with {} % data ".format(
```
kernc: This line (and the subsequent lines) should be indented one space further. The string is _inside_ the tuple.

PR #1138, opened 2016-03-25
[FIX] Fixed no report for test and score widget due to legacy Bootstr…

wibrt:
Follow up to ( PR #1127 and  #1136), which had a github automated user checking and branching problem, my apologies

Reporting should work now using ShuffleSplit 
(with indentation)


kernc:
Thanks!

PR #1139, opened 2016-03-25
[FIX] Settings: Warn when setting values are not present on instance

astaric:
Fixes a problem that can be reproduced by deleting scatter plot from the following schema:

screen shot 2016-03-29 at 23 32 53

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -174,7 +174,12 @@ class SettingProvider:
         instance : OWWidget
         """
         if setting.packable:
-            yield setting.name, getattr(instance, setting.name)
+            if hasattr(instance, setting.name):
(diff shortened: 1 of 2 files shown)


Orange/widgets/settings.py
```diff
@@ -174,7 +174,12 @@ def _default_packer(setting, instance):
         instance : OWWidget
         """
         if setting.packable:
-            yield setting.name, getattr(instance, setting.name)
+            if hasattr(instance, setting.name):
+                yield setting.name, getattr(instance, setting.name)
+            else:
+                warnings.warn("{0} is declared as setting on {1} "
+                              "but not present on instance."
+                              .format(setting.name, instance))
```
kernc: When can that happen?

PR #1140, opened 2016-03-25
Context settings refactoring

astaric:


Orange/widgets/settings.py
```diff
@@ -860,6 +904,8 @@ def settings_to_widget(self, widget):
             setattr(widget, self.reservoir, ll)
 
     def fast_save(self, widget, name, value):
+        super().fast_save(widget, name, value)
```
janezd: `super().fast_save(widget, name, value)`

This is faster than fast, it's *super fast* :+1:

PR #1146, opened 2016-03-31
Cleaning unittests

rekonder:
Changes:
- Adding setUpClass method where is this possible.
- Changing assertTrue, assertFalse into assertEqual, assertGreater...
- Deleting some duplicated lines
- Deleting bad-whitespaces and unused imports


Orange/tests/test_classification.py
```diff
@@ -77,10 +77,10 @@ def test_value_from_probs(self):
         clf = learn(t)
         clf.ret = Model.Probs
         y2 = clf(x, ret=Model.Value)
-        self.assertTrue(y2.shape == (nrows,))
+        self.assertEqual(y2.shape, (nrows,))
         y2, probs = clf(x, ret=Model.ValueProbs)
-        self.assertTrue(y2.shape == (nrows, ))
-        self.assertTrue(probs.shape == (nrows, 2))
+        self.assertEqual(y2.shape, (nrows, ))
```
kernc: This bad-whitespace seems to be one of the [21 errors you introduced](https://travis-ci.org/biolab/orange3/jobs/119871036).

Orange/tests/test_ada_boost.py
```diff
@@ -8,7 +8,9 @@
 
 
 class SklAdaBoostTest(unittest.TestCase):
-    def setUp(self):
+
+    @classmethod
+    def setUpClass(self):
```
kernc: The first argument of classmethods is called `cls` by convention.

Orange/tests/test_classification.py
janezd: Well, it's still [WIP]. :) I didn't realize that we had so much to clean up in unit tests. It was about time that somebody starts cleaning this mess.

astaric:
Thanks for taking the time for cleaning the unittests. As for the WIP tag, I would prefer to merge this PR sooner than later, as it touches a lot of modules. If you plan to perform other cleanups, please open a separate PR for them.

rekonder:
@astaric  Oki I agree. I add two new commits. In first one I correct mistakes with  self in setUpClass functions. In second I delete bad-whitespaces and unused import which was found by pylint test. If you wish I can add commit with deleting bad-whitespaces and unused imports  into other PR.

kernc:
> If you wish I can add commit with deleting bad-whitespaces and unused imports  into other PR.

What you can try instead, is [squash the two fixup commits](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history) with their respective change-introducing ancestors. :smiley:

rekonder:
@kernc  I squash fixup commits ;).

Orange/tests/test_distances.py
```diff
@@ -436,148 +442,155 @@ def test_spearmanr_distance_one_example(self):
...
-                                       np.array([[ 0.3833333],
-                                                 [ 0.4666667],
-                                                 [ 0.3666667],
-                                                 [ 0.       ]]))
+                                       np.array([[0.3833333],
+                                                 [0.4666667],
+                                                 [0.3666667],
+                                                 [0.]]))
         np.testing.assert_almost_equal(self.dist(self.breast[:3], self.breast[:3], axis=0),
-                                       np.array([[ 0.  ,  0.25,  0.  ,  0.25,  0.25,  0.25,  0.25,  0.25,  0.25],
```
zidarsk8: @kernc With arrays like this, I believe that having the previous formatting is better (readability is usually more important). The case here is, in the old style with spaces, it's easy to see that the diagonal is all zeros, in the new style it's not.

As for pylint, all these tables could be moved to a file that would just contain them as constants with good names. and then in that file we could ignore multiple spaces in arrays and even lines longer than 80.

kernc: I'm leaning towards adding a few pylint exceptions, like `bad-whitespace, line-too-long, too-many-lines, too-many-public-methods, invalid-name` into all tests files. The last one to skip lengthy method names.

astaric: If matrices contain values that you need to understand, I agree. But nice formatting can also be achieved by using a fixed number of decimals (0.00 instead of 0.)

+1 for constants in a separate module, but not as a part of this PR. Distance tests need a more thorough cleanup anyway.

Orange/tests/test__orange.py
```diff
@@ -13,10 +13,6 @@ def test_valuecount(self):
         b = _valuecount.valuecount(a)
         np.testing.assert_almost_equal(b, [[1, 2], [0.6, 0.4]])
 
-        a = np.array([[0, 1, 1, 1], [0.1, 0.2, 0.3, 0.4]])
-        b = _valuecount.valuecount(a)
-        np.testing.assert_almost_equal(b, [[0, 1], [0.1, 0.9]])
```
zidarsk8: While I see that the same `a` was already used, This assert line was never duplicated and you basically just removed one test if I'm reading this correctly.

The only duplicate lines here that should have been removed are:

```
a = np.array([[0, 1, 1, 1], [0.1, 0.2, 0.3, 0.4]])
b = _valuecount.valuecount(a)
```
But I don't understand why the different assert statements still pass.

Orange/tests/test__orange.py
```diff
@@ -13,10 +13,6 @@ def test_valuecount(self):
         b = _valuecount.valuecount(a)
         np.testing.assert_almost_equal(b, [[1, 2], [0.6, 0.4]])
 
-        a = np.array([[0, 1, 1, 1], [0.1, 0.2, 0.3, 0.4]])
```
zidarsk8: Why was this removed? I might be looking at this wrongly, but I don't see where this code is duplicated.

PR #1151, opened 2016-04-01
[WIP] pythagoras tree widget

pavlin-policar:
Pythagoras tree widget for visualizing trees.

Currently only works for classification trees. Only draws the tree on the QGraphicsScene at this stage.

### DIFF
--- a/Orange/widgets/visualize/owpythagorastree.py
+++ b/Orange/widgets/visualize/owpythagorastree.py
@@ -17,7 +17,7 @@ Point = namedtuple('Point', ['x', 'y'])
 
 
 class OWPythagorasTree(OWWidget):
-    name = "Generalized Pythagoras Tree"
+    name = "Pythagoras Tree"
     description = "Generalized Pythagoras Tree for visualizing trees."
     priority = 100
 
@@ -70,7 +70,8 @@ class OWPythagorasTree(OWWidget):
     def set_ctree(self, ctree=None):
         self.clear()
         self.raw_tree = ctree
-        self.tree = SklTreeAdapter(ctree.skl_model.tree_)
+        if ctree is not None:
+            self.tree = SklTreeAdapter(ctree.skl_model.tree_)
 
         self._update()
 
--- a/Orange/widgets/visualize/owpythagorastree.py
+++ b/Orange/widgets/visualize/owpythagorastree.py
@@ -1,4 +1,5 @@
-import math
+# coding=utf-8
+from math import pi, sqrt, cos, sin, degrees
 from collections import namedtuple, defaultdict
 
 import Orange
@@ -10,8 +11,6 @@ from Orange.data.table import Table
 
 from PyQt4 import QtGui, QtCore
 from PyQt4.QtCore import Qt
-
-PI = math.pi
 
 # Please note that all angles are in radians
 Square = namedtuple('Square', ['center', 'length', 'angle'])
@@ -95,7 +94,7 @@ class OWPythagorasTree(OWWidget):
                     draw_square_rec(child)
 
             draw_square_rec(tree_builder.pythagoras_tree(
-                self.tree, 0, Square(Point(0, 0), 200, PI / 2)
+                self.tree, 0, Square(Point(0, 0), 200, pi / 2)
             ))
         else:
             pass
@@ -138,7 +137,7 @@ class SquareGraphicsItem(QtGui.QGraphicsRectItem):
         # no need to perform extra transformations where they won't be seen
         if angle % 90 != 0:
             self.setTransformOriginPoint(self.boundingRect().center())
-            self.setRotation(math.degrees(angle))
+            self.setRotation(degrees(angle))
 
     def _get_rect_attributes(self):
         """Get the rectangle attributes requrired to draw item.
@@ -237,9 +236,9 @@ class PythagorasTree:
         """
         weight = tree.weight(node)
         # the angle of the child from its parent
-        alpha = weight * PI
+        alpha = weight * pi
         # the child side length
-        length = parent_square.length * math.sin(alpha / 2)
+        length = parent_square.length * sin(alpha / 2)
         # the sum of the previous anlges
         prev_angles = sum(self._slopes[parent_square])
 
@@ -249,7 +248,7 @@ class PythagorasTree:
         # the angle of the square is dependent on the parent, the current
         # angle and the previous angles. Subtract PI/2 so it starts drawing at
         # 0rads.
-        angle = parent_square.angle - PI / 2 + prev_angles + alpha / 2
+        angle = parent_square.angle - pi / 2 + prev_angles + alpha / 2
         square = Square(center, length, angle)
 
         self._slopes[parent_square].append(alpha)
@@ -285,9 +284,9 @@ class PythagorasTree:
         t0 = self._get_point_on_square_edge(
             parent_center, parent_length, parent_angle)
         # get the edge point that we will rotate around t0
-        square_diagonal_length = math.sqrt(2 * parent_length ** 2)
+        square_diagonal_length = sqrt(2 * parent_length ** 2)
         edge = self._get_point_on_square_edge(
-            parent_center, square_diagonal_length, parent_angle - PI / 4)
+            parent_center, square_diagonal_length, parent_angle - pi / 4)
         # if the new square is not the first child, we need to rotate the edge
         if base_angle != 0:
             edge = self._rotate_point(edge, t0, base_angle)
@@ -298,7 +297,7 @@ class PythagorasTree:
         # calculate the middle point between the rotated point and edge
         t2 = Point((t1.x + edge.x) / 2, (t1.y + edge.y) / 2)
         # calculate the slope of the new square
-        slope = parent_angle - PI / 2 + alpha / 2
+        slope = parent_angle - pi / 2 + alpha / 2
         # using this data, we can compute the square center
         return self._get_point_on_square_edge(t2, length, slope + base_angle)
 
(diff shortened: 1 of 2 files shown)


kernc:
If you want, you can tend to some of these pylint warnings: https://travis-ci.org/biolab/orange3/jobs/120147877

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -0,0 +1,387 @@
...
+    ----------
+    square : Square
+        The square the represents the tree node.
+    parent : TreeNode
+        The parent of the current node.
+    children : list of TreeNode, optional
+        All the children that belong to this node.
+
+    """
+    def __init__(self, square, parent, children=[]):
```
kernc: If default `children` has to be an iterable, should be a read-only type, e.g. tuple.

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -0,0 +1,387 @@
...
+    argv = sys.argv
+    if len(argv) > 1:
+        filename = argv[1]
+    else:
+        filename = "iris"
+
+    app = QtGui.QApplication(argv)
+    ow = OWPythagorasTree()
+    data = Orange.data.Table(filename)
+    clf = TreeLearner(max_depth=10)(data)
```
kernc: @astaric, @zidarsk8 What do we do with pylint complaining about these semantically local variables [interpreted as constants](https://travis-ci.org/biolab/orange3/jobs/120147877)?

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -0,0 +1,387 @@
...
+            temp.x * math.sin(-alpha) + temp.y * math.cos(-alpha)
+        )
+        # translate rotated point back to its original spot
+        t1 = Point(t1.x + t0.x, t1.y + t0.y)
+
+        # calculate the middle point between the rotated point and edge
+        t2 = Point((t1.x + edge.x) / 2, (t1.y + edge.y) / 2)
+        return self._get_point_on_square_edge(t2, length, angle)
+
+    def _compute_center_right(self, initial_square, length, alpha, angle):
```
kernc: The code for `_center_center_left()` and `_compute_center_right()` is almost the same. If you add another two parameters `op1` and `op2` (named appropriately) where they stand for `operator.add` and `operator.sub` (or vice versa), this can be a single function?

pavlin-policar: I am in the process of changing this to a more generalized form that will support multiple children, but it isn't working fully yet.

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -0,0 +1,387 @@
...
+    """
+    def __init__(self, square, parent, children=[]):
+        self.square = square
+        self.parent = parent
+        self.children = children
+
+
+class PythagorasTree:
+    def pythagoras_tree(self, tree, node=0,
+                        square=Square(Point(0, 0), 200, 0)):
```
zidarsk8: It's usually really bad to have any mutable state as a default argument value, since it will persist through multiple function calls. For such cases you should use `None` and then a simple `if kwarg is None:` statement.

here's a section from google python styleguide: https://google.github.io/styleguide/pyguide.html?showone=Default_Argument_Values#Default_Argument_Values

Orange/widgets/visualize/owpythagorastree.py
zidarsk8: There are multiple things wrong with this code here, for one the imports are not at the top of the class and such. So A propper way to solve all of these issues would be to put this code in the main function.

```python
def main():
    # your code here

if __name__ == "__main__":
    main()
```

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -0,0 +1,387 @@
...
+        TreeNode
+            The root node which contains the rest of the tree.
+
+        """
+        children = []
+        if tree.has_children(node):
+            left_child = tree.left_child(node)
+            right_child = tree.right_child(node)
+
+            w1, w2 = tree.weight(left_child), tree.weight(right_child)
```
zidarsk8: I tried to follow the `tree.weight` but I can't see if it's possible or for both `w1` and `w2` to be zero. So if that's the case, can you make sure this does not fail with `ZeroDivisionError`. And if it's not possible, I'd be happy if you can give a quick reason. Thanks.

pavlin-policar:
I'm afraid I've managed to accidentally do strange things to the git tree. I am very sorry. It would probably be best if I created a new pull request.

Orange/widgets/visualize/owpythagorastree.py
pavlin-policar: I was not aware of that. I've change this to a tuple, so mutability should not be an issue anymore.

kernc:
No need. Go to your `widget-pythagoras-tree` branch and do:
```bash
git checkout widget-pythagoras-tree

# If you don't already have Biolab's Orange fork set as upstream remote
git remote add upstream https://github.com/biolab/orange3

git fetch upstream master
git rebase -i upstream/master
# Now reorder/fix stuff, remove commits that aren't yours. Finally

git push -f
```

pavlin-policar:
I think I've resolved my mess with the git tree. Thank you for the help.

Orange/widgets/visualize/tests/test_owpythagorastree.py
```diff
@@ -0,0 +1,63 @@
...
+        point = self.builder._get_point_on_square_edge(
+            center=Point(3.06, 2.79), length=2.39, angle=math.radians(30)
+        )
+        expected_point = Point(4.08, 3.41)
+        self.assertAlmostEqual(point.x, expected_point.x, places=1)
+        self.assertAlmostEqual(point.y, expected_point.y, places=1)
+
+    def test_get_point_on_square_edge_with_angle_2(self):
+        point = self.builder._get_point_on_square_edge(
+            center=Point(2.7, 2.77), length=1.65, angle=math.radians(20.97)
```
kernc: These three tests all seem to test `_get_point_on_square_edge`. They could be merged into a single test. And human-interpretable values like `(0, 0)`, `1`, `pi/2` should probably be preferred.

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -0,0 +1,416 @@
...
+        return self._tree.n_node_samples[node]
+
+    def parent(self, node):
+        for idx, el in enumerate(self._tree.children_left):
+            if el == node:
+                return idx
+        for idx, el in enumerate(self._tree.children_right):
+            if el == node:
+                return idx
+        return -1
```
kernc: If you say this is slow, and the `_tree` is from sklearn; can you test if this works and is faster?
```py
for children in (self._tree.children_left,
                 self._tree.children_right):
    try:
        return (children == node).nonzero()[0][0]
    except IndexError:
        continue
return -1
```
It lays the iterative burden onto numpy.

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -0,0 +1,416 @@
...
+
+        # Flip y axis for sane drawing
+        matrix = QtGui.QMatrix()
+        matrix.scale(1, -1)
+        self.view.setMatrix(matrix)
+
+        # self.view.setAlignment(Qt.AlignBottom | Qt.AlignHCenter)
+        self.view.setVerticalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
+        self.view.setHorizontalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
+        self.view.setRenderHint(QtGui.QPainter.Antialiasing)
```
kernc: If these are static properties of QGraphicsView, you should set them as kwargs to the constructor. See: https://github.com/biolab/orange3/blob/master/CONTRIBUTING.md#coding-style

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -0,0 +1,416 @@
...
+from Orange.widgets import gui
+from Orange.widgets.widget import OWWidget
+from Orange.widgets.settings import Setting
+from Orange.classification.tree import TreeClassifier
+from Orange.data.table import Table
+
+from PyQt4 import QtGui, QtCore
+from PyQt4.QtCore import Qt
+
+PI = math.pi
```
kernc: You seem to define this but then later use `math.pi` as well. If you import everything like
```py
from math import pi as PI, sin, cos, radians
```
you don't need this line anymore, get shorter lines and should also be able to shave some 10ns from each math call. :smile:

Orange/widgets/visualize/owpythagorastree.py
zidarsk8: This is just FYI, but here's what happens with mutable default vars:
```
In [1]: def hello(world=[]):
   ...:     world.append("omg")
   ...:     print(world)
   ...:     

In [2]: hello()
['omg']

In [3]: hello()
['omg', 'omg']
```
This means that even with tupples, if the elements in the tupple are mutable, you should not use that.

Orange/widgets/visualize/owpythagorastree.py
pavlin-policar: This sped up the function around 34 times. Very noticeable with larger datasets e.g. adult.

kernc:
The lab agreed you should open this PR against Orange3-Prototypes as well so it can be merged ASAP (no need to close this one, though). Sorry for misleading you. :grinning: 

To maintain history, you can move branch commits across repos with `git-cherry-pick` or `git-format-patch` & `git apply`.

pavlin-policar:
It's no problem. I've opened a PR in the prototype repo.

BlazZupan:
Moved and extended at prototypes.

PR #1152, opened 2016-04-01
Replace "horizontal" and "vertical" with Qt.Horizontal and Qt.Vertical

janezd:
Some of us have endured `"horizontal"` and `"vertical"` for more than a decade. I've had enough.

Let us use `Qt.Horizontal` and `Qt.Vertical` instead. This requires an additional import unless `Qt` is already imported, but that's cheap. String literals still work, but I still volunteer to patch all widgets. (Happy rebasing, everybody.)

@kernc, @astaric, please agree.

### DIFF
--- a/Orange/widgets/classify/owsvmclassification.py
+++ b/Orange/widgets/classify/owsvmclassification.py
@@ -33,8 +33,7 @@ class OWBaseSVM(OWBaseLearner):
         # Initialize with the widest label to measure max width
         self.kernel_eq = self.kernels[-1][1]
 
-        self.kernel_box = box = gui.widgetBox(
-            self.controlArea, "Kernel", orientation="horizontal")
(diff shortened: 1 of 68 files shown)


janezd:
Also, add deprecate warning for strings

astaric:
widget.py:201, widget.py:239

Orange/widgets/classify/owsvmclassification.py
```diff
@@ -65,8 +64,8 @@ def _add_kernel_box(self):
         box.setMinimumWidth(box.sizeHint().width())
 
     def _add_optimization_box(self):
-        self.optimization_box = gui.widgetBox(self.controlArea,
-                                              "Optimization parameters")
+        self.optimization_box = gui.vBox(
+                self.controlArea, "Optimization parameters")
```
astaric: 8 spaces?

PR #1154, opened 2016-04-02
OWSVMClassification: minor gui fix (change the width of tol widget)

s-alexey:
Tol spin widget doesn't display value properly.

### DIFF
--- a/Orange/widgets/classify/owsvmclassification.py
+++ b/Orange/widgets/classify/owsvmclassification.py
@@ -74,9 +74,9 @@ class SVMBaseMixin(OWProvidesLearner):
                 self.controlArea, "Optimization parameters")
 
         gui.doubleSpin(
-                self.optimization_box, self, "tol", 1e-7, 1.0, 5e-7,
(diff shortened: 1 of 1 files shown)


Orange/widgets/classify/owsvmclassification.py
```diff
@@ -175,9 +175,9 @@ def _add_type_box(self):
 
     def _add_optimization_box(self):
         super()._add_optimization_box()
-        gui.spin(self.optimization_box, self, "max_iter", 0, 1e6, 100,
+        gui.spin(self.optimization_box, self, "max_iter", 1, 1e6, 100,
```
janezd: 1 is not a good lower limit: if the user goes down and up again, he has 101, 201 ... I would suggest 50 as the lower limit and the step, that is, 

    gui.spin(self.optimization_box, self, "max_iter", 50, 1e6, 50,

Alternative would be 20 as the minimum and the step.

s-alexey: Thanks, I have changed it.

PR #1157, opened 2016-04-05
[FIX] Feature constructor Python 3.5 ast support

ales-erjavec:
* Fix an error in Feature Constructor widget due to a change in Python 3.5 ast (`AttributeError: 'Call' object has no attribute 'starargs'`)
* Add tests
* Some other minor fixes and code cleanup

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -583,6 +583,25 @@ import ast
 
 
 def freevars(exp, env):
+    """
+    Return names of all free variables in a parsed (expression) AST.
+
+    Parameters
+    ----------
+    exp : ast.AST
+        An expression ast (ast.parse(..., mode="single"))
+    env : List[str]
+        Environment
+
+    Returns
+    -------
+    freevars : List[str]
+
+    See also
+    --------
+    ast
+
+    """
     etype = type(exp)
     if etype in [ast.Expr, ast.Expression]:
         return freevars(exp.body, env)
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owfeatureconstructor.py
```diff
@@ -541,8 +524,14 @@ def apply(self):
 
         desc = list(self.featuremodel)
 
+        def validate(source):
+            try:
+                return validate_exp(ast.parse(source, mode="eval"))
```
kernc: Could't we just eval the expression and see if it fails with SyntaxError while the rest is on the user?

Orange/widgets/data/owfeatureconstructor.py
```diff
@@ -663,6 +652,68 @@ def freevars(exp, env):
         raise ValueError(exp)
 
 
+def validate_exp(exp):
+    """
+    Validate an `ast.AST` expression.
+
+    Only expressions with no list,set,dict,generator comprehensions
```
kernc: I've been meaning to introduce a ListVariable type. Seems like I wouldn't be able to populate it with feature constructor?

kernc: I see other uses cases where the expression evaluated to a list of values would be useful (like setting the values of all instances at once (though for that I'd also need to be given all instances at once)). How might constructing a `np.diff` work?

ales-erjavec: > I've been meaning to introduce a ListVariable type. Seems like I wouldn't be able to populate it with feature constructor?

No, unless you enable/change the accepted syntax for your variable type.

> I see other uses cases where the expression evaluated to a list of values would be useful (like setting the values of all instances at once (though for that I'd also need to be given all instances at once))

That should be done by a vectorized expression, which is an argument for limiting the syntax further.

Issue #1052, opened 2016-03-03
Table: copy method


s-alexey:
Current implementation looks quite strange:
```python
    def copy(self):
        """
        Return a copy of the table
        """
        t = self.__class__(self)
        t.ensure_copy()
        return t

    def ensure_copy(self):
        """
        Ensure that the table owns its data; copy arrays when necessary
        """
        if self.X.base is not None:
            self.X = self.X.copy()
        if self._Y.base is not None:
            self._Y = self._Y.copy()
        if self.metas.base is not None:
            self.metas = self.metas.copy()
        if self.W.base is not None:
            self.W = self.W.copy()
``` 

`self.__class__(self)` call runs init method with self as a second (typicaly X) argument. This isn't desired behavior. 

**Possible solutions**: use `copy` module
```python
import copy 

...
    def copy(self):
        """
        Return a copy of the table
        """
        return copy.deepcopy(self)
```

or just change to `self.__class__.__init__(self)` 

I can make PL, but I don't know what is the best solution.

kernc:
> Current implementation looks quite strange

I agree.

> I don't know what is the best solution.

I agree.

Does anything break?

s-alexey:
Curent

s-alexey:
Current Text-Mining add-on implementation use it in Corpus class and it leads to some problems there.

kernc:
Indeed. Do Orange's unit tests (`python ./setup.py test`) still pass? If they do, please submit a pull request.

janezd:
Although it sounds reasonable and obvious, I'd be careful with this. @astaric and @lanzagar have to confirm that this change has no side effects. I may be the original author of this code (or perhaps it's @astaric), but I can't remember why it's written like this. It can be a blunder, or it may be intentional.

Do numpy array implement deep copy efficiently? (I just quickly checked - it seems that they have a `__deepcopy__` method.)

s-alexey:
I'm not sure whether deepcopy is the best solution. 

```python
t = self.__class__(self)
```
here we create a new object with self as a second argument. It's dangerous.

Here a simple example of problems that may occur:
```python 
class Foo(object):
    def __init__(self, a, b):
        self.a = a
        
        if isinstance(a, Foo):
            raise ValueError("a can't be Foo inctance")

        self.b = b
        
        # in additional it may take a lot of time
        self.c = hard_computation_problem(a, b)
    
    def copy(self):
        # here a = self, b is not provided
        # so it raises a TypeError
[...]

PR #1158, opened 2016-04-06
Table: fix indexing and copy

s-alexey:
Refs #1052. 
New indices for every new table.
Copy domain in Table.copy.

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -330,10 +330,7 @@ class Table(MutableSequence, Storage):
             else:
                 self.W = np.empty((n_rows, 0))
             self.name = getattr(source, 'name', '')
-            if hasattr(source, 'ids'):
-                self.ids = np.array(source.ids[row_indices])
-            else:
-                cls._init_ids(self)
+            cls._init_ids(self)
             cls.conversion_cache[(id(domain), id(source))] = self
             return self
         finally:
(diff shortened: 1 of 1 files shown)


Orange/data/table.py
```diff
@@ -963,6 +959,8 @@ def copy(self):
         Return a copy of the table
         """
         t = self.__class__(self)
+        t.domain = Domain(t.domain.attributes, t.domain.class_vars,
+                          t.domain.metas, getattr(t.domain, 'source', None))
```
s-alexey: Or should I use `copy.copy(t.domain)`?

Orange/data/table.py
```diff
@@ -11,7 +11,6 @@
 from tempfile import NamedTemporaryFile
 from urllib.parse import urlparse, unquote as urlunquote
 from urllib.request import urlopen
-from urllib.error import URLError
```
kernc: I think this might have been meant to be caught in the `from_url` method. :worried:

s-alexey: But it's unused (caught outside `from_url` in `OWFile`). 
Actually with `from .instance import *` it's a mess here. This import shadows `chain`, `Real` and `Integral` and imports `numpy`.

kernc: > But it's unused

It just might have been meant for that. Indeed, doesn't really hurt if we just let urllib complain more verbosely than we would probably know to.

> it's a mess here.

astaric:
I am not really sure what you are trying to accomplish here.

We tend to consider domain as immutable, any modifications are usually done by constructing a new domain. (i.e. rearranging columns, moving variables from metas to attributes, ...). This is why it was not copied in Table.copy.

Ids work in a similar way. They allow us to use rows from a discretised, filtered table as selections on original data (i.e. using ScatterPlot). Generating new ids for each selection would break this feature.

Issue #1134, opened 2016-03-23
Save Graph doesn't save axes


ajdapretnar:
Save Graph functionality doesn't save axes properly (name and scale of axis). Also png files are saved on the basis of current window size and thus become difficult to scale if saved with small window.

kumargaurav2596:
I fixed the scaling issue by passing the dimensions of Qgraphicsview instead of scene to create the buffer and then scaled the the  painter in _get_target() function in orange/widgets/io.py , as show below

```
def write_image(cls, filename, scene):
        try:
            exporter = cls._get_exporter()
            cls._export(exporter(scene), filename)
        except Exception:
            if isinstance(scene,  QGraphicsView):
                rect = scene.sceneRect()
            if isinstance(scene, QGraphicsScene):
                rect = scene.views()[0].sceneRect()
            elif isinstance(scene, QWidget):
                rect = scene.rect()
            rect = rect.adjusted(-15, -15, 15, 15)
            buffer = cls._get_buffer(rect.size(), filename)
```
Now the scaling of painter in the _get_target()
```
class PngFormat(ImgFormat):
    def _get_buffer(size, filename):
        return QtGui.QPixmap(400, 300)

    @staticmethod
    def _get_target(scene, painter, buffer, source):
        try:
            brush = scene.backgroundBrush()
            if brush.style() == QtCore.Qt.NoBrush:
                brush = QtGui.QBrush(scene.palette().color(QtGui.QPalette.Base))
        except AttributeError:  # not a QGraphicsView/Scene
            brush = QtGui.QBrush(QtCore.Qt.white)
        painter.fillRect(buffer.rect(), brush)

        painter.scale(( 400 * ( source.height() / 300)) / source.width(), 1)
        return QtCore.QRectF(0, 0, 400, 300)
```


But as you can see using painter is not at priority, instead it is the exporter that tries first.

With exporter I tried 
`exporter.parameters()['width'] = 400;`

As long as the window size is larger than 400 it is working perfectly keeping the aspect ratio but the graph starts to shrink ones the window size goes below 400.

Any help?

kernc:
Fixed in https://github.com/biolab/orange3/commit/6b57116a3d57ca16397b4cae5bf20e232c583e63

PR #1162, opened 2016-04-07
[FIX] Save Graph doesn't save axes #1134

kumargaurav2596:
Fixed "Save Graph doesn't save axes" issue #1134. The graphs are now saved with a white background. Scaling issue still exist.


astaric:
This PR cannot be applied onto the latest master. Please rebase.

Also, try saving the graph on the distributions widget. That is the one that is currently not saving correctly.

PR #1167, opened 2016-04-08
[FIX] Save Graph doesn't save axes #1134 (updated base)

kumargaurav2596:
Fixed "Save Graph doesn't save axes" issue #1134. The graphs are now saved with a white background. Scaling issue still exist.

### DIFF
--- a/Orange/widgets/io.py
+++ b/Orange/widgets/io.py
@@ -35,8 +35,16 @@ class ImgFormat(FileFormat):
     @classmethod
     def write_image(cls, filename, scene):
         try:
+            scene = scene.scene()
+            scenerect = scene.sceneRect()   #preserve scene bounding rectangle
+            viewrect = scene.views()[0].sceneRect()
+            scene.setSceneRect(viewrect)
+            backgroundbrush = scene.backgroundBrush()  #preserve scene background brush
+            scene.setBackgroundBrush(QtCore.Qt.white)
             exporter = cls._get_exporter()
             cls._export(exporter(scene), filename)
+            scene.setBackgroundBrush(backgroundbrush)  # reset scene background brush
+            scene.setSceneRect(scenerect)   # reset scene bounding rectangle
         except Exception:
             if isinstance(scene, (QGraphicsScene, QGraphicsView)):
                 rect = scene.sceneRect()


ajdapretnar:
@kumargaurav2596 I've tested out the fix and I really like how it's now on a white background and with axes properly displayed.
It's a good fix, but scaling remains problematic.
Any thoughts on that @kernc, @janezd?

kumargaurav2596:
For scaling problem I proposed a method, it has some minor issues , specially when window width is very small. For details please check my comment at https://github.com/biolab/orange3/issues/1134 

Also, please add a gsoc level to this pull request.

kernc:
> but scaling remains problematic.

I have nothing against WYSIWYG. If you need scalable graphics, opt for vector.

ajdapretnar:
@kernc Not scaling as such. I used it in relation to Kumar's 
> Scaling issue still exist.

but I guess I misunderstood what he had in mind here. What I'm referring to is this (huge white space above the actual plot, happens every time):

xtys

kumargaurav2596:
The white spaces are occurring in case views's width is first increased and then decreased, as the scene dimension do not shrink during resizing. Fixed this in commit https://github.com/biolab/orange3/pull/1167/commits/54550393fd299118adcd3ce236c2600336165cca

ajdapretnar:
@kumargaurav2596 Works well for me. :+1:

Orange/widgets/io.py
```diff
@@ -35,8 +35,16 @@ def _export(self, exporter, filename):
     @classmethod
     def write_image(cls, filename, scene):
         try:
+            scene = scene.scene()
+            scenerect = scene.sceneRect()   #preserve scene bounding rectangle
+            viewrect = scene.views()[0].sceneRect()
+            scene.setSceneRect(viewrect)
+            backgroundbrush = scene.backgroundBrush()  #preserve scene background brush
+            scene.setBackroundBrush(Qt.White)
```
kernc: Hmm... Is `Qt` imported somewhere?

kumargaurav2596: thank you for pointing that out, apparently  it was still working because the scene was already updated before the error occurred and hence we were getting the expected results. Fixed it...

astaric:
Works on OSX as well.

Can you squash the commits and rebase onto master so we can merge?

kumargaurav2596:
Squashed all the commits and rebased it onto master.

kernc:
Thanks!

PR #1171, opened 2016-04-11
[ENH] Add support for PyQt5

ales-erjavec:
Add full support for PyQt5


Orange/canvas/utils/settings.py
```diff
@@ -82,9 +79,9 @@ def qt_to_mapped_type(value):
         return value
 
 
-class QABCMeta(pyqtWrapperType, abc.ABCMeta):
+class QABCMeta(_QObjectType, abc.ABCMeta):
     def __init__(self, name, bases, attr_dict):
-        pyqtWrapperType.__init__(self, name, bases, attr_dict)
+        _QObjectType.__init__(self, name, bases, attr_dict)
         abc.ABCMeta.__init__(self, name, bases, attr_dict)
```
kernc: So ... what does this metaclass do?

ales-erjavec: https://docs.python.org/3.5/library/abc.html#abc.ABCMeta
It is the metaclass of the python's 'abstract' containers (`collections.Mapping`, `collection.Sequence`, ...), and needed here in order to properly subclass QObject and MutableMapping later on.

Orange/canvas/application/canvastooldock.py
```diff
@@ -87,33 +89,39 @@ def setSplitterAndWidget(self, splitter, widget):
...
+        if self.__widget is not None:
+            self.__widget.removeEventFilter(self)
+        if self.__splitter is not None:
+            self.__splitter.removeEventFilter(self)
+
         self.__splitter = splitter
         self.__widget = widget
 
-        if widget:
+        if widget is not None:
```
zidarsk8: Going with google code style, it's beter to use implicit false here. See  https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

Orange/canvas/application/canvastooldock.py
```diff
@@ -183,6 +198,11 @@ def eventFilter(self, obj, event):
                 self.__action.setChecked(True)
                 self.__expanded = True
 
+        if obj is self.__splitter and event.type() == QEvent.Show:
+            if self.__updateOnShow:
```
zidarsk8: Why is this condition not just a part of the previous if statement?

Orange/canvas/application/canvastooldock.py
astaric: I am not sure how the document relates to the preference. The document explicitly states:

`Beware of writing if x: when you really mean if x is not None`

Orange/widgets/utils/webview.py
```diff
@@ -0,0 +1,122 @@
+import textwrap
```
kernc: Why not replace Orange/widgets/webview.py with this module, and have an additional WebView class that extends one or the other?

Orange/canvas/report/owreport.py
```diff
@@ -301,9 +328,16 @@ def save_report(self):
...
+            page = self.report_view.page()
+
+            def save_html(contents):
+                with open(filename, "w", encoding="utf-8") as f:
+                    f.write(contents)
+
+            if webview.HAVE_WEBENGINE:
+                page.toHtml(save_html)
+            else:
+                save_html(page.mainFrame().toHtml())
```
kernc: Isn't a minimal compatible wrapping interface at one single place better than "ifdefs"  all around the codebase?

astaric:
I would like to merge this PR and start providing nightly builds that use PyQt5.

Could you rebase onto the latest master (for the final time)?

Orange/canvas/__main__.py
```diff
@@ -163,6 +169,9 @@ def main(argv=None):
 
     qt_argv += args
 
+    # Add --disable-web-security to enable Qt WebEngine to load local resources
+    qt_argv.append('--disable-web-security')
```
ales-erjavec: @kernc Why is this necessary?

Qt WebEngine can (or at least should be able to) load local resources for locally originating content. At lest that is what the documentation for [QWebEnginePage](http://doc.qt.io/qt-5/qwebenginepage.html) says:
> ... By default, local schemes like file:// and qrc:// are considered to be in the same security origin, and can access each other's resources

kernc: Yes, that appears to not have been needed, thanks.

kernc:
Please give it another rebase just to trigger CI.

PR #1174, opened 2016-04-12
OWFile: Use sample datasets on first use

astaric:


Orange/widgets/data/owfile.py
```diff
@@ -80,6 +80,13 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
     LOCAL_FILE, URL = range(2)
 
     settingsHandler = XlsContextHandler()
+
+    # Overload RecentPathsWidgetMixin.recent_paths to set defaults
+    recent_paths = Setting([
+        RecentPath("","sample-datasets", "iris.tab"),
+        RecentPath("","sample-datasets", "titanic.tab"),
+        RecentPath("","sample-datasets", "housing.tab"),
```
kernc: Pylint issues three true-positive bad-whitespace warnings.

PR #1175, opened 2016-04-12
OWWidget: Do not restore widget geometry into full screen mode

ales-erjavec:
~~Possibly~~ Fix a problem where a widget is restored into full screen mode
with no obvious way to exit.

### DIFF
--- a/Orange/widgets/utils/__init__.py
+++ b/Orange/widgets/utils/__init__.py
@@ -33,3 +33,180 @@ def getdeepattr(obj, attr, *arg, **kwarg):
 
 def getHtmlCompatibleString(strVal):
     return strVal.replace("<=", "&#8804;").replace(">=","&#8805;").replace("<", "&#60;").replace(">","&#62;").replace("=\\=", "&#8800;")
+
+
+# Utilities for inspecting/modifying widget geometry state as returned/set by
+# QWidget's saveGeometry and restoreGeometry
+
+import struct
+from collections import namedtuple
+
+#: Parsed geometry state (version 1; Qt < 5.4)
+_geom_state_v1 = namedtuple(
+    "geom_state",
+    ["magic",   # int32_t magic constant
+     "major_version", "minor_version",  # (uint16_t, uint16_t) format version
+     #: 4x int32_t ( == frameGeometry().getCoords())
+     "frame_left", "frame_top", "frame_right", "frame_bottom",
+     #: 4x int32_t ( == geometry().getCoords() WHEN NOT in FullScreen)
+     "normal_left", "normal_top", "normal_right", "normal_bottom",
+     "screen",      # screen number
+     "maximized",   # windowState() & Qt.WindowMaximized
+     "full_screen"  # windowState() & Qt.WindowFullScreen
+     ]
+)
+
+
+#: Parsed geometry state (version 2; Qt >= 5.4)
+_geom_state_v2 = namedtuple(
+    "geom_state",
+    _geom_state_v1._fields +
+    ("screen_width", )  # int32_t screen's width
+)
+
+
+def geom_state(magic, major_version, *args, **kwargs):
+    if magic != geom_state.MAGIC:
+        raise ValueError
+
+    if major_version == 1:
+        return geom_state.v1(magic, major_version, *args, **kwargs)
+    elif major_version == 2:
+        return geom_state.v2(magic, major_version, *args, **kwargs)
+    else:
+        raise ValueError
+
+geom_state.MAGIC = 0x1D9D0CB
+geom_state.v1 = _geom_state_v1
+geom_state.v2 = _geom_state_v2
+
+
+def geom_state_from_bytes(state):
+    """
+    Parse the QWidget.saveGeometry return value
+
+    Parameters
+    ----------
+    state : bytes
+        Saved widget geometry state as returned by :func:`QWidget.saveGeometry`
+
+    Returns
+    -------
+    parsed : namedtuple
+        The parsed geometry state as a named tuple
+    """
+    state = bytes(state)
+    MAGIC = 0x1D9D0CB
+    header_fmt = (
+        "I"   # magic number
+        "HH"  # minor.major format version
+    )
+    header_len = struct.calcsize(">" + header_fmt)
+    magic, major, minor = struct.unpack(">" + header_fmt, state[:header_len])
+    if magic != MAGIC:
+        raise ValueError("Magic value does not match")
+    if major not in {1, 2}:
+        raise ValueError("Do not know how to handle version {}".format(major))
+
+    payload_fmt = (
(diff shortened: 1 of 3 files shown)


ales-erjavec:
@ajdapretnar Please check if this fixes the problem

ajdapretnar:
This fix works for me!

kernc:
I can assure as much that I didn't have the window in full-screen mode, otherwise I wouldn't be able to save the scheme. Is only the widgets's geometry rect saved or any other flags as well?

This is xwininfo result for a random Orange widget window.
```
% xwininfo -all

xwininfo: Please select the window about which you
          would like information by clicking the
          mouse in that window.

xwininfo: Window id: 0x560008c "Preprocess"

  Root window id: 0xab (the root window) "i3"
  Parent window id: 0x1204379 "[i3 con] container around 0x1a09790"
     0 children.

  Absolute upper-left X:  641
  Absolute upper-left Y:  37
  Relative upper-left X:  1
  Relative upper-left Y:  0
  Width: 638
  Height: 1042
  Depth: 24
  Visual: 0x20
  Visual Class: TrueColor
  Border width: 0
  Class: InputOutput
  Colormap: 0x22 (installed)
  Bit Gravity State: NorthWestGravity
  Window Gravity State: NorthWestGravity
  Backing Store State: NotUseful
  Save Under State: no
  Map State: IsViewable
  Override Redirect State: no
  Corners:  +641+37  -641+37  -641-1  +641-1
  -geometry 638x1042+640-0

  Bit gravity: NorthWestGravity
  Window gravity: NorthWestGravity
  Backing-store hint: NotUseful
  Backing-planes to be preserved: 0xffffffff
  Backing pixel: 0
  Save-unders: No

  Someone wants these events:
      KeyPress
      KeyRelease
      ButtonPress
      ButtonRelease
      EnterWindow
      LeaveWindow
      PointerMotion
      ButtonMotion
      KeymapState
      Exposure
      StructureNotify
      FocusChange
      PropertyChange
  Do not propagate these events:
  Override redirection?: No

  Window manager hints:
      Client accepts input or input focus: Yes
      Initial state is Normal State
      Window type:
          Normal
      Process id: 24006 on host zoob

  Normal window size hints:
      User supplied location: 370, 236
      Program supplied location: 370, 236
      User supplied size: 589 by 492
      Program supplied size: 589 by 492
      Program supplied minimum size: 578 by 315
      Program supplied window gravity: NorthWestGravity
  No zoom window size hints defined

  No window shape defined
  No border shape defined
```
Might be, that since the reported geometry is `638x1042+640-0`, Windos positions the widget's visible area at y = 0, meaning window decorations are clipped off the screen above it.

kernc:
Why do we save the geometry anyway; seems prone to such DoS. If I open a window on a 4K screen in the bottom right corner, where does Qt reopen it on a 800x600 display?

ales-erjavec:
> Is only the widgets's geometry rect saved or any other flags as well?

The widget's own geometry, the window frame geometry, screen number and maximized/fullscreen state.

> Windos positions the widget's visible area at y = 0, meaning window decorations are clipped off the screen above it.

No. For top level widgets (i.e. windows) the positioning includes the window decoration (ie. moving a window to (0, 0), positions the window frame at the top left screen corner (with the title bar and frame visible). Also that would not hide the taskbar as happened in this case.

In any case the observed behavior was identical to 
```
from PyQt4.QtGui import QApplication, QDialog
from PyQt4.QtCore import Qt
app = QApplication([])
w = QDialog()
w.showFullScreen()
app.exec_()
```

> Why do we save the geometry anyway; seems prone to such DoS. If I open a window on a 4K screen in the bottom right corner, where does Qt reopen it on a 800x600 display?

It positions the widget inside screen's the [available geometry](http://doc.qt.io/qt-4.8/qdesktopwidget.html#availableGeometry-2)

kernc:
> Also that would not hide the taskbar as happened in this case.

If you mean window title decorations, then there is a reasonable possibility that I had the table widget full screen and then killed it with Escape. In that case, if full-screen mode is indeed the culprit, [disabling `WindowFullScreen` flag](https://doc.qt.io/qt-4.8/qwidget.html#setWindowState) in `__restoreWidgetGeometry()` might work?

ales-erjavec:
I have changed the fix. The `__restoreWidgetGeometry` now parses the stored geometry state, modifying it so that it is restored into a maximized state instead (another option would be to restore it into 'normal' mode with (approximately) equal geometry it had before entering full screen).

Orange/widgets/widget.py
```diff
@@ -329,11 +329,27 @@ def setStatusBarText(self, text, timeout=5000):
...
+                    pass
+                else:
+                    if state.full_screen:
+                        # remove the full screen state replacing it with
+                        # maximized state instead
+                        state = state._replace(
+                            full_screen=0, maximized=Qt.WindowMaximized
+                        )
+                        geometry = state.to_bytes()
                 restored = self.restoreGeometry(QByteArray(geometry))
```
kernc: The other two changed files above surely must raise the question:

Why didn't you settle with something like:
```py
self.setWindowState(self.windowState() & ~Qt.WindowFullScreen)
geom, avail = self.geometry(), qApp.desktop().availableGeometry(self)
if geom.width() > avail.width() and geom.height() > avail.height():
    self.setWindowState(self.windowState() | Qt.WindowMaximized)
```
The docs say:
> If the window is not visible, the window state will take effect when show() is called. For visible windows, the change is immediate.

ales-erjavec: The geometry is being restored from `showEvent`. That event is a result of `show` already been (or maybe being?) called. In `__restoreWidgetGeometry()` `self.isVisible()` is already `True`

kernc: But for visible windows, the change is immediate? I'd want proof flicker is noticeable or problematic. :stuck_out_tongue:

Orange/widgets/utils/__init__.py
```diff
@@ -33,3 +33,180 @@ def getdeepattr(obj, attr, *arg, **kwarg):
...
+geom_state.from_bytes = geom_state_from_bytes
+geom_state.v1.to_bytes = geom_state_to_bytes
+geom_state.v2.to_bytes = geom_state_to_bytes
+
+
+from PyQt4.QtCore import Qt, QRect, QMargins
+from PyQt4.QtGui import QWidget
+
+
+def geom_state_normal(state, widget_or_margins=None):
```
astaric: is this used anywhere? (beside tests)

ales-erjavec: No.

janezd:
This PR is much cleaner than #1183, which constructs a new widget just to reset the geometry. But both seems like they could fail on some future version of Qt, so I decided for #1183 just because it is shorter and probably easier to maintain.

PR #1176, opened 2016-04-13
[ENH] OWFile: Open dragged files

astaric:
Allow files to be opened by dragging them to OWFile window. 

Does not work on OSX10.10+ with Qt4 (see https://bugreports.qt.io/browse/QTBUG-40449).
When applied on top of #1171, it should work as expected.

Could use some testing on other platforms.

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -193,6 +193,8 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
         # explicitly re-enters the event loop (by a progress bar)
         QtCore.QTimer.singleShot(0, self.load_data)
 
+        self.setAcceptDrops(True)
+
     def reload(self):
         if self.recent_paths:
             basename = self.file_combo.currentText()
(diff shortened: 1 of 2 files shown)


lanzagar:
Works for me with Arch linux, dwm, pcmanfm

ajdapretnar:
Works on Win8.

Orange/widgets/data/owfile.py
```diff
@@ -404,6 +406,30 @@ def get_ext_name(filename):
...
+
+        if event.mimeData().hasUrls():
+            # On OSX Yosemite and later, dragged files are passed
+            # as file references (file://.file/id=... urls).
+            # As PyQt4 toLocalFile does not resolve such urls,
+            # we do not allow them to be dropped on widget.
+            if _is_osx_file_reference_url():
+                return
+
+            event.acceptProposedAction()
```
kernc: Before accepting, it should probably ensure there exists a reader for the associated file. Otherwise if breaks at least on ZIP archives when I press _Reload_.

Orange/widgets/data/owfile.py
```diff
@@ -404,6 +406,25 @@ def get_ext_name(filename):
...
+                event.acceptProposedAction()
+            break
+
+    def dropEvent(self, event):
+        """Handle file drops"""
+        for url in event.mimeData().urls():
+            self.add_path(url.toLocalFile())
+            self.source = self.LOCAL_FILE
+            self.load_data()
+            break
```
lanzagar: I vote for the more understandable:

```python
urls = event.mimeData().urls()
if urls:
    self.add_path(urls[0].toLocalFile())  # add first file
    self.source = self.LOCAL_FILE
    self.load_data()
```

PR #1177, opened 2016-04-13
Refactor data.io

astaric:
Current singleton implementation of file readers/writers is hard to extend when additional features are needed (such as selecting sheet for excel files and overwriting header in #1012).

This PR replaces
    FileFormat.read(filename)
with
    FileFormat.get_reader(filename).read()

The new way enables us to set additional properties on reader before the read method is called.

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -210,7 +210,7 @@ class FileFormat(metaclass=FileFormatMeta):
 
         def read(self):
             ...  # load headers, data, ...
-            return wrapper(self.data_table(data, headers))
+            return self.data_table(data, headers)
 
         @classmethod
         def write_file(cls, filename, data):
@@ -219,8 +219,7 @@ class FileFormat(metaclass=FileFormatMeta):
             writer.writerows(data)
 
     Wrapper FileFormat.data_table() returns Orange.data.Table from `data`
-    iterable (list (rows) of lists of values (cols)). `wrapper` is the
-    desired output class (if other than Table).
+    iterable (list (rows) of lists of values (cols)).
     """
 
     PRIORITY = 10000  # Sort order in OWSave widget combo box, lower is better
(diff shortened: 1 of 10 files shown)


Orange/data/io.py
```diff
@@ -626,43 +651,46 @@ class BasketFormat(FileFormat):
...
 
-    @classmethod
-    def read_file(cls, filename, storage_class=None):
-        from Orange.data import _io, Table, Domain
-        import sys
-        if storage_class is None:
-            storage_class = Table
+    def read(self):
+        if self.wrapper is None or self.wrapper is _IDENTITY:
+            self.wrapper = Table
```
kernc: I think we should consider throwing these wrappers out. Is `wrapper` ever not `Table`? If it isn't then the caller is obviously aware of it and can invoke `wrapper(reader.read())` on their end.

Orange/data/table.py
```diff
@@ -517,7 +517,10 @@ def from_file(cls, filename, wrapper=None):
 
         if not os.path.exists(absolute_filename):
             raise IOError('File "{}" was not found.'.format(filename))
-        data = FileFormat.read(absolute_filename, wrapper or cls)
+        reader = FileFormat.get_reader(absolute_filename)
+        if wrapper or cls != Table:
+            reader.set_wrapper(wrapper or cls)
+        data = reader.read()
```
kernc: Aware ... like ... here.

Orange/data/table.py
```diff
@@ -500,24 +500,12 @@ def from_file(cls, filename, wrapper=None):
...
-            absolute_filename = ext = ""
 
-        if not os.path.exists(absolute_filename):
-            raise IOError('File "{}" was not found.'.format(filename))
-        data = FileFormat.read(absolute_filename, wrapper or cls)
+        absolute_filename = FileFormat.locate(filename, dataset_dirs)
+        reader = FileFormat.get_reader(absolute_filename)
+        if wrapper or cls != Table:
+            reader.set_wrapper(wrapper or cls)
+        data = reader.read()
```
kernc: Aware ... like ... here.

Orange/data/io.py
astaric: I agree. See ce70666

PR #1183, opened 2016-04-15
OWWidget: restore fullscreen into maximized instead

kernc:
Alternative to https://github.com/biolab/orange3/pull/1175, works beautifully on Windos 10.


astaric:
This causes noticable (and problematic) flicker on OSX (or should I start calling it macOS)

Issue #1143, opened 2016-03-30
Icons missing in Orange Canvas


MEL27:
##### Orange version

3.3.1.dev0+5d39ea2

##### Expected bahavior
Icons showing up.

##### Actual behavior
Icons didn't show up, except the ones in Help.

##### Steps to reproduce the behavior
Launch orange-canvas

##### Additional info (worksheets, data, screenshots, ...)

OS: Windows 10 Home
Orange installed from installer from official website

![screenshot](https://cloud.githubusercontent.com/assets/10842148/14145440/6f91884a-f6c6-11e5-90b2-ad5b0020101c.png)

ajdapretnar:
@MEL27 Thank you very much for such precise description!
I'm running the same version on Win8 and am not experiencing the same issue.
Perhaps @thocevar can check for Win10?

kernc:
Does this help?

https://stackoverflow.com/questions/34629635/orange-canvas-does-not-show-any-icon/

MEL27:
@kernc
Yes, it does. After I created a qt.conf according to the instruction, the problem is solved.
I didn't see the answer when I searched. Thanks for the link!

MEL27:
Instead of adding a qt.conf, using `QtCore.QCoreApplication.addLibraryPath('./Lib/site-packages/PyQt4/plugins')` may be a better solution. [(from stackoverflow)](http://stackoverflow.com/a/23525640)

kernc:
Thanks!

@ales-erjavec, can we use this?

ajdapretnar:
@kernc @ales-erjavec Are we using the above fix? Perhaps @MEL27 you could submit a PR?

MEL27:
I've submitted a pull request #1194 . The hack is dirty given that we still don't know what exactly happens under the hood. But yeah it works.

kernc:
Should be fixed now. Thanks again!

PR #1194, opened 2016-04-20
Manually specify plugins path

MEL27:
A hack for issue #1143, where Qt failed to find the correct path for SVG plugin.

### DIFF
--- a/Orange/canvas/config.py
+++ b/Orange/canvas/config.py
@@ -43,6 +43,9 @@ def init():
     QCoreApplication.setApplicationName("Orange Canvas")
     QCoreApplication.setApplicationVersion(version)
     QSettings.setDefaultFormat(QSettings.IniFormat)
+    
+    # Hack to make sure the correct plugin search path is added. (Github issue #1143)
+    QCoreApplication.addLibraryPath('./Lib/site-packages/PyQt4/plugins')
 
     # Make it a null op.
     global init
--- a/Orange/canvas/config.py
+++ b/Orange/canvas/config.py
@@ -5,6 +5,7 @@ Orange Canvas Configuration
 
 import os
 import sys
+import site
 import logging
 import pickle as pickle
 import itertools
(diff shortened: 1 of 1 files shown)


Orange/canvas/config.py
```diff
@@ -44,6 +45,11 @@ def init():
     QCoreApplication.setApplicationVersion(version)
     QSettings.setDefaultFormat(QSettings.IniFormat)
 
+    # Hack to make sure the correct plugin search path is added. (Github issue #1143)
+    for path in site.getsitepackages():
+        if path.endswith("site-packages"):
+            QCoreApplication.addLibraryPath(path + "\\PyQt4\\plugins")
```
kernc: Does this work with POSIX paths on Mac OS X and GNU/Linux? Perhaps better to use `os.path.join()`.

Orange/canvas/config.py
```diff
@@ -44,6 +45,11 @@ def init():
     QCoreApplication.setApplicationVersion(version)
     QSettings.setDefaultFormat(QSettings.IniFormat)
 
+    # Hack to make sure the correct plugin search path is added. (Github issue #1143)
+    for path in site.getsitepackages():
+        if path.endswith("site-packages"):
+            QCoreApplication.addLibraryPath(os.path.join(path, "PyQt4", "plugins"))
```
kernc: It's not ok to simply set the default library path and effectively override any user's set preferred path, if one exists. It would probably be better to _extend_ the library path with the one in site-packages. I think you can do it like:
```py
QCoreApplication.setLibraryPaths(
    QCoreApplication.libraryPaths() + [os.path.join(path, "PyQt4", "plugins")])
```
Does this work?

kernc:
Sorry for the late review. This is great and should help us tremendously with Windos users. I hope it's not too late for you to consider my above suggestion?

Orange/canvas/config.py
MEL27: @kernc It works. Did you mean that we should append the path to the end instead of the front? I am not so sure but it sounds reasonable.

kernc: > we should append the path to the end instead of the front?

Exactly, we get to honor any user overrides, but if our requirements aren't met in any of the already set paths, the fallback default will fill them in.

MEL27:
@kernc see https://github.com/biolab/orange3/pull/1194/commits/56d79acfbc5f58f206ea54ac1a9f1efeb53a8200.

MEL27:
It's been a while. I just want to ask, will it be used?

kernc:
Oh, it _definitely_ will! This is really nitpicking and just my personal preference, but would you consider using a list comprehension for the iteration? :smiley: That way, it's more evident that the loop and branching don't drive program flow but just set the correct value; it all reduces to a single statement (calling of Qt's function) with a single value (the amended list of paths). I'm also happy to merge as-is.

Issue #1036, opened 2016-02-29
Widgets don't retain settings after worksheet reload


gcatalfamo:
Whenever I reopen a workflow I have to to reconfigure every widget (not recreate, just reconfigure) because they get messed up after reload.
It doesn't happen after every single reload but I can't track down what makes it happen.

kernc:
What widgets are those?

gcatalfamo:
I use mainly data and visualization widgets. The one I am certain the problem is happening are:
- DATA, select rows
- DATA, feature constructor

The widgets seems to reset:
- feature constructor is empty
- select rows offers just one condition (regardless of how many there were before) with the first available option selected

I have used Orange3.2 on mac with several snapshots of the last 2 weeks.

kernc:
I can't seem to reproduce. If you open [this worksheet](https://dl.dropboxusercontent.com/u/8775166/untitled.ows), the preferences are cleared?

gcatalfamo:
The preference seems to be fine.

I found a way to reproduce it: if you reload a second dataset (that has the same features), widget will reset.

ajdapretnar:
@gcatalfamo Could you please let us know if the issue persists with the latest build?

kernc:
I made a subset of iris with Select Rows and saved it to iris2.tab. I add another variable in feature constructor and save _that_ as iris3.

When setting File to load iris2, it works. With iris3, it doesn't. Simple replacing of `PerfectDomainContextHandler` with `DomainContextHandler` didn't work, obviously. The domain is just not the same (even if it includes all the relevant variables). @astaric What can we do?

astaric:
We can fix it :)

I'll prepare a PR.

astaric:
A fix and imperfect matching of domains for FeatureConstructor widget is implemented in #1196. Please see if the second part of the workflow above now works with changing datasets.

I haven't managed to check the Select Rows widget yet.

astaric:
After playing around with the Select Rows widget, I could not find any problems. The fact that it does not restore filters for similar datasets is by design, as using it on adult and adult-sample keeps the filters just fine.

If one were to use many different datasets, which share only few of the variables, trying to apply filters from one to the other might not be a wisest choice, as they will be meaningful. If I was observing a group of male germans on one dataset, why would I want to have this as a default filter on a different datasets which contains variables gender and nationality?

[...]

PR #1196, opened 2016-04-20
[ENH] Support context cloning in FeatureConstructor and SelectRows

astaric:
Fix a bug when changing dataset with descriptors in FeatureConstructor would cause an error.
Implement partial matching of contexts (feature constructed on adult_sample that uses variable age can also be used on titanic dataset)


astaric:
When settings from similar domain are used to initialize defaults, descriptors with names that would clash with existing variable names are not included in the list. https://github.com/biolab/orange3/issues/1036#issuecomment-212794395

PR #1199, opened 2016-04-21
[WIP] Add control for editing domains, include it in OWFile

janezd:
Rebased #1012.

The control (table view, corresponding models and delegates) have been moved from OWFile.py to a separate module and can be included in other widgets.

Contexts are broken: OWFile widget now has two separate contexts and we need to decide how to handle this. Since this is the first time it happened in a decade, a hack will suffice.

Apply button cannot be implemented before i/o refactoring.


janezd:
Continued in #1319.

PR #1201, opened 2016-04-22
[ENH] OWMergeData: merge by instance ID.

markotoplak:


lanzagar:
As discussed, the new option (renamed to "Same source" or sth?) could appear only when possible.
There could also be another option to merge by indices.

PR #1207, opened 2016-04-22
FilterContinuous: Add tests and fix bugs

janezd:


Orange/data/filter.py
```diff
@@ -351,28 +353,29 @@ def __call__(self, inst):
...
+            column = "feature({})".format(self.column)
+
+        names = {self.Equal: "=", self.NotEqual: "≠",
+                 self.Less: "<", self.LessEqual: "≤",
+                 self.Greater: ">", self.GreaterEqual: "≥"}
+        if self.oper in names:
+            return "{} {} {}".format(column, names[self.oper], self.ref)
         if self.oper == self.Between:
-            return "feature_%s <= %s <= %s" % (self.column, self.ref, self.max)
+            return "{} ≤ {} ≤ {}".format(self.min, self.column, self.max)
```
lanzagar: `column` instead of `self.column`

Orange/data/filter.py
```diff
@@ -351,28 +353,29 @@ def __call__(self, inst):
...
         if self.oper == self.Outside:
-            return "NOT %s <= feature_%s <= %s" % (self.column, self.ref, self.max)
+            return "not {} ≤ {} ≤ {}".format(self.min, self.column, self.max)
```
lanzagar: `column` instead of `self.column`

Orange/data/filter.py
```diff
@@ -351,28 +353,29 @@ def __call__(self, inst):
...
         if self.oper == self.IsDefined:
-            return "feature_%s IS DEFINED" % self.column
+            return "{} is defined".format(self.column)
```
lanzagar: `column` instead of `self.column`

PR #1209, opened 2016-04-22
Preprocess: Add number of intervals to discretize slider

rokgomiscek:
Preprocess: Add number of intervals to discretize slider


janezd:
I forgot to add a comment here. The position of the label is bad. I made a pull request to this pull request.

@rokgomiscek, can you move the label and make another PR?

PR #1210, opened 2016-04-22
SameValue filter tests

mstrazar:
Moved SameValue tests to test_filter.py
- Added test with Mock objects and tested attribute, value pairs so that coverage is 100 % for SameValue class.
- Added tests for combining two filters.
- Minor bugfix (Remove invalid attribute inst._values).


kernc:
Merged in https://github.com/biolab/orange3/pull/1251.

PR #1211, opened 2016-04-23
SklLearner: add supports_weights property.

s-alexey:
This flag was missed in some learners (e.g. LogisticRegression, RandomForest)

### DIFF
--- a/Orange/tests/test_base.py
+++ b/Orange/tests/test_base.py
@@ -1,0 +1,32 @@
+# Test methods with long descriptive names can omit docstrings
+# pylint: disable=missing-docstring
+import unittest
+
+from Orange.base import SklLearner
+from Orange.classification import LogisticRegressionLearner
+
+
+class TestSklLearner(unittest.TestCase):
+    def test_supports_weights(self):
+        class DummySklLearner:
+            def fit(self, X, y, sample_weight=None):
+                pass
+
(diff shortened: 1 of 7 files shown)


Orange/tests/test_base.py
```diff
@@ -0,0 +1,24 @@
...
+class TestSklLearner(unittest.TestCase):
+    def test_supports_weights(self):
+        class DummySklLearner:
+            def fit(self, X, y, sample_weight=None):
+                pass
+
+        class DummyLearner(SklLearner):
+            __wraps__ = DummySklLearner
+
+        self.assertTrue(DummyLearner().supports_weights)
```
kernc: Could we base this test on one real sklearn's learner, so it will fail when, if, they decide to change the `sample_weight` argument name.

Orange/tests/test_base.py
```diff
@@ -0,0 +1,24 @@
+import unittest
```
kernc: Please put somewhere at the top of the module:
```py
# Test methods with long descriptive names can omit docstrings
```
And if possible, make sure the [PyLint test](https://travis-ci.org/biolab/orange3/jobs/125246229#L555) passes.

s-alexey: Ok, I have made some changes and it passes PyLint test now.

PR #1212, opened 2016-04-24
[RFC] Attribute names in gui functions

janezd:
Widget has an attribute `my` whose `__getattr__` checks for existence of attribute and returns its name.

Intended use is illustrated in the code of `OWContinuize`.

    my = self.my  # type: OWContinuize

This pulls `my` into `__init__`'s namespace and tricks PyCharm to provide completion. Later we use, e.g.

     gui.radioButtonsInBox(box, self, my.multinomial_treatment, btnLabels=labels)

String literals still work as before, of course.

Advantages: no string literals (ehm, so what?); PyCharm completion; potential for more control, e.g. warning if an attribute is not declared as setting

Disadvantages: one more piece of magic (but not too mystic)

### DIFF
--- a/Orange/widgets/data/owcontinuize.py
+++ b/Orange/widgets/data/owcontinuize.py
@@ -59,33 +59,34 @@ class OWContinuize(widget.OWWidget):
 
     def __init__(self):
         super().__init__()
+        my = self.my  # type: OWContinuize
 
         box = gui.vBox(self.controlArea, "Multinomial attributes")
         gui.radioButtonsInBox(
-            box, self, "multinomial_treatment",
+            box, self, my.multinomial_treatment,
             btnLabels=[x[0] for x in self.multinomial_treats],
             callback=self.settings_changed)
 
         box = gui.vBox(self.controlArea, "Continuous attributes")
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owcontinuize.py
```diff
@@ -59,33 +59,34 @@ class OWContinuize(widget.OWWidget):
 
     def __init__(self):
         super().__init__()
+        my = self.my  # type: OWContinuize
```
astaric: Why not
```
my = OWContinuize
```
? Shorter, and 0% magic free?

janezd: Doesn't work. `my.class_treatment` must resolve to `str`. `OWContinuize.class_treatment` is `bool`, and `gui` can't use it. It would work if `OWContinuize.class_treatment` were still a `Setting` and if `gui` would accept settings (besides strings, for backward compatibility).

(((On the other hand, `gui` can take the `id` of the object and parse `master.__dict__` to discover the name of the setting. Then even `self.class_treatment` would work, but this would be black magic. It's attractive, though. :)))

Issue #1184, opened 2016-04-15
Integrate Orange with OpenML.org


brylie:
The OpenML.org website provides buttons to export flows for Weka, MOA, and RapidMiner. E.g.
http://www.openml.org/t/1940

As an Orange user, how do I import a data-flow from OpenML.org into an Orange data flow graph?
# Related issues
- #1264 OpenML Authentication widget
- #1265 OpenML Data Set widget

kernc:
We don't support loading Weka, MOA, or RapidMiner projects. Orange has its own workflow scheme (.ows). Incidentally, OpenML doesn't (yet) support Orange. If you care, perhaps you can write them and express your interest? :smiley:

brylie:
@kernc good suggestion. There is some middle ground here it seems.

brylie:
I posed the Orange Data Mining + OpenML question on the [OpenML discussion list](https://groups.google.com/forum/#!topic/openml/W_QeCy3BP8w), and got a response:

> We definitely hope to support Orange, as well as other Python libraries. We have a Python interface under development: http://www.openml.org/guide#!python

> It currently allows you to download data, run experiments, and upload results. We also have some examples in scikit-learn. What is does not yet support are Orange-specific procedures to export Orange workflows.

> As far as I know nobody is currently working on the latter but of course we greatly encourage work in this direction and will help as much as we can. It just takes someone (or a few people) to get this started.

kernc:
Thanks!

> It just takes someone (or a few people) to get this started.

Would you like to work on it?

brylie:
Well, I am rather new to the entire ML realm, having just started learning Orange in the past couple of weeks. I can certainly try, but may need a someone to break the task into small chunks.

brylie:
I got basic approval to pursue this task at work, so can dedicate some time to the task given guidance.

brylie:
One starter idea is to create a UI widget that can be used to enter OpenML credentials.

brylie:
Also, where can I find instructions describing how to set up a development environment? I am running Linux, with primary experience in JavaScript development.

cheral:
https://github.com/biolab/orange3/wiki/Installation-on-Ubuntu-13.04
http://orange-development.readthedocs.org/index.html
http://orange.biolab.si/docs/

kernc:
Actually, these are the up-to-date instructions: https://github.com/biolab/orange3#installing If you seem to be missing any packages, install them from your distribution channels.

[...]

PR #1213, opened 2016-04-25
[REF] Remove unnecessary parentheses

brylie:
As discovered in [discussion around issue #1184](https://github.com/biolab/orange3/issues/1184#issuecomment-214215811), the tuple is not necessary here. Only a single string is necessary, as is documented in the setuptools official [entry_points example](https://pythonhosted.org/setuptools/setuptools.html#dynamic-discovery-of-services-and-plugins).

### DIFF
--- a/doc/development/source/orange-demo/setup.py
+++ b/doc/development/source/orange-demo/setup.py
@@ -5,5 +5,5 @@ setup(name="Demo",
       package_data={"orangedemo": ["icons/*.svg"]},
       classifiers=["Example :: Invalid"],
       # Declare orangedemo package to contain widgets for the "Demo" category
-      entry_points={"orange.widgets": ("Demo = orangedemo")},
+      entry_points={"orange.widgets": "Demo = orangedemo"},
       )


doc/development/source/orange-demo/setup.py
```diff
@@ -5,5 +5,5 @@
       package_data={"orangedemo": ["icons/*.svg"]},
       classifiers=["Example :: Invalid"],
       # Declare orangedemo package to contain widgets for the "Demo" category
-      entry_points={"orange.widgets": ("Demo = orangedemo")},
-      )
\ No newline at end of file
+      entry_points={"orange.widgets": "Demo = orangedemo"},
```
kernc: In Python, these parentheses should make no difference:
```py
assert 'Demo = orangedemo' == ("Demo = orangedemo")
```
It's only a tuple when the item list (of a single item) ends with a comma: `('whatever',)`

The reason for it not working before must have been somewhere else.

lanzagar:
It is true that just the string is enough - the parentheses were not necessary.
BUT they were not a tuple so they could not have caused the issue this tried to solve.

I will just accept this PR, as it is still an improvement, but please do continue to resolve the issue.

PR #1214, opened 2016-04-25
Fix broken link to setuptools entry_points docs

brylie:
Changed:
- Fix broken link to setuptools `entry_points` documentation


ajdapretnar:
@brylie Thank you so much for discovering and fixing this!

PR #1216, opened 2016-04-26
Confusion matrix: Add context settings, store selection

janezd:
Confusion matrix now stores the selection in context settings and keeps the learner selection after `set_results`. @ajda, @BlazZupan, schemata with multiple confusion matrices and selections now work. PR is ready, I guess, unless...

@astaric The context is determined by the class, so the learner selection is not context dependent. Keeping the learner selection after `set_results` is a bit convoluted and could perhaps be improved, but we first have to decide whether we should move it to context settings (and have it stored for free). In this case, we need a useful `EvaluationResultsContextHandler`. There is one in `widgets.evaluation.contexthandler`, but it's written for some specific case (requires target class) and not used anywhere.

### DIFF
--- a/Orange/widgets/evaluate/owconfusionmatrix.py
+++ b/Orange/widgets/evaluate/owconfusionmatrix.py
@@ -86,7 +86,7 @@ class OWConfusionMatrix(widget.OWWidget):
 
     settingsHandler = settings.ClassValuesContextHandler()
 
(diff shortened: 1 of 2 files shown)


janezd:
Sorry for the monologue. Learner selection can't be context setting. In a common use, switching the data should not change the learner selection -- consider a case in which somebody uses multiple confusion matrices...

PR #1219, opened 2016-05-03
[RFC] Updated sh script for Linux

ajdapretnar:
Including an updated sh script for Linux. Needs testing.
@ales-erjavec @astaric I'm not even sure this is a good folder to put it in, suggestions welcome.

### DIFF
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+p=$PWD
+sudo apt-get update
+sudo apt-get -y install git python-pip python-virtualenv python3-dev python3-numpy python3-scipy python3-pyqt4 python-qt4-dev python3-sip-dev libqt4-dev
+
+virtualenv -p python3 --system-site-packages orange3env
+source orange3env/bin/activate
+
+echo "/usr/lib/python3/dist-packages/" > "orange3env/lib/python3.4/site-packages/0.pth"
+pip install --upgrade numpy
+
+git clone https://github.com/biolab/orange3
+cd orange3
+pip install -r requirements-core.txt  # For Orange Python library
+pip install -r requirements-gui.txt   # For Orange GUI
+pip install -r requirements-sql.txt   # To use SQL support
+pip install -r requirements-opt.txt   # Optional dependencies, may fail
+pip install Orange-Bioinformatics
+python setup.py develop


scripts/setup-ubuntu.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+p=$PWD
+sudo apt-get update
+sudo apt-get -y install git python-pip python-virtualenv python3-dev python3-numpy python3-scipy python3-pyqt4 python-qt4-dev python3-sip-dev libqt4-dev
+
+virtualenv -p python3 --system-site-packages orange3env
+source orange3env/bin/activate
+
+echo "/usr/lib/python3/dist-packages/" > "orange3env/lib/python3.4/site-packages/0.pth"
```
kernc: What does this hack do?

scripts/setup-ubuntu.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+p=$PWD
```
kernc: Unquoted is unsafe if any spaces are in path leading to current working directory. Also isn't used anywhere.

scripts/setup-ubuntu.sh
```diff
@@ -0,0 +1,20 @@
...
+echo "/usr/lib/python3/dist-packages/" > "orange3env/lib/python3.4/site-packages/0.pth"
+pip install --upgrade numpy
+
+git clone https://github.com/biolab/orange3
+cd orange3
+pip install -r requirements-core.txt  # For Orange Python library
+pip install -r requirements-gui.txt   # For Orange GUI
+pip install -r requirements-sql.txt   # To use SQL support
+pip install -r requirements-opt.txt   # Optional dependencies, may fail
+pip install Orange-Bioinformatics
```
kernc: Why bio and not the rest of add-ons?

scripts/setup-ubuntu.sh
acopar: This enables access to system-wise numpy and scipy installed with apt-get and avoids the need to compile them. It also solves confusion on stock ubuntu and debian systems when you have existing numpy installations on the system.

astaric:
I am not sure I would run this as a linux user. The script runs sudo commands, assumes apt-get to manage packages, and creates virtualenv in current directory.

I would rather have an up to date installation guides on [wiki](https://github.com/biolab/orange3/wiki/Installation-on-Ubuntu-13.04) and proper installation for supported distributions (such as .deb, .rpm or aur package).

PR #1227, opened 2016-05-05
[ENH] Simplify objects wrapping pipeline.

s-alexey:


janezd:
We discussed this PR at today's meeting. We like what you've done, for two reasons. First, we'd like to *declare* interfaces instead of programming them. Second, `gui.py` needs refactoring (reimplementation, actually), and your direction is appealing.

Your PR is limited to learners. We are thinking about something along this line for all widgets.

The bigger problem is that you have basically moved the definition of the interface to the implementation of the learner. If the learner only declares the possible options, the gui is going to look ugly; widgets will always require some manual tweaking of the layout. If, however, the options in the learner are declared with such details that the gui looks nice, the learner will include too many details about the layout. We thus lose the separation between the implementation of the learner and the interface.

We said we'd close the PR in this form, though the classes for options may be very useful in some future refactoring of `gui.py`. If you wish, you can try it (perhaps we should discuss it a bit more before), or you can work on this when you're here with us.

s-alexey:
I agree with interface definition. I was going to extend `options_layout` and remove `GUI` classes (move them to widget classes). This will separate gui definition from the implementation. 
And, indeed, complex structures looks ugly so far. But they can be customized using options directly (without `options_layout` method). For instance (consider example above):
```python 
layout.add_widget(lr.values['C'].as_widget())
layout.add_widget(lr.values['penalty'].label())
layout.add_widget(lr.values['penalty'].as_widget())
```
The main advanatage is that no addition methods are needed (`lr.penalty` and `lr.C` will be syncronized with their widgets).

It also can be applied elsewhere, but of course, it takes some time. I chose learners just to show how it works. 

I was planning to use it in my work, so I hope it will be merged.

PR #1231, opened 2016-05-06
OWTable: sort columns of discrete variables that look like ints as ints

janezd:
I'm ignoring cyclomatic complexity: `data` methods are essentially switch statements.

@ajda, you asked for this. Can you check it on your data?


ajdapretnar:
@janezd You might want to check your tag. :grin: But yes, I'll give it a look.

ajdapretnar:
@janezd Not sure this works. I'm still getting the table wrongly sorted. Is this intended?

fix

janezd:
Which tag should I check?

It indeed doesn't work. I don't know how I checked this. And unsurprisingly, @ales-erjavec's brilliant models already have almost everything we need for this.

PR #1237, opened 2016-05-09
[ENH] Table.attributes field.

markotoplak:

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -278,6 +278,18 @@ class FileFormat(metaclass=FileFormatMeta):
         return cls.write_file(filename, data)
 
     @classmethod
+    def write_table_metadata(cls, filename, data):
+        if isinstance(filename, str) and data.attributes:
+            with open(filename + '.metadata', 'wb') as f:
+                pickle.dump(data.attributes, f, pickle.HIGHEST_PROTOCOL)
+
+    @classmethod
+    def set_table_metadata(cls, filename, table):
+        if isinstance(filename, str) and path.exists(filename + '.metadata'):
+            with open(filename + '.metadata', 'rb') as f:
+                table.attributes = pickle.load(f)
+
+    @classmethod
     def locate(cls, filename, search_dirs=('.',)):
         """Locate a file with given filename that can be opened by one
         of the available readers.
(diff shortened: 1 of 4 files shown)


Orange/data/io.py
```diff
@@ -674,6 +677,9 @@ def write_file(cls, filename, data):
             writer = csv.writer(file, delimiter=cls.DELIMITERS[0])
             cls.write_headers(writer.writerow, data)
             cls.write_data(writer.writerow, data)
+        if isinstance(filename, str) and data.attributes:
+            with open(filename + '.apck', 'wb') as f:
+                pickle.dump(data.attributes, f, pickle.HIGHEST_PROTOCOL)
```
kernc: Could these two chunks be methods on `FileFormat` class? There might be use for them with other readers/writers?

Orange/data/io.py
```diff
@@ -662,6 +662,9 @@ def read(self):
                                    '{}{}').format(pos,
                                                   ('-' + str(endpos)) if (endpos - pos) > 1 else '')
                         warnings.warn(warning)
+                    if isinstance(self.filename, str) and path.exists(self.filename + '.apck'):
+                        with open(self.filename + '.apck', 'rb') as f:
```
kernc: I vote for _.attributes_ or _.metadata_ file extensions.

Orange/data/table.py
```diff
@@ -364,6 +369,10 @@ def from_table_rows(cls, source, row_indices):
         self.W = source.W[row_indices]
         self.name = getattr(source, 'name', '')
         self.ids = np.array(source.ids[row_indices])
+        if hasattr(source, "attributes"):
+            self.attributes = source.attributes
+        else:
+            self.attributes = {}
```
kernc:
```py
self.attributes = getattr(source, 'attributes', {})
```

Orange/data/table.py
```diff
@@ -334,6 +335,10 @@ def get_columns(row_indices, src_cols, n_rows, dtype=np.float64):
                 self.ids = np.array(source.ids[row_indices])
             else:
                 cls._init_ids(self)
+            if hasattr(source, 'attributes'):
+                self.attributes = source.attributes
+            else:
+                self.attributes = {}
```
kernc: As below.

BlazZupan:
I have tested this and works well. Please update the documentation as well (Data Table section of the Reference documentation).

BlazZupan:
This feature is needed urgently. I have merged it, though documentation is still required. The request for documentation was in this case (as an exception) added to Orange to do list.

PR #1239, opened 2016-05-09
TableModel: sort int-like discrete variables like ints

janezd:
In `Table` widget and other widgets that use `TableModel`, a column with discrete values that look like integer numbers will now be sorted like integer numbers.

@ajdapretnar, this should work now. (I also realized '*which tag*'. :))

@ales-erjavec, can you please check that the change does not have any unintended side effects?

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -17,7 +17,7 @@ from PyQt4.QtGui import (
 
 import numpy
 
-from Orange.data import Variable, Storage
+from Orange.data import Variable, DiscreteVariable, Storage
 from Orange.widgets import gui
 from Orange.widgets.utils import datacaching
 from Orange.statistics import basic_stats
@@ -719,6 +719,21 @@ class TableModel(QAbstractTableModel):
                 partial(format_dense, var)
             )
 
+        def _create_sort_keys():
+            keys = []
+            for var in self.vars:
+                vkeys = None
+                if isinstance(var, DiscreteVariable):
+                    try:
+                        vkeys = [int(x) for x in var.values]
+                    except ValueError:
+                        pass
+                    else:
+                        vkeys.append(max(vkeys) + 1)
+                        vkeys = numpy.array(vkeys)
+                keys.append(vkeys)
+            return keys
+
         columns = []
 
         if self.X_density != Storage.DENSE:
@@ -775,6 +790,7 @@ class TableModel(QAbstractTableModel):
         self.__sortInd = None
         # The inverse of __sortInd
         self.__sortIndInv = None
+        self.__sortValueKeys = _create_sort_keys()
 
     def sort(self, column, order):
         """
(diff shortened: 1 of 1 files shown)


ajdapretnar:
@janezd Works like a charm. :)

Orange/widgets/utils/itemmodels.py
```diff
@@ -853,6 +859,9 @@ def columnSortKeyData(self, column, role):
             col_data = numpy.asarray(col_view)
             if self.__sortInd is not None:
                 col_data = col_data[self.__sortInd]
+            value_keys = self.__sortValueKeys[column]
+            if value_keys is not None:
+                col_data = value_keys[col_data.astype(numpy.int)]
```
ales-erjavec: This will raise and error if col_data contains any NaNs.

Orange/widgets/utils/itemmodels.py
```diff
@@ -775,6 +775,12 @@ def row_instance(index):
         self.__sortInd = None
         # The inverse of __sortInd
         self.__sortIndInv = None
+        self.__sortValueKeys = [
+            numpy.array([int(x) for x in var.values])
+            if isinstance(var, DiscreteVariable) and
+            all([str.isdigit(value) for value in var.values])
```
ales-erjavec: Should negative integers also be accepted? (`str.isdigit("-1") == False`)

ajdapretnar: I think so. Now they are sorted erroneously (e.g. 5,4,3,2,1,0,-3,-2,-1).

janezd:
@ajdapretnar, can you check if it still works like charm? It also works with negative integers now.

@ales-erjavec, I assigned this to you, but feel free to remove the assignment if you wish.

kernc:
@janezd Why does it only work with integers?

Orange/widgets/utils/itemmodels.py
```diff
@@ -849,10 +865,17 @@ def columnSortKeyData(self, column, role):
...
+            col_view, _ = self.source.get_column_view(var)
             col_data = numpy.asarray(col_view)
             if self.__sortInd is not None:
                 col_data = col_data[self.__sortInd]
+            value_keys = self.__sortValueKeys[column]
+            if value_keys is not None:
+                if col_data is col_view:
+                    col_data = numpy.array(col_data)
+                col_data[numpy.isnan(col_data)] = len(var.values)
+                col_data = value_keys[col_data.astype(numpy.int)]
```
kernc: Sorting discrete floats could work if above cast was made to `float` and here `np.searchsorted(value_keys, col_data)`?

janezd:
This is meant for values that are integers, but encoded as discrete. It's trivial to extend this to floats by just replacing `int` with `float`, but I'm not sure we want this. @ajdapretnar, your opinion?

ajdapretnar:
I don't see how or why you would have discrete floats. Even this fix is pushing it a bit, since there's little value in having 10+ discrete integers per feature (the only use case I can think of is having, say, US states encoded as ints and would like to sort them), but it is still probably more useful than introducing also floats.

@janezd Sorting works for me.

Orange/widgets/utils/itemmodels.py
janezd: This is `DiscreteVariable`, so `col_data` contains integer indices (stored as floats). It would suffice to change `int(x)` to `float(x)` in function `_create_sort_keys`. But I don't see a sensible use case.

kernc:
@janezd, @ajdapretnar Please look at the _ecoli_ dataset, 4th and 5th columns. Surely, one can imagine numerous similar cases.

ajdapretnar:
@kernc Makes sense. With Janez's fix sorting of positive floats works, but not for negative values.

janezd:
Closed in favour of #1255.

PR #1241, opened 2016-05-10
[WIP] Cleaning unittest: test_table.py

rekonder:
Adding iterators where is possible, deleting duplicated rows in test_table.py

### DIFF
--- a/Orange/tests/test_table.py
+++ b/Orange/tests/test_table.py
@@ -37,12 +37,11 @@ class TableTestCase(unittest.TestCase):
 
     def test_indexing_class(self):
         d = data.Table("test1")
-        self.assertEqual([e.get_class() for e in d], ["t", "t", "f"])
-        cind = len(d.domain) - 1
-        self.assertEqual([e[cind] for e in d], ["t", "t", "f"])
-        self.assertEqual([e["d"] for e in d], ["t", "t", "f"])
         cvar = d.domain.class_var
-        self.assertEqual([e[cvar] for e in d], ["t", "t", "f"])
+        cind = len(d.domain) - 1
+        self.assertEqual([e.get_class() for e in d], ["t", "t", "f"])
+        for val in (cind, "d", cvar):
+            self.assertEqual([e[val] for e in d], ["t", "t", "f"])
 
     def test_filename(self):
         dir = data.table.get_sample_datasets_dir()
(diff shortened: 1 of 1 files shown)


janezd:
We discussed this at some length. Using loops instead of duplicating is of course better in general, but here we don't agree about which code makes it easier to see what is tested and what fails. Despite some voices and arguments in favour, these changes had no clear advocate so we decided to close it for now.

We really appreciate your effort; unit tests are crucial. We however discussed that it would be better to have more tests, so perhaps you could focus on writing new ones instead.

Orange/tests/test_table.py
```diff
@@ -53,285 +54,146 @@ def test_filename(self):
...
-            metae = d.domain.index("e")
-            self.assertEqual(d[0, metae], "i")
-            self.assertEqual(d[0, vare], "i")
-            self.assertEqual(d[0, "e"], "i")
-            self.assertEqual(d[0][metae], "i")
-            self.assertEqual(d[0][vare], "i")
-            self.assertEqual(d[0]["e"], "i")
-            self.assertEqual(d[np.int_(0), np.int_(metae)], "i")
-            self.assertEqual(d[np.int_(0)][np.int_(metae)], "i")
+            for dom, indexOne, indexTwo, comp in (("c", 0, 1, "0"),  # regular, discrete
```
kernc: Picking names according to their function can be much clearer:
```py
for var, inst, var_idx, value in ...
```

Orange/tests/test_table.py
```diff
@@ -53,285 +54,146 @@ def test_filename(self):
...
+                                                  ("b", 0, 0, 0),  # regular, continuous
+                                                  ("b", -2, 0, 3.333),  # negative
+                                                  ("a", 0, 0, "A"),  # meta, discrete
+                                                  ("e", 0, 0, "i")):  # meta, string
+                if dom == "a" or dom == "e":
+                    indexTwo = d.domain.index(dom)
```
kernc: This looks pretty weird at first glance. Perhaps you can use this second type of index in all iterations (regardless of how it was before only in the last two). Like so:
```py
var_domain_index = d.domain.index(var)
self.assertEqual(d[inst, var_domain_index], value)
...
```

Orange/tests/test_table.py
```diff
@@ -53,285 +54,146 @@ def test_filename(self):
         self.assertTrue(d.__file__.endswith("test2.tab"))  # platform dependent
 
     def test_indexing(self):
-        import warnings
-
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore")
+        with catch_warnings():
+            simplefilter("ignore")
```
kernc: I'd actually prefer having just `warnings` imported at the top and these used like `warnings.simplefilter('ignore')`

Orange/tests/test_table.py
```diff
@@ -53,285 +54,146 @@ def test_filename(self):
...
-            self.assertEqual(e[metae], "i")
-            self.assertEqual(e[vare], "i")
-            self.assertEqual(e["e"], "i")
-            self.assertEqual(e[np.int_(metae)], "i")
+            for dom, pos, comp in (("c", 1, "0"),  # regular, discrete
+                                   ("b", 0, 0),  # regular, continuous
+                                   ("a", 0, "A"),  # meta, discrete
+                                   ("e", 0, "i")):  # meta, string
+                if dom == "a" or dom == "e":
+                    pos = d.domain.index(dom)
```
kernc: Everything as above.

Orange/tests/test_table.py
```diff
@@ -53,285 +54,146 @@ def test_filename(self):
...
+                                                            (-1, 2),  # remove last
+                                                            (-2, 2))):  # remove one before last
+                del d[delIndex]
+                self.assertEqual(len(d), initlen - index)
+                self.assertEqual(d[pos, "e"], "4ex")
+                if index is 4:
+                    self.assertNotEqual(d[-1, "e"], "was last")
+                    d[-1, "e"] = "was last"
+                elif index is 5:
+                    self.assertEqual(d[-1, "e"], "was last")
```
kernc: This refactoring is not ok. Loops are only clearer when there aren't countless obfuscating exceptions breaking their flow.

Orange/tests/test_table.py
```diff
@@ -53,285 +54,146 @@ def test_filename(self):
...
-                del d[100]
-            self.assertEqual(len(d), initlen - 6)
-
-            with self.assertRaises(IndexError):
-                del d[-100]
-            self.assertEqual(len(d), initlen - 6)
+            for index in (100, -100):
+                with self.assertRaises(IndexError):
+                    del d[index]
+                self.assertEqual(len(d), initlen - 6)
```
kernc: This test (not only this line) is idiotic. If you don't opt to revert it, refactor the whole function.

Orange/tests/test_table.py
```diff
@@ -53,285 +54,146 @@ def test_filename(self):
...
-            d[0] = ["3.14", "1", "f"]
-            almost_equal_list(d[0].values(), [3.14, "1", "f"])
+            for index, value, realValue in ((0, ["3.14", "1", "f"], [3.14, "1", "f"]),
+                                            (0, [3.15, 1, "t"], [3.15, "0", "t"]),
+                                            (np.int_(0), [3.15, 2, "f"], [3.15, 2, "f"]),
+                                            (0, data.Instance(d.domain, ["3.16", "1", "f"]), [3.16, "1", "f"]),
+                                            (np.int_(0), data.Instance(d.domain, ["3.16", 2, "t"]), [3.16, 2, "t"]),
+                                            (0, data.Instance(d.domain, ["3.16", "1", "f"]), [3.16, "1", "f"])):
+                d[index] = value
+                almost_equal_list(d[0].values(), realValue)
```
kernc: Inline that function call:
```py
self.assertAlmostEqual(list(d[0]).values(), realValue)
```

Orange/tests/test_table.py
```diff
@@ -53,285 +54,146 @@ def test_filename(self):
...
-            self.assertEqual([e[0] for e in x], [2.24, 2.23, 2.22])
-
-            x = d[-3:]
-            self.assertEqual(len(x), 3)
-            self.assertEqual([e[0] for e in x], [2.26, 3.333, Unknown])
+            for val, realValue in ((d[:3], [0, 1.1, 2.22]),
+                                   (d[2:5], [2.22, 2.23, 2.24]),
+                                   (d[4:1:-1], [2.24, 2.23, 2.22]),
+                                   (d[-3:], [2.26, 3.333, Unknown])):
+                x = val
```
kernc: What does this line do? :smiley:

Orange/tests/test_table.py
```diff
@@ -691,22 +535,15 @@ def test_convert_through_append(self):
...
-        s = pickle.dumps(d)
-        d2 = pickle.loads(s)
-        self.assertEqual(d[0], d2[0])
 
-        self.assertEqual(d.checksum(include_metas=False),
-                         d2.checksum(include_metas=False))
+        for name in ("zoo", "iris"):
+            d = data.Table(name)
+            s = dumps(d)
+            d2 = loads(s)
```
kernc: Have `pickle` imported and `pickle.dumps()` used.

Orange/tests/test_table.py
```diff
@@ -1657,23 +1418,16 @@ def test_can_copy_table(self):
...
-
-    def test_can_filter_row_with_slice(self):
-        for slice_ in self.interesting_slices:
-            new_table = data.Table.from_table(
-                self.domain, self.table, row_indices=slice_)
-            self.assert_table_with_filter_matches(
-                new_table, self.table, rows=slice_)
+    def test_filter_row(self):
+        for indices in (([0], [1, 5, 6, 7]),  # with list
+                        self.interesting_slices):  # with slice
```
kernc: Put these where they are only used, into this functions.

Orange/tests/test_table.py
```diff
@@ -1704,10 +1458,10 @@ def test_can_use_metas_as_new_columns(self):
...
     def test_can_use_combination_of_all_as_new_columns(self):
         a, c, m = column_sizes(self.table)
-        order = ([random.randrange(a) for _ in self.domain.attributes] +
-                 [random.randrange(a, a + c) for _ in self.domain.class_vars] +
-                 [random.randrange(-m + 1, 0) for _ in self.domain.metas])
-        random.shuffle(order)
+        order = ([randrange(a) for _ in self.domain.attributes] +
+                 [randrange(a, a + c) for _ in self.domain.class_vars] +
+                 [randrange(-m + 1, 0) for _ in self.domain.metas])
+        shuffle(order)
```
kernc: If you also just `import random` at the top, these needless changes will get reverted. It's ok to import modules when that is more clear, i.e. when you are just calling their functions. Otoh, it's better to import types standalone. It just always depends. :smile:

kernc:
These changes, and the ones on top of them, have the required quorum of advocates.

janezd:
Sorry for mixed signals. :)

Please follow @kernc's suggestions wrt this PR. But the recommendation to write new tests instead of refactoring the existing still stands.

kernc:
> to write new tests instead of refactoring the existing still stands.

Please discuss it with @astaric on Monday. :smile:

Orange/tests/test_table.py
rekonder: Actually nothing smart :). I will delete it and in next lines include val where is needed instead x :)

Orange/tests/test_table.py
```diff
@@ -2019,7 +1772,7 @@ def test_delete_rows(self):
     def test_clear(self):
         self.table.clear()
         self.assertEqual(len(self.table), 0)
-        for i in self.table:
+        for _ in self.table:
```
zidarsk8: isn't this the same as `if self.table:` and it makes it more readable.

astaric: It is not the same. **For** will call `__iter__` (or `__getitem__` if `__iter__` is not defined). **If** will call `__bool__`.

But we could (should?) test both.

rekonder: I add both cases now

astaric:
This WIP PR has not been updated for almost two months so I am closing it. If you decide to continue your work on this, feel free to reopen.

PR #1246, opened 2016-05-12
[FIX] owdistributions: Check that the data is not empty

lanzagar:
Currently the Distributions widget crashes on empty data.
This shows a warning instead.

### DIFF
--- a/Orange/widgets/visualize/owdistributions.py
+++ b/Orange/widgets/visualize/owdistributions.py
@@ -211,8 +211,12 @@ class OWDistributions(widget.OWWidget):
     def set_data(self, data):
         self.closeContext()
         self.clear()
+        self.warning(0)
         self.data = data
         if self.data is not None:
+            if not self.data:
+                self.warning(0, "Empty input data cannot be visualized")
+                return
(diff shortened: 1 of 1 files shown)


Orange/widgets/visualize/owdistributions.py
```diff
@@ -211,8 +211,12 @@ def update_views(self):
     def set_data(self, data):
         self.closeContext()
         self.clear()
+        self.warning(0)
         self.data = data
         if self.data is not None:
+            if not self.data:
```
kernc: Can you be a tiny bit more verbose:
```py
if not len(self.data):
```

lanzagar: No.

To be more verbose:
This is not a question of verbosity - the two conditions are different.
`not len` is a shortcut (which is not shorter) that calls and computes `__len__` instead of `__bool__`.
Computing the size of the table and checking if that is nonzero might be as fast for numpy tables, but is not in general (e.g. large SQL tables). And in this case I do not care about the size, just the non-emptiness.

PR #1247, opened 2016-05-13
[FIX] kmeans: Compute silhouette only when possible

lanzagar:
Silhouette score can be computed only for 2 <= k < n. Prevent its call in other cases.
Reproduce the bug by e.g. setting k=n in the kmeans widget.

Side note: I do not like the fact that this and other scores are computed automatically after every clustering run. Currently, other code depends on the scores being there, but a future PR could address this and at least make it optional, but preferably separate the scoring methods into individual functions, which can be called after clustering.

### DIFF
--- a/Orange/clustering/kmeans.py
+++ b/Orange/clustering/kmeans.py
@@ -21,11 +21,12 @@ class KMeans(SklProjector):
     def fit(self, X, Y=None):
         proj = skl_cluster.KMeans(**self.params)
         if isinstance(X, Table):
-            proj = proj.fit(X.X, Y)
-            proj.silhouette = silhouette_score(X.X, proj.labels_)
-        else:
-            proj = proj.fit(X, Y)
+            X = X.X
+        proj = proj.fit(X, Y)
+        if 2 <= proj.n_clusters < len(X):
             proj.silhouette = silhouette_score(X, proj.labels_)
+        else:
+            proj.silhouette = 0
         proj.inertia = proj.inertia_ / len(X)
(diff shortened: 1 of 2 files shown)


janezd:
Taking 10 instances from Iris and optimizing k-means for 2 - 11 clusters gives

```
  File "/Users/janez/Dropbox/orange3/Orange/widgets/unsupervised/owkmeans.py", line 247, in run
    self.run_optimization()
  File "/Users/janez/Dropbox/orange3/Orange/widgets/unsupervised/owkmeans.py", line 229, in run_optimization
    self.show_results()
  File "/Users/janez/Dropbox/orange3/Orange/widgets/unsupervised/owkmeans.py", line 257, in show_results
    scores = list(zip(*k_scores))[1]
IndexError: list index out of range
```

if the table is shown.

If I instead optimize for 2 to 10 clusters, I get

```
  File "/Users/janez/Dropbox/orange3/Orange/widgets/unsupervised/owkmeans.py", line 247, in run
    self.run_optimization()
  File "/Users/janez/Dropbox/orange3/Orange/widgets/unsupervised/owkmeans.py", line 229, in run_optimization
    self.show_results()
  File "/Users/janez/Dropbox/orange3/Orange/widgets/unsupervised/owkmeans.py", line 280, in show_results
    item.setData(fmt.format(score), Qt.DisplayRole)
ValueError: Precision not allowed in integer format specifier
```

These issues are at least related to this PR, so perhaps you can include the fixes for them, too.

PR #1249, opened 2016-05-13
Clear report on schema reload, save on close

janezd:
When closing the schema, canvas now also asks whether to save & clear, clear or keep the report.

### DIFF
--- a/Orange/canvas/application/canvasmain.py
+++ b/Orange/canvas/application/canvasmain.py
@@ -858,7 +858,7 @@ class CanvasMainWindow(QMainWindow):
     #################
     def pre_close_save(self):
         """
-        Ask whether to save the schenma (if changed) and the report (if
+        Ask whether to save the schema (if changed) and the report (if
         not empty).
 
         Returns: `False` if the user cancelled, `True` otherwise
@@ -1068,13 +1068,15 @@ class CanvasMainWindow(QMainWindow):
         if report.is_empty():
             return QDialog.Accepted
 
-        msgbox = QMessageBox(QMessageBox.Question, "Report", "Clear report?",
-                             parent=self)
+        msgbox = QMessageBox(QMessageBox.Question,
+                             "Report", "Report window has unsaved changes.",
+                             parent=self,
+                             informativeText="Save the report?")
         # Cancel must have AcceptRole, otherwise Os X reorders the buttons
-        cancel = msgbox.addButton("Cancel", QMessageBox.AcceptRole)
-        keep = msgbox.addButton("Keep", QMessageBox.AcceptRole)
-        msgbox.addButton("Clear", QMessageBox.AcceptRole)
         save = msgbox.addButton("Save && Clear", QMessageBox.AcceptRole)
+        msgbox.addButton("Clear", QMessageBox.DestructiveRole)
+        keep = msgbox.addButton("Keep", QMessageBox.AcceptRole)
+        cancel = msgbox.addButton("Cancel", QMessageBox.RejectRole)
         msgbox.exec()
         button = msgbox.clickedButton()
         if button is cancel or \
@@ -1093,11 +1095,12 @@ class CanvasMainWindow(QMainWindow):
         """
         from Orange.canvas.report.owreport import OWReport
         report = OWReport.get_instance()
-        if report.is_empty():
+        if not report.is_changed():
             return QDialog.Accepted
 
         answ = message_question(
-            "Save report?", "Report",
+            "Report window contains unsaved changes", "Report window",
+            "Save the report?",
             buttons=QMessageBox.Yes | QMessageBox.No | QMessageBox.Cancel,
             parent=self)
         if answ == QMessageBox.Cancel:
--- a/Orange/canvas/application/canvasmain.py
+++ b/Orange/canvas/application/canvasmain.py
@@ -856,16 +856,24 @@ class CanvasMainWindow(QMainWindow):
     #################
     # Action handlers
     #################
+    def pre_close_save(self):
+        """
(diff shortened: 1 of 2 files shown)


Orange/canvas/application/canvasmain.py
```diff
@@ -1057,6 +1061,23 @@ def set_new_scheme(self, new_scheme):
 
         old_scheme.deleteLater()
 
+    def ask_clear_report(self):
+        from Orange.canvas.report.owreport import OWReport
```
zidarsk8: I'm not sure what your policy on imports is, but in the contributing.txt it says "Whenever PEP-8 is undefined, adhere to Google Python Style Guide." That would mean all imports must be on the top of the file.

Orange/canvas/report/owreport.py
```diff
@@ -191,6 +191,10 @@ def _remove_item(self, row):
         self.table_model.removeRow(row)
         self._build_html()
 
+    def _clear(self):
```
zidarsk8: I don't see this function used anywhere. Is there a good reason for introducing dead code here?

Orange/canvas/application/canvasmain.py
janezd: You're right in general, but in this case it wouldn't work since the two modules would mutually import each other.

Orange/canvas/report/owreport.py
janezd: Oops, in my code it's `clear`, without underscore. I don't know how this happened; I now (force-)pushed again.

ajdapretnar:
@janezd So far I've tested a few options and here's the result.
Now the workflow asks you if you want to keep the report. It keeps it only if you open another workflow over an existing one, but it loses it if you close Orange.
Also if you reload a workflow where you selected to keep the report, report won't be there if you reopen the workflow at a later point (but have closed Orange in the meantime).

Is this what we want or not? I imagine it would be nice to associate a report with an .ows. Also 'keep' in this case is a bit confusing, as it implies it will reopen the report upon reopen Orange (which it doesn't).

Comments?

lanzagar:
If the user selects Save&Clear and then cancels the save operation, she is not returned to the previous state, but loses the report.

Also the code for closing workflows could benefit from refactoring, to avoid calls to ask_clear_report in so many places.

janezd:
- I added Cancel to `ask_clear_report`.
- Cancelling when saving files now works everywhere.
- The code is refactored.

ajdapretnar:
@janezd I've checked all the possible scenarios I can think of and it works.
Probably @lanzagar should be in charge of merging here to follow up on his comment.

lanzagar:
It does work.
However, I will do some nitpicking anyway - sorry in advance.
In order of annoyingness:

1) The order of buttons in `ask_clear_report`: Cancel should be the last one (on the right) as in all other dialogs. I like the rest (Keep first), but would then change the question from "Clear report?" to "Keep report?" - like it is for workflows (question asks whether to save, first button is save)
Also see 3)

2) If the user saves the workflow and the report and then tries to quit orange, she is asked if she wants to save the report. How much work would it be to change the "empty report" check with "unsaved changes" check?

3) The questions in the new dialogs for saving reports are very brief, which leads to a noticeably different style to the previous dialog for saving workflows. I like brevity, but maybe the styles could be brought a bit closer, so the difference does not scream at the user as much.

janezd:
**Ad 1.** I defined the cancel button as `AcceptRole` precisely to be able to push it to the right. On Os X it already looks like this:

screen shot 2016-05-19 at 21 43 12

[This page](https://doc.qt.io/archives/qq/qq19-buttons.html) explains what's happening, but your window manager is different. If fixing the order of buttons requires using QDialog instead of QMessageBox ... we can, but I'm not sure it's worth it.

In my case, it starts with Save & Clear. I changed the question to "Save the report", but it make less sense if the first button is "keep".

However, `ask_save_report`, which is called on Quit, has standard buttons so Cancel (with `RejectRole`) is pushed to the left of the `AcceptRole`.

I considered giving the buttons in `ask_clear_report` their proper roles and let the each OS position them according to its particular rules. But on Os X this gives the nonsensical

screen shot 2016-05-19 at 22 12 29

If I give all buttons except the cancel button the `AcceptRole`, it's even worse: Cancel becomes the second button on the right?!

**Ad 2.** You're right and it wasn't not much work at all.

**Ad 3.** Short messages bothered me, too. I split them into text and informative text. They look better on Os X.

Please check the changes w.r.t. 2 and 3. I don't know what to do with 1. Tomorrow we can try to find the configuration of roles that work on all OSs.

Orange/canvas/application/canvasmain.py
```diff
@@ -1057,6 +1056,59 @@ def set_new_scheme(self, new_scheme):
...
+        """
+        from Orange.canvas.report.owreport import OWReport
+        report = OWReport.get_instance()
+        if report.is_empty():
+            return QDialog.Accepted
+
+        msgbox = QMessageBox(QMessageBox.Question,
+                             "Report", "Report window contains some data",
+                             parent=self)
+        msgbox.setInformativeText("Save the report?")
```
kernc: property => kwarg

janezd: > property => kwarg

I keep forgetting this. Thanks for reminders. :)

Orange/canvas/application/canvasmain.py
```diff
@@ -856,16 +856,24 @@ def __update_scheme_margins(self):
     #################
     # Action handlers
     #################
+    def pre_close_save(self):
+        """
+        Ask whether to save the schenma (if changed) and the report (if
```
lanzagar: schenma

PR #1253, opened 2016-05-13
[FIX] report: Saving scenes in reports

janezd:
Unnecessary instantiation of PngWriter caused an error when saving scenes to reports.


ajdapretnar:
@janezd Probably need to merge this one first before testing #1249?

janezd:
The two PRs are unrelated. And Lan already merged it. ;)

PR #1255, opened 2016-05-13
[ENH] DiscreteVariable: sort values as numbers if possible

janezd:
Discrete values used to be sorted alphabetically. If they look numeric, they are now sorted as numbers.


ajdapretnar:
@janezd It works for me (tested on positive ints, negative ints, positive floats, negative floats and both with missing values).

PR #1262, opened 2016-05-17
[WIP] Tests at owscatterplot.py

rekonder:
Adding new tests at owscatterplot.py


Orange/widgets/visualize/owscatterplot.py
```diff
@@ -619,5 +620,38 @@ def test_main(argv=None):
 
     return rval
 
+
+class Test(unittest.TestCase):
```
astaric: Gui tests should extend the GuiTest class, which makes sure that all gui tests are run using the same QApplication.

https://github.com/biolab/orange3/blob/master/Orange/widgets/tests/base.py#L8

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -619,5 +620,38 @@ def test_main(argv=None):
...
+    @classmethod
+    def setUpClass(cls):
```
astaric: Only use setUpClass when it makes sense.

setUp method is called before each test is run for a reason. Try removing the last two tests (*_none) and run all remaining tests. Now the test_subset_data fails, as self.data is still set to iris.

PR #1263, opened 2016-05-17
[ENH] [sic] Add auto-apply checkboxes to learner dialogs

kernc:
Sometimes, the data sets are small, the models are simple, and recalculations in light of changed parameters are fast. No need to have to navigate up and down all the time.

### DIFF
--- a/Orange/widgets/classify/owlogisticregression.py
+++ b/Orange/widgets/classify/owlogisticregression.py
@@ -55,13 +55,10 @@ class OWLogisticRegression(OWBaseLearner):
         self.set_c()
 
     def set_c(self):
-        self.settings_changed()
         self.C = self.C_s[self.C_index]
-        if self.C >= 1:
-            frmt = "C={}"
-        else:
-            frmt = "C={:.3f}"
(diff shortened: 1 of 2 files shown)


kernc:
We should probably disable warnings as they make the dialog too jumpy. We also don't use them anywhere else to inform about the controls' state being "dirty" and the flow not "commited" yet. We should use them everywhere. Or we should make the Apply button disabled unless the state is dirty. Probably the latter. Probably worth thinking about in the general GUI revamping effort, and an ad-hoc solution _now_ is not really necessary.

kernc:
@s-alexey Please review. :smile:

janezd:
I don't like disabled buttons - I like the save icon to be active although I just saved the document...

I would prefer some visual cue -- occasional (not too aggressive) slow flashing of the button, a visible but unaggressive icon ... 

A good thing with auto-commit buttons is that we get this for free, without needing to change the widget.

astaric:
I am merging this as-is. Redesign of auto-commit buttons an be a part of a future pull request.

PR #1269, opened 2016-05-19
[ENH] Install add-on by dragging zip/tgz/wheel onto the addons dialog

kernc:
Needs testing.

### DIFF
--- a/Orange/canvas/gui/utils.py
+++ b/Orange/canvas/gui/utils.py
@@ -13,7 +13,7 @@ from PyQt4.QtGui import (
     QPainter, QStyleOption, QStyle
 )
 
-from PyQt4.QtCore import QPointF
+from PyQt4.QtCore import QPointF, QUrl
 
 import sip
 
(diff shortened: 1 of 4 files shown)


Orange/canvas/gui/utils.py
```diff
@@ -305,3 +305,18 @@ def message(icon, text, title=None, informative_text=None, details=None,
...
+    """Return OS X NSURL file reference as local file path or '' if not NSURL"""
+    if isinstance(url, QUrl):
+        url = url.toString()
+    if not url.startswith('file:///.file/id='):
+        return ''
+    from subprocess import Popen, PIPE, DEVNULL
+    # OS X Foundation, and with it NSURL, is available for default Python 2
+    return Popen('''python -c '
+import sys
+from Foundation import NSURL
```
kernc: Ahh, seems this [only works if PyObjC package is installed](https://stackoverflow.com/questions/1614648/importerror-no-module-named-foundation) (requires building with Xcode).

astaric:
I can confirm that this works for installation of wheel addons on osx. I could not install source distribution since it has a .tar.gz extension, which is not supported.

PR #1270, opened 2016-05-20
[ENH] Silhouette: Add report, improve GUI

janezd:


Orange/widgets/unsupervised/owsilhouetteplot.py
```diff
@@ -308,6 +317,11 @@ def commit(self):
         self.send("Selected Data", selected)
         self.send("Other Data", other)
 
+    def send_report(self):
+        self.report_plot()
+        self.report_caption("Silhouette plot with ({} distance)".
```
lanzagar: Why is the second part of this sentence in parentheses?
I think either the word "with" or the parentheses need to be removed.

lanzagar:
The variable which was used for grouping is not reported!

lanzagar:
I am guessing the annotation var should also be reported, but could not even try it out - do annotations work at all?
The combo does not include discrete variables, and string variables are not shown in the plot.

This could be a bug fix for another PR or added here.

janezd:
I omitted the names of the clustering and annotation variables thinking that these would be obvious from the values appearing in the graph. But they're probably not, so I added them to the report.

Annotation needs string meta variables. You can try them with zoo, except that you have to continuize it first.

A matter for another PS is that the plot is scaled to some fixed height, I guess. If you increase the bar widths, the labels get smaller.

lanzagar:
currently: clustered by 'type', annotated by 'name'
better: grouped by 'type', annotated with 'name'

Other than that I think the report is now OK.

As for the annotations - I almost wrote that I tried on zoo and they did not work. However I discovered the problem. If bar width is less than 5 (default 3) they do not appear. Without any indication or explanation. Very confusing and misleading. But probably to be handled in another PR?

janezd:
by -> with: yes, sure.

I added the indication in a separate commit (but this same PR). I didn't use `self.warning` since the move the interface, pushing the slider from below the mouse. The label also looks better, and there is enough space for it, too.

lanzagar:
I like the label.
@janezd can you take care of the CLA (maybe you don't have all emails you're using added to the github account?) so this can be merged.

janezd:
I was using the same mail. I pushed the last commit again; now it's OK.

PR #1273, opened 2016-05-20
[ENH] Predictions: Report, Data <--> Predictions, fixups

nikicc:
Added report button to Predictions widget. 

Also, modified `report_table` in `report.py` to:
* use `QTableView.itemDelegate()` for formatting values, 
* skip hidden columns.

### DIFF
--- a/Orange/canvas/report/index.html
+++ b/Orange/canvas/report/index.html
@@ -70,6 +70,12 @@
         padding: 2px 6px 2px 6px;
     }
 
+    @media screen {
+        td, th {
+            white-space: nowrap;
+        }
+    }
+
     .caption {
         text-align: center;
         font-size: 90%;
--- a/Orange/canvas/report/report.py
+++ b/Orange/canvas/report/report.py
@@ -1,8 +1,8 @@
 import itertools
 import time
-from collections import OrderedDict
+from collections import OrderedDict, Iterable
 from itertools import chain
-from PyQt4.QtCore import Qt, QAbstractItemModel, QByteArray, QBuffer, QIODevice
+from PyQt4.QtCore import Qt, QAbstractItemModel, QByteArray, QBuffer, QIODevice, QLocale
 from PyQt4.QtGui import QGraphicsScene, QStandardItemModel, QColor
 from Orange.widgets.io import PngFormat
 from Orange.data.sql.table import SqlTable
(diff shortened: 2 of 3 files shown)


janezd:
This report is not of much use: data and predictions (the rows from the two tables) should be put side by side.

nikicc:
I though about that but wasn't quite sure how to merge two `QTableViews`. Will look into it.

janezd:
I suggest you still use `report.report_table`, so any changes made to this function will also apply to Predictions.

What about writing a generator that will generate concatenated rows from the two tables? If `send_report` passes this generator to `report.report_table`, `report_table` will treat it as a "list" (see the local function `report_list` within `report_table`), which should work just fine.

nikicc:
`isinstance(table, list)` doesn't seem to treat generator as a list. Also, I don't see an elegant solution for writing header if I pass a `list`. Anyway, will figure something out.

janezd:
`report_list` can handle generators (`report_standard_model` and `report_abstract_model` actually pass generators to `report_list`. Perhaps we should change `isinstance(table, list)` to `isinstance(table, collections.Iterable)` --- or even to `table is not None`. If the table cannot be printed out, this should trigger an error.

nikicc:
@janezd this looks better now. Changes to `report_abstract_model ` are currently not being used but maybe we should keep them anyway. Now we skip hidden columns and format according to `itemDelegate`.

janezd:
This is much better and more useful.

There's a problem when the last column in predictions shows class distributions: the text is wrapped.

screen shot 2016-05-23 at 22 45 10

This can be prevented by adding `white-space: nowrap;` to `th, td` selector in *index.html* used for reports. I am not sure whether we really want to prevent text wrapping in all tables. @VesnaT, @kernc, @astaric?

kernc:
Given how X can be quite wide ...

janezd:
... wrapping td's will make it quite high ...

In this particular case, I think we want to prevent wrapping, but I'm not sure this holds for everything else that turns into a table. Maybe we should add something like `nowrap=False` argument to `report_table`.

astaric:
What should happen if the width of the table does not fit on the page? For online viewing, overflow=scroll is an option, but probably not when printing to pdf.

nikicc:
On brown-selected this looks even worse since minus signs are in their own rows. In this case I think that wrapping should be disabled. Also grey background does not stretch when scrolling.

screen shot 2016-05-24 at 11 51 29

Likewise on `StringVariables` wrapping looks strange:
screen shot 2016-05-24 at 12 18 05

Maybe we should disable wrapping altogether for online viewing and handle PDSs separately. Though, I am not quite sure how.

nikicc:
@janezd: I put predictions before data. It's seem easier to compare them with class variables IMO.

If we keep this version maybe we should also change this in widget – now that class comes first.

ajdapretnar:
@janezd I'll interfere here and agree with @nikicc. We've recently moved class column to the head of the table and all tables should match this change (namely you want classes and predictions juxtaposed).

janezd:
I agree with everything. @nikicc, can you make these changes? That is (I guess)

- swap the tables in the widget,
- disable wrapping for online viewing using the appropriate media block in the stylesheet.

nikicc:
@janezd I think this is ready for another look 😄

ajdapretnar:
@nikicc @janezd Storing some reports in .html produces UnicodeEncodeError (try with Iris, SVM for Predictions and save to .html). Also .report format seems to be broken.

PR #1278, opened 2016-05-23
Win installer: upgrade pip and scikit-learn in the windows application installer

ales-erjavec:
Upgrade the minimum required pip version needed for installation
    
Require pip 8.* in order to install .whl files created by wheel>=0.27
Keep the wheel==0.26.* in the build env just to be sure.

Upgrade included scikit-learn to 0.17.1

### DIFF
--- a/scripts/windows/build-win-application.sh
+++ b/scripts/windows/build-win-application.sh
@@ -274,7 +274,7 @@ function prepare_orange {
     # ensure that correct numpy and scipy are installed in the build env
     pip install --no-index -f "$BUILDBASE/wheelhouse" \
                 --only-binary numpy,scipy \
-                numpy==$NUMPY_VER, scipy==$SCIPY_VER
+                numpy==$NUMPY_VER scipy==$SCIPY_VER
 
     # ensure that the wheel package in the build env creates .whl files that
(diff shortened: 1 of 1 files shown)


scripts/windows/build-win-application.sh
```diff
@@ -199,7 +199,7 @@ function prepare_msvcr100 {
 
 function prepare_pip {
     local version=${PIP_VER:?}
-    local url=https://pypi.python.org/packages/py2.py3/p/pip/pip-${version}-py2.py3-none-any.whl
+    local url=https://pypi.python.org/packages/9c/32/004ce0852e0a127f07f358b715015763273799bd798956fa930814b60f39/pip-8.1.2-py2.py3-none-any.whl
```
lanzagar: So it is not possible anymore to get the file just with the name and version?
Maybe it would still be better to not hardcode the whole url and read at least the version from the constant (like before)? To avoid updating the constants in lines 70-71 but forgetting this url.
Actually, if we already have the md5 in the constant above, we could probably have the url there as well?

ales-erjavec: I moved the url closer to the pip version definition.

PR #1282, opened 2016-05-25
Fixups

kernc:
From here and there.

### DIFF
--- a/Orange/canvas/application/addons.py
+++ b/Orange/canvas/application/addons.py
@@ -40,7 +40,7 @@ from ..config import ADDON_KEYWORD
 from ..gui.utils import message_warning, message_information, \
                         message_critical as message_error, \
                         OSX_NSURL_toLocalFile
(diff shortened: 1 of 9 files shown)


astaric:
I get the following error when I open the add-on window:

```
version = max(pypi.package_releases(name), key=version.LooseVersion)
UnboundLocalError: local variable 'version' referenced before assignment
```

kernc:
Indeed. Please see revised.

Orange/canvas/application/addons.py
```diff
@@ -150,6 +150,32 @@ def editorEvent(self, event, model, option, index):
         return model.setData(index, checkstate, Qt.CheckStateRole)
 
 
+def get_meta_from_archive(path):
+    """Return project name, version and summary extracted from
```
kernc: If there is some way to do this more elegantly using _pkg_resources_ or _pip_, I'd sure like to know. :unamused:

PR #1300, opened 2016-06-02
Some PyTableModel & PyListModel touch-ups

kernc:
These are actually pretty joyful to use. We should do this more often. :smiley:

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -91,6 +91,8 @@ class PyTableModel(QAbstractTableModel):
             value = self[index.row()][index.column()]
         except IndexError:
             return
+        if role == Qt.DecorationRole and isinstance(value, Variable):
+            return gui.attributeIconDict[value]
         if role == Qt.DisplayRole:
             if (isinstance(value, Number) and
                     not (isnan(value) or isinf(value) or isinstance(value, Integral))):
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -418,6 +418,9 @@ class PyListModel(QAbstractListModel):
         self.__delitem__(i)
         return item
 
+    def indexOf(self, value):
+        return self._list.index(value)
+
     def clear(self):
         del self[:]
 
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -218,7 +218,7 @@ class PyTableModel(QAbstractTableModel):
 
     def __delitem__(self, i):
         if isinstance(i, slice):
-            start, stop, step = _as_contiguous_range(i, len(self))
+            start, stop, _ = _as_contiguous_range(i, len(self))
             stop -= 1
         else:
             start = stop = i = i if i >= 0 else len(self) + i
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -322,7 +322,7 @@ class PyListModel(QAbstractListModel):
         row = index.row()
         if role in [self.list_item_role, Qt.EditRole] \
                 and self._is_index_valid_for(index, self):
-            return self[row]
+            return str(self[row]) if role == Qt.DisplayRole else self[row]
         elif self._is_index_valid_for(row, self._other_data):
             return self._other_data[row].get(role, None)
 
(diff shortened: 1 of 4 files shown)


kernc:
@matevzkren, @PrimozGodec Could one or preferably both of you guys take a look at and review this PR. Consider it mandatory part of your assignments. You both will also have tableviews to work with eventually, I suppose.

Orange/widgets/utils/itemmodels.py
```diff
@@ -76,11 +85,29 @@ def columnCount(self, parent=QModelIndex()):
...
+                absval = abs(value)
+                strlen = len(str(int(absval)))
+                value = '{:.{}{}}'.format(value,
+                                          2 if absval < .001 else
+                                          3 if strlen < 2 else
+                                          1 if strlen < 5 else
+                                          0 if strlen < 6 else
+                                          3,
+                                          'f' if strlen < 6 and absval > .001 else 'e')
+            return str(value)
```
janezd: `value` may already be a string here. Perhaps `return` inside `if`?

On a related note, I'm concerned about the speed of this conversion if the table is huge.

Orange/widgets/utils/itemmodels.py
```diff
@@ -76,11 +85,29 @@ def columnCount(self, parent=QModelIndex()):
...
+            if (isinstance(value, Number) and
+                    not (isnan(value) or isinf(value) or isinstance(value, Integral))):
+                absval = abs(value)
+                strlen = len(str(int(absval)))
+                value = '{:.{}{}}'.format(value,
+                                          2 if absval < .001 else
+                                          3 if strlen < 2 else
+                                          1 if strlen < 5 else
+                                          0 if strlen < 6 else
+                                          3,
```
janezd: These if's are difficult to follow, at least for me. Is this perhaps cleaner?
```
2 if absval < 0.001 else
1 if 2 <= strlen <= 4 else
0 if strlen == 5 else
3
```

Orange/widgets/utils/itemmodels.py
```diff
@@ -145,6 +154,7 @@ def sort(self, column, order=Qt.AscendingOrder):
                          key=lambda i: self._table[i][column],
                          reverse=order != Qt.AscendingOrder)
         self._table[:] = [self._table[i] for i in indices]
+        self._roleData[:] = [self._roleData[i] for i in indices]
```
janezd: I may have missed something, but what if `_roleData` is `None`? Same in other methods that follow.

kernc: Possibly. Where would it be set to `None`?

Orange/widgets/utils/itemmodels.py
kernc: No.

Orange/widgets/utils/itemmodels.py
janezd: As I said, I may have missed something. I overlooked that the constructor calls `wrap`, and that calling `wrap` is also the only (legal) way of assigning to `_table`.

Orange/widgets/utils/itemmodels.py
```diff
@@ -222,39 +239,34 @@ def __setitem__(self, i, value):
     def wrap(self, table):
         self.beginResetModel()
         self._table = table
+        rows, cols = range(len(table)), range(len(table[0]) if table else 0)
+        self._roleData = [[{} for _ in cols] for _ in rows]
```
janezd: I'm not sure this is a good idea. An empty dictionary takes 288 bytes on my system. A very modest table with 10 columns and 100 rows thus takes 288 KB for empty dicts. 10000 columns and 100 rows -- which is still not outrageously big -- will take 288.000,000 bytes. If `table` is, say, a numpy array of doubles, the data will take 8 MB and the `_roleData`, which is create automatically, just for the case we might need it, will take 288 MB.

kernc: I was following PyListModel example (except where noticeably cleaner). If 100,000,000 rows are added to PyListModel, its `_other_data` grows to overwhelming 29 GB.

janezd: I don't know whether this was sarcasm or not. To be on the safe side, I'll reply. :)

It is unlikely that `PyListModel` will have 1e8 rows, but quite likely that a table will have 1e7 (=1000 x 10000) cells and take 2,9 GB - without the widget actually using other roles.

We use `PyListModel` for settings or attribute lists, which are reasonably small, so no harm done here. I support the idea of adding other roles, but setting individual cell roles will make sense only for small tables (like confusion matrices, perhaps). If larger tables define roles, they will overload the method `data` and compute the role values on the fly. I guess `_roleData` should be `None` unless set explicitly.

kernc: PyTableModel was, likewise, meant for smaller tables like confusion matrices, Test & Score results etc. But given how useful it appears to be, I'll take you up on it and make it even better.

PR #1306, opened 2016-06-03
[FIX] Support Sparse Matrices

nikicc:
* Modified RemoveNaNColumns to skip sparse data sets so most classifiers (except Naive Bayes) can work on sparse data.
* Modified `Table.from_table` to work if `table.X` is sparse.
* Changed a few `len(data.X)` to `data.X.shape[0]` since len does not work on scipy sparse matrices.

### DIFF
--- a/Orange/preprocess/fss.py
+++ b/Orange/preprocess/fss.py
@@ -1,6 +1,7 @@
 import random
 import Orange
 import numpy as np
+from scipy.sparse import issparse
 
 from itertools import takewhile
 from operator import itemgetter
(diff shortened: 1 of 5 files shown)


Orange/preprocess/fss.py
```diff
@@ -139,6 +140,9 @@ def __init__(self, threshold=None):
         self.threshold = threshold
 
     def __call__(self, data, threshold=None):
+        if issparse(data.X):
+            return data
```
astaric: This return could use a comment stating why is it safe to return data.

PR #1307, opened 2016-06-03
[FIX] Open correct help file for add-on widgets

astaric:


ajdapretnar:
As for Preprocess, it works. Can't think of another example to test on, so I'm merging.

PR #1308, opened 2016-06-03
[FIX] Canvas item selection indicators

### LINKED ISSUE #1301: Selection of multiple objects (nodes, annotations, arrows) is not always visible
##### Orange version

3.3.5

##### Expected bahavior

cmd + click to select multiple objects should highlight each object

##### Actual behavior

when selecting certain objects, (especially multiple arrows and textboxes) objects often appear "unselected" but when moving the objects they are apparently selected. 

try selecting multiple objects (nodes, textboxes, and arrows) and drag them around.

thanks @astaric for helping me clarify the problem

ales-erjavec:
* Fix an error in text annotation's selection state GUI indicator where, during user interaction, the frame could be removed but the item stays selected
* Make annotation items selection states somewhat more visible (this could still be improved)

Fixes #1301

### DIFF
--- a/Orange/canvas/canvas/items/annotationitem.py
+++ b/Orange/canvas/canvas/items/annotationitem.py
@@ -550,7 +550,7 @@ class ArrowAnnotation(Annotation):
         """
         if self.__color != color:
             self.__color = QColor(color)
-            self.__updateBrush()
+            self.__updateStyleState()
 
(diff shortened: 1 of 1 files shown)


astaric:
Travis is misbehaving again. Can you just amend and force push so that the tests will be run?

PR #1319, opened 2016-06-07
OWFile: Enable domain editing

VesnaT:

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -1,21 +1,21 @@
 import os
-from itertools import chain
-from itertools import count
+from itertools import chain, count
 from warnings import catch_warnings
-from xlrd import open_workbook
+
+import numpy as np
 from PyQt4 import QtGui, QtCore
 from PyQt4.QtGui import QSizePolicy as Policy
 
 from Orange.canvas.gui.utils import OSX_NSURL_toLocalFile
-from Orange.data import Domain
+from Orange.data import Domain, DiscreteVariable, StringVariable
+from Orange.data.table import Table, get_sample_datasets_dir
+from Orange.data.io import FileFormat, UrlReader
 from Orange.widgets import widget, gui
 from Orange.widgets.settings import Setting, ContextHandler, ContextSetting, \
     PerfectDomainContextHandler
 from Orange.widgets.utils.domaineditor import DomainEditor
 from Orange.widgets.utils.itemmodels import PyListModel
 from Orange.widgets.utils.filedialogs import RecentPathsWComboMixin
-from Orange.data.table import Table, get_sample_datasets_dir
-from Orange.data.io import FileFormat, ExcelReader, UrlReader
 
 # Backward compatibility: class RecentPath used to be defined in this module,
 # and it is used in saved (pickled) settings. It must be imported into the
@@ -169,7 +169,6 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
         url_model.wrap(self.recent_urls)
         url_combo.setModel(url_model)
         url_combo.setSizePolicy(Policy.MinimumExpanding, Policy.Fixed)
-        url_combo.setMaximumWidth(500)
         url_combo.setEditable(True)
         url_combo.setInsertPolicy(url_combo.InsertAtTop)
         url_edit = url_combo.lineEdit()
@@ -207,6 +206,7 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
         QtCore.QTimer.singleShot(0, self.load_data)
 
         self.setAcceptDrops(True)
+
     def sizeHint(self):
         return QtCore.QSize(600, 550)
 
@@ -241,7 +241,7 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
             self, 'Open Orange Data File', start_file, self.dlg_formats)
         if not filename:
             return
-
+        self.loaded_file = filename
         self.add_path(filename)
         self.source = self.LOCAL_FILE
         self.load_data()
@@ -259,6 +259,7 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
                 errors.append("An error occured:")
                 errors.append(str(ex))
                 data = None
+                self.editor_model.reset()
             self.warning(
                 33, warnings[-1].message.args[0] if warnings else '')
 
@@ -271,6 +272,8 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
 
         add_origin(data, self.loaded_file)
         self.send("Data", data)
+        self.editor_model.set_domain(data.domain)
+        self.data = data
 
     def _get_reader(self):
         """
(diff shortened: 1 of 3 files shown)


Orange/widgets/data/owfile.py
```diff
@@ -341,22 +344,34 @@ def apply_domain_edit(self):
...
             else:
                 var = tpe(name)
             places[place].append(var)
-            cols[place].append(column)  # This assumes that no columns were skipped in the original file!
+            cols[place].append(col_data)
         domain = Domain(attributes, class_vars, metas)
-        # data = Table.from_file(self.loaded_file, domain=domain, columns=cols)
+        data = [d for col in cols for d in col]
+        data_t = [[data[i][j] for i in range(len(data))]
+                  for j in range(len(data[0]))]
```
janezd: Isn't this slow? Can you check how long it takes on some large data set? Can you use numpy?

VesnaT: Yes, it is faster with numpy. But from_numpy doesn't allow to build a table with an empty X. Than, maybe I should enable that as well.

janezd: Yes, maybe enable it. ;)

kernc: IIUC, these three lines are equal to:
```py
data_t = np.row_stack(cols)
```
?
It's so damn convoluted I can't tell for sure. >D

janezd:
Load heart_disease. Change the type of "gender" from discrete to string. Instead of values "male" and "female", you get (string) values 1.0 and 0.0.

janezd:
Row colors should correspond to the colors of columns in widget Table, that is, class should be dark and features should be white and not the opposite.

janezd:
The last column shouldn't appear to be editable. (I know, I know. *It's not you, it's me.* :))

Orange/widgets/utils/domaineditor.py
```diff
@@ -0,0 +1,164 @@
...
+                return val
+        if role == QtCore.Qt.DecorationRole:
+            if col == 1:
+                return gui.attributeIconDict[self.vartypes.index(val) + 1]
+        if role == QtCore.Qt.ForegroundRole:
+            if self.variables[row][2] == 3 and col != 2:
+                return QtGui.QColor(160, 160, 160)
+        if role == QtCore.Qt.BackgroundRole:
+            place = self.variables[row][2]
+            return TableModel.ColorForRole.get(place, None)
```
janezd: This should fix the colors. I don't like it, but ... It works.

```
            mapping = [2, 0, 1, None]
            return TableModel.ColorForRole.get(mapping[place], None)
```

Orange/widgets/data/owfile.py
```diff
@@ -249,130 +243,139 @@ def browse_file(self, in_demos=False):
...
+            col_data = list(chain(*self.data[:, orig_var]))
+            if orig_plc == 2:
+                col_data = list(chain(*self.data[:, orig_var].metas))
+            if name == orig_var.name and tpe == type(orig_var):
+                var = orig_var
+            elif tpe == DiscreteVariable:
+                values = list(str(i) for i in set(col_data))
+                var = tpe(name, values)
+                col_data = [values.index(str(x)) for x in col_data]
+            else:
```
janezd:
```
            elif tpe == StringVariable and \
                    isinstance(orig_var, DiscreteVariable):
                var = tpe(name)
                col_data = [orig_var.values[int(x)] for x in col_data]
```

ajdapretnar:
@VesnaT Tested a bit and it looks amazing! That's exactly what I needed for summer school! :+1: :star:

ajdapretnar:
P.S. @janezd Could you also check this PR so we can merge? :smiley:

janezd:
I can, but not before late afternoon. Since the PR contains Vesna's, Anže's and my commits, perhaps someone fourth should merge it. Otherwise, I'll do it in the afternoon anyway.

Orange/widgets/data/owfile.py
```diff
@@ -1,16 +1,23 @@
 import os
+from itertools import chain
+from itertools import count
```
lanzagar: These 2 imports should be on a single line.

Orange/widgets/data/owfile.py
```diff
@@ -1,16 +1,23 @@
...
 from warnings import catch_warnings
+import numpy as np
```
lanzagar: Empty line before numpy import to separate py std lib from external libs.

ajdapretnar:
@janezd Looks like @lanzagar volunteered. :grin:

Orange/widgets/data/owfile.py
```diff
@@ -1,16 +1,23 @@
...
 from Orange.canvas.gui.utils import OSX_NSURL_toLocalFile
+from Orange.data import Domain
 from Orange.widgets import widget, gui
-from Orange.widgets.settings import Setting, ContextHandler, ContextSetting
+from Orange.widgets.settings import Setting, ContextHandler, ContextSetting, \
+    PerfectDomainContextHandler
+from Orange.widgets.utils.domaineditor import DomainEditor
 from Orange.widgets.utils.itemmodels import PyListModel
 from Orange.widgets.utils.filedialogs import RecentPathsWComboMixin
+from Orange.data import DiscreteVariable, StringVariable
```
lanzagar: This could also be merged with the import of Domain (from the same module) above.

It is not so noticeable, because the imports are not sorted - could you do that as well, while you are changing things here (sort everything inside the 3 groups of imports).

Orange/widgets/data/owfile.py
```diff
@@ -249,130 +243,143 @@ def browse_file(self, in_demos=False):
...
+        X, y, m = [], [], []
+        cols = [X, y, m]  # Xcols, Ycols, Mcols
+        for column, (name, tpe, place, vals, is_con), (orig_var, orig_plc) in \
+            zip(count(), self.editor_model.variables,
+                chain([(at, 0) for at in self.data.domain.attributes],
+                      [(cl, 1) for cl in self.data.domain.class_vars],
+                      [(mt, 2) for mt in self.data.domain.metas])):
+            if place == 3:
+                continue
+            col_data = list(chain(*self.data[:, orig_var]))
```
lanzagar: Should this be inside some if/elif/else also?
Otherwise it seems like it is computed here and overwritten 2 lines below (if orig_plc==2).
(At first glance, at least it can't hurt to put it in an else of the next if, possibly something even better)

Orange/widgets/data/owfile.py
```diff
@@ -74,13 +81,12 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
         doc="Attribute-valued data set read from the input file.")]
 
     want_main_area = False
-    resizing_enabled = False
```
lanzagar: Now that resizing is enabled again:
does line L174 `url_combo.setMaximumWidth(500)` still make sense?

`file_combo` does not have a max set and extends to fill the widget, which would be even more useful for potentially long URLs...

Orange/widgets/utils/domaineditor.py
```diff
@@ -0,0 +1,186 @@
...
+
+class Place:
+    feature = 0
+    class_var = 1
+    meta = 2
+    skip = 3
+
+
+class VarTableModel(QtCore.QAbstractTableModel):
+    places = "feature", "class", "meta", "skip"
```
lanzagar: "class" is used in this widget, but "Target Variable" in Select Columns. We should probably be consistent to avoid confusion.

As I remember, "Target Variable" was chosen there deliberately over "class", because it makes more sense for regression as well.
I think just using "target" instead of "class" here would be fine.

Orange/widgets/utils/domaineditor.py
```diff
@@ -0,0 +1,186 @@
...
+    typenames = "nominal", "numeric", "string"
+    vartypes = DiscreteVariable, ContinuousVariable, StringVariable
```
lanzagar: We should probably include TimeVariable too.
With "time" or "datetime" typename maybe.

Confer with @kernc if more input is needed about this.

lanzagar:
The last column of the editor shows values for nominal vars and is empty for others.
When a user changes to a nominal var from something else (e.g. numeric) it would be nice to see the list of values this would produce.

lanzagar:
The editor does not get cleared/reset after an incorrect file/url is selected - no data is output, but the editor contains information from the previous data.

janezd:
> When a user changes to a nominal var from something else (e.g. numeric) it would be nice to see the list of values this would produce.

This can't be done. :) Now I recalled why we needed the more complicated approach (passing the domain to the loader) instead of @VesnaT's simpler solution: when data is read as numeric, string values are lost. For instance, "0.1" and "0.10" are blended into the same value, as are `12.000` and `12` (what is given by `12` in the file would actually quite probably appear as `12.000` when the type is changed to nominal). This is a bug that can't be avoided in the current implementation, which does not re-read the file.

I suggest that we nevertheless merge this PR (@ajdapretnar needs it) and hope the bug doesn't cause too many problems. However, we add a high priority ticket for fixing this bug.

kernc:
> This is a bug that can't be avoided in the current implementation, which does not re-read the file.

No problem. We will document it.

PR #1320, opened 2016-06-07
OWTestLearners: Add auto apply

rokgomiscek:
Added auto apply

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -158,6 +158,8 @@ class OWTestLearners(widget.OWWidget):
     TARGET_AVERAGE = "(Average over classes)"
     class_selection = settings.ContextSetting(TARGET_AVERAGE)
 
+    auto_apply = settings.Setting(False)
+
     def __init__(self):
         super().__init__()
 
@@ -199,8 +201,7 @@ class OWTestLearners(widget.OWWidget):
         gui.appendRadioButton(rbox, "Test on test data")
 
         rbox.layout().addSpacing(5)
-        self.apply_button = gui.button(
-            rbox, self, "&Apply", callback=self.apply, default=True)
+        self.apply_button = gui.auto_commit(rbox, self, "auto_apply", "&Apply")
 
         self.cbox = gui.vBox(self.controlArea, "Target Class")
         self.class_selection_combo = gui.comboBox(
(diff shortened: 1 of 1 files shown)


Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -573,18 +574,15 @@ def _invalidate(self, which=None):
                         item.setData(None, Qt.DisplayRole)
                         item.setData(None, Qt.ToolTipRole)
 
-        self.apply_button.setEnabled(True)
+        if self.auto_apply:
```
janezd: Skip this `if`.

`gui.auto_commit` replaces the original commit method, which checks this condition.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -329,7 +330,7 @@ def set_preprocessor(self, preproc):
     def handleNewSignals(self):
         """Reimplemented from OWWidget.handleNewSignals."""
         self._update_class_selection()
-        self.apply()
+        self.commit()
```
janezd: I suppose it would be better to call `self.unconditional_commit`. We probably want to force apply here.

kernc: It's sometimes _very_ useful to be able to disable flow propagation in slow Test & Score. Imagine I want to connect multiple stuff and only run the simulation at the end, not wait 20 stupendous seconds between my every action.

Whether the widgets always auto apply when the input signal has new data is something that should be uniformly determined and in the guidelines?

janezd: I believe that at least when the schema is first loaded, all signals need to be processed.

Also, I think that when the widget receives data, it should process and output it. "Apply" button should refer to application of changes in the user interfaces. (This could be different if the name is "Commit" and "Send", which imply sending the data, not just applying changes.)

kernc: I agree it's not unreasonable. But just saying that recently in Orange 2, it proved useful to be able to disable automatic recalculation when connecting multiple classifiers.

ajdapretnar:
@rokgomiscek Could you please discuss with @janezd or @kernc on the commit/apply issue and fix the PR? We would preferably merge this by tomorrow. :smile: Thanks!

janezd:
I also wanted to play a bit with the layout (the box around the button, the button's name), so I added another commit and moved yours to PR #1344.

PR #1325, opened 2016-06-10
[FIX] gui.auto_commit: Fix crash when caller gives  argument

janezd:
`addToLayout` was given twice, both in `**misc` and as keyword argument.


astaric:
How can I reproduce the crash?

PR #1328, opened 2016-06-10
Discretize: Improve gui

janezd:
To consider: the orders of options for general and individual setting are no longer the same. We can

- keep them different
- reorder them and disregard the compatibility with saved settings
- reorder them and keep the compatibility, which would require having some trivial, but annoying code for reshuffling.

Before:
screen shot 2016-06-10 at 12 38 28

After:
screen shot 2016-06-10 at 12 36 19

### DIFF
--- a/Orange/widgets/data/owdiscretize.py
+++ b/Orange/widgets/data/owdiscretize.py
@@ -161,7 +161,11 @@ class OWDiscretize(widget.OWWidget):
         box = gui.vBox(self.controlArea, self.tr("Default Discretization"))
         self.default_bbox = rbox = gui.radioButtons(
             box, self, "default_method", callback=self._default_disc_changed)
-
+        rb = gui.hBox(rbox)
+        self.left = gui.vBox(rb)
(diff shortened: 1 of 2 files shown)


astaric:
Can we make it more obvious that number of intervals applies to both equal freq an equal width?

Regarding the settings, I currently do not see, how we could reshuffle old settings. How do you distinguish between old and new settings?

janezd:
In the upper windows it's kind of obvious (at least *obviouser*). In specific options, I could reorder the settings so that we have "Default" then some space, then the two options that have the number of intervals + the slider, than some more space, and then the rest of the options.

As for old and new settings, I currently fill the radio button group box in the old order and then reshuffle the widgets within the layout. This should keep the old enumeration. I could do the same for specific options.

An alternative is to move/hide the slider.

kernc:
> An alternative is to move/hide the slider.

Hide/disable.

janezd:
It now moves/disables. I don't want to hide; even with just moving, the widget resizes a bit.

PR #1333, opened 2016-06-12
[FIX] Pass data attributes after preprocess.

BlazZupan:
Data table can include "attributes" attribute that stores object to be passed along with the data. SklImpute created a new data set from numpy objects, thus not storing the "attributes" from original data table. This has now been fixed. 

I was not able to find a better, general fix that would modify Preprocess class to do so for any kind of preprocessing.

The unwanted behavior fixed in this pull request was found through combination of Distance -> MDS widgets where the "attributes" attribute of the incoming data was lost when selecting data from MDS, and the loss was contributed to management of the data in Distance widget.


astaric:
The rest of the preprocessors use Table.from_table, so they should work already.

PR #1339, opened 2016-06-15
Distances: Add Mahalanobis distance.

thocevar:


Orange/distance/__init__.py
```diff
@@ -152,3 +152,29 @@ def __call__(self, e1, e2=None, axis=1, impute=False):
 
 PearsonR = PearsonDistance(absolute=False)
 PearsonRAbsolute = PearsonDistance(absolute=True)
+
+
+class Mahalanobis():
+    """Mahalanobis distance."""
+    def __init__(self, data, axis=1):
+        """
+        :param data: The dataset used for calculating covariances.
```
kernc: You can set PyCharm to generate docstrings in preferred format in Settings > Tools > Python Integrated Tools > Docstring format.

kernc:
Does it make sense to add this distance in the Distances widget as well?

PR #1342, opened 2016-06-15
[ENH] OWPaintData: Fix and improve handling of input data

janezd:
Paint data crashed when the data contained no (discrete) class. @ajdapretnar, this is to-do ticket 311. Please check.

The widget used to label classes as C1, C2 ... instead of using the labels present in the data.

The widget didn't disable attr2 (and probably crashed) if the data contained only a single attribute.

The widget used the input data for setting the data once, but then the only way to reset it was to re-emit the signal. I added a button Reset to Input Data, which is enabled when the input data is present. (I initially showed/disabled it, but prefer enabling/disabling since the presence of the button hints to the user that she can feed the data into the widget.)

The widget now also issues proper information and warning messages.

I removed redundant import (`File Format`).

### DIFF
--- a/Orange/widgets/data/owpaintdata.py
+++ b/Orange/widgets/data/owpaintdata.py
@@ -24,7 +24,6 @@ import Orange.data
 from Orange.widgets import widget, gui
 from Orange.widgets.settings import Setting
 from Orange.widgets.utils import itemmodels, colorpalette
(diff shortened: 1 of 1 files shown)


ajdapretnar:
@janezd Tested all sorts of crazy scenarios and everything works now! :grin: Thank you! Merging.

PR #1352, opened 2016-06-20
Changed defaults in Test&Score

ajdapretnar:
Probably we want to keep all the defaults to `auto_apply = True`. No?

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -158,7 +158,7 @@ class OWTestLearners(widget.OWWidget):
     TARGET_AVERAGE = "(Average over classes)"
     class_selection = settings.ContextSetting(TARGET_AVERAGE)
 
-    auto_apply = settings.Setting(False)
+    auto_apply = settings.Setting(True)
 
     def __init__(self):
         super().__init__()


janezd:
Not necessarily and particularly not here, especially with the current implementation of the widget.

Construct a simple schema (File, Test & Score, Bayes). Set auto apply on. Select random sampling and drag the slider for Relative training set size left and right for some time. After you're done, the widget will still run the bayes for quite some time, for every size you dragged across.

This is easily fixed by disabling tracking on the slider, so that learning is run only when the user releases the slider. The two spin boxes above are a problem. We can have the extra enter-like button, but the resulting layout may be really bad. Or we can add a timeout - learning is run when the user stops changing the spin, that is, one second after the last change...

ajdapretnar:
I think the idea is to quickly prototype with defaults. So when I connect the widgets, the data is propagated instantly. 
Selecting random sampling and dragging sliders around is already a 'parameter setting' step and can be fixed by disabling 'auto-apply' first. From the user standpoint, I think this would be a better solution. In this case, could we just disable 'auto-apply' when selecting 'random sampling' or fix the recalculation as mentioned above?
(Having to tick auto-apply every single time really bugs me, I have to admit.)

janezd:
Spin should not disable auto-apply. We need to find a better solution.

We don't really need a spin box there --- 7-fold CV is kind of weird, nobody ever used it, I suppose. Perhaps we can have a combo with values 2, 3, 5, 10, 20. With that, we can even get rid of the apply button and always apply the change.

ajdapretnar:
@janezd So what have we decided on auto-apply=True? I still go for enabling the user to quickly connect the components without having to open the widget and select 'auto-apply', but I agree we should go with a proper fix for re-computation. Should I close the PR and we're going to sort it on Friday?

janezd:
Continued in #1362.

PR #1353, opened 2016-06-20
[FIX] OWSelectColumns: auto-apply

matevzkren:
Beforehand, changes on auto-apply were not reflected correctly.
Drag&drop did not work as expected. The fix includes calls
to `self.update_completer_model()` whenever either of the
VariablesListItemModels is changed (`rowsRemoved`).


ajdapretnar:
@matevzkren That's exactly what I needed, however, a small glitch is still there (at least for me). Use iris and drag whatever attribute you want to metas. Open Data Table to see the real-time result. When dragging attributes out of metas, they don't seem to update. It only happens when dragging from metas to elsewhere. Other fields look good.

ajdapretnar:
Wunderbar. :+1:

PR #1358, opened 2016-06-20
[FIX] Revert PyListModel display role change

ales-erjavec:
Granted the Qt documentation can be a bit misleading at times, however the Qt's item view framework does not impose the restriction that the data associated with a DisplayRole be a string. After all a MVC abstraction is intended to separate the data from its presentation.

For instance there are many situations where preserving the type of the item intended for display is beneficial (when displaying numerical data, dates, ...), as the view can then benefit from the default behavior of QSortFilterProxyModel model and have proper sorting with the default configuration out of the box (i.e. no `setSortRole` necessary, as the item to be displayed is also the one defining the ordering), locale specific date formating, decimal separator, .... (courtesy of `QStyleItemDelegate.displayText`). 
In extreme cases the data presentation is not even textual (https://doc.qt.io/qt-4.8/qt-itemviews-pixelator-example.html, https://doc.qt.io/qt-4.8/qt-itemviews-chart-example.html)

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -273,14 +273,16 @@ def variable_icon(dtype):
         return QtGui.QIcon()
 
 
+class FeatureItemDelegate(QtGui.QStyledItemDelegate):
+    def displayText(self, value, locale):
+        return value.name + " := " + value.expression
+
+
 class DescriptorModel(itemmodels.PyListModel):
     def data(self, index, role=Qt.DisplayRole):
         if role == Qt.DecorationRole:
             value = self[index.row()]
             return variable_icon(type(value))
-        elif role == Qt.DisplayRole:
-            value = self[index.row()]
-            return value.name + " := " + value.expression
         else:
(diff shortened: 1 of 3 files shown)


kernc:
If some `PyListModel` is made to hold arbitrary objects, it can't be used with any view widget (QListView, QComboBox) out of the box.

Why is it desirable for data associated with `Qt.DisplayRole` to not be a string? The [documentation](https://doc.qt.io/qt-4.8/qt.html#ItemDataRole-enum) is pretty clear:
> The roles are used by the view to indicate to the model which type of data it needs.  Custom models should return data  in these types.
> Qt::DisplayRole ... QString.

ales-erjavec:
> Why is it desirable for data associated with Qt.DisplayRole to not be a string? The documentation is pretty clear: ...

See the updated first comment.

kernc:
Seems like the behavior is undefined when the returned value is not a QVariant. I see how it makes it easier to just pass around the original data, and to make the model compatible with Q*View, it's easy enough to subclass the model or use a delegate.

PR #1362, opened 2016-06-21
Test score - Changes in GUI

janezd:
Continues from #1352 -- see the discussion there.

In a nutshell, we would like to have auto-apply as default, but for this we need to disable tracking and change spins into combos.

Class attributes were renamed to avoid compatibility problems with older saved settings.

@kernc, please comment on 820d34c8bd1072f98978fb7209d3b6e7f91287d1. I think we need something like this; the implementation is a bit of a hack, but this is easier than passing arguments to constructor. Besides, it fixes all controls in `gui` with a single change.

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -186,7 +186,7 @@ class OWComponent:
 
 def miscellanea(control, box, parent,
                 addToLayout=True, stretch=0, sizePolicy=None, addSpace=False,
-                disabled=False, tooltip=None):
+                disabled=False, tooltip=None, **kwargs):
     """
     Helper function that sets various properties of the widget using a common
     set of arguments.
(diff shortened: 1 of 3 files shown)


Orange/widgets/gui.py
```diff
@@ -228,6 +228,8 @@ def miscellanea(control, box, parent,
     :param sizePolicy: the size policy for the box or the control
     :type sizePolicy: PyQt4.QtQui.QSizePolicy
     """
+    for prop, val in kwargs.items():
+        getattr(control, "set" + prop[0].upper() + prop[1:])(val)
```
kernc: Perhaps a line of comment would be nice.

kernc:
I don't mind it. It is future compatible. :stuck_out_tongue: 
With tracking disabled, the slider's label only updates on release.

Orange/widgets/gui.py
janezd: I now added it to the docstring and the general documentation for the module.

PR #1364, opened 2016-06-21
[ENH] [WIP] Orange.data: argmaxrnd

matevzkren:
A cython function that returns the index of the maximum value for a given 1D numpy array. As is, the code only interprets system default data types and raises a ValueError otherwise. ~~Alternatively, `argmaxrnd_all' can be called (included in the commit).~~ Without exact typing, it runs very slow.

As recommended below, benchmarks will be done.


kernc:
Please paste some [%timeit](https://stackoverflow.com/questions/29280470/what-is-timeit-in-python) benchmarks comparing `numpy.argmax`, `bottleneck.nanargmax`, and your implementation's run time.

Then please also provide any qualitative comparisons that you feel could help you justify this case.

janezd:
I can provide the justification: in case of ties, we sometimes (= often = always in Orange 2) need a random index, not the first (or the last?), which is returned by `numpy` methods. As for speed, I'd just check whether the Cython code compiles into C without any unnecessary native Python types.

kernc:
So just in case anybody hasn't tried it yet; on my machine:
```py
>>> R = np.random.random(1e4)
>>> %timeit argmaxrnd(R)
100000 loops, best of 3: 12.2 µs per loop
>>> %timeit np.random.choice((R == np.nanmax(R)).nonzero()[0])
100000 loops, best of 3: 18.1 µs per loop

>>> R = np.random.random(1e5)
>>> %timeit argmaxrnd(R)
10000 loops, best of 3: 120 µs per loop
>>> %timeit np.random.choice((R == np.nanmax(R)).nonzero()[0])
10000 loops, best of 3: 107 µs per loop
```
Never mind. This is just for me to understand what kind of optimization we are dealing with again.

janezd:
@matevzkren also computes hash from the data to fix the random seed. You're right that we don't gain much. But since we already use Cython for other purposes, it doesn't hurt using it here, too.

Orange/data/_argmaxrnd.pyx
```diff
@@ -0,0 +1,104 @@
...
+        Py_ssize_t i
+        Py_ssize_t m_i
+        int c
+        np.int_t curr
+        np.int_t m
+
+    m = vec[0]
+    c = 1
+    m_i = 0
+    for i from 1 <= i < vec.shape[0]:
```
kernc: This is [deprecated syntax](http://docs.cython.org/src/userguide/language_basics.html#integer-for-loops).

Orange/data/_argmaxrnd.pyx
```diff
@@ -0,0 +1,104 @@
...
+    cdef:
+        Py_ssize_t i
+        Py_ssize_t m_i
+        int c
+        np.int_t curr
+        np.int_t m
+
+    m = vec[0]
+    c = 1
+    m_i = 0
```
kernc: You can initialize values in declarations above.

Orange/data/_argmaxrnd.pyx
```diff
@@ -0,0 +1,104 @@
...
+        elif curr == m:
+            c += 1
+            if 1 + <int>(1.0 * c * rand() / RAND_MAX) == c:
+                m_i = i
+    return m_i
+
+
+@boundscheck(False)
+@cdivision(True)
+cdef Py_ssize_t argmaxrnd_float(np.ndarray[np.float_t, ndim=1] vec):
```
kernc: You can use [fused types](http://docs.cython.org/src/userguide/fusedtypes.html) to mimic C++ templates and have one single function work with either type of array.

Orange/data/_argmaxrnd.pyx
```diff
@@ -0,0 +1,104 @@
...
+        int c
+        np.float_t curr
+        np.float_t *m
+
+    m = NULL
+    for i from 0 <= i < vec.shape[0]:
+        curr = vec[i]
+        if not isnan(curr):
+            if not m:
+                m = <np.float_t *>malloc(1 * sizeof(np.float_t))
```
kernc: Why does this need to be allocated? Why can't it be handled on the stack like in the argmaxrnd_int case?

Orange/data/_argmaxrnd.pyx
```diff
@@ -0,0 +1,104 @@
...
+
+def argmaxrnd(np.ndarray vec not None, object random_seed=None):
+
+    """
+    Returns the index of the maximum value for a given 1D array.
+    In case of multiple indices corresponding to the maximum value,
+    the result is chosen randomly among those. The random number
+    generator, used by the C lang function 'rand', can be seeded
+    by forwarding a hashable python object. If -1 is passed, the
+    input array is hashed instead.
```
kernc: Please add a comment that what this does is actually:
```py
np.random.choice((vec == np.nanmax(vec)).nonzero()[0])
```
where one can optionally pass a buffer-protocol-compliant object as random seed to seed the choice. Try to make it sound less trivial.

Please also explain in the comment **why** is _this_ used as opposed the line above
and **where**.

Orange/data/setup.py
```diff
@@ -13,22 +13,13 @@ def configuration(parent_package='', top_path=None):
...
-                         sources=['_io.c'],
-                         include_dirs=[numpy.get_include()],
-                         libraries=libraries)
-    config.add_extension('_variable',
-                         sources=['_variable.c'],
-                         include_dirs=[numpy.get_include()],
-                         libraries=libraries)
+
+    for source in ('_valuecount.c', '_contingency.c', '_io.c',
+                   '_variable.c', '_argmaxrnd.c',):
```
kernc: Perhaps add this new file closer to where it's used. Does it fit well in Orange.data namespace?

Orange/data/_argmaxrnd.pyx
matevzkren: Fused types turned out (significantly) slower. If optimisation is goal, this should stay.

kernc:
As @matevzkren proposes, closing in favor of:
```py
def argmaxrnd(X, random_state=None):
    random = np.random if random_state is None else np.random.RandomState(random_state)
    return random.choice((X == bn.nanmax(X)).nonzero()[0])
```

PR #1366, opened 2016-06-22
[ENH] Scatter plot and MDS: Show labels only for selected points

janezd:
The ticket was about doing this for all widgets. I found only these two.

The first and the thirs commit implement the option for the two widgets.

The second commit reorganizes the Scatter plot GUI. The original layout had separate boxes for combos and for checkboxes. The new checkbox so obviously belongs to the label combo that I've put it there. By the same logic, other checkboxes belong to different places in the widget. I'm not totally satisfied with the result; I think it's better than the original (cleaner and smaller), but I can revert it if you dislike it.

Original:
screen shot 2016-06-22 at 17 48 44

Changed:
screen shot 2016-06-22 at 17 48 47

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -141,6 +141,7 @@ class OWMDS(widget.OWWidget):
     shape_value = settings.ContextSetting("")
     size_value = settings.ContextSetting("")
     label_value = settings.ContextSetting("")
+    label_only_selected = settings.Setting(False)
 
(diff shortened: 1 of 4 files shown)


ajdapretnar:
My vote goes for the original. I especially like the boxes in the control area, which make everything organized. And also the order for me worked well. My 2 cents. :smiley:

ajdapretnar:
@janezd Have we reached a consensus on GUI or am I the only one voting here? :)

PR #1373, opened 2016-06-23
[ENH] OWColors: Implement copy-to-all for continuous pallettes

janezd:
@ajdapretnar, was this (ticket 301) your idea?

Please see whether you like this implementation.

It is only for continuous attributes. For discrete, it may not make much sense since the number of values varies across features.


ajdapretnar:
@janezd Yes, this was my idea. I simply thought setting a color palette one variable at the time is awful. :smiley: I like the implementation, I think it makes sense (and yes, it was meant only for continuous indeed).
I'm merging.

ajdapretnar:
@janezd Oh wait, what do we do with 'Ridicule Lint' commit? :)

PR #1374, opened 2016-06-24
Addons: Add Orange3-Timeseries to OFFICIAL_ADDONS

kernc:


kernc:
PyPI search apparently works.

PR #1375, opened 2016-06-24
[FIX] Fix and error in `PyListModel.__setitem__`

ales-erjavec:
Before it would overwrite elements which were not supposed to be overwritten when setting a slice in the middle of a sequence.

Add tests.

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -475,6 +475,7 @@ class PyListModel(QAbstractListModel):
             start, stop, step = _as_contiguous_range(s, len(self))
             self.beginRemoveRows(QModelIndex(), start, stop - 1)
         else:
+            s = operator.index(s)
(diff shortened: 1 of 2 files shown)


Orange/widgets/utils/itemmodels.py
```diff
@@ -490,13 +491,14 @@ def __setitem__(self, s, value):
                 value = list(value)
             separators = [start + i for i, v in enumerate(value) if v is self.Separator]
             self.beginInsertRows(QModelIndex(), start, start + len(value) - 1)
-            self._list[s] = value
```
kernc: I don't understand, when would this overwrite elements that were not supposed to be overwritten?

ales-erjavec: The 'items' corresponding to the slice which should have been overwritten were already removed 5 lines before.

kernc: Thanks.

PR #1385, opened 2016-06-26
[FIX] Remove '(from Orange3)' from widget names in canvas

janezd:
Tooltips on widget toolbars showed "(from Orange3)" after the name of each widget due to mismatching project name.

@ales-erjavec, @astaric, is there a better, long-term solution? Comparing with `Orange.setup.NAME` doesn't seem appropriate - or is it?

### DIFF
--- a/Orange/canvas/registry/qt.py
+++ b/Orange/canvas/registry/qt.py
@@ -296,7 +296,7 @@ def tooltip_helper(desc):
     tooltip = []
     tooltip.append("<b>{name}</b>".format(name=escape(desc.name)))
 
-    if desc.project_name and desc.project_name != "Orange":
+    if desc.project_name:
         tooltip[0] += " (from {0})".format(desc.project_name)
 
     if desc.description:


kernc:
_setup.py_ doesn't get installed. You could check for existence of '-':
```py
desc.project_name and (not desc.project_name.startswith('Orange') or '-' in desc.project_name)
```

janezd:
Why `'-' in desc.project_name`?

kernc:
So the addendum won't show for Orange core widgets but will show for widgets from add-ons, normally named Orange-Something. It's what the PR is about, no?

janezd:
Changed as told. And let's please attribute my question to sleep deprivation or something. Thanks.

Orange/canvas/registry/qt.py
```diff
@@ -296,7 +296,8 @@ def tooltip_helper(desc):
     tooltip = []
     tooltip.append("<b>{name}</b>".format(name=escape(desc.name)))
 
-    if desc.project_name and desc.project_name != "Orange":
+    if desc.project_name and (not desc.project_name.startswith('Orange') or
```
astaric: We could also remove the additional condition and just show (from package_name) for all projects. This might come in handy if (when) we decide to separate canvas and core widgets.

janezd:
Done.

PR #1386, opened 2016-06-26
[ENH] Implement copying graph to clipboard using Ctrl-C (Cmd-C)

janezd:
This should work for all widgets that have `graph_name` attribute (~ the Save Image button).

Can somebody check it on Linux and Windows? Scatter plot and Box plot, for instance.

@VesnaT, you're an expert for Linear Projection - and saving graphs in general. Can you check why it copying doesn't work for Linear Projection and Distributions?

### DIFF
--- a/Orange/widgets/classify/owtreeviewer2d.py
+++ b/Orange/widgets/classify/owtreeviewer2d.py
@@ -361,7 +361,7 @@ class OWTreeViewer2D(OWWidget):
     _DEF_NODE_WIDTH = 24
     _DEF_NODE_HEIGHT = 20
 
-    graph_name = True
+    graph_name = "scene"
 
     def __init__(self):
         super().__init__()
--- a/Orange/widgets/widget.py
+++ b/Orange/widgets/widget.py
@@ -18,6 +18,7 @@ from Orange.widgets import settings, gui
 from Orange.canvas.registry import description as widget_description
 from Orange.canvas.report import Report
 from Orange.widgets.gui import ControlledAttributesDict, notify_changed
+from Orange.widgets.io import ClipboardFormat
 from Orange.widgets.settings import SettingsHandler
 from Orange.widgets.utils import saveplot, getdeepattr
 from .utils.overlay import MessageOverlayWidget
@@ -198,6 +199,9 @@ class OWWidget(QDialog, Report, metaclass=WidgetMetaClass):
         sc = QShortcut(QKeySequence(Qt.ShiftModifier | Qt.Key_F1), self)
         sc.activated.connect(self.__quicktip)
 
+        sc = QShortcut(QKeySequence.Copy, self)
+        sc.activated.connect(self.copy_to_clipboard)
+
         return self
 
     def __init__(self, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


Orange/widgets/widget.py
```diff
@@ -198,6 +199,9 @@ def __new__(cls, *args, **kwargs):
         sc = QShortcut(QKeySequence(Qt.ShiftModifier | Qt.Key_F1), self)
         sc.activated.connect(self.__quicktip)
 
+        sc = QShortcut(QKeySequence(Qt.ControlModifier | Qt.Key_C), self)
```
kernc:
```py
QKeySequence(QKeySequence.Copy)
```

kernc:
Currently segfaults on GNU/Linux.

Orange/widgets/widget.py
```diff
@@ -316,6 +320,12 @@ def save_graph(self):
             return
         saveplot.save_plot(graph_obj, self.graph_writers)
 
+    def __copy_to_clipboard(self):
+        graph_obj = getdeepattr(self, self.graph_name, None)
+        if graph_obj is None:
+            return
+        ClipboardFormat.write_image(None, graph_obj)
```
kernc: Can't step through it after here. :confused:

janezd: With which visualization? All? Perhaps you should ask @VesnaT. Linear projection report used to fail on Os X in a similar call, and she fixed it.

kernc: Interesting. Doesn't segfault anymore, yet nothing has changed code-wise. But anyway, now the images are borked:
distributions

janezd: Can you try now? For me it now works on all visualizations.

kernc: Works here too.

janezd: Can somebody (@thocevar?) try this PR on Windows?

ajdapretnar:
@janezd Tested it. Works on all visualizations, apart from Classification Tree Viewer, which I reckon is of a different type. It also works on Word Cloud in Text add-on, but not on GeoMap. Networks don't work either. But the rest work like a charm, so it's up to you to decide whether we patch those as well or not. :smile: Still, I'm over the moon with this one! :)

janezd:
I fixed tree view. If this is the only one that doesn't work, we're done here.

As for add-ons, GeoMap does not seem so trivial. Networks should be simple, but if I apply the same fix as for trees, the points and edges are misaligned.

Orange/widgets/classify/owtreeviewer2d.py
```diff
@@ -551,3 +551,6 @@ def save_graph(self):
         save_plot(data=dict(scene=self.scene, tree=self.tree),
                   file_formats=dict(chain(FileFormat.img_writers.items(),
                                           FileFormat.graph_writers.items())))
+
+    def copy_to_clipboard(self):
+        ClipboardFormat.write_image(None, self.scene)
```
kernc: Might setting `OWTreeViewer2D.graph_name = 'scene'` above also work.

kernc:
Both are pretty trivial. Just a matter of assigning `OWNxExplorer.graph_name = 'view'` and `OWGeoMap.graph_name = 'webview'`. Works for me.

janezd:
I changed the tree widget as suggested.

For networks, the problem was that I had an old version of the widget. With the last version, `OWNxExplorer.graph_name` works for me, too. But it's in add-on, so it doesn't belong to this PR.

kernc:
Do you want to push the changes and remove the WIP tag?

janezd:
Sorry, I forgot. I pushed them now.

PR #1389, opened 2016-06-30
[FIX] OWLearnerWidget: Don't crash when no features 

nikicc:
If we provide a data set without any features (as unpreprocessed Corspus) learners crashed with a red screen. This PR adds a check for this and shows an error message instead.

### DIFF
--- a/Orange/widgets/utils/owlearnerwidget.py
+++ b/Orange/widgets/utils/owlearnerwidget.py
@@ -186,6 +186,9 @@ class OWBaseLearner(OWWidget, metaclass=OWBaseLearnerMeta):
                 self.error(self.DATA_ERROR_ID,
                            "Data contains a single target value. "
                            "There is nothing to learn.")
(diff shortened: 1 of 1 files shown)


Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -186,6 +186,9 @@ def check_data(self):
                 self.error(self.DATA_ERROR_ID,
                            "Data contains a single target value. "
                            "There is nothing to learn.")
+            elif self.data.X.size == 0:
+                self.error(self.DATA_ERROR_ID,
+                           "Found array with 0 feature(s) while a minimum of 1 is required.")
```
kernc: "Data has no features to learn from."? Zero features are plural. :stuck_out_tongue:

nikicc: I was being sloppy and just copied the error from scikit 🙈 Corrected.

PR #1395, opened 2016-07-01
[ENH] Automatically create required SQL extensions

### LINKED ISSUE #1388: sample_time requires installation of extension in orange3/Orange/data/sql/table.py
##### Orange version

3.3.5

##### Expected behavior

Sampling of large table for attribute discovery
screen shot 2016-06-29 at 08 48 43

##### Actual behavior

Error that method system_time doesn't exist

Connect to postgres database, Answer "Yes, on a sample" to the popup box.

screen shot 2016-06-29 at 08 48 28

The solution is to [create the system_time extension](https://www.postgresql.org/docs/9.5/static/tsm-system-time.html) in Postgres 9.5+ (not available before then).

Suggested solution is to check for Postgres 9.5+ and the existence of the system_time method. If it doesn't exist, create the method using the following:

`CREATE EXTENSION tsm_system_time;`

After installing the extension, the sampling works properly.

lanzagar:
Try to create the required extensions if they do not exist on the selected database.
Show error if unable to do so.

Fixes #1388

### DIFF
--- a/Orange/widgets/data/owsql.py
+++ b/Orange/widgets/data/owsql.py
@@ -133,16 +133,17 @@ class OWSql(widget.OWWidget):
 
     def error(self, id=0, text=""):
         super().error(id, text)
+        err_style = 'QLineEdit {border: 2px solid red;}'
         if 'server' in text or 'host' in text:
-            self.servertext.setStyleSheet('QLineEdit {border: 2px solid red;}')
+            self.servertext.setStyleSheet(err_style)
         else:
             self.servertext.setStyleSheet('')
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owsql.py
```diff
@@ -176,6 +178,7 @@ def connect(self):
 
     def refresh_tables(self):
         self.tablecombo.clear()
+        [self.error(42 + i) for i in range(len(EXTENSIONS))]
```
kernc: This is not clever. :stuck_out_tongue:

PR #1396, opened 2016-07-01
[FIX] PCA: Fix the variance spin.

thocevar:

### DIFF
--- a/Orange/widgets/unsupervised/owpca.py
+++ b/Orange/widgets/unsupervised/owpca.py
@@ -262,7 +262,7 @@ class OWPCA(widget.OWWidget):
             y = curve[cutidx]
             line.setData([-1, cutidx], 2 * [y])
             label.setPos(cutidx, y)
-            label.setPlainText("{:.2f}".format(y))
+            label.setPlainText("{:.3f}".format(y))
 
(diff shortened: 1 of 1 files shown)


Orange/widgets/unsupervised/owpca.py
```diff
@@ -262,11 +262,11 @@ def _set_horline_pos(self):
             y = curve[cutidx]
             line.setData([-1, cutidx], 2 * [y])
             label.setPos(cutidx, y)
-            label.setPlainText("{:.2f}".format(y))
+            label.setPlainText("{:.3f}".format(y))
 
     def _on_cut_changed(self, line):
         # cut changed by means of a cut line over the scree plot.
-        value = int(round(line.value()))
+        value = int(line.value())
```
janezd: I don't like this change: the value should be rounded, not rounded down.

janezd:
I understand why three decimals -- to explain the difference due to rounding.

I don't understand what was wrong with the line. Can you tell me how to reproduce the problem that this PR fixes?

thocevar:
The displayed number of components does not match with the cut line (it's sometimes off by one) when changing the variance spin.

PR #1397, opened 2016-07-01
[ENH] Rule induction (CN2)

matevzkren:
A more general framework of replaceable individual components that can be fine-tuned to specific needs. To induce rules from examples, divide-and-conquer strategy is applied. CN2 and CN2Unordered algorithms have been implemented so far.. Next up -> widgets!


Orange/classification/rules.py
```diff
@@ -0,0 +1,938 @@
...
+    An array of indices corresponding to the maximum value of each row
+    is then returned.
+    """
+    if a.ndim > 2:
+        raise ValueError("argmaxrnd only accepts arrays of up to 2 dim")
+
+    random = (np.random if random_seed is None
+              else np.random.RandomState(random_seed))
+
+    def f(x): return random.choice((x == np.nanmax(x)).nonzero()[0])
```
kernc: Please use some vspace here.

You can `import bottlechest as bn; bn.nanmax()` for 50% faster.

Orange/classification/rules.py
```diff
@@ -0,0 +1,938 @@
...
+    return f(a) if a.ndim == 1 else np.apply_along_axis(f, axis=1, arr=a)
+
+
+def entropy(x):
+    x = x[x != 0]
+    x = x / np.sum(x, axis=0)
```
kernc: `x /= x.sum(0)`

Orange/classification/rules.py
```diff
@@ -0,0 +1,938 @@
...
+    x = -x * np.log2(x)
```
kernc: `x *= -np.log2(x)`

Orange/classification/rules.py
```diff
@@ -0,0 +1,938 @@
...
+    return np.sum(x)
```
kernc: `x.sum()`

Orange/classification/rules.py
```diff
@@ -0,0 +1,938 @@
...
+
+
+def likelihood_ratio_statistic(x, y):
+    x[x == 0] = 1e-5
+    y[y == 0] = 1e-5
+    lrs = np.sum(x * np.log(x / y))
+    lrs = 2 * (lrs - np.sum(x) * np.log(np.sum(x) / np.sum(y)))
```
kernc: If `x` and `y` are arrays, which they are`, you can call `.sum()` methods directly, which is marginally faster and shorter to write.

Orange/classification/rules.py
```diff
@@ -0,0 +1,938 @@
...
+        The source array (observed classification).
+    domain : Orange.data.domain.Domain
+        Data domain.
+
+    Returns
+    -------
+    dist : ndarray, int
+        Distribution of classes.
+    """
+    return np.bincount(Y.astype(dtype=np.int),
```
kernc: If this is called a lot, you should ensure `Y.dtype` is integer to begin with.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+    An array of indices corresponding to the maximum value of each row
+    is then returned.
+    """
+    if a.ndim > 2:
+        raise ValueError("argmaxrnd only accepts arrays of up to 2 dim")
+
+    random = (np.random if random_seed is None
+              else np.random.RandomState(random_seed))
+
+    def f(x): return random.choice((x == bn.nanmax(x)).nonzero()[0])
```
kernc: Return on a separate line. Thanks.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+    ----------
+    a : array_like
+        The source array.
+
+    Returns
+    -------
+    hash : int
+        Hash function result.
+    """
+    return int(sha1(bytes(a)).hexdigest(), base=16) & 0xffffffff
```
kernc:
```py
return hash(bytes(a))
```
Thanks.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+        raise NotImplementedError("descendants of Evaluator must "
+                                  "overload method 'evaluate_rule'")
+
+
+class EntropyEvaluator(Evaluator):
+    def evaluate_rule(self, rule):
+        x = rule.curr_class_dist.astype(dtype=np.float)
+        if rule.target_class is not None:
+            x = np.array([x[rule.target_class],
+                          np.sum(x) - x[rule.target_class]])
```
kernc: If this is called a lot, `x[list_of_indices]` makes a copy, so you can cache it in a variable to avoid making it twice.

`x.sum()`

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+    chance.
+    """
+    def __init__(self, alpha=1.0):
+        self.alpha = alpha
+
+    def validate_rule(self, rule):
+        if self.alpha >= 1.0 or rule.parent_rule is None:
+            return True
+
+        x = rule.curr_class_dist.astype(dtype=np.float)
```
kernc: You can use Python's types (int, float) and Numpy does it correctly. I find it more readable without superfluous `np.` everywhere.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+        y = rule.parent_rule.curr_class_dist.astype(dtype=np.float)
+        if rule.target_class is not None:
+            x = np.array([x[rule.target_class],
+                          np.sum(x) - x[rule.target_class]])
+            y = np.array([y[rule.target_class],
+                          np.sum(y) - y[rule.target_class]])
```
kernc: If this is called a lot, cache these `x` and `y` subsets.

Orange/classification/rules.py
```diff
@@ -0,0 +1,938 @@
...
+        result[no_clash] = classifications[no_clash]
+        result[resolve_clash] = argmaxrnd(clashes[resolve_clash])
+        return result
+
+
+def main():
+    data = Orange.data.Table('titanic')
+    row_defined = ~np.isnan(data.X).any(axis=1)
+    data.Y = data.Y[row_defined]
+    data.X = data.X[row_defined]
```
sstanovnik: `data = data.from_table_rows(row_defined)`

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+
+        Returns
+        -------
+        candidate_rules : Rule list
+            Chosen rules.
+        rules : Rule list
+            Rules not chosen, i.e. the remainder.
+        """
+        raise NotImplementedError("descendants of SearchAlgorithm must "
+                                  "overload method 'select_candidates'")
```
kernc: In future, you can just `raise NotImplementedError`. This is for programmers anyway; raising this exception makes it obvious the method needs to be reimplemented.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+
+    def filter_rules(self, rules):
+        """
+        Filter rules to be considered in the next iteration.
+
+        Parameters
+        ----------
+        rules : Rule list
```
kernc: Type is, more correctly, list of ...

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+        return rules
+
+    def refine_rule(self, X, Y, candidate_rule):
+        covered_X = X[candidate_rule.covered_examples]
+        possible_selectors = []
+
+        for i, attribute in enumerate(candidate_rule.domain.attributes):
+            column = covered_X[:, i]
+            if attribute.is_discrete:
+                for val in [int(x) for x in set(column)]:
```
kernc: `np.unique(column).astype(int)` will be much faster when `column` is already an array.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+            # to ensure that the covered_examples matrices are of the
+            # same size throughout the rule_finder iteration
+
+            if new_rule.validity:
+                new_rules.append(new_rule)
+
+        return new_rules
+
+
+class Selector:
```
kernc: Make Selector extend namedtuple.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+    """
+    Define a single rule condition.
+    """
+    operators = {
```
kernc: Class-level constants should be UPPERCASE.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+        '>=': operator.ge
+    }
+
+    def __init__(self, column, op, value):
+        self.column = column
+        self.op = op
+        self.value = value
+
+    def filter_instance(self, x):
+        return self.operators[self.op](x[self.column], self.value)
```
kernc: If self is a tuple, `return self.OPERATORS[self[1]](x[self[0]], self[2])` will be somewhat faster.

Orange/classification/rules.py
kernc: You don't need `.astype(int)` anyway.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+        (target_class, selectors, domain, prior_class_dist,
+         quality_evaluator, complexity_evaluator, significance_validator,
+         general_validator) = candidate_rule.seed()
+
+        for curr_selector in possible_selectors:
+            copied_selectors = copy(selectors)
+            copied_selectors.append(curr_selector)
+
+            new_rule = Rule(selectors=copied_selectors,
+                            parent_rule=candidate_rule, domain=domain,
```
kernc: domain= on its own line.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+
+    def filter_instance(self, x):
+        return self.operators[self.op](x[self.column], self.value)
+
+    def filter_data(self, X):
+        return self.operators[self.op](X[:, self.column], self.value)
+
+    def __eq__(self, other):
+        return all((self.op == other.op, self.value == other.value,
+                    self.column == other.column))
```
kernc: You only have three conds:
```py
cond1 and cond2 and cond3
```

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+            self.covered_examples = np.copy(self.parent_rule.covered_examples)
+            start = len(self.parent_rule.selectors)
+        else:
+            self.covered_examples = np.ones(X.shape[0], dtype=np.bool)
+            start = 0
+        for selector in self.selectors[start:]:
+            self.covered_examples &= selector.filter_data(X)
+
+        self.curr_class_dist = get_dist(Y[self.covered_examples], self.domain)
+        self.validity = self.general_validator.validate_rule(self)
```
kernc:
```py
is_valid = self.is_valid = ...
if is_valid:
```
Prefer boolean flags start with a verb (`is_`, `has_`, `have_`, `want_`, ...).

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+        x : array_like
+            Evaluate this instance.
+
+        Returns
+        -------
+        res : bool
+            True, if the rule covers 'x'.
+        """
+        # return True if the given instance matches the rule condition
+        return all((selector.filter_instance(x) for selector in self.selectors))
```
kernc: FYI, sole-argument generator need not be parenthesized.

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+            An optional list of initial rules to constrain the search.
+        domain : Orange.data.domain.Domain
+            Data domain, used to calculate class distributions.
+
+        Returns
+        -------
+        best_rule : Rule
+            Highest quality rule discovered.
+        """
+        def rcmp(rule): return rule.quality, rule.complexity
```
kernc: vspace

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+
+    The over-fitting of noisy data is avoided by preferring simpler,
+    shorter rules even if the accuracy of more complex rules is higher.
+
+    References
+    ----------
+    .. [1] "Separate-and-Conquer Rule Learning", Johannes Fürnkranz,
+           Artificial Intelligence Review 13, 3-54, 1999
+    """
+    __metaclass__ = ABCMeta
```
kernc: Python 3 can just specify metaclass in the header:
```py
class RuleLearner(Learner, metaclass=ABCMeta):
    ...
```

kernc:
https://stackoverflow.com/questions/12590058/python-performance-with-global-variables-vs-local

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+        return (Y.size == 0 or (target_class is not None and
+                                target_class not in Y))
+
+    @staticmethod
+    def cover_and_remove(X, Y, new_rule):
+        examples_to_keep = np.logical_not(new_rule.covered_examples)
+        return X[examples_to_keep], Y[examples_to_keep]
+
+
+class RuleClassifier(Model):
```
kernc: `class RuleClassifier(Model, metaclass=ABCMeta)`

Orange/classification/rules.py
```diff
@@ -0,0 +1,947 @@
...
+    """
+    A rule induction classifier. Instances are classified following
+    either an unordered set of rules or a decision list.
+    """
+    __metaclass__ = ABCMeta
+
+    def __init__(self, domain=None, rule_list=None):
+        super().__init__(domain)
+        self.domain = domain
+        self.rule_list = rule_list if rule_list is not None else []
```
kernc:
```py
self.rule_list = rule_list or []
```

BlazZupan:
Trying to run on brown-selected.tab (only 186 data instances, 79 attributes). The induction is very slow on this data set, perhaps a problem with missing values?

matevzkren:
Missing values are always handled by the preprocessor. Induction is slow due to a lot of unique values in numerous continuous attributes. I've included a small fix to lead the example that is now included in the main function (please re-download).

`learner.rule_finder.search_strategy.discretise_continuous = True`

The example uses continuous discretisation and speeds up the algorithm significantly. However, I have intentions to further improve the speed slightly in regard to continuous variables.

janezd:
I'm sorry for not paying more attention earlier: I am not sure that discretization is the way to go here. We expected similar conditions in the rules as in the trees, hence one would expect something like *age < 30* and not *age within some interval*. In a sense, using discretization here would mean repeating the skl's mistake backwards: skl treats discrete attributes as numeric, while this implementation of cn2 does the opposite.

In the last two weeks I reimplemented classification and regression trees, but with proper treatment of all kinds of attributes. I had the problem with this same data set, brown-selected, until I implemented finding the thresholds (and some other functions) in Cython. Please see #1443, and in particular https://github.com/janezd/orange3/blob/709c23aa59d10ddd121c79ddc51e65685f457a6e/Orange/classification/_tree_scorers.pyx. The first function in this file is really similar to what you need: given an attribute and class column, and indices obtained from argsorting the attribute, it returns the optimal threshold w.r.t entropy. You may need to modify it to observe just one half of the data (below and/or above the threshold) and to compute some other statistics.

matevzkren:
The misunderstanding stems from an inadequate function name on my part, which somehow endured from the early stages of development. Continuous variables are not discretized into bins, rather the continuous value space is 'discretized'. Relevant cut-off values are determined using entropy & MDL.

That said, I will definitely re-use your code (with modifications to support weights & target class).  Thanks^

Orange/widgets/classify/owrules.py
```diff
@@ -0,0 +1,263 @@
...
+            # add the default rule
+            rule_list.append(self.generate_default_rule(X, Y, W, self.domain))
+
+        return CustomRuleClassifier(domain=self.domain, rule_list=rule_list,
+                                    rule_ordering=self.rule_ordering,
+                                    covering_algorithm=self.covering_algorithm)
+
+
+class OWRuleLearner(OWBaseLearner):
+    name = "Rule Induction"
```
kernc: Might be called CN2? Trees also induce rules.

Orange/widgets/classify/owrules.py
```diff
@@ -0,0 +1,263 @@
...
+    description = "Induce rules from data."
```
kernc: > Induce decision rules from data using CN2 algorithm.

Orange/widgets/visualize/owruleviewer.py
```diff
@@ -0,0 +1,251 @@
...
+                         QMainWindow, QMouseEvent, QGraphicsView,
+                         QPen, QBrush, QColor, QAbstractItemView,
+                         QItemDelegate, QStyledItemDelegate, QPainter,
+                         QFontMetrics, QStyle)
+
+
+class Rule:  # temp - DELETE
+    curr_class_dist = [0, 28, 0]
+
+    def __str__(self):
```
kernc: You can put arbitrary objects into PyTableModel. Yet, if you want their `__str__()` values to be used for Qt.DisplayRole, you need to extend the model like so:
```py
class MyTableModel(PyTableModel):
    def data(index, role=Qt.DisplayRole):
        value = super().data(index, role)
        return str(value) if role == Qt.DisplayRole else value
```
Otherwise, the view will just show blanks as it doesn't know how to convert arbitrary objects into strings.

Orange/widgets/visualize/owruleviewer.py
```diff
@@ -0,0 +1,251 @@
...
+        #  widget.Message()
+    ]
+
+    headers = ["IF cond", "", "THEN class", "quality", "class dist"]
+
+    def __init__(self):
+        raise NotImplementedError
+        self.rule_list = None
+
+        self.tableModel = PyTableModel(sequence=[], parent=self, editable=True)
```
kernc: `editable=True` makes the view respond to double-click so the user can write their own values.

Orange/widgets/visualize/owruleviewer.py
```diff
@@ -0,0 +1,251 @@
...
+        self.tableModel.setHorizontalHeaderLabels(self.headers)
+        self.tableModel.dataChanged.connect(self.on_changed)
+
+        self.tableView = gui.TableView
```
kernc: is a class

Orange/widgets/visualize/owruleviewer.py
```diff
@@ -0,0 +1,251 @@
...
+        def __lt__(self, other):
+            return self.data() < other.data()
+
+    class ProxyModel(QSortFilterProxyModel):
+        def filterAcceptsRow(self, row, parent):
+            pass
+
+
+class DistributionItemDelegate(QStyledItemDelegate):
+    # taken from http://bit.ly/2b0NBqU
```
kernc: Replace the bitly link with its target.

kernc: Long URLs go on a separate line.

doc/data-mining-library/source/tutorial/classification.rst
```diff
@@ -135,4 +135,64 @@ Logistic regression wins in area under ROC curve::
...
+chosen rule. The process is repeated while learning instances remain.
+To evaluate found hypotheses and to choose the best rule in each
+iteration, search heuristics are used. Primarily, rule class
+distribution is the decisive determinant.
+
+Classic CN2
++++++++++++
+
+Classic CN2 inducer (:any:`CN2Learner`) that constructs a list of
+ordered rules. To evaluate found hypotheses, entropy measure is used.
```
kernc: What does it mean that the rules are ordered vs. unordered?

doc/data-mining-library/source/tutorial/classification.rst
```diff
@@ -135,4 +135,64 @@ Logistic regression wins in area under ROC curve::
...
+Returns a :any:`CN2Classifier` if called with data.
+
+The code below loads the *titanic* data set (three discrete attributes
+and discrete class) and fits the learner. The classifier is
```
kernc: a discrete class.

Orange/widgets/gui.py
```diff
@@ -62,6 +62,10 @@ def __init__(self, parent=None, **kwargs):
         v.setVisible(False)
         v.setResizeMode(QHeaderView.ResizeToContents)
 
+    def setModel(self, model):
+        super().setModel(model)
+        model.view = self
```
kernc: This solution is incorrect. A single data model can be attached to multiple views.

Did you have a look at QSortFilterProxyModel?

janezd:
Define the `sizeHint` method to set a suitable default widget size.

janezd:
Having the data input is a good idea, but ...

For classification trees, I can have schema File -> Classification tree -> Classification tree viewer -> Table. When I select a node in the viewer, I see the corresponding subset from the training data. If I try to do the same with CN2 rules, I have to connect File to the Viewer.

Have the data stored with the rules (either always, or add it just in the widget), so the viewer can get it from there.

astaric:
Could this PR be merged in the present state and additional features added in subsequent PRs? Are there any obvious parts missing?

janezd:
That was my plan, too -- merge and perhaps add features later. If you'd do it, yes, please give it a quick check and merge.

matevzkren:
The existing code is fully functional and ready to be merged as is. Some features are being worked on (including additional algorithms) and should be ready by the end of next week. Those can however be handled in subsequent PRs.

markotoplak:
Merging this increased "python setup.py test" on my computer from 17 to 33 seconds. @matevzkren , could you look into this?

PR #1401, opened 2016-07-01
[FIX] Lin and Log Regression: Prevent double commit

BlazZupan:
Both linear and logistic regression committed the learner twice upon initialization. The first time when setting (changing) the attributes for initial setup, and the second time once the initialization was over and when the auto commit was set to True. If the widgets were connected to Test & Score, this resulted in running the testing (e.g. cross validation) twice.

Debugging of these widget lead to discovery a bug of crashing the parallel execution of cross-validation, but only when in debug mode and running through PyCharm on OS X. The fix is to change to sequential execution  (`n_jobs = 1`) but issuing a warning instead of raising an error.

### DIFF
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -347,7 +347,9 @@ class Results:
             # multiprocessing (otherwise it shits itself at least on Windos)
             mp_queue = mp_ctx.Manager().Queue() if n_jobs > 1 else mp.Queue()
         except (EOFError, RuntimeError):
-            raise RuntimeError('''
+            mp_queue = mp.Queue()
+            n_jobs = 1
+            warnings.warn('''
 
         Can't run multiprocessing code without a __main__ guard.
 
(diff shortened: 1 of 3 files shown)


Orange/evaluation/testing.py
```diff
@@ -365,8 +365,10 @@ def _is_picklable(obj):
         recursion ensues.
 
         Guard your executed code with above Python idiom, or pass n_jobs=1
-        to evaluation methods, i.e. {}(..., n_jobs=1).
-            '''.format(self.__class__.__name__)) from None
+        to evaluation methods, i.e. {}(..., n_jobs=1). Setting n_jobs to 1.
+            '''.format(self.__class__.__name__), OrangeWarning)
+        mp_queue = mp.Queue()
+        n_jobs = 1
```
kernc: These two lines are probably under-indented. They should be under `except`. Perhaps move them directly under except above and before warning.

PR #1402, opened 2016-07-01
[ENH] ImageViewer: Add a 'Preview' like window

ales-erjavec:
Introduce a 'OSX Preview' like window to step through the images.

### DIFF
--- a/Orange/widgets/data/owimageviewer.py
+++ b/Orange/widgets/data/owimageviewer.py
@@ -7,6 +7,8 @@ import sys
 import os
 import weakref
 import logging
+import enum
+import itertools
 from xml.sax.saxutils import escape
 from collections import namedtuple
 from functools import partial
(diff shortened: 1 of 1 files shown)


BlazZupan:
This is a great addition to Image Viewer. Has problems though if the images are very big (say 3000 x 3000). Then the preview window would cover the entire area of the screen. I can make preview window smaller, but as soon as I click on the main Image Viewer (or change the image by using arrow keys) the preview comes into focus in a full size, obscuring everything else.

This could be fixed with:

- preview window having the same size (does not scale automatically to the size of the image)
- preview can be made bigger or smaller, by the user, but stays the same with all the images
- preview window remembers its position and size

When preview opens, it shows at front (correct), but then I am not able to click on other windows, say on Image Viewer. It would be better if Preview is treated just as a regular window that can also fall back when other windows are brought into focus.

Check also changing the images. If i do (1) bring the preview window, (2) click on the Image View window, the selected image changes with arrow keys, but the preview image stays the same.

astaric:
I kind of like it how it work on osx.

Preview opens large automatically adapts it size to the image it displays until I resize it manually. Once manually resized, it will keep the size (and position) until closed.

ales-erjavec:
The image window now remembers it it was resized by the user and keeps that size when switching between images. Hiding/showing the window will reset this state, reverting back to automatic resize.

PR #1405, opened 2016-07-03
gui: Add listView

janezd:
Working with models is cleaner and usually more convenient than with lists of strings. This PR adds `gui.listView`, which is similar to `gui.listBox` but uses a model instead of list, so values can be instances of `Variable` instead of their names. The code is tested on the box plot widget.

This currently does not work with context settings. We need to modify them to store the `name` if the value as an instance of `Variable` and to restore the `Variable`. For the latter, we need a flag either in the definition of the setting or in the saved settings. @astaric, any recommendations?

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -1217,7 +1217,6 @@ def listView(widget, master, value=None, model=None, box=None, callback=None,
         bg = widget
     view = ListViewWithSizeHint(preferred_size=sizeHint)
     view.setModel(model)
-    view.selectionModel().selectionChanged.connect(callback)
     if value is not None:
         connectControl(master, value, callback,
                        view.selectionModel().selectionChanged,
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -12,9 +12,9 @@ from types import LambdaType
 import pkg_resources
 
 from PyQt4 import QtGui, QtCore
-from PyQt4.QtCore import Qt, pyqtSignal as Signal
+from PyQt4.QtCore import Qt, pyqtSignal as Signal, QSize
 from PyQt4.QtGui import QCursor, QApplication, QTableView, QHeaderView, \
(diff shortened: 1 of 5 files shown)


Orange/widgets/gui.py
```diff
@@ -1195,6 +1195,36 @@ def attributeItem(var):
     return attributeIconDict[var], var.name
 
 
+class ListViewWithSizeHint(QListView):
+    def __init__(self, *args, preferred_size=None, **kwargs):
+        super().__init__(*args, **kwargs)
```
kernc: Given how one can pass a preferred size as a kwarg into QWidget constructor, e.g.
```py
QListView(parent, sizeHint=QSize(100, 80), **other_properties)
```
is this class necessary?

Yes, QSize needs to be imported.

janezd: I suppose I didn't know about the `sizeHint` property. I'll remove the class. Thanks.

No, `QSize` doesn't need to be imported. The class is (was) intended to be used in `gui.listView`, which can accept a tuple and convert it.

kernc: I meant in above case when setting sizeHint from kwarg. This makes it inconvenient compared to tuples, but is a rather small price to pay if it gets you rid of a whole class. :smiley:  

> "It is better to have 100 functions operate on one data structure than 10 functions on 10 data structures."
> — Alan Perlis

ales-erjavec:
> Given how one can pass a preferred size as a kwarg into QWidget constructor, e.g.
> ```py
> QListView(parent, sizeHint=QSize(100, 80), **other_properties)
> ```
> is this class necessary?

That does not work as `sizeHint` is not a settable property. It might warrant a bug report to PyQt, though.

kernc:
Ah, sorry. Thanks.

astaric:
Regarding the ContextSettings, they should work. encode and decode convert between variables and serializable representation (name, type).

If they still do not work after rebasing this onto the latest master, I can take a look given the reproducible example.

PR #1406, opened 2016-07-03
VizRank-like function for Box Plot

janezd:
<del>Based on #1405 (not master).</del>

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -159,20 +159,31 @@ class OWBoxPlot(widget.OWWidget):
         self.attrs = VariableListModel()
         view = gui.listView(
             self.controlArea, self, "attribute", box="Variable",
-            model=self.attrs, callback=self.attr_changed, sizeHint=(200, 150))
+            model=self.attrs, callback=self.attr_changed)
+        view.setMinimumSize(QSize(30, 30))
+        # Any other policy than Ignored will let the QListBox's scrollbar
+        # set the minimal height (see the penultimate paragraph of
+        # http://doc.qt.io/qt-4.8/qabstractscrollarea.html#addScrollBarWidget)
+        view.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Ignored)
+        gui.separator(view.box, 6, 6)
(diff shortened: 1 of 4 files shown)


ajdapretnar:
@janezd Is this still WIP?

janezd:
Blocked by #1405.

ajdapretnar:
Ah, yes, of course. :panda_face:

Orange/widgets/visualize/owboxplot.py
```diff
@@ -224,14 +232,56 @@ def set_data(self, dataset):
...
+        def compute_score(attr):
+            if attr is group_var:
+                return 3
+            if attr.is_continuous:
+                # One-way ANOVA
+                col = data.get_column_view(attr)[0]
+                groups = (col[group_col == i] for i in range(n_groups))
+                groups = (col[~np.isnan(col)] for col in groups)
+                groups = [group for group in groups if len(group)]
+                return f_oneway(*groups)[1] if groups else 2
```
kernc: To wet the appetite, in pandas, this could be:
```py
if attr.is_continuous:
    groups = data.groupby(group_var)[attr].apply(list)
    f_oneway(*groups)[1] if len(groups) else 2

janezd: Interestingly, it's *whet*, not *wet*.

VesnaT:
When attribute's score appears to be nan (i.e not all values of discrete variable appear in a subset, which is then sent to the Boxplot or there is only one distinct value for a continuous variable in the subset), variables doesn't seem to get sorted correctly. I would put those (with nan score) at the end of the list.

janezd:
I added commit 9be832d945c77f320eb23e3fa96af7306883916a. Does it fix the problem? For discrete variables, it can still compute the p-value if there are at least two distinct values. Otherwise, it puts the attribute at the end, as you suggested. I think I also fixed the problem with constant numeric values.

If it still doesn't work: can you send me the data (or instructions) for reproducing this?

VesnaT:
Still not working for continuous variables. What about returning 2 when the score is nan?

janezd:
Thanks. Please try now.

PR #1407, opened 2016-07-03
[RFC] Constants for status messages

janezd:
When implementing a widget which issues multiple error, warning or information messages, we have to keep track of the used id's or pick random numbers and hope for the best to avoid collisions.

This PR adds id generator `status_id` to the `OWWidget` class, and shows its use on several widgets. Most widgets have only a single type of error, hence they won't need to use these constants (see owsql and owdistances, below). For those with multiple statuses, constants will clean the mess (see owtestlearners and owheatmap).

Current `int` constants are allowed but deprecated.


janezd:
Closed in favour of #1414

PR #1412, opened 2016-07-04
[FIX] overlay: Auto disconnect when the overlay widget is deleted

ales-erjavec:
Register `__on_destroyed` as a Qt slot so it can be automatically
disconnected when the overlay widget is deleted/destroyed.
This fixes a possible 'RuntimeError C/C++ object was deleted ...' error
when the overlay is deleted before the target widget.


astaric:
Travis is misbehaving again and did not automatically run the tests. Could you rebase and push again please?

PR #1334, opened 2016-06-12
[WIP]Widget tests for svm, knn and adaboost classification

rekonder:
- Creating widget test for svm, knn and adaboost classification
- svm classifier widget on create show right kernel parameters
- Added missing parameter at creating learner in adaboost widget

PR #1417, opened 2016-07-05
[ENH] Test widgets on travis

astaric:
This PR contains fixes for (random) segfaults related to gui tests on travis, as well as (slightly modified) tests from #1262 and #1334.

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -1,14 +1,148 @@
 import unittest
 
 from PyQt4.QtGui import QApplication
+import sip
+
+from Orange.canvas.report.owreport import OWReport
 
 app = None
 
 
+class DummySignalManager:
+    def __init__(self):
+        self.outputs = {}
+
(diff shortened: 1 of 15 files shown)


Orange/widgets/tests/base.py
```diff
@@ -1,14 +1,82 @@
...
+
+        NOTE: sip.delete is mandatory. If removed, widgets (sometimes) get
+        deleted by python while some references in QApplication remain
+        (QApplication::topLevelWidgets()), causing a SEGFAULT when
+        QApplication is destroyed.
+        """
+        app.processEvents()
+        for w in cls.widgets:
+            w.onDeleteWidget()
+            sip.delete(w)
```
kernc: Does it work if you just put a destructor on `OWWidget`:
```py
def __del__(self):
    if not sip.isdeleted(self):
        sip.delete(self)
```

astaric: Sadly, no.

PR #1420, opened 2016-07-07
Tree: Common mixin for regression and classification tree

astaric:


kernc:
So ... umm ... this is something only used in tests?

Issue #1303, opened 2016-06-03
Can't "pinch to zoom" in main view


sirjamesgray:
##### Expected bahavior

in the main view (is it called the "classification tree"?) I was hoping I could zoom in / out by using "pinch to zoom" gesture on OS X
##### Actual behavior

"pinch" gesture does not affect the viewer
##### Steps to reproduce the behavior

try to "pinch to zoom" in the main window

kernc:
Does "pinch to zoom" internally translate into scroll wheel events?

ajdapretnar:
@astaric Is this a wontfix or do we plan to implement this on Tree Viewer? I have reservations about this, since Scatter Plot works with, as @kernc mentioned, scroll wheel events. I reckon 'pinch-to-zoom' is touchpad-only.

ajdapretnar:
@astaric?

astaric:
I guess it is a wont fix.

kernc:
Pinch-to-zoom is a feature on all multi-touch trackpads with supporting software with gestures enabled; also touchscreens.

Why is this a wontfix? Do we expressly _not want it_? We have an open-source community; meaning no one in particular has to do it for the general wishlist item to remain open and awaiting anyone willing to tackle it.

PR #1423, opened 2016-07-07
[RFC] TreeViewer: Pinch to zoom

astaric:
An attempt at solving #1303.

I have ported the Qt4 Pinch Zooming example (http://doc.qt.io/qt-4.8/qt-touch-pinchzoom-example.html) to work with our tree widget and modified the code a bit (if the distance between fingers changes for more than 20%, its a pinch to zoom, otherwise, it is a pan event).

Only tested in on OSX and it kind of works. Could use some testing on other platforms and general suggestions if there is another, cleaner way to do this.

### DIFF
--- a/Orange/widgets/classify/owtreeviewer2d.py
+++ b/Orange/widgets/classify/owtreeviewer2d.py
@@ -232,10 +232,32 @@ class TreeGraphicsView(QGraphicsView):
     def __init__(self, scene, *args):
         super().__init__(scene, *args)
         self.viewport().setMouseTracking(True)
+        self.viewport().setAttribute(Qt.WA_AcceptTouchEvents)
         self.setFocusPolicy(Qt.WheelFocus)
         self.setRenderHint(QPainter.Antialiasing)
         self.setRenderHint(QPainter.TextAntialiasing)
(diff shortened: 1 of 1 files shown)


kernc:
Would you make a `gui.GraphicsView(QGraphicsView)` all other widgets' `QGraphicsViews` would extend from?

astaric:
Could do. Are there any other features we could pull into this new class?

kernc:
Auto setting [properties](https://doc.qt.io/qt-5/qgraphicsview.html#properties) to most sane and cross-platform-compatible (overridable) values. Kind of like [what is done for TableView](https://github.com/biolab/orange3/blob/20a860c684f510203d8f2dcd915e359f25b804ae/Orange/widgets/gui.py#L41).

kernc:
Feel like working on this further? :smiley:

astaric:
Nah. The current version is not good enough for practical use and I really hate that I have to manually decide whether a gesture is a pan or a zoom one (I fitted the parameters to my trackpad, but they might be completely off for different hardware). I would be really nice, if this functionality would be provided by OS/Qt

kernc:
With [`QGesture`](https://doc.qt.io/qt-4.8/gestures-overview.html), for example.

PR #1427, opened 2016-07-08
[WIP] Changes to splash screen

ajdapretnar:
What at first seemed to be a trivial task is actually a vast pool of refactoring. I've made some of it already, changed instances of documentation() to tutorials() and tutorial() to examples(), but this is far from complete. Before merging please review the required refactoring in order to avoid inconsistencies and confusion. Also - Tutorials in Help now goes to YT channel. Is that what we want? Tnx.


BlazZupan:
> Tutorials in Help now goes to YT channel. Is that what we want

yes. I also agree with all other changes. The icon for Examples is missing, should have it during the weekend, will send it to you. Then it's ready for a commit.

BlazZupan:
This all works fine. So despite WIP, which should actually be ENH, I will merge this PR.

PR #1431, opened 2016-07-08
[FIX] DataTable: Sends None when no data is selected

BlazZupan:
Fixed a bug where Data Table send an empty data table (table with no instances) when no data was selected. No sends None.


lanzagar:
Is this still needed after #1433?

Also, looks like this duplicates logic of lines [844](https://github.com/biolab/orange3/pull/1431/files?diff=split#diff-39be72554856e6c399803af6ad541157R844) and below.

astaric:
Since this is a workaround for the problem that was fixed in #1433, I am closing this PR without merging.

PR #1432, opened 2016-07-08
[FIX] statistics.utils: Fix stats for sparse when last column missing

nikicc:
`np.bincount` has a length of the maximal element and can hence be shorter than the number of columns (e.g. when some of the last columns have all zeros). This forces it to count non zero elements for all columns.


astaric:
Could you add a test for this?

nikicc:
Test was added 😄

PR #1435, opened 2016-07-08
[FIX] MDS and Distances widges fix

ales-erjavec:
* Add more input data checks in MDS widget
* Preserve selected metric in 'Distances' widget


BlazZupan:
This nicely solves the problem of remembering the same distance metrics for the data in the same domain. I have tried this on dense data. Could somebody, say @nikicc, test on sparse data and then close.

nikicc:
@BlazZupan I tested it on sparse and did not find any problems so I'm merging.

PR #1437, opened 2016-07-10
[FIX] Welcome screen (tutorial) schemas renamed to (example) workflows

BlazZupan:
Cosmetic changes in wording. Examples from Welcome Screen now open a window called Example Workflows. Before, this window was (wrongly) referred to as Tutorials. Now Tutorials refer to YouTube videos. To avoid further confusion I have also renamed the directory with workflows (before, this directory was called tutorials).


Orange/canvas/application/workflows/__init__.py
```diff
@@ -32,55 +30,57 @@ def default_entry_point():
...
 
-def tutorial_entry_points():
-    """Return an iterator over all tutorials.
+def workflow_entry_points():
+    """Return an iterator over all example workflows.
     """
     default = default_entry_point()
     return chain([default],
-                 pkg_resources.iter_entry_points("orange.widgets.tutorials"))
+                 pkg_resources.iter_entry_points("orange.widgets.workflows"))
```
lanzagar: This breaks all add-ons that register their own tutorials/workflows using this entry point.

lanzagar: I suggest we use both entry points for an intermediate time, while add-ons can be updated and start using the new entry point.
Later, when everything is up to date, we can remove the 'tutorials' entry point.

I.e. for now we use:
```
    return chain([default],
                 pkg_resources.iter_entry_points("orange.widgets.tutorials"),
                 pkg_resources.iter_entry_points("orange.widgets.workflows"))
```

PR #1440, opened 2016-07-11
[FIX] OWScatterPlot: Fix crash on classless data

VesnaT:


lanzagar:
I think at first the information message was meant for the VizRank widget. After some refactoring it is now a more lightweight QDialog and can't have warnings anymore.

But using the same message on the main ScatterPlot widget is not OK.
`"Data with a class variable is required."` is wrong and confusing there.

An acceptable half-solution would be to use a better message, e.g. "Automatic scoring requires data with a class variable".
Even better in my opinion is to not use the information mechanism at all, and just make the message a tooltip on the disabled button.

PR #1441, opened 2016-07-11
[ENH] Pythagorean tree and Pythagorean forest widgets

pavlin-policar:
Pythagorean tree widget for visualising trees. Pythagorean forest widget for visualising random forests.

Both work with the `Tree` and `RandomForest` interfaces introduced recently in 2230b7c.

Commit history available in prototypes.

The util files I have for graphic views and components should probably be moved to a better named folder (widgetutils isn't the best, but utils was taken and my utils and the existing ones don't really fit into the same context).


Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -0,0 +1,610 @@
...
+    def _update_legend_visibility(self):
+        if self.legend is not None:
+            self.legend.setVisible(self.show_legend)
+
+    def _update_log_scale_slider(self):
+        """On calc method combo box changed."""
+        if self.SIZE_CALCULATION[self.size_calc_idx][0] == 'Logarithmic':
+            self.log_scale_box.parent().setEnabled(True)
+        else:
+            self.log_scale_box.parent().setEnabled(False)
```
kernc:
```py
self.log_scale_box.parent().setEnabled(
    self.SIZE_CALCULATION[self.size_calc_idx][0] == 'Logarithmic')
```

Orange/widgets/visualize/widgetutils/tree/skltreeadapter.py
```diff
@@ -0,0 +1,302 @@
...
+        self._adjust_weight = adjust_weight
+
+        # clear memoized functions
+        self.weight.cache_clear()
+        self._adjusted_child_weight.cache_clear()
+        self.parent.cache_clear()
+
+        self._all_leaves = None
+
+    @lru_cache(maxsize=1024)
```
kernc: `lru_cache`-ing of methods may prevent disposing of self. https://stackoverflow.com/questions/33672412/python-functools-lru-cache-with-class-methods-release-object

Orange/widgets/visualize/widgetutils/tree/rules.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def merge_with(self, rule):
+        if not isinstance(rule, ContinuousRule):
+            raise NotImplemented('Continuous rules can currently only be '
+                                 'merged with other continuous rules')
+        # Handle when both have same sign
+        if self.sign == rule.sign:
+            # When both are GT
+            if self.sign is True:
+                larger = self.value if self.value > rule.value else rule.value
```
kernc:
```py
max(self.value, rule.value)
```

Orange/widgets/visualize/widgetutils/tree/rules.py
```diff
@@ -0,0 +1,181 @@
...
+    Notes
+    -----
+      - Continuous rules can currently only be merged with other continuous
+        rules.
+
+    """
+
+    def __init__(self, attr_name, gt, value, inclusive=False):
+        self.attr_name = attr_name
+        self.sign = gt
```
kernc: Is `sign` an ok name for a `True`/`False` value? I though it was about the number sign, e.g. signed zero (-0). Would `greater_than`, `greater`, or even `gt` work?

Orange/widgets/visualize/widgetutils/tree/rules.py
```diff
@@ -0,0 +1,181 @@
...
+                larger = self.value if self.value > rule.value else rule.value
+                return ContinuousRule(self.attr_name, self.sign, larger)
+            # When both are LT
+            else:
+                smaller = self.value if self.value < rule.value else rule.value
+                return ContinuousRule(self.attr_name, self.sign, smaller)
+        # When they have different signs we need to return an interval rule
+        else:
+            lt_rule = self if self.sign is False else rule
+            gt_rule = self if lt_rule != self else rule
```
kernc: le and gt or lt and ge cover the interval.
```py
le_rule, gt_rule = (rule, self) if self.sign else (self, rule)
```

Orange/widgets/visualize/widgetutils/tree/rules.py
```diff
@@ -0,0 +1,181 @@
...
+    def __init__(self, attr_name, eq, value):
+        self.attr_name = attr_name
+        self.sign = eq
+        self.value = value
+
+    def merge_with(self, rule):
+        # It does not make sense to merge discrete rules, since they can only
+        # be eq or not eq.
+        from sys import stderr
+        print('WARNING: Merged two discrete rules `%s` and `%s`'
```
kernc: There's a print here. Perhaps you meant `warnings.warn(...)`?

Orange/widgets/visualize/widgetutils/tree/rules.py
```diff
@@ -0,0 +1,181 @@
...
+    eq : bool
+        Should indicate whether or not the rule equals the value or not.
+    value : object
+
+    Examples
+    --------
+    Age = 30
+    >>> rule = DiscreteRule('age', True, 30)
+    Name ≠ John
+    >>> rule = DiscreteRule('name', False, 'John')
```
kernc: This doctest example is in reverse?

kernc: Note to self: doctest tests should be failing on this docstring. Why aren't they?

Orange/widgets/visualize/widgetutils/tree/rules.py
```diff
@@ -0,0 +1,181 @@
...
+        Parameters
+        ----------
+        rule : Rule
+
+        Returns
+        -------
+        Rule
+
+        """
+        raise NotImplemented()
```
kernc:
```py
raise NotImplementedError
```

kernc:
Please enable tests in _Orange/widgets/visualize/widgetutils/tree/tests_.

kernc:
Please do what you find reasonable to [make pylint more happy](https://travis-ci.org/biolab/orange3/jobs/143900630#L316).

Orange/widgets/visualize/widgetutils/common/view.py
```diff
@@ -0,0 +1,180 @@
...
+
+        super().keyPressEvent(ev)
+
+    def __set_padding(self, padding):
+        # Allow for multiple formats of padding for convenience
+        if isinstance(padding, int):
+            padding = list(repeat(padding, 4))
+        elif isinstance(padding, list) or isinstance(padding, tuple):
+            if len(padding) == 2:
+                padding = (*padding, *padding)
```
kernc: This is (unfortunately) not Python 3.4 syntax. Accommodating people using LTS software (including myself), we should keep this Python 3.4-compatible.

Orange/widgets/visualize/widgetutils/tree/rules.py
pavlin-policar: The reason I left it as `sign` is in order to keep some consistency between the subclasses of `Rule`.

Seeing as I did never actually use this general property and that it will most likely cause nothing but confusion, I will change the `sign` in the `ContinuousRule` to `gt` and `sign` in `DiscreteRule` to `eq`.

Orange/widgets/visualize/widgetutils/tree/rules.py
pavlin-policar: The docstring example is not in reverse. The second argument of `DiscreteRule` indicates the `EQ` relation, therefore `Age is equal(True) to 30` and `Name is not equal (False) to John`.

Maybe you meant the line order, so the result should come after the example code?

pavlin-policar:
How would I go about enabling tests? Does this mean I move them to the *Orange/tests* directory, or do I have to somehow indicate that the tests should be included?

Orange/widgets/visualize/widgetutils/tree/rules.py
kernc: The order, yes. For such examples, it is customary to simulate the python interpreter session:
```py
>>> something I type in
something the interpreter spews out
```
For example:
```py
>>> rule = DiscreteRule('name', False, 'John')
>>> print(rule)
Age = 30
```
(Could also be without `print()` if `rule` prints that on `repr()` call.)

kernc:
> How would I go about enabling tests?

Nevermind. Your tests _are_ running on Travis. :blush:

Orange/widgets/visualize/widgetutils/tree/rules.py
```diff
@@ -88,11 +90,11 @@ class ContinuousRule(Rule):
...
     --------
-    x ≤ 30
-    >>> rule = ContinuousRule('age', False, 30, inclusive=True)
+    >>> ContinuousRule('age', False, 30, inclusive=True)
+    age ≤ 30.000
 
-    x > 30
-    >>> rule = ContinuousRule('age', True, 30)
+    >>> ContinuousRule('age', True, 30)
+    age > 30.000
```
kernc: Sorry to have mislead you, this is not ok. `__repr__()` [should return a Python expression from which the object can be recreated](https://docs.python.org/3/reference/datamodel.html#object.__repr__).  So best use separate repr and str implementations.

pavlin-policar: Please see if the `__repr__` implementation is now all right. I've also changed the docstrings to use `print()` since I find this conveys the meaning of the statement

> Rule ∈ [1.000, 3.000)

much clearer than this

> IntervalRule(attr_name='Rule', left_rule=ContinuousRule(attr_name='Rule', greater=True, value=1, inclusive=True), right_rule=ContinuousRule(attr_name='Rule', greater=False, value=3, inclusive=False))

If this is not ok, I can fix it.

kernc: Looks fine to me. :+1:

Orange/widgets/visualize/utils/tree/rules.py
```diff
@@ -0,0 +1,211 @@
...
+                                      'merged with other continuous rules')
+        # Handle when both have same sign
+        if self.greater == rule.greater:
+            # When both are GT
+            if self.greater is True:
+                larger = max(self.value, rule.value)
+                return ContinuousRule(self.attr_name, self.greater, larger)
+            # When both are LT
+            else:
+                smaller = self.value if self.value < rule.value else rule.value
```
kernc:
```py
min(self.value, rule.value) ?
```

BlazZupan:
This is almost ready to merge. I have tested it on Mac, and everything works fine. Some minor suggestions before merging:

- when showing regression trees, "Target class" label in GUI should change to "Node color" (make sure this is changed back to Target class when the widget gets classification tree on the input). This goes for both Tree and Forest widgets
- rename widget name display in the canvas from Pythagorean forest to Pythagorean Forest (widget names use camel case)

Orange/widgets/visualize/pythagorastreeviewer.py
```diff
@@ -0,0 +1,759 @@
...
+    pen : QPen, optional
+        The pen to be used for the border.
+
+    """
+
+    timer = QtCore.QTimer()
+
+    MAX_OPACITY = 1.
+    SELECTION_OPACITY = .5
+    HOVER_OPACITY = .1
```
kernc: I think this value is too small. When one branch is hovered, the other branches are almost invisible here.

pavlin-policar: I made it that way because for larger trees, where there is a lot of overlap (e.g. *adult* with sqrt sizing) the almost invisible-ness of other branches makes the branch you are hovering over very clear. If I change the hover opacity to anything larger (e.g. 0.2), it becomes less clear which branch is being highlighted.

If you feel this is unnecessary and having a higher opacity would be better I can change this.


[0.1 opacity](https://i.gyazo.com/5b5dcdd53df161b840e1e4b92e3ffcdc.png) vs [0.2 opacity](https://i.gyazo.com/2f1bf2a4015435664323eea0f556c6a3.png)

kernc: I see (thanks). In that case, it's ok.

PR #1443, opened 2016-07-12
Add own implementation of tree learner

janezd:
We need our own implementation of tree learner. The emphasis is not on speed (we have simple tree learner for this) and on accuracy (where SKL may (or may not) be better) but on interpretability.

Scikit's tree does not properly handle nominal attributes, and simple tree learner lacks features that would slow it down. This PR (WIP!) has an implementation that is rather slow, in particular for numeric attributes, but will be sped up with a grain of cython.

**Note**: I believe that `Learner.fit` and `Learner.fit_storage` were not implemented as intended. Why would we force all learners to implement `fit`? Wasn't `fit` meant as a way to implement a learner without going through Orange's `Table`? But if somebody prefers `Table`, (s)he can use it. In particular, the new tree learner needs the domain, hence it cannot implement `fit` - except by wrapping the three matrices back into `Table` by using `self.domain`.

Besides, condition `if type(self).fit is Learner.fit` was obviously always `False`.


astaric:
> Besides, condition if type(self).fit is Learner.fit was obviously always False.

```
from Orange.classification import Learner

class A(Learner):
    def test(self): return type(self).fit is Learner.fit    

class B(A):
    def fit(self): pass     

print(A().test())
print(B().test())
```
Outputs
```
True
False
````

**Comment by janezd** [comment]
Sorry, I read `type(self.fit)` instead of `type(self).fit` :)

File: Orange/base.py [hunk 1]
```diff
@@ -27,11 +27,15 @@ def __init__(self, preprocessors=None):
         self.preprocessors = list(preprocessors)
 
     def fit(self, X, Y, W=None):
-        raise NotImplementedError(
-            "Descendants of Learner must overload method fit")
+        raise SystemError(
```
astaric: Based on documentation of [SystemError](https://docs.python.org/3.5/library/exceptions.html#SystemError) and [NotImplementedError](https://docs.python.org/3.5/library/exceptions.html#NotImplementedError), the latter seems more suitable for this particular error.

Orange/classification/tree.py
```diff
@@ -1,20 +1,254 @@
+"""Tree inducers: SKL and Orange's own inducer"""
+
+import numpy as np
+import bottleneck as bn
 import sklearn.tree as skl_tree
+from Orange.statistics import distribution, contingency
 
-from Orange.base import Tree
+from Orange.base import Tree, Model, Learner
```
astaric: Did you want to extend Orange.base.Learner, thus supporting both classification and regression within the same class, or Orange.classification.Learner, which can perform only classification?

janezd: This Learner will handle both, classification and regression. I've put it to folder `classification` for lack of a better place.

Orange/base.py
janezd: Neither is appropriate. `NotImplemented` is described as

> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method.

We do not require overriding `fit`; derived class may provide `fit_storage` instead. Lint doesn't know that, so if we use `NotImplementedError`, lint complains about undefined abstract method in derived class. I used `SystemError` in the sense "this shouldn't have happened". But you're right, it wasn't meant for this. I'll go for [RuntimeError](https://docs.python.org/3.5/library/exceptions.html#RuntimeError).

janezd:
Classification trees are now basically reimplemented (they need some tests etc). Our trees handle nominal attributes - with non-binary trees or with binarization, and tolerate missing attribute values. Unless I don't know something, the only thing they lack in comparison with skl is gini-index. It can be added, but I don't know how it differs from entropy. If the user can't make an informed choice, let's not give him any.

Prediction with our trees is usually (a few times) faster.

Training times are incomparable since our trees do more than skl's.

@BlazZupan What do we do with SKL's trees? I propose we keep them in scripts, but the tree widget will induce our trees (alternative: a checkbox saying something non-technical, like "Prefer speed over readability"). I added the code for mixin (@astaric, it's more like an interface :) that allows for showing them in widgets. I would not remove that code, but it will remain unused since there is no widget to generate SKL trees.

janezd:
This is ready for merging (after I squash the commits), unless we decide to change any of the following:

- @astaric defined an interface that allows showing different kinds of trees with the same widget. Classification trees and regression trees are now one and the same tree, so they are now the only class that implements this interface - and the only model appropriate for the tree visualization widget. (I removed the interface for SKL classification trees since there is no widget that produces them and thus nobody would use it and it would be unmaintained; for SKL regression, I haven't even implemented that interface.) We could simplify the code of the tree and the widget by removing the interface.

- Orange still wraps the classification and regression trees from SKL. I don't think they offer anything useful. Remove them?

- The two learners produce the same model and have almost the settings (except for one additional in classification), and run the same tests. We can merge them into the same learner that will handle classification and regression data.

I'd appreciate any comments on this.

Orange/base.py
```diff
@@ -27,11 +27,15 @@ def __init__(self, preprocessors=None):
         self.preprocessors = list(preprocessors)
 
     def fit(self, X, Y, W=None):
-        raise NotImplementedError(
-            "Descendants of Learner must overload method fit")
+        raise RuntimeError(
+            "Descendants of Learner must overload method fit or "
+            "fit_storage")
```
kernc: I don't think this change is reasonable or required. Abstract methods `raise NotImplementedError` by convention. Even the soon-to-be-obsolete message is not required. The developer experiencing this exception should know or know where to look what to do.

Orange/base.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, domain=None):
         self.domain = domain
 
     def predict(self, X):
-        if self.predict_storage == Model.predict_storage:
+        if type(self).predict_storage is Model.predict_storage:
```
kernc: Overly pedantic. :stuck_out_tongue:

Orange/regression/tree.py
```diff
@@ -1,12 +1,142 @@
...
+    NumericNode, OrangeTreeModel
+from Orange.regression import SklLearner, SklModel, Learner
+from Orange.classification import _tree_scorers
 
-__all__ = ["TreeRegressionLearner"]
+__all__ = ["TreeRegressionLearner", "OrangeTreeLearner"]
 
 
-class TreeRegressor(SklModel, Tree):
+class OrangeTreeLearner(Learner):
```
kernc: This class bears the same name as `Orange.classification.OrangeTreeLearner`, yet it is [imported as RegressionTreeLearner](https://github.com/biolab/orange3/pull/1443/files#diff-8f8447f63aca75c586862b01460f16a2R7)?

Orange/base.py
janezd: I didn't change it because of `is` but because of `type` --- the condition couldn't be `True` (I guess) since it compared bound and unbound method.

```
>>> class A:
...     def f():
...         pass
...
>>> a = A()
>>> a.f == A.f
False
>>> type(a).f == A.f
True
```

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
+#cython: embedsignature=True
+
+import numpy
```
kernc:
```py
import numpy as np
```
Somewhat oddly but surely, there is no clash.

Orange/regression/tree.py
janezd: ... yes, but only when you need to import both in the same file. I would prefer, though, to merge both classes into a single class for classification and regression. Same for widget.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
+#cython: embedsignature=True
```
kernc: Also `boundscheck=False, wraparound=False` (ensure they can be applied module-wide). If you can vouch never to divide by zero, also `cdivision=True`.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+
+import numpy
+cimport numpy as np
+import cython
+
+from libc.stdlib cimport calloc, free
+
+cdef extern from "math.h":
+    double log(double x)
```
kernc:
```cython
from libc.math cimport log
```

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+
+cdef extern from "numpy/npy_math.h":
+    bint npy_isnan(double x)
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def find_threshold_entropy(np.ndarray[np.float64_t, ndim=1] x,
+                           np.ndarray[np.float64_t, ndim=1] y,
+                           np.ndarray[np.int64_t, ndim=1] idx,
+                           int n_classes, int min_leaf):
```
kernc:
```py
def find_threshold_entropy(double[:] x,
                           double[:] y,
                           int[:] idx,
                           int n_classes,
                           int min_leaves):
```
I find this syntax somewhat more readable. It is compatible and potentially faster.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+    free(distr)
+    return (class_entro - best_entro) / N / log(2), x[idx[best_idx]]
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def find_binarization_entropy(np.ndarray[np.float64_t, ndim=2] cont,
+                              np.ndarray[np.float64_t, ndim=1] class_distr,
+                              np.ndarray[np.float64_t, ndim=1] val_distr,
+                              int min_leaf):
```
kernc: Please add comprehensive docstrings to these functions.

Orange/classification/_tree_scorers.pyx
kernc: Oh, and also `infer_types=True`.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+    bint npy_isnan(double x)
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def find_threshold_entropy(np.ndarray[np.float64_t, ndim=1] x,
+                           np.ndarray[np.float64_t, ndim=1] y,
+                           np.ndarray[np.int64_t, ndim=1] idx,
+                           int n_classes, int min_leaf):
+    cdef unsigned int *distr = <unsigned int *>calloc(2 * n_classes, sizeof(int))
+    cdef Py_ssize_t i, j
```
kernc: If you use Cython's `infer_types=True`, you don't need lines like this one.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+    cdef double entro, class_entro, best_entro
+    cdef unsigned int p, curr_y
+    cdef unsigned int best_idx = 0
+    cdef unsigned int N = idx.shape[0]
```
kernc: You can wrap all cdefs in a single `cdef` block: http://docs.cython.org/en/latest/src/userguide/language_basics.html#grouping-multiple-c-declarations

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+
+    # Initial split (min_leaf on the left)
+    if N <= min_leaf:
+        return 0, 0
+    for i in range(min_leaf - 1):  # one will be added in the loop
+        # without temporary int, cython uses PyObjects here
+        curr_y = int(y[idx[i]])
+        distr[n_classes + curr_y] += 1
```
kernc: It shouldn't construct PyObjects. Perhaps one of above proposals will fix it. You can also cast with:
```cython
distr[n_classes + <int>curr_y] += 1
```

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+    for i in range(min_leaf - 1, N):
+        distr[int(y[idx[i]])] += 1
```
kernc: Likewise.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+cdef extern from "numpy/npy_math.h":
+    bint npy_isnan(double x)
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def find_threshold_entropy(np.ndarray[np.float64_t, ndim=1] x,
+                           np.ndarray[np.float64_t, ndim=1] y,
+                           np.ndarray[np.int64_t, ndim=1] idx,
+                           int n_classes, int min_leaf):
+    cdef unsigned int *distr = <unsigned int *>calloc(2 * n_classes, sizeof(int))
```
kernc: Don't do your own memory management. Use `np.zeros(2 * n_classes, dtype=int)`. It's only called once. The overhead is negligible.

kernc: So then the declaration changes:
```cython
cdef:
    unsigned int[:] distr = ...
```

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+    best_inter = (sum * sum) / N
+    for i in range(min_leaf - 1, N - min_leaf):
+        sleft += y[idx[i]]
+        if x[idx[i]] == x[idx[i + 1]]:
+            continue
+        inter = sleft * sleft / (i + 1) + (sum - sleft) * (sum - sleft) / (N - i - 1)
+        if inter > best_inter:
+            best_inter = inter
+            best_idx = i
+    # punishment for missing values is delivered outside
```
kernc: What does that mean? Elsewhere? Why? Where?

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+                          int n_values, int min_leaf):
+    cdef double sleft, sum, val
+    cdef unsigned int left
+    cdef unsigned int i, change, to_right, allowed, m
+    cdef unsigned int best_mapping, move, mapping, previous
+    cdef double inter, best_inter, start_inter
+    cdef unsigned int N
+
+    cdef np.ndarray[np.int32_t, ndim=1] group_sizes = \
+        numpy.zeros(n_values, dtype=numpy.int32)
```
kernc: So again:
```cython
cdef:
    np.int32_t[:] group_sizes = np.zeros(n_values, dtype=np.int32)
```
```

File: Orange/classification/_tree_scorers.pyx [hunk 18]
```diff
@@ -0,0 +1,293 @@
...
+    free(distr)
+    return (class_entro - best_entro) / N / log(2), best_mapping
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def find_threshold_MSE(np.ndarray[np.float64_t, ndim=1] x,
+                       np.ndarray[np.float64_t, ndim=1] y,
+                       np.ndarray[np.int64_t, ndim=1] idx,
+                       int min_leaf):
```
kernc: Once you have functions use buffers and no python (numpy) code, you can tag them:
```cython
def func(...) nogil:
    ...
```
or
```cython
with nogil:
    ... code that is run with released GIL  ...
```

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+    # factor n / x.shape[0] is the punishment for missing values
+    #return (inter - sum * sum / n) / n * n / x.shape[0]
+    return (inter - sum * sum / n) / x.shape[0]
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def compute_predictions(np.ndarray[np.float64_t, ndim=2] X,
+                        np.ndarray[np.int32_t, ndim=1] code,
+                        np.ndarray[np.float64_t, ndim=2] class_distrs,
```
kernc: And this becomes
```cython
double[:, :] class_distrs
```

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,293 @@
...
+            val = X[i, code[node_ptr + 2]]
+            if npy_isnan(val):
+                break
+            if code[node_ptr] == 3:
+                node_idx = code[node_ptr + 1]
+                val_idx = int(val > thresholds[node_idx])
+            else:
+                val_idx = int(val)
+            next_node_ptr = code[node_ptr + 3 + val_idx]
+            if next_node_ptr == -1:
```
kernc: What's a `-1`?

Orange/classification/setup.py
```diff
@@ -17,9 +17,12 @@ def configuration(parent_package='', top_path=None):
                              "build_tree", "destroy_tree", "new_node",
                              "predict_classification", "predict_regression"]
                          )
+    config.add_extension('_tree_scorers',
```
kernc: What about @jzbontar's *_simple_tree.c* a few lines above?

Orange/classification/tree.py
```diff
@@ -1,18 +1,191 @@
...
+
+    # Binarization is exhaustive, so we set a limit on the number of values
+    MAX_BINARIZATION = 16
+
+    def __init__(
+            self, *args, binarize=True, max_depth=None,
+            min_samples_leaf=1, min_samples_split=2, sufficient_majority=0.95,
+            **kwargs):
+        super().__init__(*args, **kwargs)
+        self.binarize = binarize
```
kernc: `binarize=True` is what sklearn does? Why on by default?

Orange/classification/tree.py
```diff
@@ -1,18 +1,191 @@
...
+    __returns__ = OrangeTreeModel
+
+    # Binarization is exhaustive, so we set a limit on the number of values
+    MAX_BINARIZATION = 16
+
+    def __init__(
+            self, *args, binarize=True, max_depth=None,
+            min_samples_leaf=1, min_samples_split=2, sufficient_majority=0.95,
+            **kwargs):
+        super().__init__(*args, **kwargs)
```
kernc: Missing docstring.

Orange/classification/tree.py
```diff
@@ -1,18 +1,191 @@
...
+
+class TreeClassifier(SklModel):
+    """Wrapper for SKL's tree classifier with the interface API for
+    visualizations"""
+    def __init__(self, *args, **kwargs):
+        SklModel.__init__(self, *args, **kwargs)
+        self._cached_sample_assignments = None
 
 
 class TreeLearner(SklLearner):
```
kernc: Why not make our tree learner the default one and rename this one to `SklTreeLearner`?

Orange/regression/tree.py
```diff
@@ -1,12 +1,142 @@
...
+    __returns__ = OrangeTreeModel
+
+    # Binarization is exhaustive, so we set a limit on the number of values
+    MAX_BINARIZATION = 16
+
+    def __init__(
+            self, *args,
+            binarize=True, min_samples_leaf=1, min_samples_split=2,
+            max_depth=None, **kwargs):
+        super().__init__(*args, **kwargs)
```
kernc: Missing docstring.

Orange/widgets/visualize/owtreeviewer.py
```diff
@@ -0,0 +1,425 @@
...
+    GraphicsNode, GraphicsEdge, OWTreeViewer2D
+from Orange.data import Table
+
+from Orange.widgets.settings import ContextSetting, ClassValuesContextHandler, \
+    Setting
+from Orange.widgets import gui
+from Orange.widgets.utils.colorpalette import ContinuousPaletteGenerator
+
+
+class PieChart(QGraphicsRectItem):
```
kernc: Why is this not `QGraphicsEllipseItem`?

Orange/widgets/visualize/owtreeviewer.py
```diff
@@ -0,0 +1,425 @@
...
+        for i in range(len(self.dist)):
+            angle = self.dist[i] * 16 * 360. / dist_sum
+            if angle == 0:
+                continue
+            painter.setBrush(QBrush(colors[i]))
+            painter.setPen(QPen(colors[i]))
+            painter.drawPie(-self.r, -self.r, 2 * self.r, 2 * self.r,
+                            int(start_angle), int(angle))
+            start_angle += angle
+        painter.setPen(QPen(Qt.white))
```
kernc: Can this pie outline be made black?

Orange/widgets/classify/owclassificationtree.py
```diff
@@ -1,71 +1,103 @@
...
-        gui.comboBox(self.controlArea, self, "attribute_score",
-                     box='Feature Selection',
-                     items=[name for name, _ in self.scores],
+        box = gui.vBox(self.controlArea, True)
+        # the checkbox is put into vBox for alignemnt with other checkboxes
+        gui.checkBox(gui.vBox(box), self, "binary_trees", "Induce binary tree",
                      callback=self.settings_changed)
+        for label, check, setting, fromv, tov in self.spin_boxes:
+            gui.spin(box, self, setting, fromv, tov, label=label, checked=check,
+                     alignment=Qt.AlignRight, callback=self.settings_changed)
```
kernc: The callback seems to not get applied when the checkbox state changes.

kernc:
Viewer crashes on adult and anneal.

Orange/regression/tree.py
kernc: I don't think anyone would object to this. We've already been discussing to unify learners like that.

janezd:
I guess I'm done. The remaining issues could be handled in future PRs. 

- Merging classification and regression: let this be another PR, which will introduce a new directory "modelling", which will also include other such methods, like knn etc.

- After this is done, we also make this the default and push the SKL algorithm to the back.

- I haven't looked into speeding it up. If we keep SKL the "default" for now, there's no such rush to fix that. The priority at the moment is interactive tree builder (needed by the end of August), which requires this PR.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,403 @@
...
+                node_idx = code[node_ptr + 1]
+                val_idx = int(val > thresholds[node_idx])
+            else:
+                val_idx = int(val)
+            next_node_ptr = code[node_ptr + 3 + val_idx]
+            if next_node_ptr == NULL_BRANCH:
+                break
+            node_ptr = next_node_ptr
+        node_idx = code[node_ptr + 1]
+        predictions[i] = values[node_idx]
```
kernc:
```cython
for j in range(values.shape[1]):
    predictions[i, j] = values[node_idx, j]
```
Will allow it to be `nogil` and is just what `__pyx_memoryview_copy_contents()` does otherwise.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,403 @@
...
+        a matrix of values
+    """
+    cdef:
+        unsigned int node_ptr, i, val_idx
+        signed int next_node_ptr, node_idx
+        np.float64_t val
+        double[: ,:] predictions = numpy.empty(
+            (X.shape[0], values.shape[1]), dtype=numpy.float64)
+        # np.ndarray[np.float64_t, ndim=2] predictions = numpy.empty(
+        #     (X.shape[0], class_distrs.shape[1]), dtype=numpy.float64)
```
kernc: This is a comment.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,403 @@
+#cython: embedsignature=True
+#cython: boundscheck=False
+#cython: wraparound=False
+#cython: infer_types=True
+#cython: cdivision=True
+
+
+import numpy as np
+numpy = np
```
kernc: Why `import numpy as np` to then call `numpy.empty()`? Use `np.empty()` or just `import numpy`?

kernc:
Can these trees be made to account for instance weights?

janezd:
Yes, but must they? All learners in Orange 2 supported weights, and we almost never used them.

kernc:
Perhaps you didn't, but general machine learning workflows may rely on them (I have data with a use case right here). If we don't intend to make weights first class citizens, might as well throw them out completely. If we keep (and extend upon) them, all sklearn models support `instance_weights`, so I suggest we don't necessarily switch the default until these trees support it also.

janezd:
I consider this PR done.

I understand that this will not make it into the next release. But unless there are any reasonable objections (I expect there may be some, but this time I will ignore the unreasonable ones) ... Well, I won't merge my own PR, but I'll ask @BlazZupan to do it the minute after the release.

These trees are waaaaay better than shitkit trees: they properly handle nominal attributes and unknown values. That is, unless you consider *gender=male<0.5* a proper synonym for *female*. Given all I list above, it was totally stupid that this was not merged before the workshops we had in September.

Classification accuracy / MSE is the same (+- random) as that of sh-learn.

Regression and classification trees are united; although there are two widgets, one is derived from the other and they are almost the same. The viewer is the same for both, too.

The test coverage is 99.51%; one of the two missing lines is related to SKL and one is "unreachable".

- For tree.py, pylint zero issues in tree.py, and radon reports no method with complexity C, and a few B's.
- For owclassificationtree, pylint reports a single issue (too many ancestors?!), and radon is perfect.
- owtreeviewer has zero issues and only three B's in radon.
- owtreeviewer2d has 2 issues and only one B.

Overall, the changes reduces the number of issues from 268 to 146 (-122), and cyclomatic complexity from 209 to 198 (-11).

I jumped through all the *reasonable* hoops I was asked for. Note also that the PR is neatly divided into no less than 19 commits. (I can split a few of them further if needed. Just ask, I now know how to do it.) TreeModel, which we proposed before I started this work, is implemented in early commits but later removed. I didn't squash this just in case.

@kernc discovered that induction of a tree on the entire adult takes 5s. While this is true, nowadays trees are not used because they are fast but because they are interpretable. The trees in this PR are interpretable, while sh-learn's are fast. Who wins? (Random forest is another issue; there we use sh-learn. (For now.))

Furthermore, I must stress that *it has been suggested that there will be nothing wrong with pandas slowing down most widgets by a factor of three*, with no obvious great advantages beyond following the pandas-hype. The fact that trees are going to be slower (by four seconds on a large data set resulting in a tree with 8000 nodes) is, on the other hand, inconsequential, while these trees have obvious advantages over the existing sh-kit's *gender=male<0.5* -trees.

If this all sounds like I'm still pissed about something ... this may indeed be a safe *assertion*. I can even write a short test for this.

janezd:
Changed status to WIP; I implemented support for Pythagorean trees and I'll add some tests tomorrow.

doc/development/source/orange-demo/orangedemo/OWLearningCurveA.py
```diff
@@ -280,7 +280,7 @@ def main(argv=sys.argv):
     l2.name = 'Logistic Regression'
     ow.set_learner(l2, 2)
 
-    l4 = Orange.classification.TreeLearner()
+    l4 = Orange.classification.SklTreeLearner()
```
ajdapretnar: I checked this and it has some old imports that don't work (nothing related to this PR). I will update accordingly.

Orange/classification/_tree_scorers.pyx
```diff
@@ -0,0 +1,413 @@
...
+                if entro < best_entro:
+                    best_entro = entro
+                    best_idx = i
+    return (class_entro - best_entro) / N / log(2), x[idx[best_idx]]
+
+
+def find_binarization_entropy(double[:, :] cont, double[:] class_distr,
+                              double[:] val_distr, int min_leaf):
+    """
+    Find the split of discrete values into two groups the optimizes information
```
lanzagar: the -> that

Orange/base.py
```diff
@@ -49,11 +53,7 @@ def __call__(self, data):
 
         self.domain = data.domain
 
-        if type(self).fit is Learner.fit:
-            model = self.fit_storage(data)
-        else:
-            X, Y, W = data.X, data.Y, data.W if data.has_weights() else None
-            model = self.fit(X, Y, W)
+        model = self.fit_storage(data)
```
lanzagar: IIRC, the if was there to allow having both `fit` and `fit_storage` defined in which case the more specific (numpy optimized) `fit` would be used.
Is there a reason this was removed?

PR #1445, opened 2016-07-12
SilhouettePlot: Move to Visualize pane

BlazZupan:
As we are moving visualisation of trees within the visualize pane, and actually place there all the vidgets do visualisation, we are also moving there the Silhouette plot.


lanzagar:
The widget (`owsilhouetteplot.py`) was moved, but not the icon. So now it shows a question mark instead.

astaric:
@ales-erjavec If I understand correctly, this move will break all old schemas that use this widget. Is there a way to tell the widget that it used to be known under a different name/path?

ales-erjavec:
`replaces = ["{oldfullqualifiedpath}", ...]`

An example use is even in the module being moved: https://github.com/biolab/orange3/blob/d0d9b84c279c6bd040085a25342c50de7a62452f/Orange/widgets/unsupervised/owsilhouetteplot.py#L36

lanzagar:
The new commit adds a new icon to widgets/visualize/icons instead of moving the existing one from widgets/unsupervised/icons.

PR #1448, opened 2016-07-12
[FIX] Stop advertising support for weights in LogisticRegression.

sstanovnik:
The liblinear solver does not support weights in SKL, even though the parameter is there. Override the property to prevent errors.


lanzagar:
Since codecov is also complaining, you might want to leave a test for logreg (and the new one with linreg replacing the previous one).

assertFalse(LogisticRegressionLearner().supports_weights, ...)
and maybe also assert that learning with weights would raise the exception (so that the override is justified).

astaric:
+1 for second assert, this way we will get a failing test if (when) liblinear starts accepting weights.

PR #1452, opened 2016-07-13
[ENH] Save painted data to schema

### LINKED ISSUE #1451: [ENH] Paint data should store its state when workflow is saved
##### Orange version:
3.3.6

##### Expected behavior: 
When Orange workflow is saved, it should also save the data painted in the Paint Data widget

##### Actual behavior: 
If I paint the data using Paint Data widget and save the workflow, the painted data is not saved. That is, when I close Orange and open the saved workflow, Paint Data initializes with the empty data set. 

Paint the data in the Paint Data widget, store the workflow, close Orange, open the workflow. The data set in the Paint Data widget is empty.

astaric:
Introduce a concept of schema only setting - a setting that does not update its defaults, but is still saved with schema.

Use schema only setting to store painted points in PaintData widget.

Fixes #1451

### DIFF
--- a/Orange/widgets/data/owpaintdata.py
+++ b/Orange/widgets/data/owpaintdata.py
@@ -897,6 +897,7 @@ class OWPaintData(widget.OWWidget):
         redo.setShortcut(QtGui.QKeySequence.Redo)
 
         self.addActions([undo, redo])
+        self.undo_stack.indexChanged.connect(lambda _: self.invalidate())
 
         gui.separator(tBox)
         indBox = gui.indentedBox(tBox, sep=8)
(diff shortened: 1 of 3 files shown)


BlazZupan:
Works great, thanks! Tiny bug: paint the data, set Send to Send Automatically, connect scatter plot to Paint Data, save the workflow. Open the saved worflow, the data is in Paint Data, but although Send Automatically is on the data has not been sent out from the widget (scatter plot is empty).

BlazZupan:
Perhaps while fixing the bug above, there is another one somehow related: after Undo, this widget does not commit the data when Send Automatically is on.

PR #1453, opened 2016-07-13
[FIX] OWBaseLearner: Learner name is changed on output when user changes it and auto apply selected 

PrimozGodec:
Learner name field didn't have callback when it is changed. Problem was when user changed name and auto apply was on, because name haven't changed on output. It is fixed with this PR.

### DIFF
--- a/Orange/widgets/utils/owlearnerwidget.py
+++ b/Orange/widgets/utils/owlearnerwidget.py
@@ -1,5 +1,5 @@
 import numpy as np
-from PyQt4.QtCore import QTimer
+from PyQt4.QtCore import QTimer, Qt
 
 from Orange.classification.base_classification import LearnerClassification
 from Orange.data import Table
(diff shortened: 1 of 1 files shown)


lanzagar:
Now it sends a new output after every key press. This can make it really slow when connected to e.g. Test&Score.
I think in other places we apply the change on focus-out only (or when pressing enter after editing)?

PrimozGodec:
@lanzagar I changed it that now apply happens only on focus-out. One think that I do not like too much is button that shows that field is updated but not committed. It is part of gui script. 
screenshot from 2016-07-13 13-38-22

lanzagar:
It might be acceptable if the button appeared to the right of the text field.
Definitely not above, as in the screenshot :)

PrimozGodec:
It has been fixed

lanzagar:
Looks good to me now.
You can squash all commits into the first and I'll merge.

PrimozGodec:
Done

PR #1456, opened 2016-07-14
[FIX] Table: Fix ensure_copy for sparse matrices

nikicc:
`ensure_copy` is checking whether an array is a view through the `.base` argument which doesn't exist on `scipy.sparse` matrices. Since `scipy.sparse` don't work with views (e.g. indexing returns a copy of the matrix) the check should only be performed for dense matrices.

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -912,15 +912,20 @@ class Table(MutableSequence, Storage):
 
     def ensure_copy(self):
         """
-        Ensure that the table owns its data; copy arrays when necessary
+        Ensure that the table owns its data; copy arrays when necessary.
         """
-        if self.X.base is not None:
+        def is_view(x):
+            # Sparse matrices don't have views like numpy arrays. Since indexing on
+            # them creates copies in constructor we can skip this check here.
+            return not sp.issparse(x) and x.base is not None
+
+        if is_view(self.X):
             self.X = self.X.copy()
-        if self._Y.base is not None:
+        if is_view(self._Y):
             self._Y = self._Y.copy()
-        if self.metas.base is not None:
+        if is_view(self.metas):
             self.metas = self.metas.copy()
-        if self.W.base is not None:
+        if is_view(self.W):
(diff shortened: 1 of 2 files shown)


Orange/data/table.py
```diff
@@ -912,15 +912,16 @@ def is_copy(self):
 
     def ensure_copy(self):
         """
-        Ensure that the table owns its data; copy arrays when necessary
+        Ensure that the table owns its data; copy arrays when necessary.
+        The check is skipped for sparse matrices since they don't have views like numpy arrays.
         """
-        if self.X.base is not None:
+        if not sp.issparse(self.X) and self.X.base is not None:
             self.X = self.X.copy()
```
kernc: But if they are sparse, no copy is made?

nikicc: Currently not, this is left to the constructor. As far as I understand `ensure_copy` only ensures that the data is not a view and makes a copy if it is. Since sparse matrices do not have views (at least I think so) I propose to skip the check. Currently the copy is made in the constructor anyway and hence we probably do not need to make an other here.

Orange/data/table.py
```diff
@@ -912,15 +912,16 @@ def is_copy(self):
 
     def ensure_copy(self):
         """
-        Ensure that the table owns its data; copy arrays when necessary
+        Ensure that the table owns its data; copy arrays when necessary.
+        The check is skipped for sparse matrices since they don't have views like numpy arrays.
         """
-        if self.X.base is not None:
+        if not sp.issparse(self.X) and self.X.base is not None:
```
astaric: What about introducing a functions like

def is_view(x):
    return not sp.issparse(x) and x.base is not None

and using it in ifs?

nikicc: Yup, that's nicer. Corrected.

astaric:
Could you also add a test for copy with sparse data?

nikicc:
A simple test was added 😄

PR #1458, opened 2016-07-14
[FIX] Invalid number format in KMeans widget

janezd:
The format string was, for instance, `{:.2.0f}` instead of `{:.2f}`.


janezd:
@kernc, was there any particular reason for replacing `int` with `np.floor`?

kernc:
Already [fixed](https://github.com/biolab/orange3/pull/1458/files#diff-0913fd017e312578148451e62fe5d98eL269) in https://github.com/biolab/orange3/pull/1434. `np.floor()` propagates `nan` (= silhouette score if not computed, see https://github.com/biolab/orange3/commit/074cecadf3c55af5e476694851aad50b8ba42a0f).

PR #1460, opened 2016-07-15
[FIX] statistics.util: Fix applying weights

nikicc:
This commit fixes three issues about weights. First, multiplication does work since X is 2d and weights are 1d (e.g. `operands could not be broadcast together with shapes (1000,2) (1000,)`). This is fixed by transforming weights to 2d. Further, weights should only be used for calculating mean. Third, we should not apply weights if X contains objects (like StringVariable).

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -5,7 +5,7 @@ and once used from the bottlechest package (fork of bottleneck).
 It also patches bottleneck to contain these functions.
 """
 import numpy as np
-from scipy.sparse import issparse
+import scipy.sparse as sp
 import bottleneck as bn
 
 
@@ -116,7 +116,7 @@ def contingency(X, y, max_X=None, max_y=None, weights=None, mask=None):
             continue
         col = X[..., i]
         nx = np.unique(col[~np.isnan(col)]).size if max_X is None else max_X + 1
-        if issparse(col):
+        if sp.issparse(col):
             col = np.ravel(col.todense())
         contingencies.append(
             bincount(y + ny * col,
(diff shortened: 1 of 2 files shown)


Orange/statistics/util.py
```diff
@@ -160,8 +160,8 @@ def stats(X, weights=None, compute_variance=False):
         If the length of the weight vector does not match the length of the
         array
     """
-    if weights is not None:
-        X = X * weights
+    if weights is not None and X.dtype != object:
+        X = X * weights[:, None]
```
kernc: Use `np.c_[weights]`. It will ensure it's a column as opposed to raising a dimension (which won't work if weights is already passed as column).

nikicc: That's handy. Fixed.

Orange/statistics/util.py
```diff
@@ -160,8 +160,8 @@ def stats(X, weights=None, compute_variance=False):
         If the length of the weight vector does not match the length of the
         array
     """
-    if weights is not None:
-        X = X * weights
+    if weights is not None and X.dtype != object:
+        X = X * np.c_[weights]
```
kernc:
```py
from scipy.sparse import diags
X = X * diags(weight.ravel())

File: Orange/statistics/util.py [hunk 3]
```diff
@@ -160,11 +160,17 @@ def stats(X, weights=None, compute_variance=False):
...
-    if weights is not None:
-        X = X * weights
     is_numeric = np.issubdtype(X.dtype, np.number)
     is_sparse = issparse(X)
 
+    if weights is not None and X.dtype != object:
+        if is_sparse:
+            # How row-wise multiply sparse matrices:
+            # http://stackoverflow.com/questions/12237954/multiplying-elements-in-a-sparse-array-with-rows-in-matrix
+            X.data *= weights.repeat(np.diff(X.indptr))
```
kernc: This only works if `issparse(X) and isinstance(X, csr_matrix)`.

PR #1464, opened 2016-07-15
Changed kMeans default parameters and added some tests

rokgomiscek:


Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -39,7 +39,7 @@ class OWKMeans(widget.OWWidget):
 
     resizing_enabled = False
 
-    k = Setting(8)
+    k = Setting(3)
```
astaric: Why is 3 a better default value than 8?

Orange/widgets/unsupervised/tests/test_owkmeans.py
```diff
@@ -0,0 +1,43 @@
...
+
+from Orange.widgets.tests.base import WidgetTest
+from Orange.widgets.unsupervised.owkmeans import OWKMeans
+
+from Orange.data import Table
+class TestOWKMeans(WidgetTest):
+
+    def setUp(self):
+        self.widget = self.create_widget(OWKMeans)
+        self.combo_box = self.widget.findChildren(QtGui.QComboBox)
```
astaric: A better way would be to modify the widget to store qt components to instance attributes and then use those. That way, the test would still work if new controls of the same type are added to the widget.

Orange/widgets/unsupervised/tests/test_owkmeans.py
```diff
@@ -0,0 +1,43 @@
...
+        self.spinner = self.widget.findChildren(QtGui.QSpinBox)
+        self.radio_button = self.widget.findChildren(QtGui.QRadioButton)
+        self.iris = Table("iris")
+
+    def test_boxes_visible(self):
+        """ Check if all boxes visible """
+        self.assertEqual(self.combo_box[0].isHidden(), False)
```
astaric: When would boxes not be visible? Does anyone ever hide them?

Orange/widgets/unsupervised/tests/test_owkmeans.py
```diff
@@ -0,0 +1,43 @@
...
+        self.assertEqual(self.combo_box[1].isHidden(), False)
+        self.assertEqual(self.combo_box[2].isHidden(), False)
+        self.assertEqual(self.spinner[0].isHidden(), False)
+        self.assertEqual(self.spinner[1].isHidden(), False)
+        self.assertEqual(self.spinner[2].isHidden(), False)
+
+
+    def test_optimization_report_display(self):
+        """ Check visibility of optimization report after we select the number of clusters  """
+        for button_pos, value in ((0, True),
```
astaric: I would rewrite this without a for, so it will look like a "scenario". Something like:

1. Select fix number of clusters
2. Check that optimization part is hidden
3. Select optimized number of clusters
4. Check that optimization part is shown
5. Click apply
6. Check that some lines were added to the optimization pane

Orange/widgets/unsupervised/tests/test_owkmeans.py
```diff
@@ -0,0 +1,43 @@
...
+                                  (1, False)):
+            self.radio_button[button_pos].click()
+            self.radio_button[button_pos].isChecked()
+            self.assertEqual(self.widget.mainArea.isHidden(), value)
+
+    def test_data_on_output(self):
+        """ Check if data is on output after create widget and run """
+        # Connect iris to widget
+        self.send_signal("Data", self.iris)
+        self.widget.run()
```
astaric: Rather than call the run method, I would call commit method (or even click the Apply button).

Orange/widgets/unsupervised/tests/test_owkmeans.py
```diff
@@ -0,0 +1,43 @@
...
+        self.assertNotEqual(self.widget.data, None)
+        # Disconnect the data
+        self.send_signal("Data", None)
+        # removing data should have cleared the output
+        self.assertEqual(self.widget.data, None)
```
astaric: There is a missing newline at the end of file (see the little red image shown on github).

Orange/widgets/unsupervised/owkmeans.py
rokgomiscek: I was following task 321 (Default parameters should be 3 fixed clusters and automatic apply.)

Orange/widgets/unsupervised/tests/test_owkmeans.py
rokgomiscek: I guess not... :laughing: Removed

Orange/widgets/unsupervised/tests/test_owkmeans.py
rokgomiscek: Modified the widget.

Orange/widgets/unsupervised/tests/test_owkmeans.py
rokgomiscek: Changed the scenario.

Orange/widgets/unsupervised/tests/test_owkmeans.py
rokgomiscek: Changed.

Orange/widgets/unsupervised/tests/test_owkmeans.py
rokgomiscek: Added newline at the end

Issue #1450, opened 2016-07-12
[owmergedata] Merge Data widget does not save selected attribute


stuart-cls:
##### Orange version
3.3.7.dev0+d6a324e

##### Expected behavior
Merge two datasets by a selected attribute, save `.ows`, load `.ows` file: worksheet remains the same.

##### Actual behavior
Merge two datasets by a selected attribute, save `.ows`, load `.ows` file: Merge Data widget has forgotten the selected merge attribute and all downstream widgets have no data.

##### Steps to reproduce the behavior
Load `zoo.tab` and `zoo-with-images.tab` datasets, merge by `name`, save worksheet, open worksheet.

PR #1468, opened 2016-07-18
OWMergeData: replace list views with combos, save settings

kernc:
Fixes https://github.com/biolab/orange3/issues/1450

### DIFF
--- a/Orange/widgets/data/owmergedata.py
+++ b/Orange/widgets/data/owmergedata.py
@@ -1,19 +1,20 @@
 import math
 import itertools
+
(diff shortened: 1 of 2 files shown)


ajdapretnar:
Workflow is now saved properly, but I'm facing another issue. When I input data with the same number of instances (say split titanic --> age, sex == A, status, survived == B), the default is merge by position (index). However, I get no input. Should't it just zip columns together as well? If I use merge by source position (index), data are output properly. Either this should be the default or merge by position should work.

kernc:
@ajdapretnar Can you have another look at this?

PR #1474, opened 2016-07-20
[FIX] Fix AdaBoost widgets and add some tests

VesnaT:


Orange/widgets/classify/owadaboost.py
```diff
@@ -26,39 +26,44 @@ class OWAdaBoostClassification(OWBaseLearner):
     learning_rate = Setting(1.)
     algorithm = Setting(0)
 
+    DEFAULT_BASE_ESTIMATOR = TreeLearner
```
kernc: Since this default base learner is always the same, might as well create a single class instance right here?

PR #1475, opened 2016-07-21
[FIX] KNN: Fix crash when Mahanalobis metric is used

VesnaT:

### DIFF
--- a/Orange/classification/knn.py
+++ b/Orange/classification/knn.py
@@ -1,15 +1,10 @@
 import sklearn.neighbors as skl_neighbors
+from Orange.base import KNNBase
 from Orange.classification import SklLearner
 
 __all__ = ["KNNLearner"]
 
 
-class KNNLearner(SklLearner):
+class KNNLearner(KNNBase, SklLearner):
     __wraps__ = skl_neighbors.KNeighborsClassifier
     name = 'knn'
-
-    def __init__(self, n_neighbors=5, metric="euclidean", weights="uniform",
-                 algorithm='auto',
-                 preprocessors=None):
-        super().__init__(preprocessors=preprocessors)
-        self.params = vars()
--- a/Orange/tests/test_knn.py
+++ b/Orange/tests/test_knn.py
@@ -7,23 +7,25 @@ import numpy as np
 
 from Orange.data import Table, Domain, ContinuousVariable, DiscreteVariable
 from Orange.classification import KNNLearner
-from Orange.evaluation import CA, CrossValidation
+from Orange.regression import KNNRegressionLearner
+from Orange.evaluation import CA, CrossValidation, MSE
 
 
 class TestKNNLearner(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.iris = Table('iris')
-        cls.learn = KNNLearner()
+        cls.housing = Table('housing')
 
     def test_KNN(self):
-        results = CrossValidation(self.iris, [self.learn], k=10)
+        results = CrossValidation(self.iris, [KNNLearner()], k=3)
         ca = CA(results)
         self.assertGreater(ca, 0.8)
         self.assertLess(ca, 0.99)
 
     def test_predict_single_instance(self):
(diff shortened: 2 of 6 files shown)


Orange/tests/test_knn.py
```diff
@@ -7,23 +7,27 @@
...
+from Orange.regression import KNNRegressionLearner
 from Orange.evaluation import CA, CrossValidation
 
 
 class TestKNNLearner(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.iris = Table('iris')
-        cls.learn = KNNLearner()
+        cls.learn = KNNLearner
```
kernc: What's wrong with a single learner instance?

Orange/classification/knn.py
```diff
@@ -11,5 +11,7 @@ class KNNLearner(SklLearner):
     def __init__(self, n_neighbors=5, metric="euclidean", weights="uniform",
                  algorithm='auto',
                  preprocessors=None):
+        if metric == "mahalanobis":
+            algorithm = "brute"
```
kernc: Please add a comment explaining why this is necessary. For the record, this works for me:
```py
>>> from sklearn.neighbors import KNeighborsClassifier
>>> nn = KNeighborsClassifier(metric='mahalanobis', algorithm='auto')
>>> X = np.random.random((10, 2))
>>> y = (np.random.random(10) > .5).astype(int)
>>> nn.fit(X, y)
>>> nn.predict(X)
array([1, 1, 1, 1, 0, 0, 0, 0, 0, 0])
```

VesnaT: For me as well. But i.e. X = np.random.random((12, 2)) doesn't.

Orange/tests/test_knn.py
```diff
@@ -7,23 +7,27 @@
...
+from Orange.regression import KNNRegressionLearner
+from Orange.evaluation import CA, CrossValidation, MSE
 
 
 class TestKNNLearner(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.iris = Table('iris')
-        cls.learn = KNNLearner()
+        cls.learn = KNNLearner
```
astaric: I would skip assigning classes in to cls.learn and cls.learn_reg and just use their names in tests.

self.learn() is not much shorter than KNNLearner() (in fact it has the same number of characters)

Orange/widgets/classify/tests/test_owknnclassification.py
```diff
@@ -1,53 +1,30 @@
...
-from PyQt4 import QtGui
-
-from Orange.widgets.tests.base import WidgetTest
 from Orange.widgets.classify.owknn import OWKNNLearner
+from Orange.widgets.tests.base import (WidgetTest, WidgetLearnerTestMixin,
+                                       GuiToParam)
 
 
-class TestOwKnnClassification(WidgetTest):
+class TestOWKNN(WidgetTest, WidgetLearnerTestMixin):
```
astaric: If the test case name matches TestClassName syntax, PyCharm will allow you to jump between tests and code using a keyboard shortcut (CMD + T on mac)

Orange/widgets/classify/tests/test_owknnclassification.py
```diff
@@ -1,53 +1,30 @@
...
-    def test_learner_on_output(self):
-        """ Check if learner is on output after create widget  and apply """
-        self.widget.apply()
-        self.assertNotEqual(self.widget.learner, None)
-
+        self.widget = self.create_widget(OWKNNLearner,
+                                         stored_settings={"auto_apply": False})
+        self.init()
+
+        def combo_set_value(i, x):
```
astaric: I have seen this in multiple test classes, should we change it to a static method on WidgetLearnerTestMixin?

VesnaT: Yes, we should. Was planning to do it after it is merged, to avoid conflicts.

astaric: ok

PR #1476, opened 2016-07-21
[ENH] TestWidgetTree: Cls and reg tree widgets test upgrade

VesnaT:

### DIFF
--- a/Orange/widgets/classify/tests/test_owclassificationtree.py
+++ b/Orange/widgets/classify/tests/test_owclassificationtree.py
@@ -1,7 +1,8 @@
 # Test methods with long descriptive names can omit docstrings
 # pylint: disable=missing-docstring
 from Orange.widgets.classify.owclassificationtree import OWClassificationTree
-from Orange.widgets.tests.base import WidgetTest, WidgetLearnerTestMixin
+from Orange.widgets.tests.base import (WidgetTest, WidgetLearnerTestMixin,
+                                       GuiToParam)
 
 
 class TestOWClassificationTree(WidgetTest, WidgetLearnerTestMixin):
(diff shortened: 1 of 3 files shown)


Orange/widgets/regression/tests/test_owregressiontree.py
```diff
@@ -1,11 +1,48 @@
...
+        """
+        self.widget.max_depth_spin[0].setCheckState(False)
+        self.widget.min_internal_spin[0].setCheckState(False)
+        self.widget.min_leaf_spin[0].setCheckState(False)
+        for i, val, f in zip(range(1, 4), [None, 2, 1],
+                             [lambda x: None, lambda x: 2, lambda x: 1]):
+            el = self.gui_to_params[i]
+            self.gui_to_params[i] = GuiToParam(
+                el.name, el.gui_el, f, el.set, [val], el.set_values[0:1])
+        super().test_parameters()
```
astaric: no need to use super() here. self will suffice.

Orange/widgets/regression/tests/test_owregressiontree.py
```diff
@@ -1,11 +1,48 @@
...
+                       lambda i, x: x.setValue(i), ml_min_max, ml_min_max)]
+
+    def test_parameters_unchecked(self):
+        """Check learner and model for various values of all parameters
+        when pruning parameters are not checked
+        """
+        self.widget.max_depth_spin[0].setCheckState(False)
+        self.widget.min_internal_spin[0].setCheckState(False)
+        self.widget.min_leaf_spin[0].setCheckState(False)
+        for i, val, f in zip(range(1, 4), [None, 2, 1],
```
astaric: I would rewrite this as a tuple of tuples,
 ((1, None, lambda x: None),
  (2, 2, lambda x:2),
  (3, 1, lambda x:1))
to make it more obvious which values go together.

After doing that, it looks like the third parameter (the function) just returns the value of second parameter. If this is the case,  could we just create the function inside for loop?

Orange/widgets/classify/tests/test_owclassificationtree.py
```diff
@@ -1,11 +1,48 @@
...
+                       lambda i, x: x.setValue(i), ml_min_max, ml_min_max)]
+
+    def test_parameters_unchecked(self):
+        """Check learner and model for various values of all parameters
+        when pruning parameters are not checked
+        """
+        self.widget.max_depth_spin[0].setCheckState(False)
+        self.widget.min_internal_spin[0].setCheckState(False)
+        self.widget.min_leaf_spin[0].setCheckState(False)
+        for i, val, f in zip(range(1, 4), [None, 2, 1],
```
astaric: Same comments apply as for the regression tree below.

PR #1477, opened 2016-07-22
[FIX] OWRandomForest: Fix, refactor and widget tests

VesnaT:

### DIFF
--- a/Orange/widgets/classify/owrandomforest.py
+++ b/Orange/widgets/classify/owrandomforest.py
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-
 from PyQt4 import QtGui
-from PyQt4.QtGui import QLabel, QGridLayout
-from PyQt4.QtCore import Qt
 
 from Orange.data import Table
 from Orange.classification.random_forest import RandomForestLearner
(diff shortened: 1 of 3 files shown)


Orange/widgets/regression/tests/test_owrandomforestregression.py
```diff
@@ -2,11 +2,65 @@
...
+        self.widget.max_features_spin[0].setCheckState(False)
+        self.widget.random_state_spin[0].setCheckState(False)
+        self.widget.max_depth_spin[0].setCheckState(False)
+        self.widget.min_samples_split_spin[0].setCheckState(False)
+        el = self.gui_to_params[4]
+        self.gui_to_params[4] = GuiToParam(el.name, el.gui_el, lambda x: 2,
+                                           lambda i, x: x.setValue(i), [2], [0])
+        super().test_parameters()
+        # reset checkbox to default
+        self.widget.min_samples_split_spin[0].setCheckState(True)
```
astaric: Why is this line needed? Each test should get its own instance of a widget with controls already set to default.

Orange/widgets/classify/owrandomforest.py
```diff
@@ -113,12 +71,16 @@ def create_learner(self):
...
-        self._min_samples_split_spin.setEnabled(self.use_min_samples_split)
+    def check_data(self):
+        self.valid_data = super().check_data()
+        if self.valid_data:
+            n_features = len(self.data.domain.attributes)
+            if self.use_max_features and self.max_features > n_features:
+                self.error(self.DATA_ERROR_ID,
+                           "Number of splitting attributes should "
+                           "be smaller than number of features.")
+                return False
```
astaric: should self.valid_data be set as well?

Orange/widgets/regression/tests/test_owrandomforestregression.py
VesnaT: I thought so too, but the settings are probably saved. However, I did not work without it.

Orange/widgets/classify/owrandomforest.py
VesnaT: Ups. I guess it should.

PR #1480, opened 2016-07-27
gui: Fix crash in doubleSpin when checked parameters is set

matevzkren:
The fix prevents the gui element to crash when 'checked' parameter is forwarded, ie. is not None.

### DIFF
--- a/Orange/widgets/tests/test_gui.py
+++ b/Orange/widgets/tests/test_gui.py
@@ -1,0 +1,15 @@
+from Orange.widgets import gui
+from Orange.widgets.tests.base import GuiTest
+from Orange.widgets.widget import OWWidget
+
+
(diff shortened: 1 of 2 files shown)


astaric:
Could you also write a test for this fix?

Orange/widgets/tests/test_gui.py
```diff
@@ -0,0 +1,15 @@
...
+from Orange.widgets.tests.base import GuiTest
+from Orange.widgets.widget import OWWidget
+
+
+class OWTestDoubleSpin(GuiTest):
+    some_param = 0
+    some_option = False
+
+    # make sure that the gui element does not crash when
+    # 'checked' parameter is forwarded, ie. is not None
```
kernc: Why don't make this a docstring?

PR #1481, opened 2016-07-28
[FIX] Table names set by readers

lanzagar:
Although Tables should ideally be constructed using Table methods (e.g. from_file), sometimes readers are used directly (as in owfile.py). Currently names are not set properly in that case, leading to 'untitled' datasets.
This changes readers (csv,tab,excel) to return more complete data Tables with names set.

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -669,8 +669,14 @@ class CSVReader(FileFormat):
                 try:
                     reader = csv.reader(file, dialect=dialect)
                     data = self.data_table(reader)
-                    data.name = os.path.splitext(
-                        os.path.split(self.filename)[-1])[0]
+
+                    # TODO: Name can be set unconditionally when/if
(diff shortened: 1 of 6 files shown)


astaric:
(Some) readers can work with data which is passed as a StreamIO (instead of the filename), which is used in tests to avoid writing unnecessary temp files. For those cases, name should still default to untitled (instead of crashing).

astaric:
And some tests would be nice :)

lanzagar:
Yuck!

```
Parameters
----------
filename : str
    name of the file to open
```
And then one subclass (without a comment) abuses the filename parameter for StringIO objects and we hack open to not open in some cases. And we want to maintain this although it is never used anywhere in the code apart from tests...?

But that should be handled in another PR / refactoring if we want this PR closed before winter :)

kernc:
It's not uncommon for file-like objects to be supported in place of filename paths.

PR #1487, opened 2016-07-28
Highcharts queued JS, license, and attribution

kernc:
`QApplication.processEvents()` is no longer called in the Highcharts constructor.

@PrimozGodec please review and confirm whether this works for you.


PrimozGodec:
Everything seems fine. Thanks @kernc

PR #1489, opened 2016-07-29
Move preprocess widget helpers into utils

markotoplak:


PrimozGodec:
May you also write some unit test for it?

janezd:
@markotoplak, @ales-erjavec, this PR seems harmless to merge, but I wonder whether these classes are not perhaps more general? Do they really belong `Orange.widgets.data.utils.preprocess` or could they be in something like `Orange.widgets.utils`, after perhaps some renaming?

ales-erjavec:
The StandardItemModel, Controler and SequenceFlow are (or could be made) more general.

The current patch does not move the `PreprocessAction` class defined in owpreprocess which is used by Controller class in `createWidgetFor` (that is one of the places that should be fixed, if it is to be used more generally).

janezd:
@markotoplak, would you move these classes to `Orange.widgets.utils`, together with the classes suggested by @ales-erjavec? If not, we can merge this PR as it is, and make them more general when we need it.

janezd:
@markotoplak, have we decided whether we want to generalize this or keep it as it is? In any case, can you rebase, please?

PR #1491, opened 2016-07-29
Highcharts - Map contains complete functionality of heatmap, so heatmap is removed. 

PrimozGodec:
Due to response on highcharts/highcharts#5546 (comment) map contains complete functionality of heatmap so heatmaps is removed. It also solves the problem with duplicating the legend for ColorAxis on heatmap chart.


PrimozGodec:
@kernc may you check it?

ajdapretnar:
Oh I so dream of having a highcharts heatmap one day instead of the one we have now. :heart:

kernc:
Remove the JS file as well.

PrimozGodec:
Done.

PR #1495, opened 2016-08-02
[FIX] Gini impurity: formula and docstring fixed.

ajdapretnar:
Gini coefficient and Gini impurity are not the same. Docstring fixed to proper Wikipedia source.

As observed in many [sources](https://en.wikipedia.org/wiki/Decision_tree_learning#Gini_impurity), the formula for Gini impurity never contains division by 2. @BlazZupan, please check whether this is the right formula or not.

Also, "Gini" was a misleading attribute name in Rank widget. Here we're actually measuring Gini decrease. So even perhaps "Gini Gain" might not be the right name for this. Comments? @lanzagar @astaric 

Probably some fixes to names should be made elsewhere, too.

### DIFF
--- a/Orange/preprocess/score.py
+++ b/Orange/preprocess/score.py
@@ -205,9 +205,9 @@ def _entropy(D):
 
 def _gini(D):
     """Gini index of class-distribution matrix"""
-    P = D / np.sum(D, axis=0)
-    return sum((np.ones(1 if len(D.shape) == 1 else D.shape[1]) - np.sum(np.square(P), axis=0))
-               * 0.5 * np.sum(D, axis=0) / np.sum(D))
+    P = np.asarray(D / np.sum(D, axis=0))
+    return np.sum((1 - np.sum(P ** 2, axis=0)) *
+                  np.sum(D, axis=0) / np.sum(D))
 
 
 def _symmetrical_uncertainty(X, Y):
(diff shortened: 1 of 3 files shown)


ajdapretnar:
Tests are now fixed as well.

Orange/preprocess/score.py
```diff
@@ -207,7 +207,7 @@ def _gini(D):
     """Gini index of class-distribution matrix"""
     P = D / np.sum(D, axis=0)
     return sum((np.ones(1 if len(D.shape) == 1 else D.shape[1]) - np.sum(np.square(P), axis=0))
-               * 0.5 * np.sum(D, axis=0) / np.sum(D))
+               * np.sum(D, axis=0) / np.sum(D))
```
kernc: This sums up [the proposed equation](https://en.wikipedia.org/wiki/Decision_tree_learning#Gini_impurity) somewhat more clearly:
```py
return 1 - (P*P).sum()
```

kernc: Among the several equivalent representations of [the proposed equation](https://en.wikipedia.org/wiki/Decision_tree_learning#Gini_impurity), this one seems nicest:
```py
return 1 - (P*P).sum()
```

ajdapretnar: I agree it looks the nicest and most concise. Can maybe @lanzagar also comment?

lanzagar: At first glance it looks to me like at least the second line is needed to get the weighted impurity we want.
It is possible that 1 can be used instead of the complications with np.ones (should be tested).

ajdapretnar: This is out of league. Perhaps merge this PR to fix division by two and make another one with improved formula? Otherwise I'll just copy-paste whatever you'll tell me to. :grinning:

lanzagar: Copy paste:
    P = np.asarray(D / np.sum(D, axis=0))
    return np.sum((1 - np.sum(P**2, axis=0))
               * np.sum(D, axis=0) / np.sum(D))

ajdapretnar: Fixed.

PR #1497, opened 2016-08-02
[FIX] OWHeatMap: Fix crash when creating an instance of the widget

VesnaT:
The widget crashed, because its properties (which use settings) were accessed before the settings were initialized.

### DIFF
--- a/Orange/widgets/visualize/owheatmap.py
+++ b/Orange/widgets/visualize/owheatmap.py
@@ -1295,8 +1295,12 @@ class OWHeatMap(widget.OWWidget):
             col_clust_msg = "Column clustering was disabled due to the " \
                             "input matrix being to big"
 
-        self.Information.row_clust(row_clust_msg)
-        self.Information.col_clust(col_clust_msg)
+        self.Information.row_clust.clear()
+        self.Information.col_clust.clear()
+        if row_clust_msg:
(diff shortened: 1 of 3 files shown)


ajdapretnar:
And you've written a test! :heart: Merging!

s-alexey:
> The widget crashed, because its properties (which use settings) were accessed before the settings were initialized.

You solution is work well for this widget, but I believe the major problem is `gettitem` [here](https://github.com/biolab/orange3/blob/a5dd7cb00876ee63a9084a28c96993d440622f09/Orange/widgets/utils/messages.py#L266) runs the getter method of properties. That's unsafe till `__init__` is called.

astaric:
I guess a better alternative would be to use [inspect.getmembers](https://docs.python.org/3.5/library/inspect.html#inspect.getmembers) and a [isclass](https://docs.python.org/3.5/library/inspect.html#inspect.isclass) predicate instead of calling getattr on names returned by dir().

ajdapretnar:
Perhaps this is my fault. We discussed with @VesnaT regarding the PR and decided we need a quick fix for Heat Map, to prevent it from crashing. HeatMap per se needs a major rehaul and I think Vesna did a lot of the work here that we decided to do in another major PR. This is just a quick fix for the immediate problem. But I might be wrong and there's a better quick fix for this. :open_mouth:

PR #1499, opened 2016-08-05
[ENH] Upgrade OWSvm unittests

VesnaT:

### DIFF
--- a/Orange/widgets/classify/owsvmclassification.py
+++ b/Orange/widgets/classify/owsvmclassification.py
@@ -68,7 +68,7 @@ class OWBaseSVM(OWBaseLearner):
     def _add_optimization_box(self):
         self.optimization_box = gui.vBox(
             self.controlArea, "Optimization Parameters")
-        gui.doubleSpin(
+        self.tol_spin = gui.doubleSpin(
             self.optimization_box, self, "tol", 1e-6, 1.0, 1e-5,
             label="Numerical tolerance:",
             decimals=6, alignment=Qt.AlignRight, controlWidth=100,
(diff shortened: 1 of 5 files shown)


Orange/widgets/classify/owsvmclassification.py
```diff
@@ -173,7 +176,7 @@ def create_learner(self):
         common_args = dict(
             kernel=kernel,
             degree=self.degree,
-            gamma=self.gamma,
+            gamma=self.gamma if self.gamma else self._default_gamma,
```
kernc: This is the same as
```py
self.gamma or self._default_gamma
```

Orange/widgets/classify/tests/test_owsvmclassification.py
```diff
@@ -1,107 +1,69 @@
...
-
-    def test_output_signal_learner(self):
-        """ Check if we have on output learner """
-        self.widget.kernel_box.buttons[0].click()
-        self.widget.set_data(Table("iris")[:100])
-        self.widget.apply()
-        self.assertNotEqual(self.widget.learner, None)
+    def test_kernel_equation(self):
+        """Check if the right equation is written according to kernel """
+        for i in range(4):
```
astaric: The old class included a test case that checked, whether correct combination of kernel selection, parameters and formula is used when widget is shown. (This was a regression test for bug we have encountered before.) Could the same check be added to this test (and the next one for parameter box visibiity)?

astaric: Resolved

PR #1503, opened 2016-08-08
[FIX] PyTableModel: DisplayRole value presentation

matevzkren:
When data is requested (and DisplayRole forwarded), value presentation is now correct & consistent, even if abs(value) equals to 0. This handles a display bug (consequently, no tests are included).


kernc:
This is not a [FIX] since it was not a bug. It's a representation enhancement, but it doesn't deserve [ENH] as we use it since it's not something to be featured in the changelog (or [brag to mother about](https://github.com/biolab/orange3/commit/2cd5e28a1b01055c52daf422b8fef2f5995e1ec1#diff-195a635ad245ded9076330e31134bd80R20)).

PR #1506, opened 2016-08-10
Created `__repr__` functions for some preprocessors

Ameobea:
This commit creates `__repr__` functions for the majority of Orange's learner and preprocessor classes.  It uses standardized and generic functions wherever possible.  These changes pave the way for my code generator (PR upcoming), which makes extensive use of these functions to create runnable Python scripts based on the contents of the workflow.

I moved some data processing functions out of widget code and into more appropriate directories, updating the widget code as appropriate.  I converted some functions to classes in order to make them more similar to others and created __repr__ functions for them as well.  Their usages were updated accordingly.  I then squashed all commits into a single commit.

All tests are passing (except those that weren't passing before) and the linter doesn't show anything significant caused by this commit.  As far as I can tell, this commit is ready to merge and completely working.  If any changes need to be made, please let me know.

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -210,10 +210,11 @@ class SklLearner(Learner, metaclass=WrapperMeta):
     _params = {}
 
     name = 'skl learner'
-    preprocessors = [RemoveNaNClasses(),
-                     Continuize(),
-                     RemoveNaNColumns(),
-                     SklImpute()]
+    preprocessors = default_preprocessors = [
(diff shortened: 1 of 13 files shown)


kernc:
You needn't close a PR if you mess something up and intend to open a new one. You can just add additional commits to it, or amend existing commits on the branch and then force push.

Ameobea:
I had tried that, but it seems that the PR hasn't updated even though the branch has.

PR #1508, opened 2016-08-11
Fix repetition of descriptions 

s-alexey:
image

### DIFF
--- a/Orange/canvas/registry/qt.py
+++ b/Orange/canvas/registry/qt.py
@@ -339,7 +339,7 @@ def whats_this_helper(desc, include_more_link=False):
 
     template = ["<h3>{0}</h3>".format(escape(title))]
 
-    if description:
+    if description and not long_description.startswith(description):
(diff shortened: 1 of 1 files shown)


janezd:
@ajdapretnar, should I check and merge this PR or -- as we discussed today -- move the existing docstrings into descriptions, wait for @s-alexey to do the same with his widgets and then remove this functionality (and thus make this PR redundant)?

ajdapretnar:
I personally vote for keeping descriptions in descriptions variable. 

I'm happy to alter the widgets and submit a new PR if that's required. But this would also mean fixing whats_this_helper() function. Again, my vote goes again displaying long_description in the first place, since it's limited to 100 characters anyway, which sometimes cuts off the description mid-way, and since the user is generally simply wanting to see what the widget does (short description). For more, there's always help. My 2c.

janezd:
@ajdapretnar will move texts from docstrings to short descriptions and remove the long description from display.

PR #1510, opened 2016-08-13
[ENH] Simple benchmarking suite.

sstanovnik:
To ease my benchmarking of pandas code, I created a simple benchmark suite. It's based on `unittest` and is `timeit`-like, but does not use `timeit`.

I've included some basic tests. See if this is something we want included with Orange. I'm only not sure about the location, for now, I've put this in `/benchmark/`

Example output on current master:
```
Running benchmark with CPython v3.5.2 on Darwin-15.6.0-x86_64-i386-64bit (OSX 10.11.6 x86_64)
[adult_create_X] with 50 loops, best of 3:
	min 0.518 usec per loop
	avg 0.528 usec per loop

[adult_filter] skipped: Not a pandas environment.
[adult_read] with 5 loops, best of 3:
	min 748 msec per loop
	avg 786 msec per loop
[iris_basic_stats] with 50 loops, best of 3:
	min 206 usec per loop
	avg 247 usec per loop
[iris_contingency] with 10 loops, best of 3:
	min 200 usec per loop
	avg 211 usec per loop
[iris_create_X] with 100 loops, best of 3:
	min 0.491 usec per loop
	avg 0.525 usec per loop
[iris_discretize] with 10 loops, best of 3:
	min 662 usec per loop
	avg 991 usec per loop
[iris_distributions] with 20 loops, best of 3:
	min 110 usec per loop
	avg 115 usec per loop

[iris_iteration_pandas] skipped: Not a pandas environment.
[iris_iteration_pre_pandas] with 10 loops, best of 3:
	min 681 usec per loop
	avg 782 usec per loop
[iris_read] with 100 loops, best of 3:
	min 2.49 msec per loop
	avg 2.87 msec per loop
```

### DIFF
--- a/benchmark/bench_basic.py
+++ b/benchmark/bench_basic.py
@@ -1,0 +1,65 @@
+from .base import Benchmark, benchmark, pandas_only, non_pandas_only
+from Orange.data import Table
+from Orange.preprocess import Discretize
+from Orange.preprocess.discretize import EqualFreq
+
+
+# noinspection PyStatementEffect
+class BenchBasic(Benchmark):
+    def setUp(self):
+        self.setup_test_string = "sttss"
+        self.iris = Table('iris')
+        self.adult = Table('adult')
+        self.discretizer = Discretize(EqualFreq(n=3))
+
+    def function_setup(self):
+        self.func_setup_str = "fsstr"
+
+    @benchmark(number=100)
+    def bench_iris_read(self):
+        Table('iris')
+
+    @benchmark(number=5, warmup=1)
+    def bench_adult_read(self):
+        Table('adult')
+
+    @benchmark(number=100)
+    def bench_iris_create_X(self):
+        self.iris.X
+
+    @benchmark(number=50)
+    def bench_adult_create_X(self):
+        self.adult.X
+
+    @pandas_only
+    @benchmark(number=20)
+    def bench_adult_filter(self):
+        self.adult[(self.adult.age > 30) & (self.adult.workclass == 'Private')]
+
+    @benchmark(number=50)
+    def bench_iris_basic_stats(self):
+        self.iris._compute_basic_stats()
+
+    @benchmark(number=20)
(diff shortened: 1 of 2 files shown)


kernc:
Could this be shaped into a set of performance guarantee tests which always run and fail if some change results in poorer performance over some threshold?

kernc:
> but does not use `timeit`.

You say this as if it were a feature? :smiley:

sstanovnik:
I don't think it can be done, atleast without significant effort. There is no travis-like tool that integrates with github, the only similar one I found was [asv](https://github.com/spacetelescope/asv), which seems local-only. Running on travis is impossible because it does not provide a consistent performance environment as far as I'm aware. The threshold is the other problem: you have to allow for some noise and, even then, you'd get a bazillion PRs that worsen the performance slightly and then another one that, maybe even because of noise, finally fails, without it even being its fault.

`timeit`-likeness is a feature only in that the output is similar and it runs tests `number`x`repeat` times :)

benchmark/bench_basic.py
```diff
@@ -0,0 +1,80 @@
...
+    from Orange.data.filter import FilterContinuous, FilterDiscrete, Values
+except:
+    # legacy only
+    pass
+
+
+# noinspection PyStatementEffect
+class BenchBasic(Benchmark):
+    def setUp(self):
+        self.setup_test_string = "sttss"
```
astaric: What is this?

benchmark/bench_basic.py
```diff
@@ -0,0 +1,80 @@
...
+        self.iris = Table('iris')
+        self.adult = Table('adult')
+        self.discretizer = Discretize(EqualFreq(n=3))
+
+    def function_setup(self):
+        self.func_setup_str = "fsstr"
```
astaric: And this?

sstanovnik: Whoops, this is a leftover from when I was testing whether everything works.

PR #1515, opened 2016-08-23
[ENH] Enable Ward clustering in Hierarchical clustering widget 

ales-erjavec:
* Enable ward clustering in Hierarchical clustering widget
* Implement a workaround for scipy's input checking in scipy.cluster.hierarchy.linkage in scipy 0.16 and 0.17 (scipy issue 5220)
* Adjust limits for clustering in 'Distance map' and 'Heatmap' widgets.

### DIFF
--- a/Orange/clustering/hierarchical.py
+++ b/Orange/clustering/hierarchical.py
@@ -392,12 +392,11 @@ def top_clusters(tree, k):
     heap = [item(tree)]
 
     while len(heap) < k:
-        key, cl = heapq.heappop(heap)
+        _, cl = heap[0]  # peek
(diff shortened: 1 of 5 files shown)


tomazc:
Looks good. @ales-erjavec, should we include a unit test for the currently untested [special handling](https://codecov.io/gh/biolab/orange3/src/098258549f9f0ddd80bf39229ec4f8a0771ea360/Orange/clustering/hierarchical.py#L104) of Ward linkage in scipy prior to 0.18? If yes, then we need to modify travis so that it will test also with older versions of scipy (current requirement is scipy>=0.11.0).

PR #1516, opened 2016-08-23
Changes to widget descriptions.

ajdapretnar:
As discussed with @janezd, I'm introducing standard widget descriptions that get displayed in canvas. Now only `description `variable gets displayed - it needs to be short and telling. `long_description `is no longer displayed but remains in the code. This also somehow prevents showing docstrings, which is good.

### DIFF
--- a/Orange/widgets/data/owcolor.py
+++ b/Orange/widgets/data/owcolor.py
@@ -275,14 +275,8 @@ class ContinuousTable(ColorTable):
 
 
 class OWColor(widget.OWWidget):
-    """
-    Set color legend for variables
-
-    Widget for defining specific color palettes for discrete and continuous
-    features
-    """
-
     name = "Color"
+    description = "Set color legend for variables."
     icon = "icons/Colors.svg"
     inputs = [("Data", Orange.data.Table, "set_data")]
     outputs = [("Data", Orange.data.Table)]
--- a/Orange/canvas/registry/description.py
+++ b/Orange/canvas/registry/description.py
@@ -198,9 +198,6 @@ class WidgetDescription(object):
         version.
     description : str, optional
         A short description of the widget, suitable for a tool tip.
-    long_description : str, optional
-        A longer description of the widget, suitable for a 'what's this?'
-        role.
     qualified_name : str
         A qualified name (import name) of the class implementing the widget.
     package : str, optional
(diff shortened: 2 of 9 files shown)


Orange/widgets/data/owcolor.py
```diff
@@ -275,14 +275,10 @@ def change_color(self, index):
...
-    """
-    Set color legend for variables
-
-    Widget for defining specific color palettes for discrete and continuous
-    features
-    """
-
     name = "Color"
+    description = "Set color legend for variables."
+    long_description = "Widget for defining specific color palettes " \
```
kernc: So what does `long_description` do anyway? Where is it used?

ajdapretnar: Just in the code now (description.py and setup.py). If you consider it obsolete, I'll throw it out altogether.

astaric: Long description is displayed in the box below the widgets toolbar when a widget is selected.

ajdapretnar: Not after this PR. I removed long_description from display as it sometimes fell out of bounds and it also somehow took docstrings with it (I don't know where that was set). I discussed it with @janezd and we decided to keep only short descriptions. If you have better suggestions, you're welcome.

astaric: Then +1 for removing them completely.

ajdapretnar: Fixed. I hope it's correct. Can you check especially description.py and then merge?

janezd:
Please also remove `long_description = None` from `OWWidget` class.

Could you/Have you checked whether any of our extensions use docstrings to give short descriptions and warned the authors (e.g. @PrimozGodec) to move them to `short_description`?

ajdapretnar:
> Please also remove long_description = None from OWWidget class.

Done.
No widgets use long_description anymore, not even in add-ons. And since [this fix](https://github.com/biolab/orange3/pull/1516/files#diff-7d694008bd30356fdf46c25e95e03c64L312), no docstrings get pulled into long_description. I'll make appropriate PRs to add-ons where descriptions have to be sorted.

janezd:
Thanks for fixing the mess I made. :)

PR #1517, opened 2016-08-24
Removed Univariate Regression widget from Orange core 

PrimozGodec:
This widget has been moved to Educational Add-on as a Polynomial Regression widget.


astaric:
Before we merge this, could you add replaces attribute to the widget in edicational add-on?

Replaces attribute lists paths where widget was located in the paths. When set correctly, opening a workflow that uses this widget from Orange will automatically update the workflow to use the widget from educational add-on.

(For example, see Silhouette plot: https://github.com/biolab/orange3/blob/master/Orange/widgets/visualize/owsilhouetteplot.py#L38)

PrimozGodec:
@astaric Done. Fixed as a part of this pull request https://github.com/biolab/orange3-educational/pull/17

ajdapretnar:
I merged PR#17, so I guess this can be merged as well?

ajdapretnar:
p.s. Wouldn't it make sense to remove it from Prototypes as well?

lanzagar:
In that case this can be merged too.

Yes, it would make sense to remove it from Prototypes.
If the version there had any modifications we would want to keep, I guess @PrimozGodec made sure to transfer them to the new version.

PrimozGodec:
I didn't know that this widget also exist in Prototypes. 

Version in prototypes is older than version which was moved to Educational add-on. I checked there is nothing more too keep, so I removed this widget from Prototypes. https://github.com/biolab/orange3-prototypes/pull/42

PR #1522, opened 2016-08-25
Updates to Loading your data subpage

ajdapretnar:
Small updates.


doc/visual-programming/source/loading-your-data/index.rst
```diff
@@ -6,19 +6,18 @@ also handle native Excel (.xlsx or .xls), comma- or tab-delimited data files. Th
 set is usually a table, with data instances (samples) in rows and
 data attributes in columns. Attributes can be of different type
 (continuous, discrete, and strings) and kind (input features, meta attributes, and class). Data attribute type and kind can be provided
-in the data table header. This can be changed later, after reading the
-data with :doc:`Select Columns<../widgets/data/selectcolumns>` widget.
+in the data table header. They can also be subsequently changed in the :doc:`File<../widgets/data/file>` widget,
```
kernc: subsequently → when importing data

doc/visual-programming/source/loading-your-data/index.rst
```diff
@@ -6,19 +6,18 @@ also handle native Excel (.xlsx or .xls), comma- or tab-delimited data files. Th
 set is usually a table, with data instances (samples) in rows and
 data attributes in columns. Attributes can be of different type
 (continuous, discrete, and strings) and kind (input features, meta attributes, and class). Data attribute type and kind can be provided
-in the data table header. This can be changed later, after reading the
-data with :doc:`Select Columns<../widgets/data/selectcolumns>` widget.
+in the data table header. They can also be subsequently changed in the :doc:`File<../widgets/data/file>` widget, 
+while data kind can also be modified with :doc:`Select Columns<../widgets/data/selectcolumns>` widget.
```
kernc: reword data kind?

BlazZupan: No, at least before we have something better. `data kind` and `data type` were intentionally put in the original writing to distinguish between the two header rows.

kernc:
*Role*, not kind.

doc/visual-programming/source/loading-your-data/index.rst
kernc: How do you feel about _role_ instead of kind?

BlazZupan: Perhaps, to be debated, but not for this pull request (would require changes elsewhere).

lanzagar:
Merged current improvements.
But I also think attribute _kind_ is probably not very clear and understandable for readers. I quite like the suggested _role_. Also I think it would be better to use "variable role" instead of "attribute role/kind", because attribute was usually used specifically for independent variables (i.e. it is already a specific role/kind). So an "attribute role/kind" of `class` is sort of a contradiction.

Issue #1527, opened 2016-08-28
Edit Domain crashes if discrete attributes are re-colored


janezd:
<img width="316" alt="screen shot 2016-08-28 at 18 19 44" src="https://cloud.githubusercontent.com/assets/2387315/18037515/10fca97a-6d4c-11e6-8d17-d12d6ebf5e5c.png">

```
  File "/Users/janez/Dropbox/orange3/Orange/widgets/data/oweditdomain.py", line 459, in transform
    if vdesc in self.domain_change_hints:
TypeError: unhashable type: 'list'
```

since the Variable contains colors as a list.

Thanks to bug #1526, this also crashes if Edit Domain is connected to File, as in the screenshot in #1526.

astaric:
Another take on this is in PR#1552. The problem occurs because the descriptions of variables in the old domain are not hashable, but we can perform the matching of variables on type, name and values alone.

markotoplak:
Related: color widget should survive even when colors are invalid.

PR #1530, opened 2016-08-29
[WIP][FIX] OWEditDomain: Fix crash when variable's attributes are represented as…

VesnaT:
… lists

Fixes https://github.com/biolab/orange3/issues/1527

### DIFF
--- a/Orange/widgets/data/oweditdomain.py
+++ b/Orange/widgets/data/oweditdomain.py
@@ -46,18 +46,20 @@ def variable_description(var):
 
     """
     var_type = type(var)
+    attrs = tuple((k, tuple(v) if isinstance(v, list) else v)
+                  for k, v in sorted(var.attributes.items()))
     if var.is_discrete:
         return (var_type.__module__,
                 var_type.__name__,
                 var.name,
                 (("values", tuple(var.values)),),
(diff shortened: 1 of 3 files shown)


Orange/widgets/data/oweditdomain.py
```diff
@@ -51,13 +51,13 @@ def variable_description(var):
                 var_type.__name__,
                 var.name,
                 (("values", tuple(var.values)),),
-                tuple(sorted(var.attributes.items())))
+                tuple((k, tuple(v)) for k, v in sorted(var.attributes.items())))
```
astaric: What will happen if a variable has an attribute that has an int-value?

janezd:
EditDomain is not the only widget that crashes in this way. Perhaps it would be better if we decide that all Variable's values must be hashable and fix the coloring to use tuples, instead of fixing all other widgets.

kernc:
What Variable's values are those? Variable's attributes' keys?

janezd:
@VesnaT, @astaric, #1552 also fixed the problem in `get_dict`, right? Can we close this PR?

astaric:
The tests from this PR have been cherry-picked to #1552, the rest was solved in a different way.

PR #1531, opened 2016-08-30
OWDistances: warn about imputing missing values

kernc:


Orange/widgets/unsupervised/owdistances.py
```diff
@@ -121,6 +123,8 @@ def commit(self):
             if numpy.product(shape) == 0:
                 self.Error.empty_data(shape)
             else:
+                if bn.anynan(data.X):
```
lanzagar: Can this be moved closer to the actual check for missing data and imputation. Reusing the same conditional check if possible (currently appears to be coupled with discretization, so might need some exercise).

ajdapretnar:
Isn't this supposed to show warning when imputing missing values in Distances widget? Why can't I see this message?

lanzagar:
Fixed in #1551

kernc:
I don't see how this is _replaced_ by #1551. The point of warning of imputation is that you can get different results depending on imputation axis.

kernc:
Nevermind.

PR #1537, opened 2016-09-01
Linear regression: Move mixing ratio under the slider

janezd:
At some point (perhaps when reorganizing the bottom parts of the widgets), the label move to some weird place in the layout.


lanzagar:
Slightly related - I find it frustrating that after moving the slider it is impossible for me to move it back to 0.50:0.50. It jumps from 0.49 to 0.51, _all_ other proportions can be set :)

ajdapretnar:
I can totally set the slider back to 0.50:0.50. :open_mouth:

PR #1538, opened 2016-09-02
ScatterPlot: Filter metas in Shape, Size, ...

astaric:
Once upon a time, metas were treated as inferior and were ignored by the ContextSettings unless explicitly told not to.

This PR fixes ScatterPlot, but in general, if we are going to treat metas as first class citizens (like we already do in many widgets), it makes sense to allow them to be used in all ContextSettings. This can be done by changing the default (see the TODO).


janezd:
This would be a life saver if there was a way to distribute it immediately. We moved the deadline for homeworks due to this bug.

I added a PR with tests. Merge them if you like them, and I'm merging this PR.

janezd:
@astaric, travis is stuck. Can you push again?

PR #1539, opened 2016-09-02
[ENH] PCA transformation speedup

markotoplak:
Speedup of transformations into the PCA space on my laptop:
- adult:
  6.07s -> 0.21s
- an infrared dataset with shape (16384, 1608):
  (more than 40 minutes - cancelled) -> 6.69s

### DIFF
--- a/Orange/data/util.py
+++ b/Orange/data/util.py
@@ -28,3 +28,30 @@ def scale(values, min=0, max=1):
     if ptp == 0:
         return np.clip(values, min, max)
     return (-minval + values) / ptp * (max - min) + min
+
+
+class SharedComputeValue:
+    """Structures compute_value functions that repeat parts of computation
+    for different variables into shared and specific parts.
+
+    Parameters
+    ----------
+    compute_shared: a callable that takes a Orange.data.Table
+        A callable that performs computation that is shared between
+        multiple variables. Variables sharing computation need to set
+        the same instance.
+    """
+
+    def __init__(self, compute_shared):
+        self.compute_shared = compute_shared
+
+    def __call__(self, data, shared_data=None):
+        """Fallback if common parts are not passed."""
+        if shared_data is None:
+            shared_data = self.compute_shared(data)
+        return self.compute(data, shared_data)
+
+    def compute(self, data, shared_data):
+        """Given precomputed shared data, perform variable-specific
+        part of computation and return new variable values."""
+        raise NotImplementedError
--- a/Orange/data/util.py
+++ b/Orange/data/util.py
@@ -31,12 +31,12 @@ def scale(values, min=0, max=1):
 
 
 class SharedComputeValue:
-    """Structures compute_value functions that repeat parts of computation
+    """A base class that separates compute_value computation
     for different variables into shared and specific parts.
 
     Parameters
     ----------
-    compute_shared: a callable that takes a Orange.data.Table
+    compute_shared: Callable[[Orange.data.Table], object]
         A callable that performs computation that is shared between
         multiple variables. Variables sharing computation need to set
         the same instance.
--- a/Orange/projection/pca.py
+++ b/Orange/projection/pca.py
@@ -11,6 +11,7 @@ except ImportError:
         pass
 
 import Orange.data
+from Orange.data.util import SharedComputeValue
 from Orange.data import Variable
 from Orange.misc.wrapper_meta import WrapperMeta
 from Orange.preprocess import Continuize
@@ -89,11 +90,27 @@ class _LinearCombination:
             for a, m, w in zip(self.attrs, self.mean, self.weights))
 
 
+class _PCATransformDomain:
+    """Computation common for all PCA variables."""
+
+    def __init__(self, pca):
+        self.pca = pca
+
+    def __call__(self, data):
+        if data.domain != self.pca.pre_domain:
+            data = data.from_table(self.pca.pre_domain, data)
+        return self.pca.transform(data.X)
+
+
 class PCAModel(Projection, metaclass=WrapperMeta):
+
     def __init__(self, proj, domain):
+        pca_transform = _PCATransformDomain(self)
+
         def pca_variable(i):
             v = Orange.data.ContinuousVariable(
-                'PC%d' % (i + 1), compute_value=Projector(self, i))
+                'PC%d' % (i + 1),
+                compute_value= Projector(self, i, pca_transform))
             v.to_sql = _LinearCombination(
                 domain.attributes, self.components_[i, :],
                 getattr(self, 'mean_', None))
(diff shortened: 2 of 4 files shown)


Orange/data/util.py
```diff
@@ -28,3 +28,27 @@ def scale(values, min=0, max=1):
     if ptp == 0:
         return np.clip(values, min, max)
     return (-minval + values) / ptp * (max - min) + min
+
+
+class ComputeValueWCommon:
```
kernc: Does not belong here. Put it in pca.py, where it's used.

Orange/data/util.py
```diff
@@ -28,3 +28,27 @@ def scale(values, min=0, max=1):
...
+    def __call__(self, data, common=None):
+        """Fallback if common parts are not passed."""
+        if common is None:
+            common = self.commonfn(data)
+        return self.specific(common, data)
+
+    def specific(self, common, data):
+        """Given the common computation results, perform variable-specific
+        part of computation and return new variable values."""
+        raise NotImplementedError("specific computation must be implemented")
```
kernc: Just if you'd care to know, I, personally, find `NotImplementedError` messages like this one 100% useless.

kernc:
While I enjoy the performance increase, I don't like how the change introduces considerable branching on the consumer end.

> The key to performance is elegance, not battalions of special cases. 
> — Jon Bentley and Doug McIlroy

Orange/projection/pca.py
```diff
@@ -137,16 +153,18 @@ def partial_fit(self, data):
...
         self.projection = projection
         self.feature = feature
         self.transformed = None
+        self.commonfn = commonfn #domain conversion to original data set,
+                                 #transform to PCA space
 
-    def __call__(self, data):
-        if data.domain != self.projection.pre_domain:
-            data = data.from_table(self.projection.pre_domain, data)
-        return self.projection.transform(data.X)[:, self.feature]
```
kernc: The result would be pretty much the same if you just somehow intelligently memoized this `self.projection.transform()` call?

Orange/data/util.py
markotoplak: This will be used for other transformations. It belongs in some base orange classes as it is referenced in table.py.

Orange/data/util.py
markotoplak: It is written as a lot of the code already in Orange. What would you recommend instead? Just "pass"?

Orange/projection/pca.py
markotoplak: This could ob course be memoized and shared between different compute_value functions for the same domain. But for lesser memory use that cache would have to be invalidated after transformation. Therefore, I did it so that the common part is exposed to the tranformer, which can destroy the intermediate results after transformation.

I do not get the your first comment, but is probably also irrelevant.

Orange/data/util.py
kernc: No, sure, there's nothing wrong with the line above. I just get to assume every developer who sees a `NotImplementedError` raised in a method of some base or mixin class understands that that is (one of) the method they are intended to override, thoughtfully seeking further information in the method's docstring, most likely. In this view, there is very little that
```py
raise NotImplementedError('you have to override this method')
```
adds compared to just
```py
raise NotImplementedError
```
Again, nothing _wrong_ with the code above. Sorry.

Orange/data/util.py
kernc: I don't know. Will be used for other transformations as part of this PR? Utils are tools, like screwdrivers and wrenches. This doesn't look like a tool, rather a very specific piece of plug I, unrelatedly, happen to not appreciate at all. :smiley:

Orange/projection/pca.py
kernc: I don't currently have a better idea. :confused: I just ~~feel~~know this isn't something a user of the projector should worry about.

Orange/data/util.py
janezd: I agree with @kernc. In general, I don't like the style in which we directly address the user or developer. Many errors/warnings in widgets used to say something like "Too few features. You need to provide at least two features to thingamagog". I prefer "Thingamagog needs at least two features."

In this case, even if you keep the message, rephrase it to "Method 'specific' is not implemented" (but this is obvious) or "... must be implemented", or something along this line. Just without "you". :)

Orange/data/util.py
markotoplak: Not, not as a part of this PR, but I'll definitely use it a lot in an add-on. This thing tries to at least partially solve the problem where a lot of features are based on the same computation. This also frequently occurs in text mining - also @nikicc was questioning it once.

Orange/data/util.py
markotoplak: I really like the NotImplementedError without messeges.

Orange/projection/pca.py
markotoplak: My line of reasoning. I would like to clear the cache after transformation. The only thing that knows when transformation ends is the user. Therefore the user has to be aware of it.

An alternative solution would be to call something like clear_cache() for all the variables after transformation, but this has the same awareness-drawback while being less direct. Also, theoretically, such caches would have problems with concurrent transforms...

Orange/data/util.py
```diff
@@ -28,3 +28,26 @@ def scale(values, min=0, max=1):
...
+class SharedComputeValue:
+    """Structures compute_value functions that repeat parts of computation
+    for different variables into shared and specific parts.
+
+    Attributes:
+        compute_shared  Function that performs computation that is shared
+                        between multiple variables. Variables sharing
+                        computation need to set the same function instance.
+    """
+    compute_shared = lambda data: None
```
astaric: def __init__(self, compute_shared):
   self.compute_shared = compute_shared

kernc:
```py
assert callable(compute_shared)
```

Orange/tests/test_data_util.py
```diff
@@ -16,3 +17,30 @@ def test_one_hot(self):
...
+        np.testing.assert_equal(res, res2)
+
+    def test_with_row_indices(self):
+        obj = PlusCommon()
+        obj.compute_shared = lambda data: 1.
+        data = Orange.data.Table("iris")
+        domain = Orange.data.Domain([Orange.data.ContinuousVariable("cv", compute_value=obj)])
+        data1 = Orange.data.Table.from_table(domain, data)[:10]
+        data2 = Orange.data.Table.from_table(domain, data, range(10))
+        np.testing.assert_equal(data1.X, data2.X)
```
astaric: Could you add another test that checks that shared computation is only executed once? You can use [mocks](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_count) as your shared computation functions and then check their call_count.

Orange/data/util.py
```diff
@@ -28,3 +28,30 @@ def scale(values, min=0, max=1):
...
     return (-minval + values) / ptp * (max - min) + min
+
+
+class SharedComputeValue:
+    """Structures compute_value functions that repeat parts of computation
+    for different variables into shared and specific parts.
+
+    Parameters
+    ----------
+    compute_shared: a callable that takes a Orange.data.Table
```
astaric: Callable[[Orange.data.Table], object] [PEP484](https://www.python.org/dev/peps/pep-0484/#callable)

Orange/data/util.py
```diff
@@ -28,3 +28,30 @@ def scale(values, min=0, max=1):
     if ptp == 0:
         return np.clip(values, min, max)
     return (-minval + values) / ptp * (max - min) + min
+
+
+class SharedComputeValue:
+    """Structures compute_value functions that repeat parts of computation
```
astaric: I would include the words Base class in this description

Orange/tests/test_data_util.py
astaric: Done in the test below

PR #1541, opened 2016-09-06
[FIX] Fixed svg function to return svg chart together with container div for highcharts

PrimozGodec:
It is required that also axis labels are provided in a report.  

Also fixed svg function in webview to take return whole svg. Before svg was returned without last `>`

### DIFF
--- a/Orange/widgets/highcharts.py
+++ b/Orange/widgets/highcharts.py
@@ -325,6 +325,12 @@ class Highchart(WebView):
         self._selection_callback([np.sort(selected).astype(int)
                                   for selected in points])
 
+    def svg(self):
+        """ Return SVG string of the first SVG element on the page, or
+        raise ValueError if not any. """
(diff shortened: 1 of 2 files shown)


PrimozGodec:
@kernc It is corrected now.

Orange/widgets/highcharts.py
```diff
@@ -325,6 +325,12 @@ def _on_selected_points(self, points):
         self._selection_callback([np.sort(selected).astype(int)
                                   for selected in points])
 
+    def svg(self):
+        """ Return SVG string of the first SVG element on the page, or
+        raise ValueError if not any. """
```
kernc: Add a comment why this is overridden and why a non-SVG is returned.

kernc: Please add a comment why this is overridden and why a non-SVG is returned.

PrimozGodec:
Done.

PR #1543, opened 2016-09-06
[ENH] widget: Allow subclasses to disable the default message bar widget

ales-erjavec:
<s> * Move the message bar to the bottom of the widget.</s>

* Allow subclasses to disable the default constructed message widget.

### DIFF
--- a/Orange/canvas/scheme/widgetsscheme.py
+++ b/Orange/canvas/scheme/widgetsscheme.py
@@ -711,7 +711,7 @@ def user_message_from_state(message_group):
         severity=message_group.severity,
         message_id=message_group,
         contents="<br/>".join(msg.formatted
-                              for msg in message_group.active.values()) or None,
+                              for msg in message_group.active) or None,
         data={"content-type": "text/html"})
 
 
--- a/Orange/widgets/tests/test_widget.py
+++ b/Orange/widgets/tests/test_widget.py
@@ -1,9 +1,12 @@
 # Test methods with long descriptive names can omit docstrings
 # pylint: disable=missing-docstring
+
+from unittest.mock import patch
+
 from Orange.widgets.gui import CONTROLLED_ATTRIBUTES, OWComponent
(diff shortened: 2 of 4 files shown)


janezd:
Having the bar on the top is bad because it moves the content of the widget and because it prevents us from listing multiple errors (so we need to use tooltips instead). I vote for moving the bar to the bottom, without an option of keeping them on the top. With this, we can remove the tooltips completely and instead expand the bar to contain multiple messages.

As for disabling the messages, the idea was that the user can disable a certain message for a certain widget. E.g. I no longer wish to be warned that Mosaic discretized some features. It is thus per-message.

I started implementing something a week ago, but then left if unfinished. You are inifinitely better in handling Qt, so I can leave this to you. See if there's anything useful in https://github.com/biolab/orange3/commit/51a732f8d9c79f9d703766f8316178841d99d67e. The idea is that each message gets a unique id composed of the widget name and the name of the attribute in which it is stored. This id is used as a key in QSettings.

It almost works, it just needs a button (commented out, connected to a string....) with which the user can silent the message + some visual redesign. 

We will also need a way to reenable the messages; probably in the Settings dialog.

janezd:
`self._set_hidden(msg, True)` in the commit I mention above is just for testing, of course. This would be triggered by user, not after showing the message for the first time.

One more thing: the user can only disable information and warning messages, not errors.

ales-erjavec:
> As for disabling the messages, the idea was that the user can disable a certain message for a certain widget. E.g. I no longer wish to be warned that Mosaic discretized some features. It is thus per-message.

This pull request is not about suppressing individual messages. Only about their display method in the OWWidget. It is in a sense equivalent to `wantStateInfoWidget` parameter in Orange 2.

janezd:
In `MessageGroup.clear()`: shouldn't `for msg in list(self.active):` be changed to `msg in list(self._active):`, too?

PR #1545, opened 2016-09-06
[FIX] Context attributes with metas in Sieve and Mosaic

janezd:
Fixes a similar problem as #1538, but for Sieve and Mosaic.

The fix for Mosaic has no effect since context settings in Mosaic do not work. @astaric, can you check why?

We can modify Mosaic to use models (like Sieve), which would make the problem go away, but we still need to fix the bug in settings (if there is one).

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -36,11 +36,11 @@ class OWMosaicDisplay(OWWidget):
 
     settingsHandler = DomainContextHandler()
     use_boxes = Setting(True)
-    variable1 = ContextSetting("")
-    variable2 = ContextSetting("")
-    variable3 = ContextSetting("")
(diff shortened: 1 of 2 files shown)


astaric:
What about setting the default value in ContextSetting to also include meta values?

astaric:
What is the problem with settings in Mosaic? If I modify settings and switch between two datasets, the modified settings are restored for each dataset. What do I need to do to replicate the problem?

janezd:
> What about setting the default value in ContextSetting to also include meta values?

You're the context expert. If it doesn't break anything, let's change the default. Will you do it and close this PR?

This PR also fixes a smaller glitch: the default value for the selection should be `set()` not `{}`. It doesn't affect anything, though.

> What is the problem with settings in Mosaic?

I can no longer replicate it either.

PR #1550, opened 2016-09-09
[FIX] owimageviewer: Open local images directly

ales-erjavec:
Using QNetworkAccessManager to 'retrieve' many local files can lead
to temporarily running out of available file descriptors.

Presumably this is because the associated open fd is not closed
until the QNetworkReply is actually deleted (and not just closed).

### DIFF
--- a/Orange/widgets/data/owimageviewer.py
+++ b/Orange/widgets/data/owimageviewer.py
@@ -1005,7 +1005,23 @@ class OWImageViewer(widget.OWWidget):
                 thumbnail.instance = inst
                 self.thumbnailView.addThumbnail(thumbnail)
 
-                if url.isValid():
+                if url.isValid() and url.isLocalFile():
+                    reader = QImageReader(url.toLocalFile())
+                    image = reader.read()
+                    if image.isNull():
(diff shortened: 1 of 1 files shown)


astaric:
I cannot reproduce the original problem on my machine. @BlazZupan, can you check if this solves the issue for you?

ales-erjavec:
> I cannot reproduce the original problem on my machine. @BlazZupan, can you check if this solves the issue for you?

Try running orange-canvas like so:
```sh
(ulimit -n 150; orange-canvas;)
```
(adjusting the limit appropriately to trigger the error)

janezd:
Perhaps related: trying to view the images from https://www5.cs.fau.de/fileadmin/research/datasets/fundus-images/all.zip in the Image Viewer on Windows shows the first 20 or so out of 45 images and then complains that it cannot open the file.

I've seen this on somebody's computer, but cannot replicate it neither on Mac nor on Windows.

BlazZupan:
Works fine on Mac. Should be checked on Win.

astaric:
I am merging this. Testing on windows can be done with daily builds, when they become available.

PR #1553, opened 2016-09-09
[WIP] report: report_plot works with QWidget

kernc:


astaric:
Is this used on any of the widgets already so I can see it in action?

kernc:
QGraphicsView from Network Explorer (report code https://github.com/biolab/orange3-network/pull/14).

astaric:
Using the suggested widget, I got the following report when reporting on a zommed in version of a graph.

screen shot 2016-09-15 at 14 17 33

Can you check whether the problem lies in report generating code or in the widget itself?

astaric:
Have you managed to take a look at the problem I mentioned above? If the problem is not related to this PR, are there any other widgets that I can see the change in action?

kernc:
The only case I had for plotting QWidgets in Report was in Network Explorer, but there the plot source can easily be switched from QGraphicsView to its scene, which is already supported. Therefore this PR is unneeded.

ajdapretnar:
> but there the plot source can easily be switched from QGraphicsView to its scene

I'd appreciate that so much! 🙌 Would it not make plotting a legend also easier (or at least bearable)?

PR #1554, opened 2016-09-09
[RFC] [ENH] canvas: Indicate runtime state on links

ales-erjavec:
Track and display the state of signal propagation on links in the canvas.


astaric:
Have you considered switching the state that is shown on widgets? (Like highlighting the channels that do not pass any data)?

janezd:
It looks great.

PR #1557, opened 2016-09-13
SharedComputeValue: functions called only once also with later transformations

markotoplak:

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -286,15 +286,15 @@ class Table(MutableSequence, Storage):
                                  [x - n_src_attrs for x in src_cols])
 
             a = np.empty((n_rows, len(src_cols)), dtype=dtype)
-            shared_cache = {}
+            shared_cache = cls.conversion_cache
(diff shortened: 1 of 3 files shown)


markotoplak:
@nikicc @astaric

markotoplak:
@nikicc @astaric This should work now.

nikicc:
It works on my data. Thanks!

Orange/data/table.py
```diff
@@ -35,6 +35,12 @@ def get_sample_datasets_dir():
 dataset_dirs = ['', get_sample_datasets_dir()]
 
 
+"""Domain conversion cache used in from_table. It is global so that
```
astaric: from_table -> Table.from_table (although it is the only such method, I would be explicit)

astaric:
I have restarted the Travis build since it failed on an unrelated error.

The rest looks ok to me. Is this still considered to be WIP?

markotoplak:
@astaric I think it is ready for merging.

Issue #1556, opened 2016-09-10
Rank widgets on data without the class variable


janezd:
... should still show results if it has scorers (such as PCA) that do not need the class attribute.

mstrazar:
When connecting the signal from PCA, the signal doesn't get processed by the signalmanager, and consequently the owrank.set_learner is never reached. 

Stderr output whn connecting PCA to owrank input:
2016-09-16 10:29:19.910 python[22808:3503627] modalSession has been exited prematurely - check for a reentrant call to endModalSession

Any ideas?

PR #1560, opened 2016-09-15
[ENH] OWRank: Widget improvements

### LINKED ISSUE #1547: Rank Widget Improvements
##### Orange version
3.3

##### Expected behavior
1. Add “Scoring” box. Shows scoring methods, full name, with checkboxes to decide if the column associated with the scoring method is displayed. Like in Orange2, but plain, with no parameters (always use defaults). Content (classification or regression) based on type of input data set (default, when no data, shows for classification). By default, following methods are checked: Classification: Gain Ratio, Gini Index, Regression: Linear regression, RReliefF (there are no others for now).
2. Scores should have a full name (shown with the checkboxes) and shorter name (column headers).
3. Scores should compute when the scoring method is enabled (checked). If the user unchecks a method, keep the scoring method in cache.
4. Selection of attributes is context dependent (applicable only if attributes are selected manually, not with None, All or Best ranked).
5. Adding scoring methods to the input of the widget only adds a new scoring column (does not alter the content of the box with checkboxes). These columns are always displayed.
6. Implement progress bar (with fat data sets, feature scoring takes quite some time).
7. Consider implementing the score in the background, check how this is done in Test & Score (this feature is optional, but may take substantial time if we postpone it after other issues with this widget are fixed).

##### Actual behavior
Currently the widget displays scores for all the measures. For classification, there are many, and can be quite confusing for the user. Plus computing them all can take substantial time.

VesnaT:
Enable selection of scoring method. Scores are computed only when checked and when deselected stored in cache for later use.
Attribute selection is context dependent and progress bar is added to the widget to show percentage of computed scores.

Fixes #1547 and #1556

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -14,12 +14,14 @@ from scipy.sparse import issparse
 from PyQt4 import QtGui
 from PyQt4.QtCore import Qt
 
-import Orange
 from Orange.base import Learner
-from Orange.data import ContinuousVariable, DiscreteVariable, Domain, StringVariable
+from Orange.data import (Table,  Domain, ContinuousVariable, DiscreteVariable,
+                         StringVariable)
 from Orange.preprocess import score
 from Orange.canvas import report
-from Orange.widgets import widget, settings, gui
+from Orange.widgets import widget, gui
+from Orange.widgets.settings import (DomainContextHandler, Setting,
+                                     ContextSetting)
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.widget import OWWidget, Msg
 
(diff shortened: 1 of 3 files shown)


ajdapretnar:
I like this very much. Not sure if this is a part of this PR or not, but there's an issue with Report. When reporting on all possible scoring methods for discrete class, the last few columns fall out of view. When saving to .pdf, columns names are shifted. Could probably be fixed in another PR I reckon.

ajdapretnar:
Important note: load *housing.tab* and connect it to Rank. Then connect PCA to Rank and Logistic Regression (it is wrong, I know). The widget crashes. Before it it displayed only the error message. Please check.

janezd:
Rebase, please.

While you're at it, can you also handle the case in which the data has no class, but it has an outside scorer (e.g. PCA). See issue #1556. As shown in PR #1566, it suffices to remove the error message and setting the data to None - except not in that you mustn't do it in `set_learner`.

kernc:
> When saving to .pdf, columns names are shifted.

Will be fixed in https://github.com/biolab/orange3/pull/1573/commits/71b62c8fd62292510cdff656fc52509a74ee5602

ajdapretnar:
Rank is not working for me anymore. @VesnaT can you please check?

janezd:
You can wait but you don't have to. It can be fixed in another PR.

PR #1562, opened 2016-09-16
[FIX] owlearnerwidget: Fix output initialization

ales-erjavec:
Always output the initial learner, disregarding the `auto_apply`.

### DIFF
--- a/Orange/widgets/utils/owlearnerwidget.py
+++ b/Orange/widgets/utils/owlearnerwidget.py
@@ -123,7 +123,7 @@ class OWBaseLearner(OWWidget, metaclass=OWBaseLearnerMeta):
         self.preprocessors = None
         self.outdated_settings = False
         self.setup_layout()
-        QTimer.singleShot(0, self.apply)
+        QTimer.singleShot(0, self.update_learner)
 
     def create_learner(self):
         """Creates a learner with current configuration.
--- a/Orange/widgets/utils/owlearnerwidget.py
+++ b/Orange/widgets/utils/owlearnerwidget.py
@@ -123,7 +123,7 @@ class OWBaseLearner(OWWidget, metaclass=OWBaseLearnerMeta):
         self.preprocessors = None
         self.outdated_settings = False
         self.setup_layout()
-        QTimer.singleShot(0, self.update_learner)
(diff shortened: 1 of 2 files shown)


Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -123,7 +123,7 @@ def __init__(self):
         self.preprocessors = None
         self.outdated_settings = False
         self.setup_layout()
-        QTimer.singleShot(0, self.apply)
+        QTimer.singleShot(0, self.update_learner)
```
janezd: Wouldn't it be safer to call `self.unconditional_apply` here? It doesn't change anything for the base class since `update_model` doesn't do anything without the data, but derived classes can add some code to the `apply` method or change `update_model` to do something without data.

ales-erjavec: I did try that. At the least that breaks 'Linear Regression' widget because it overrides the `add_bottom_buttons` method and does not 'define' `unconditional_apply` so it fails to even initialize.

janezd: I see. Still, I think it's too dangerous to skip the potentially overridden `apply`. It can lead to a situation in which the widget will, for instance, ignore some setting when force-applied.

Let's require all widgets derived from `OWBaseLearner` to have `unconditional_apply`. All that don't, need to be fixed. Linear regression, in particular, doesn't need to define `commit` at all (partially my fault :()

janezd: @ales-erjavec, @astaric, please see #1601, which should also fix this problem. Two other widgets in education remove buttons, while (AFAIK) no widgets in core, bio, datafusion, network, prototypes or text extend `BaseLearner`.

What about replacing the above line with
`QTimer.singleShot(0, getattr(self, "unconditional_apply", self.apply))`?

If the widget does not have `unconditional_apply` it has either left `apply` as it is, hence calling it has the same effect as `unconditional_apply` would, or it has overridden `apply`, so we must call it anyway. If the widget creates its own Apply button with a callback with a different name, then it will have to reimplement the output initialization, too.

janezd: 'this problem' = the problem with linear regression, not the problem fixed by this PR here.

PR #1564, opened 2016-09-16
CN2: simplify documentation

janezd:
A quick fix to move the documentation about CN2, which is too specific, out from the basic tutorial.


Orange/classification/rules.py
```diff
@@ -1,3 +1,12 @@
+"""
+Induction of rules works by finding a rule that covers some learning instances,
+removing these instances, and repeating this until all instances are covered.
+The rules are scored by heuristics such as impurity of class distribution of
+the covered instances. The module includes the common rule-learning algorithms,
```
ajdapretnar: the covered instances --> covered instances :)

ajdapretnar: The module includes the common rule-learning algorithms -->
The module includes common rule-learning algorithms

Orange/classification/rules.py
```diff
@@ -1,3 +1,12 @@
...
+and allows for replacing search rule search strategies, scoring and other
```
ajdapretnar: > search rule search strategies

?

Orange/classification/rules.py
janezd: I understand the first one, but not the second. There are some common rule-learning algorithms, namely two variants of CN2. Wouldn't these then be *the* common rule-learning algorithms?

janezd: Fixed (all three).

ajdapretnar:
Probably it would make sense to fix index.rst with this PR, too. **CN2 Rule Viewer** is in Visualize, while **CN" Rule Induction** is in Classify. Index.rst puts both of them in Classify (incorrect). Or perhaps we should indeed move Viewer to classify, just like we have Classification Tree Viewer there.

Orange/classification/rules.py
ajdapretnar: Since you're not mentioning what these two algorithms are, I reckon no. I'm no expert, mind you, but I'd go with either *two common rule-learning algorithms* or without *the*. Especially since it's plural. My 2c.

janezd:
I fixed the documentation. @BlazZupan suggested moving *the model viewers* (yes,  we know which viewers we are talking about :) to visualization, including, eventually, the tree viewer.

PR #1566, opened 2016-09-16
Rank widget computes scores when domain has no class variable, provided a Scorer. 

### LINKED ISSUE #1556: Rank widgets on data without the class variable

mstrazar:
Error reporting has been modified accordingly.

Fixes #1556


janezd:
Works only if the widget gets the scorer after receiving the data. Even if the scorer comes later, the error re-appears when the data is reloaded.

I'm closing the PR because a fix in this fashion can't work and because #1560 turned the widget upside down anyway, so this PR is no longer applicable.

PR #1567, opened 2016-09-16
pylintrc: make more lax to show fewer errors

kernc:
Linter is meant to be helpful, not ignored.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -70,7 +70,9 @@ disable=
     standarderror-builtin,import-star-module-level,old-octal-literal,
     metaclass-assignment,reduce-builtin,indexing-exception,
     parameter-unpacking,long-suffix,cmp-method,using-cmp-argument,
-    useless-suppression
+    useless-suppression,import-error,
+    empty-docstring,missing-docstring,redefined-outer-name,redefined-builtin,
+    attribute-defined-outside-init
 
 
 [REPORTS]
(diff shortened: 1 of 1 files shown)


pylintrc
```diff
@@ -335,19 +326,19 @@ max-args=20
 
 # Argument names that match this expression will be ignored. Default to name
 # with leading underscore
-ignored-argument-names=_.*
+ignored-argument-names=^_
 
 # Maximum number of locals for function / method body
-max-locals=20
+max-locals=100
```
janezd: 100 local variables looks huge (it actually exceeds the maximal number of statements). I hated this warning, but it still made me think whether there's something I can simplify.

pylintrc
```diff
@@ -335,19 +326,19 @@ max-args=20
 
 # Argument names that match this expression will be ignored. Default to name
 # with leading underscore
-ignored-argument-names=_.*
+ignored-argument-names=^_
```
janezd: Sure? Wouldn't we like to keep an option to have "undocumented" arguments that may be removed in the future?

pylintrc
```diff
@@ -210,61 +211,51 @@ include-naming-hint=yes
...
 # Regular expression matching correct function names
 function-rgx=[a-z_][a-zA-Z0-9_]{2,30}$
-
 # Naming hint for function names
 function-name-hint=[a-z_][a-zA-Z0-9_]{2,30}$
 
 # Regular expression matching correct variable names
-variable-rgx=[a-zA-Z_][a-zA-Z0-9_]{1,30}$
-
+variable-rgx=[a-zA-Z_][a-zA-Z0-9_]{0,30}$
```
janezd: Thanks!

pylintrc
kernc: `_.*` and `^_` match the same things.

pylintrc
kernc: Reduced to 60. If code looks generally untidy, the reviewer will catch this at manual inspection anyway.

The rationale here in general is to make our linter useful instead of ignored.

pylintrc
```diff
@@ -70,7 +70,8 @@ disable=
     standarderror-builtin,import-star-module-level,old-octal-literal,
     metaclass-assignment,reduce-builtin,indexing-exception,
     parameter-unpacking,long-suffix,cmp-method,using-cmp-argument,
-    useless-suppression
+    useless-suppression,import-error,
+    empty-docstring,missing-docstring,redefined-outer-name
```
astaric: I would also add

attribute-defined-outside-init, redefined-builtin

janezd: I would keep attribute-defined-outside-init. It saved me from lots of bugs with undefined attributes in widgets, and only seldom did I have to add something to `__init__` that really didn't belong there.

astaric: pylint is stupid and does not understand that `__init__` calls methods that in turn set attributes (i.e. when widgets use methods to construct gui)

PR #1571, opened 2016-09-16
travis: disable python_nightly tests

astaric:


.travis.yml
```diff
@@ -12,9 +12,7 @@ matrix:
           cache: {}
         - python: '3.4'
         - python: '3.5'
-        - python: 'nightly'
```
kernc: Would you just comment this line instead and leave the other one in?

astaric: I could, but I prefer to delete the code and make git remember what was there before.

PR #1573, opened 2016-09-16
[FIX] Reporting tabular in Data Table and Rank widgets

kernc:
QBrush. https://doc.qt.io/qt-4.8/qt.html#ItemDataRole-enum

Fixes report in Data Table.

### DIFF
--- a/Orange/canvas/report/report.py
+++ b/Orange/canvas/report/report.py
@@ -3,7 +3,7 @@ import time
 from collections import OrderedDict, Iterable
 from itertools import chain
 from PyQt4.QtCore import Qt, QAbstractItemModel, QByteArray, QBuffer, QIODevice, QLocale
-from PyQt4.QtGui import QGraphicsScene, QTableView, QColor
+from PyQt4.QtGui import QGraphicsScene, QTableView, QColor, QBrush
 
 from Orange.util import try_
 from Orange.widgets.io import PngFormat
@@ -222,10 +222,12 @@ class Report:
                             if view and row is not None and col is not None else False)
 
                 fgcolor = data(Qt.ForegroundRole)
-                fgcolor = fgcolor.color().name() if fgcolor else 'black'
+                fgcolor = (QBrush(fgcolor).color().name()
+                           if isinstance(fgcolor, (QBrush, QColor)) else 'black')
 
                 bgcolor = data(Qt.BackgroundRole)
-                bgcolor = bgcolor.color().name() if bgcolor else 'transparent'
+                bgcolor = (QBrush(bgcolor).color().name()
+                           if isinstance(bgcolor, (QBrush, QColor)) else 'transparent')
                 if bgcolor.lower() == '#ffffff':
                     bgcolor = 'transparent'
 
--- a/Orange/canvas/report/report.py
+++ b/Orange/canvas/report/report.py
@@ -257,6 +257,8 @@ class Report:
 
             if has_horizontal_header:
                 stream.append('<tr>')
+                if has_vertical_header:
+                    stream.append('<th></th>')
                 stream.extend(item_html(None, col) for col in columns)
                 stream.append('</tr>')
 
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -379,7 +379,7 @@ class OWDataTable(widget.OWWidget):
     def __init__(self):
         super().__init__()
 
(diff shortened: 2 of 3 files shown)


ales-erjavec:
Ok. But I think you should still handle the case where the data(Qt.BackgroundRole) is not a QBrush in `report_table`, as there is no guaranty what the return value from `QAbstractDataModel.data` actually is. At the very least it needs to check for `None` values (i.e. an empty/null `QVariant()`).

kernc:
In [Qt](https://github.com/qt/qtbase/search?p=1&q=backgroundrole&type=Code&utf8=%E2%9C%93), QBrush is always returned for this role. No guarantee, but a [_convention_](https://doc.qt.io/qt-4.8/qt.html#ItemDataRole-enum). If the user passes arbitrary types for predefined roles, that's a bug waiting to be discovered.

> For _user roles_, it is up to the developer to decide which types to use and ensure that components use the correct types when accessing and setting data.

Emphasis added.

`None` case [is handled](https://github.com/biolab/orange3/blob/f18f6a6fd0dad780e6ebacb02ae450ff145c3f89/Orange/canvas/report/report.py#L228).

ales-erjavec:
* https://github.com/qt/qtxmlpatterns/blob/678502cce013a01a344e7a54d06bbf2aa816af13/tests/auto/xmlpatternssdk/TestGroup.cpp#L75

* https://github.com/qt/qtcharts/blob/4cbd063d221985f1d659b50f7353902a26bcfd7d/tests/manual/chartdesigner/engine.cpp#L175

* https://github.com/qt/qtcharts/blob/4cbd063d221985f1d659b50f7353902a26bcfd7d/examples/charts/barmodelmapper/customtablemodel.cpp#L91

* https://github.com/qt/qtcharts/blob/4cbd063d221985f1d659b50f7353902a26bcfd7d/examples/charts/modeldata/customtablemodel.cpp#L90

* https://github.com/qt/qtcharts/blob/facc2941efbfb8c9f40e363f0ea881653f6b4393/tests/manual/candlesticktester/customtablemodel.cpp

* https://github.com/qtproject/qt-creator/blob/7f8acd8fda7c831a4085c167c16a54cea4702fa2/src/plugins/texteditor/colorschemeedit.cpp#L103

* https://github.com/qtproject/qt-creator/blob/15fbfaf2e94b72cfd54bb0bd6f2b9ab87b870795/src/plugins/cppeditor/cppinsertvirtualmethods.cpp#L443

Really anything which is `qvariant_cast<QBrush>()` convertible can be a returned/set for a Qt::BackgroundRole as far as Qt is concerned. This includes QColor and even QPixmap.

It is an unfortunate consequence of [PyQt QVariant api 2](http://pyqt.sourceforge.net/Docs/PyQt4/python_v3.html#qvariant) that the abstraction that is provided by the QVariant is no longer available.

Note that `qvariant_cast<T>()` or one of the equivalent converters is the only way to get any value out of a QVariant, and that if passed a value it cannot convert it returns a default constructed T(). So you might say that graceful fallback when presented with arbitrary types in Qt item/model view framework is not only a convention but a rule.

> None case is handled.

Sorry I did not see that.

kernc:
Thanks for the elaborate response. Please find the PR completely revised.

ajdapretnar:
Is this solving crash for Report? Because the fix works, but there's also the issue of this:

fall-out

Also it makes reporting very slow. I don't remember how it was before, but it takes about 30s-1m to report on adult.tab (Win10).

Orange/canvas/report/report.py
```diff
@@ -216,16 +219,18 @@ def data(role=Qt.DisplayRole,
                     if row is None or col is None:
                         return model.headerData(col if row is None else row,
                                                 orientation, role)
-                    return model.data(model.index(row, col), role)
+                    return try_fmtnum(model.data(model.index(row, col), role))
```
janezd: Do not reformat the number here, at least not like this.

Orange used to keep track of the number of decimal places it should print for a specific variable. For instance, the "ST by exercise" has only a single decimal in the heart_disease.tab file, and the "age" and the "number of vessels colored" have none, obviously. Orange used to record this when loading the data. Now all numeric values have three decimals.

I consider this a bug (#1581). After this is fixed, `model.data` that is used in, say, the Table widget, will (again) show the numeric values with appropriate number of decimals, so this function should not reformat it.

kernc: Rebased and without the offending commit.

PR #1580, opened 2016-09-20
[FIX] OWBaseLearner: Do not re-fit if name has changed

VesnaT:
When the name of the learner had been changed, the entire model was re-calculated.


kernc:
It probably doesn't re-calculate the model. Instead, the recipient widgets are sent the model again (Test & Score runs again, for example).

PR #1583, opened 2016-09-21
Sparse fixups

nikicc:
Various bug-fixes required for sparse data:

- `preprocess.Transformation` should not call .squeeze since sparse matrices does not support it.
- `ReplaceUnknown` should be skipped for sparse since _missing_ are considered to be zero.
- `Discretizer` support sparse data.
- `Normalizer` support zero-based sparse data.
- `Table.from_table` should consider sparsity of initial array. For example, when adding columns to sparse X it should remain sparse.
- `Table._compute_contingency` fix for sparse on bottleneck.
- `OWHeatMap` widget converts sparse data to dense. Though this is probably not optimal it at least enables the use of heat map on sparse data where number of attributes is reasonable. On larger we probably couldn't see anything anyway. I tried and failed to make it work on sparse. The problem is in pyqtgraph's `makeARGB` function which only accepts numpy arrays. Even if that was bypassed we would probably  need to create a dense pixmap array at some point.

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -264,15 +264,14 @@ class Table(MutableSequence, Storage):
 
         global _conversion_cache
 
-        def sparse_to_flat(x):
-            if sp.issparse(x):
-                x = np.ravel(x.toarray())
-            return x
-
-        def get_columns(row_indices, src_cols, n_rows, dtype=np.float64):
+        def get_columns(row_indices, src_cols, n_rows, dtype=np.float64,
+                        is_sparse=False):
 
             if not len(src_cols):
-                return np.zeros((n_rows, 0), dtype=source.X.dtype)
+                if is_sparse:
+                    return sp.csr_matrix((n_rows, 0), dtype=source.X.dtype)
+                else:
+                    return np.zeros((n_rows, 0), dtype=source.X.dtype)
 
             n_src_attrs = len(source.domain.attributes)
             if all(isinstance(x, Integral) and 0 <= x < n_src_attrs
@@ -280,7 +279,7 @@ class Table(MutableSequence, Storage):
                 return _subarray(source.X, row_indices, src_cols)
             if all(isinstance(x, Integral) and x < 0 for x in src_cols):
                 arr = _subarray(source.metas, row_indices,
-                                 [-1 - x for x in src_cols])
+                                [-1 - x for x in src_cols])
                 if arr.dtype != dtype:
                     return arr.astype(dtype)
                 return arr
@@ -289,7 +288,20 @@ class Table(MutableSequence, Storage):
                 return _subarray(source._Y, row_indices,
                                  [x - n_src_attrs for x in src_cols])
 
-            a = np.empty((n_rows, len(src_cols)), dtype=dtype)
+            if is_sparse:
+                a = sp.dok_matrix((n_rows, len(src_cols)), dtype=dtype)
+            else:
+                a = np.empty((n_rows, len(src_cols)), dtype=dtype)
+
+            def match_type(x):
+                """ Assure that matrix and column are both dense or sparse. """
+                if is_sparse == sp.issparse(x):
+                    return x
+                elif is_sparse:
+                    return sp.csc_matrix(x.reshape(-1, 1).astype(np.float))
+                else:
+                    return np.ravel(x.toarray())
+
             shared_cache = _conversion_cache
             for i, col in enumerate(src_cols):
                 if col is None:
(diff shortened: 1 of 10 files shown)


ajdapretnar:
Works for me. This doesn't fix HeatMap, does it? My problem now is how to visualize the results of SimHash since hashes are appended to the end of a sparse matrix and are practically invisible.... 😮 
However, I got no other errors apart from HeatMap, so we might want to merge this asap.

nikicc:
@ajdapretnar as of visualizing results of SimHash, why is your data even sparse? SimHash produces dense matrix. Or did you add similarity hashing after BoW? If that's the case, than just use SimHash on data before BoW was applied and doing so you can take SimHash matrix into HeatMap.

nikicc:
@ajdapretnar this now also includes a fix for heat map on sparse data. Notice, however, that sparse input data will be converted to dense format.

nikicc:
I'm not sure why the tests fail but it doesn't seem to be related to changes introduced here.

ajdapretnar:
@nikicc Heat Map now works. However, I'm getting bizarrely strange results when connecting dense matrices to it, e.g. TopicModelling. Instances are no longer grouped by class and they're all mixed, meaning one rows somehow represents more than one instance (document). 😮

Orange/preprocess/transformation.py
```diff
@@ -107,7 +107,10 @@ def __init__(self, variable, offset, factor):
         self.factor = factor
 
     def transform(self, c):
-        return (c - self.offset) * self.factor
+        if sp.issparse(c):
+            return c
```
janezd: I don't understand this: why don't you normalize sparse values? In this case this transformation doesn't work as intended, does it?

Orange/data/table.py
```diff
@@ -264,23 +264,21 @@ def from_table(cls, domain, source, row_indices=...):
 
         global _conversion_cache
 
-        def sparse_to_flat(x):
-            if sp.issparse(x):
-                x = np.ravel(x.toarray())
-            return x
-
-        def get_columns(row_indices, src_cols, n_rows, dtype=np.float64):
+        def get_columns(row_indices, src_cols, n_rows, dtype=np.float64, is_sparse=False):
```
janezd: Not a huge deal, but: table.py mostly obey the 80-column limit. Can you break the longer lines in this PR?

Orange/widgets/visualize/owheatmap.py
```diff
@@ -462,6 +463,7 @@ class Information(widget.OWWidget.Information):
         discrete_ignored = Msg("{} discrete column{} ignored")
         row_clust = Msg("{}")
         col_clust = Msg("{}")
+        sparse_densified = Msg("Data was converted to dense format")
```
janezd: This information will not tell anything to a non-technical user. We should either find a way to explain it (e.g. a warning saying that: Showing this data in heat map may require a lot of memory) or, if we can't, just remove the info.

Orange/data/table.py
nikicc: I think I got them all under 80 characters now.

Orange/preprocess/transformation.py
nikicc: You're right, it doesn't 🙈 I was thinking about this and the only think I came up with was to normalise just non-zero values. Though, this is probably also not correct since non-present values are defined as zeros. Normalising everything would break the sparsity structure which is really undesirable for bag of words approaches.

So, should I just normalise non-zero values?

Orange/widgets/visualize/owheatmap.py
nikicc: I've shorten your version to simply: Showing this data may require a lot of memory.

Orange/preprocess/transformation.py
janezd: I don't like normalizing of non-zero values only, since you're putting them on a different scale than zero-values. What about just multiplying the (non-zero) values by `self.factor`?

This of course only makes sense when the factor is computed with offset=0. We can either accept that sometimes the transformation won't make sense or, perhaps better, raise an exception of `self.offset != 0`.

```
if sp.issparse(c):
    if self.offset != 0:
        raise ValueError("Non-zero offset in normalization of sparse data")
    return c * self.factor  # ... or whatever is the optimal way to multiply a sparse matrix with a scalar
```

nikicc: The above suggestion seems the right thing to do; however, it does not solve my problem —applying SVMs on sparse bag of words data — due to non-zero offsets.

Thinking a bit further, maybe my problem is not in _how_ we normalise, but _that_ we normalise. Should we even apply normalisation in this case? It only contains counts which in a sense already are in the same range (though this assumption breaks when there are also other, non BoW features present). Could we just turn normalisation off in this case? Though, I'm not sure how this could be elegantly achieved.

janezd: I'd then suggest that we normalize as I suggested above; the method must not silently fail, hence is must use the factor (because it can) and raise an error when offset is non-zero (because it can't do anything else).

The problem with the SVM is not directly related and can be solved in another PR.

nikicc: OK. You can check the current version and we can discuss later what to do with SVMs.

Orange/widgets/visualize/owheatmap.py
```diff
@@ -680,6 +682,11 @@ def set_dataset(self, data=None):
                 data_sample.download_data(2000, partial=True)
                 data = Table(data_sample)
 
+        if data is not None and sp.issparse(data.X):
+            data = data.copy()
+            data.X = data.X.toarray()
```
kernc: Catch potential MemoryError.

nikicc: Please check again.

Orange/widgets/visualize/owheatmap.py
```diff
@@ -462,9 +463,11 @@ class Information(widget.OWWidget.Information):
         discrete_ignored = Msg("{} discrete column{} ignored")
         row_clust = Msg("{}")
         col_clust = Msg("{}")
+        sparse_densified = Msg("Showing this data may require a lot of memory")
 
     class Error(widget.OWWidget.Error):
         no_continuous = Msg("No continuous feature columns")
+        no_enough_memory = Msg("Not enough memory to show this data")
```
janezd: Rename to `not_enough_memory`.

nikicc: Ups 🙈 Fixed.

PR #1584, opened 2016-09-21
[FIX] Load Classifier widget sends classifier on init

nikicc:
Filenames are stored in settings but when the new instance is created the most recent file was not loaded - it was only added to dropdown. The manual click on reload was necessary to load the classifier.


astaric:
@nikicc, @kernc: data processing should not be part of the widget's `__init__`, but rather scheduled for a later execution using a singleshot timer.

nikicc:
@astaric is #1586 better?

astaric:
yes, I have already merged it

PR #1587, opened 2016-09-22
OWPythonScript: Update locals when script is executed

VesnaT:
The widget didn't clear variables from previously executed scripts.

### DIFF
--- a/Orange/widgets/data/tests/test_owpythonscript.py
+++ b/Orange/widgets/data/tests/test_owpythonscript.py
@@ -1,0 +1,51 @@
+# Test methods with long descriptive names can omit docstrings
+# pylint: disable=missing-docstring
+from Orange.data import Table
+from Orange.classification import LogisticRegressionLearner
+from Orange.widgets.data.owpythonscript import OWPythonScript
+from Orange.widgets.tests.base import WidgetTest
+
+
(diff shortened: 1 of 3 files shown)


Orange/widgets/data/tests/test_owpythonscript.py
```diff
@@ -0,0 +1,123 @@
...
+        self.assertEqual(self.widget.in_classifier, None)
+        self.send_signal("in_classifier", self.model)
+        self.assertEqual(self.widget.in_classifier, self.model)
+
+    def test_in_classifier_disconnect(self):
+        """Check widget's classifier after disconnecting classifier on the input"""
+        self.send_signal("in_classifier", self.model)
+        self.assertEqual(self.widget.in_classifier, self.model)
+        self.send_signal("in_classifier", None)
+        self.assertEqual(self.widget.in_classifier, None)
```
kernc: `*_disconnect` tests seem to share at least two lines with their connect counterparts.

I'd personally prefer these tests more condensed, like:
```py
def test_inputs(self):
    for input in ('in_data', 'in_learner', ...):
        ... # full connect and disconnect scenario

def test_outputs(self):
    ...
```
The basic structure of all these tests looks compatible so there is no reason to quintuplicate.

PR #1589, opened 2016-09-23
[FIX] Variable: Fix Variable.copy for StringVariable and TimeVariable

ales-erjavec:
Due to static constructors in Variable.copy and ContinuousVariable.copy
the StringVariable.copy and TimeVariable.copy used to return Variable
and ContinuousVariable instances respectively.

Fix this by dispatching on `type(self)`.

Also add the missing entry for `vartype(Variable)` in
`gui.attributeIconDict`

### DIFF
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -427,7 +427,7 @@ class Variable(metaclass=VariableMeta):
         return make_variable, (self.__class__, self._compute_value, self.name), self.__dict__
 
     def copy(self, compute_value):
-        var = Variable(self.name, compute_value)
+        var = type(self)(self.name, compute_value=compute_value)
         var.attributes = dict(self.attributes)
         return var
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/gui.py
```diff
@@ -1169,6 +1169,8 @@ def __getitem__(self, key):
                                     "S", (0, 0, 0)),
                                    (vartype(TimeVariable()),
                                     "T", (68, 170, 255)),
+                                   (vartype(Variable),
+                                    "?", (128, 128, 128)),
```
kernc: But isn't `Variable` just a base class? Won't this open the door for certain errors to pass silently?

ales-erjavec: Yes it would. I removed it.

Orange/data/variable.py
```diff
@@ -899,6 +898,12 @@ def __init__(self, *args, **kwargs):
         self.have_date = 0
         self.have_time = 0
 
+    def copy(self, compute_value=None):
+        copy = super().copy(compute_value=compute_value)
+        copy.have_data = self.have_date
```
astaric: *have_date

ales-erjavec: Thanks. Fixed.

PR #1593, opened 2016-09-23
File widget: Add heart_disease to default data sets

janezd:


kernc:
Justification should be evident, I suppose?

janezd:
It's a data set with a mixture of discrete and continuous attributes, unlike those we had before. It's convenient for testing and for teaching/demonstration.

PR #1597, opened 2016-09-23
OWRank: Sort attributes by first visible score

VesnaT:
But only if header has not been clicked. In that case, attributes remains sorted the way user wanted.

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -68,7 +68,7 @@ class OWRank(OWWidget):
     auto_apply = Setting(True)
 
     # Header state for discrete/continuous/no_class scores
-    headerState = Setting((None, None, None))
+    headerState = Setting([None, None, None])
 
     settingsHandler = DomainContextHandler()
     selected_rows = ContextSetting([])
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owrank.py
```diff
@@ -193,15 +192,13 @@ def button(text, buttonid, toolTip=None):
 
         if self.headerState[0] is not None:
             self.discRanksView.horizontalHeader().restoreState(
-            self.headerState[0]
-        )
+            self.headerState[0])
```
kernc: Indent or inline.

Orange/widgets/data/owrank.py
```diff
@@ -671,6 +671,11 @@ def create_scores_table(self, labels):
         new_table.name = "Feature Scores"
         return new_table
 
+    @staticmethod
+    def sort_by_col(measures, def_sel):
+        cols = [i + 1 for i, m in enumerate(measures) if m.name in def_sel]
+        return cols[0] if cols else len(measures) + 1
```
kernc: What is `def_sel`? Passed in are `selected_checks`.
This method is named `sort_by_col()` but it does no sorting (as opposed to `self.ranksView.sortByColumn()`. I'd rename it to `select_col`.
The method is used in a single place and is short enough to bear moving it to where it's used.

Orange/widgets/data/owrank.py
```diff
@@ -610,6 +606,21 @@ def updateVisibleScoreColumns(self):
             self.ranksView.setColumnHidden(i + 1, not shown)
             self.ranksView.setColumnWidth(i + 1, 100)
 
+        if self.headerState[self.rankMode] is not None:
```
lanzagar: This if condition seems to be unnecessary and if removed would make the code (together with the next if) a bit more readable.

PR #1603, opened 2016-09-23
[ENH] Metadata for data files

janezd:
@astaric, you meant something like this?

screen shot 2016-09-24 at 00 51 54

Orange already read and wrote .metadata files, but in pickle format. This is highly inconvenient since the user can't edit it. I changed the format to a (very) simplified YAML: the file is expected to contain a single block mapping. Data is read as string. If we wish, we can either try to evaluate it (with fallback to string) or use PyYAML, although I don't think we need it.

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -283,10 +283,15 @@ class FileFormat(metaclass=FileFormatMeta):
 
     @classmethod
     def write_table_metadata(cls, filename, data):
-        if isinstance(filename, str) and getattr(data, 'attributes', {}):
-            with open(filename + '.metadata', 'w') as f:
-                f.write("\n".join(
-                        "{}: {}".format(*kv) for kv in data.attributes.items()))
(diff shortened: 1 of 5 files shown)


astaric:
@BlazZupan, @markotoplak does this change break your use of metadata?

kernc:
It does break reading any existing metadata files.

janezd:
I had it implemented (the other way around, I first checked whether it's a pickle). Then I "discovered" (don't ask me how) that nobody uses .metadata and decided to switch to a "better" format while it's still time.

I re-added it. For writing, pickle is used if there already exists a metadata file that is a pickle, or if the current attributes are not an `OrderedDict` with strings as keys and values.

This is imperfect, but I think that having a plain text format is better than a binary - when possible. I wouldn't like to have two metadata files. Any better ideas?

janezd:
Fixed as discussed:

- when loading, it tries to unpickle, then to read as "YAML"
- when saving, it uses "YAML" if the attributes are a dict with string keys and values

PR #1608, opened 2016-09-25
Implement DomainModel

janezd:
This PR does two things.

It introduces `DomainModel`, which is primarily intended for combo boxes. The model is tested on the Scatter plot widget. The model supports arbitrary arrangements of attributes, classes, metas, other strings and separators.

Second, it cleans up the `ScaleData` class. The class had to be modified to use `Variable` instead of `str`. The class contained a lot of dead and rustic code, so I removed it instead of adapting it. The only other widget that uses it is Parallel Coordinates. I fixed it, except for the bugs that were, I guess, already there before.

The most interesting files in the PR are itemmodels.py, which contains the model, and scatterplot.py (particularly method `init_attr_values`), which shows how the model simplifies the code while making the combo boxes nicer.


kernc:
This PR has one commit, yet touches 12 files and introduces all sorts of changes?

janezd:
Unfortunately yes. If I split it in two, one commit would include all the changes in gui.py and itemmodels.py, and the other would change everything else since the changes are interdependent.

If it helps, you can imagine there are two commits and observe the former two files separately from others. :)

kernc:
It would be best if commits were structured:

1. itemmodels.DomainModel + tests
2. widgets using DomainModel
2. unrelated refactoring

Orange/widgets/utils/itemmodels.py
```diff
@@ -646,6 +647,60 @@ def python_variable_tooltip(self, var):
...
+
+    def set_domain(self, domain):
+        self.beginResetModel()
+        content = []
+        # The logic related to separators is a bit complicated: it ensures that
+        # even when a section is empty we don't have two separators in a row
+        # or a separator at the end
+        add_separator = False
+        for section in self.order:
+            if section is self.Separator:
```
janezd: I thought about that, too. But then you'd get a separator between `("(None)", ATTRIBUTES)`.

One possible rule would be that a separator automatically put between the "categories", so `("(None)", ATTRIBUTES, CLASSSES)` would have a separator between the latter but not between former two. Yet this rule is rather complicated, so I'd vote for explicit separators.

janezd:
Agreed, but (a) it wouldn't be much better, (b) I'm not disciplined enough for that (when I'm modifying a function I'm refactoring it as well; and you never know how far will some innocent refactoring go) and (c) it's too late. I'll try harder next time.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -179,22 +182,36 @@ def __init__(self):
         self.sampling.setVisible(False)
 
         box = gui.vBox(self.controlArea, "Points")
+        color_model = DomainModel(
+            order=("(Same color)", dmod.Separator) + dmod.SEPARATED,
+            valid_types=dmod.PRIMITIVE)
         self.cb_attr_color = gui.comboBox(
             box, self, "graph.attr_color", label="Color:",
             emptyString="(Same color)", callback=self.update_colors,
```
kernc: Is there any way this `emptyString` wouldn't need to be redefined? It's defined in the model, that should be enough and should work then with any users of DomainModel, not just combos.

Like:
```py
model = DomainModel(..., default='(Same color)')  # default is always first
# and also
assert model.default == '(Same color)'
```
and the gui "front" handlers can check:
```py
...
value = model.data(index)
return value if value is not model.default else None

**Comment by kernc** [comment]
Re
(a) I disagree,
(b) Sure, you refactor the touched function. But the changes in _datacaching.py_ and _scaling.py_, in particular, seem a bit out-of-place.
(c) It's never too late: :smiley: 
```bash
git co domain-model
git reset upstream/master

# Repeat until no changes left
git add --patch
git commit
```

Orange/widgets/utils/itemmodels.py
```diff
@@ -646,6 +647,60 @@ def python_variable_tooltip(self, var):
         text += self.variable_labels_tooltip(var)
         return text
 
+
+class DomainModel(VariableListModel):
```
kernc: In either case, please add some unit tests.

Orange/widgets/utils/itemmodels.py
kernc: > I thought about that, too. But then you'd get a separator between ("(None)", ATTRIBUTES).

Indeed, that's what you coded in manually for scatter plot.

kernc: > > But then you'd get a separator between ("(None)", ATTRIBUTES).

> Indeed, that's what you coded in manually for scatter plot.

Which is totally fine by me. A separator adds a 12px line that is easily noticeable and conveys information. Cost-wise, with 20 attributes, the 12px separator line is negligible. With 20.000, just as so.

Orange/widgets/visualize/owscatterplot.py
janezd: I hated this repetition, too.

In my first implementation, the default was the first string value in `order`. It had a similar effect as what you propose, but does not introduce yet another argument, and it also works if you want to put the default at the end. But I didn't do it since perhaps sometimes there is no default, and the first value is just a string.

I don't like your solution any better because it automagically adds something to `order`. Also, I have no way to decide whether I'd like to have it separated or not.

I currently see no better solution than `emptyString`, so for now I'm keeping what we had before.

Orange/widgets/utils/itemmodels.py
janezd: I did it that way in the scatter plot because the rest is also separated. If I had just `"(None)", ATTRIBUTES)`, I wouldn't separate it.

It's not about space, it's about aesthetics. And freedom. :)

Orange/widgets/utils/itemmodels.py
janezd: > In either case, please add unit tests.

OK, I'll squash them in. :)

Orange/widgets/visualize/owscatterplot.py
kernc: When is the first value "just a string"? This is DomainModel — It's about Variables of the Domain. For arbitrary contents, PyListModel applies. Why would we not want a uniform UX throughout Orange? Why would you put the default option last? Why wouldn't you separate the default from the rest?

You hate this repetition too? Invent an argument for it.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -179,22 +182,32 @@ def __init__(self):
...
+        label_model = DomainModel(
+            placeholder="(No labels)", valid_types=dmod.PRIMITIVE)
         self.cb_attr_label = gui.comboBox(
             box, self, "graph.attr_label", label="Label:",
-            emptyString="(No labels)", callback=self.graph.update_labels,
-            **common_options)
+            callback=self.graph.update_labels,
+            model=label_model, **common_options)
+        shape_model = DomainModel(
+            placeholder=("Same shape"), valid_types=DiscreteVariable)
```
kernc: Parentheses inside.

janezd: LOL. Done.

Orange/widgets/utils/itemmodels.py
janezd: Tests added and also put in separate commit las week (4c8a628800641ec6a89e315549e2d30b74233870).

kernc:
With exception of one overly vague commit message ("Placeholders for Nones"), this is one good-looking PR. :+1:

PR #1612, opened 2016-09-28
CrossValidation: use forkserver multiprocessing also on Linux

kernc:


astaric:
@markotoplak, can you check? I guess we want this to make it to the 3.3.8 release.

PR #1614, opened 2016-09-29
[FIX] gui: Add a push button class adapted for variable width text

ales-erjavec:
Use it to fix `auto_commit` generated layout.

### DIFF
--- a/Orange/widgets/utils/buttons.py
+++ b/Orange/widgets/utils/buttons.py
@@ -1,0 +1,61 @@
+from PyQt4.QtGui import QPushButton, QStyle, QStyleOptionButton
+from PyQt4.QtCore import Qt, QSize
+
+
+class VariableTextPushButton(QPushButton):
+    """
+    QPushButton subclass with an sizeHint method to better support settable
+    variable width button text.
+
+    Use this class instead of the QPushButton when the button will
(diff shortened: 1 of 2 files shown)


Orange/widgets/utils/buttons.py
```diff
@@ -0,0 +1,61 @@
+from PyQt4.QtGui import QPushButton, QStyle, QStyleOptionButton
+from PyQt4.QtCore import Qt, QSize
+
+
+class VariableTextPushButton(QPushButton):
+    """
+    QPushButton subclass with an sizeHint method to better support settable
+    variable width button text.
+
+    Use this of class instead of the QPushButton when the button will
```
lanzagar: Use this _-of_ class

lanzagar:
"fix auto_commit generated layout" - what is the best way to see the difference? / where were previous problems noticeable?

ales-erjavec:
> what is the best way to see the difference? / where were previous problems noticeable?

When the  auto commit checkbox is toggled the layout would change. Most noticeable on widgets with fixed size constraint like most learners.

lanzagar:
OK, apparently that was an issue on OSX (looks annoying).
Was already fine on my linux machine, and it remains the same after this fix.
So, :+1: , merging.

PR #1615, opened 2016-09-30
DistanceMatrix: Fix context settings, add tests

astaric:


Orange/widgets/settings.py
```diff
@@ -635,9 +635,23 @@ def open_context(self, widget, *args):
         widget.current_context, is_new = \
             self.find_or_create_context(widget, *args)
         if is_new:
-            self.settings_from_widget(widget, *args)
+            if self.settings_from_widget.__code__.co_argcount == 2:
```
janezd: `f(a, b, *args)` still has `co_argcount==2`. I don't know why this doesn't raise warnings though - do we mute them somewhere?

The proper check is `if not self.settings_from_widget.__code__.co_flags & 4:`.

PR #1619, opened 2016-09-30
Subset selection in Scatter Plot now made bolder.

ajdapretnar:


Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -808,7 +808,9 @@ def make_pen(color, width):
                 self.pen_colors = pens[c_data]
                 self.brush_colors = np.array([
                     [QBrush(QColor(0, 0, 0, 0)),
-                     QBrush(QColor(col[0], col[1], col[2], self.alpha_value))]
+                     QBrush(QColor(col[0], col[1], col[2],
+                                   255 if subset is not None
```
janezd: Can you store `255 if subset is not None else self.alpha_value` into a variable that is computed before this line, so the condition doesn't need to be checked for each data instance separately?

PR #1622, opened 2016-10-02
[FIX] Scatterplot: Reintroduce sliders for size and opacity

### LINKED ISSUE #1621: Symbol size and Opacity have disappeared from Scatter Plot
##### Orange version
3.3.8 candidate

##### Expected behavior
Scatter Plot allowed to set Symbol size and Opacity in Points box (see [screenshots](http://orange.biolab.si/screenshots/).

##### Actual behavior
In 3.3.8 candidate Symbol size and Opacity are missing. They are very useful when teaching with Orange or working with data sets of various number of instances.

janezd:
Fixes #1621.

@ajdapretnar also complained about the missing sliders, but I told her that it was intentional. We want to simplify the GUI's by selecting optimal colors/sizes/shapes... ourselves instead of having dozens of options.

I remember discussing this, but I'm no longer sure that we reached a conclusion and that the change was indeed intentional: the commit in which this was removed, 71ccfcdd3de7dc0bdc0170dd752800cdf806ab99, is named "Lint" and doesn't make any other non-lint-related changes.

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -197,6 +197,7 @@ class OWScatterPlot(OWWidget):
             **common_options)
 
         g = self.graph.gui
+        g.point_properties_box(self.controlArea, box)
         box = gui.vBox(self.controlArea, "Plot Properties")
         g.add_widgets([g.ShowLegend, g.ShowGridLines], box)
         gui.checkBox(


janezd:
You think so? I guess that opacity serves (mostly) aesthetics, while when showing a subset you'd prefer the subset being as visible as possible. I'd keep #1619.

ajdapretnar:
Well, it leads to confusion since the slider is still on and it changes nothing. Then perhaps block the slider on Selected Data input? But I think what @BlazZupan had in mind was changing the slider exactly when the subset is highlighted. Waiting for his input.

astaric:
Required changes can be addressed in a subsequent pull request.

PR #1623, opened 2016-10-03
[FIX] win-installer: Build scikit-learn in windows installer build script

ales-erjavec:
scikit-learn no longer provides wheels for Python 3.4 on windows for the latest 0.18 release.


astaric:
Thanks

PR #1624, opened 2016-10-03
[ENH] Manifold Learning

VesnaT:


Orange/widgets/unsupervised/owmanifoldlearning.py
```diff
@@ -0,0 +1,284 @@
...
+    max_iter = Setting(300)
+    random_state = Setting(0)
+    random_state_values = (None, 0)
+
+    def __init__(self, parent):
+        super().__init__(parent)
+        self.max_iter_spin = self._create_spin_parameter(
+            "max_iter", 10, 10 ** 4, "Max iterations:")
+        self.random_state_check = self._create_check_parameter(
+            "random_state", "Replicable (deterministic) learning")
```
lanzagar: "Replicable (deterministic) learning" is a bit awkward and too long - does not fit into fixed widget size for me.
Suggestion: invert the option and use "Random initialization" checkbox.
Bonus question: does replicable learning use PCA initialization? because the results look different than in the MDS widget...

Orange/widgets/unsupervised/owmanifoldlearning.py
```diff
@@ -0,0 +1,284 @@
...
+    class Error(OWWidget.Error):
+        n_neighbors_too_small = Msg("Neighbors must be greater than {}.")
+        negative_distances = Msg("All distances should be positive.")
+        manifold_error = Msg("{}")
+
+    def __init__(self):
+        self.data = None
+
+        # GUI
+        info_box = gui.vBox(self.controlArea, "Info")
```
lanzagar: While the Info box at the top isn't inappropriate it is also not necessary IMHO.
And comparing this widget with similar ones: PCA, MDS, CA, k-means, hierarchical, or even any classification/regression widget - none of them have an Info box.

BlazZupan:
For t-SNE, some metric are redundant and some are hard to understand from default names given in sklearn. My proposal is:

Euclidean - remove (sklearn uses Euclidean squared)
L1 - remove (same as Manhattan)
L2 - keep, but rename to Euclidean
Manhattan - keep
Cosine - keep, but check what is wrong
Chebyshev - keep
Jaccard - keep
Mahalanobis - keep
Minkowski - remove (would need a parameter p, if it stays, but we actually do not need it here)
Seuclidean - remove (needs parameter)
Sqeuclidean - remove

With t-SNE and cosine metric, I get a complaint "All distances should be positive." The metric is 1-cos(fi), so by definition this should always yield positive. But due to numerical error, cos(fi) on iris indeed returns a negative number (-1e-16). Quick fix for this pull request: do not include cosine (@lanzagar will report an issue to sklearn).

lanzagar:
Fix for TSNE with cosine distances was merged in scikit-learn 2 days ago so it should work in the next version.
https://github.com/scikit-learn/scikit-learn/pull/7732

PR #1625, opened 2016-10-03
[FIX] errorreporting: Fix an KeyError for a missing 'Widget Module' entry

ales-erjavec:


janezd:
@ales-erjavec, is this the bug that triggered endless recursion? (@kernc)

kernc:
Not entirely sure what triggers the recursion since the excepthook is patched to prevent reentry. But the code was bug and this fixes it. :smile:

PR #1628, opened 2016-10-04
Refactor SqlTable

astaric:
Sql code is currently coupled with SqlTable, making support for other sql backends hard to implement. This PR moves (some of the) code that is dependent on sql driver to a separate class (Psycopg2Backend) which could (in theory) be replaces by another implementation of the database backend.

### DIFF
--- a/Orange/data/sql/table.py
+++ b/Orange/data/sql/table.py
@@ -5,6 +5,7 @@ import functools
 import logging
 import re
 import threading
+import warnings
 
 from contextlib import contextmanager
 from itertools import islice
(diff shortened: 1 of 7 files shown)


astaric:
@lanzagar I have added some tests for the sql calls that were not covered before and added docstrings for the public Backend interface. Please have another look.

Orange/widgets/data/owsql.py
```diff
@@ -309,10 +309,10 @@ def get_table(self):
             QApplication.setOverrideCursor(QCursor(Qt.WaitCursor))
             if sample:
                 s = table.sample_time(1)
-                domain = s.get_domain(guess_values=True)
+                domain = s.create_domain(inspect_values=True)
```
lanzagar: The new create domain has a required positional argument

Orange/tests/sql/test_misc.py
```diff
@@ -0,0 +1,28 @@
+"""Test for miscelaneous sql queries in widgets
+
+Please not that such use is deprecated.
```
lanzagar: note

Orange/data/sql/table.py
```diff
@@ -75,53 +289,28 @@ def __init__(
         """
         if isinstance(connection_params, str):
             connection_params = dict(database=connection_params)
-        self.connection_params = connection_params
```
lanzagar: was used in (at least) owdatainfo.py, L142

PR #1632, opened 2016-10-06
[FIX] OWFeatureConstructor: Fix domain check for only meta data sets

nikicc:
This is ugly, but since domain does not iterate over metas, I do not know of a better solution.

@ajdapretnar this should fix your feature constructor issues with twitter data set.


kernc:
Anything that is ugly / hackish / potentially non-evident warrants a short comment.

PR #1634, opened 2016-10-06
Focus adjustments

janezd:
Pressing Enter in line edit should remove the focus. This signals that the data has been "entered". At least on Os X, keeping the focus looks as if the field is still being edited.

Second, when the widget open, the first field shouldn't have focus - especially if this is line edit. Editing should be triggered by the user.

For an example, open the Paint widget. The first line edit has focus and pressing Enter does not release it.

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -787,6 +787,7 @@ class LineEditWFocusOut(QtGui.QLineEdit):
         self.__changed = True
 
     def returnPressedHandler(self):
+        self.clearFocus()
         if self.__changed:
             self.__changed = False
             if hasattr(self, "cback") and self.cback:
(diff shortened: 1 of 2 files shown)


Orange/widgets/widget.py
```diff
@@ -205,6 +205,8 @@ def __new__(cls, *args, captionTitle=None, **kwargs):
 
         sc = QShortcut(QKeySequence.Copy, self)
         sc.activated.connect(self.copy_to_clipboard)
+        if self.controlArea is not None:
+            self.controlArea.setFocus()  # Otherwise, the first control has focus
```
kernc: Can this be `setFocus(Qt.ActiveWindowFocusReason)`?

Orange/widgets/gui.py
```diff
@@ -787,6 +787,7 @@ def __textEdited(self):
         self.__changed = True
 
     def returnPressedHandler(self):
+        self.clearFocus()
```
janezd: No, this is usually not how you edit settings. If you change a value in one line edit, it doesn't mean you're going to continue with the next.

Orange/widgets/widget.py
janezd: OK, I added `Qt.ActiveWindowFocusReason`. It probably doesn't change anything, though.

PR #1635, opened 2016-10-06
[FIX] OWFile: Report errors for incorrect file formats instead of crashing

janezd:
If the file was in incorrect format, the exception was sometimes thrown already by `FileFormat.get_reader`. To see this, rename an arbitrary file to .xlsx and try to load it.

The code would be simpler with a single try-except. However, since it catches any exception, I prefer wrapping a single call, hence two separate try-except blocks.

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -257,29 +257,36 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
 
     # Open a file, create data from it and send it over the data channel
     def load_data(self):
+        # We need to catch any exception type since anything can happen in
+        # file readers
+        # pylint: disable=broad-except
         self.editor_model.set_domain(None)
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owfile.py
```diff
@@ -257,29 +257,36 @@ def browse_file(self, in_demos=False):
 
     # Open a file, create data from it and send it over the data channel
     def load_data(self):
+        # We need to catch any exception type since anything can happen in
+        # file readers
+        # pylint: disable=broad-except
         self.editor_model.set_domain(None)
-        self.reader = self._get_reader()
-        if not self.reader:
-            self.data = None
```
lanzagar: `self.data` used to be cleared in case of errors, now it is not anymore...

This is used in reports for example (maybe elsewhere too), which now report on previous data, even after trying to open something else and failing.

PR #1637, opened 2016-10-07
OWFile: Set self.data to None on error

janezd:
As noted by @lanzagar, #1635 no longer sets `self.data` to `None` in case or error. Fixed.


janezd:
You're right. I added the missing `self.data = None`.

PR #1639, opened 2016-10-07
OWPythonScript: Don't trigger reporting on errors

nikicc:

### DIFF
--- a/Orange/widgets/data/owpythonscript.py
+++ b/Orange/widgets/data/owpythonscript.py
@@ -4,6 +4,7 @@ import code
 import keyword
 import itertools
 import unicodedata
+from unittest.mock import patch
 
 from PyQt4 import QtGui, QtCore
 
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owpythonscript.py
```diff
@@ -206,7 +207,8 @@ def push(self, line):
         saved = sys.stdout, sys.stderr
         try:
             sys.stdout, sys.stderr = self, self
-            return code.InteractiveConsole.push(self, line)
+            with patch('sys.excepthook', sys.__excepthook__):   # prevent console errors to trigger error reporting
```
lanzagar: Maybe you could put the comment on its own line?

nikicc: Fixed. Also stdout & stderr was patched the same way as the hook.

PR #1645, opened 2016-10-07
[FIX] OWScatterPlot: Fix progress bar percentages

### LINKED ISSUE #1536: progress bar shows percentage in thousands
##### Orange version
3.3.7

##### Expected bahavior
progress should be between 0 and 100

##### Actual behavior
progress in the window bar of Score Plots goes to thousands, say to 12300%

Open fresh Orange, in File read the data from http://bit.ly/1EEB4FX, connect Scatter Plot, click Score Plots and then Start. It is interesting that the wrong scoring appears only the first time this function is used (pressing Pause and then Continue would display the correct counting).

tomazc:
Fixes #1536


kernc:
Is failing.

PR #1651, opened 2016-10-09
Make TreeAdapters for Pythagorean trees more general

janezd:
TreeAdapters are great and potentially useful elsewhere, too.

When implementing another adapter, I guess I found two (cosmetic) slips:

- `get_instances_in_nodes` accepts instances of `TreeNode`, which is specific Pythagorean trees,
- `ROOT_PARENT` is defined as `-1`, which is specific to SKL trees. A value of `None` would be more pythonic. Besides, the code in pythagorasviewer.py compares the node with `-1` instead of with `ROOT_PARENT`. Finally, the constant is defined in the base `TreeAdapter` as well as in the derived `SklTreeAdapter`.

@pavlin-policar, can you take a look at these two cosmetic changes? Did I overlook any place that needs to be changed w.r.t this?

### DIFF
--- a/Orange/widgets/visualize/owpythagorastree.py
+++ b/Orange/widgets/visualize/owpythagorastree.py
@@ -350,7 +350,7 @@ class OWPythagorasTree(OWWidget):
                        self.scene.selectedItems())
 
         data = self.tree_adapter.get_instances_in_nodes(
-            self.clf_dataset, [item.tree_node for item in items])
+            self.clf_dataset, [item.tree_node.label for item in items])
         self.send('Selected Data', data)
 
     def send_report(self):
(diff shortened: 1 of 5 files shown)


pavlin-policar:
This is great, but this breaks the regression tree tooltips (which were apparently wrong and overlooked in the first place).

Changing the regression tree tooltip to simply show the number of samples in the node is probably best and can be done easily by changing the regression tooltip function (L#550-L#553) from
```python
    def _regression_get_tooltip(self, node):
        total = self.tree_adapter.num_samples(
            self.tree_adapter.parent(node.label))
        samples = self.tree_adapter.num_samples(node.label)
        ratio = samples / total
```
to
```python
        num_samples = self.tree_adapter.num_samples(node.label)
```

and lines L#571-L#572 from
```python
            + '<br>{}/{} samples ({:2.3f}%)'.format(
              int(samples), total, ratio * 100) \
```
to
```python
            + '<br>{} samples'.format(num_samples) \
```

It previously tried to show the ratio between what number of samples belonged to the node and the number of samples in the parent node, which is inconsistent with the classification tooltips, which show the ratio of the majority class sample and all the samples in the node.

It would probably be better to keep things consistent so that the tooltips only show data for the given node and not care about its parent.

janezd:
Apologies for letting you wait; I had 11 hours of lecturing this week.

I've spotted the problem with regression trees, but attributed it to some other changes I've made in another PR I'm preparing. I've merged #1660. You can now merge this one if you think it's OK.

pavlin-policar:
No problem. It should be okay to merge now. I've looked it over again and found no apparent problems.

PR #1655, opened 2016-10-12
[WIP][ENH] Scatterplot, HeatMap, TreeGraph, ConfusionMatrix and Unsupervised widgets: Output Flagged Data 

VesnaT:

### DIFF
--- a/Orange/tree.py
+++ b/Orange/tree.py
@@ -224,9 +224,14 @@ class TreeModel(Model):
         return _count(self.root)
 
     def get_instances(self, nodes):
+        indices = self.get_indices(nodes)
+        if indices is not None:
+            return self.instances[indices]
+
+    def get_indices(self, nodes):
         subsets = [node.subset for node in nodes]
         if subsets:
-            return self.instances[np.unique(np.hstack(subsets))]
+            return np.unique(np.hstack(subsets))
 
     @staticmethod
     def climb(node):
--- a/Orange/widgets/evaluate/tests/test_owconfusionmatrix.py
+++ b/Orange/widgets/evaluate/tests/test_owconfusionmatrix.py
@@ -4,22 +4,28 @@ from Orange.data import Table
 from Orange.classification import NaiveBayesLearner, TreeLearner
 from Orange.evaluation.testing import CrossValidation
 from Orange.widgets.evaluate.owconfusionmatrix import OWConfusionMatrix
-from Orange.widgets.tests.base import WidgetTest
+from Orange.widgets.tests.base import WidgetTest, WidgetOutputsTestMixin
 
 
-class TestOWConfusionMatrix(WidgetTest):
+class TestOWConfusionMatrix(WidgetTest, WidgetOutputsTestMixin):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        WidgetOutputsTestMixin.init(cls)
+
         bayes = NaiveBayesLearner()
         tree = TreeLearner()
-        iris = Table("iris")
+        iris = cls.data
         titanic = Table("titanic")
         common = dict(k=3, store_data=True)
         cls.results_1_iris = CrossValidation(iris, [bayes], **common)
         cls.results_2_iris = CrossValidation(iris, [bayes, tree], **common)
         cls.results_2_titanic = CrossValidation(titanic, [bayes, tree],
                                                 **common)
+
+        cls.signal_name = "Evaluation Results"
+        cls.signal_data = cls.results_1_iris
+        cls.same_input_output_domain = False
 
     def setUp(self):
         self.widget = self.create_widget(OWConfusionMatrix,
@@ -40,19 +46,11 @@ class TestOWConfusionMatrix(WidgetTest):
         self.send_signal("Evaluation Results", self.results_1_iris)
         self.widget.selected_learner[:] = [0]
 
-    def test_outputs(self):
-        self.send_signal("Evaluation Results", self.results_1_iris)
-
-        # check selected data output
-        self.assertIsNone(self.get_output("Selected Data"))
-
-        # select data instances
+    def _select_data(self):
         self.widget.select_correct()
-
-        # check selected data output
-        selected = self.get_output("Selected Data")
-        self.assertGreater(len(selected), 0)
-
-        # check output when data is removed
-        self.send_signal("Evaluation Results", None)
-        self.assertIsNone(self.get_output("Selected Data"))
+        indices = self.widget.tableview.selectedIndexes()
+        indices = {(ind.row() - 2, ind.column() - 2) for ind in indices}
+        selected = [i for i, t in enumerate(zip(
+            self.widget.results.actual, self.widget.results.predicted[0]))
+                    if t in indices]
+        self.selected_indices = self.widget.results.row_indices[selected]
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -29,6 +29,8 @@ from Orange.clustering.hierarchical import \
 
 from Orange.widgets import widget, gui, settings
 from Orange.widgets.utils import colorpalette, itemmodels
+from Orange.widgets.utils.annotated_data import (create_annotated_table,
+                                                 ANNOTATED_DATA_SIGNAL_NAME)
 from Orange.widgets.io import FileFormat
 
 __all__ = ["OWHierarchicalClustering"]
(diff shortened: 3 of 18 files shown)


janezd:
I haven't yet checked the content of the PR (obviously), but before we merge it we should perhaps talk about the signal name again. I don't think that the verb "to flag" appears in the dictionary with that meaning. Even the use of word "flag" for some kind of marker is IMHO limited to computer science, so the name "flagged data" wouldn't mean anything to an outsider.

What is wrong with "marked"?

Vesna, sorry if this will require some (hopefully trivial) refactoring.

VesnaT:
No worries.. I've intentionally made this PS 'small' (only four widgets are modified).

Orange/misc/flagged_data.py
```diff
@@ -0,0 +1,44 @@
...
+    and is generated using name parameter, appended by smallest possible index.
+
+    :param names: list
+    :param name: str
+    :return: str
+    """
+    if name not in names:
+        return name
+    counts = [int(re.match(r"(" + name + " )(\d{1,}$)", n).group(2))
+              for n in names if re.match(r"(" + name + " )(\d{1,}$)", n)] + [1]
```
janezd: Something like this:
```
counts = max((int(mo.group(2))
              for mo in re.finditer(r"(" + name + " )(\d{1,}$)", n)),
             default=0)
```

janezd: Matter of taste: I'd write `r"({})(\d{{1,}}$".format(name)`

Orange/tree.py
```diff
@@ -228,6 +228,11 @@ def get_instances(self, nodes):
         if subsets:
             return self.instances[np.unique(np.hstack(subsets))]
 
+    def get_indices(self, nodes):
```
janezd: `get_instances` could call this function. Don't forget to handle the case when `get_indices` returns `None`.

Orange/tests/test_misc.py
```diff
@@ -28,3 +33,60 @@ def test_memoize_method(self):
...
+
+    def test_create_flagged_table_none_data(self):
+        self.assertIsNone(create_flagged_table(None, None))
+
+    def test_create_flagged_table_none_indices(self):
+        flagged = create_flagged_table(self.zoo, None)
+        self.assertEqual(len(flagged), len(self.zoo))
+        self.assertEqual(0, np.sum([i[FLAGGED_FEATURE_NAME] for i in flagged]))
+
+    def test_cascade_flagged_tables(self):
```
janezd: I could be too smart by half and replace your function that uses regular expressions with one that uses the first unoccupied name. This test wouldn't fail because there are no "holes". Can you add some code to this test to remove the second meta and then "flag" the table again, so my smart idea would fail?

Orange/misc/flagged_data.py
```diff
@@ -0,0 +1,44 @@
+import re
+import numpy as np
+from Orange.data import Table, Domain, DiscreteVariable
+
+FLAGGED_SIGNAL_NAME = "Flagged Data"
```
janezd: Thinking about it again, I started liking the constant because it will indeed make us stick to the same name in all widgets.

The name is not perfect, though. Not only will we change "flagged" to something else, but it also suggests that it is a name of a flagged signal. `ANNOTATED_DATA_SIGNAL_NAME = "Data"` is better (than `ANNOTATED_SIGNAL_NAME`) but awfully long. Think about it...

janezd: If nothing else, the module belongs within `Orange.widgets` because it includes a name of the signal, hence it **is** related to widgets. :)

Orange/widgets/unsupervised/tests/test_owdistancemap.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # check selected data output
+        self.assertIsNone(self.get_output("Data"))
+
+        # check flagged data output
+        flagged = self.get_output(FLAGGED_SIGNAL_NAME)
+        self.assertEqual(0, np.sum([i[FLAGGED_FEATURE_NAME] for i in flagged]))
+
+        # select data points
+        points = random.sample(range(0, len(self.iris)), 20)
```
janezd: I prefer deterministic tests. Randomly choose some indices instead of choosing some indices at random. (https://xkcd.com/221/).

Orange/widgets/unsupervised/tests/test_owdistancemap.py
```diff
@@ -0,0 +1,50 @@
...
+        self.widget._selection = points
+        self.widget.commit()
+
+        # check selected data output
+        selected = self.get_output("Data")
+        self.assertEqual(len(selected), len(points))
```
janezd: What about testing that the correct instances were chosen? There may be a better way to do it, but I somewhere used something like `np.testing.assert_almost_equal(selected.X, self.iris.X[points])`.

Orange/widgets/unsupervised/tests/test_owmds.py
```diff
@@ -0,0 +1,53 @@
...
+
+        # check selected data output
+        self.assertIsNone(self.get_output("Selected Data"))
+
+        # check flagged data output
+        flagged = self.get_output(FLAGGED_SIGNAL_NAME)
+        self.assertEqual(0, np.sum([i[FLAGGED_FEATURE_NAME] for i in flagged]))
+
+        # select data points
+        points = random.sample(range(0, len(self.iris)), 20)
```
janezd: Same as in `DistanceMap`.

janezd:
Github is showing some of my comments as outdated, although you haven't made any further commits. Please check those, too.

Apart from these minor suggestions, I like the PR, in particular your factoring out of the parts of the tests. It would be even greater if you could simulate, say, selection action, but I know this is probably too hard.

Please tell me/us when you make the changes, so we don't wait too long with merging, since rebasing dozens of widgets is not that much fun.

VesnaT:
The comments are outdated because the code was moved to a Mixin in one of the following commits (441b15a), since it was very similar for all widgets. I could have rebased, but wanted to keep the code in case someone didn't like the Mixin idea.
I thought I did simulate the selection...

VesnaT:
Done.

Orange/widgets/utils/annotated_data.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    if data is None:
+        return None
+    names = [var.name for var in data.domain.variables + data.domain.metas]
+    name = _get_next_name(names, ANNOTATED_DATA_FEATURE_NAME)
+    metas = data.domain.metas + (DiscreteVariable(name, ("No", "Yes")),)
+    domain = Domain(data.domain.attributes, data.domain.class_vars, metas)
+    annotated = np.zeros((len(data), 1))
+    if selected_indices is not None:
+        annotated[selected_indices] = 1
```
janezd: Should this be 0 or 1? If nothing is selected, all instances have to have Selected=No, no?

Orange/widgets/evaluate/tests/test_owconfusionmatrix.py
```diff
@@ -40,19 +46,11 @@ def test_selected_learner(self):
...
-
-        # check output when data is removed
-        self.send_signal("Evaluation Results", None)
-        self.assertIsNone(self.get_output("Selected Data"))
+        indices = self.widget.tableview.selectedIndexes()
+        indices = {(ind.row() - 2, ind.column() - 2) for ind in indices}
+        selected = [i for i, t in enumerate(zip(
+            self.widget.results.actual, self.widget.results.predicted[0]))
+                    if t in indices]
+        self.selected_indices = self.widget.results.row_indices[selected]
```
janezd: Would it be nicer if `_select_data` returned `selected_indices` instead of (ab?)using the instance's attributes for semi-global data storage?

Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -1092,10 +1099,12 @@ def commit(self):
 
         if not selected_indices:
             self.send("Selected Data", None)
-            self.send("Other Data", None)
+            annotated_data = create_annotated_table(items, selected_indices) \
```
janezd: Not really an issue, but since you're going to make another commit anyway, can you replace `selected_indices` with `[]`, so that it's obvious this call will always select all (or no) instances.

Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -1135,22 +1144,33 @@ def commit(self):
 
             if selected_indices:
                 selected_data = data[mask]
-            if unselected_indices:
-                unselected_data = data[~mask]
+                if self.append_clusters:
+                    def remove_other_value(vars_):
+                        vars_ = [var for var in vars_]
```
janezd: Why not `copy`?

Orange/widgets/utils/annotated_data.py
janezd: > Should this be 0 or 1? If nothing is selected, all instances have to have Selected=No, no?

I'm stupid. Please ignore.

VesnaT:
I will fix the suggested in the next PR.

PR #1657, opened 2016-10-12
Merge learner widgets

pavlin-policar:
A much simpler approach to handling data with continuous or discrete target variables inside learner widgets.

The `BaseLearner` now defines a `problem_type` context setting that tells us whether the data represents a classification or regression problem. This setting can be manually set with radio buttons if there is no data (this is required to work properly with the Test&Score widget). If there is data on the input, we disable the radio buttons and assign the appropriate type. The radio buttons are completely opt-in and can be enabled with a flag on the widget, so widgets that only handle one kind of target variable don't have to deal with any extra logic.

Each individual learner can then override the `on_type_change` method, where they can do whatever updates they need to the UI or anything else.

I've made the modifications necessary for this approach to work to the `TreeLearner` widget. Please let me know if this is an okay approach and if I should continue this route and port the other widgets to something like this.


janezd:
This was an example of bad communication: tree classifier and regressor were already merged into a single widget, but unfortunately within a PR that was neglected from July. Sorry about that... The PR was finally merged today; please update to today's master and see the regression tree widget.

Orange/widgets/__init__.py
```diff
@@ -17,6 +17,7 @@ def widget_discovery(discovery):
         "Orange.widgets.visualize",
         "Orange.widgets.classify",
         "Orange.widgets.regression",
+        "Orange.widgets.learners",
```
janezd: @BlazZupan, will the merged category (classify + regression) be called "learners"?

- We have "Unsupervised", so to keep in line with that the new category would have to be "Supervised". It's weird, I don't like it.
- A better option is "Modelling". The form doesn't fit, but we already have a mess -- "Visualize", "Classify" and "Regression", and not "Visualization", "Classification" and "Regression" (or "Visualize", "Classify" and "Regress"!?).
- The problem with "Modelling" is that it also covers unsupervised modelling.

Still, I vote for "Model" or "Modelling".

janezd:
I don't like the radio buttons. Besides using space, they unnecessarily burden the user. Say the the user connect a File widget and a learner widget to Test Learners. From the user's perspective, the learner widget outputs a "modelling algorithm". The algorithm "*should be able to handle classification and regression data*". Having to specify the kind of data seems "*unreasonable*", since the algorithm "*sees*" what kind of data it got.

The philosophy is that the user *can decide* how the algorithm should run and not that is has to *set the algorithm properly* to run for a particular data type (when the decision is trivial and can be automated).

The need for the radio buttons hence doesn't come because the user needs to decide something but just because classifier and regressor are implemented as separate classes and Test Learners needs to be given the correct class.

I suggest two solutions.
- Ideally, learning algorithms should be able to handle both regression and classification, in particular if they share most of the code, such as in knn.
- Otherwise, we can have a general dispatcher for learning. It would be derived from `Fitter`, and contain a classifier and regressor, e.g. `fitter = FitterDispatcher(tree_classifier, tree_regressor)`, and then it would `self.send('Learner', fitter)` to Test and Score. The dispatcher would then delegate the learning to either classifier or regressor, depending on the data type.

pavlin-policar:
I agree, this would be a much better approach.

But merging the learners into a single one still leaves a problem. While some of the learners (KNN, AdaBoost, RF) can be easily merged since they their UIs are identical, others take different parameters that need to be set in the UI or have different allowed parameters.

- Linear regression supports none, L1, L2 and elastic net regularization whereas logistic regression only supports L1 and L2 regularization.
- In the new tree widgets, the classification has an extra option to stop when the majority reaches a certain percent while the regression one does not have this option.
- The SVM can output either a C-SVM model for classification or an epsilon-SVM for regression.

Of course it makes sense that regression and classification problems can have slightly different parameters, but if we merge them into single learners where should these problem specific parameters appear or should they disappear altogether? Should the parameters be renamed to something more generic so they refer to either problems (but that might lead to confusion)?

If the learner widget has an input table, then displaying the correct options is simple and requires no input from the user, but if we just connect them to the Test&Score widget, how would the user be able adjust the specific parameters for instance regression since the learner widget has no awareness of the data (which makes sense)? This was the reasoning behind the radio buttons, since I haven't been able to come up with a better solution other than removing the specific parameters altogether.

*As an aside, I've noticed that the SGDClassifier can be turned into logistic regression and does support all the types of regularization that linear regression does so this may not be a problem at all since then they could have the same parameters.*

janezd:
Huh, I see. I'd say that it two methods (classification/regression) require substantially different GUIs that cannot be nicely merged into one, they should remain separate widgets. An extra option is not a problem, even a few options in separate (always visible?) boxes for classification and regression are OK, IMHO.

However, we will always have pure classification or regression methods (CN2?), and we could even keep logistic and linear regression in separate widgets if we had to.

pavlin-policar:
I do think that options that are always visible could be confusing and not the most user friendly, since when I change an option I expect a change, but that may not always happen in this case. Maybe help text, tooltips, help icons, ...

Anyway, I'll get on merging the ones that make sense first.

kernc:
Could remain visible but become disabled.

janezd:
@kernc, read more carefully. Disabled based upon ... what?

kernc:
Input, provided provided.

janezd:
I'm wasting time, I guess, but I'll bite. :)

> If the learner widget has an input table, then displaying the correct options is simple and requires no input from the user, but if we just connect them to the Test&Score widget, how would the user be able adjust the specific parameters for instance regression since the learner widget has no awareness of the data (which makes sense)?

kernc:
Right.

pavlin-policar:
What might be a good solution when there are different options in the UI in dropdown menus that are specific to classification and regression problems?

For example, AdaBoost is identical in all aspects except the algorithm and should definitely be merged, but since the learner is completely unaware of the data (until it gets some data, but that doesn't matter since it has to work with Test&Score).

I'm not sure how I would go about solving this.

janezd:
I would have boxes "Discrete target (Classification)" and "Numeric target (Regression)", and put the specific controls into them. (I may change my mind about the box names, but at the moment I like them. :)

While you're changing the interface, could you also tidy it up a bit? Make the spin boxes the same width; see the classification tree widget for an example. I guess the two combo boxes should also have the same width (but not the same as the (narrower) spins).

Orange/widgets/__init__.py
janezd: @BlazZupan agrees with `Model` as the category name.

Orange/tests/test_fitter.py
```diff
@@ -0,0 +1,169 @@
...
+        self.assertEqual(
+            fitter.classification_learner.active_preprocessors, (pp,),
+            'Classification learner should override default preprocessors '
+            'when specified in constructor')
+        self.assertEqual(
+            fitter.regression_learner.active_preprocessors, (pp,),
+            'Regression learner should override default preprocessors '
+            'when specified in constructor')
+
+    def test_use_deafult_preprocessors_property(self):
```
janezd: Typo.

pavlin-policar:
@janezd After looking over what we discussed more carefully, there should actually not be any need to set the domain on the learner inside the fitter [`__call__`](https://github.com/pavlin-policar/orange3/blob/merge-learners/Orange/modelling/fitter.py#L40) method, since each individual learner sets its own property in their own [`__call__`](https://github.com/pavlin-policar/orange3/blob/merge-learners/Orange/base.py#L54) method. It's actually the fitter instance that remains without the domain, which should be fine, since it doesn't need it and shouldn't need it.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -223,6 +232,20 @@ def send_report(self):
     def setup_layout(self):
         self.add_learner_name_widget()
         self.add_main_layout()
+        # Options specific to target variable type, if supported
+        if self.__handles_any_input_type():
+            # Only add a classification section if the method is overridden
+            if type(self).add_classification_layout is not \
+                    OWBaseLearner.add_classification_layout:
```
janezd: @astaric, we're seeing more and more of this checking whether a class has overrides an inherited method. (Not that it's nice, but it's useful.) Is it time to implement something like
```
def overrides(method, base):
    return getattr(type(method.__self__), method.__name__) != getattr(base, method.__name__)
```
to be used like (here above) `if overrides(self.add_classification_layout, OWBaseLearner)`?

janezd:
I see, of course.

It looks safe to merge, unless @kernc has any objections. Just thing: can you remove the change to ada_boost, so the pr doesn't touch more files than necessary?

pavlin-policar:
Sure thing.

Orange/modelling/fitter.py
```diff
@@ -0,0 +1,94 @@
...
+    """Handle multiple types of target variable with one learner.
+
+    Subclasses of this class serve as a sort of dispatcher. When subclassing,
+    we provide a `LearnerTypes` instance which contain actual learner classes
+    that handle appropriate data types. The fitter can then be used on any
+    data and will delegate the work to the appropriate learner.
+
+    If the learners that handle each data type require different parameters,
+    you should pass in all the possible parameters to the fitter. The fitter
+    will then determine which parameters have to passed to individual learners.
```
kernc: ... have to _be_ passed.

Orange/modelling/fitter.py
```diff
@@ -0,0 +1,94 @@
...
+        if not isinstance(dispatches, LearnerTypes):
+            raise AssertionError(
+                'The `__fits__` property must be an instance of '
+                '`Orange.base.LearnerTypes`.')
+
+    @property
+    def learner(self):
+        return self.classification_learner if \
+            self.problem_type == self.CLASSIFICATION else \
+            self.regression_learner
```
kernc: And if you _do_ this, [prefer parenthesis to backslashes](https://www.python.org/dev/peps/pep-0008/#maximum-line-length), perhaps with more logical grouping:
```py
return (self.classification_learner
        if problem_type == self.CLASSIFICATION else
        self.regression_learner)
```

Orange/modelling/fitter.py
```diff
@@ -0,0 +1,94 @@
...
+    will then determine which parameters have to passed to individual learners.
+
+    """
+    __fits__ = None
+    __returns__ = Model
+
+    # Constants to indicate what kind of problem we're dealing with
+    CLASSIFICATION, REGRESSION = range(2)
+
+    def __init__(self, *args, preprocessors=None, **kwargs):
```
kernc: Unless you know of any _now_, Fitter should accept no `*args` to pass to learners.

Orange/modelling/fitter.py
```diff
@@ -0,0 +1,94 @@
...
+    def __get_kwargs(self, kwargs, problem_type):
+        if problem_type == self.CLASSIFICATION:
+            params = self._get_learner_kwargs(self.__fits__.classification)
+            kwarg_keys = params & set(kwargs.keys())
+            kwargs = {k: kwargs[k] for k in kwarg_keys}
+        elif problem_type == self.REGRESSION:
+            params = self._get_learner_kwargs(self.__fits__.regression)
+            kwarg_keys = params & set(kwargs.keys())
+            kwargs = {k: kwargs[k] for k in kwarg_keys}
+        return kwargs
```
kernc: The body of this function can be nicely refactored into three lines or less.

Orange/modelling/fitter.py
```diff
@@ -0,0 +1,94 @@
...
+        if self.__classification_learner is None:
+            self.__classification_learner = self.__fits__.classification(
+                *self.args,
+                **self.__get_kwargs(self.kwargs, self.CLASSIFICATION))
+        self.__classification_learner.use_default_preprocessors = \
+            self.use_default_preprocessors
+        return self.__classification_learner
+
+    @property
+    def regression_learner(self):
```
kernc: The two properties can be a single, simple function: `get_learner(problem_type)`

Orange/modelling/fitter.py
```diff
@@ -0,0 +1,94 @@
...
+    # Constants to indicate what kind of problem we're dealing with
+    CLASSIFICATION, REGRESSION = range(2)
+
+    def __init__(self, *args, preprocessors=None, **kwargs):
+        super().__init__(preprocessors=preprocessors)
+        self.args = args
+        self.kwargs = kwargs
+        # Make sure to pass preprocessor params to individual learners
+        self.kwargs['preprocessors'] = preprocessors
+        self.problem_type = self.CLASSIFICATION
```
kernc: Set this to `None` initially if it's bound to be reset on some condition anyway.

Orange/base.py
```diff
@@ -63,15 +63,29 @@ def __call__(self, data):
...
         """
         Apply the `preprocessors` to the data.
         """
-        for pp in self.preprocessors:
+        for pp in self.active_preprocessors:
             data = pp(data)
         return data
 
+    @property
+    def active_preprocessors(self):
```
kernc: What could also work, likely better with fewer symbols to track, is defining this property as simply `preprocessors`. Then its setter stores the user's, and this getter returns the union.

Orange/modelling/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .fitter import *
```
kernc: Rename module to _base_.

Orange/modelling/knn.py
```diff
@@ -0,0 +1,10 @@
+from Orange.classification import KNNLearner as KNNClassification
+from Orange.modelling.fitter import Fitter, LearnerTypes
+from Orange.regression import KNNRegressionLearner
+
+
+class KNNLearner(Fitter):
+    name = 'knn'
+
+    __fits__ = LearnerTypes(classification=KNNClassification,
+                            regression=KNNRegressionLearner)
```
kernc: Will all Fitter modules be this short?

Orange/base.py
kernc: Nevermind, this is not how Python works.

Orange/base.py
```diff
@@ -18,13 +19,12 @@ class Learner:
...
     #: fitting the model
-    preprocessors = ()
+    preprocessors = None
     learner_adequacy_err_msg = ''
 
     def __init__(self, preprocessors=None):
-        if preprocessors is None:
-            preprocessors = type(self).preprocessors
-        self.preprocessors = list(preprocessors)
+        self.use_default_preprocessors = False
```
janezd: I disagree with everything except that it deserves a comment to explain why all this is necessary -- which would also answer your considerations about unnecessarily complicating things, about mandatory preprocessors and about `use_default_processors` (which is not a property of the class that should be given as an argument but rather a flag that would essentially need to be given in the `__call__` - except that it can't).

@pavlin-policar explained the workings of this two days ago and I already forgot. Better documentation would help an eager pull request reviewer understand the pull request better and thus give more informed comments. :)

Orange/base.py
```diff
@@ -63,15 +63,29 @@ def __call__(self, data):
...
         return data
 
+    @property
+    def active_preprocessors(self):
+        if self.preprocessors is not None:
+            pps = []
+            if isinstance(self.preprocessors, Iterable):
+                pps += list(self.preprocessors)
+            else:
+                pps += [self.preprocessors]
```
janezd: I thought about this too, but we do similar things in other places, and Python does it too, like in, for instance, the second argument of `isinstance`. Practicality beats purity.

Orange/modelling/fitter.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @staticmethod
+    def __check_dispatches(dispatches):
+        if not isinstance(dispatches, LearnerTypes):
+            raise AssertionError(
+                'The `__fits__` property must be an instance of '
+                '`Orange.base.LearnerTypes`.')
+
+    @property
+    def learner(self):
```
janezd: @kernc, see line 94. That's the crux of the ugly business of this PR, method `__getattr__`.

Orange/widgets/model/owknn.py
```diff
@@ -0,0 +1,65 @@
...
+
+from Orange.data import Table
+from Orange.modelling import KNNLearner
+from Orange.widgets import gui
+from Orange.widgets.settings import Setting
+from Orange.widgets.utils.owlearnerwidget import OWBaseLearner
+
+
+class OWKNNLearner(OWBaseLearner):
+    name = "K-Nearest Neighbors"
```
janezd: Not in this PR. This PR merges/move kNN from classification and regression into modelling, and doesn't make any (although potentially necessary) fixes in the kNN itself.

You of all people shouldn't request changing kNN in this PR. :)

Orange/modelling/fitter.py
janezd: +1

Orange/base.py
pavlin-policar: You're right of course, this deserves explanation. I would like to address the reason I didn't add `use_default_preprocessors` here as a constructor argument.

The reason for this was purely practical. Every learner reimplements the `__init__` method and they only specify which arguments that specific learner needs + preprocessors ([example](https://github.com/biolab/orange3/blob/master/Orange/classification/logistic_regression.py#L42-L44)). Since `owleanerwidget` would then call the learner constructor with this new argument, we would have to fix every learner implementation to support this argument as well, exactly like every learner handles the `preprocessor` argument. Since this argument would only ever be used in rare cases, I don't think it warrants changing every learner and adding that additional argument, since then each learner constructor becomes more complicated, when it should really only care about its own specific arguments and not about the underlying learner logic.

pavlin-policar:
This PR adds the `Fitter` class, which acts a sort of delegator between learners that handle classification and regression problems. This will be used in learner widgets so each widget could handle both classification and regression tasks.

One key change I've made to learners is the preprocessors. In the existing implementation, the preprocessors were kept in [`OwBaseLearner`](https://github.com/biolab/orange3/blob/master/Orange/widgets/utils/owlearnerwidget.py#L147). This could not work in the new setting where we don't know which learner we are actually going to use until the last second, when we receive the data. Therefore a new mechanism needed to be implemented. My solution was to add an `active_preprocessors` property to learners, and a `use_default_preprocessors` attribute.

The reason I didn't just make default preprocessors work by default is because this behaviour is only desired in widgets and not in the programmatic part since if somebody sets about overriding the default preprocessors, we assume they know what they are doing (I spoke to @astaric about this and he clarified that this was indeed the desired behaviour).

I did also contemplate stuffing all this logic inside the `preprocessor` attribute, but this would most likely require me to implement some kind of descriptor, and would make the entire process seem even more magical than it already is. I believe that having the `active_preprocessor` property separate from the general `preprocessors` property makes things more clear.

Orange/base.py
pavlin-policar: The thing is, it's never actually more than one preprocessor in widgets, since if we connect the `Preprocess` widget to a learner we actually get an instance of `LearnerList` or something like that. So as far as widgets go, it would make more sense to accept a single preprocessor.

As far as a programmatic API, having to wrap preprocessors in a special `PreprocessorList` whenever you want to pass in multiple preprocessors would get really annoying, really quickly.

Orange/modelling/fitter.py
```diff
@@ -0,0 +1,94 @@
...
+        """Get a `set` of kwarg names that belong to the given learner."""
+        # Get function params except `self`
+        return set(learner.__init__.__code__.co_varnames[1:])
+
+    @staticmethod
+    def __check_dispatches(dispatches):
+        if not isinstance(dispatches, LearnerTypes):
+            raise AssertionError(
+                'The `__fits__` property must be an instance of '
+                '`Orange.base.LearnerTypes`.')
```
pavlin-policar: I agree, a metaclass would be better, but as far as the dict goes, I don't see how that would be beneficial. The only thing that would achieve is that now someone, who wants to use the fitter has to pay special attention to how they spell the dict keys so it works correctly. If we use a named tuple on the other hand, we get autocompletion in IDEs and becomes basically impossible to make a mistake like that.

Orange/widgets/classify/tests/test_owlogisticregression.py
```diff
@@ -55,6 +55,6 @@ def setter(val):
     def test_output_coefficients(self):
         """Check if coefficients are on output after apply"""
         self.assertIsNone(self.get_output("Coefficients"))
-        self.send_signal("Data", self.data)
+        self.send_signal("Data", self.data[0])
```
pavlin-policar: Yes, I agree, that was a quick and dirty fix. Will change right away.

Orange/base.py
kernc: Right, thanks. I think this explanation along with the one given in https://github.com/biolab/orange3/pull/1657#issuecomment-263121263 should be added as a comment there somewhere. Otherwise I'll need to rediscover this repeatedly. :smiley:

Orange/widgets/model/owknn.py
kernc: It's a _new_, unified widget. I have no qualms about refreshing it. :stuck_out_tongue:

Orange/modelling/knn.py
pavlin-policar: Yes, I think most of them will be fairly short, with the possible exception of AdaBoost and perhaps SVM and LR. I'm fairly certain that further changes will need to be made to the Fitter, but the current implementation should cover at least half of the learners in Orange atm.

Orange/modelling/base.py
```diff
@@ -0,0 +1,89 @@
...
+    def __call__(self, data):
+        # Set the appropriate problem type from the data
+        self.problem_type = self.CLASSIFICATION if \
+            data.domain.has_discrete_class else self.REGRESSION
+        return self.get_learner(self.problem_type)(data)
+
+    def get_learner(self, problem_type):
+        """Get the learner for a given problem type."""
+        # Prevent trying to access the learner when there's no actual learner
+        if problem_type not in self.__handled_problem_types():
```
kernc: If `__fits__` were a simple dict, this could be:
```py
class Fitter:
    CLASSIFICATION, REGRESSION = 'classification', 'regression'
    def get_learner(self, problem_type):
        if problem_type not in self.__fits__:
```
saving additional 9 lines below. :smile:

Orange/modelling/knn.py
kernc: Will it make sense to scatter 4-line classes across the filesystem?

Orange/modelling/fitter.py
pavlin-policar: All right. That does make sense.

Orange/modelling/knn.py
pavlin-policar: No, but what's the alternative? Should all the fitters be in `base.py`? That doesn't make too much sense either...

kernc: Unless very complex, perhaps in _fitters.py_, dunno. ¯\\_(ツ)_/¯

Orange/base.py
```diff
@@ -63,15 +63,29 @@ def __call__(self, data):
...
         """
-        for pp in self.preprocessors:
+        for pp in self.active_preprocessors:
             data = pp(data)
         return data
 
+    @property
+    def active_preprocessors(self):
+        if self.preprocessors is not None:
+            pps = []
```
pavlin-policar: I was not aware of this feature of python. That is really cool :)

Orange/base.py
```diff
@@ -22,14 +62,15 @@ class Learner:
     learner_adequacy_err_msg = ''
 
     def __init__(self, preprocessors=None):
-        if preprocessors is None:
-            preprocessors = type(self).preprocessors
-        self.preprocessors = list(preprocessors)
+        self.use_default_preprocessors = False
+        if isinstance(preprocessors, Iterable):
+            self.preprocessors = preprocessors
```
kernc: Generators are legitimate iterables but will unwind only once. Safest to cast it to tuple explicitly.

pavlin-policar: Yes, but the processors passed in the constructor should not be generators. Also, `self.preprocessors` isn't a generator, it's a tuple. Only `self.active_preprocessors` is a generator. Unless we want to support generators being passed into the constructor?

Orange/base.py
```diff
@@ -65,13 +106,21 @@ def __call__(self, data):
...
 
+    @property
+    def active_preprocessors(self):
+        # Check if preprocessors were overriden
+        if self.preprocessors is not type(self).preprocessors:
+            yield from self.preprocessors
+            if self.use_default_preprocessors:
+                yield from tuple(type(self).preprocessors)
+        else:
+            yield from type(self).preprocessors or ()
```
kernc: Really please excuse me, but I find this slightly unintuitive to follow. Isn't it functionally the same, but more readable:
```py
# yield instance preprocessors first. they exist and are definitely a tuple, as ensured by the constructor
yield from self.preprocessors
# if the instance preprocessors aren't the same as class', apply the latter also
if (self.use_default_preprocessors
        and self.preprocessors is not type(self).preprocessors)):
```
Yielding from an empty tuple is, of course, a non-issue.

Orange/base.py
kernc: No, of course, we just want to support iterables. But generators are iterables, and the user might efficiently do:
```py
all_preprocessors = [all the preprocessors Orange supports]
valid = (p for p in all_preprocessors if requires(SomeLearner, p))
learner = SomeLearner(preprocessors=valid)
```
He might obtain those preprocessors otherwise. In any case, we currenlty require the input (if a collection of preprocessors) to be an [Iterable, not a Sequence](https://docs.python.org/3/library/collections.abc.html).
And with above user's code, this happens:
```py
>>> list(learner.active_preprocessors) == list(p for p in all_preprocessors
...                                            if requires(SomeLearner, p))
True
>>> list(learner.active_preprocessors) == list(p for p in all_preprocessors
...                                            if requires(SomeLearner, p))
False
>>> list(learner.active_preprocessors)  # Wat happen??
[]
```
Just wrapping that input, potentially a generator iterable, into tuple avoids this.

kernc: Basically, whenever you check something is an iterable, consider it might be a generator instance which can be iterated over only once.

pavlin-policar: By doing this, I also realised most of my tests concerning preprocessors actually belonged to the learner tests. I've moved them to `Orange/tests/test_base.py` since I was unable to find any other tests regarding the `Learner` class.

pavlin-policar:
@kernc Is there anything else that needs to be fixed? If not, I'd like to start merging other learners together.

Orange/modelling/base.py
```diff
@@ -0,0 +1,81 @@
+from Orange.base import Learner, Model
+
+
+class FitterMeta(type):
+    """Ensure that each subclass of the `Fitter` class overrides the `__fits__`
+    attribute with a valid value."""
+    def __new__(mcs, name, bases, kwargs):
+        # Check that a fitter implementation defines a valid `__fits__`
+        if kwargs.get('name', False):
```
kernc: I still don't think checking `name` attribute is adequate. https://github.com/biolab/orange3/pull/1771/commits/e6a14f0728f632371a213d69e8c4e8f59fc371e7 in https://github.com/biolab/orange3/pull/1771 obsoletes redundant short names on `Learner` types. Can you instead check that the class' name equals 'Fitter' or that none of the bases is named 'Fitter' or similar?

Also, `kwargs` is actually _a dict of class attributes_, so `attrs` or `namespace` or similar might be more appropriate.

PR #1659, opened 2016-10-13
[FIX] Statistics.util.stats: Fix negative #nans for sparse

nikicc:
Computing the number of nans for sparse matrices was broken and sometimes returned negative numbers.

Variable `non_zero` contains the number of defined values in each column. So to calculate the number of undefined (nans) for each column we have to substract `non_zero` from the number of values in the column i.e. `X.shape[0]` and not `X.shape[1]`. The bug was noticed when the number of defined values in some column was larger than the number of features and consequenlty the number of nans was negative. This caused density in OWTable to exceeded 100%.

screen shot 2016-10-13 at 11 55 57

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -192,7 +192,7 @@ def stats(X, weights=None, compute_variance=False):
             X.max(axis=0).toarray().ravel(),
             np.asarray(X.mean(axis=0)).ravel() if not weighted else weighted_mean,
             np.zeros(X.shape[1]),      # variance not supported
-            X.shape[1] - non_zero,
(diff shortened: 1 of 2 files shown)


nikicc:
@lanzagar IMHO we already have a test for this — the one that I corrected in this PR — just that it was incorrect.

For sparse data "nans" are actually values not present in sparse matrix and the `test_stats_sparse` already contain those since `csr_matrix` does not store zeros.

lanzagar:
You are right, the (not very clear) test that was already there (but broken) does test this too. I guess it is enough to cover the basics now that it is fixed.

PR #1662, opened 2016-10-14
[FIX] Fix an error when using latest pyqtgraph develop snapshot

ales-erjavec:
* Fix an AttributeError when using latest pyqtgraph develop branch.
* Lock aspect ratio on the plot view.

### DIFF
--- a/Orange/widgets/visualize/owlinearprojection.py
+++ b/Orange/widgets/visualize/owlinearprojection.py
@@ -408,6 +408,7 @@ class OWLinearProjection(widget.OWWidget):
         self.view.setRenderHint(QtGui.QPainter.Antialiasing, True)
         self.view.setFrameStyle(QtGui.QFrame.StyledPanel)
         self.viewbox = pg.ViewBox(enableMouse=True, enableMenu=False)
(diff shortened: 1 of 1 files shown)


lanzagar:
Looks OK.
I just have a question about locked aspect ratio - do we want it to behave differently here than elsewhere (e.g. scatterplot)? It was probably done to avoid issues/complications with axis labels and their angles?
I don't mind it, just think that it would be better to have it behave consistently across similar visualizations.

ales-erjavec:
Yes labels and text would get misaligned, so locking the aspect fixes it. However biplots should/need to preserve aspect ratio (i.e. preserve angles).

Note that MDS also uses locked aspect ratio (to preserve distances). So not all visualizations are consistent in this regard.

Come to think of it, Correspondence Analysis might also need this (it is a form of a biplot).

lanzagar:
I agree that it should be locked here, let's do it in others as well (where needed).
I guess we can leave scatter plot unlocked, where it is not needed, and might even be desirable. The different behaviour should not confuse users too much (and it is justified).

PR #1664, opened 2016-10-14
[FIX] OWHeatMap: Resend 'Selected Data' when settings change

VesnaT:


Orange/widgets/visualize/tests/test_owheatmap.py
```diff
@@ -1,5 +1,6 @@
 # Test methods with long descriptive names can omit docstrings
 # pylint: disable=missing-docstring
+import random
```
janezd: Add an empty line between python imports and Orange imports.

janezd: Just pulling your leg.

PR #1665, opened 2016-10-14
Expose widget controls by attribute names

janezd:
The motivation behind this is that we refer to the control by the name of the attribute it controls. 

This should reduce the number of attributes per widget and save us from similar names, like `binarize` for the attribute and `cb_binarize` for the checkbox that controls binarization.

If we decide we like it, I have to close one possible problem; in the dictionary that is used here to map attributes to controls, an attribute can be related to multiple controls. This is rare and can be fixed easily.

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -184,6 +184,7 @@ class OWComponent:
     """
     def __init__(self, widget):
         setattr(self, CONTROLLED_ATTRIBUTES, ControlledAttributesDict(self))
+        self.controls = ControlGetter(self)
 
         if widget.settingsHandler:
             widget.settingsHandler.initialize(self)
@@ -3240,3 +3241,27 @@ class FloatSlider(QSlider):
         # For compatibility with qwtSlider
         # TODO If it's related to Qwt, remove it
         self.setScale(minValue, maxValue, step)
+
+
+class ControlGetter:
+    """
+    Provide access to GUI elements based on their corresponding attributes
+    in widget.
+
+    Every widget has an attribute `controls` that is an instance of this
+    class, which uses the `controlledAttributes` dictionary to retrieve the
+    control (e.g. `QCheckBox`, `QComboBox`...) corresponding to the attribute.
+    For `OWComponents`, it returns its controls so that subsequent
+    `__getattr__` will retrieve the control.
+    """
+    def __init__(self, widget):
+        self.widget = widget
+
+    def __getattr__(self, name):
+        widget = self.widget
+        if hasattr(widget, "settingsHandler") and \
+                name in widget.settingsHandler.provider.providers:
+            # Control within an OWComponent are retrieved by its `controls`
+            return getattr(widget, name).controls
+        else:
+            return widget.controlledAttributes[name][0].control
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -112,7 +112,22 @@ def resource_filename(path):
 
 class OWComponent:
     """
-    Mixin for classes that contain controlled attributes.
+    Mixin for classes that contain settings and/or attributes that trigger
+    callbacks when changed.
+
+    The class initializes the settings handler, provides `__setattr__` that
+    triggers callbacks, and provides `control` attribute for access to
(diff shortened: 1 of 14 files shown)


Orange/widgets/gui.py
```diff
@@ -3240,3 +3241,16 @@ def setRange(self, minValue, maxValue, step=1.0):
         # For compatibility with qwtSlider
         # TODO If it's related to Qwt, remove it
         self.setScale(minValue, maxValue, step)
+
+
+class ControlGetter:
+    def __init__(self, widget):
```
astaric: This class could use a docstring describing what it does and how does it actually find controls.

Orange/widgets/gui.py
```diff
@@ -3240,3 +3241,16 @@ def setRange(self, minValue, maxValue, step=1.0):
...
+        self.widget = widget
+
+    def __getattr__(self, name):
+        widget = self.widget
+        if hasattr(widget, "settingsHandler") and \
```
astaric: I guess that the hasattr part of the check is there to handle OWComponents, and the second part is there to handle multilevel gets?

If so, could you please add a comment stating that, as it not clear from the code. (at first I commented that the hasattr could be removed as widget always has a settingsHandler, but that is not the case for owcomponents)

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -372,6 +372,7 @@ def set_subset_data(self, subset_data):
                 self.warning("Data subset does not support large Sql tables")
                 subset_data = None
         self.subset_data = self.move_primitive_metas_to_X(subset_data)
+        self.controls.graph.alpha_value.setEnabled(subset_data is None)
```
astaric: Could you add some tests, specifically for multilevel access, checking both the cases when the "graph" is declared as settings provider and when it is not?

(tests should **not** be a part of scatterplot tests, I am just writing a comment here because it is convenient :))

janezd:
@astaric Looks like done.

I also tried settings as properties, but I gave up. It adds another layer of magic without any substantial benefit. We have settings that are not controlled, controls that are not settings, and attributes that are neither but appear in format strings. The current mechanism (with simplifications from this PR) is necessary and sufficient.

astaric:
I have played around with the Setting-property in https://github.com/astaric/orange3/commit/98dc8178d83cac0e45b4d8843fb18114feb5ccb2#diff-059bac55767274615c764f3c76104d3dR53

It proved useful for list settings that are used in models (as they currently require custom SettingsHandler that copy the list from/to model). The code is not general though and I do not like the way it looks yet.

Orange/widgets/gui.py
```diff
@@ -111,86 +110,42 @@ def resource_filename(path):
     return pkg_resources.resource_filename(__name__, path)
 
 
-class ControlledAttributesDict(dict):
+class OWComponent:
     """
-    Dictionary with names of attributes for which the `master` needs to be
-    notified of their changes
+    Mixin for classes that contain controlled attributes.
```
astaric: What are controlled attributes? Why would I want to use this mixin?

Orange/widgets/gui.py
```diff
@@ -111,86 +110,42 @@ def resource_filename(path):
...
+        self.controlled_attributes = defaultdict(list)
+        self.controls = ControlGetter(self)
+        if widget is not None and widget.settingsHandler:
+            widget.settingsHandler.initialize(self)
 
-    def __setitem__(self, key, value):
-        if key not in self:
-            super().__setitem__(key, [value])
+    def connect_control(self, name, func):
+        if "." in name:
```
astaric: docstring?

janezd:
> The code is not general though and I do not like the way it looks yet.

Generalization is a problem. Not all controlled attributes are settings, so you can't subclass Setting.  Making the Setting a subclass of sth like `ControlledAttribute` doesn't work either since you have different kinds of controlled attributes, e.g. they have different callbacks depending on whether they control check boxes or line edits or ... You can solve this by having attributes `getter` and `setter` on `ControlledAttribute`. As you start really working on it, it becomes a great mess.

Thanks for the comment on documentation and docstrings. Done.

Issue #1668, opened 2016-10-17
MDS: faulty visualization with 'similar pairs' option


ajdapretnar:
This is what I see in MDS with 'Show similar pairs' option on. When disabling it (sliding to 0), grey squares no longer appear.

![mds-bug](https://cloud.githubusercontent.com/assets/12524972/19428440/77fb5e80-944a-11e6-9f26-611d623169f3.png)

PR #1670, opened 2016-10-17
[FIX] MDS Widget: Fix zero length line, gray square bug

ales-erjavec:
Filter out QGraphicsLineItem with zero length. When using the Qt's
raster graphics system they cause bad rendering artifacts.

Fixes gh-1668


ales-erjavec:
I can only duplicate the bug (gh-1668) on OSX if I set the `QT_GRAPHICSSYSTEM=raster` env variable when running orange. Please check that this fix also works on Windows.

PR #1672, opened 2016-10-18
[FIX] Scatter Plot: Fix a error when restoring from pre DomainModel workflows

ales-erjavec:
... when the attr_* were stored by name.

For example open the second workflow (Interactive Visualizations) in the 'MenuBar->Help->Examples'

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -182,32 +182,32 @@ class OWScatterPlot(OWWidget):
         self.sampling.setVisible(False)
 
         box = gui.vBox(self.controlArea, "Points")
-        color_model = DomainModel(
+        self.color_model = DomainModel(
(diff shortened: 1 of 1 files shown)


janezd:
I knew that #1608 breaks compatibility, but decided against introducing legacy patches.

I changed my mind. I don't oppose to adding this PR. Yet this example shows we need a general solution. 7e454bd2b43eeb4fe618532c6d549796968e3b2a is going to create the same problem in Sieve.

I suppose we need settings versions and a method `OWWidget.settings_migration(old_settings)` that returns migrated settings (or patches the given setting in-place). The method would be called before settings are applied to the widget. Classes like `DomainContextHandler` would provide utility functions such as those in this PR. @astaric?

astaric:
One day is gone. A general solution turned out to be more complicated. While it may come at a later time, I am merging this now as it fixes a known bug.

PR #1673, opened 2016-10-18
[ENH] Canvas: Always show the link dialog if the user holds Shift

janezd:
Numerous widgets now provide the signal with annotated data (#1655), which will be used quite often, but it is not the default output. Selected data should stay the default; even always showing the dialog would be annoying. The user thus has to connect the widgets, double click the link, and reroute. This annoyed me a lot.

I suggest that holding shift while connecting should trigger opening the link dialog. To not confuse the user, this should happen even when there's just a single signal on both sides.

I know this feature is a bit of easter egg, but a really useful one.

@ales-erjavec, the change is really simple. Have I overlooked anything?

### DIFF
--- a/Orange/canvas/document/interactions.py
+++ b/Orange/canvas/document/interactions.py
@@ -228,6 +228,7 @@ class NewLinkAction(UserInteraction):
         self.sink_item = None
         self.from_item = None
         self.direction = None
+        self.force_link_dialog = False
 
         # An `NodeItem` currently under the mouse as a possible
         # link drop target.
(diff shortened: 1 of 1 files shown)


ales-erjavec:
Please initialize the flag in the class `__init__` and clear it after the link is created.

You might also make it slightly less of an easter egg and mention this in the `QuickHelpTipEvent` that is posted in the [`mousePressEvent`](https://github.com/biolab/orange3/blob/master/Orange/canvas/document/interactions.py#L322) method

janezd:
Done, I guess. Thanks for the tip about the tip!

kernc:
The feature stops being an easter egg if you document it.

PR #1674, opened 2016-10-19
[ENH] SQL Server support in SQL widget

astaric:
Backends moved to a separate package, moving backend-specific code from SqlTable/widgets to the appropriate backend

### DIFF
--- a/Orange/data/sql/__init__.py
+++ b/Orange/data/sql/__init__.py
@@ -1,2 +1,0 @@
-import Orange.misc
-psycopg2 = Orange.misc.import_late_warning("psycopg2")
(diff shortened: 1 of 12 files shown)


astaric:
I have no idea how to test mssql on travis, since the database is not provided there, so I am merging this, even while it lowers the coverage. Ideas are welcome.

Orange/data/sql/__init__.py
```diff
@@ -1,2 +0,0 @@
-import Orange.misc
-psycopg2 = Orange.misc.import_late_warning("psycopg2")
```
kernc: Since this change, `import_late_warning` is used nowhere.
We could test if at least one of the backends is available and warn if none.

PR #1677, opened 2016-10-19
[ENH] Visualize widgets: Output Annotated data and Fixups

VesnaT:

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -1101,7 +1101,7 @@ class OWHierarchicalClustering(widget.OWWidget):
 
         if not selected_indices:
             self.send("Selected Data", None)
-            annotated_data = create_annotated_table(items, selected_indices) \
+            annotated_data = create_annotated_table(items, []) \
                 if self.selection_method == 0 and self.matrix.axis else None
             self.send(ANNOTATED_DATA_SIGNAL_NAME, annotated_data)
             return
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -1148,7 +1148,7 @@ class OWHierarchicalClustering(widget.OWWidget):
                 selected_data = data[mask]
                 if self.append_clusters:
                     def remove_other_value(vars_):
-                        vars_ = [var for var in vars_]
+                        vars_ = list(vars_)
                         clust_var = vars_[-1].copy()
                         clust_var.values.pop()
                         vars_[-1] = clust_var
--- a/Orange/widgets/unsupervised/tests/test_owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/tests/test_owhierarchicalclustering.py
@@ -28,13 +28,12 @@ class TestOWHierarchicalClustering(WidgetTest, WidgetOutputsTestMixin):
         return [14, 15, 32, 33]
 
     def _compare_selected_annotated_domains(self, selected, annotated):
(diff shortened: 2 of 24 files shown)


Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -1146,7 +1146,7 @@ def commit(self):
                 selected_data = data[mask]
                 if self.append_clusters:
                     def remove_other_value(vars_):
-                        vars_ = [var for var in vars_]
+                        vars_ = list(vars_).copy()
```
janezd: Ooops, in my earlier comment I thought that `vars_` was a list. Sorry. Now that you have to convert it to a list anyway, `list(vars_)` will suffice; you can skip `copy()`.

Orange/widgets/unsupervised/tests/test_owhierarchicalclustering.py
```diff
@@ -28,13 +28,12 @@ def _select_data(self):
...
-                             for var in selected.domain.variables)))
+        self.assertEqual(annotated.domain.variables,
+                         selected.domain.variables)
         self.assertNotIn("Other", selected.domain.metas[0].values)
         self.assertIn("Other", annotated.domain.metas[0].values)
-        self.assertTrue(
-            all((var in [var.name for var in annotated.domain.metas]
-                 for var in [var.name for var in selected.domain.metas])))
+        self.assertLess(set(var.name for var in selected.domain.metas),
+                        set(var.name for var in annotated.domain.metas))
```
janezd: No need to change anything here, but: you can also use set comprehensions in such cases. (I keep forgetting this, too.)

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -224,6 +229,8 @@ def set_tree(self, model=None):
             # if hasattr(model, 'meta_depth_limit'):
             #     self.depth_limit = model.meta_depth_limit
             #     self.update_depth()
+        self.send(ANNOTATED_DATA_SIGNAL_NAME,
```
janezd: Is this needed? The "Selected data" signal is not sent here. I commented it out and it seems it still works.

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -347,6 +355,11 @@ def commit(self):
         data = self.tree_adapter.get_instances_in_nodes(
             self.clf_dataset, [item.tree_node.label for item in items])
         self.send('Selected Data', data)
+        nodes = [i.tree_node.label for i in self.scene.selectedItems()
```
janezd: Isn't this already computed above, as the second argument in the call of `get_instances_in_nodes`?

Orange/widgets/visualize/owruleviewer.py
```diff
@@ -18,8 +20,9 @@ class OWRuleViewer(widget.OWWidget):
     inputs = [("Data", Table, 'set_data'),
               ("Classifier", _RuleClassifier, 'set_classifier')]
 
-    data_output_identifier = "Filtered data"
-    outputs = [(data_output_identifier, Table)]
+    data_output_identifier = "Selected Data"
```
janezd: I hate this.

First, it should be upper case -- if for nothing else, to match the case of `ANNOTATED_DATA_SIGNAL_NAME`.

Second, it's probably the only widget which defines a constant for the signal name. This could be a good idea, but only if we do it elsewhere too.

It's not your fault (it was here before), and removing it is not a part of this PR. But since you have to rename it anyway, you could perhaps simply get rid of it. Just a suggestion, you don't have to.

janezd:
@VesnaT, a small rebase, please.

PR #1678, opened 2016-10-20
[FIX] OWTable: Sort Continuous metas as floats; not strings

nikicc:
Continuous variables inside meta attributes are currently sorted by their string represenation and not by float values. This happens since dtype for metas is object and we map all object to strings before sorting. This PR checks whether the sorting variable is ContinuousVariable and casts values to floats.

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -1015,6 +1015,9 @@ class TableModel(QAbstractTableModel):
                 and role == TableModel.ValueRole:
             col_view, _ = self.source.get_column_view(coldesc.var)
             col_data = numpy.asarray(col_view)
+            if coldesc.var.is_continuous:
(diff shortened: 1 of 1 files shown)


Orange/widgets/utils/itemmodels.py
```diff
@@ -1015,6 +1015,8 @@ def columnSortKeyData(self, column, role):
                 and role == TableModel.ValueRole:
             col_view, _ = self.source.get_column_view(coldesc.var)
             col_data = numpy.asarray(col_view)
+            if coldesc.var.is_continuous:   # continuous from metas have dtype object; cast it to float
```
lanzagar: Please avoid comments on the same line when that results in line length > 80.

nikicc:
@lanzagar I moved it inside the if. Is this better?

lanzagar:
It is.

PR #1679, opened 2016-10-21
Fix pythagorean weights

pavlin-policar:
Fix weight adjustment in Pythagorean tree widget - they were broken with the new trees.

### DIFF
--- a/Orange/widgets/visualize/pythagorastreeviewer.py
+++ b/Orange/widgets/visualize/pythagorastreeviewer.py
@@ -653,6 +653,9 @@ class PythagorasTree:
             The parent square of the given child.
         node : int
             The node label of the child.
+        weight : float
+            The weight of the node relative to its parent e.g. two children in
+            relation 3:1 should have weights .75 and .25, respectively.
 
         Returns
         -------
--- a/Orange/widgets/visualize/pythagorastreeviewer.py
+++ b/Orange/widgets/visualize/pythagorastreeviewer.py
@@ -105,7 +105,6 @@ class PythagorasTreeViewer(QGraphicsWidget):
         self.__calc_node_color_func = kwargs.get('node_color_func')
         self.__get_tooltip_func = kwargs.get('tooltip_func')
         self._interactive = kwargs.get('interactive', True)
-        self._weight_adjustment = kwargs.get('weight_adjustment', lambda x: x)
 
         self._square_objects = {}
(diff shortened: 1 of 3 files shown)


Orange/widgets/visualize/pythagorastreeviewer.py
```diff
@@ -111,7 +112,7 @@ def __init__(self, parent=None, adapter=None, depth_limit=0, padding=0,
         if adapter is not None:
             self.set_tree(adapter)
 
-    def set_tree(self, tree_adapter):
+    def set_tree(self, tree_adapter, weight_adjustemnt=None):
```
janezd: Typo (here and below, where it's used).

Orange/widgets/visualize/pythagorastreeviewer.py
```diff
@@ -114,13 +114,15 @@ def __init__(self, parent=None, adapter=None, depth_limit=0, padding=0,
...
-    def set_tree(self, tree_adapter):
+    def set_tree(self, tree_adapter, weight_adjustment=lambda x: x):
         """Pass in a new tree adapter instance and perform updates to canvas.
 
         Parameters
         ----------
         tree_adapter : TreeAdapter
             The new tree adapter that is to be used.
+        weight_adjustment : callable
+            A weight adjustment function that with signature `x -> x`
```
janezd: Remove 'that'

janezd:
Load some data and change the "Size" combo to Square root. The tree updates accordingly. Now go to the File widget and click Reload. The Size combo stays at Square root (which is correct) but the tree is rendered as if the size was set normal. It seems that the problem is fixed by adding argument `weight_adjustment=self.SIZE_CALCULATION[self.size_calc_idx][1]` to the call of `self.ptree.set_tree` in `OWPythagorasTree.set_tree`.

janezd:
One other detail to think about: the name "Size" and the values in the combo are uninformative, even for me. Perhaps rename it to "Square size"? (Although it still doesn't say whether it's the square's area or side.) For the values of the combo, I don't know what's Normal. Is it linear?

Perhaps add a box "Square size". It would start with a combo "Subset size corresponds to area", "Subset size corresponds to side", "Logarithmic scaling". Below it is the slider for logarithmic mapping.

I don't know whether this would look good or it's too long or wordy.

This doesn't have to be changed in this commit.

PR #1693, opened 2016-10-28
[ENH] Add appveyor configuration

ales-erjavec:
Add [appveyor](https://appveyor.com) CI service configuration for testing on Windows.
An example output can be [viewed here](https://ci.appveyor.com/project/ales-erjavec/orange3).

### DIFF
--- a/.ci_tools/appveyor/build.cmd
+++ b/.ci_tools/appveyor/build.cmd
@@ -1,0 +1,21 @@
+ @echo off
+:: To build extensions for 64 bit Python 3, we need to configure environment
+:: variables to use the MSVC 2010 C++ compilers from GRMSDKX_EN_DVD.iso of:
+:: MS Windows SDK for Windows 7 and .NET Framework 4
+::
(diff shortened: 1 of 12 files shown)


ales-erjavec:
The test unfortunately take a lot of time (at least in their current state) so I do not think running on every pull request is an option.

ales-erjavec:
Blocked on #1698

astaric:
@ales-erjavec If the slowness of the test is the only problem, we can enable the integration, but mark it as optional. This way, appveyor builds would run, but if they take too long, we can still merge.

ales-erjavec:
> If the slowness of the test is the only problem, we can enable the integration, but mark it as optional. This way, appveyor builds would run, but if they take too long, we can still merge.

Sure, why not.

astaric:
The trusty build that caused the failing python3.4 tests on travis has beed disabled. Can you rebase onto the latest master so we can check if everything is in order now?

PR #1698, opened 2016-10-28
[FIX] Fix tests

ales-erjavec:
- Fix tests so they run with PyQt5, without sql backend and on Windows.
- Fix QtWebEngine availability test in canvasmain.py


ales-erjavec:
These were originally part of #1693

PR #1699, opened 2016-10-28
[ENH] Add VizRank to Mosaic

janezd:
VizRank for Mosaic. Computes chi-squares, either across class distributions for each cell (when the widget shows class distributions) or across cells.


janezd:
Note to self: Limit the number of attributes to two. Run and let it finish. Change it to 3 and back to 2. The button enables and changes to Start. Push it and it will add rows without removing anything, hence duplicating each row.

PR #1605, opened 2016-09-24
[ENH] Allow removing a single condition in Select Rows

RachitKansal:

PR #1701, opened 2016-11-01
Hierarchical clustering dendogram now shows the correct labels when the distances are calculated from columns

RachitKansal:


Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -966,6 +966,7 @@ def _set_items(self, items, axis=1):
         elif not axis:
             model[:] = ["None", "Enumeration", "Attribute names"]
             self.annotation_idx = 2
+            self._update_labels()
```
kernc: If this is required here, after `annotation_idx` changes, how come it is not required on [line 985 below](https://github.com/biolab/orange3/pull/1701/commits/193b693b6af1554a61ecabf97659120e56657f8c#diff-22fbad01e8c84125308a0a7fe72ba0f1R985)?

kernc:
The problem with this PR is that it shares some commits with your other PR https://github.com/biolab/orange3/pull/1605. Could you follow the standard [git flow](https://guides.github.com/introduction/flow/) and create a separate [topic branch](https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows#Topic-Branches) for each PR?

janezd:
@RachitKansal, thanks for the fix. While adding `_update_labels` does correct the problem, it shouldn't have been necessary. Fixing a bug in module `gui` (#1764) solves this, plus there is another PR (#1748) which changes the treatment of the annotation attribute altogether. Hence I'm closing this PR. Your contribution is appreciated, although it didn't make it in the distribution!

PR #1704, opened 2016-11-03
OWPreprocess: Randomize - Enable replicable shuffling

VesnaT:
##### Issue

Randomize should have a check box for replicable results

##### Description of changes
Enable setting random seed for Randomize preprocessor.
Add checkbox to OWPreprocess (Randomize editor).

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/preprocess.py
+++ b/Orange/preprocess/preprocess.py
@@ -326,6 +326,9 @@ class Randomize(Preprocess):
         If Randomize.RandomizeAttributes, attributes are shuffled.
         If Randomize.RandomizeMetas, metas are shuffled.
 
+    rand_seed : int (optional)
+        Random seed
+
     Examples
     --------
     >>> from Orange.data import Table
@@ -339,8 +342,9 @@ class Randomize(Preprocess):
                      "RandomizeMetas")
     (RandomizeClasses, RandomizeAttributes, RandomizeMetas) = RandTypes
 
-    def __init__(self, rand_type=RandomizeClasses):
+    def __init__(self, rand_type=RandomizeClasses, rand_seed=None):
         self.rand_type = rand_type
+        self.rand_seed = rand_seed
 
     def __call__(self, data):
         """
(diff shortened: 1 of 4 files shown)


VesnaT:
@lanzagar screenshot should be updated

lanzagar:
I opened a running issue (#1705), where we can keep track of screenshots that need to be updated.
You can add stuff to that list in the future.

Also this PR changes code and includes a test, but does not update documentation. So it should have appropriate things checked in the description.

"Issue" and "Description of changes" in the PR template should be filled out with at least a few words (can be very brief for obvious changes)

Orange/preprocess/preprocess.py
```diff
@@ -339,8 +342,9 @@ class Randomize(Preprocess):
                      "RandomizeMetas")
     (RandomizeClasses, RandomizeAttributes, RandomizeMetas) = RandTypes
 
-    def __init__(self, rand_type=RandomizeClasses):
+    def __init__(self, rand_type=RandomizeClasses, rand_seed=None):
         self.rand_type = rand_type
+        np.random.seed(rand_seed)
```
lanzagar: When `rand_seed` is provided, it should be used for _every_ call. Here you set the seed in the constructor so only the first call will have the expected output, but subsequent calls will depend on previous invocations.

So `init` should store the seed and it should be set in `randomize` before calls to random.shuffle.

As this is an important issue that was almost overlooked we might want to add a test for this as well. E.g. construct a randomizer and use it on A and B. Then use it on C and B. The randomization of B should be the same both times.

Orange/widgets/data/owpreprocess.py
```diff
@@ -827,25 +828,34 @@ def __init__(self, parent=None, **kwargs):
...
     @staticmethod
     def createinstance(params):
         rand_type = params.get("rand_type", Randomize.RandomizeClasses)
-        return _Randomize(rand_type=rand_type)
+        rand_seed = params.get("rand_seed", 1)
+        return _Randomize(rand_type=rand_type, rand_seed=rand_seed)
 
     def __repr__(self):
-        return self.__rand_type_cb.currentText()
+        return "{}, {}".format(self.__rand_type_cb.currentText(),
```
lanzagar: The string constructed in this `__repr__` is not very consistent - it uses human readable names for `rand_type` (not indexes), but the direct representation of `rand_seed` attribute (instead of its meaning).

Not sure if this is used for anything else, but it is definitely not OK for reports, which should be understandable to humans looking at them. Currently this is shown for example:

> Randomize: Classes, 1
> Randomize: Features, None

VesnaT:
Done.

PR #1711, opened 2016-11-04
[FIX] owfile: Hide apply button after resetting editor_model

lanzagar:
##### Issue

After changes to domain editor the Apply button appears. It was not hidden again when different data was selected.

##### Description of changes
Hide apply button when load_data is called and editor_model reset.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -217,9 +217,10 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
 
         self.apply_button = gui.button(
             box, self, "Apply", callback=self.apply_domain_edit)
-        self.apply_button.hide()
+        self.apply_button.setEnabled(False)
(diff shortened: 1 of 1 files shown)


janezd:
Congratulations: you will have to rebase this one-line PR before it can be merged. :)

Orange/widgets/data/owfile.py
```diff
@@ -263,6 +263,7 @@ def load_data(self):
         # file readers
         # pylint: disable=broad-except
         self.editor_model.set_domain(None)
+        self.apply_button.hide()
```
kernc: Can we please agree disabling is better UX than hiding?
```py
self.apply_button.setEnabled(False)
```

lanzagar:
Rebased and switched to disabling/enabling the button instead of hiding/showing.

PR #1722, opened 2016-11-04
PCA: make self.variance_covered always an integer

markotoplak:
##### Issue

There were bug reports on some Windows and OS X system when
setting the corresponding QSpinBox's value with numpy.float.
I could not repreduce the bug. For me QSpinBox.setValue also
worked with floats.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owpca.py
+++ b/Orange/widgets/unsupervised/owpca.py
@@ -278,7 +278,8 @@ class OWPCA(widget.OWWidget):
         self._set_horline_pos()
 
         if self._pca is not None:
-            self.variance_covered = self._cumulative[components - 1] * 100
+            # convert to int as we had bug reports with QSpin.setValue on some matchines
+            self.variance_covered = int(self._cumulative[components - 1] * 100)
 
         if current != self._nselected_components():
             self._invalidate_selection()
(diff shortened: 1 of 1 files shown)


Orange/widgets/unsupervised/owpca.py
```diff
@@ -278,7 +278,8 @@ def _on_cut_changed(self, line):
         self._set_horline_pos()
 
         if self._pca is not None:
-            self.variance_covered = self._cumulative[components - 1] * 100
+            # convert to int as we had bug reports with QSpin.setValue on some matchines
+            self.variance_covered = int(self._cumulative[components - 1] * 100)
```
janezd: Perhaps round before casting? (Same elsewhere.)

ales-erjavec:
The type error was probably caused by `variance_covered` being a NaN. I.e. `spin.setValue(float("nan"))` raises an `TypeError: setValue(self, int): argument 1 has unexpected type 'float'`. The proposed changeset will only replace the TypeError with an `ValueError: cannot convert float NaN to integer`

markotoplak:
Thanks @ales-erjavec

PR #1723, opened 2016-11-04
owpca: fix error when variance is set to 100

markotoplak:
##### Issue

When an user set covered variance to 100, the widget crashed.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


astaric:
If I use PCA widget on iris dataset and select all 4 components, the Variance Covered box shows only 99%. I would expect that 100% of the variance would be covered if the number of selected components is the same as the number of attributes in the original dataset.

PR #1724, opened 2016-11-04
[ENH] Setting migration

astaric:
##### Issue
When widgets are opened with old version of settings, they break in many different ways.

##### Description of changes
This PR introduces the concept of setting versions and migrations. When a widget changes its settings in a way that is not compatible with old version, it should raise its settings_version attribute and implement the migrate_settings (migrate_context for context settings) method.

##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation

### DIFF
--- a/Orange/widgets/widget.py
+++ b/Orange/widgets/widget.py
@@ -154,6 +154,13 @@ class OWWidget(QDialog, Report, ProgressBarMixin, WidgetMessagesMixin,
     settingsHandler = None
     """:type: SettingsHandler"""
 
+    #: Version of the settings representation
+    #: Subclasses should increase this number when they make breaking
+    #: changes to settings representation (a settings that used to store
+    #: int now stores string) and handle migrations in migrate and
+    #: migrate_context settings.
+    settings_version = 1
+
     savedWidgetGeometry = settings.Setting(None)
 
     #: A list of advice messages (:class:`Message`) to display to the user.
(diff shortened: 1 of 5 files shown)


Orange/widgets/widget.py
```diff
@@ -738,6 +745,32 @@ def _userconfirmed():
 
         self.__msgwidget.accepted.connect(_userconfirmed)
 
+    @classmethod
+    def migrate_settings(cls, settings, version):
```
astaric: This belongs to whoever defines the settings handler (the class passed in parameter to SettingHandler.create call). As components can be used as a part of the widget, they do not have their own settingHandlers and reuse the widget's. The widget is therefore responsible for migration of all the settings.

astaric:
I have added a mention of migrations to the widget development tutorial.

Orange/widgets/widget.py
janezd: > The widget is therefore responsible for migration of all the settings.

Say that some component is reused in multiple widgets. If we change a setting in the component, we have to add the corresponding migration to all widgets? Including, say, widgets in third-party add-ons that use our components? With the later, any change in component's settings would break compatibility with widgets; with introducing component-level migrations, this can be avoided.

Wouldn't it thus be better if settings handler traversed all components (including the widget itself) and called the corresponding migration method in each? This would also require setting versioning on components, not widgets.

astaric: If the settings break with a change, so might the widget that uses the component. If for instance a component used to store a list of indices in a setting and now it stores a single number, a widget trying to index the appropriate component attribute will crash.

While I can move the migrate_settings method to the component, what can we do with migrate_context? Leave it on the widget? If this is the case, I would prefer to keep the methods in the same place. And if the migrations are placed on a component, who defines the settings version?

IMO, if component is complicated enough to warrant its own migration, it can define its own migrate_settings, migrate_context, but the widget should be aware of that and call the method explicitly in its own migrate_settings method with proper parameters.

PR #1642, opened 2016-10-07
[FIX] Switch Sieve to DomainModel, which also fixes VizRank crash on meta attributes

janezd:
Fixes #1546.
<strike>Breaks compatibility with old settings. See #1672 for a similar problem with the scatter plot.
Blocked by #1724 and #1727.</strike>

PR #1727, opened 2016-11-05
Context: Add rename and migrate_str_to_variable

janezd:
<strike>Needs #1724.</strike> Usage example in #1642 (8227b626609a5cae23430ef2dbfde4d7584c246e)

##### Description of changes

Add utility functions for two common setting migrations.

##### Includes
- [X] Code changes
- [x] Tests
- [X] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owdatasampler.py
+++ b/Orange/widgets/data/tests/test_owdatasampler.py
@@ -17,8 +17,7 @@ class TestOWDataSampler(WidgetTest):
     def test_error_message(self):
         """ Check if error message appears and then disappears when
         data is removed from input"""
-        self.widget.controlledAttributes["sampling_type"][0].control.buttons[
-            2].click()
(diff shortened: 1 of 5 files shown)


astaric:
I am not so fond of putting this code on Context. Context is a data object that is manipulated by ContextHandlers.

Both methods could be converted to a functions, moved to a separate module (migrations?) and imported from there when needed. rename_setting could also be modified to support both context and ordinary settings.

janezd:
Generalization of `rename_setting` makes sense. I'll do it.

I'm not a fan of module-level functions, though. I'm also not sure we'd have many utility functions for migrations, so they may not deserve a separate module.

The utility function for migrations were initially class methods of context handlers, but then I noted they always have context as the first element. What do you say about changing them to class methods of `SettingsHandler` (which already includes methods for writing/reading to/from files)?

Issue #1714, opened 2016-11-04
TimeVariable copy issues


lanzagar:
##### Orange version

3.3.9.dev0+ee79a34

##### Expected behavior
TimeVariables work

##### Actual behavior
Confusion with ContinuousVariables causes problems somewhere

##### Steps to reproduce the behavior
File widget -> load cyber-security-breaches -> change all datetime to numeric -> apply -> reload

##### Additional info (worksheets, data, screenshots, ...)
'ContinuousVariable' object has no attribute 'parse'

PR #1734, opened 2016-11-07
Fix confusion

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/1714

##### Description of changes
Prefer `Variable.make`-made `var` instead of the specified `coltype`.
Use logging to log file loading exceptions for simpler debugging.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -570,6 +570,7 @@ class FileFormat(metaclass=FileFormatMeta):
 
             cols, domain_vars = append_to
             cols.append(col)
+            var = None
             if domain_vars is not None:
                 if names and names[col]:
                     # Use existing variable if available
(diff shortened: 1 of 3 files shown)


lanzagar:
While testing I encountered an infinite recursion problem. Not sure what I did, but I was changing from `datetime` to `numeric` and back and reloading data (cyber security)

```
  File "/home/lan/dev/orange3/Orange/widgets/utils/itemmodels.py", line 1112, in data
  File "/home/lan/dev/orange3/Orange/widgets/utils/itemmodels.py", line 863, in format_dense
    return str(instance[var])
  File "/home/lan/dev/orange3/Orange/data/variable.py", line 166, in __str__
  File "/home/lan/dev/orange3/Orange/data/variable.py", line 924, in repr_val
    return str(val)
  File "/home/lan/dev/orange3/Orange/data/variable.py", line 166, in __str__
...
```

(might not even be a direct issue of this PR, but something related)

kernc:
> might not even be a direct issue of this PR

Yeah, probably isn't. Can you reproduce?

lanzagar:
No, I am unable to reproduce now.
And it does fix the original bug, so I am merging, but we should keep an eye for strange behaviour.

PR #1735, opened 2016-11-07
[ENH] Map widget

kernc:
##### Issue
Add map widget for plotting geospatial data.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


astaric:
Can you skip the owmap tests when only PyQt4 is available? As we will not drop support for PyQt4 for some time, it would be nice if the tests still passed.

ajdapretnar:
Documentation is prepared but waiting for some additional fixes (heatmap for classification and resolution regarding the test data set).

I am experiencing the same issues as @lanzagar with additional issue of labeling by name on airports data set. I get AttributeError: 'LeafletMap' object has no attribute '_label_values'. In essence, labeling by string doesn't work for me.

kernc:
@ajdapretnar Have yet to confirm, but there may not be a classification heatmap so soon.

> Zoom to rectangle selection does not work for me.

@lanzagar Any errors in the error console?

lanzagar:
Bumping this to 3.3.10 as it is still under active development and needs documentation etc.

kernc:
> @ajdapretnar Have yet to confirm, but there may not be a classification heatmap so soon.

Managed something.

screenshot - 12022016 - 11 34 23 am

ajdapretnar:
That's exactly what I had in mind. Thanks! :) Will test it again next week!

kernc:
Blocked on #1799.

Orange/widgets/visualize/owmap.py
```diff
@@ -0,0 +1,807 @@
...
+        QTimer.singleShot(10, add_points)
+
+    def set_subset_ids(self, ids):
+        self._subset_ids = ids
+        self.redraw_markers_overlay_image(new_image=True)
+
+
+class OWMap(widget.OWWidget):
+    name = 'Map'
+    description = 'Show data poits on a world map.'
```
ajdapretnar: data poits --> data points

kernc:
Perhaps blocked by https://github.com/biolab/orange3/pull/1816.

kernc:
Comments addressed, except:

> Starting zoom is not correct (clicking the center button after that works)

No idea.

> correctly hidden on systems without QT5

Re-run Orange with `--force-discovery`.

ajdapretnar:
Documentation ready in #1806.

ajdapretnar:
As far as I'm concerned, it works.
Just one minor thing: when the widget is renamed, the docs should also be renamed, else they don't match.

Orange/widgets/visualize/owmap.py
```diff
@@ -0,0 +1,837 @@
...
+
+        self.closeContext()
+
+        if data is None:
+            return self.clear()
+
+        all_vars = list(chain(self.data.domain, self.data.domain.metas))
+        continuous_vars = [var for var in all_vars if var.is_continuous]
+        discrete_vars = [var for var in all_vars if var.is_discrete]
+        primitive_vars = [var for var in all_vars if var.is_primitive]
```
nikicc: This filtering doesn't seem to work — i.e. `primitive_vars` is the same as `all_vars` even when StringVariables are present. Isn't `is_primitive` a method rather than a property?

nikicc:
On a side note, should jittering really be recalculated every time a map is moved? For example use 5% jittering and zoom quite closely. Then drag a map slightly up and all points will jump to their new "jittered" location. Could we keep the previous "jittered" location to remove the effect of points jumping around?

nikicc:
The jumping effect is even more disturbing when changing size of the points.

BlazZupan:
.xz files are not readable on OSX. please replace with something more standard that can will be read on all platforms.

kernc:
> .xz files are not readable on OSX. please replace with something more standard that can will be read on all platforms.

I don't think this PR is at issue. Fire up your fav editor and observe how you can load that file no problem.

kernc:
> should jittering really be recalculated every time a map is moved?

Addressed.

PR #1738, opened 2016-11-10
[ENH] OWTranspose: Add a new widget

VesnaT:
##### Issue

Transpose widget

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


Orange/widgets/data/owtranspose.py
```diff
@@ -0,0 +1,117 @@
...
+            if len(self.class_model) else "None"
+        self.report_items(
+            "", [("Create feature names from", self.feature_names_column),
+                 ("Class variable as", class_var)])
+        if self.data:
+            self.report_data("Data", self.data)
+
+
+if __name__ == "__main__":
+    from PyQt4.QtGui import QApplication
```
janezd: AnyQt?

ajdapretnar:
Two comments:
- The widget doesn't show class variable to retrieve feature names by, but it does the retrieval. In other words, iris-setosa appears in feature name, but it doesn't appear as a drop-down in the widget.
- Fails on sparse data. Well, the widget itself doesn't fail, Data Table does, which makes sense, but I think a warning would be nice ('Can't transpose sparse data.')

Big thumbs up for the widget! 👍

Orange/widgets/data/owtranspose.py
```diff
@@ -0,0 +1,107 @@
...
+            btnLabels=["Generic", "From meta attributes"])
+
+        self.feature_model = DomainModel(
+            order=DomainModel.METAS, valid_types=StringVariable,
+            alphabetical=True)
+        self.feature_combo = gui.comboBox(
+            box, self, "feature_names_column", orientation=Qt.Horizontal,
+            labelWidth=100, sendSelectedValue=True, callback=self.apply,
+            contentsLength=12, enabled=self.feature_type)
+        self.feature_combo.setModel(self.feature_model)
```
janezd: Indent the combo like this:

```
        self.feature_combo = gui.comboBox(
            gui.indentedBox(
                box, gui.checkButtonOffsetHint(self.feature_radio.buttons[0])),
            self, "feature_names_column", orientation=Qt.Horizontal,
            labelWidth=100, sendSelectedValue=True, callback=self.apply,
            contentsLength=12, enabled=self.feature_type)
```

Orange/widgets/data/owtranspose.py
```diff
@@ -0,0 +1,107 @@
...
+        self.feature_type = int(bool(len(self.feature_model)))
+
+    def apply(self):
+        self.clear_messages()
+        transposed = None
+        if self.data:
+            options = dict()
+            if self.feature_type:
+                options["feature_names_column"] = self.feature_names_column
+
```
janezd: Feel free to remove this empty line.

Orange/widgets/data/owtranspose.py
```diff
@@ -0,0 +1,107 @@
...
+    icon = "icons/Transpose.svg"
+    priority = 2000
+
+    inputs = [("Data", Table, "set_data")]
+    outputs = [("Data", Table)]
+
+    resizing_enabled = False
+    want_main_area = False
+
+    settingsHandler = PerfectDomainContextHandler(metas_in_res=True)
```
janezd: Why `PerfectDomainContextHandler`? I prefer `DomainContextHandler` when there is no need to match all attributes, because it matches more domains.

Orange/widgets/data/owtranspose.py
```diff
@@ -0,0 +1,107 @@
...
+    feature_type = ContextSetting(0)
+    feature_names_column = ContextSetting("")
```
janezd: This is `Variable`, not `str`. You should use `None` as default.

Orange/widgets/data/owtranspose.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.data = None
+
+        # GUI
+        box = gui.vBox(self.controlArea, "Feature names")
+        self.feature_radio = gui.radioButtonsInBox(
+            box, self, "feature_type", callback=self._feature_type_changed,
+            btnLabels=["Generic", "From meta attributes"])
```
janezd: `"From meta attribute: "` (singular, colon)

Orange/widgets/data/owtranspose.py
```diff
@@ -45,8 +43,7 @@ def __init__(self):
             gui.indentedBox(
                 box, gui.checkButtonOffsetHint(self.feature_radio.buttons[0])),
             self, "feature_names_column", callback=self._feature_combo_changed,
-            model=self.feature_model, sendSelectedValue=True,
```
janezd: `sendSelectedValue` is inapplicable when the combo uses a model --- which always returns the object from the model.

Orange/widgets/data/owtranspose.py
```diff
@@ -45,8 +43,7 @@ def __init__(self):
...
-            labelWidth=100, contentsLength=12, orientation=Qt.Horizontal)
```
janezd: You don't need this since there is no label.

Orange/widgets/data/owtranspose.py
```diff
@@ -57,20 +54,23 @@ def _feature_combo_changed(self):
         self.apply()
 
     def set_data(self, data):
-        self.closeContext()
+        # Skip the context if the combo is empty: a context with
```
janezd: I spent an hour hunting this down. The problem is that if you take Iris, it sets `feature_names_column` to `None`. The you take zoo. The Iris' context matches for zoo domain since it doesn't require any attribute - hence it will set `feature_names_column` to `None`. For the bug to be harder to track, the combo will not reflect this since it does not contain a `None` value.

This is not a bug in `DomainContextHandler`, but it will appear in all such cases. @astaric, what about preventing a match when no attribute were matched since no matcher were required?

`PerfectDomainContextHandler` would work here. :) But I still think that widgets shouldn't use `PerfectDomainContextHandler` unless unavoidable.

Orange/widgets/data/owtranspose.py
```diff
@@ -57,20 +54,23 @@ def _feature_combo_changed(self):
...
-        self.openContext(data)
+        if self.data is not None and self.feature_model:
+            self.openContext(data)
         self.apply()
 
     def update_controls(self):
         self.feature_model.set_domain(None)
         if self.data:
             self.feature_model.set_domain(self.data.domain)
-            if len(self.feature_model):
```
janezd: `PyListModel` has `__bool__` defined.

Orange/widgets/data/owtranspose.py
```diff
@@ -57,20 +54,23 @@ def _feature_combo_changed(self):
...
-                _names = [m.name for m in self.data.domain.metas if m.is_string]
```
janezd: The beauty of the domain model is that you don't need to do this. Besides, it should use `m`, not `m.name`.

Orange/widgets/data/owtranspose.py
```diff
@@ -57,20 +54,23 @@ def _feature_combo_changed(self):
...
-                self.feature_names_column = _names[0]
-        enabled = bool(len(self.feature_model))
```
janezd: Again, `bool` suffices here.

Orange/widgets/data/owtranspose.py
```diff
@@ -79,11 +79,9 @@ def apply(self):
         self.clear_messages()
         transposed = None
         if self.data:
-            options = dict()
-            if self.feature_type:
-                options["feature_names_column"] = self.feature_names_column
             try:
-                transposed = Table.transpose(self.data, **options)
+                transposed = Table.transpose(
+                    self.data, feature_names_column=self.feature_names_column)
```
janezd: This is a bit wrong, but I wouldn't like to `push -f` over it, and lose the comments I made above. Please change this to .... `feature_names_column = self.feature_type and self.feature_names_column`.

Orange/widgets/data/owtranspose.py
```diff
@@ -101,6 +99,8 @@ def send_report(self):
 
     app = QApplication([])
     ow = OWTranspose()
+    d = Table("iris")
```
janezd: This is the case where it doesn't work. (To see this, add `print(self.feature_names_column)` after `self.openContext()`.)

janezd:
The comments above are meaningless unless you read them in a diff for 99b80e860952d38d2ad2c9e8f4840adbc3e145cc.

Orange/widgets/data/owtranspose.py
```diff
@@ -0,0 +1,108 @@
...
+        value_error = Msg("{}")
+
+    def __init__(self):
+        super().__init__()
+        self.data = None
+
+        # GUI
+        box = gui.vBox(self.controlArea, "Feature names")
+        self.feature_radio = gui.radioButtonsInBox(
+            box, self, "feature_type", callback=self.apply,
```
janezd: One more thing (sorry!): changing the radio button always applies even if the auto-apply is off. Change `callback=self.apply` to `callback=lambda: self.apply()`.

(The problem is that `callback` is bound to the original `self.apply` before `gui.auto_commit` replaces it with the conditional apply.)

janezd:
Icon contains invisible lines to the left, and the file is rather long. The picture contains two smaller tables; larger, less complex symbols work better, imho. Please consider the icon in the attachment.

[Transpose.svg.zip](https://github.com/biolab/orange3/files/617504/Transpose.svg.zip)

janezd:
I added a commit with a fix for auto apply, and took the liberty to change the icon.

PR #1739, opened 2016-11-10
OWHierarchialClustering: Annotation index is context dependent setting

VesnaT:
##### Issue

OWHierarchialClustering: Annotation drop down does not remember selection when domain attribute (or class variable) is selected.

##### Description of changes
Annotation index  Setting is replaced with ContextSetting.
To distinguish between types of passed distance matrix, an additional column is added to domain.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -24,7 +24,7 @@ import pyqtgraph as pg
 
 import Orange.data
 from Orange.data.domain import filter_visible
-from Orange.data import Domain
+from Orange.data import Domain, StringVariable
 import Orange.misc
 from Orange.clustering.hierarchical import \
     postorder, preorder, Tree, tree_from_linkage, dist_matrix_linkage, \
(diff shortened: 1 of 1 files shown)


astaric:
Is this PR still relevant or is it being replaced by #1748? If it is not meant to be merged, please close.

janezd:
I believe that #1748 now has the same effect without "tricking" the domain context handler by storing fake attributes used as flags (which I dislike).

#1748 is a bit longer, but mostly because it simplifies the code related to the model that is touched by the PR anyway.

Another problem with this PR is that it assumes that `items` (that is `self.matrix.row_items`) are an instance of `Table` and thus have a `domain` attribute. This doesn't need to be the case, although probably all widgets currently output such matrices. (Another place with this problem is `_update_labels`, as well as the call `self._set_items(matrix.row_items, matrix.axis)`. If `axis=0`, the data should be read from `col_items` and should already be a list of labels, not `Table`. I'm not touching this, though, and I'm not opening this as an issue until it actually becomes an issue.)

PR #1746, opened 2016-11-11
[ENH] Vizrank indicators and filters

janezd:
Vizrank for two attributes now also shows results in a single column, with comma-separated attribute names.

Vizrank now indicates the visualization quality with a bar at the bottom of each row.

Visualizations can be filtered.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
Try Sieve Diagram on "imports-85" data.
I am guessing (but could be wrong) that missing values result in nan scores which the sorting does not handle correctly.

janezd:
@lanzagar, great catch, thanks. I fixed it.

Trying the same on Mosaic (unexpectedly) seems to work, but the top-ranked combination crashes the widget. I reported that; it doesn't belong to this PR.

lanzagar:
We should do some more testing with missing values for different vizranks (and elsewhere :))
But the stuff in this PR seems good to me, works, and the changes are very nice. So lets merge, start using them, and see if we can spot some more corner cases which can be improved.

PR #1748, opened 2016-11-11
[FIX] Hierarchical clustering: Make annotation a context setting

janezd:
Alternative to #1739.

##### Issue

OWHierarchialClustering: Annotation drop down does not keep context settings.

##### Description of changes

Annotation index is replaced with a context setting that contains an attribute or a string. Context are used only if the data are distances between rows. Otherwise, separate settings for axis=0 and axis=1 are used.

Settings are not migrated: the previous settings didn't work anyway.

##### Includes
- [X] Code changes
- [x] Tests
<strike>- [ ] Documentation</strike>

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -723,10 +723,16 @@ class OWHierarchicalClustering(widget.OWWidget):
     outputs = [("Selected Data", Orange.data.Table, widget.Default),
                (ANNOTATED_DATA_SIGNAL_NAME, Orange.data.Table)]
 
+    settingsHandler = settings.DomainContextHandler()
+
     #: Selected linkage
     linkage = settings.Setting(1)
     #: Index of the selected annotation item (variable, ...)
-    annotation_idx = settings.Setting(0)
+    annotation = settings.ContextSetting("Enumeration")
+    #: Out-of-context setting for the case when the "Name" option is available
+    annotation_if_names = settings.Setting("Name")
+    #: Out-of-context setting for the case with just "Enumerate" and "None"
+    annotation_if_enumerate = settings.Setting("Enumerate")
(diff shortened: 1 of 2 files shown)


VesnaT:
Values in the combo should be the same when placing a widget on canvas (disconected) and after disconnecting data from it.

VesnaT:
I think selection in the combo, when data are distances between columns, should be somehow saved as well.

janezd:
- Empty widget should have `Enumeration` as default.
- If domain has a class, use it as default, otherwise Names, otherwise Evaluation.

astaric:
Would it be possible to add a tests that checks that the settings are stored and retrieved? Both, for row an attribute distances?

janezd:
The widget already implemented the above rules we discussed with @VesnaT: class was chosen if it existed, and Enumeration was preferred over None. The latter didn't work (and an empty widget had None instead of Enumeration) because of the bug in `gui.comboBox`, which is fixed in #1764. Fetch my repo and cherry pick bb88aad862caa4f041f97d3b562d8106a7f09544 if you want to test.

Widget now behaves like in #1739: it has context settings for matrices with attached data, and separate settings for axis=0 and axis=1.

I added tests.

(Note to self: ea30352 is the commit with the previous version.)

VesnaT:
Make a workflow: File (iris) -> Select columns -> Distances -> Hieararhical Clustering.
It works fine. Than use Select Columns to remove Target Variable from domain. After clicking "Send", Hieararhical Clustering crashes.

janezd:
@VesnaT, thanks. Fixed ([lines 997 and 998](https://github.com/biolab/orange3/pull/1748/files#diff-22fbad01e8c84125308a0a7fe72ba0f1R997) in the revised commit) and tested ([test_domain_loses_class](https://github.com/biolab/orange3/pull/1748/files#diff-d1c6a5670057d04f9d1030c5eb881193R100)).

PR #1749, opened 2016-11-13
Add Error Reporting

prasastoadi:
##### Issue
TODO: error reporting

##### Description of changes
Add error reporting

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/classify/owloadclassifier.py
+++ b/Orange/widgets/classify/owloadclassifier.py
@@ -97,9 +97,9 @@ class OWLoadClassifier(widget.OWWidget):
         try:
             classifier = pickle.load(open(filename, "rb"))
         except pickle.UnpicklingError:
-            raise  # TODO: error reporting
+            raise pickle.UnpicklingError("Failed to load '{}'.".format(filename))
         except os.error:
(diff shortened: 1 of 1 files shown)


Orange/widgets/classify/owloadclassifier.py
```diff
@@ -97,9 +97,9 @@ def load(self, filename):
         try:
             classifier = pickle.load(open(filename, "rb"))
         except pickle.UnpicklingError:
-            raise  # TODO: error reporting
+            raise  pickle.UnpicklingError("Failed to load '{}'.".format(filename))
```
jayvdb: Only one space should appear between `raise` and argument, here and below..
Please amend your commit.

janezd:
Thanks for noticing this and for contributing to Orange.

However, the change you've made still raises an exception instead of showing a nicely formatted end-user-readable message, as widgets are supposed to.

The actual fix is short, but requires understanding how error messages in widgets work. For demonstration, I implemented the fix for this bug myself. See the changes in PR #1752.

Can you now find another widget that crashes due to user's fault (like this one) and add a similar fix there?

PR #1751, opened 2016-11-14
OWConfusionMatrix: Fix predicitons order

nikicc:
##### Issue

In confusion matrix, as of #1655 (commit deb8585295eaed414c9ec2a09e511b7e4bb907fc), predictions were appended to the data in incorrect order. The core of the issue is due to the fact that `predictions` from `Results.predicted` are not in the same order as instances in data table.

For example, select only misclassified examples on iris and pass them to a table. Some of the misclassified examples have a predicted class that is the same as the correct one.
screen shot 2016-11-14 at 16 27 06

##### Description of changes
Predictions from `Results.predicted` are first sorted into the order of data instances and only then appended. 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owconfusionmatrix.py
+++ b/Orange/widgets/evaluate/owconfusionmatrix.py
@@ -344,7 +344,13 @@ class OWConfusionMatrix(widget.OWWidget):
             metas = self.data.domain.metas
 
             if self.append_predictions:
-                extra.append(predicted.reshape(-1, 1))
+                # predictions are not in the same order as data (i.e. they are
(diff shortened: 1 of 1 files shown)


nikicc:
Tests seem to fail due to an unrelated issue with RandomForestRegression.

janezd:
@nikicc is there any relation or interaction between this and #1720?

nikicc:
@janezd yes, #1720 also fixes this issues hence I'm closing this PR.

PR #1752, opened 2016-11-14
[FIX] OWLoadClassifier: Show message on unpickling error

janezd:
##### Issue

Load Classifier widget crashes instead of reporting an error when unpickling fails.

##### Description of changes

Show the message.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/classify/owloadclassifier.py
+++ b/Orange/widgets/classify/owloadclassifier.py
@@ -12,6 +12,7 @@ from Orange.widgets.settings import Setting
 from Orange.widgets.utils import stdpaths
 
 from Orange.widgets.classify import owsaveclassifier
+from Orange.widgets.widget import Msg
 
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/classify/owloadclassifier.py
```diff
@@ -96,11 +100,10 @@ def load(self, filename):
         """Load the object from filename and send it to output."""
         try:
             classifier = pickle.load(open(filename, "rb"))
-        except pickle.UnpicklingError:
-            raise  # TODO: error reporting
-        except os.error:
-            raise  # TODO: error reporting
+        except (pickle.UnpicklingError, os.error, EOFError):
```
kernc: `os.error` → `OSError`

janezd: Gladly. It was there before, so I kept it, but I didn't like it either. Fixed.

PR #1755, opened 2016-11-15
WebView: Fix evalJS for scripts ending with comments

nikicc:
##### Issue

When loading multiple JS scripts in WebView and one of the first n-1 ends with a comment syntax error is raised even when the code is syntactically correct. This happens since these scripts are joined with a `;` and hence the first line of the following script is commented out.

##### Description of changes
Join multiple JS scripts with `;\n` instead.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/webview.py
+++ b/Orange/widgets/utils/webview.py
@@ -310,7 +310,8 @@ class _WebViewBase:
             if not self.__is_init and self.__js_queue:
                 return QTimer.singleShot(1, _later)
             if self.__js_queue:
-                code = ';'.join(self.__js_queue)
+                # '/n' is required when the last line is a comment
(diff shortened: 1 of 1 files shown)


Orange/widgets/utils/webview.py
```diff
@@ -310,7 +310,8 @@ def _later():
             if not self.__is_init and self.__js_queue:
                 return QTimer.singleShot(1, _later)
             if self.__js_queue:
-                code = ';'.join(self.__js_queue)
+                # '/n' is required when the last line is a comment
+                code = ';\n'.join(self.__js_queue)
```
kernc: In that case, it would be better to `'\n;'`.

PR #1756, opened 2016-11-16
OWManifoldLearning: Add cosine distance to t-SNE

VesnaT:
##### Issue

The distance was not added before, since there was a bug in sklearn, which is now fixed.
https://github.com/scikit-learn/scikit-learn/pull/7732

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmanifoldlearning.py
+++ b/Orange/widgets/unsupervised/owmanifoldlearning.py
@@ -85,7 +85,7 @@ class ManifoldParametersEditor(QWidget, gui.OWComponent):
 
 
 class TSNEParametersEditor(ManifoldParametersEditor):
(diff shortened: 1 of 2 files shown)


astaric:
The linked scikit-learn issue is marked as a part of the  1.18.1 release. Do we need to bump the required version in requirements file?

VesnaT:
Yes. I'll to that.

lanzagar:
I think this PR is OK now. (tested & works)
But while getting sklearn 0.18.1 I had some problems with compiling it (during pip installation). Packaged sklearn works well, so I don't see a reason to wait. Just writing this down, if anyone gets a chance to double check (I won't be able to test further for a couple of days).

Issue #1758, opened 2016-11-17
Can't save a Classifier that was created with Features that come from a Feature Constructor


lopezco:
## Orange version
3.3.9.dev0+1036e49

## Expected behavior
Be able to save a classifier that was created with features that come from a feature constructor

## Actual behavior
Orange3 crashes  because lambda functions can't be serialized using pickle

**Exception:**```_pickle.PicklingError: Can't pickle <function <lambda> at 0x7fe4d99df1e0>: attribute lookup <lambda> on __main__ failed```

**Module:**```Orange.widgets.classify.owsaveclassifier:84```

**Widget Name:** ```Save Classifier```

**Widget Module:** ```Orange.widgets.classify.owsaveclassifier:116```

**Environment:** ```Python 3.5.2 on Linux 4.4.0-47-generic #68-Ubuntu SMP Wed Oct 26 19:39:52 UTC 2016 x86_64```

**Stack Trace:**
```Traceback (most recent call last):
  File "Orange/widgets/classify/owsaveclassifier.py", line 116, in browse
    self.save(filename)
  File "Orange/widgets/classify/owsaveclassifier.py", line 84, in save
    pickle.dump(self.model, f)
_pickle.PicklingError: Can't pickle <function <lambda> at 0x7fe4d99df1e0>: attribute lookup <lambda> on __main__ failed
```
**Local Variables:**
```
OrderedDict([('f', <_io.BufferedWriter name='/home/jose.lopez/qsdqsdqs.pkcls'>),
             ('filename', '/home/jose.lopez/qsdqsdqs.pkcls'),
             ('self',
              <Orange.widgets.classify.owsaveclassifier.OWSaveClassifier object at 0x7fe4d871f168>)])

```

## Steps to reproduce the behavior
1- Load a file
2- Create some features
3- Add a classifier
4- Try to save ir

## Additional info (worksheets, data, screenshots, ...)
![orange_error](https://cloud.githubusercontent.com/assets/6637618/20387563/fb41e488-acc0-11e6-9bbf-a119fee2260d.png)

kernc:
Thanks for a tidy bug report!

A proposed solution in https://github.com/biolab/orange3/pull/1759.

PR #1759, opened 2016-11-17
OWSaveClassifier: Pickle stuff with dill

kernc:
... which supports (nearly) arbitrary Python objects.
Fixes https://github.com/biolab/orange3/issues/1758

##### Includes
- [X] Code changes
- [x] Tests

### DIFF
--- a/Orange/widgets/classify/tests/test_owloadclassifier.py
+++ b/Orange/widgets/classify/tests/test_owloadclassifier.py
@@ -1,12 +1,14 @@
 # Test methods with long descriptive names can omit docstrings
 # pylint: disable=missing-docstring
 import os
-import pickle
 from tempfile import mkstemp
 
 from Orange.classification.majority import ConstantModel
 from Orange.widgets.classify.owloadclassifier import OWLoadClassifier
 from Orange.widgets.tests.base import WidgetTest
+
+import pickle
+import dill    # Import dill after Orange because patched
 
 
 class TestOWMajority(WidgetTest):
@@ -21,10 +23,11 @@ class TestOWMajority(WidgetTest):
         fd, fname = mkstemp(suffix='.pkcls')
         os.close(fd)
         try:
-            with open(fname, "wb") as f:
-                pickle.dump(clsf, f)
-            self.widget.load(fname)
-            self.assertFalse(self.widget.Error.load_error.is_shown())
+            for pickle_impl in (pickle, dill):
+                with open(fname, 'wb') as f:
+                    pickle_impl.dump(clsf, f)
+                self.widget.load(fname)
+                self.assertFalse(self.widget.Error.load_error.is_shown())
 
             with open(fname, "w") as f:
                 f.write("X")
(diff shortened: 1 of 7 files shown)


astaric:
Can you include some new test that show that dill can pickle lambdas? focus on lambdas that use variables from different scopes.

astaric:
And tests that show that dill can "unpickle" saved classifiers created with pickle.

kernc:
Done.

ajdapretnar:
Getting a crash if using a classifier, built on a different domain. Should get an error. Just writing it down here, not meant as a part of this PR. I'm opening it as a new issue.

p.s. I also got some recurring error (log just wouldn't go away and had to force-close the program), but I cannot reproduce it currently.

Orange/widgets/data/tests/test_owfeatureconstructor.py
```diff
@@ -71,6 +73,27 @@ def test_construct_variables_string(self):
...
+
+        lambda_func = lambda x, LOCAL_CONST=7: \
+            x * LOCAL_CONST * NONLOCAL_CONST * self.CLASS_CONST * GLOBAL_CONST
+
+        def nested_func(x, LOCAL_CONST=7):
+            return x * LOCAL_CONST * NONLOCAL_CONST * self.CLASS_CONST * GLOBAL_CONST
+
+        self.assertEqual(lambda_func(11),
+                         pickle.loads(pickle.dumps(lambda_func))(11))
+        self.assertEqual(nested_func(11),
```
kernc: Interestingly, it remembers the exact values of locals and nonlocals at pickling, so the computed result after unpickling is different. Changing the global after pickling errors with:
```
  File "/home/jk/PycharmProjects/orange3/Orange/widgets/data/tests/test_owfeatureconstructor.py", line 94, in test_lambdas_pickle
    pickled_lambda = pickle.dumps(lambda_func)
  File "/home/jk/Downloads/dill/dill/dill.py", line 256, in dumps
    dump(obj, file, protocol, byref, fmode, recurse)#, strictio)
  File "/home/jk/Downloads/dill/dill/dill.py", line 249, in dump
    pik.dump(obj)
  File "/usr/lib/python3.4/pickle.py", line 412, in dump
    self.save(obj)
  File "/usr/lib/python3.4/pickle.py", line 479, in save
    f(self, obj) # Call unbound method with explicit self
  File "/home/jk/Downloads/dill/dill/dill.py", line 1268, in save_function
    globs = globalvars(obj, recurse=True, builtin=True)
  File "/home/jk/Downloads/dill/dill/detect.py", line 200, in globalvars
    _vars = globalvars(obj.cell_contents, recurse, builtin) or {}
ValueError: Cell is empty
```
It works if a global is defined `global`, but this statement is impossible in lambdas. The difference `global` statement makes is in `LOAD_DEREF` vs `LOAD_GLOBAL` opcodes. I don't know if it is or will become an issue in a practical case.

Orange/widgets/classify/tests/test_owloadclassifier.py
```diff
@@ -1,13 +1,15 @@
...
 # pylint: disable=missing-docstring
 import os
-import pickle
 from tempfile import mkstemp
 
 from Orange.classification.majority import ConstantModel
 from Orange.widgets.classify.owloadclassifier import OWLoadClassifier
 from Orange.widgets.tests.base import WidgetTest
 
+import pickle
```
lanzagar: A comment here would be nice, to explain why this needs to be after Orange imports and not grouped with python imports at the top.

PR #1760, opened 2016-11-18
[FIX] TimeVariable: don't crash Data Table when reloading and Visualize ...

kernc:
##### Issue
The issue was with DataTable widget with Visualize continuous values enabled and reloading a timevariable as numeric and back. Now seems to work.

Fixes https://github.com/biolab/orange3/pull/1734#issuecomment-260641386

	  File "/home/lan/dev/orange3/Orange/widgets/utils/itemmodels.py", line 1112, in data
	  File "/home/lan/dev/orange3/Orange/widgets/utils/itemmodels.py", line 863, in format_dense
		return str(instance[var])
	  File "/home/lan/dev/orange3/Orange/data/variable.py", line 166, in __str__
	  File "/home/lan/dev/orange3/Orange/data/variable.py", line 924, in repr_val
		return str(val)
	  File "/home/lan/dev/orange3/Orange/data/variable.py", line 166, in __str__
	...
	RecursionError: maximum recursion depth exceeded while calling a Python object

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -921,7 +921,7 @@ class TimeVariable(ContinuousVariable):
             return '?'
         if not self.have_date and not self.have_time:
             # The time is relative, unitless. The value is absolute.
-            return str(val)
+            return str(val.value) if isinstance(val, Value) else str(val)
 
         # If you know how to simplify this, be my guest
         seconds = int(val)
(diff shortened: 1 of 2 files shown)


astaric:
Can you perhaps reproduce the issue in a test and include that test in the PR?

kernc:
I can't.

kernc:
~~I can't.~~ Managed. Painful to invest too much time into what I'm eagerly working to replace with pandas. :smile:

PR #1766, opened 2016-11-19
[ENH] Create Class widget

janezd:
In line with my widely known and well established tradition of impeccable pull requests and thoroughly tested code, I am presenting a new widget that creates a class attribute with values based on substrings of the chosen string or discrete attribute. The widget is needed for infraorange and presumably also in bioinformatics.

The widget has zero pylint warnings, was assigned a straight A by radon, and has a 100 % coverage with unit tests of the highest quality. The latter rely on simulated user interaction with the widget, and restrict themselves from testing the implementational details except where appropriate. The abundance of documentation strings explaining even the smallest nuances of the private methods - but without going into unnecessary details - will make the future maintenance of the widget pure joy.

This whole grand endeavour is topped by a carefully crafted icon that complements the widget's amazing user interface.

----

A minor shortcoming that may need to be mentioned is that I can't make the widget properly shrink vertically when lines are removed. Not for lack of trying. Any help appreciated.

I have not written any documentation for the widget catalog so far. This will wait for any suggestions and for @ajdapretnar's help with screenshots.

The code of the widget contains two general classes derived from `Transform` and `Lookup` -- but perhaps not general enough to pollute the namespace of the `transformation` module.

The pull request includes a fix for `Orange.preprocess.transformation.Lookup`, which probably nobody used since it was plain broken. This commit could belong to another pull request, but processing/merging chains of pull requests tends to take too long in my experience. Being broken and useless, the class has two relatives of the same name in other modules, one of which does exactly the same thing as the original `Lookup` did not do.

----

@markotoplak, can you test this on files with hyperspectral data? After this glorious presentation, let's hope the widget doesn't crash on the very first use. ;)

##### Includes

- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/widgets/data/owcreateclass.py
+++ b/Orange/widgets/data/owcreateclass.py
@@ -13,31 +13,52 @@ from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.widget import Msg
 
 
-def map_by_substring(a, patterns):
+def map_by_substring(a, patterns, case_sensitive, at_beginning):
     res = np.full(len(a), np.nan)
+    if not case_sensitive:
+        a = np.char.lower(a)
+        patterns = (pattern.lower() for pattern in patterns)
     for val_idx, pattern in reversed(list(enumerate(patterns))):
-        res[np.char.find(a, pattern) != -1] = val_idx
+        indices = np.char.find(a, pattern)
+        matches = indices == 0 if at_beginning else indices != -1
+        res[matches] = val_idx
     return res
 
 
-class ClassFromStringSubstring(Transformation):
-    def __init__(self, variable, patterns):
+class ValueFromStringSubstring(Transformation):
+    def __init__(self, variable, patterns,
+                 case_sensitive=False, match_beginning=False):
         super().__init__(variable)
         self.patterns = patterns
+        self.case_sensitive = case_sensitive
+        self.match_beginning = match_beginning
 
     def transform(self, c):
         nans = np.equal(c, None)
         c = c.astype(str)
         c[nans] = ""
-        res = map_by_substring(c, self.patterns)
+        res = map_by_substring(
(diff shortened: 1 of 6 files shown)


Orange/widgets/data/owcreateclass.py
```diff
@@ -13,19 +14,48 @@
 from Orange.widgets.widget import Msg
 
 
-def map_by_substring(a, patterns, case_sensitive, at_beginning):
+def map_by_substring(a, patterns, case_sensitive, match_beginning):
+    """
+    Map values in a using a list of patterns. The patterns are considered in
```
ajdapretnar: Map values ~in a~ using a list of patterns.

+ What patterns?

ajdapretnar:
I like this widget a lot, but GUI totally confuses me. The + symbol is very small and renders as cross on Win, instance count odd, widget is a bit long... I'm not sure, it just doesn't lie well with me.
I'd make it more Select-Rows-esque, with X on the right, info box perhaps (?) and so on. And no, it's probably not the best idea either.

As for functionality, I'd like to have sth like Other, Rest or *... Something that says 'everything else belongs to this class'.

Also as for pattern, how about Regex? If not, I'd have a tooltip for 'Pattern' giving an example of use. Will be featured in docs, too.

I'd add Report and have the two bottom buttons as in Feature Constructor.

janezd:
Note to` self` from today's discussion with @ajdapretnar:

Besides what Ajda wrote above:

- equal width of apply and report buttons
- narrower widget
- align the combo with line edits
- placeholder in the first empty pattern that tells that this will cover the remaining instances
- tooltip on counts
- replace the add button with normal `+`; compute the necessary width
- perhaps remove the box title (*Pattern*)

astaric:
Could you also include a help file for the widget? Even if it is just a stub.

PR #1768, opened 2016-11-22
LinkItem: Split long signal names to two lines

VesnaT:
##### Issue

Long signal names should be split to two lines.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
The breaks are now inserted purely based on width and can happen at awkward places. Also it can break in 3 lines which usually looks awkward.

How about just breaking long link names into two parts instead (output & input)?
E.g. after line 526:
```
                if len(text) > 20:
                    text = "{0}\n\u2192 {1}".format(self.__sourceName,
                                                    self.__sinkName)
```

For future consideration, we can take the length of the link into account. But that would complicate the code, so let's go with something simpler first.

ales-erjavec:
Please also see: VesnaT/orange3#11

lanzagar:
I guess the future is here already :)
The solution by @ales-erjavec seems worth the couple of extra lines.

PR #1507, opened 2016-08-10
[WIP] Created __repr__ functions for learners + preprocessors

Ameobea:
This commit creates `__repr__` functions for the majority of Orange's learner and preprocessor classes.  It uses standardized and generic functions wherever possible.  These changes pave the way for my code generator (PR upcoming), which makes extensive use of these functions to create runnable Python scripts based on the contents of the workflow.

I moved some data processing functions out of widget code and into more appropriate directories, updating the widget code as appropriate.  I converted some functions to classes in order to make them more similar to others and created **repr** functions for them as well.  Their usages were updated accordingly.  I then squashed all commits into a single commit.

All tests are passing (except those that weren't passing before) and the linter doesn't show anything significant caused by this commit.  As far as I can tell, this commit is ready to merge and completely working.  If any changes need to be made, please let me know.

EDIT: I've changed the title to WIP due to issues between the results of tests on my computer and those returned by Travis.

PR #1449, opened 2016-07-12
[WIP] Exporting OWS to Python code, RFC please

Ameobea:
## Project Overview

I have been working over the past few weeks to implement code generation for Orange workflows.  I believe my work is currently at a decision point and that my method of code generation is set up to the point where its functionality is fully implemented.  Before moving forward to the bulk of the work in creating code generator functions for as many of the individual widgets as possible, I wanted to check in with the Orange community at large and get feedback on what I've made so far.

I've been corresponding with @kernc and I've done my best to change my code to match his feedback as well as I could without changing my base design, but I want to be sure that what I have is acceptable before continuing work.
## Code Generator Design

The code generator that I've build is designed as a system by which code from widgets and the Orange library can be extracted, modified, and formatted before being linked with other widgets and inserted into the output script.  It aims to allow for code to be added to the output with as little management code as possible while at the same time keeping duplicated code to as low a level as possible.  

It provides several helper functions that do things such as generate import statements, inline external functions, register widget outputs, and remove/replace strings and lines in the output.  The main code generator itself is held in `Orange/widgets/utils/codegen.py` and an instance is created for every widget.  
## Example

I use this workflow for testing of the code generator:

And here is the code that is generated for it:
https://ameo.link/u/bin/2mq
## Code Generation Process

The code generation process starts with the conversion of the scheme into a topographically sorted linear list of widgets ordered by the order in which they should be processed.  This ensures that all widget inputs are generated before being used.  The code generator for each of the widgets in the list is then initialized in order and the result is added to the output string.  That string is then written to file and can be run as a regular Python script.
## Closing Remarks

This WIP PR represents the current result of my efforts so far.  I undertook this project under the instruction and funding of @Pelonza to create a tool for him to use in the classroom with his students.  At the same time, I absolutely intend to create a result that meets all the requirements for merging and will continue to work closely with the Orange team to produce a result that everyone is happy with.

If the current code generator as a whole is felt to be unsatisfactory or an alternative that would do the job better is agreed upon, I'll do what's necessary to make it happen.

PR #1771, opened 2016-11-22
Common __repr__ for common types

kernc:
##### Issue
Our types all around.

##### Description of changes
Makes most of our types repr-able, as recommended by [python docs](https://docs.python.org/3/reference/datamodel.html#object.__repr__) and as needed eventually by https://github.com/biolab/orange3/pull/1449.
A continuation of https://github.com/biolab/orange3/pull/1507.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

This "traits" programming is quite clean. I reckon we shall soon also have types that are `_Copyable` and `_Enumerable`!

Would _really_ appreciate a speedy review!


Ameobea:
I'm glad that this is finally getting the attention it deserves.  I was not cut out for the level of expertise necessary to make this happen and on top of that got very distracted by other things once summer ended.  

In case my work on code generation could be of any use in this new effort, here's the branch in which I was working (this is other work than my efforts in `__repr__` generation): https://github.com/Ameobea/orange3/tree/code-generation

Sorry I wasn't able to make this happen, and good luck with your efforts.

kernc:
Oh, you did great. If it weren't for you, we likely wouldn't be looking at anything like this at all. :wink:

Orange/base.py
```diff
@@ -262,13 +294,14 @@ def fit(self, X, Y, W=None):
...
-    def __str__(self):
-        return '{} {}'.format(self.name, self.params)
-
     def __repr__(self):
-        return '{}({})'.format(type(self).__name__,
-                               ", ".join("{}={}".format(k, v)
-                                         for k, v in self.params.items()))
+        # In addition to saving values onto self, SklLearners save into params
+        vars = self.__dict__
+        self.__dict__ = dict(vars, **self.params)
```
astaric: This looks alot like patching the self.__dict__. Any specific reason why not using [patch.object](https://docs.python.org/3/library/unittest.mock.html#patch-object)?

Orange/classification/logistic_regression.py
```diff
@@ -36,7 +36,7 @@ def coefficients(self):
 class LogisticRegressionLearner(SklLearner, _FeatureScorerMixin):
     __wraps__ = skl_linear_model.LogisticRegression
     __returns__ = LogisticRegressionClassifier
-    name = 'logreg'
+    name = 'logit'
```
astaric: Why?

Orange/data/filter.py
```diff
@@ -379,16 +375,11 @@ def __str__(self):
             return "{} is defined".format(column)
         return "invalid operator"
 
-    __repr__ = __str__
-
-
-    # For PyCharm:
-    Equal = NotEqual = Less = LessEqual = Greater = GreaterEqual = 0
```
astaric: I guess this was here to allow IDEs to suggest these names for code completion. Does this work without the dummy members in the current version of PyCharm?

Orange/preprocess/continuize.py
```diff
@@ -1,12 +1,13 @@
 from Orange.data import ContinuousVariable, Domain
 from Orange.statistics import distribution
+from Orange.util import Reprable
 from .transformation import Identity, Indicator, Indicator1, Normalizer
 from .preprocess import Continuize
 
-__all__ = ["DomainContinuizer", "MultinomialTreatment"]
+__all__ = ["DomainContinuizer"]
```
astaric: Why was this removed?

Orange/preprocess/preprocess.py
```diff
@@ -419,7 +502,7 @@ class PreprocessorList:
         A list of preprocessors.
     """
 
-    def __init__(self, preprocessors):
+    def __init__(self, preprocessors=[]):
```
astaric: lint probably complains that having mutable default values is not a good idea.

Orange/tests/test_classification.py
```diff
@@ -8,11 +8,15 @@
 
 import traceback
 import numpy as np
+import Orange
```
astaric: Is this used anywhere?

Orange/widgets/data/owdatasampler.py
```diff
@@ -219,6 +220,7 @@ def commit(self):
         self.send("Data Sample", sample)
         self.send("Remaining Data", other)
 
+
```
astaric: One empty line is preferred between methods.

Orange/widgets/data/owpreprocess.py
```diff
@@ -762,8 +700,8 @@ def __init__(self, parent=None, **kwargs):
         self.__scalecb.activated.connect(self.edited)
 
     def setParameters(self, params):
-        center = params.get("center", Scale.CenterMean)
-        scale = params.get("scale", Scale.ScaleBySD)
+        center = params.get("center", Scale.CenterMean - 1)
```
astaric: Why - 1?

Orange/widgets/data/owpreprocess.py
```diff
@@ -762,8 +700,8 @@ def __init__(self, parent=None, **kwargs):
...
+        scale = params.get("scale", Scale.ScaleBySD - 1)
```
astaric: Why -1?

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -194,6 +194,7 @@ def check_data(self):
                 self.valid_data = True
         return self.valid_data
 
+
```
astaric: One line between methods is usually sufficient.

Orange/preprocess/fss.py
```diff
@@ -89,6 +89,7 @@ def __call__(self, data):
                                     data.domain.class_vars, data.domain.metas)
         return data.from_table(domain, data)
 
+
```
astaric: One empty line is usually enough.

astaric:
Another note, do we state anywhere (docs, github, wiki) that the strings returned by __repr__ should be executable? If not, should we, and where?

Orange/base.py
```diff
@@ -72,14 +74,46 @@ def preprocess(self, data):
...
         return True
 
+    @property
+    def name(self):
+        """Return a short name derived from Learner type name"""
+        try:
+            return self.__name
+        except AttributeError:
+            name = self.__class__.__name__
+            assert name.endswith('Learner'), \
```
kernc: To enforce some conformity. Now stripped.

Orange/classification/logistic_regression.py
kernc: Number of results for
https://www.google.si/search?num=2&q=logit
vs.
https://www.google.si/search?num=2&q=logreg

Been bugging me for a while.

Orange/preprocess/continuize.py
kernc: Unused.

Orange/tests/test_classification.py
kernc: Seems to be via `Orange.evaluation.AUC` below. I don't know how that worked before, or how Python's imports work, obviously. :confused:

Orange/widgets/data/owpreprocess.py
kernc: The enum values are no longer comparable to integers, which IMHO is a good thing. Replaced with two short utility functions.

kernc:
Comments addressed, `__repr__` mentioned. Thanks.

Orange/classification/logistic_regression.py
lanzagar: a) the number of results for "trump" is even greater, lets not call it that!
More seriously:
b) the number of results for "logistic regression" is over 5x that of "logit"
Even more seriously:
c) while I don't like nitpicking over terminology (especially when not very well defined), I think logit is more frequently used for the function, and logistic regression for the predictive model (which uses the function). Wikipedia has "logit regression" as an alternative, but I don't think it is used as much. And logit shows an article which has this at the top: "For the basic regression technique that uses the logit function, see logistic regression"

Anyway as long as we are not renaming the class to LogitRegression the name change gets a -1 from me. But, you could argue for a non abbreviated name if that bothers you...

Orange/tests/test_classification.py
astaric: Could you replace with explicit import?

Orange/preprocess/continuize.py
```diff
@@ -8,15 +8,18 @@
...
+                 transform_class=False, **kwargs):
         self.zero_based = zero_based
         self.multinomial_treatment = multinomial_treatment
         self.transform_class = transform_class
 
-        return self if data is None else self(data)
+        # Backward-compat polite notice
+        if isinstance(zero_based, (Table, Domain)) or 'data' in kwargs:
+            raise TypeError('DomainContinuizer no longer accepts data in the '
+                            'constructor. Instea, first make an object instance, '
```
astaric: *Instead

Orange/preprocess/util.py
```diff
@@ -0,0 +1,16 @@
...
+class _RefuseDataInConstructor:
+    # TODO: drop this as soon as possible
+    def __new__(cls, *args, **kwargs):
+        if (args and isinstance(args[0], Table) or
+                isinstance(kwargs.get('data'), Table)):
+            warnings.warn(
+                '%s no longer accepts data in the constructor. Instead, '
+                'first make an instance, i.e. %s(), then call it with data.'
+                % (cls, cls), OrangeDeprecationWarning, stacklevel=2)
+        return super().__new__(cls)
```
astaric: IMHO DeprecationWarnings are used to point that a feature/method/class is going away and that a developer should modify the code as it will not work with the future versions of the library. Here, a warning is used to point that the code does not work any more.

I would rather have a ValueError raised (with the same message), or keep the old behaviour working until some future version (probably mentioned in the warning).

Orange/projection/pca.py
```diff
@@ -154,8 +154,7 @@ def compute(self, data, pca_space):
         return pca_space[:, self.feature]
 
 
-class RemotePCA:
-    def __new__(cls, data, batch=100, max_iter=100):
+def remotePCA(data, batch=100, max_iter=100):
```
astaric: Please update the import in widget as well.
https://github.com/biolab/orange3/blob/master/Orange/widgets/unsupervised/owpca.py#L154

Orange/preprocess/util.py
kernc: > or keep the old behaviour working until some future version

That's what I meant. :blush:  Done. Thanks.

Orange/projection/pca.py
kernc: Reverted because it doesn't play along with Orange remote.

Orange/widgets/unsupervised/owpca.py
```diff
@@ -151,9 +151,9 @@ def start(self):
         else:
             self.address = self.addresstext.text()
             with remote.server(self.address):
-                from Orange.projection.pca import RemotePCA
+                from Orange.projection.pca import remotePCA
```
astaric: This should be reverted as well

Orange/widgets/data/owcontinuize.py
```diff
@@ -346,17 +347,15 @@ def normalize_by_sd(var, data_or_dist):
     return normalized_var(var, mean, 1 / sd)
 
 
-class DomainContinuizer:
-    def __new__(cls, data=None, zero_based=True,
-                multinomial_treatment=Continuize.Indicators,
-                continuous_treatment=Continuize.Leave,
-                class_treatment=Continuize.Leave):
-        self = super().__new__(cls)
+class DomainContinuizer(Reprable):
```
astaric: Should this class also be backwards compatible (with a warning)?

Orange/widgets/unsupervised/owpca.py
kernc: This creeped in in the wrong place. Thanks.

PR #1772, opened 2016-11-23
Fix Remove Constant Filter for Sparse Data

nikicc:
##### Issue

When calling `Orange.preprocess.remove.remove_constant` on sparse data columns containing just one different value (besides the undefined ones) are wrongly interpreted as constant columns and are hence removed.

This can easily be reproduced in Text add-on using Twitter widget. Load tweets for two authors and pass them through Bag of Words widget. Then connect this to Select Rows and as condition select one of the authors first and then the second one. Observe the number of retained features and you can see that most of them are thrown away. For example, out of initial 1k+ features for the first author I get ~20 and for the second ~30.

##### Description of changes
Changes introduced on three ends:
* `Orange.statistics.utils`: `countnans` and `bincount` modified to support sparse arrays.
* `Table._compute_distributions` is corrected to also count undefined values for sparse data (before it just returned 0).
* `Orange.preprocess.remove_constant` function adopted for sparse data. Precisely, `has_at_least_two_values` now also accounts for missing data and counts them as one of the values. For example columns of `sp.sparse.eye(4)` should not be treated as constant columns any more.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


lanzagar:
The changes are totally justified for sparse matrices.
For dense data, I am not so sure if we want to consider columns with a single value and missing values as constant and let them be removed or not. This PR changes the existing rules, so we should be a bit more careful.
However, I have not found any cases where we remove constant columns that would have problems after the change. Unless someone comes up with a reason not to, I am fine with the change to dense data and merging this as is.

nikicc:
@lanzagar please check revised. I modified it to remove columns like `np.array([1, np.nan, 1])` for dense matrices as we decided on the meeting and prettify the tests.

PR #1774, opened 2016-11-24
[FIX] Fix varius deprecation (and other) warnings

ales-erjavec:
##### Issue

Current output from test shows quite a lot of warnings, which means that new ones are hard to spot.

##### Description of changes

Fix most of the easily avoidable warnings.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/regression/tree.py
+++ b/Orange/regression/tree.py
@@ -101,8 +101,9 @@ class TreeLearner(Learner):
             if score == 0:
                 return  REJECT_ATTRIBUTE
             score *= non_nans / len(col_x)
-            branches = (col_x > cut).astype(int)
-            branches[np.isnan(col_x)] = -1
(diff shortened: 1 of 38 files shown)


Orange/regression/tree.py
```diff
@@ -101,8 +101,11 @@ def _score_cont():
             if score == 0:
                 return  REJECT_ATTRIBUTE
             score *= non_nans / len(col_x)
-            branches = (col_x > cut).astype(int)
+
+            with np.errstate(invalid="ignore"):  # NaN in col_x handled later
```
lanzagar: Wouldn't it be cleaner to initialize branches to -np.ones and set only the non-nan values after that.
Maybe it's just me, but I prefer not triggering errors to declaring which ones to ignore.

lanzagar:
Great stuff 👍

PR #1783, opened 2016-11-29
[FIX] OWBoxPlot: Fixups

VesnaT:
##### Issue

The widget crashed if continuous variables were only in metas. 

##### Description of changes
To reproduce, use File (iris) -> Select columns -> Feature constructor -> Box plot, to move continuous features to metas and to create a new StringVariable, which causes data.metas.dtype = object.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/domain.py
+++ b/Orange/data/domain.py
@@ -283,25 +283,25 @@ class Domain:
         except KeyError:
             raise ValueError("'%s' is not in domain" % var)
 
-    def has_discrete_attributes(self, include_class=False):
+    def has_discrete_attributes(self, include_class=False, include_metas=False):
         """
-        Return `True` if domain has any discrete attributes. If `include_class`
-                is set, the check includes the class attribute(s).
+        Return `True` if domain has any discrete attributes. If
+        `include_class` is set, the check includes the class attribute(s). If
+        `include_metas` is set, the check includes the meta attributes.
         """
-        if not include_class:
-            return any(var.is_discrete for var in self.attributes)
-        else:
-            return any(var.is_discrete for var in self.variables)
+        vars = self.variables if include_class else self.attributes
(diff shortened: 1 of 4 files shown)


Orange/data/domain.py
```diff
@@ -283,25 +283,35 @@ def index(self, var):
...
         """
-        Return `True` if domain has any discrete attributes. If `include_class`
-                is set, the check includes the class attribute(s).
+        Return `True` if domain has any discrete attributes. If
+        `include_class` is set, the check includes the class attribute(s). If
+        `include_metas` is set, the check includes the meta attributes.
         """
         if not include_class:
-            return any(var.is_discrete for var in self.attributes)
+            attributes = list(self.attributes) if not include_metas \
```
lanzagar: a) Isn't the conversion to lists unnecessary?
b) This whole function can be shortened even further, e.g.:
```
vars = self.variables if include_class else self.attributes
vars += self.metas if include_metas else ()
return any(var.is_discrete for var in vars)
```

Orange/data/domain.py
```diff
@@ -283,25 +283,35 @@ def index(self, var):
...
+    def has_continuous_attributes(self, include_class=False, include_metas=False):
         """
         Return `True` if domain has any continuous attributes. If
-        `include_class` is set, the check includes the class attribute(s).
+        `include_class` is set, the check includes the class attribute(s). If
+        `include_metas` is set, the check includes the meta attributes.
         """
         if not include_class:
-            return any(var.is_continuous for var in self.attributes)
+            attributes = list(self.attributes) if not include_metas \
```
lanzagar: Same as for discrete above.

VesnaT:
This should be done in another pull request, since it requres some changes with Context settings.

PR #1785, opened 2016-11-29
OWBoxPlot: Test saved selection

VesnaT:
##### Issue

Selection is saved to Context setting and therefore should be tested

##### Includes
- [ ] Code changes
- [X] Tests
- [ ] Documentation


lanzagar:
Selections (in general, not just here) should probably be schema-only settings.
They should definitely be saved with workflows, and it isn't bad to remember them when changing inputs in a widget. But once the widget is removed and a new one created, the selections should probably be reset.

PR #1799, opened 2016-12-02
Small changes here and there

kernc:
Random stuff. See individual commits.

The commit that broke WebviewWidget has been moved to https://github.com/biolab/orange3/pull/1816.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,15 +23,16 @@ This version of Orange requires Python 3.4 or newer. To build it and install
 it in a development environment, run:
 
     # Install some build requirements via your system's package manager
-    sudo apt-get install virtualenv git python3-dev g++ gfortran \
-                         libblas-dev liblapack-dev libatlas-base-dev
+    sudo apt-get install virtualenv git build-essential
 
     # Also install Qt dependencies for the GUI
     sudo apt-get install python3-pyqt4
+    # or if python version is >= 3.5
+    #  pip install pyqt5 
 
-    # Create a separate Python environment for Orange and its dependencies,
-    # and make it the active one
+    # Create a separate Python environment for Orange and its dependencies ...
     virtualenv --python=python3 --system-site-packages orange3venv
+    # ... and make it the active one
     source orange3venv/bin/activate
 
     # Clone the repository and move into it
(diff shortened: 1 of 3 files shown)


janezd:
The purpose of the PR's description is to explain what the PR is about.

- Please replace "Fixups" in the title with "[FIX]". Then, provide a title.
- Change "Random stuff" to something better than *foo*.
- Explain what is meant by "One of the commits breaks some WebviewWidget users like add-ons". It's cryptic. What is a "WebviewWidget user"? An actual person or another widget, possibly from an add-on? Does one of *these* commits do break it? Please say which and how and why and what to do about it.

If the whatever this PR fixes can and should be tested, please add tests.

It would seem that this PR fixes at least three totally unrelated things: dependencies, something related to webview and results of several itemmodels' methods. Should it be broken into multiple PRs?

Please either correct me if I'm wrong or improve this PR according to what meets your criteria for an acceptable PR.

README.md
```diff
@@ -24,7 +24,8 @@ it in a development environment, run:
 
     # Install some build requirements via your system's package manager
     sudo apt-get install virtualenv git python3-dev g++ gfortran \
-                         libblas-dev liblapack-dev libatlas-base-dev
+                         libblas-dev liblapack-dev libatlas-base-dev \
+                         libssl-dev libffi-dev  # ...
```
astaric: Which package needs these libraries?

kernc:
@janezd Thanks, you're right. I have batched the most offending commit along with others related into https://github.com/biolab/orange3/pull/1816. This now contains three totally unrelated things, any of which by itself seems too light to bear its own whole PR. They're small and unobtrusive changes; kind of something one would just push herself if it weren't for policy. I don't have a problem with that.

README.md
kernc: No idea, but I have them installed. Anyway, I think these instructions are obsolete. Will update.

ajdapretnar:
5a0287f is probably what I need to review, but I have no idea how...

PR #1803, opened 2016-12-07
Table.from_table: match_type handle python's lists

nikicc:
##### Issue

When using Feature Constructor to add discrete features to sparse array the code crashed inside `Table.from_table` in `match_type` function. Most easily reproducible on Text addon. Load some tweets and use Feature Constructor to create a discrete variable if some word is present in the content.
##### Description of changes
`match_type` was modified to assure the input, when dense, is a numpy array. That is, if python's list is provided (as is the case for Feature Constructor) it is transformed to a numpy array.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -299,6 +299,7 @@ class Table(MutableSequence, Storage):
                 if is_sparse == sp.issparse(x):
                     return x
                 elif is_sparse:
+                    x = np.asarray(x)
                     return sp.csc_matrix(x.reshape(-1, 1).astype(np.float))
                 else:
                     return np.ravel(x.toarray())


ajdapretnar:
I'm experiencing a strange behaviour. Feature Constructor now works, but outputs feature name instead of the set value.
E.g.
'smiley'
`0 if ':)' in Content else 1`
outputs a column with an occasional 'smiley' entry. 'Smiley' seems to be the entry for False.

nikicc:
I suppose that's because for a corpus the default value for X is an empty sparse matrix. Hence, when you add a discrete feature it is also stored in sparse format. 

About why you only see `Smiley` I'm not completely sure. I guess this is how we show binary sparse data in basket format?

ajdapretnar:
A screenshot might better explain my problem:

wrong

nikicc:
@ajdapretnar I checked and that's just how we show sparse data. Noting wrong here.

What's bothering you, I assume, is that this is in sparse format at all, while it should probably be in dense. This seems like an unrelated issue which should be fixed on Text.

Issue #1794, opened 2016-12-02
Some example workflows included in Orange do not open


ales-erjavec:
Some Example workflows from Orange do not load

* 'Classification Tree'
```UnknownWidgetDefinition('Orange.widgets.classify.owclassificationtreegraph.OWClassificationTreeGraph',)```

* 'Feature Ranking'
```
Traceback (most recent call last):
  File "/Users/aleserjavec/workspace/orange3/Orange/canvas/scheme/widgetsscheme.py", line 454, in create_widget_instance
    widget.__init__()
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/data/owrank.py", line 260, in __init__
    if self.headerState[2] is not None:
IndexError: tuple index out of range
```

* 'Where are Missclassifications?'
```
Traceback (most recent call last):
  File "/Users/aleserjavec/workspace/orange3/Orange/canvas/scheme/widgetsscheme.py", line 822, in process_signals_for_widget
    handler(*args)
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/evaluate/owconfusionmatrix.py", line 225, in set_results
    prev_sel_learner = self.selected_learner.copy()
AttributeError: 'int' object has no attribute 'copy'
```

##### Orange version

3.3.9.dev0+...

##### Expected behavior

Example workflows load

##### Actual behavior

Example workflows do not load

##### Steps to reproduce the behavior

Menu -> Help -> Examples

PR #1804, opened 2016-12-08
[FIX] TreeGraph: Compatibility with old schemas

astaric:
##### Issue
Classification tree example does not open (#1794)

##### Description of changes
Declare which widgets are replaced with the new TreeGraph widget.

### DIFF
--- a/Orange/widgets/visualize/owtreeviewer.py
+++ b/Orange/widgets/visualize/owtreeviewer.py
@@ -163,6 +163,11 @@ class OWTreeGraph(OWTreeViewer2D):
     target_class_index = ContextSetting(0)
     regression_colors = Setting(0)
 
+    replaces = [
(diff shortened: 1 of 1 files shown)


astaric:
@ales-erjavec after this fix is applied, the schema shows all widgets, but the signal connecting the Tree widget and the TreeViewer is not created, as the name of the input has changed from Classification Tree to Tree.

Do we have a mechanism that would allow us to declare mapping between old and new signal names? (the way replaces works for widget paths)

ales-erjavec:
> Do we have a mechanism that would allow us to declare mapping between old and new signal names? (the way replaces works for widget paths)

No, not yet. Work in progress at [wip/channel-ids](https://github.com/ales-erjavec/orange3/tree/wip/channel-ids)

PR #1813, opened 2016-12-09
Orange.data: Clean up the namespace

lanzagar:
##### Issue

Orange.data namespace included a lot of junk.

##### Description of changes
Define `__all__` in data.io.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -29,6 +29,9 @@ from Orange.data import (
 from Orange.util import Registry, flatten, namegen
 
 
+__all__ = ["Flags", "FileFormat"]
+
+
 _IDENTITY = lambda i: i
 
 


Orange/data/io.py
```diff
@@ -29,6 +29,9 @@
 from Orange.util import Registry, flatten, namegen
 
 
+__all__ = ["Flags"]
```
kernc: I don't think that's ok. The module defines a lot more "public" stuff than just `Flags`. Perhaps rather make `import *` in _Orange.data.\_\_init\_\_.py_ explicit?

astaric: FileFormat might be considered public as well, as someone might use it to define their own readers, but what else? Reader implementations are not supposed to be used directly, they are accessible through FileReader.get_reader.

lanzagar: Added `FileFormat`. As said, the individual readers are really just current implementations that should not be used directly elsewhere, so I don't think they belong in `__all__`. And they can be imported from data.io if needed (for exceptional cases only).

kernc: >  but what else?

@astaric always more than one thinks.

https://github.com/biolab/orange3-text/pull/176/files

PR #1814, opened 2016-12-09
Box plot: Fix labels outside of view

### LINKED ISSUE #1737: Box Plot labels fall out of view
##### Orange version
3.3.9.dev0+7dca993

##### Expected behavior
All important information in Box Plot are in place (visible).

##### Actual behavior
Sometimes labels and even visualizations fall out of view.

Housing.tab with Box Plot. CHAS attribute.

box-plot-fallout

janezd:
Fixes #1737.

Widget computes scaling based on positions of labels and lines; as I recall, I did this to avoid uncontrolled scaling of text and other elements. This complicates everything, but works ... except that the scene was to wide and the widget cropped it to avoid scrolling. If instead of cropping I try scaling, the view get a lot of white space, scale is too small and everything looks ugly.

##### Description of changes

Increases the width by 60.

It looks OK on Mac. @ajdapretnar, can you check it on Windows?

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -646,7 +646,7 @@ class OWBoxPlot(widget.OWWidget):
 
         bv = self.box_view
         viewrect = bv.viewport().rect().adjusted(15, 15, -15, -30)
-        self.scale_x = scale_x = viewrect.width() / (gtop - gbottom)
+        self.scale_x = scale_x = (viewrect.width() - 60) / (gtop - gbottom)
 
         # In principle we should repeat this until convergence since the new
         # scaling is too conservative. (No chance am I doing this.)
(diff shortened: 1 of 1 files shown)


astaric:
macOS on external monitor (not-retina)

screen shot 2016-12-12 at 09 16 58

ajdapretnar:
Not quite there on Windows either, but better than before.

box-plot

lanzagar:
Linux has same issues.
But in addition to elements falling out of viewport, there seem to be factual errors in box plot, e.g. CHAS variable shows median at 1, when it should be 0 (lots of 0 values, few 1s)

janezd:
Doesn't work. We may need to expect the worst case (the left-most label is extended to the left, the right-most to the right) in computation of scaling.

PR #1817, opened 2016-12-13
Merge OWRandomForest

pavlin-policar:
##### Issue
Merge RF Widget to handle both classification and regression problems.

##### Description of changes
Straightforward merge. Keep widgets in classify and regression categories until all the learners are merged. No noticeable difference except that both learners can handle any data type.

I've also had to implement a `RandomForestModel` class since the widget now has to output a model and not a classifier or regressor. I could also merge this into existing `RandomForest` class if you think that would be better.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/model/owrandomforest.py
```diff
@@ -27,7 +27,8 @@ class OWRandomForest(OWBaseLearner):
     index_output = settings.Setting(0)
 
     class Error(OWBaseLearner.Error):
-        not_enough_features = Msg("Insufficient number of attributes ({})")
+        not_enough_features = Msg(
+            'Number of considered attributes is too high (max {})!')
```
janezd: I don't know who requested this change. It could have been me. I feel like an old professor. Anyway, I don't like it any better than the old message. I find it equally uninformative. :) 

What about `"Number of considered attributes exceeds the number of attributes"`? It's long, but the widget is probably wide enough.

Orange/modelling/base.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, preprocessors=None, **kwargs):
         self.kwargs = kwargs
         # Make sure to pass preprocessor params to individual learners
         self.kwargs['preprocessors'] = preprocessors
-        self.problem_type = None
+        self.problem_type = self.CLASSIFICATION
```
janezd: Why is this necessary? Haven't we discussed that it's better to leave this undefined until you get the data? (Apologies if we decided otherwise and I forgot it.)

Orange/widgets/model/owrandomforest.py
pavlin-policar: I looked at the message and found it strange so I tried to change it to something a bit more informative. I tried your message, but sadly it's just a bit too long. What about `"Considering more attributes than available"` or `"Considering more attributes than total number of attributes"`?

Orange/modelling/base.py
pavlin-policar: I've now fixed this after quite a bit of headache,  but now it is testable.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -177,7 +177,6 @@ def update_learner(self):
             self.learner.use_default_preprocessors = True
         if self.learner is not None:
             self.learner.name = self.learner_name
-        self.learner.use_default_preprocessors = True
```
kernc: Wasn't this supposed to be [enabled in widgets](https://github.com/biolab/orange3/pull/1657#issuecomment-263121263)?

kernc: Never mind. I'm blind. :see_no_evil:

pavlin-policar: I assume this means everything's ok?

janezd:
I checked it again and I'd merge it, except that it needs rebasing to solve conflicts. It looks simple (but I still don't dare doing it myself).

pavlin-policar:
Done.

PR #1818, opened 2016-12-13
Merge OWTree

pavlin-policar:
##### Issue
Merge Tree learner Widget to handle both classification and regression problems.

##### Description of changes
Fairly straightforward merge. Keep widgets in classify and regression categories until all the learners are merged. No noticeable difference except that both learners can handle any data type.

We did discuss that it would be a good idea to move `Orange.tree` into `Orange.modelling.tree` but I haven't been able to figure this out yet due to some cyclic imports. Perhaps that should go into its own PR?

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/modelling/__init__.py
+++ b/Orange/modelling/__init__.py
@@ -1,3 +1,4 @@
 from .base import *
 
 from .knn import *
+from .tree import *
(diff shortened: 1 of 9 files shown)


pavlin-policar:
Sure thing. Could you check if the text is ok?

PR #1819, opened 2016-12-13
Merge OWAdaBoost

pavlin-policar:
##### Issue
Merge AdaBoost to handle both classification and regression problems.

##### Description of changes
This merge was messier than the previous ones. A few things to note:

1. It is based on top of the merged tree widget since if AdaBoost can handle any data type then it obviously needs a base estimator that handles any input type. Basing it on top of the already merged tree widget makes this simple and keeps the base estimator a tree model.

2. I've had to comment out some tests that check that the parameters are correctly passed down to the learners. This is necessary since the param testing as it is right now expects all the parameters to be passed down to the learner. Since AdaBoost takes two different set of parameters (`loss` for regression and `algorithm` for classification), they are delegated appropriately and only the classification learner receives the `algorithm` parameter. This is not what the learner test mixin currently expects, hence the todo. I will update the tests appropriately, otherwise this will be a problem in multiple learners.

3. I've also added *random seed* to the parameter list in the widget. This seems useful for replicating results for demonstration (or testing) purposes. The aesthetic of the option is questionable though at best (see screenshot of merged widget below).

AdaBoost widget

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/ensembles/ada_boost.py
```diff
@@ -1,24 +1,37 @@
...
+    'SklAdaBoostRegressionLearner',
+]
 
 
 class SklAdaBoostClassifier(SklModelClassification):
     pass
 
 
-class SklAdaBoostLearner(SklLearnerClassification):
+class SklAdaBoosClassificationLearner(SklLearnerClassification):
```
kernc: SklAdaboos**t**Classification.

Orange/ensembles/ada_boost.py
```diff
@@ -1,24 +1,37 @@
...
     __wraps__ = skl_ensemble.AdaBoostClassifier
     __returns__ = SklAdaBoostClassifier
-    name = 'skl adaBoost'
+    name = 'skl adaBoost classification'
```
kernc: Setting `.name` is somewhat redundant since https://github.com/biolab/orange3/blob/40bbcde23b32c5c61f29052bd381876a4f3dfef8/Orange/base.py#L144-L161

Orange/ensembles/ada_boost.py
```diff
@@ -36,7 +49,21 @@ class SklAdaBoostRegressionLearner(SklLearnerRegression):
...
+                base_estimator.REGRESSION)
+        # If sklearn learner, get the underlying sklearn representation
         if isinstance(base_estimator, SklLearner):
             base_estimator = base_estimator.__wraps__(**base_estimator.params)
         super().__init__(preprocessors=preprocessors)
         self.params = vars()
+
+
+class SklAdaBoostLearner(Fitter):
+    name = 'skl adaboost'
```
kernc: Here too. `class SklAdaboostLearner(Fitter)` objects will act as having `.name` "adaboost".

Orange/modelling/base.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, preprocessors=None, **kwargs):
         self.kwargs = kwargs
         # Make sure to pass preprocessor params to individual learners
         self.kwargs['preprocessors'] = preprocessors
-        self.problem_type = None
+        self.problem_type = self.CLASSIFICATION
```
kernc: This.

Orange/ensembles/ada_boost.py
pavlin-policar: My bad, I must've missed when this was added. So I should just skip defining the `name` attributes altogether?

Orange/modelling/base.py
pavlin-policar: My bad, I had actually fixed this, but had forgotten to push it...

pavlin-policar:
I've moved the random seed option to the bottom of the section. It is better than before. It's still not very good, but I have no ideas on how to improve it.

at bottom

Orange/ensembles/ada_boost.py
kernc: As far as I see it, it's less redundant this way, so I do prefer it. I don't know whether this plays into @astaric's adored "Implicit is better than explicit" adage, but I find explicit often breaks DRY and is pain to maintain or keep in sync (cf. [_rule of generation_](https://en.wikipedia.org/wiki/Unix_philosophy#Eric_Raymond.E2.80.99s_17_Unix_Rules)), whereas some of the best software that has ever been compiled relies heavily on sane defaults. :smiley:

pavlin-policar: It's very clever, but since it's a property on the class, it will only work as expected when dealing with an instance. It doesn't actually work if you tried to do `SomeLearner.name` which happens [here](https://github.com/biolab/orange3/pull/1819/files#diff-ac6f5b024d3d33a616daa15703ec0a64R21). This is the only instance where I've come across this, and it's very easily fixable here. Just an observation.

Orange/ensembles/ada_boost.py
```diff
@@ -32,11 +44,22 @@ class SklAdaBoostRegressor(SklModelRegression):
...
+            base_estimator = base_estimator.get_learner(
+                base_estimator.REGRESSION)
+        # If sklearn learner, get the underlying sklearn representation
         if isinstance(base_estimator, SklLearner):
             base_estimator = base_estimator.__wraps__(**base_estimator.params)
         super().__init__(preprocessors=preprocessors)
         self.params = vars()
+
+
+class SklAdaBoostLearner(Fitter):
```
kernc: Fitters for KNN and Tree went into _Orange/modelling_. They might as well (or even rather) be defined closer to where their underlying learners are, but let's just be consistent.

Orange/modelling/base.py
```diff
@@ -68,5 +68,15 @@ def __kwargs(self, problem_type):
             self.__fits__[problem_type].__init__.__code__.co_varnames[1:])
         return {k: v for k, v in self.kwargs.items() if k in learner_kwargs}
 
+    @property
+    def supports_weights(self):
+        """The fitter supports weights if both the classification and
+        regression learners support weights."""
+        return self.get_learner(self.CLASSIFICATION).supports_weights and \
+            self.get_learner(self.REGRESSION).supports_weights
```
kernc: Only `SklLearner` instances define `supports_weights` property, other learners do not. Will that be a problem?

Orange/widgets/model/owadaboost.py
```diff
@@ -0,0 +1,116 @@
...
+                  "and adapts to the 'hardness' of each training sample. "
+    icon = "icons/AdaBoost.svg"
+    priority = 80
+
+    LEARNER = SklAdaBoostLearner
+
+    inputs = [("Learner", Learner, "set_base_learner")]
+
+    #: Algorithms for classification problems
+    algorithms = ["SAMME", "SAMME.R"]
```
kernc: According to http://scikit-learn.org/stable/auto_examples/ensemble/plot_adaboost_multiclass.html, SAMME.R is better. It is also the default.

Orange/modelling/base.py
pavlin-policar: Thank, I wouldn't have noticed that. I've now changed it so that it returns false if the learner doesn't have a `supports_weights` attribute to avoid any errors.

Orange/widgets/model/owadaboost.py
pavlin-policar: I've changed the default selection to now be SAMME.R

janezd:
@astaric, @lanzagar, is pylint now a required check? I'd like to merge, but I don't want to force it as admin.

astaric:
It is. As this is an old pull request, I would say use your god powers to merge it anyway, but in the future, we want to be more strict about the code style.

PR #1828, opened 2016-12-16
Pandas forward compat

kernc:
##### Description of changes

Minimal (non-exhaustive) pandas forward-compatible API with our current Table technology. Current API still (mostly) works but is deprecated and updated in tests. 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/misc/distmatrix.py
+++ b/Orange/misc/distmatrix.py
@@ -97,12 +97,12 @@ class DistMatrix(np.ndarray):
             col_items = row_items
         obj = self[np.ix_(row_items, col_items)]
         if self.row_items is not None:
-            obj.row_items = self.row_items[row_items]
+            obj.row_items = self.row_items.iloc[row_items]
         if self.col_items is not None:
             if self.col_items is self.row_items and row_items is col_items:
                 obj.col_items = obj.row_items
             else:
-                obj.col_items = self.col_items[col_items]
+                obj.col_items = self.col_items.iloc[col_items]
         return obj
 
     @classmethod
@@ -212,8 +212,9 @@ class DistMatrix(np.ndarray):
 
     @staticmethod
     def _trivial_labels(items):
-        return items and \
+        return items is not None and \
                isinstance(items, Table) and \
+               not items.empty and \
                len(items.domain.metas) == 1 and \
                isinstance(items.domain.metas[0], StringVariable)
 
--- a/Orange/preprocess/impute.py
+++ b/Orange/preprocess/impute.py
@@ -81,8 +81,7 @@ class DropInstances(BaseImputeMethod):
     description = ""
 
     def __call__(self, data, variable):
-        index = data.domain.index(variable)
-        return numpy.isnan(data[:, index]).reshape(-1)
+        return numpy.isnan(data.get_column_view(variable)[0])
 
 
 class Average(BaseImputeMethod):
@@ -165,7 +164,7 @@ class ReplaceUnknownsModel(Reprable):
         if isinstance(data, Orange.data.Instance):
             predicted = self.model(data)
         else:
-            predicted = self.model(data[mask])
+            predicted = self.model(data.iloc[mask])
         column[mask] = predicted
         return column
 
--- a/Orange/statistics/contingency.py
+++ b/Orange/statistics/contingency.py
@@ -1,12 +1,12 @@
 import math
 import numpy as np
-from Orange import data
+from Orange.data import Storage, Variable
 
 
 def _get_variable(variable, dat, attr_name,
                   expected_type=None, expected_name=""):
     failed = False
-    if isinstance(variable, data.Variable):
+    if isinstance(variable, Variable):
         datvar = getattr(dat, "variable", None)
         if datvar is not None and datvar is not variable:
             raise ValueError("Variable does not match the variable in the data.")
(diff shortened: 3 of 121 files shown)


Orange/misc/distmatrix.py
```diff
@@ -261,10 +262,10 @@ def save(self, filename):
         with open(filename, "wt") as fle:
             fle.write(data + "\n")
             if col_labels is not None:
-                fle.write("\t".join(str(e.metas[0]) for e in col_labels) + "\n")
+                fle.write("\t".join(str(m[0]) for m in col_labels.metas) + "\n")
```
astaric: why?

Orange/misc/distmatrix.py
```diff
@@ -261,10 +262,10 @@ def save(self, filename):
...
             for i, row in enumerate(self):
                 if row_labels is not None:
-                    fle.write(str(row_labels[i].metas[0]) + "\t")
+                    fle.write(str(row_labels.metas.T[0][i]) + "\t")
```
astaric: why?

Orange/preprocess/impute.py
```diff
@@ -81,8 +81,7 @@ class DropInstances(BaseImputeMethod):
     description = ""
 
     def __call__(self, data, variable):
-        index = data.domain.index(variable)
-        return numpy.isnan(data[:, index]).reshape(-1)
+        return numpy.isnan([list(row) for _, row in data[variable].iterrows()]).reshape(-1)
```
astaric: slow?

Orange/statistics/contingency.py
```diff
@@ -261,20 +242,19 @@ def get_contingency(dat, col_variable, row_variable=None, unknowns=None, unknown
     elif variable.is_continuous:
         return Continuous(dat, col_variable, row_variable, unknowns, unknown_rows)
     else:
-        raise TypeError("cannot compute distribution of '%s'" %
-                        type(variable).__name__)
+        raise TypeError("cannot compute distribution of '%s'" % type(variable).__name__)
 
 
-def get_contingencies(dat, skipDiscrete=False, skipContinuous=False):
+def get_contingencies(dat, skip_discrete=False, skip_continuous=False):
```
astaric: why?

Orange/tests/test_contingency.py
```diff
@@ -39,8 +39,8 @@ def test_discrete(self):
 
     def test_discrete_missing(self):
         d = data.Table("zoo")
-        d.Y[25] = float("nan")
-        d[0][0] = float("nan")
+        d.loc[d.index[25], d.domain.class_var] = float("nan")
```
astaric: I just love how pandas makes the code cleaner :)

Orange/widgets/data/owfeatureconstructor.py
```diff
@@ -544,8 +544,7 @@ def duplicateFeature(self):
     def check_attrs_values(self, attr, data):
         for i in range(len(data)):
             for var in attr:
-                if not math.isnan(data[i, var]) \
-                        and int(data[i, var]) >= len(var.values):
+                if data[var].iloc[data.index[i]] not in range(var.values):
```
astaric: check

Orange/widgets/data/owmergedata.py
```diff
@@ -183,7 +183,7 @@ def seen(val):
         ((-i, m) for i, m in enumerate(B.domain.metas, start=1))
     )
     reduced_indices_B = [i for i, var in iter_vars_B if not var in all_vars_A]
-    reduced_B = B[:, list(reduced_indices_B)]
+    reduced_B = B.loc[list(reduced_indices_B)]
```
astaric: iloc?

Orange/widgets/evaluate/owpredictions.py
```diff
@@ -278,11 +278,6 @@ def _update_predictions_model(self):
             results = []
             for p in slots:
                 values, prob = p.results
-                if p.predictor.domain.class_var.is_discrete:
```
astaric: why?

Orange/preprocess/impute.py
kernc: Haven't measured it, but I suppose not much slower than it was before. `numpy.isnan(data[:, index])` also did a full iteration over data (via `Sequence.__iter__()` via `table.__getitem__(i)`).

In pandas, this will be: `return data[variable].isnull()`.

I'll fix it up to use `Table.get_column_view()` interim.

Orange/tests/test_contingency.py
kernc: This is not code. These are tests, made compatible with as little change as possible. The useful function code will always be much cleaner.

Ideally, this would be:
```py
d.type.iloc[25] = np.nan
```
But our current code makes it hard to adapt it so. I suppose it will have to get a bit more ugly before it gets neat.

astaric:
As this PR touches too many files in the repository and has not been updated for almost a year, it is highly unlikely that it will be merged. We are still interested in porting the Table to pandas, but it will have to be done in a more gradual way.

If anyone is interested in working on this, I would suggest the following steps:

1. Add new methods to Table object that match pandas interface
(`empty()`, `iterrows()`, ... - see commit 9a1a35a6d2 for more ideas)
2. Modify the code to call the new methods while deprecating the old ones
(code calls `empty()`, `__bool__` becomes deprecated, ...)
3. Figure out if Table can be ported to pandas without modifying more than 10 files, if not, return to 1. :)

Each "pandas compatible" method should be added in a separate pull request, as that eases reviewing and raises the chances of PR actually being merged. I would also split 1. and 2. into two PRs as 1 should only touch a single file and rebasing it should be much easier than the modifications of all places the code is used in 2.

PR #1830, opened 2016-12-18
Merge OWSGD

pavlin-policar:
##### Title

Merge SGD Widget to handle both classification and regression problems.

##### Description of changes
Orange currently only had SGD for regression, but after a bit of digging, I found sklearn supports SGD for classification problems as well, so I added that.

I've also completely redone the widget itself since the old one didn't support many of the parameters that sklearn let's us set. I've also renamed some things in order to keep things more consistent across the learners e.g. I've renamed *penalty* to *regularization*, *absolute norm* and *euclidean norm* to *lasso (L1)* and *ridge (L2)* to keep it consistent with regression learners. The new names are hopefully more informative, if nothing else, they are more consistent with the existing learners.

The classification learner also has loss functions that are equivalent to other learning methods but using a stochastic learning approach ([sklearn docs](http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html#sklearn.linear_model.SGDClassifier)). Those include: hinge = linear SVM, log = logistic regression and perceptron. I've tried to name things accordingly here as well.

I've also added a function that converts booleans to "Yes" and "No" in `report.py`. This could be completely pointless, since we could hide a setting in the report altogether if it's a bool, but "Yes/No" seems better than "True/False".

Before vs after:
Before After

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
Looks great. I have just a bit of nitpicking: the last two checkboxes are not aligned on Os X.

screen shot 2017-01-20 at 11 13 27

This happens because of additional box around the second checkbox. Ales says this is unavoidable, the only (clean) solution is to put everything in a grid layout. A simpler, dirtier solution is to add a box around the first checkbox, i.e. `gui.hBox(box), self, 'shuffle', 'Shuffle data after each iteration'`.

pavlin-policar:
Done.

PR #1832, opened 2016-12-21
Error message for multitarget prediction

### LINKED ISSUE #1823: Wrong error message for multitarget prediction
##### Expected behavior

Error when having two or more target variables should be "Too many target variables."

##### Actual behavior

Error says "Data has no target variable."

Set two attributes in File widget to be target variable and connect Classification Tree or Test&Score to it.

ajdapretnar:
##### Issue
Fixes #1823.
When there was more than one target variable in the data, Test&Score reported a 'missing target variable'.

##### Description of changes
New error message for 'too many variables' introduced.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -165,17 +165,18 @@ class OWTestLearners(OWWidget):
     class_selection = settings.ContextSetting(TARGET_AVERAGE)
 
     class Error(OWWidget.Error):
-        class_required = Msg("Train data input requires a class variable")
-        class_required_test = Msg("Test data input requires a class variable")
+        class_required = Msg("Train data input requires a target variable.")
+        too_many_classes = Msg("Too many target variables.")
+        class_required_test = Msg("Test data input requires a target variable.")
         too_many_folds = Msg("Number of folds exceeds the data size")
         class_inconsistent = Msg("Test and train data sets "
-                                 "have different classes")
+                                 "have different target variables.")
 
     class Warning(OWWidget.Warning):
         missing_data = \
-            Msg("Instances with unknown target values were removed from{}data")
-        test_data_missing = Msg("Missing separate test data input")
-        scores_not_computed = Msg("Some scores could not be computed")
+            Msg("Instances with unknown target values were removed from{}data.")
+        test_data_missing = Msg("Missing separate test data input.")
+        scores_not_computed = Msg("Some scores could not be computed.")
         test_data_unused = Msg("Test data is present but unused. "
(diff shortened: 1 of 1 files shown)


ajdapretnar:
Fixes #1823.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -166,6 +166,7 @@ class OWTestLearners(OWWidget):
 
     class Error(OWWidget.Error):
         class_required = Msg("Train data input requires a class variable")
+        too_many_classes = Msg("Too many target variables.")
```
lanzagar: I like the use of "target variable" as that corresponds to the use in File and Select Columns widgets.
Could you change the messages above and below to also use that instead of "class variable"?

ajdapretnar: Will do! 👍

lanzagar: I do too. And for exactly the reasons you mentioned, I asked if she can correct the other two existing messages to be more like hers.
We all want the same thing :)

janezd: Sorry. Misread. Not enough sleep. Too much lecturing. Coffee too schwach. All of them.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -165,17 +165,18 @@ class OWTestLearners(OWWidget):
...
     class Error(OWWidget.Error):
-        class_required = Msg("Train data input requires a class variable")
-        class_required_test = Msg("Test data input requires a class variable")
+        class_required = Msg("Train data input requires a target variable.")
+        too_many_classes = Msg("Too many target variables.")
+        class_required_test = Msg("Test data input requires a target variable.")
         too_many_folds = Msg("Number of folds exceeds the data size")
         class_inconsistent = Msg("Test and train data sets "
-                                 "have different classes")
+                                 "have different target values.")
```
lanzagar: Should also be target _variables_ not _values_.

ajdapretnar: I understood that the content of target variable (values) is not matching. But it can indeed be also target variables, I just thought 'target values' makes it more clear.

lanzagar: The previous message could mean anything, but checking the code reveals that variables are being compared.

ajdapretnar: Ok, I'll change it.

PR #1844, opened 2016-12-29
[ENH] Table.transpose: Use heuristic to guess data type of attributes of attributes

VesnaT:
##### Issue

Attributes' data type is StringVariable by default. Use heuristic to guess it.

##### Description of changes
Refactor io.py to make guessing heuristic reusable. 
Use heuristic to guess data type of attributes to be saved to metas.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -13,6 +13,7 @@ import numpy as np
 import bottleneck as bn
 from scipy import sparse as sp
 
+import Orange.data  # import for io.py
 from Orange.data import (
     _contingency, _valuecount,
     Domain, Variable, Storage, StringVariable, Unknown, Value, Instance,
@@ -1453,6 +1454,7 @@ class Table(MutableSequence, Storage):
             feature names are mapped
         :return: Table - transposed table
         """
+
         self = cls()
         n_cols, self.n_rows = table.X.shape
         old_domain = table.attributes.get("old_domain")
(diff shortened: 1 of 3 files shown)


ajdapretnar:
This fails with Corpus (sparse issues, @nikicc) and with ImportImages (likely because data only has metas).

ajdapretnar:
Sorry, not part of this PR. Just random bugs I discovered, should be in a different PR.

ajdapretnar:
On the topic of this PR: shouldn't this make transposed zoo.tab also discrete? Or at least titanic, since it indeed has mostly 0 and 1 (if that's the criterium)? Dunno....

Orange/data/table.py
```diff
@@ -1453,6 +1453,8 @@ def transpose(cls, table, feature_names_column="",
             feature names are mapped
         :return: Table - transposed table
         """
+        from Orange.data.io import guess_data_type, sanitize_variable
```
lanzagar: This import does not seem particularly slow or special in a way that would warrant local importing inside a function instead of at the top of the file. Or am I missing something?

Orange/tests/test_table.py
```diff
@@ -2546,7 +2576,7 @@ def _compare_tables(self, table1, table2):
         self.assertEqual(table1.n_rows, table2.n_rows)
         np.testing.assert_array_equal(table1.X, table2.X)
         np.testing.assert_array_equal(table1.Y, table2.Y)
-        np.testing.assert_array_equal(table1.metas, table2.metas)
+        self.assertTrue((table1.metas == table1.metas).all())
```
lanzagar: Is this OK?

VesnaT: I guess not.

PR #1845, opened 2016-12-29
OWTranspose: Show warning when table.X is empty

VesnaT:
##### Issue

OWTable crashed, when passed data table with no data (produced by OWTranspose).
To reproduce: transpose table that consists only of meta attributes and pass it to OWTable.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
I think this avoids a bug, but does not fix it.
Table.transpose should not create a one-dimensional metas array.

lanzagar:
Not needed after #1855 was merged.

PR #1850, opened 2016-12-30
[WIP] Treeviewer sklearn tree compatibility

pavlin-policar:
##### Issue
Trees output from Pythagorean forest are not compatible with the TreeViewer widget.

##### Description of changes

This builds upon the changes in #1786.

I've started to try and make sklearn tree models compatible with the TreeViewer widget. As of now, the tree structure is displayed properly.

My current method may be problematic, since essentially what I'm doing is converting the sklearn tree into the Orange tree when the Classifier is built. This definitely slows down anything using these trees, which kind of defeats the purpose of using sklearn trees in the first place.

Another thing I've noticed is that the models require the actual training data to function properly in visualisations. The Orange tree implementation does this as well. In order to provide this to sklearns random forests and trees, I've had to override the `fit_storage` method in the appropriate learners, and the `__init__` methods in the models. If this is needed in so many places, it might be a good idea to put it the base `SklModel` and `SklLearner` classes.

This is a first stab at it, but I don't think its very good, so any suggestions would help @janezd @astaric.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -286,7 +286,7 @@ class TreeModel(Tree, Model):
     pass
 
 
-class RandomForest:
+class RandomForestModel(Model):
     """Interface for random forest models
     """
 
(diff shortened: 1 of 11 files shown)


pavlin-policar:
Having established that this would complicate things far too much, a better solution is available in #1870.

PR #1857, opened 2017-01-04
Change defaults in Select Rows

ajdapretnar:
##### Issue
Purging results in non-matching data and causes many issues.

##### Description of changes
Default are now set to False for purging.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -47,8 +47,8 @@ class OWSelectRows(widget.OWWidget):
     settingsHandler = SelectRowsContextHandler()
     conditions = ContextSetting([])
     update_on_change = Setting(True)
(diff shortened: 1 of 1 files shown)


lanzagar:
What kind of issues does it cause?
We should have them written down and do something about them if possible - just changing the defaults will not prevent users from checking the boxes again and getting the same issues.

ajdapretnar:
Uh, some selections aren't matched. Use two Select Rows, create two subsets and connect to Venn. All data sets will be 0. Untick all purging and the data appears. And there was at least one other I can't remember.

p.s. I've rebased and the AppVeyor is still misbehaving.

ajdapretnar:
One of the new commits solves this issue, so I'm closing this.

PR #1863, opened 2017-01-04
[ENH] OWRandomize: Add a new widget

VesnaT:
##### Issue

Create a new widget for shuffling columns in a data able.

##### Description of changes
Enable multiple types of randomization at the same time.
New widget Randomize.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation


Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def set_data(self, data):
+        self.data = data
+        self.apply()
+
+    def apply(self):
+        data = None
+        if self.data:
+            rand_seed = self.random_seed or None
+            _type = Randomize.RandomizeClasses * self.shuffle_class | \
```
janezd: `type_` instead of `_type` (like `exec_` in main. Where, by the way, the underscore is no longer necessary in Python 3, AFAIK).

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+            indices = sorted(random.sample(range(len(self.data)), size))
+            randomized = Randomize(_type, rand_seed)(self.data[indices])
+            data = self.data.copy()
+            for i, instance in zip(indices, randomized):
+                data[i] = instance
+        self.send("Data", data)
+
+    def send_report(self):
+        text = "No shuffling"
+        labels = ["Classes", "Features", "Metas"]
```
janezd: Use lowercase here (*Shuffle: Classes, Metas* doesn't look nice)

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+        include = [self.shuffle_class, self.shuffle_attrs, self.shuffle_metas]
+        if sum(include):
+            text = ", ".join([l for i, l in zip(include, labels) if i])
```
janezd: I prefer printing "a, b, c, d and e", where possible. Try something like

    def send_report(self):
        labels = ["classes", "features", "metas"]
        include = [self.shuffle_class, self.shuffle_attrs, self.shuffle_metas]
        include = [label for labels, i in zip(labels, include)
        text = " and ".join(filter(None, (", ".join(include[:-1]), include[-1:]))) or "none"
        self.report_items ...

This doesn't require `if`. If you find the last line too cryptic, you can get rid of `filter` by handling `len(include) == 1` separately.

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+            random.seed(rand_seed)
+            indices = sorted(random.sample(range(len(self.data)), size))
+            randomized = Randomize(_type, rand_seed)(self.data[indices])
+            data = self.data.copy()
+            for i, instance in zip(indices, randomized):
+                data[i] = instance
+        self.send("Data", data)
+
+    def send_report(self):
+        text = "No shuffling"
```
janezd: I'd prefer `text = "none"`.

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+    scope_prop = Setting(80)
+    random_seed = Setting(0)
+    auto_apply = Setting(True)
+
+    def __init__(self):
+        super().__init__()
+        self.data = None
+
+        # GUI
+        box = gui.vBox(self.controlArea, "Shuffle")
```
janezd: Vertical arrangement of check boxes with such short labels doesn't look nice. I think this is nicer:

        box = gui.hBox(self.controlArea, "Shuffled columns")
        box.layout().setSpacing(20)

That is, `hBox`, longer (more informative) title, and spacing.

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+        self.class_check = gui.checkBox(
+            box, self, "shuffle_class", "Classes",
+            callback=self._shuffle_check_changed)
+        self.attrs_check = gui.checkBox(
+            box, self, "shuffle_attrs", "Features",
+            callback=self._shuffle_check_changed)
+        self.metas_check = gui.checkBox(
+            box, self, "shuffle_metas", "Metas",
+            callback=self._shuffle_check_changed)
+        gui.separator(box, 10, 10)
```
janezd: I'd move this checkbox (and the separator) to the other box. It could be in either of them, but in the second it's the only checkbox, so it's less confusing.

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+        self.replicable_check = gui.checkBox(
+            box, self, "random_seed", "Replicable shuffling",
+            callback=self._shuffle_check_changed)
+
+        box = gui.vBox(self.controlArea, "Scope")
```
janezd: The title is not too informative. *Shuffled rows* would work better - it parallels the title of the box above, plus it allows for simpler labels for the slider (see below)

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+        hbox = gui.hBox(box)
+        gui.widgetLabel(hbox, "No data")
```
janezd: I would understand *No data* as *No data comes out* not as *No data is shuffled*. Changing the box title allows us to change this label to "None"...

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+        self.scope_slider = gui.hSlider(
+            hbox, self, "scope_prop", minValue=0, maxValue=100, width=140,
+            createLabel=False, callback=self._scope_slider_changed)
+        gui.widgetLabel(hbox, "All data")
```
janezd: ... and this one to "All".

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+        self.send("Data", data)
+
+    def send_report(self):
+        text = "No shuffling"
+        labels = ["Classes", "Features", "Metas"]
+        include = [self.shuffle_class, self.shuffle_attrs, self.shuffle_metas]
+        if sum(include):
+            text = ", ".join([l for i, l in zip(include, labels) if i])
+        self.report_items("Settings",
+                          [("Shuffle", text),
```
janezd: If the box title is changed, perhaps also change `"Shuffle"` to `"Shuffled columns"`.

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+                           ("Scope", "{}% of data".format(self.scope_prop)),
```
janezd: "of rows" would be more accurate than "of data". However, you can change "Scope" is changed to "Proportion of shuffled rows", and then put just `"{} %".format(self.scope_prop)`.

Orange/preprocess/preprocess.py
```diff
@@ -326,8 +326,7 @@ class Randomize(Preprocess):
     >>> randomizer = Randomize(Randomize.RandomizeClasses)
     >>> randomized_data = randomizer(data)
     """
-    Type = Enum("Randomize",
-                "RandomizeClasses, RandomizeAttributes, RandomizeMetas")
+    Type = 1, 2, 4
```
janezd: This is now no longer an enum. Why not something like this?

    Type = Enum("Randomize", dict(RandomizeClasses=1, RandomizeAttributes=2, RandomizeMetas=4))

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+    icon = "icons/Random.svg"
+    priority = 2100
+
+    inputs = [("Data", Table, "set_data")]
+    outputs = [("Data", Table)]
+
+    resizing_enabled = False
+    want_main_area = False
+
+    shuffle_class = Setting(1)
```
janezd: I'd prefer `True` and `False`, so everybody knows that these are boolean flags, not numbers. The trick in `apply` will still work.

Orange/widgets/data/owrandomize.py
```diff
@@ -0,0 +1,119 @@
...
+        self.data = data
+        self.apply()
+
+    def apply(self):
+        data = None
+        if self.data:
+            rand_seed = self.random_seed or None
+            _type = Randomize.RandomizeClasses * self.shuffle_class | \
+                    Randomize.RandomizeAttributes * self.shuffle_attrs | \
+                    Randomize.RandomizeMetas * self.shuffle_metas
```
janezd: This is exactly how I'd do it and I like it. But perhaps

    type_ = sum(part for part, flag in zip(Randomize.Type, [self.shuffle_class, shuffle_attrs, shuffle_metas]) if flag)

is better. (If you use `[self.shuffle_class, shuffle_attrs, shuffle_metas]` in a lot of places, you can create a property

    @property
     def parts(self):
         return [self.shuffle_class, shuffle_attrs, shuffle_metas]

and simplify this expression and also `send_report`.

janezd:
I'd prefer a gui like this.

screen shot 2017-01-14 at 17 32 16

Orange/preprocess/preprocess.py
VesnaT: Why should it be an enum? It is no longer enum, because they don't support bitwise operators.
If you insist with enums, I should probably add two new parameters to the method (for each type of randomization). Or am I missing something here?

VesnaT:
Fixed all apart from the enum part.

Orange/preprocess/preprocess.py
kernc: Enums [accept a type parameter](https://docs.python.org/3/library/enum.html#functional-api), e.g. ```Enum(..., type=int)```.

VesnaT: Thanks!

janezd: Otherwise, no, it doesn't have to be `Enum`. We're mostly not using it. I just liked it and asked why it was removed.

PR #1864, opened 2017-01-04
[FIX] OWSave: Safer Check if Writer Support Sparse

nikicc:
##### Issue
Some writers in addons did not have `SUPPORT_SPARSE_DATA` flag which caused `OWSave` to crash.

##### Description of changes
Check if writer supports sparse more safely; i.e. if there is no flag assume it does not support sparse.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


nikicc:
https://sentry.io/biolab/orange3/issues/198595992/

PR #1870, opened 2017-01-05
[FIX] Treeviewer sklearn tree compatibility

pavlin-policar:
##### Issue
Trees output from Pythagorean forest are not compatible with the TreeViewer widget.

##### Description of changes
Having established that #1850  would cause far too much confusion and make the overall tree model code completely non-understandable (Adapter for sklearn trees, but now also vice versa, tree model for sklearn trees?), I've taken another stab at it.

It seems that converting the TreeViewer widget to use the TreeAdapter actually doesn't require any major changes at all. So this is what I did in this PR, TreeViewer now uses the TreeAdapter, just like PythagoreanTree, so at least it's less confusing.

This is still based on #1786 since that fixes the PythagoreanTree compatibility.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owtreeviewer.py
+++ b/Orange/widgets/visualize/owtreeviewer.py
@@ -376,9 +376,9 @@ class OWTreeGraph(OWTreeViewer2D):
     def update_node_info_reg(self, node):
         """Update the printed contents of the node for regression trees"""
         node_inst = node.node_inst
-        print(self.tree_adapter.get_distribution(node_inst)[0])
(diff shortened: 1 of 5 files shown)


pavlin-policar:
@kernc Could you please take a look at this as well. This builds upon the PR closed yesterday that fixed compatibility between the Pythagorean forest and Pythagorean trees. This one changes TreeViewer to accept the input of Pythagorean forests as well.

This is kind of high priority.

Issue #1875, opened 2017-01-06
Silhouette Plot: TypeError:  ufunc 'isnan' not supported for the input types


ales-erjavec:
https://sentry.io/biolab/orange3/issues/196744733/

```
TypeError:  ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
  Module "Orange.widgets.gui", line 2237, in __call__

    self.func(**kwds)
  Module "Orange.widgets.visualize.owsilhouetteplot", line 242, in _invalidate_scores

    self._update()
  Module "Orange.widgets.visualize.owsilhouetteplot", line 263, in _update

    mask = numpy.isnan(labels)
```
##### Steps to reproduce the behavior

File (brown-selected.tab) -> Select Columns (move 'gene' column to metas) -> Silhouette Plot

PR #1876, opened 2017-01-06
[FIX] owsilhouetteplot: Fix TypeError when cluster column is an object array

ales-erjavec:
##### Issue

Fixes gh-1875

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


ajdapretnar:
The fix is great. But the eternal question is, could we add a tiny tiny test to avoid this in the future?

Issue #1869, opened 2017-01-04
Concatenate widget: append source ID as class


markotoplak:
##### Orange version
master

##### Expected behavior
Should append source ID as class.

##### Actual behavior
Raises exception.

##### Steps to reproduce the behavior
Connect Iris and Titanic to Concatenate, set appending source ID as class.

Issue #1865, opened 2017-01-04
Concatenate widget: no union of attributes


markotoplak:
##### Orange version
At least 3.3.8+, also affects master

##### Expected behavior
Given two different data sets to Concatenate and chosen option to output union of attributes would make a new data set with domain containing attributes from both data sets.

##### Actual behavior
Only attributes from the primary data set are on the output.

markotoplak:
Hmmm, looking at the widget text this seems like this was actually a feature.

markotoplak:
So the widget works as follows: union (or intersection) of features only works when there are no Primary Data. If there are Primary data, its domain is always used for the concatenated data set. Which was at least for me not what I expected when I sloppily used this widget. To make this clearer I propose deactivating "Domain Merging" box when "Primary Data" is on input in #1886. I also changed default inputs to "Additional Data".

I think the best would be to rename the multiple input "Additional data" to "Data"  and "Primary Data" to "Domain" (and only use the Domain of the data set connected to this input). Because this would break compatibility of saved schemas, I propose this change for a new feature release.

PR #1886, opened 2017-01-06
[FIX] Concatenate bugfix

### LINKED ISSUE #1872: Concatenate widget: autoapply problems
##### Orange version
3.3.8+

##### Expected behavior
Changes in the concatenate widget should be applied automatically.

##### Actual behavior
Changed input data is applied automatically, other changes are not.

markotoplak:
##### Issue

Fixes #1872, #1869, and tries to fix a usability problem in #1865 by saner defaults and by deactivating a part of the widget settings that are not used in certain input combinations.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owconcatenate.py
+++ b/Orange/widgets/data/owconcatenate.py
@@ -266,7 +266,7 @@ def append_columns(data, attributes=(), class_vars=(), metas=()):
     metas = [ascolumn(col) for _, col in metas]
 
     X = numpy.hstack((data.X,) + tuple(attr_cols))
-    Y = numpy.hstack((data.Y,) + tuple(class_cols))
+    Y = numpy.hstack((data._Y,) + tuple(class_cols))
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owconcatenate.py
```diff
@@ -64,7 +64,7 @@ def __init__(self):
         self.primary_data = None
         self.more_data = OrderedDict()
 
-        mergebox = gui.vBox(self.controlArea, "Domain Merging")
+        self.mergebox = mergebox = gui.vBox(self.controlArea, "Domain Merging")
```
lanzagar: `mergebox` is only used once, so instead of defining it, you could just use `self.mergebox` in L69.
Simpler and in this case even shorter.

markotoplak: Fixed. Thanks!

PR #1891, opened 2017-01-06
Merge OWSVM

pavlin-policar:
##### Issue
Merge SVM Widget to handle both classification and regression problems.

##### Description of changes
1. Implement three fitters, one for regular SVM (which merges C-SVM and epsilon-SVM), one for LinearSVM and one for NuSVM. I also implemented LinearSVM for regression, since it was missing and added tests (this was simple, since sklearn provides this).
2. Merge the two widgets together, I didn't change the widget layout at all, I did use the layout from SVR, which contains 2 pairs of param boxes (as seen on image of widget). I had to include C twice, because its used for both classification and regression with the regular SVM, but only for regression in the NuSVM, so that could be confusing - hence we include it twice, once with the `Regression` keyword to (hopefully) clear that up.
3. The only change I made to the UI was that the parameters of the non-active SVM type are now disabled. This could help avoid confusion. Also, the classification SVM did not have an iteration limit parameter (which is supported, and also present in the SVR), so I included that as well.

OWSVM

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/model/owsvm.py
```diff
@@ -0,0 +1,244 @@
...
+        if self.svm_type == self.SVM:
+            self.C_spin.setEnabled(True)
+            self.epsilon_spin.setEnabled(True)
+            self.nu_C_spin.setEnabled(False)
+            self.nu_spin.setEnabled(False)
+        else:
+            self.C_spin.setEnabled(False)
+            self.epsilon_spin.setEnabled(False)
+            self.nu_C_spin.setEnabled(True)
+            self.nu_spin.setEnabled(True)
```
janezd: Add `self.settings_changed()`

janezd:
I haven't found any other problems, except for the missing call `self.settings_changed`, so I'd merge. Could you please rebase.

pavlin-policar:
Done.

janezd:
@astaric, can you nudge travis to finish this?

astaric:
no, you need to amend the latest commit and force push it (or add another commit or sth).

janezd:
I knew this. But what if it's not my PR?

pavlin-policar:
Sorry for noticing this so late. I've ammended and pushed, so travis should do his thing now.

janezd:
No problem. I'd do it, but couldn't (or didn't know how to).

Now it segfaulted?! Can you do it again, so we see whether it's just a travis fluke or something real?

pavlin-policar:
Hmm, it segfaulted again, and seemingly in the same spot as before. Is this just Travis doing strange things, or have I accidentally done something wrong? As far as I can tell, I haven't touched the reports...

astaric:
I have restarted the travis build (again).

Report tests are problematic, as they authomatically generate reports for all widgets in a category, so a minor change somewhere cascades and results in a crash somewhere else. We should probably remote these general tests and add separate report tests for each widget. But not today :)

PR #1901, opened 2017-01-09
[FIX] OWSelectRows: Output None when data has 0 rows

VesnaT:
##### Issue

When conditions result in a table with 0 rows, None should be sent to the output.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


lanzagar:
I am not so sure about this. I think Select Rows and Data Sampler were intentionally handled differently from some other widgets and are returning empty data by design (instead of None).
We should discuss what we want and decide on some general guidelines about this.

janezd:
Following our discussion yesterday, I guess we should close this PR without merging and fix other widgets instead.

PR #1902, opened 2017-01-09
[FIX] owrank: Remove `super()` call from `migrate_settings`

ales-erjavec:
##### Issue

super cannot be used in migrate_settings because the method is invoked
during type creation by the widget's metaclass while the implicit
`__class__` cell is not yet bound.

Prevents a warning during class import:

    Could not read defaults for widget <class 'Orange.widgets.data.owrank.OWRank'>
    The following error occurred:

    super(): empty __class__ cell

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -691,7 +691,6 @@ class OWRank(OWWidget):
 
     @classmethod
     def migrate_settings(cls, settings, version):
-        super().migrate_settings(settings, version)
         if not version:
             # Before fc5caa1e1d716607f1f5c4e0b0be265c23280fa0
             # headerState had length 2


Orange/widgets/data/owrank.py
```diff
@@ -691,7 +691,6 @@ def create_scores_table(self, labels):
 
     @classmethod
     def migrate_settings(cls, settings, version):
-        super().migrate_settings(settings, version)
```
astaric: Could we use explicit name of the parent class instead of the super() or would that also fail?

ales-erjavec: OWWidget.migrate_settings would work. But what use would it really be? It can't do anything with the information provided by the version number (it's not its version).

astaric: Good point

Issue #1782, opened 2016-11-29
OWSelectRows: Error it is not possible to use "." as decimal separator and "," produces an error


lopezco:
## Context
__Exception:__ ```ValueError: could not convert string to float: '0,6'```
__Module:__ ```Orange.widgets.data.owselectrows:392```
__Widget Name:__ ```Select Rows```
__Widget Module:__ ```Orange.widgets.data.owselectrows:362```
__Version:__ ```3.3.9.dev0+f3f5ca0```
__Environment:__ ```Python 3.5.2 on Linux 4.4.0-47-generic #68-Ubuntu SMP Wed Oct 26 19:39:52 UTC 2016 x86_64```
__Local Variables:__ ```OrderedDict([('.0', <tuple_iterator object at 0x7f285621ab70>), ('v', '0,6')])```

### Expected behavior

To work with '.' decimal separator

### Actual behavior

An exception is risen

```python
Traceback (most recent call last):
  File "/home/workspace/git/orange3/Orange/widgets/data/owselectrows.py", line 362, in conditions_changed
    self.commit()
  File "/home/workspace/git/orange3/Orange/widgets/gui.py", line 1972, in unconditional_commit
    do_commit()
  File "/home/workspace/git/orange3/Orange/widgets/gui.py", line 1980, in do_commit
    commit()
  File "/home/workspace/git/orange3/Orange/widgets/data/owselectrows.py", line 392, in commit
    attr_index, oper, *[float(v) for v in values])
  File "/home/workspace/git/orange3/Orange/widgets/data/owselectrows.py", line 392, in <listcomp>
    attr_index, oper, *[float(v) for v in values])
ValueError: could not convert string to float: '0,5'
```

### Steps to reproduce the behavior

1 - Load a dataset with continuous variables
2 - Use a select columns to subset the dataset with decimal values
3 - Try to use '.' as separator = wont work
4 - Try to use ',' as separate = error

### Possible Solution
We can change the numeric validator with a QRegExpValidator

```python
        def add_numeric(contents):
            le = add_textual(contents)
            le.setValidator(QRegExpValidator(QRegExp("[-]{0,1}\d{1,}[\.]{0,1}\d{0,}e{0,1}[-]{0,1}\d{1,}[\.]{0,1}\d{0,}")))
            return le
```

ajdapretnar:
Dot works for me normally.
Comma indeed produces an error and it should not.

lopezco:
I think it has to do with my machine locale (FR).

PR #1903, opened 2017-01-09
OWSelectRows: locale-independent numeric value validation

kernc:
Fixes https://github.com/biolab/orange3/issues/1782

##### Description of changes
Replace locale-dependent `QDoubleValidator` with custom, hopefully correct regex validator.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -51,7 +51,7 @@ class SelectRowsContextHandler(DomainContextHandler):
                 var = attr in domain and domain[attr]
                 if var and var.is_continuous and not isinstance(var, TimeVariable):
                     value[i] = (attr, op,
-                                list([QLocale().toString(i, 'f')
+                                list([QLocale().toString(float(i), 'f')
                                       for i in values]))
         return value
 
@@ -88,7 +88,7 @@ class OWSelectRows(widget.OWWidget):
             (FilterContinuous.NotEqual, "is not"),
             (FilterContinuous.Less, "is below"),
             (FilterContinuous.LessEqual, "is at most"),
-            (FilterContinuous.Greater,"is greater than"),
+            (FilterContinuous.Greater, "is greater than"),
             (FilterContinuous.GreaterEqual, "is at least"),
             (FilterContinuous.Between, "is between"),
             (FilterContinuous.Outside, "is outside"),
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -11,7 +11,7 @@ from AnyQt.QtGui import (
     QDoubleValidator, QRegExpValidator, QStandardItemModel, QStandardItem,
     QFontMetrics, QPalette
 )
-from AnyQt.QtCore import Qt, QPoint, QRegExp, QPersistentModelIndex
+from AnyQt.QtCore import Qt, QPoint, QRegExp, QPersistentModelIndex, QLocale
 
 from Orange.data import (ContinuousVariable, DiscreteVariable, StringVariable,
                          Table, TimeVariable)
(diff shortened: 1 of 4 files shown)


ajdapretnar:
This seems to work for me. I've set an attribute to contain comma-separated values (e.g. 1,23) and then converted it to nominal feature (from meta string). And the values were indeed displayed in Scatter Plot properly! Kudos!

Before merging, probably someone else should check it as well. Perhaps @astaric?

astaric:
IMO, the problem is that parsing is not locale aware and not the other way around.
https://docs.python.org/2/library/locale.html#locale.atof

ajdapretnar:
So.... will this be converted to locale-aware parsing or we go with regex?

lanzagar:
I agree locale-aware numbers would appeal to certain people.

Also - avoiding a custom regex is always good :)
Incompatibility with python 3.6 `_` is probably not going to bother anyone, but the fact that `6.022E23` doesn't work might.
https://docs.python.org/3.6/reference/lexical_analysis.html#floating-point-literals

Orange/widgets/data/owselectrows.py
```diff
@@ -269,12 +272,31 @@ def _get_value_contents(box):
...
+            locale.setNumberOptions(QLocale.RejectGroupSeparator)
+            validator.setLocale(locale)
+            self._validators.append(validator)
+
         def validate(self, input_, pos):
             if not input_:
                 return (QDoubleValidator.Acceptable, input_, pos)
-            else:
-                return super().validate(input_, pos)
+            for validator in self._validators:
```
astaric: This looks complicated. I would just keep the validation using _DEFAULT_LOCALE_NUMERIC (the way numbers look in all spin boxes). And you can then simply setLocale in __init__ and skip the validator-inside-validator dance.

Orange/widgets/data/owselectrows.py
```diff
@@ -427,8 +449,14 @@ def commit(self):
                             self.error(e.args[0])
                             return
 
+                    try:
+                        floats = [float(v) for v in values]
```
astaric: What about putting all the parsing (time and ordinary) to a separate method that returns a list of float values or None if the parsing failed? This would improve the readability a bit and probably pass the failing travis test.

Orange/widgets/data/owselectrows.py
```diff
@@ -427,8 +449,14 @@ def commit(self):
...
+                    except ValueError:
+                        floats = [_DEFAULT_LOCALE_NUMERIC.toDouble(v)[0]
```
astaric: The second value in the tuple indicates whether the parsing was successful. It might be a good idea to inspect the value and show an error in parsing fails (the way it is done when parsing floats a couple of lines above).

Orange/widgets/data/owselectrows.py
```diff
@@ -416,19 +460,11 @@ def commit(self):
                 operators = self.Operators[type(attr)]
                 opertype, _ = operators[oper_idx]
                 if attr.is_continuous:
-                    if any(not v for v in values):
+                    floats = self._values_to_floats(attr, values)
+                    if floats is None:
```
astaric: This is ok for no values, but not if an error occurred. The old code for error (in time variable) called return.

Orange/widgets/data/owselectrows.py
```diff
@@ -416,19 +465,15 @@ def commit(self):
...
-
-                    # Parse datetime strings into floats
-                    if isinstance(attr, TimeVariable):
-                        try:
-                            values = [attr.parse(v) for v in values]
-                        except ValueError as e:
-                            self.error(e.args[0])
-                            return
-
+                    floats = self._values_to_floats(attr, values)
```
astaric: This should probably be moved inside the try block :)

PR #1907, opened 2017-01-10
[WIP] OWOutliers: Fix crash for multiclass dataset

### LINKED ISSUE #1906: TypeError:  OneClassSVMLearner doesn't support multiple class variables
https://sentry.io/biolab/orange3/issues/202498152/

```
TypeError:  OneClassSVMLearner doesn't support multiple class variables
  Module "Orange.widgets.data.owoutliers", line 141, in commit

    y_pred = self.detect_outliers()
  Module "Orange.widgets.data.owoutliers", line 169, in detect_outliers

    model = learner(self.data)
  Module "Orange.base", line 253, in __call__

    m = super().__call__(data)
  Module "Orange.base", line 52, in __call__

    self.__class__.__name__)
```

VesnaT:
##### Issue

Fixes #1906

##### Description of changes
Show error message when dataset has multiple class variables.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owoutliers.py
+++ b/Orange/widgets/data/owoutliers.py
@@ -8,6 +8,7 @@ from Orange.classification import OneClassSVMLearner, EllipticEnvelopeLearner
 from Orange.data import Table, Domain, ContinuousVariable
 from Orange.widgets import widget, gui
 from Orange.widgets.settings import Setting
+from Orange.widgets.widget import Msg
 from Orange.widgets.utils.sql import check_sql_input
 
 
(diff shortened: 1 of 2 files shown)


lanzagar:
OneClassSVM does not mind multiclass data as it is an unsupervised method and ignores all target variables anyway.
But we need to discuss how best to extend it from the base Learner and make it work.

astaric:
I agree with @lanzagar, but not as a part of this PR. We would like to fix the crash now, while the redesigned learner can be a part of the next feature release.

PR #1923, opened 2017-01-13
[RFC] ErrorReporting: Report developer names to Sentry

nikicc:
##### Issue
A while ago we discussed it would be nice if developers could report errors with their names rather than machine ids.

##### Description of changes
Machine ID line edit was added to Preferences menu. If the name is provided use it for reporting errors else fallback to reporting machine ids as before.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/errorreporting.py
+++ b/Orange/canvas/application/errorreporting.py
@@ -190,6 +190,8 @@ class ErrorReporting(QDialog):
         packages = ', '.join(sorted("%s==%s" % (i.project_name, i.version)
                                     for i in pip.get_installed_distributions()))
 
+        machine_id = QSettings().value('error-reporting/machine-id', '', type=str)
+
         # If this exact error was already reported in this session,
         # just warn about it
         if (err_module, widget_module) in cls._cache:
@@ -227,7 +229,7 @@ class ErrorReporting(QDialog):
             platform.python_version(), platform.system(), platform.release(),
             platform.version(), platform.machine())
         data[F.INSTALLED_PACKAGES] = packages
-        data[F.MACHINE_ID] = str(uuid.getnode())
+        data[F.MACHINE_ID] = machine_id or str(uuid.getnode())
         data[F.STACK_TRACE] = stacktrace
         if err_locals:
             data[F.LOCALS] = err_locals
--- a/Orange/canvas/application/settings.py
+++ b/Orange/canvas/application/settings.py
@@ -16,7 +16,7 @@ from ..utils.propertybindings import (
 from AnyQt.QtWidgets import (
     QWidget, QMainWindow, QComboBox, QCheckBox, QListView, QTabWidget,
     QToolBar, QAction, QStackedWidget, QVBoxLayout, QHBoxLayout,
-    QFormLayout, QSizePolicy
+    QFormLayout, QSizePolicy, QLineEdit,
 )
 
 from AnyQt.QtCore import (
@@ -357,6 +357,17 @@ class UserSettingsDialog(QMainWindow):
         box.setLayout(layout)
         form.addRow(self.tr("Help window"), box)
 
+        tab.setLayout(form)
+
+        # Error Reporting Tab
+        tab = QWidget()
+        self.addTab(tab, self.tr("Error Reporting"),
+                    toolTip="Settings related to error reporting")
+
+        form = QFormLayout()
+        line_edit_mid = QLineEdit()
+        self.bind(line_edit_mid, "text", "error-reporting/machine-id")
+        form.addRow("Machine ID:", line_edit_mid)
         tab.setLayout(form)
 
         if self.__macUnified:
--- a/Orange/canvas/config.py
+++ b/Orange/canvas/config.py
@@ -114,7 +114,10 @@ spec = \
      ("logging/dockable", bool, True, "Allow log window to be docked"),
 
      ("help/open-in-external-browser", bool, False,
-      "Open help in an external browser")
+      "Open help in an external browser"),
+
+     ("error-reporting/machine-id", str, '',
+     "Report custom name instead of machine ID"),
      ]
 
 spec = [config_slot(*t) for t in spec]


Orange/canvas/application/errorreporting.py
```diff
@@ -38,6 +38,19 @@ class OWWidget: pass
...
+    """Check if developer provided his name for Sentry reports."""
+    try:
+        from Orange.misc.environ import data_dir_base
+        fname = os.path.join(data_dir_base(), 'Orange', 'developer_id.txt')
+        if os.path.exists(fname):
+            with open(fname, 'r') as f:
+                username = f.readline().strip()
+                return username
+    except Exception:
+        return ''
```
astaric: This methods can return three different things.

username, if file was found and in correct format
'', if there was some error (something went wrong when opening/reading file)
None, it the file does not exist.

Could we merge the last two cases?

astaric:
Can you think of a way a new developer could set this id (settings/command line flag/instructions somewhere)?

In a month or two, the only way to "discover" this feature will be to stumble upon this code.

Orange/canvas/application/errorreporting.py
nikicc: Adressed.

Orange/canvas/application/errorreporting.py
```diff
@@ -38,6 +38,20 @@ class OWWidget: pass
...
+def get_developer_id():
+    """Check if developer provided his name for Sentry reports."""
+    try:
+        from Orange.misc.environ import data_dir_base
+        fname = os.path.join(data_dir_base(), 'Orange', 'developer_id.txt')
+        if os.path.exists(fname):
+            with open(fname, 'r') as f:
+                username = f.readline().strip()
+                return username
+    except Exception:
```
lanzagar: `except Exception` should always be avoided if possible.
Here you check if the file exists before doing anything, so users should be safe. And if developers manage to break something with a strange file, we probably want to see the error.

Orange/canvas/application/errorreporting.py
```diff
@@ -38,6 +38,20 @@ class OWWidget: pass
 log = logging.getLogger()
 
 
+def get_developer_id():
+    """Check if developer provided his name for Sentry reports."""
+    try:
+        from Orange.misc.environ import data_dir_base
+        fname = os.path.join(data_dir_base(), 'Orange', 'developer_id.txt')
```
kernc: How do you (we) feel about environment variables, like `ORANGE_MACHINE_ID='whatever'`?

https://github.com/biolab/orange3/pull/1929/commits/6e244005c4202b73a8e1248212159aaacf2eff55 much too late proposes `ORANGE_DEBUG` and https://github.com/biolab/orange3/pull/1771/commits/bdf6abc68b0dfd7067d70d77cbd7216283cfae45 introduced `ORANGE_DEPRECATIONS_ERROR`.

Orange/canvas/application/errorreporting.py
nikicc: Agreed. If we'll stick with storing this name to a file this will be refactored.

Orange/canvas/application/errorreporting.py
nikicc: I think I like this approach better and the code will simplify as well.

Though this complicates things if someday we would like to set these names through the GUI.

astaric: I would prefer it as a setting I can set in preferences :)

kernc: Setting it via GUI is just as easy:
```py
import os
os.environ['ORANGE_MACHINE_ID'] = 'whatever'
```

astaric: By first reading the value from file :)

astaric: Btw, ORANGE_DEPRECATIONS_ERROR causes tests to fail locally if it is not set, so I do not think it is a good idea. Same goes for ORANGE_DEBUG, I would prefer to have a runtime argument (that is listed when canvas is run with --help)

nikicc:
We set the names through the Preferences menu now. Hence, we probably do not need the environment variables, however, if we decide to support them they can be added later.

Please, check the current version.

Orange/canvas/application/errorreporting.py
kernc: The case for environment variables vs. just program arguments is that the behavior is tied to the environment instead of program invocation. A developer, regardless of if I run Orange from terminal, through PyCharm, through gdb or valgrind, or if I run standalone widgets sans canvas, I want Orange to expose the debug interface, whatever that be, and to object loudly to deprecated interfaces. But I can _add_ program arguments for those, if you wish.

> causes tests to fail locally if it is not set

That's a bug. :+1:

Orange/canvas/application/settings.py
```diff
@@ -359,6 +360,24 @@ def __setupUi(self):
...
+                super().__init__()
+                self.setText(get_machine_id())
+
+            def focusOutEvent(self, QFocusEvent):
+                super().focusOutEvent(QFocusEvent)
+                set_machine_id(self.text())
+
+        form = QFormLayout()
+        form.addRow("Machine ID:", MachineIDLineEdit())
+        tab.setLayout(form)
```
kernc: How about more like the surrounding code:
```py
self.bind(machine_id_edit, 'text', 'debug/machine-id')
```
And then you don't need the helper module _Orange/misc/reporting.py_; you can use:
```py
QSettings().value('debug/machine-id', '', type=str) or str(uuid.getnode())
```

nikicc: I was thinking about that but those settings are stored under the version directory and are hence reset with each new version.

If there are other things that should persist through versions we should probably develop something more general (like those settings above) that persists through updates.

Orange/canvas/application/errorreporting.py
kernc: > > causes tests to fail locally if it is not set
> 
> That's a bug. :+1:

Fixed in https://github.com/biolab/orange3/pull/1929/commits/d458310c600d28b3140ab578ce52ac6434280cc4

Orange/canvas/application/settings.py
ales-erjavec: The settings stored by QSettings() do not reset between versions (in fact they persist since Orange 2.7).

Orange/misc/reporting.py
```diff
@@ -0,0 +1,18 @@
+import os
+from Orange.misc.environ import data_dir_base
+
+SETTINGS_FILE = os.path.join(data_dir_base(), 'Orange', 'machine_id.txt')
+
+
+def get_machine_id():
+    if os.path.exists(SETTINGS_FILE):
+        with open(SETTINGS_FILE, 'r') as f:
```
ales-erjavec: Please always use
```python
try:
    with open(..., 'r') as f:
         ...
except FileNotFoundError:
    ...
```

Orange/canvas/application/settings.py
nikicc: My bad 🙈

Orange/misc/reporting.py
nikicc: Thanks, will note for future occasions. In this PR this was removed anyhow.

nikicc:
OK, I think this is (finally!) in a decent state; thought, I probably still don't fully comprehend how our QSettings work. Please, check again.

PR #1928, opened 2017-01-17
[FIX] Tree: Fix crash when two attributes equal number of values

janezd:
When two attributes had an equal number of values, the test before binarization tried to compare instances of `DiscreteVariable`.

##### Includes
- [X] Code changes


janezd:
https://sentry.io/biolab/orange3/issues/206543234/

PR #1929, opened 2017-01-17
Geomap & webview updates

kernc:
##### Description of changes
See individual commits.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,6 +111,16 @@ implicitly.
 [doctests]: https://en.wikipedia.org/wiki/Doctest
 
 
+Environment variables
+---------------------
+Set these environment variables for value-added behavior:
+
+* `ORANGE_DEBUG=1` - general developing and debugging. Influences stuff like
+  DOM Inspector in QWebView right-click menu, etc.
+* `ORANGE_DEPRECATIONS_ERROR=1` - whether warnings of type
+  `OrangeDeprecationWarning` should be raised as exceptions.
+
+
 Commit messages
 ---------------
 Make a separate commit for each logical change you introduce. We prefer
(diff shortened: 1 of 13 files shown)


lanzagar:
Changes work well for me 👍 

Can you also fix the following warnings after removing data:
`unable to set <Orange.widgets.gui.OrangeComboBox object at 0x7fdbcccf8438> to value 'None'. Possible values are (Same color).`
etc.

ajdapretnar:
Is this PR intended to also solve the mischievous behaviour of Corpus Viewer? E.g.

```
Property 'sizeGripEnabled'' of object 'OWCorpusViewer' has no notify signal and is not constant, value updates in HTML will be broken!
Property 'modal'' of object 'OWCorpusViewer' has no notify signal and is not constant, value updates in HTML will be broken!
Property 'progressBarValue'' of object 'OWCorpusViewer' has no notify signal and is not constant, value updates in HTML will be broken!
Property 'processingState'' of object 'OWCorpusViewer' has no notify signal and is not constant, value updates in HTML will be broken!
```

I get a bunch of these in the console. I'm just asking since I see some WebView fixes here. If this is for a different PR, I'll talk to @nikicc about it.

kernc:
> I get a bunch of these in the console

This is the relevant warning:
```text
Orange/widgets/utils/webview.py:406: OrangeDeprecationWarning: Don't expose QWidgets in WebView. Construct minimal QObjects instead.
```
This the offending line: https://github.com/biolab/orange3-text/blob/473968c5274c0c2f27cf14811e645191f00b3f5b/orangecontrib/text/widgets/owcorpusviewer.py#L115
If the webview doesn't callback into Qt, a bridge needn't be set; in particular not a QWidget one.

Issue #1930, opened 2017-01-17
owmds: ValueError:  cannot convert float NaN to integer


ales-erjavec:
https://sentry.io/biolab/orange3/issues/204303966/

```
---------------------------- ValueError Exception -----------------------------
Traceback (most recent call last):
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/Orange/widgets/gui.py", line 2237, in __call__
    self.func(**kwds)
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/Orange/widgets/unsupervised/owmds.py", line 722, in _on_size_index_changed
    self._update_plot()
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/Orange/widgets/unsupervised/owmds.py", line 736, in _update_plot
    self._setup_plot()
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/Orange/widgets/unsupervised/owmds.py", line 937, in _setup_plot
    antialias=True
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/Orange/widgets/unsupervised/owmds.py", line 56, in __init__
    super().__init__(*args, **kwargs)
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/pyqtgraph/graphicsItems/ScatterPlotItem.py", line 262, in __init__
    self.setData(*args, **kargs)
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/pyqtgraph/graphicsItems/ScatterPlotItem.py", line 308, in setData
    self.addPoints(*args, **kargs)
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/pyqtgraph/graphicsItems/ScatterPlotItem.py", line 409, in addPoints
    self.updateSpots(newData)
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/pyqtgraph/graphicsItems/ScatterPlotItem.py", line 562, in updateSpots
[...]

PR #1931, opened 2017-01-17
[FIX] MDS Widget: Handle NaN values for plot point styling

ales-erjavec:
##### Issue

Fixes gh-1930

##### Description of changes

Make `_setup_plot` more robust for NaN values in the data.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -742,8 +742,10 @@ class OWMDS(OWWidget):
 
         size = self._effective_matrix.shape[0]
 
-        def column(data, variable):
+        def column(data, variable, dtype=None):
             a, _ = data.get_column_view(variable)
+            if dtype is not None:
+                a = a.astype(dtype)
             return a.ravel()
 
         def attributes(matrix):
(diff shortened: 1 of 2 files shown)


Orange/widgets/unsupervised/owmds.py
```diff
@@ -861,9 +863,10 @@ def jitter(x, factor=1, rstate=None):
                 size_data = MinPointSize + size_data * point_size
             elif have_data and size_index > 0:
                 size_var = self.sizevar_model[size_index]
-                size_data = column(self.data, size_var)
+                size_data = column(self.data, size_var, dtype=float)
                 size_data = scale(size_data)
                 size_data = MinPointSize + size_data * point_size
+                size_data[numpy.isnan(size_data)] = 1  # maybe 0?
```
lanzagar: How about `MinPointSize` (or equivalently 0, but one line higher)

ales-erjavec: Would prefer to keep the points with unknown size visually distinguished from the rest (in some way)

lanzagar: OK, then the current value of 1 should be fine. It's already smaller than anything else.

PR #1935, opened 2017-01-20
Domain editor speedup

### LINKED ISSUE #1866: File widget: slow loading of large files
##### Orange version
master

##### Expected behavior
A file with ~1000 continuous attributes and ~10000 samples is loaded in less than a second.

##### Actual behavior
It is way to slow.

Currently the changes to the domain in the bottom part of the file widget are always applied (even if there are none). Perhaps speed up loading by giving the modified domain info to the file reader (suggestion by @astaric ).

markotoplak:
Use numpy instead of python lists. Faster name changes.
Fixes #1866.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/domaineditor.py
+++ b/Orange/widgets/utils/domaineditor.py
@@ -217,13 +217,20 @@ class DomainEditor(QTableView):
             if place == Place.skip:
                 continue
             if orig_plc == Place.meta:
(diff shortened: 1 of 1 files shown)


astaric:
Speeds up opening of adult dataset from ~5s to ~1s

Issue #1938, opened 2017-01-20
ValueError:  zero-size array to reduction operation fmin which has no identity


ales-erjavec:
https://sentry.io/biolab/orange3/issues/206319094/

```
ValueError:  zero-size array to reduction operation fmin which has no identity
  Module "Orange.widgets.visualize.owlinearprojection", line 595, in customEvent

    self._setup_plot()
  Module "Orange.widgets.visualize.owlinearprojection", line 724, in _setup_plot

    X = plotutils.normalized(X)
  Module "Orange.widgets.visualize.owlinearprojection", line 1580, in normalized

    amin, amax = numpy.nanmin(a), numpy.nanmax(a)
  Module "numpy.lib.nanfunctions", line 237, in nanmin

    res = np.fmin.reduce(a, axis=axis, out=out, **kwargs)
```
To reproduce test on:
```
X1,X2,X3
c,c,c

?,1,2
1,?,2
1,2,?
```

PR #1939, opened 2017-01-20
[ENH] Widget testing utilities

ales-erjavec:
##### Issue

Add some utilities for testing widgets, some datasets, ...
Also includes fix (and test) for gh-1938

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -1,5 +1,6 @@
+import os
 import unittest
-import sip
+
 
 import numpy as np
 from AnyQt.QtWidgets import (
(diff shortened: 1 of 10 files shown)


Orange/widgets/tests/base.py
```diff
@@ -561,3 +566,61 @@ def _compare_selected_annotated_domains(self, selected, annotated):
...
+    def data_2(cls):
+        # type: () -> Orange.data.Table
+        """
+        Data set with 3 continuous features (X{1,2,3}) where all the columns
+        and rows contain at least one NaN value and X1, X2 are constant.
+
+        One discrete constant class D with NaN values.
+
+        Mixed continuous/discrete/string class metas ({X,D,S}M)
+        One discrete constant class (D) with NaN values
```
astaric: duplicated line

Orange/widgets/tests/base.py
```diff
@@ -561,3 +566,61 @@ def _compare_selected_annotated_domains(self, selected, annotated):
...
+
+
+class datasets:
+    @staticmethod
+    def path(filename):
+        dirname = os.path.join(os.path.dirname(__file__), "datasets")
+        return os.path.join(dirname, filename)
+
+    @classmethod
+    def data_1(cls):
```
astaric: These methods could use more descriptive names. maybe missing_data_1? Telling what exactly is in the dataset would probably make them longer that 50 characters.

Orange/widgets/tests/base.py
```diff
@@ -1,11 +1,13 @@
...
 import unittest
-import sip
+
 
 import numpy as np
 from AnyQt.QtWidgets import (
     QApplication, QComboBox, QSpinBox, QDoubleSpinBox, QSlider
 )
 
+import Orange.data
```
astaric: Table is already imported in the line 15. Maybe the rest of the symbols could be imported in the same fashion?

Orange/widgets/tests/base.py
ales-erjavec: I agree, but cant think of a naming scheme besides missing_data_{1,2,3}.

astaric: So maybe just missing_data_n (with matching/modified filenames of the test files)?

I am afraid that otherwise the pattern will continue and we will get data_{4,5,6,7,8} in the future, making it impossible to tell what kind of data is used in the test.

ales-erjavec: Ok

PR #1941, opened 2017-01-23
[FIX] OWMosaic: Fix crash for attribute with no values

### LINKED ISSUE #1940: ValueError:  minlength must be positive
https://sentry.io/biolab/orange3/issues/206599981/

```
ValueError:  minlength must be positive
(5 additional frame(s) were not displayed)
...
  Module "Orange.statistics.distribution", line 295, in get_distribution

    return Discrete(dat, variable, unknowns)
  Module "Orange.statistics.distribution", line 39, in __new__

    return cls.from_data(dat, variable)
  Module "Orange.statistics.distribution", line 63, in from_data

    dist, unknowns = data._compute_distributions([variable])[0]
  Module "Orange.data.table", line 1314, in _compute_distributions

    dist, unknowns = bincount(m, len(var.values) - 1, W)
  Module "Orange.statistics.util", line 52, in bincount

    minlength=minlength).astype(float),
```

VesnaT:
##### Issue

Fixes #1940

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -898,6 +898,13 @@ class OWMosaicDisplay(OWWidget):
         else:
             self.Warning.no_valid_data.clear()
 
+        attrs = [attr for attr in attr_list if not data.domain[attr].values]
+        if attrs:
+            CanvasText(self.canvas,
+                       "Feature {} has no values".format(attrs[0]),
+                       (self.canvas_view.width() - 120) / 2,
+                       self.canvas_view.height() / 2)
+            return
         if self.interior_coloring == self.PEARSON:
             apriori_dists = [get_distribution(data, attr) for attr in attr_list]
         else:
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -47,10 +47,10 @@ def bincount(X, max_val=None, weights=None, minlength=None):
         nans = 0. if X.ndim == 1 else np.zeros(X.shape[1], dtype=float)
     if minlength is None and max_val is not None:
         minlength = max_val + 1
-    return (np.bincount(X.astype(np.int32, copy=False),
-                        weights=weights,
-                        minlength=minlength).astype(float),
-            nans)
+    bc = np.array([]) if minlength is not None and minlength <= 0 else \
+        np.bincount(X.astype(np.int32, copy=False),
+                    weights=weights, minlength=minlength).astype(float)
+    return bc, nans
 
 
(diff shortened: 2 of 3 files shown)


astaric:
Not sure how to test this. What I tried was adding a discrete variable with all missing values (using feature constructor) and using that in mosaic. It still crashed, but it might be a different problem.

VesnaT:
Looks like that is another problem.

To test this use an existing dataset with feature with all missing values, since the problem occurred when the variable's "values" is empty.
It is impossible to construct discrete feature with no values with 
feature constructor.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -899,6 +899,13 @@ def draw_legend(x0_x1, y0_y1):
             self.Warning.no_valid_data.clear()
 
         if self.interior_coloring == self.PEARSON:
+            attrs = [attr for attr in attr_list if not data.domain[attr].values]
+            if attrs:
+                CanvasText(self.canvas,
+                           "Feature {} has no values".format(attrs[0]),
+                           (self.canvas_view.width() - 120) / 2,
+                           self.canvas_view.height() / 2)
+                return
```
janezd: I would put this before `if`, that is, fail in both cases, with Pearson or with class distribution. The visualization does not make sense if any attribute has no values.

janezd: screen shot 2017-01-24 at 21 13 31

Orange/widgets/visualize/owmosaic.py
```diff
@@ -948,6 +955,12 @@ def get_max_label_width(attr):
                 get_conditional_distribution(subset, attr_list)
 
         # draw rectangles
+        if self.interior_coloring != self.PEARSON and not class_var.values:
+            CanvasText(self.canvas,
+                       "Class variable {} has no values".format(class_var),
+                       (self.canvas_view.width() - 140) / 2,
+                       self.canvas_view.height() / 2)
+            return
```
janezd: And then you can remove this.

janezd:
@astaric, try this data: [x.tab.zip](https://github.com/biolab/orange3/files/727860/x.tab.zip)

lanzagar:
I am merging this, but two bugs in Mosaic still remain (unrelated to this PR):
- using `anneal` data and choosing feature `bt` crashes (I think this is what @astaric noticed)
- Find Informative Mosaics (also on anneal) does not sort correctly (nan scores incomparable) - I think @janezd fixed the same issue in another widget's vizrank?

PR #1945, opened 2017-01-24
ErrorReporting: Escape string representation of value

astaric:


astaric:
https://sentry.io/biolab/orange3/issues/207211742/

PR #1949, opened 2017-01-24
[FIX] Catch errors during learning in learner widgets

janezd:
##### Issue

The base widget for learners did not intercept errors occurring during fitting. 

##### Description of changes

Errors are now caught and a basic message that something went wrong is displayed. Widgets can provide more specific error messages when possible.

Some widgets tested `self.valid_data` to verify that fitting succeeded. This did not really work. Although it now would, the tests are replaced to check whether `self.model` is not `None`. This is more appropriate since the data may be valid when the model is not.

A minor fix needed for tests: `Model` checked whether the `domain` parameter is given by checking `if domain`. This is wrong since `Domain` does not have `__bool__`, so `__len__` is used instead, and so `Domain([])` is treated as `False`.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -178,7 +178,7 @@ class Model(Reprable):
     def __init__(self, domain=None):
         if isinstance(self, Learner):
             domain = None
-        elif not domain:
+        elif domain is None:
             raise ValueError("unspecified domain")
         self.domain = domain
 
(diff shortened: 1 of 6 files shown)


astaric:
https://sentry.io/biolab/orange3/issues/206321113/

astaric:
https://sentry.io/biolab/orange3/issues/206320798/

astaric:
https://sentry.io/biolab/orange3/issues/206320389/

jerneju:
https://sentry.io/biolab/orange3/issues/196743657/

PR #1952, opened 2017-01-25
[FIX] DataSampler: Fix crash when stratifying unbalanced datasets

astaric:
##### Issue
DataSampler crashed when stratified was selected, but one class contained only one element. Bootstrap did not work (invalid signature).

##### Description of changes
If stratified sampling fails, show a warning and sample without.
Add argument data to Bootstrap to satisfy signature.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owdatasampler.py
+++ b/Orange/widgets/data/owdatasampler.py
@@ -46,6 +46,9 @@ class OWDataSampler(OWWidget):
     number_of_folds = Setting(10)
     selectedFold = Setting(1)
 
+    class Warning(OWWidget.Warning):
+        could_not_stratify = Msg("Stratification failed\n{}")
+
     class Error(OWWidget.Error):
         too_many_folds = Msg("Number of folds exceeds data size")
         sample_larger_than_data = Msg("Sample must be smaller than data")
@@ -252,10 +255,17 @@ class OWDataSampler(OWWidget):
             self.indices = None
             return
 
-        rnd = self.RandomSeed if self.use_seed else None
         stratified = (self.stratify and
                       type(self.data) == Table and
                       self.data.domain.has_discrete_class)
+        try:
+            self.indices = self.sample(data_length, size, stratified)
+        except ValueError as ex:
+            self.Warning.could_not_stratify(str(ex))
+            self.indices = self.sample(data_length, size, stratified=False)
+
+    def sample(self, data_length, size, stratified):
+        rnd = self.RandomSeed if self.use_seed else None
         if self.sampling_type == self.FixedSize:
             self.indice_gen = SampleRandomN(
                 size, stratified=stratified, replace=self.replacement,
@@ -268,7 +278,7 @@ class OWDataSampler(OWWidget):
         else:
             self.indice_gen = SampleFoldIndices(
                 self.number_of_folds, stratified=stratified, random_state=rnd)
-        self.indices = self.indice_gen(self.data)
+        return self.indice_gen(self.data)
 
     def send_report(self):
         if self.sampling_type == self.FixedProportion:
--- a/Orange/widgets/data/owdatasampler.py
+++ b/Orange/widgets/data/owdatasampler.py
@@ -385,7 +385,14 @@ class SampleBootstrap(Reprable):
         self.size = size
         self.random_state = random_state
 
-    def __call__(self):
+    def __call__(self, table=None):
+        """Bootstrap indices
+
+        Args:
+            table: Not used (but part of the signature)
+        Returns:
+            tuple (out_of_sample, sample) indices
+        """
         rgen = np.random.RandomState(self.random_state)
         sample = rgen.randint(0, self.size, self.size)
         sample.sort()  # not needed for the code below, just for the user
--- a/Orange/widgets/data/tests/test_owdatasampler.py
+++ b/Orange/widgets/data/tests/test_owdatasampler.py
@@ -12,7 +12,7 @@ class TestOWDataSampler(WidgetTest):
         cls.iris = Table("iris")
 
     def setUp(self):
-        self.widget = self.create_widget(OWDataSampler)
+        self.widget = self.create_widget(OWDataSampler)  # type: OWDataSampler
 
     def test_error_message(self):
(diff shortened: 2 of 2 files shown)


astaric:
https://sentry.io/biolab/orange3/issues/206658574/

astaric:
https://sentry.io/biolab/orange3/issues/208834147/

ajdapretnar:
This works well for me, however Bootstrap always returns the same number of instances as on the input. I don't think this is the intended behaviour.

lanzagar:
That is the intended behaviour. This method samples with replacement, so the size is always the same, but the instances in the sample are different (some are repeated).
But it is different from other sampling options, and maybe we should make it clearer to the user (e.g. tooltip; a horizontal rule after other methods and/or small caption Sampling with replacement; or even just rename to "Bootstrap sampling with replacement).

For this PR, however, I think this could be an excellent addition to the new tests for bootstrap - assert that len(sample) == len(input). So we will know in the future that this was indeed intended.

astaric:
done. Also added comments clarifying the asserts.

Orange/widgets/data/owdatasampler.py
```diff
@@ -46,6 +46,9 @@ class OWDataSampler(OWWidget):
     number_of_folds = Setting(10)
     selectedFold = Setting(1)
 
+    class Warning(OWWidget.Warning):
+        could_not_stratify = Msg("Could not stratify\n{}")
```
janezd: Could you turn this message into a sentence, something like "Stratification failed"?

Orange/widgets/data/tests/test_owdatasampler.py
```diff
@@ -26,3 +26,36 @@ def test_error_message(self):
...
+
+        # Bootstrap should sample len(input) instances
+        self.assertEqual(len(sample), len(self.iris))
+        # Sample and remaining should cover all instances, while none
+        # should be present in both
+        self.assertEqual(len(in_sample | in_remaining), len(in_input))
+        self.assertEqual(len(in_sample & in_remaining), 0)
+        # Sampling with replacement will always produce at least one distinct
+        # instance in sample, and at least one instance in remaining with
+        # high probability (1-(1/150*2/150*...*145/150) ~= 1-2e-64)
```
janezd: Then you should probably use `self.assertProbablyGreater`? :)

Orange/widgets/data/tests/test_owdatasampler.py
```diff
@@ -12,7 +12,7 @@ def setUpClass(cls):
         cls.iris = Table("iris")
 
     def setUp(self):
-        self.widget = self.create_widget(OWDataSampler)
+        self.widget = self.create_widget(OWDataSampler)  # type: OWDataSampler
```
astaric: I saw this in tests aleš writes :)

Orange/widgets/data/tests/test_owdatasampler.py
astaric: I do not like the implicit nature of self.assertProbablyGreater, as it is not obvious how the check is performed. I am more inclined towards a decorator

@should_pass(3, out_of=5)["times"]

that would run the test for the required amount of retries and then compute the probability of it being correct.

astaric: But just in case we ever need it:
```python
def assertProbablyGreater(self, a, b, msg=None):
    """Assert that a in greater than b in at least one out of two runs."""
    import inspect

    try:
        return self.assertGreater(a, b, msg)
    except AssertionError:
        pass

    frames = inspect.stack()[1:]
    test_frames = [f for f in frames if f.function.startswith("test_")]
    if any(f for f in frames if f.function == "c") or not test_frames:
        # Retried, but still not working or nothing to retry. Fail
        return self.assertGreater(a, b, msg)

    getattr(self, test_frames[0].function)()
```

janezd: First, your decorator, cunning as it is, still leaves a non-zero probability that the test fails. To increase the probability the it will work, use

```
@should_pass(9, out_of=10)["times"]
@should_pass(3, out_of=5)["times"]
```

Also, it should be called `should_pass_at_least`.

Second, I believe your formula is incorrect. Denominators should go to 150; the probability of failing is 150! / 150^150. (I guess you probably used the correct formula; this indeed gives 2.2e-64).

Orange/widgets/data/tests/test_owdatasampler.py
```diff
@@ -26,3 +26,36 @@ def test_error_message(self):
...
+        # Bootstrap should sample len(input) instances
+        self.assertEqual(len(sample), len(self.iris))
+        # Sample and remaining should cover all instances, while none
+        # should be present in both
+        self.assertEqual(len(in_sample | in_remaining), len(in_input))
+        self.assertEqual(len(in_sample & in_remaining), 0)
+        # Sampling with replacement will always produce at least one distinct
+        # instance in sample, and at least one instance in remaining with
+        # high probability (1-(1/150*2/150*...*145/150) ~= 1-2e-64)
+        self.assertGreater(len(in_sample), 0)
```
janezd: Third, isn't `len(in_sample)` always at least 1? :)

Orange/widgets/data/tests/test_owdatasampler.py
janezd: > But just in case we ever need it:

I don't think so. :) Please just check the "third" remark before merging, the one about `len(in_sample)` being at least 1. I actually meant that one.

Orange/widgets/data/tests/test_owdatasampler.py
astaric: It is, it is also stated in the comment above:
`will always produce at least one distinct instance in sample`
The high probability part refers to the second assertion
`and at least one instance in remaining with high probability`

Orange/widgets/data/tests/test_owdatasampler.py
astaric: > Second, I believe your formula is incorrect. 

Nice catch. I have updated the formula.

Issue #1899, opened 2017-01-09
It is no longer possible to add comments in reports


lopezco:
##### Orange version
3.3.10.dev0+917b3cb

##### Expected behavior
User can add comments in reports

##### Actual behavior
It is no longer possible to add comments in reports

##### Steps to reproduce the behavior
1. Open Orange
2. Add a widget
3. Click Report

ajdapretnar:
I'd add that report for File also fails to report properly on the file name and format. I've tried housing.tab and titanic.tab and got unknown in both cases.

lopezco:
I've tried to add a comment after the update but couldn't do it because the textbox doesn't appear.

**Version:** 3.4.dev0+5373929

kernc:
Closed incorrectly via commit log, sorry.

PR #1956, opened 2017-01-25
[FIX] report comments and OWFile reporting filename

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/1899

##### Description of changes
Not sure what the problem was re comments. Removing `required=required` in HTML now works for me. in OWFile, `self.loaded_file` is now correctly set when the file is loaded, not when browsed for.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/report/index.html
+++ b/Orange/canvas/report/index.html
@@ -99,7 +99,7 @@
     }
 
     textarea {
-        border: none;
+        border: 1px solid silver;
         font-style: italic;
         width: 100%;
         height: 4em;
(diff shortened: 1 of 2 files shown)


ajdapretnar:
All looks good except for one thing. When I create a report, write some comments and save the report, the final comment isn't saved.
The rest works great.

kernc:
Need more reproductions steps. Can't seem to replicate. The comment saves, or, if unsaved, asks to be.

ajdapretnar:
I loaded iris.tab, reported File, reported Distributions, reported Box Plot. Wrote comments in each section and saved the report. Comments in File and Distributions were there upon opening html, while the comment for Box Plot disappeared.

I tested this three times, once also with only two reported widgets and it's the same - the final comment is missing. We can check tomorrow.

kernc:
I can't reproduce (the issue).

kernc:
@ajdapretnar Please try if this now works.

PR #1961, opened 2017-01-26
[FIX] File: Fix crash when last_path is None

janezd:
https://sentry.io/biolab/orange3/issues/196743788/

The error also appeared when I tried to reproduce some unrelated bugs.

##### Description of changes

Check whether `last_path` is `None` before checking its size.

##### Includes
- [X] Code changes
- [X] Tests


janezd:
https://sentry.io/biolab/orange3/issues/196743788/

astaric:
Could you also write a test that reproduces the problem?

janezd:
I didn't think so. It turned out it's trivial.

PR #1971, opened 2017-01-27
[FIX] OWPolynomialRegression: Send learner to output even when no data

pavlin-policar:
##### Issue
Learner was not on output signal unless there was valid data. This is not consistent with the rest of Orange's learner widgets.

##### Description of changes
Fix the issue and some general improvements to code style.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/regression/owunivariateregression.py
+++ b/Orange/widgets/regression/owunivariateregression.py
@@ -1,21 +1,20 @@
-from AnyQt.QtWidgets import QSizePolicy
-from AnyQt.QtGui import QColor, QPen, QFont, QPalette
-from AnyQt.QtCore import Qt, QRectF
-
-import pyqtgraph as pg
 import numpy as np
+import pyqtgraph as pg
+from AnyQt.QtCore import Qt, QRectF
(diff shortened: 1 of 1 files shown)


Orange/widgets/regression/owunivariateregression.py
```diff
@@ -186,20 +182,19 @@ def plot_regression_line(self, x_data, y_data):
...
+            learner=LinearRegressionLearner() if self.learner is None else self.learner)
 
+        if self.data is not None:
             attributes = self.x_var_model[self.x_var_index]
             class_var = self.y_var_model[self.y_var_index]
-            data_table = Table(Domain([attributes], class_vars=[class_var]), self.data)
+            data_table = Table(Domain([attributes], class_vars=[class_var]),
+                               self.data)
 
             learner.name = self.learner_name
```
janezd: Doesn't this belong before `if`, too? The name should be set even if there's no data.

pavlin-policar:
I closed this since the Polynomial regression widget has been moved to the educational addon, and will be deleted when I merge the widgets in #1994

PR #1976, opened 2017-01-28
K-Means improvements

janezd:
Based on #1974 (6c0a8d1 belongs there, not here).

##### Description of changes

- Use custom model for showing optimization results instead of a standard model. This replaced a messy function for setting up the items in the view with a cleaner model.
- Remove the options for selecting the output name and type. The name is now always "Cluster", and the column is always the class column.
- Minor refactoring and reformatting.

No new tests are introduced.

##### Includes
- [X] Code changes


Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -423,6 +429,8 @@ def set_data(self, data):
             self.run()
 
     def send_report(self):
+        # False positives (Setting is not recognized as int)
+        # pylint: disable=invalid-sequence-index
```
astaric: Did you consider ignoring this for the whole class/file?

janezd: I prepared two more commits (the last is not yet completed and pushed), and this remains only in two places, so I'd prefer being conservative wrt disabling warnings.

janezd:
@ales-erjavec, I need your Qt expertise (again). Why does `self.table_view.sizeHint().width()` give a wrong width in this widget? It currently sets column widths with `QTableView.setColumnWidth`, but I also unsuccessfully tried size hint role in the model and the size hint in the delegate.

ales-erjavec:
None of the generic Qt item view widgets propagate the size hints from model/delegates to their own `sizeHint` property. If you need that you will have to subclass QTableView and reimplement the sizeHint method.

janezd:
This makes perfect sense. I'm so stupid. I spent hours and hours with this. Thanks.

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -49,333 +142,277 @@ class Error(widget.OWWidget.Error):
...
+        try:
+            self.optimization_runs[k] = KMeans(
+                n_clusters=k,
+                init=['random', 'k-means++'][self.smart_init],
+                n_init=self.n_init,
+                max_iter=self.max_iterations,
+                compute_silhouette_score=True)(self.data)
+        except BaseException as exc:
```
astaric: Is there a special reason why BaseException is needed or would Exception suffice?

janezd:
Why has this failed!?

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -27,355 +67,254 @@ class OWKMeans(widget.OWWidget):
...
         layout.addWidget(sb, 0, 1)
         gui.lineEdit(
             sb, self, "n_init", controlWidth=60,
-            valueType=int, validator=QIntValidator(),
-            callback=self.update)
-        layout.addWidget(gui.widgetLabel(None, "Maximal iterations: "),
-                         1, 0, Qt.AlignLeft)
+            valueType=int, validator=QIntValidator(), callback=self.invalidate)
+        layout.addWidget(
+            gui.widgetLabel(None, "Maximal iterations: "), 1, 0, Qt.AlignLeft)
```
lanzagar: This was not changed and probably does not need fixing (maximal should be correct in this use), but I think "Maximum iterations" is a bit more standard...

PR #1977, opened 2017-01-29
[FIX] Fitter: Fix infinite recursion in __getattr__

pavlin-policar:
##### Issue
Running fitters on large enough datasets through the test and score widget and setting it to cross validation would result in a stack overflow error.

I am not particularly sure why this happened, my theory is that when test and score receives a large enough dataset, it runs cross validation in parallel, and that the learner attributes were probably accessed statically and failing.

Unfortunately, I have no idea what caused this and how exactly it was called, so I don't really know how to test against this. But since I will most likely change this in the near future, I wouldn't spend too much time on this.

##### Description of changes
No longer rely on `self.kwargs` being on the object, but raise attribute error by explicitly getting `kwargs` with `__getattribute__`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/classification/tree.py
+++ b/Orange/classification/tree.py
@@ -64,6 +64,8 @@ class TreeLearner(Learner):
         self.min_samples_split = min_samples_split
         self.sufficient_majority = sufficient_majority
         self.max_depth = max_depth
+        self.params = {k: v for k, v in vars().items()
+                       if k not in ('args', 'kwargs')}
 
     def _select_attr(self, data):
         """Select the attribute for the next split.
--- a/Orange/classification/tree.py
+++ b/Orange/classification/tree.py
@@ -59,13 +59,12 @@ class TreeLearner(Learner):
             min_samples_leaf=1, min_samples_split=2, sufficient_majority=0.95,
             **kwargs):
         super().__init__(*args, **kwargs)
-        self.binarize = binarize
-        self.min_samples_leaf = min_samples_leaf
-        self.min_samples_split = min_samples_split
-        self.sufficient_majority = sufficient_majority
-        self.max_depth = max_depth
-        self.params = {k: v for k, v in vars().items()
-                       if k not in ('args', 'kwargs')}
+        self.params = {}
+        self.binarize = self.params['binarize'] = binarize
+        self.min_samples_leaf = self.params['min_samples_leaf'] = min_samples_leaf
+        self.min_samples_split = self.params['min_samples_split'] = min_samples_split
+        self.sufficient_majority = self.params['sufficient_majority'] = sufficient_majority
+        self.max_depth = self.params['max_depth'] = max_depth
 
     def _select_attr(self, data):
         """Select the attribute for the next split.
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -114,18 +114,20 @@ class Learner(_ReprableWithPreprocessors):
                             self.__class__.__name__)
 
         self.domain = data.domain
-
-        if type(self).fit is Learner.fit:
-            model = self.fit_storage(data)
-        else:
-            X, Y, W = data.X, data.Y, data.W if data.has_weights() else None
-            model = self.fit(X, Y, W)
+        model = self._fit_model(data)
         model.domain = data.domain
         model.supports_multiclass = self.supports_multiclass
         model.name = self.name
         model.original_domain = origdomain
         model.original_data = origdata
         return model
+
+    def _fit_model(self, data):
+        if type(self).fit is Learner.fit:
+            return self.fit_storage(data)
+        else:
+            X, Y, W = data.X, data.Y, data.W if data.has_weights() else None
+            return self.fit(X, Y, W)
 
     def preprocess(self, data):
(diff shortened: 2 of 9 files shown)


kernc:
1. joblib tries to pickle its arguments.
2. pickling accesses `obj.__getstate__` for "instructions".
3. `__getattr__` handler calls `self.get_learner(None)` since `self.problem_type` is not yet set, since we haven't yet fitted anything.
4. This raises `AttributeError` in `get_learner()`.
5. Python thinks `obj` doesn't have the attribute `__getattr__` was called with, so it calls `__getattr__` with it.
...

I guess the comment in `get_learner()` is wrong and a `TypeError` should be raised. With this change, _test_evaluation_testing.TestCrossValidation.test_njobs()_ with learner set to one of the fitters doesn't crash anymore.

pavlin-policar:
Wow, thank you very much. I don't think I would've figured that out by myself.

pavlin-policar:
So I noticed this was failing in some cases, so I tried to try to make things a bit more straightforward.

Note: This is based on #1968 because I made some changes to the learner widget tests that enable testing fitters, which turned out to be useful here as well.

I've made two rather big changes from last time:
1. To make `Learners` simpler in general, we now assume that they will always have a `params` attribute (this is added to the main learner docstring, but I don't really have a good way to enforce this). The only change this required was on Tree learners, since everything else already had params. It may be nicer to have all the learner params as actual attributes on the learners, which could probably be done easily, but this required less changes. @kernc what do you think about this? The api on learners is still kind of a mess, but this could be an improvement.
2. I've removed all state from Fitters. Storing the problem type on the fitter was causing all kinds of unpredictable behaviour and was difficult to understand, so I've removed that entirely. @janezd what do you think about this? This should make the fitters simpler.

This does also fix the error that @lanzagar pointed out last week.

Orange/classification/tree.py
```diff
@@ -64,6 +64,8 @@ def __init__(
         self.min_samples_split = min_samples_split
         self.sufficient_majority = sufficient_majority
         self.max_depth = max_depth
+        self.params = {k: v for k, v in vars().items()
+                       if k not in ('args', 'kwargs')}
```
janezd: What is in `vars()` -- except globals, which we don't like anyway?! I don't like this blind copying.

pavlin-policar:
Sorry this has taken me so long to get to, but I'm at it again.

I'm guessing you meant changes related to #1968. The only thing these two have in common is some shared code to make the tests work with fitters.

Orange/classification/tree.py
pavlin-policar: This was an attempt to make the Learner API more intuitive. Currently, Orange learners have their settings (or parameters) saved to the object themselves, whereas sklearn learners have a special `params` field that stores all that data. This field is used heavily in the tests and only rarely in any other code. This makes dealing with the tests a bit confusing. It would probably be better that to switch the field to be protected, and then doing some magic with `__getattr__` to make those fields accessible to tests. If this were done, the API to the learners would be a bit more straightforward, but if you think it's unnecessary, I can remove it altogether. This would probably be cleaner than the blind copying that I've done here.

janezd:
I will try to merge all your PRs today, so they can be a part of the next release, which is due soon.

The first three commits of this PR are already in the master. So if you just rebase to master, they will disappear here and make it easier to review the code.

Orange/classification/tree.py
janezd: I still don't understand what is the content of `vars`. It contains `binarize`, `max_depth`, `min_samples_leaf`, `min_samples_splet` and `sufficient_majority`; if these need to be copied to `params`, I'd copy them explicitly. Besides, it probably contains `self`, and besides that, all global names in the module, which you don't want to copy.

I am probably missing something here.

pavlin-policar:
Yeah, I'm sorry, I had forgotten to rebase, it's done now.

Orange/classification/tree.py
pavlin-policar: No you're not missing anything at all, I simply tried to emulate what the constructors in the sklearn learners were doing, which use `vars()`. I'll change it so the params are explicitly copied. I suppose the same issue is then present in all the sklearn learners. I could fix that if you'd like, in a separate PR, this is kind of a mess already.

janezd: You're right, it's uglier (although perhaps better).

Leave it like this for now, and perhaps change this in all SKL learners. Perhaps by adding something like this to `SKLLearner`

    from inspect import signature

    ...

    def update_params(self, values):
        param_names = signature(type(self).__init__).parameters
        self.params.update({name: values[name] for name in param_names if name not in {"args", "kwargs"})

which would then be called from `__init__` of derived classes with `self.update_params(vars())`. This would only copy the names that were given as arguments.

I see that there's already some similar magic in our SKL wrappers.

Not in this PR.

ajdapretnar:
This fixes the issue I had when training a classification tree on [employee attrition data](https://www.ibm.com/communities/analytics/watson-analytics-blog/hr-employee-attrition/).

PR #1983, opened 2017-02-01
[FIX] WebviewWidget: WebEngine Don't Grab Focus on setHtml

nikicc:
##### Issue

If running on WebEngine the call to `setHtml` grabs focus and sets it to WebViewWidget. This is quite annoying behaviour when for example one would like to filter some data set and show the results in WebView. Due to WebViewWidget's focus grabbing the input filter looses focus after every character is typed.

##### Description of changes
Override `setHtml` to that the widget's state is first set to disabled, then call `super().setHtlm` and finally restore the initial state.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/webview.py
+++ b/Orange/widgets/utils/webview.py
@@ -518,3 +518,18 @@ elif HAVE_WEBENGINE:
         def exposeObject(self, name, obj):
             obj = _to_primitive_types(obj)
             self._jsobject_channel.send_object(name, obj)
+
+        def setHtml(self, html, base_url=''):
+            # TODO: remove once anaconda will provide PyQt without this bug.
+            #
+            # At least on some installations of PyQt 5.6.0 with anaconda
+            # WebViewWidget grabs focus on setHTML which can be quite annoying.
+            # For example, if you have a line edit as filter and show results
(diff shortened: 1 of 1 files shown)


lanzagar:
I don't have the mentioned problem (using qt 5.8). But the fix does not seem to break anything for me either. Could you add a TODO or something like that to the comment and mention that this might be fixed in newer versions, so that we can remove the fix in the future if it will not be necessary anymore.

astaric:
I could not reproduce this either (I'm using Qt 5.6).

ajdapretnar:
I have the bug. (Qt 5.6)

nikicc:
@astaric running on anaconda?

nikicc:
@lanzagar I added the comment.

@astaric these are the packages I have installed in my conda env. Are you running the same versions and cannot reproduce? Are you running WebKit or WebEngine?
```
anyqt                     0.0.7                    py35_1    conda-forge
pyqt                      5.6.0                    py35_2
pyqtgraph                 0.10.0                   py35_0    conda-forge
qt                        5.6.2                         0
```

PR #1984, opened 2017-02-02
[FIX] OWScatterMap: Fix crash when displaying constant feature

VesnaT:
##### Issue

Fixes https://sentry.io/biolab/orange3/issues/196744426/

##### Description of changes
Display data when dataset with a single instance or dataset with a constant feature is send to input.
Support missing values as well.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


janezd:
The widget was moved to prototypes and these fixes will be applied there.

PR #1989, opened 2017-02-03
testing: Use sklearn model_selection module instead of cross_validation module

VesnaT:
##### Issue

Sklearn cross_validation module is deprecated and will be removed in 0.20, so module model_selection should be used instead.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -8,7 +8,7 @@ import warnings
 import numpy as np
 
 import joblib
-import sklearn.cross_validation as skl_cross_validation
+import sklearn.model_selection as skl
 
 from Orange.util import OrangeWarning
 from Orange.data import Table, Domain, ContinuousVariable, DiscreteVariable
@@ -507,7 +507,7 @@ class CrossValidation(Results):
     def __init__(self, data, learners, k=10, stratified=True, random_state=0, store_data=False,
                  store_models=False, preprocessor=None, callback=None, warnings=None,
                  n_jobs=1):
-        self.k = k
+        self.k = int(k)
         self.stratified = stratified
         self.random_state = random_state
         if warnings is None:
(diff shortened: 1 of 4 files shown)


Orange/widgets/data/owdatasampler.py
```diff
@@ -331,11 +331,15 @@ def __init__(self, folds=10, stratified=False, random_state=None):
     def __call__(self, table):
         if self.stratified and table.domain.has_discrete_class:
             # XXX: StratifiedKFold does not support random_state
```
astaric: This comment looks misleading :) Could you remove it?

Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -1657,18 +1657,18 @@ def filter_vars(vars):
...
     data = Orange.data.Table("brown-selected")
     data = append_column(data, "M", Orange.data.StringVariable("Test"),
                          numpy.arange(len(data)).reshape(-1, 1) % 30)
 
-    indices = skl_cross_validation.ShuffleSplit(
-        len(data), n_iter=5, test_size=0.7
-    )
-
-    indices = iter(indices)
+    ss = skl_model_selection.ShuffleSplit(
```
astaric: Could this be changed to use orange ShuffleSplit?

Orange/evaluation/testing.py
```diff
@@ -8,7 +8,7 @@
 import numpy as np
 
 import joblib
-import sklearn.cross_validation as skl_cross_validation
+import sklearn.model_selection as skl_model_selection
```
astaric: What about importing this as "skl"? As it is used a lot, a shorter name might be better. The name "skl.StratifiedKFold" signifies nicely that it is the scikit-learn class being used.

Orange/evaluation/testing.py
```diff
@@ -523,16 +523,20 @@ def setup_indices(self, train_data, test_data):
         self.indices = None
         if self.stratified and test_data.domain.has_discrete_class:
             try:
-                self.indices = skl_cross_validation.StratifiedKFold(
-                    test_data.Y, self.k, shuffle=True, random_state=self.random_state
+                skf = skl_model_selection.StratifiedKFold(
```
astaric: I would rename variables skf, kf, loo, sss, ss and similar below to splitter (so that they would all use the same name).

PR #1990, opened 2017-02-03
[FIX] scatterplot: Fix density image

thocevar:
##### Issue
The density image (crashes) in scatterplot doesn't display with the latest version of pyqtgraph 0.10.

##### Description of changes
Change the dtype of the computed numpy array of pixel values to reflect the actual range of values and prevent "automatic" scaling (crashing) by pyqtgraph.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


astaric:
It works, thanks.

PR #1994, opened 2017-02-03
Finalize Model category

pavlin-policar:
##### Description of changes
This merges together the Mean and Majority learner widgets, and removes the `Classify` and `Regression` category. The Constant widget and new Model category have no icons yet and will be added by @BlazZupan.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
Add replaces for signals. I tried to load the example scheme for classification trees and got: `ValueError("'Classifier' is not a valid output channel name for 'Tree'.",)`

janezd:
In Orange/canvas/report/owreport.py: Fix `from Orange.widgets.classify.owrandomforest import OWRandomForest`

Orange/widgets/model/owunivariateregression.py
```diff
@@ -22,6 +22,9 @@ class OWUnivariateRegression(OWBaseLearner):
     name = "Univariate Polynomial Regression"
     description = "Univariate regression with polynomial expansion."
     icon = "icons/UnivariateRegression.svg"
+    replaces = [
+        "Orange.widgets.regression.owunivariateregression.OWUnivariateRegression",
+    ]
```
janezd: This widget has been moved to educational add-on (and already improved there). Remove it from here. You can add `replaces` to the widget in the education add-on.

pavlin-policar: All right, I was unaware of this. The replaces is already [present](https://github.com/biolab/orange3-educational/blob/master/orangecontrib/educational/widgets/owpolynomialregression.py#L35), so no update needed.

pavlin-policar:
As far as I can tell, most things should be ok, the only things still missing are the icons and the documentation still needs to be adapted. Or have I missed something?

janezd:
@pavlin-policar, pylint failed because of `class OWTreeLearner(OWTreeLearner):` or something similar. I've just found a similar problem in some other PR.

pavlin-policar:
That's very strange, I remember doing something like that before, when the classify and regression categories still existed for some importing reason and it was never an issue then.
But that really shouldn't be an issue here, since I've removed those categories and any instances where I did something like that. I'll try find the issue.

ajdapretnar:
Could you please check that old workflows also work on these new changes? We solved an issue for Classification Tree with #2021, could you just make sure it works well with your new changes too? Thanks!

pavlin-policar:
Yes, I had seen your fix, and I'd copied it for the most part (93d01da). The only difference is that I didn't check if the fitter has a handler for classification and regression, because you really shouldn't be using the fitter if you can't handle both cases and checks are in place to make sure this doesn't happen ([here](https://github.com/pavlin-policar/orange3/blob/move-nonmergable-learners/Orange/modelling/base.py#L12)).

Issue #1997, opened 2017-02-03
Error loading iris.tab dataset when LC_ALL=C


ales-erjavec:
##### Orange version

3.3.11

##### Expected behavior

Can load iris.tab dataset when LC_ALL=C (This is the effective default locale when starting an application on macOS by LaunchServices)

##### Actual behavior

```
/Users/aleserjavec/workspace/orange3/Orange/data/io.py:773: UserWarning: Skipped invalid byte(s) in position 299
  warnings.warn(warning)
Traceback (most recent call last):
  File "/Users/aleserjavec/workspace/orange3/Orange/data/io.py", line 392, in set_table_metadata
    table.attributes = pickle.load(f)
_pickle.UnpicklingError: unpickling stack underflow

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/aleserjavec/workspace/orange3/Orange/data/io.py", line 774, in read
    self.set_table_metadata(self.filename, data)
  File "/Users/aleserjavec/workspace/orange3/Orange/data/io.py", line 399, in set_table_metadata
    for line in f.readlines()))
  File "/Users/aleserjavec/virtual/orange3/bin/../lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 299: ordinal not in range(128)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/aleserjavec/workspace/orange3/Orange/data/table.py", line 198, in __new__
    return cls.from_file(args[0], **kwargs)
  File "/Users/aleserjavec/workspace/orange3/Orange/data/table.py", line 561, in from_file
    data = reader.read()
  File "/Users/aleserjavec/workspace/orange3/Orange/data/io.py", line 779, in read
    raise ValueError('Cannot parse dataset {}: {}'.format(self.filename, error)) from error
ValueError: Cannot parse dataset /Users/aleserjavec/workspace/orange3/Orange/datasets/iris.tab: 'ascii' codec can't decode byte 0xe2 in position 299: ordinal not in range(128)
```

##### Steps to reproduce the behavior

```bash
LC_ALL=C python -c 'import Orange.data; Orange.data.Table("iris.tab")'
```

PR #1998, opened 2017-02-04
[FIX] Loading iris on C locale

kernc:
##### Issue
https://github.com/biolab/orange3/issues/1997

##### Description of changes
Load *.metadata as UTF-8. Additionally, fix reporting filename and filetype in OWFile.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


astaric:
The failing test on windows seems related to the change. Can you take a look?

kernc:
Indeed, encoding should be set also when writing the file. Thanks.

astaric:
If you could add a test or two that checks this, I would be happy to merge. (Mentioning metadata in documentation could also be nice, but it is probably an overkill)

PR #2002, opened 2017-02-07
[FIX] data/io.py Metadata file not saved anymore when it is empty

jerneju:
##### Issue
Do not save the metadata file (when saving the data) if this is empty (0 Bytes)

##### Description of changes
Change if statement and add third condition. It implies that attribute should not be empty. Add test file to check if metadata is saved when empty

##### Includes
- [x] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -373,7 +373,7 @@ class FileFormat(metaclass=FileFormatMeta):
 
     @classmethod
     def write_table_metadata(cls, filename, data):
-        if isinstance(filename, str) and hasattr(data, 'attributes'):
+        if isinstance(filename, str) and getattr(data, 'attributes', None):
             if all(isinstance(key, str) and isinstance(value, str)
                    for key, value in data.attributes.items()):
                 with open(filename + '.metadata', 'w') as f:
--- a/Orange/tests/test_tab_reader.py
+++ b/Orange/tests/test_tab_reader.py
@@ -206,3 +206,23 @@ class TestTabReader(unittest.TestCase):
         table2 = TabReader(table1.__file__).read()
         self.assertEqual(table1.name, 'iris')
         self.assertEqual(table2.name, 'iris')
+
+    def test_metadata(self):
+        tempdir = tempfile.mkdtemp()
+        table = Table("titanic")
+        table.attributes = OrderedDict()
+        table.attributes["a"] = "aa"
+        table.attributes["b"] = "bb"
+        fname = path.join(tempdir, "out.tab")
+        TabReader.write_table_metadata(fname, table)
+        self.assertTrue(path.isfile(fname + ".metadata"))
+        shutil.rmtree(tempdir)
+
+    def test_no_metadata(self):
+        tempdir = tempfile.mkdtemp()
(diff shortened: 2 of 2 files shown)


Orange/tests/test_metadata_save.py
```diff
@@ -0,0 +1,36 @@
...
+class TestMetaData(unittest.TestCase):
+    def setUp(self):
+        DiscreteVariable._clear_cache()
+
+    def test_metadata(self):
+        tempdir = tempfile.mkdtemp()
+        table = Table("iris")
+        fname = path.join(tempdir, "out.tab")
+        TabReader.write_table_metadata(fname, table)
+        self.assertEqual(path.isfile(fname + ".metadata"), True)
```
astaric: You could use self.assertTrue

Orange/tests/test_metadata_save.py
```diff
@@ -0,0 +1,36 @@
...
+        shutil.rmtree(tempdir)
+
+    def test_no_metadata(self):
+        tempdir = tempfile.mkdtemp()
+        table = Table("titanic")
+        fname = path.join(tempdir, "out.tab")
+        TabReader.write_table_metadata(fname, table)
+        self.assertEqual(path.isfile(fname + ".metadata"), False)
```
astaric: You could use self.assertFalse

Orange/data/io.py
```diff
@@ -373,7 +373,7 @@ def write(cls, filename, data):
 
     @classmethod
     def write_table_metadata(cls, filename, data):
-        if isinstance(filename, str) and hasattr(data, 'attributes'):
+        if isinstance(filename, str) and hasattr(data, 'attributes') and data.attributes:
```
astaric:
```
hasattr(data, 'attributes') and data.attributes
```
could be simplified to
```
getattr(data, 'attributes', None)
```

Orange/tests/test_metadata_save.py
```diff
@@ -0,0 +1,36 @@
+# Test methods with long descriptive names can omit docstrings
```
astaric: Is there a special reason this tests were not placed in `Orange/tests/test_tab_reader.py`? They only test TabReaders functionality and all other TabReader tests are in that file.

Orange/tests/test_metadata_save.py
```diff
@@ -0,0 +1,36 @@
...
+from Orange.data import Table, DiscreteVariable
+from Orange.data.io import TabReader
+
+
+class TestMetaData(unittest.TestCase):
+    def setUp(self):
+        DiscreteVariable._clear_cache()
+
+    def test_metadata(self):
+        tempdir = tempfile.mkdtemp()
```
astaric: This tests assume that iris has metadata and titanic has not. If this would change in the future (if we added metadata for titanic as well), the second test would fail and will need to be fixed, even though the change would have nothing to do with the saving of metadata.

I suggest that you manually set the attributes dict to whatever you need in a test and then do the testing. This way, attributes dict will be the one you expect it to be, even if the datasets change in the future.

PR #2003, opened 2017-02-07
util.Reprable: more generic equals (handle array_like types)

kernc:
##### Issue
`repr()` on complex Reprable objects with reprable properties of `np.ndarray` (or similar array_like types) would  fail with standard
```text
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

##### Description of changes
Make equality comparison more robust. Can't believe how hard it is in Python to equals-compare two object's values in a generic fashion!

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/util.py
+++ b/Orange/util.py
@@ -8,6 +8,8 @@ from itertools import chain, count
 from collections import OrderedDict
 import warnings
 
+import numpy as np
+
 # Exposed here for convenience. Prefer patching to try-finally blocks
 from unittest.mock import patch  # pylint: disable=unused-import
 
@@ -234,7 +236,7 @@ class Reprable:
                     param.kind not in (param.VAR_POSITIONAL,
                                        param.VAR_KEYWORD)):
                     value = getattr(self, param.name)
-                    if value != param.default:
+                    if not self.__equal(value, param.default):
                         names_values.append((param.name, value))
 
         module = self._reprable_module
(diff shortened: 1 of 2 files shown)


Orange/util.py
```diff
@@ -246,6 +248,13 @@ def __repr__(self):
                                  ', '.join('{}={!r}'.format(*pair)
                                            for pair in names_values))
 
+    @staticmethod
+    def __equal(obj1, obj2):
+        try:
+            return obj1.size == obj2.size and np.all(obj1 == obj2)
+        except AttributeError:
+            return np.all(obj1 == obj2)
```
astaric: Why is np.all needed here?

kernc: Because just one of the two objects might be an array_like. Comparing it to a non-array_like object, this still produces an array of bools. The above try case only handled both objects being array_like _and of same size_, hereby uncovering another flaw in my reasoning; already fixed up.

astaric: Could the reasoning perhaps be written down in a comment (or two), because the code it not that obvious.

kernc: Sure thing. It wasn't obvious. :smile: Perhaps give this a little more time, I'm not convinced it's correct. :laughing:

astaric: Adding an explicit test (or more) for __equal might help :)

Although, you might have to consider renaming it to _equal as testing __private stuff is tricky.

kernc: Or I could construct complex objects and use the designated `__repr__()` interface. I'm happy to do neither here. The numpy case is covered and tested. It should work until it won't and we'll figure it out then. :stuck_out_tongue_closed_eyes:

PR #2006, opened 2017-02-10
[FIX] OWMosaic:  Fix crash for empty column

rokgomiscek:
##### Issue
Widget crashes if the columns are empty.
To reproduce: Connect data to Select Columns. Remove features, target variable and meta attributes. Connect to Mosaic Display.

##### Description of changes
Added an extra check in in set_data().

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -353,6 +353,23 @@ class OWMosaicDisplay(OWWidget):
         else:
             self.update_graph()
 
+    def _get_discrete_data(self, data):
+        """
+        Discretizes continuous attributes.
+        Returns None when there is no data, no rows, or no discrete or continuous attributes.
+        """
+        if (data is None or
+                not len(data) or
+                not any(attr.is_discrete or attr.is_continuous
+                        for attr in chain(data.domain, data.domain.metas))):
+            return None
+        elif any(attr.is_continuous for attr in data.domain):
+            return Discretize(
+                method=EqualFreq(n=4), remove_const=False, discretize_classes=True,
+                discretize_metas=True)(data)
+        else:
+            return data
+
     def init_combos(self, data):
         for combo in self.attr_combos:
             combo.clear()
@@ -400,17 +417,7 @@ class OWMosaicDisplay(OWWidget):
         self.closeContext()
         self.data = data
         self.init_combos(self.data)
-        if (self.data is None or
-                not len(self.data) or
-                not any(attr.is_discrete or attr.is_continuous
-                        for attr in chain(data.domain, data.domain.metas))):
-            self.discrete_data = None
-        elif any(attr.is_continuous for attr in data.domain):
-            self.discrete_data = Discretize(
-                method=EqualFreq(n=4), remove_const=False, discretize_classes=True,
-                discretize_metas=True)(data)
-        else:
-            self.discrete_data = self.data
+        self.discrete_data = self._get_discrete_data(self.data)
 
         self.vizrank.stop_and_reset()
         self.vizrank_button.setEnabled(
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -400,8 +400,10 @@ class OWMosaicDisplay(OWWidget):
         self.closeContext()
         self.data = data
         self.init_combos(self.data)
-        if self.data is None or not len(self.data) or not  \
-                any(attr.is_discrete or attr.is_continuous for attr in chain(data.domain, data.domain.metas)):
+        if (self.data is None or
+                not len(self.data) or
+                not any(attr.is_discrete or attr.is_continuous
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owmosaic.py
```diff
@@ -400,7 +400,7 @@ def set_data(self, data):
         self.closeContext()
         self.data = data
         self.init_combos(self.data)
-        if self.data is None or not len(self.data):
+        if self.data is None or not len(self.data) or not len(self.data[0]):
```
janezd: I think it would be better to test this on domain, not data.

Besides, meta attributes are useful only if they are discrete or continuous, not strings. I think you should check this, too.

Orange/widgets/visualize/tests/test_owmosaic.py
```diff
@@ -26,6 +26,10 @@ def test_no_data(self):
         """Check that the widget doesn't crash on empty data"""
         self.send_signal("Data", self.data[:0])
 
+    def test_empty_column(self):
+        """Check that the widget doesn't crash if the columns are empty"""
+        self.send_signal("Data", self.data[:,:0])
+
```
janezd: Add a test with a single string meta attribute.

Orange/widgets/visualize/owmosaic.py
rokgomiscek: Changed.

Orange/widgets/visualize/tests/test_owmosaic.py
```diff
@@ -40,6 +44,13 @@ def test_continuous_metas(self):
                      metas=np.arange(6).reshape(6, 1))
         self.send_signal("Data", data)
 
+    def test_string_meta(self):
+        """Check widget for dataset with only one string meta"""
+        domain = Domain([], metas=[StringVariable("m")])
+        data = Table(domain, np.arange(6).reshape(6, 1)[:, :0],
```
lanzagar: Is `np.arange(6).reshape(6, 1)[:, :0]` supposed to produce `np.empty((6,0))` ?

Orange/widgets/visualize/owmosaic.py
```diff
@@ -400,7 +400,8 @@ def set_data(self, data):
         self.closeContext()
         self.data = data
         self.init_combos(self.data)
-        if self.data is None or not len(self.data):
+        if self.data is None or not len(self.data) or not  \
+                any(attr.is_discrete or attr.is_continuous for attr in chain(data.domain, data.domain.metas)):
```
lanzagar: Lint says line too long.

Orange/widgets/visualize/tests/test_owmosaic.py
```diff
@@ -26,6 +26,10 @@ def test_no_data(self):
         """Check that the widget doesn't crash on empty data"""
         self.send_signal("Data", self.data[:0])
 
+    def test_empty_column(self):
+        """Check that the widget doesn't crash if the columns are empty"""
+        self.send_signal("Data", self.data[:,:0])
```
lanzagar: Lint would like one space after commas.

Orange/widgets/visualize/tests/test_owmosaic.py
rokgomiscek: Yes.... changed :)

Orange/widgets/visualize/owmosaic.py
rokgomiscek: Broke the line.

Orange/widgets/visualize/tests/test_owmosaic.py
rokgomiscek: Added space.

lanzagar:
Lint is still not happy...
Can you fix the 3 minor aesthetic issues so we can merge with all green.
I am partly responsible for the missing space after comma in np.empty - I edited the comment, but you copied it too quickly ;)
The other two are indentation issues - pycharm should auto-indent correctly.

rokgomiscek:
Better? :)

Orange/widgets/visualize/owmosaic.py
```diff
@@ -400,7 +400,10 @@ def set_data(self, data):
         self.closeContext()
         self.data = data
         self.init_combos(self.data)
-        if self.data is None or not len(self.data):
+        if (self.data is None or
```
astaric: Almost there :) Now radon is complaining that this function (which was not exactly a diamond before) looks worse.

I tend to agree, as the data checking part of code is growing larger and larger. I would recommend extraction the whole if-elif-else block into a method called something like "_get_discrete_data" that gets the data and returns data that consists only of discrete variables. The whole block would then turn into `self.discrete_data = self._get_discrete_data(data)`

+points if the new method has a docstring :)

rokgomiscek: Moved the block to a new method (with a docstring).

PR #2024, opened 2017-02-17
OWFile: Show error msg when file doesn't exists

VesnaT:
##### Issue

When opening a workflow whose dataset is not found, 'Iris' dataset opens.
Error message should be shown instead.

##### Description of changes
Show error msg when file doesn't exists instead of opening next available file in the list
Nonexistent file name is colored red.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/data/owfile.py
```diff
@@ -219,6 +222,9 @@ def __init__(self):
 
         if self.source == self.LOCAL_FILE:
             last_path = self.last_path()
+            if last_path and not os.path.exists(last_path):
+                self.Error.file_not_found()
```
lanzagar: Why is the error checking duplicated here? If this is removed, the error is still correctly thrown in `load_data`, right?
Of course you need to check that the path exists in the next `if` otherwise you can't safely check the size... But the cyclomatic complexity would still be lower by one ;)

PR #2025, opened 2017-02-17
[FIX] owheatmap: Prevent sliders to set Low >= High

jerneju:
##### Issue
AssertionError thrown or error message shown if Low >= High.

##### Description of changes
Prevent horizontal sliders to set Low >= High.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owheatmap.py
+++ b/Orange/widgets/visualize/owheatmap.py
@@ -441,9 +441,6 @@ class OWHeatMap(widget.OWWidget):
             "Not enough instances for k-means merging")
         not_enough_memory = Msg("Not enough memory to show this data")
 
-    class Warning(widget.OWWidget.Warning):
-        threshold_error = Msg("Low slider should be less than High")
-
     def __init__(self):
         super().__init__()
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owheatmap.py
```diff
@@ -1314,15 +1311,21 @@ def update_grid_spacing(self):
...
             self.__fixup_grid_layout()
 
-    def check_threshold_consistency(self):
-        return self.threshold_low < self.threshold_high
+    def update_lowslider(self):
+        if self.controls.threshold_low.value() == 20:
+            self.controls.threshold_low.setSliderPosition(19)
+        if self.controls.threshold_low.value() >= self.controls.threshold_high.value():
+            self.controls.threshold_high.setSliderPosition(self.controls.threshold_low.value() + 1)
+        self.update_color_schema()
```
lanzagar: I don't like that if someone changes the slider parameters (range or steps) 800 lines higher, the hard-coded constants (20 and 19) here would need to be updated too (easy to miss).
How about avoiding the constants and making the code a bit shorter, with a function like:
```python
    def update_lowslider(self):
        low, high = self.controls.threshold_low, self.controls.threshold_high
        if low.value() >= high.value():
            low.setSliderPosition(high.value() - 1)
        self.update_color_schema()
```
And similarly for the highslider below.

Orange/widgets/visualize/tests/test_owheatmap.py
```diff
@@ -98,3 +98,12 @@ def test_not_enough_data_settings_changed(self):
             self.assertFalse(msg.not_enough_features.is_shown())
             self.assertFalse(msg.not_enough_instances.is_shown())
             self.assertFalse(msg.not_enough_instances_k_means.is_shown())
+
+    def test_color_low_high(self):
+        '''
```
lanzagar: For docstrings use triple **double**-quotes `"""` (pep 8, 257)

lanzagar:
Looks good to me. You can squash the commits into one and let's merge.

PR #2028, opened 2017-02-20
Updated Classification Tree schema

### LINKED ISSUE #2005: Example schema for Classification tree
Tree Viewer should send Selected Data, not Data to the Box Plot and Scatter Plot.

ajdapretnar:
##### Issue
Fixes #2005. 

##### Description of changes
Schema set to the new Classification Tree so the learner doesn't fail.
Output of Tree Viewer set to Data Subset for both Box Plot and Scatter Plot.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


astaric:
I would prefer if the existing workflow would work with the new widgets out of the box. It should be possible, if the new widgets prroperly set the replaces attribute.

Examples can/should be updated once the widgets are fixed.

ajdapretnar:
Agreed.

PR #2031, opened 2017-02-20
[FIX] Fix Chi2 computation for variables with values with no instances

jerneju:
Chi-squared test is nan when there are attributes which are not in the data. It is caused by division by zero because  code does not calculate limits. It actually suppose to be 0.

- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owsieve.py
+++ b/Orange/widgets/visualize/owsieve.py
@@ -42,6 +42,7 @@ class ChiSqStats:
         self.expected = np.outer(self.probs_y, self.probs_x) * self.n
         self.residuals = \
             (self.observed - self.expected) / np.sqrt(self.expected)
+        self.residuals = np.nan_to_num(self.residuals)
         self.chisqs = self.residuals ** 2
         self.chisq = float(np.sum(self.chisqs))
         self.p = chi2.sf(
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owsieve.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, data, attr1, attr2):
         self.expected = np.outer(self.probs_y, self.probs_x) * self.n
         self.residuals = \
             (self.observed - self.expected) / np.sqrt(self.expected)
+        where_are_NaNs = np.isnan(self.residuals)
+        self.residuals[where_are_NaNs] = 0
```
astaric: Or you could use

https://docs.scipy.org/doc/numpy/reference/generated/numpy.nan_to_num.html

Orange/widgets/visualize/tests/test_owsieve.py
```diff
@@ -46,3 +51,26 @@ def test_keyerror(self):
         data = Table("iris")
         data = data[0:1]
         self.send_signal("Data", data)
+
+    def test_chisquare(self):
+        """
+         gh-2031
+         Check if it can calculate chi square when there are no attributes which suppose to be.
+        """
+        tempdir = tempfile.mkdtemp()
```
astaric: Instead of creating files in code, you could either commit the file along the test, or even better create the table directly with something along the lines of:

```
a, b = Orange.data.DiscreteVariable("a", values=["y", "n"]), Orange.data.DiscreteVariable("b", values=["y", "n", "other"])
t = Orange.data.Table(Orange.data.Domain([a, b], list(zip("yynny", "ynyyn"))))
```

PR #2039, opened 2017-02-21
[FIX] Prevent PickleError (owfile.py)

jerneju:
##### Issue
PickleError thrown if a column does not have a name.

##### Description of changes
Column name cannot be changed to an empty string or a string with whitespaces.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owfile.py
+++ b/Orange/widgets/data/tests/test_owfile.py
@@ -127,3 +127,15 @@ class TestOWFile(WidgetTest):
         # Open a sample dataset
         self.open_dataset("iris")
         self.assertFalse(self.widget.Error.file_not_found.is_shown())
+
+    def test_check_column_noname(self):
+        """
+        GH-2018
+        """
+        self.open_dataset("iris")
+        idx = self.widget.domain_editor.model().createIndex(1, 0)
+        temp = self.widget.domain_editor.model().data(idx, Qt.DisplayRole)
+        self.widget.domain_editor.model().setData(idx, "   ", Qt.EditRole)
+        self.assertEqual(self.widget.domain_editor.model().data(idx, Qt.DisplayRole), temp)
+        self.widget.domain_editor.model().setData(idx, "", Qt.EditRole)
+        self.assertEqual(self.widget.domain_editor.model().data(idx, Qt.DisplayRole), temp)
--- a/Orange/widgets/utils/domaineditor.py
+++ b/Orange/widgets/utils/domaineditor.py
@@ -79,7 +79,7 @@ class VarTableModel(QAbstractTableModel):
         row, col = index.row(), index.column()
         row_data = self.variables[row]
         if role == Qt.EditRole:
-            if col == Column.name:
+            if col == Column.name and not (value.isspace() or value == ""):
                 row_data[col] = value
             elif col == Column.tpe:
                 vartype = self.name2type[value]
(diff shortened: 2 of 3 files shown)


astaric:
Could you give me the steps to reproduce this issue?

jerneju:
Put File and Scatter Plot widget and then connect them. Then open some data and delete a column name. Click Apply button and close the window and then the application. You do not need to save your work. The Pickle Error should pop up.

jerneju:
@BlazZupan suggested a better way to solve this problem. That is preventing a user to even write a blank column name (string with zero length or only with white spaces). I am now doing some changes in file _domaineditor.py_.

lanzagar:
As it is, this warning should be an error and not allow the user to proceed.
But I think it is better to not burden the user with this at all. If there are columns with empty names in a file, orange could auto-set the names to Feature 1, Feature 2, ...
Also to be more certain we avoid later problems, `name` should be a required argument in Variable's constructor, which should also check that it is non-empty.

jerneju:
@lanzagar: Orange already auto-set the empty names to Feature 1, Feature 2, etc. if there are such columns in a file.

jerneju:
https://sentry.io/biolab/orange3/issues/218075133/

jerneju:
https://sentry.io/biolab/orange3/issues/218561990/

jerneju:
https://sentry.io/biolab/orange3/issues/218125246/

Orange/widgets/data/tests/test_owfile.py
```diff
@@ -127,3 +127,15 @@ def test_file_not_found(self):
         # Open a sample dataset
         self.open_dataset("iris")
         self.assertFalse(self.widget.Error.file_not_found.is_shown())
+
+    def test_check_column_noname(self):
+        '''
```
lanzagar: Docstrings: `'''` -> `"""`

Orange/widgets/utils/domaineditor.py
```diff
@@ -80,7 +80,8 @@ def setData(self, index, value, role):
         row_data = self.variables[row]
         if role == Qt.EditRole:
             if col == Column.name:
-                row_data[col] = value
+                if not (value.isspace() or value == ""):
```
lanzagar: This results in the above `if col == Column.name` succeeding and then the inside `if` doing nothing for empty values. But the function still returns True in the end.
Instead, using `if col == Column.name and not (value.isspace() or value == "")` would result either in the value being updated or nothing happens and function returns False. Which means that after an "incorrect" edit the Apply button does not become enabled (same as cancelling the edit with Esc).

Orange/widgets/utils/domaineditor.py
```diff
@@ -211,9 +212,9 @@ def is_missing(x):
 
         for (name, tpe, place, _, _), (orig_var, orig_plc) in \
                 zip(variables,
-                    chain([(at, Place.feature) for at in domain.attributes],
-                          [(cl, Place.class_var) for cl in domain.class_vars],
-                          [(mt, Place.meta) for mt in domain.metas])):
+                        chain([(at, Place.feature) for at in domain.attributes],
```
lanzagar: Why the additional indent here?

jerneju: PYLINT demands some additional spaces.

PR #2041, opened 2017-02-21
DomainContextHandler: Encode anonymous variables

astaric:
##### Issue
Variables are encoded as `name, type` pairs so there is no reason that anonymous variables could not be stored as settings.

The error was caused by calling copy.copy on all values, even those that would not be used in encoded representation.

##### Description of changes
copy only when necessary

##### Includes
- [X] Code changes
- [X] Tests


astaric:
@lanzagar can you take a look at this change?

PR #2043, opened 2017-02-21
OWMergeData: Suggest matching attributes and show output information

VesnaT:
##### Issue

Suggest matching attributes and show output information

##### Description of changes
Add output info
Use heuristic to suggest matching attributes

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation


Orange/widgets/data/owmergedata.py
```diff
@@ -83,32 +86,48 @@ def _setAttrs(self, model, data, othermodel, otherdata):
...
                     model_.insert(0, INSTANCEID)
 
+    def _find_best_match(self):
+        if self.dataA and self.dataB:
+            attr_a, attr_b, n_max_intersect = INDEX, INDEX, 0
+            str_metas_a = [m for m in self.dataA.domain.metas if m.is_string]
+            str_metas_b = [m for m in self.dataB.domain.metas if m.is_string]
+            for m_a, m_b in itertools.product(str_metas_a, str_metas_b):
+                n_inter = len(numpy.intersect1d(self.dataA[:, m_a].metas,
+                                                self.dataB[:, m_b].metas))
```
janezd: What about another heuristic: the matching attributes will tend to have the same name. This may be a stupid idea, but perhaps you can add `+ (m_a.name == m_b.name)` just to break ties.

PR #2046, opened 2017-02-22
[FIX] ZeroDivisionError owmosaic.py

jerneju:
##### Issue
ZeroDivisionError is thrown when invalid data is loaded into widget. For instance: [[0, NaN], [NaN, NaN]].
https://sentry.io/biolab/orange3/issues/204400244/

##### Description of changes
If the value "total" is zero aborts drawing the plot.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -512,6 +512,20 @@ class OWMosaicDisplay(OWWidget):
         spacing = self.SPACING
         bar_width = self.BAR_WIDTH
 
+        def get_counts(attr_vals, values):
+            """This function calculates rectangles' widths.
+            If all widths are zero then all widths are set to 1."""
+            if attr_vals == "":
+                counts = [conditionaldict[val] for val in values]
+            else:
+                counts = [conditionaldict[attr_vals + "-" + val]
+                          for val in values]
+            total = sum(counts)
+            if total == 0:
+                counts = [1] * len(values)
+                total = sum(counts)
+            return total, counts
+
         def draw_data(attr_list, x0_x1, y0_y1, side, condition,
                       total_attrs, used_attrs, used_vals, attr_vals=""):
             x0, x1 = x0_x1
@@ -542,12 +556,7 @@ class OWMosaicDisplay(OWWidget):
                 if whole == 0:
                     edge = (y1 - y0) / float(len(values) - 1)
 
-            if attr_vals == "":
-                counts = [conditionaldict[val] for val in values]
-            else:
-                counts = [conditionaldict[attr_vals + "-" + val]
-                          for val in values]
-            total = sum(counts)
+            total, counts = get_counts(attr_vals, values)
 
             # if we are visualizing the third attribute and the first attribute
             # has the last value, we have to reverse the order in which the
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/204400244/

Orange/widgets/visualize/tests/test_owmosaic.py
```diff
@@ -194,3 +194,16 @@ def test_does_not_crash(self):
         data = Table("housing.tab")
         widget.set_data(data)
         vizrank.toggle()
+
+    def test_zerodivision_error(self):
+        """
+        ZeroDivisionError should not be thrown when
```
astaric: Could you make this description more explicit. Instead of "some kind of invalid data" try to describe what the problem in data was.

Orange/widgets/visualize/tests/test_owmosaic.py
```diff
@@ -194,3 +194,16 @@ def test_does_not_crash(self):
...
+        some kind of invalid data is loaded into widget.
+        GH-2046
+        """
+        a, b = ContinuousVariable("a"), ContinuousVariable("b")
+        table = Table(Domain([a, b]), np.array([[0, np.NaN], [np.NaN, np.NaN]]))
+        try:
+            self.send_signal("Data", table)
+        except ZeroDivisionError:
+            self.assertTrue(False)
```
astaric: self.fail can be used to make the test fail. And it is also a good idea to specify the reason of the failure ("Widget crashed with ZeroDivisionError" or something like this)

Orange/widgets/visualize/owmosaic.py
```diff
@@ -512,6 +512,18 @@ def update_graph(self):
         spacing = self.SPACING
         bar_width = self.BAR_WIDTH
 
+        def get_counts(attr_vals, values):
+            if attr_vals == "":
```
astaric: Nice. Could you also add a docstring to this function describing what it does and how it handles zero counts?

Orange/widgets/visualize/owmosaic.py
```diff
@@ -512,6 +512,18 @@ def update_graph(self):
...
+                counts = [conditionaldict[val] for val in values]
```
astaric: While draw_data used conditionaldict[val], draw_text used conditionaldict.get(val, 1). Could you elaborate on why indexing is sufficient?

Orange/widgets/visualize/tests/test_owmosaic.py
```diff
@@ -194,3 +194,23 @@ def test_does_not_crash(self):
...
+            np.array([
+                [0, np.NaN, 0],
+                [0, np.NaN, 0],
+                [0, np.NaN, 0]
+            ])
+        )
+        try:
+            self.send_signal("Data", table)
+        except ZeroDivisionError:
+            self.fail("Widget crashed due to ZeroDivisionError.")
```
lanzagar: Using data with only NaNs should not result in any kind of error, not just ZeroDivision.
Why not make this test more general: rename from test_zerodivision_error to e.g. test_nan_column, remove the try, all exceptions will be shown when/if they happen.

Issue #1798, opened 2016-12-02
DataLoad-valueError: could not convert string to float.


anilkodepaka:
##### Orange version
3.3.8

##### Expected behavior

data file should load with the selected conversions (string to date and vice versa)

##### Actual behavior

Exception:
ValueError: could not convert string to float: '27.05.2015'
Module:
numpy.core.numeric:482
Widget Name:
File
Widget Module:
Orange.widgets.data.owfile:410
.
.
Stack Trace:
Traceback (most recent call last):
  File "C:\Python34\lib\site-packages\Orange\widgets\data\owfile.py", line 410, in apply_domain_edit
    table = Table.from_numpy(domain, X, y, m, self.data.W)
  File "C:\Python34\lib\site-packages\Orange\data\table.py", line 440, in from_numpy
    X, Y, W = _check_arrays(X, Y, W, dtype='float64')
  File "C:\Python34\lib\site-packages\Orange\data\table.py", line 1461, in _check_arrays
    array = np.asarray(array, dtype=dtype)
  File "C:\Python34\lib\site-packages\numpy\core\numeric.py", line 482, in asarray
    return array(a, dtype, copy=False, order=order)
ValueError: could not convert string to float: '27.05.2015'

##### Steps to reproduce the behavior

Data is present as a string in excel file, at the time of loading file, tried to update date column from string/meta to Datetime/feature.

ajdapretnar:
Datetime accepts data in ISO format only (e.g. 1990-01-31). However, the widget should not crash when encountering an unfamiliar format.

ajdapretnar:
@anilkodepaka Can you check whether the problem persists with the latest version?

ajdapretnar:
Yeah, it does. Happened to me as well...

ajdapretnar:
https://sentry.io/biolab/orange3/issues/196743724/

anilkodepaka:
hei..what is this link about?

ajdapretnar:
The corresponding issue on Sentry. Just to keep track of the resolved / unresolved issues.

ajdapretnar:
Duplicate of #1520.

Issue #1520, opened 2016-08-24
File widget: check if transformations of data type are possible


ajdapretnar:
When changing data type from numeric to datetime and connecting it to Data Table, Data Table crashes. I reckon conversion is not possible (not ISO). Such transformations should probably be only offered if conversion is possible, e.g. if numeric attribute can indeed be transformed to datetime. I think the problem is only with datetime anyway.

p.s. Can strings never be numeric? There's no such option available. Neither there is one for converting nominal to numeric.

kernc:
Strings and categoricals should be convertible too.

ajdapretnar:
I am not sure this is solved. Now I can't change string to datetime, but this is not exactly what this issue is about.
@kernc What do you think? Keep or close?

ajdapretnar:
From testing several data sets I conclude the issue is now solved properly for the time being. Nothing crashes and conversion for now make sense.

PR #2050, opened 2017-02-24
[FIX] File: Disallow changing string columns to datetime

jerneju:
##### Issue
ValueError  - could not convert string to float: '02.02.17'
https://sentry.io/biolab/orange3/issues/196743724/

##### Description of changes
Datetime option only if numerical.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owfile.py
+++ b/Orange/widgets/data/tests/test_owfile.py
@@ -130,7 +130,8 @@ class TestOWFile(WidgetTest):
 
     def test_check_column_noname(self):
         """
-        GH-2018
+        Column name cannot be changed to an empty string or a string with whitespaces.
+        GH-2039
         """
         self.open_dataset("iris")
         idx = self.widget.domain_editor.model().createIndex(1, 0)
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/198615964/

jerneju:
#1520, #1798

ajdapretnar:
This fixes the error nicely, but I was wondering whether we could provide the user with a better error message. Perhaps "Use ISO format for datetime variables (e.g. 2017-02-27)"? Say in a form of a warning? Just to tell the user how to proceed with datetime variables? :)

jerneju:
ISO 8601? Yes, program knows how to read "2017-03-01" but have no clue what means "2017-03-01T12:49:39+00:00" ("could not convert string to float").

ajdapretnar:
Complain to @kernc. It does work with _2017-03-01 12:49:39_ though.
Can we come up with a good user message?

jerneju:
Great, I see it does work with that kind of format. I changed the code in a way that a user message won't be necessary. Now only variables which can be set to numerical can also be set to datetime.

astaric:
Could you include a test?

kernc:
From the top of the head, the reason it didn't work with:

> 2017-03-01T12:49:39+00:00

is that the colon in the timezone part doesn't parse with [`datetime.strptime`](http://strftime.org/). Standards-wise, I guess we deviate a little.

PR #2051, opened 2017-02-24
[WIP] [ENH] MergeData: New GUI

janezd:
##### Issue

The Merge Data widget is not as intuitive as we'd like.

##### Description of changes

Changes in GUI.

This PR so far changes only the GUI, but does not implement (or adapt the existing) functions that actually merge the data.

Something in the spirit of @VesnaT's modifications from #2043 also need to be ported here, possibly with different heuristics for each combination.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owmergedata.py
+++ b/Orange/widgets/data/owmergedata.py
@@ -1,19 +1,16 @@
 import math
 import itertools
 from collections import defaultdict
+from itertools import chain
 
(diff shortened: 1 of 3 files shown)


janezd:
@VesnaT, can you take it from here -- as an expert for inner, outer and other joins?

PR #2058, opened 2017-02-28
OWBaseLearner.set_preprocessor: accept a single preprocessor

kernc:
Before e3dc8f212397f6b95eace371e61f809fbd39fbfb, passing a single
preprocessor was allowed. Any reason not to?

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


kernc:
Nevermind. Easier to fix on the client end. Widgets accept a single Preprocessor signal anyway and even when PreprocessorList, this is not really a Sequence.

PR #2059, opened 2017-02-28
Highcharts+WebView minor updates

kernc:
##### Description of changes
Of note:
* raise selected Highcharts points to front, and
* always suffix the evaluated JS code with `";0"` to avoid failing to interpret arbitrary complex objects the user's JS expression can evaluate to, and
* export SVG as SVG instead of HTML, and
* have `WidgetTest.wait_for()` that comes in handy with async GUI testing.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


kernc:
Shit is greeeeen!!!

PR #2057, opened 2017-02-28
[FIX] Heatmap: Fix crash on data with empty columns

jerneju:
##### Issue
Invalid number of variable columns
Columns in a domain and in a table do not match.
https://sentry.io/biolab/orange3/issues/196744285/

##### Description of changes
If a column in a table is dropped, it is erased in a domain as well.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

PR #2061, opened 2017-03-01
[FIX] ScatterPlot: Fix crash when coloring by column of unknowns

jerneju:
##### Issue
ValueError cannot convert float NaN to integer owscatterplot
In case when all column values are NaN then it throws that error.
https://sentry.io/biolab/orange3/issues/196744604/

##### Description of changes
Logarithm of zero and division by zero are prevented.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-from AnyQt.QtCore import QSize, Qt, QTimer
+from AnyQt.QtCore import Qt, QTimer
 from AnyQt.QtGui import (
     QPen, QFont, QFontInfo, QPalette, QKeySequence,
 )
@@ -162,7 +162,7 @@ class OWScatterPlot(OWWidget):
         box = gui.vBox(self.controlArea, "Axis Data")
         dmod = DomainModel
         self.xy_model = DomainModel(dmod.MIXED, valid_types=dmod.PRIMITIVE)
-        gui.comboBox(
+        self.cb_attr_x = gui.comboBox(
             box, self, "attr_x", label="Axis x:", callback=self.update_attr,
             model=self.xy_model, **common_options)
         self.cb_attr_y = gui.comboBox(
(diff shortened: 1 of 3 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/196744604/

Orange/widgets/visualize/tests/test_owscatterplot.py
```diff
@@ -102,3 +103,26 @@ def test_report_on_empty(self):
...
         self.widget.report_caption.assert_not_called()
         self.widget.report_items.assert_not_called()
+
+    def test_data_column_nans(self):
+        """
+        ValueError cannot convert float NaN to integer.
+        In case when all column values are NaN then it throws that error.
+        GH-2061
+        """
+        dat = """\
```
astaric: Is there a special reason why the dataset is generated in this way? (and not with table like in https://github.com/biolab/orange3/pull/2057/files#diff-1a66c8eae521f07f1c5d904e0cd025a7R117)?

Orange/widgets/visualize/tests/test_owscatterplot.py
```diff
@@ -102,3 +103,26 @@ def test_report_on_empty(self):
...
+            42.11\t\tneumann
+            17.01\t\tneumann
+            19.01\t\tneumann
+            33.21\t\tturing
+            34.01\t\tturing"""
+        file = io.StringIO(dat)
+        data = TabReader(file).read()
+        self.send_signal("Data", data)
+
+        self.widget.graph.attr_color = self.widget.get_shown_attributes()[1]
```
astaric: Could you select attributes using helpers found in https://github.com/biolab/orange3/blob/master/Orange/widgets/tests/utils.py#L268?

Orange/widgets/visualize/tests/test_owscatterplot.py
jerneju: No particular reason at all. Now changed.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -409,7 +409,7 @@ def handleNewSignals(self):
         self.graph.new_data(self.data_metas_X, self.subset_data)
         if self.attribute_selection_list and \
                 all(attr in self.graph.domain
-                    for attr in self.attribute_selection_list):
+                        for attr in self.attribute_selection_list):
```
kernc: This was indented ok before.

jerneju: Pylint complained and that is why I added 4 spaces.

PR #2064, opened 2017-03-02
[FIX] owpreprocess: Handle columns with only NaN values

jerneju:
##### Issue
ZeroDivisonError - Weights sum to zero, can't be normalized
In case when all rows in a column are NaN then it throws that error.
https://sentry.io/biolab/orange3/issues/227119466/

##### Description of changes
Add another except in try block and now it shows an error message in the widget.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owpreprocess.py
+++ b/Orange/widgets/data/tests/test_owpreprocess.py
@@ -9,7 +9,7 @@ from Orange.preprocess import (
 from Orange.preprocess import discretize, impute, fss, score
 from Orange.widgets.data import owpreprocess
 from Orange.widgets.data.owpreprocess import OWPreprocess
-from Orange.widgets.tests.base import WidgetTest
+from Orange.widgets.tests.base import WidgetTest, datasets
 
 
 class TestOWPreprocess(WidgetTest):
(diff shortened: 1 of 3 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/227119466/

jerneju:
https://sentry.io/biolab/orange3/issues/196743808/

Orange/widgets/data/tests/test_owpreprocess.py
```diff
@@ -44,6 +44,30 @@ def test_normalize(self):
         np.testing.assert_allclose(output.X.mean(0), 0, atol=1e-7)
         np.testing.assert_allclose(output.X.std(0), 1, atol=1e-7)
 
+    def test_data_column_nans(self):
+        """
+        ZeroDivisonError - Weights sum to zero, can't be normalized
+        In case when all rows in a column are NaN then it throws that error.
+        GH-2064
+        """
+        table = Table(
```
astaric: If the same dataset is used in multiple tests, would you consider moving it to [datasets](https://github.com/biolab/orange3/blob/master/Orange/widgets/tests/base.py#L662)?

PR #2067, opened 2017-03-02
[FIX] OWTreeViewer: Fix trees being displayed differently for same tree object

pavlin-policar:
##### Issue
OWTreeViewer was not deterministic in showing trees. This was not specific to trees where attributes had the same entropy or whatever the selection criteria was, but occured on any tree (even with iris).

##### Description of changes
This happened because the layout was updated w.r.t the graph node edges, which were stored in a set object, which is unordered.

I assume that a set was used because speed is important, so I've implemented edges as an ordered dict, to keep the speed benefits as well as getting an ordering.

I've written a test that draws the tree 10x and compares the node positions, but since the ordering of the nodes was completely random before, this is not guaranteed to fail in case the trees are drawn with random ordering. The ordering could *happen* to be the same 10x. This hasn't actually happened to me yet on iris, since there are more nodes that can fail there, but this could potentially happen.

I have no idea how to test this so that it would always fail, due to the random nature of the ordering.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/tests/test_owpythagorastree.py
+++ b/Orange/widgets/visualize/tests/test_owpythagorastree.py
@@ -314,7 +314,8 @@ class TestOWPythagorasTree(WidgetTest, WidgetOutputsTestMixin):
         # Make sure trees are deterministic with data where some variables have
         # the same entropy
         data_same_entropy = Table(path.join(
-            path.dirname(__file__), "../../tests/datasets/same_entropy.tab"))
+            path.dirname(path.dirname(path.dirname(__file__))), "tests",
+            "datasets", "same_entropy.tab"))
         data_same_entropy = TreeLearner()(data_same_entropy)
         scene_nodes = []
         for _ in range(n_tries):
--- a/Orange/widgets/visualize/tests/test_owpythagorastree.py
+++ b/Orange/widgets/visualize/tests/test_owpythagorastree.py
@@ -2,9 +2,10 @@
 import math
 import unittest
 
-from Orange.classification import TreeLearner as TreeClassificationLearner
+from os import path
+
 from Orange.data import Table
-from Orange.regression import TreeLearner as TreeRegressionLearner
+from Orange.modelling import TreeLearner
 from Orange.widgets.tests.base import WidgetTest, WidgetOutputsTestMixin
 from Orange.widgets.visualize.owpythagorastree import OWPythagorasTree
 from Orange.widgets.visualize.pythagorastreeviewer import (
@@ -93,7 +94,7 @@ class TestOWPythagorasTree(WidgetTest, WidgetOutputsTestMixin):
         WidgetOutputsTestMixin.init(cls)
 
         # Set up for output tests
-        tree = TreeClassificationLearner()
+        tree = TreeLearner()
         cls.model = tree(cls.data)
         cls.model.instances = cls.data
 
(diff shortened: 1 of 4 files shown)


pavlin-policar:
@astaric The tests are failing because apparently the `data.tab` file I added to `Orange/widgets/tests/` doesn't get uploaded and fails because the file is not found (even though it is tracked by git. How to I get it to actually upload?

astaric:
You need to include it in [Manifest](https://github.com/biolab/orange3/blob/master/MANIFEST.in#L11) and [setup.py](https://github.com/biolab/orange3/blob/master/setup.py#L179). Maybe you could move the dataset to where the other datasets reside, where this is already done? (https://github.com/biolab/orange3/tree/master/Orange/widgets/tests/datasets)

pavlin-policar:
Thanks, I've moved it to the `widgets/tests/datasets` folder.

Orange/widgets/visualize/tests/test_owpythagorastree.py
```diff
@@ -283,3 +284,46 @@ def test_label_on_tree_connect_and_disconnect(self):
...
+            self.assertTrue(
+                _check_all_same(node_row),
+                "The tree was not drawn identically in the %d times it was "
+                "sent to widget after receiving the iris dataset." % n_tries
+            )
+
+        # Make sure trees are deterministic with data where some variables have
+        # the same entropy
+        data_same_entropy = Table(path.join(
+            path.dirname(__file__), "../../tests/datasets/same_entropy.tab"))
```
kernc: Use `os.path.join()` to make the test pass on Windos.

Orange/widgets/visualize/tests/test_owtreegraph.py
```diff
@@ -19,6 +21,12 @@ def setUpClass(cls):
         cls.signal_name = "Tree"
         cls.signal_data = cls.model
 
+        # Load a dataset that contains two variables with the same entropy
+        data_same_entropy = Table(path.join(
+            path.dirname(__file__), "../../tests/datasets/same_entropy.tab"))
```
kernc: And here.

Orange/widgets/visualize/tests/test_owtreegraph.py
```diff
@@ -35,3 +43,43 @@ def test_target_class_changed(self):
...
+        n_tries = 10
+
+        def _check_all_same(data):
+            """Check that all the elements within an iterable are identical."""
+            iterator = iter(data)
+            try:
+                first = next(iterator)
+            except StopIteration:
+                return True
+            return all(first == rest for rest in iterator)
```
kernc: Would `return len(set(iter(data))) == 1` work?

pavlin-policar: Unfortunately not, because I'm comparing `QPointF`s which are an unhashable type.

Orange/widgets/visualize/tests/test_owpythagorastree.py
pavlin-policar: Is there a cleaner solution than `path.join(path.dirname(path.dirname(path.dirname(__file__))), "tests", "datasets", "same_entropy.tab")`? Could I maybe put in `".."` instead of  `path.dirname`?

PR #2070, opened 2017-03-02
[ENH] Scatterplot: Implement grouping of selections

janezd:
Instead of annotating selected data with "No" and "Yes", use G0, G1, G2 ... when data is selected is subsequent groups using Shift or Cmd/Ctrl modifier.

Connect File -> Scatter plot, and then Scatter plot to a Table and also to another Scatter plot, both through signal Data (not Selection). Now select several groups in the first scatter plot by using the Shift modifier and see the results in the other scatter plot and table.

The change does not affect the other Scatter plot's output.

We can add a check box "Group selections" to enable or disable this behaviour. However, I think this should be the default behaviour and I see little reason why the user would want to disable it.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation


janezd:
@astaric, I defined a GUI for this. :)

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -463,19 +464,43 @@ def update_graph(self, reset_view=True, **_):
...
+        metas = data.domain.metas + (
+            DiscreteVariable(
+                name,
+                ["G{}".format(i) for i in range(np.max(selection) + 1)]),
+        )
+        domain = Domain(data.domain.attributes, data.domain.class_vars, metas)
+        table = Table(
+            domain, data.X, data.Y,
+            metas=np.hstack((data.metas, selection.reshape(len(data), 1))))
+        table.attributes = data.attributes
```
markotoplak: table.ids = data.ids

janezd:
radon fails because of an additional case in the switch-ish `select` method that acts upon different possible keyboard modifiers. There's neither a reasonable way to fix this, nor is there a reason to do so as the method's flow is rather clear.

I'd rather not improve a cc score of a totally unrelated method just to make radon happy.

PR #2072, opened 2017-03-03
Fix pyqtgraphs Qt5 detection after importing Orange

markotoplak:
Orange install backport hooks for Qt4 in __init__ which
mess up pyqtgraph Qt dection. Importing pyqtgraph before
installing importhooks fixes this.

### DIFF
--- a/Orange/__init__.py
+++ b/Orange/__init__.py
@@ -33,5 +33,7 @@ except ImportError:
     pass
 else:
     if AnyQt.USED_API == "pyqt5":
+        import pyqtgraph  # import pyqtgraph first so that it can detect Qt5
+        del pyqtgraph
         AnyQt.importhooks.install_backport_hook('pyqt4')
     del AnyQt


jerneju:
#### Issue
Because of these two added lines Orange-canvas does not start. It throws an error:
_RuntimeError: the PyQt5.QtCore and PyQt4.QtCore modules both wrap the QObject class_

#### Solution
Remove package _python3-pyqt4_ from the system.

markotoplak:
Hm, a problem with two Qts. Which Qt was in that case used in pyqtgraph and which in AnyQt?

jerneju:
pyqtgraph 0.10.0
AnyQt 0.0.8

kernc:
> RuntimeError: the PyQt5.QtCore and PyQt4.QtCore modules both wrap the QObject class

Fixed in https://github.com/biolab/orange3/pull/2080.

PR #2076, opened 2017-03-07
Allow choosing sheets when constructing a Table

markotoplak:
##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation

### DIFF
--- a/Orange/tests/test_xlsx_reader.py
+++ b/Orange/tests/test_xlsx_reader.py
@@ -49,7 +49,8 @@ class TextExcelSheets(unittest.TestCase):
         self.assertEqual(table.name, 'header_0_sheet-my_sheet')
 
     def test_named_sheet_table(self):
-        table = Table(get_dataset("header_0_sheet.xlsx"), sheet="my_sheet")
+        table = Table.from_file(get_dataset("header_0_sheet.xlsx"),
+                                sheet="my_sheet")
         self.assertEqual(len(table.domain.attributes), 4)
         self.assertEqual(table.name, 'header_0_sheet-my_sheet')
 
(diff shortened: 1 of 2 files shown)


astaric:
This looks like a change that could benefit from the third checkbox (documentation).

Orange/tests/test_xlsx_reader.py
```diff
@@ -48,6 +48,11 @@ def test_named_sheet(self):
         self.assertEqual(len(table.domain.attributes), 4)
         self.assertEqual(table.name, 'header_0_sheet-my_sheet')
 
+    def test_named_sheet_table(self):
+        table = Table(get_dataset("header_0_sheet.xlsx"), sheet="my_sheet")
```
kernc: Can we avoid polluting the constructor? I.e. encouraging the use of `Table.from_file()`?

markotoplak: Constructor did not change at all with this PR, here it is just used as usual.

kernc: Usage implies `sheet` is an argument accepted in the constructor. Any future re-design needs to take this into consideration. Our Table constructor doesn't have a docstring. This is _not the usual_.

markotoplak: True, that is exactly why I thought to include it here like this: that we would know what we break. :) But here I think it would actually be better to modify the constructor not to pass kwargs to from_file so that this would not even be allowed. What do you think?

kernc: Indeed, I don't mind not passing kwargs to from_file method.

PR #2077, opened 2017-03-07
File formats: wildcard matching

markotoplak:
##### Issue

For one file format I would like to match extension ".[0-9]*" and this allows it.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


astaric:
Some tests would be nice.

Issue #2030, opened 2017-02-20
Workflows open with the last opened file


ajdapretnar:
##### Orange version
3.4.dev0+45307ad

##### Expected behavior
Workflows open with the file the workflow was created with.

##### Actual behavior
Workflows open with the last opened file instead.

##### Steps to reproduce the behavior
Create a schema. File (_iris.tab_) --> Data Table. Add a single File widget with _housing.tab_. Save.
Reopen the schema. Now both File widgets open with housing.tab.

ajdapretnar:
Or simply open a random (non-iris) data set with File and open one the example schemas. They open with your non-iris file even when they should be iris.

PR #2078, opened 2017-03-07
[FIX] File - Workflows open with the last opened file

jerneju:
##### Issue
Workflows open with the last opened file instead with the file
the workflow was created with
#2030 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Can you describe how you reproduced this?

Also, I believe that the PR doesn't fix what's broken, it fixes the consequence. The first item in `last_path` should already contain the most recent path, so `last_path()` should work. If it doesn't, the problem that needs fixing is either in `RecentPathsWidgetMixin` or in the way the File widget (and potentially some other widgets) uses it.

astaric:
Fixed in a different way in #2095

PR #2084, opened 2017-03-09
OWMergeData: New GUI, heuristic to suggest matching attributes

VesnaT:
##### Issue

The widget should support three different merging types (left, inner and outer joins).
The widget can't remember attribute settings (unless saved workflows), so heuristic should be used to find and suggest the best match.

Replaces #2043 and #2051.
##### Description of changes
New GUI.
The heuristic searches the biggest intersection amongst pairs of available (unique in most cases) meta attributes.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation


janezd:
@VesnaT, I refactored (rather: reimplemented) the part related to merging. The widget passes the tests (at least locally), but could you, as an expert, please also manually click through it a bit?

VesnaT:
In case of an empty table (len(data) == 0) we usually output None. Here should probably be the same.

janezd:
Essentially fixed with two `del extra_map[np.nan]`, after all day's work. :) Plus, it should now be more reliable for numeric variables.

Issue #2091, opened 2017-03-09
Fitter preprocessors


lanzagar:
##### Orange version

3.4.0

##### Expected behavior
Learners use preprocessors.

##### Actual behavior
Learners extending the Fitter base class do not use preprocessors.

##### Steps to reproduce the behavior
Use a learner on e.g. hearth_disease data set

lanzagar:
This is a very high priority bug in the latest version that needs a fix.
@pavlin-policar do you have time to check this out as you are most familiar with the code?

pavlin-policar:
Sure thing, I'll get right on it.

pavlin-policar:
Should be fixed with #2093

PR #2093, opened 2017-03-09
[FIX] Fitter: Properly delegate preprocessors

pavlin-policar:
##### Issue
#2091

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


pavlin-policar:
This also fixes most of the failing tests in #2092, except the ones that are related to the association rules widget, which isn't a fitter, so it isn't related to this.

PR #2098, opened 2017-03-13
[FIX] Sieve Diagram: Using datasets with meta data

jerneju:
##### Issue

Widget should intepret meta data which are continuous or discrete in the same way
as features or target.
https://sentry.io/biolab/orange3/issues/231234770/
https://sentry.io/biolab/orange3/issues/232101010/

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owsieve.py
+++ b/Orange/widgets/visualize/owsieve.py
@@ -420,9 +420,12 @@ class OWSieveDiagram(OWWidget):
         self.canvasView.setSceneRect(0, 0, view.width(), view.height())
         if not disc_x.values or not disc_y.values:
             text_ = "Features {} and {} have no values".format(disc_x, disc_y) \
-                if not disc_x.values and not disc_y.values and \
-                   disc_x != disc_y else "Feature {} has no values".format(
-                disc_x if not disc_x.values else disc_y)
+                if not disc_x.values and \
+                   not disc_y.values and \
+                          disc_x != disc_y \
+                else \
+                    "Feature {} has no values".format(
+                        disc_x if not disc_x.values else disc_y)
             text(text_, view.width() / 2 + 70, view.height() / 2,
                  Qt.AlignRight | Qt.AlignVCenter)
             return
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/231234770/

jerneju:
https://sentry.io/biolab/orange3/issues/232101010/

Orange/widgets/visualize/owsieve.py
```diff
@@ -167,13 +167,10 @@ def set_data(self, data):
             self.domain_model.set_domain(None)
         else:
             self.domain_model.set_domain(data.domain)
-            if any(attr.is_continuous for attr in data.domain):
```
janezd: The idea here was to avoid unnecessary calls to `Discretize`. Wouldn't it be better to just replace the above condition with something like

    if any(attr.is_continuous for attr in chain(data.domain, data.domain.metas)):

and leave the rest as it is?

Orange/widgets/visualize/tests/test_owsieve.py
```diff
@@ -59,3 +59,22 @@ def test_chisquare(self):
...
+                [],
+                [],
+                [ContinuousVariable("a"),
+                 DiscreteVariable("b", values=["y", "n"])]
+            ),
+            list(zip(
+                [42.48, 16.84, 15.23, 23.8],
+                "yynn"))
+        )
+        self.send_signal("Data", table)
```
janezd: This checks just that the widget does not crash. It is difficult to test this widget assuming anything about the widget internals. For the lack of better options, I'd replace `self.send_signal("Data", table)` with

```
        from unittest.mock import patch
        from Orange.widgets.visualize.owsieve import Discretize
        with patch("Orange.widgets.visualize.owsieve.Discretize",
                   wraps=Discretize) as disc:
            self.send_signal("Data", table)
            self.assertTrue(disc.called)
        metas = self.widget.discrete_data.domain.metas
        self.assertEqual(len(metas), 2)
        self.assertTrue(all(attr.is_discrete for attr in metas))
```

Move imports to the beginning of the file, of course.

Issue #2086, opened 2017-03-09
Select columns and multiple target variables problems


markotoplak:
##### Orange version
master

##### Actual behavior

If there are multiple Target variables, Select columns have a box for Target variables that is too small: in some cases I only see one variable and need to scroll for others.

If I move one target variable somewhere else, I can not move it back if there is already one target variable. So I can not even get into the original state unless I click reset.

jerneju:
Yes, the only option to have more than one target variable in a box "Target Variable" is to press reset button. Then all target variables are listed there.

Maybe we should upgrade the box "Target Variable" to "Target Variable**s**".

jerneju:
Assign @jerneju

PR #2100, opened 2017-03-14
Select Columns: multiple target variables problems

jerneju:
#### Issue
A box for Target variables is too small.
When one target variable is moved somewhere else, it cannot be moved  back if there is already one target variable.
#2086

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/test_owselectcolumns.py
+++ b/Orange/widgets/tests/test_owselectcolumns.py
@@ -1,11 +1,13 @@
 from unittest import TestCase
 from unittest.mock import Mock
-from Orange.data import ContinuousVariable, DiscreteVariable, Domain
+from Orange.data import Table, ContinuousVariable, DiscreteVariable, Domain
 from Orange.widgets.data.contexthandlers import \
     SelectAttributesDomainContextHandler
 from Orange.widgets.settings import ContextSetting
 from Orange.widgets.utils import vartype
-
+from Orange.widgets.tests.base import WidgetTest
+from Orange.widgets.data.owselectcolumns \
+    import OWSelectAttributes
 
 Continuous = vartype(ContinuousVariable())
 Discrete = vartype(DiscreteVariable())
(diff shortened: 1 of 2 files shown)


astaric:
Could you add some tests?

jerneju:
Test added.

Orange/widgets/tests/test_owselectcolumns.py
```diff
@@ -105,3 +107,33 @@ def retrieveSpecificSettings(self):
...
+    def setUp(self):
+        self.widget = self.create_widget(OWSelectAttributes)
+
+    def testall(self, location):
+        self.assertEqual(len(self.widget.available_attrs), location[0])
+        self.assertEqual(len(self.widget.used_attrs), location[1])
+        self.assertEqual(len(self.widget.class_attrs), location[2])
+
+    def test_multiple_target_variable(self):
+        '''
```
astaric: We usually use double quotes for docstrings.

Orange/widgets/tests/test_owselectcolumns.py
```diff
@@ -105,3 +107,33 @@ def retrieveSpecificSettings(self):
...
+        A box for Target variables is too small.
```
astaric: In docstring try to explain what is being tested, not what the problem is.

Orange/widgets/tests/test_owselectcolumns.py
```diff
@@ -105,3 +107,33 @@ def retrieveSpecificSettings(self):
 
     def storeSpecificSettings(self):
         pass
+
+class TestOWSelectAttributes(WidgetTest):
+    def setUp(self):
+        self.widget = self.create_widget(OWSelectAttributes)
+
+    def testall(self, location):
```
astaric: Methods that group several assertions are usually called assertSomething. In this case, I would go with
assertVariableCountsEqual(available, used, class).

This method would be called below, possibly even with keyword arguments to improve clarity.

Orange/widgets/tests/test_owselectcolumns.py
```diff
@@ -105,3 +107,33 @@ def retrieveSpecificSettings(self):
...
+        '''
+        A box for Target variables is too small.
+        When one target variable is moved somewhere else, it cannot be moved back
+        if there is already one target variable.
+        GH-2100
+        GH-2086
+        '''
+        iris = Table("iris")
+        self.send_signal("Data", iris)
+        self.widget.move_class_button.click()
```
astaric: It might be a good idea to add an assert before anything is clicked. The main point would be to show what the variable counts are when the widget is opened.

markotoplak:
Dragging (with a mouse) of multiple variables into class vars still does not work. It is also impossible to drag anything there if there is already a variable in class vars.

astaric:
Could you also check the behaviour of the left-right arrow for the target group? for attributes and metas, it seem to disable if nothing is selected, while for targets, it does not.

Issue #1753, opened 2016-11-15
No widget in Orange3


kingaza:
I install Orange3 3.3.8 on win7.
After open canvas, the main UI opens.
However, no widget lists are available.

kernc:
Does [running with `-l 4`](https://github.com/biolab/orange3/issues/1286#issuecomment-222331975) produce any helpful output?

kingaza:
here is my log

(D:\05_Software\Anaconda) C:\Users\hejiew>D:\05_Software\Anaconda\Scripts\orange
-canvas.exe -l 4
Starting 'Orange Canvas' application.
Starting CanvasApplicaiton with argv = ['D:\\05_Software\\Anaconda\\Scripts\\ora
nge-canvas-script.py'].
Adding search path 'orange' for prefix, 'canvas_icons'
Setting up Canvas main window.
'CanvasScene(objectName='', ...)' intitialized.
Setting registry 'None on 'CanvasScene(objectName='', ...)'.
Scheme node properties sync (changed: False)
'CanvasScene(objectName='', ...)' cleared.
'CanvasScene(objectName='', ...)' intitialized.
Setting registry 'None on 'CanvasScene(objectName='', ...)'.
Setting scheme '<Orange.canvas.scheme.widgetsscheme.WidgetsScheme object at 0x00
0000000539DEE8>' on 'CanvasScene(objectName='', ...)'
Stack transition start ()
Stack transition finished ()
Dock transition finished (new width 116)
Dock transition started.
Loading widget registry cache ('C:\\Users\\hejiew\\AppData\\Roaming\\Orange\\3.3
.8\\Cache\\canvas\\registry-cache.pck').
Running widget discovery process.
Could not load 'Orange Widgets = Orange.widgets' (unsatisfied dependencies).
Traceback (most recent call last):
  File "D:\05_Software\Anaconda\lib\site-packages\Orange\canvas\registry\discove
ry.py", line 89, in run
    point = entry_point.load()
  File "D:\05_Software\Anaconda\lib\site-packages\setuptools-23.0.0-py3.5.egg\pk
g_resources\__init__.py", line 2228, in load
    self.require(*args, **kwargs)
  File "D:\05_Software\Anaconda\lib\site-packages\setuptools-23.0.0-py3.5.egg\pk
g_resources\__init__.py", line 2245, in require
    items = working_set.resolve(reqs, env, installer)
[...]

PR #2103, opened 2017-03-14
[FIX] Discover widgets when some dependencies are missing

astaric:
##### Issue
When one of the requirements (of their requirements (or their requirements...)) is missing, Orange opens with no widgets.

To reproduce, uninstall keyring and open canvas

##### Description of changes
use .resolve() to load entry points

.resolve() is available from setuptools 11.3 forward (released in 2015),
does the same thing as .load(), but ommits the requirements check.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/workflows/__init__.py
+++ b/Orange/canvas/application/workflows/__init__.py
@@ -46,7 +46,7 @@ def example_workflows():
     for ep in workflow_entry_points():
         workflows = None
         try:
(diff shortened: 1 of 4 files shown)


astaric:
@ales-erjavec does this make sense? It solves the problem mentioned in #1753, but I might be missing something.

PR #2104, opened 2017-03-15
OWSilhouettePlot: Color each cluster with different color

VesnaT:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
@BlazZupan Waiting for your confirmation. I think the selection is not as prominent as I'd like, but others said gray is awful, so this was probably the best option. If you like it, I can merge.

PR #2113, opened 2017-03-17
create_changelog: Filter out already reported items

lanzagar:
##### Issue

Changes for a new release were repeating items from previous bugfix releases.

##### Description of changes
Grab all PR ids (#XXXX) from existing changelog and filter them out from the proposed new ones.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/scripts/create_changelog.sh
+++ b/scripts/create_changelog.sh
@@ -1,16 +1,21 @@
 #!/bin/bash
+set -u -e -o pipefail
 
 VERSION=$(grep '^VERSION' setup.py | sed -E "s/.*'([^']*)'/\1/")
 RELEASE_DATE=$(date "+%Y-%m-%d")
+
+# compile a list of already reported PRs
+reported="$(mktemp)"
+grep -E "#[0-9]{4}" -o CHANGELOG.md | sort > "$tempfile"
 
 echo "[$VERSION] - $RELEASE_DATE"
 echo "--------------------"
 echo '##### Enhancements'
 git log stable..master --first-parent --format='%s %b' |
-sed -E 's/.*#([0-9]+).*\[ENH\] *(.*)/\* \2 ([#\1](\.\.\/\.\.\/pull\/\1))/' |
-    grep -E '^\*'
+    sed -E 's/.*#([0-9]+).*\[ENH\] *(.*)/\* \2 ([#\1](\.\.\/\.\.\/pull\/\1))/' |
+    grep -E '^\*' | grep -v -F -f "$reported"
 echo
 echo "##### Bugfixes"
 git log stable..master --first-parent --format='%s %b' |
     sed -E 's/.*#([0-9]+).*\[FIX\] *(.*)/\* \2 ([#\1](\.\.\/\.\.\/pull\/\1))/' |
-    grep -E '^\*'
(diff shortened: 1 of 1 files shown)


scripts/create_changelog.sh
```diff
@@ -1,16 +1,20 @@
 #!/bin/bash
+set -u -e -o pipefail
 
 VERSION=$(grep '^VERSION' setup.py | sed -E "s/.*'([^']*)'/\1/")
 RELEASE_DATE=$(date "+%Y-%m-%d")
 
+# compile a list of already reported PRs
+grep -E "#[0-9]{4}" -o CHANGELOG.md | sort > /tmp/reported
```
astaric: You could use mktemp to create a temporary file.

lanzagar: Yes, much better.

scripts/create_changelog.sh
```diff
@@ -1,16 +1,21 @@
...
+tempfile="$(mktemp)"
+grep -E "#[0-9]{4}" -o CHANGELOG.md | sort > "$tempfile"
+
 echo "[$VERSION] - $RELEASE_DATE"
 echo "--------------------"
 echo '##### Enhancements'
 git log stable..master --first-parent --format='%s %b' |
 sed -E 's/.*#([0-9]+).*\[ENH\] *(.*)/\* \2 ([#\1](\.\.\/\.\.\/pull\/\1))/' |
-    grep -E '^\*'
+    grep -E '^\*' | grep -v -F -f "$tempfile"
```
astaric: Should this go into a new line?

scripts/create_changelog.sh
```diff
@@ -1,16 +1,21 @@
 #!/bin/bash
+set -u -e -o pipefail
 
 VERSION=$(grep '^VERSION' setup.py | sed -E "s/.*'([^']*)'/\1/")
 RELEASE_DATE=$(date "+%Y-%m-%d")
 
+# compile a list of already reported PRs
+tempfile="$(mktemp)"
```
astaric: Is there a better name that could be used? As /tmp/reported was used before, maybe "reported"?

scripts/create_changelog.sh
lanzagar: I don't see a reason for a new line, but if there are some bash style guidelines that someone prefers we can use them.

A better improvement would be to indent the line above.

scripts/create_changelog.sh
lanzagar: Better name for the variable or for the file?
I can change it if you have a suggestion.
`reported="$(mktemp)"` ?

scripts/create_changelog.sh
```diff
@@ -1,16 +1,21 @@
 #!/bin/bash
+set -u -e -o pipefail
 
 VERSION=$(grep '^VERSION' setup.py | sed -E "s/.*'([^']*)'/\1/")
 RELEASE_DATE=$(date "+%Y-%m-%d")
 
+# compile a list of already reported PRs
+reported="$(mktemp)"
+grep -E "#[0-9]{4}" -o CHANGELOG.md | sort > "$tempfile"
```
kernc: Where is `tempfile` assigned?

PR #2114, opened 2017-03-17
Report: Display only first 100 rows when reporting table

VesnaT:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/report/report.py
+++ b/Orange/canvas/report/report.py
@@ -197,6 +197,7 @@ class Report:
         :type header_columns: int
         :param num_format: numeric format, e.g. `{:.3}`
         """
+        row_limit = 100
         name, table = self._fix_args(name, table)
         join = "".join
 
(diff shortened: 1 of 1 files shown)


Orange/canvas/report/report.py
```diff
@@ -291,7 +292,7 @@ def report_list(data,
             return join("  <tr>\n    {}</tr>\n".format(
                 join(cells[rowi < header_rows or coli < header_columns]
                      .format(fmtnum(elm)) for coli, elm in enumerate(row))
-            ) for rowi, row in enumerate(data))
+            ) for rowi, row in zip(range(row_limit + 1), data))
```
janezd: Doesn't this output `row_limit + 1` rows?

janezd:
If the list is clipped, indicate this by adding a line (spanned across all columns)
`"+ {} more".format(len(...) - row_limit)`. Otherwise it gives an impression that there are exactly 100 items.

Orange/canvas/report/report.py
VesnaT: No, I guess header is the first.

janezd: Isn't the number of header rows given as an argument, so it can also be 0 or 3?

PR #2115, opened 2017-03-17
Mahalanobis: Remove dimension restrictions.

thocevar:
##### Issue
Mahalanobis distance (intentionally) didn't work when the number of features exceeded the number of samples.

##### Description of changes
The restriction doesn't seem to be applicable anymore.
Added error handling of too large and singular covariance matrices.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/tests/test_owdistances.py
+++ b/Orange/widgets/unsupervised/tests/test_owdistances.py
@@ -42,9 +42,21 @@ class TestOWDistances(WidgetTest):
         self.assertFalse(self.widget.Error.no_continuous_features.is_shown())
 
     def test_mahalanobis_error(self):
-        data = self.iris
-        data.X = np.vstack((data.X[0], data.X[0]))
-        data.Y = np.vstack((data.Y[0], data.Y[0]))
-        self.send_signal("Data", data)
-        self.widget.compute_distances(Mahalanobis, data)
-        self.assertTrue(self.widget.Error.mahalanobis_error.is_shown())
+        self.widget.metric_idx = METRICS.index(Mahalanobis)
+        self.widget.autocommit = True
+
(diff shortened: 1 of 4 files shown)


Orange/widgets/unsupervised/tests/test_owdistances.py
```diff
@@ -42,7 +42,9 @@ def test_error_message(self):
...
-        self.send_signal("Data", self.iris)
-        self.widget.compute_distances(Mahalanobis, self.iris)
-        self.assertTrue(self.widget.Error.too_few_observations.is_shown())
\ No newline at end of file
+        data = self.iris
+        data.X = np.vstack((data.X[0], data.X[0]))
+        data.Y = np.vstack((data.Y[0], data.Y[0]))
+        self.send_signal("Data", data)
+        self.widget.compute_distances(Mahalanobis, data)
+        self.assertTrue(self.widget.Error.mahalanobis_error.is_shown())
```
janezd: Can you also add a test that the error message disappears when

- the widget gets valid data,
- the widget gets `None` as data,
- the user changes metrics from Mahalanobis to something that can be computed?

These tests would be useful since not clearing the message is (or was) a common bug.

Since `compute_distances` obviously cleans the error message when called, a better test would be to set auto commit on and send different data sets. This way, the test would fail if somebody someday adds `if data is None` before calling `compute_distances` (or before calling a method that calls `compute_distances`).

The test doesn't need to pass invalid data. You can, but I usually prefer patching `distance.MahalanobisDistance.fit` with a mock that raises an exception.

janezd:
@thocevar, appveyor failed. Can you force-push again, so it will try again?

jerneju:
I think you need to rebase your branch.

PR #2116, opened 2017-03-17
[FIX] Widget Logistic Regression: can handle unused values

jerneju:
##### Issue

I can handle unused values
and instead of writing "coef" or sth similar I write a second value name
when only two values are considered.
https://sentry.io/biolab/orange3/issues/222366525/
https://sentry.io/biolab/orange3/issues/208659372/

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owlogisticregression.py
+++ b/Orange/widgets/model/owlogisticregression.py
@@ -92,10 +92,10 @@ class OWLogisticRegression(OWBaseLearner):
 def create_coef_table(classifier):
     i = classifier.intercept
     c = classifier.coefficients
-    if len(classifier.domain.class_var.values) > 2:
-        values = classifier.domain.class_var.values
+    if c.shape[0] > 2:
+        values = [classifier.domain.class_var.values[int(i)] for i in classifier.used_vals[0]]
     else:
-        values = ["coef"]
+        values = [classifier.domain.class_var.values[int(classifier.used_vals[0][1])]]
     domain = Domain([ContinuousVariable(value, number_of_decimals=7)
                      for value in values], metas=[StringVariable("name")])
     coefs = np.vstack((i.reshape(1, len(i)), c.T))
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/222366525/

jerneju:
https://sentry.io/biolab/orange3/issues/208659372/

Orange/widgets/model/tests/test_owlogisticregression.py
```diff
@@ -58,3 +57,37 @@ def test_output_coefficients(self):
...
+    def test_domain_with_more_values_than_table(self):
+        """
+        When the data with a domain which has more values than
+        a table was sent, the widget threw error Invalid number of variable columns.
+        GH-2116
+        """
+        data = Table("iris")
+        data = data[:80, :]
+        self.send_signal("Data", data)
+        self.widget.apply_button.button.click()
```
janezd: This only checks that the widget does not crash. Could you also test that the output is sensible?

Can you try it with keeping the first and the third class, or the second and the third?

Orange/widgets/model/owlogisticregression.py
```diff
@@ -92,10 +92,10 @@ def get_learner_parameters(self):
 def create_coef_table(classifier):
     i = classifier.intercept
     c = classifier.coefficients
-    if len(classifier.domain.class_var.values) > 2:
-        values = classifier.domain.class_var.values
+    if c.shape[0] > 2:
```
janezd: Isn't this something different? Wasn't the previous code correct? Isn't `classifier.coefficients` equal to the number of attributes + 1? Or am I overlooking something?

jerneju: Previous code was probably not correct. c.shape[0] is not always equal to len(classifier.domain.class_var.values). That is in the case when we have a domain with more class values than contained in a data table. And of course always in the case when number of that values is 2.

lanzagar:
Tests for this still do not pass. Have you looked into the issues caused by sgd reuse of the modified function?

PR #2117, opened 2017-03-17
gui.TableBarItem: Enable BarColorRole

janezd:
##### Issue

TableBarItem ignored ColorRole and always used the class color (or nothing).

##### Description of changes

The delegate now first tests whether it can use class colors, next it uses BarColorRole, otherwise it uses the default color.

##### Includes
- [X] Code changes


astaric:
How can I test this?

janezd:
Correlations widget in the prototypes add-on must show bars with different colors for positive and negative correlations. After pulling the PR referenced above by @VesnaT, of course.

PR #2118, opened 2017-03-17
[FIX] VizRankDialog: Stop computation when parent widget is deleted

VesnaT:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


astaric:
@janezd what about widgets emitting the signals and vizrank subscribing to them? Is there a special reason why vizrank did not use that in the first place?

janezd:
It would work almost exactly the same, except that we'd have to define a signal, emit it and then write a method that receives it ... and finally sets exactly the same flag that is now set directly by the widget.

The problem here is not the communication but synchronization, that is, what happens if the widget is torn down before vizrank reacts, so vizrank tries to communicate with a widget that Qt already killed, and so forth. Signals wouldn't solve this.

PR #2123, opened 2017-03-21
Pythagorean trees: Refactor

pavlin-policar:
##### Issue
Some of the Pythagrean tree and Pythagorean forest code was written very poorly with some duplication and poor handling of continuous vs discrete target classes.

##### Description of changes
Pretty major refactoring to remove all the duplicated code as well as several small fixes that were discovered along the way.

- Add `callback` to qt testing utilis `combobox_run_through_all` that enables us to call a function after the combobox changes to perform additional checks.
- Remove all the `tree_specific` stuff in the tree and forest in favour of more oop code so that messy callbacks are not needed.
- Make the `TreeAdapter` simpler by allowing it to store the data. This makes sense because since all the models have the data they were trained on, there really is no reason to complicate the adapter.
- Forest: Hide the trees while they are being drawn, since they would all bunch up in the first cell of the grid and flicker there until all would be drawn.
- Forest: Disable UI options while the trees are being drawn. This is so we can't accidentally change a setting while the trees are being drawn.
- Forest: The output is now consistent. Before, it used to be that when we changed the colors or the max depth on the forest the output wouldn't update, but when we changed the sizing, it would. This is fixed so the output changes only whenever we select a different tree.
- Tree: The tree can properly process all the display options that are passed from the forest (sizing, coloring, depth) so that when we select a tree on the forest widget, we get exactly what we are looking at in the tree widget. Before, the sizing was not functional.
- The tests now use the much better testing utils. I've also added several tests to the tree and forest to test the various UI setting combinations.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owpythagoreanforest.py
+++ b/Orange/widgets/visualize/owpythagoreanforest.py
@@ -206,7 +206,7 @@ class OWPythagoreanForest(OWWidget):
         self.ui_depth_slider.setMaximum(0)
 
     def _get_max_depth(self):
-        return max([tree.tree_adapter.max_depth for tree in self.ptrees])
+        return max(tree.tree_adapter.max_depth for tree in self.ptrees)
 
     def _get_forest_adapter(self, model):
         return SklRandomForestAdapter(model)
--- a/Orange/widgets/visualize/owpythagoreanforest.py
+++ b/Orange/widgets/visualize/owpythagoreanforest.py
@@ -268,7 +268,7 @@ class OWPythagoreanForest(OWWidget):
         if self.__prevent_commit:
             return
 
-        if len(self.scene.selectedItems()) == 0:
+        if not self.scene.selectedItems():
             self.send('Tree', None)
             # The selected tree index should only reset when model changes
             if self.model is None:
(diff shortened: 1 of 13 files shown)


janezd:
I construct a random forest for regression, showed it in the viewer and selected "Mean" for coloring. I then selected a tree and showed it in the Pythagorean tree viewer. The nodes are colored by means (which I like), but the color combo shows "Standard deviation". I guess the combo should either change when the viewer received a tree for which the node color is already defined.

Orange/widgets/visualize/owpythagoreanforest.py
```diff
@@ -192,11 +175,14 @@ def zoom_changed(self):
         self.grid.reflow(width)
         self.grid.setPreferredWidth(width)
 
-    # MODEL CHANGED METHODS
+    @contextmanager
+    def _prevent_commit(self):
```
janezd: I like this use of context managers. However, you should add try-finally to ensure that `__prevent_commit` is set to `False` in case of exception. (https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager)

Orange/widgets/visualize/owpythagoreanforest.py
```diff
@@ -218,36 +203,55 @@ def _clear_depth_slider(self):
...
+        """Temporarly disable the UI while trees may be redrawn."""
         self.ui_size_calc_combo.setEnabled(False)
+        self.ui_depth_slider.setEnabled(False)
+        self.ui_target_class_combo.setEnabled(False)
+        self.ui_zoom_slider.setEnabled(False)
+        yield
+        self.ui_size_calc_combo.setEnabled(True)
+        self.ui_depth_slider.setEnabled(True)
+        self.ui_target_class_combo.setEnabled(True)
+        self.ui_zoom_slider.setEnabled(True)
```
janezd: Same here: re-enabling should be in `finally` (and everything before that in `try`).

Issue #2125, opened 2017-03-21
Modelling issues


lanzagar:
##### Orange version

3.4.1

##### Issues

1. 'modelling' is missing in `Orange/__init__.py` and needs to be explicitly imported (unlike other submodules)
2. Some learners are not imported in `Orange/modelling/__init__.py` (i.e. Constant and SGD)
3. Other imports use *, but the modules do not define `__all__` so things are pulled into `modelling` that should not be
4. Not sure what the plan is for the `linear.py` submodule, but if it will only contain SGD it could be renamed to e.g. sgd.py?
5. SGDLearner() can't be called without parameters, because of `_change_kwargs`. I would like/expect defaults to be used when not providing specific losses etc.
6. SGDLearner uses `Learner.__call__` instead of `SklLearner.__call__` so the attribute `used_vals` is not set, and `params` does not exist. This is a problem at least in #2116, but could have other implications. I think probability predictions are not expanded correctly (see `SklModelClassification.__call__` or compare with e.g. logreg on first half of iris)!

lanzagar:
@pavlin-policar could you take a look?

pavlin-policar:
Yes, sure thing. I'll get right on it.

As far as **4.** is concerned, I don't know what would make more sense, because the `classification` and `regression` modules are not consistent with each other. In `regression`, we have a `linear.py` file that contains everything from linear regression to sgd. And there's also a `lbfgs` version of linear regression. On the other hand, in the `classification` module, logistic regression and sgd are in separate files, but they are both essentially regression... Having one file called `linear.py` made more sense to me, but it is not consistent. Logistic regression and linear regression will not be merged, so it may make more sense to call it `sgd`.

lanzagar:
I think this has the lowest priority and can wait for some discussion and consensus. I agree it makes sense in regression where it contains multiple classes. If it will remain with only one class in modelling, I think it makes less sense. But it is only a name after all, so several choices are good enough.

ajdapretnar:
@lanzagar Can this be considered resolved?

lanzagar:
Sure, why not.

PR #2128, opened 2017-03-21
[FIX] Learner: Generalize params and used_vals to all learners

pavlin-policar:
##### Issue
#2125 6.

##### Description of changes
The fix I've provided does basically one thing. We now expect every learner to have a `params` attribute. We also expect every learner to output a model with the `used_vals` field. These attributes seem very general to all learners and models and there doesn't seem to be any reason whatsoever for scikit wrapped learners to be special and have their own separate attributes on their public API. This makes everything very confusing and difficult to work with. By moving everything up to the base learner, it makes the API of any learners much simpler and more uniform.

In case this is not a desired solution, I can also provide a quick and dirty workaround that also fixes the main issue with SGD and other wrapped learners.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -82,6 +82,11 @@ class Learner(_ReprableWithPreprocessors):
     preprocessors = ()
     learner_adequacy_err_msg = ''
 
+    # Just a fallback in case learners don't specify their params. Params
+    # should be specified instance-wise, every learner can be constructed with
+    # different params, so we should store those.
+    params = {}
+
     def __init__(self, preprocessors=None):
         self.use_default_preprocessors = False
         if isinstance(preprocessors, Iterable):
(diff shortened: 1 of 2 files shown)


janezd:
I have a dream that one day Orange will be free of scikit-learn.

I have an unpleasant feeling that this makes Orange's learners more scikitish. If these parameters will be no longer needed when we replace scikit's learners with well-behaving, properly implemented algorithms, I'd prefer not putting them into base learners.

kernc:
Indeed, sklearn is a mature, stable, reliable, responsible, well-defined, disciplined and hygienic, well-tested, reasonably working, pragmatic, go-to state-of-the-art and industry-standard machine learning library with a huge developer and user community — as all such evidently awfully incompatible with what we're doing here. (Not sarcasm, actually.)

pavlin-policar:
Tbh, I think we should remove the `params` attribute all together since it is only ever useful in scikit learners and should be "protected" at the very least. Also, it's only ever used in tests in `WidgetLearnerTestMixin`. I can probably rewrite these tests so they wouldn't be needed anymore. It might however be reasonable to have attributes on learners like in [`TreeLearner`](https://github.com/biolab/orange3/blob/master/Orange/classification/tree.py#L63-L67), without the `params` of course. What do you think?

kernc:
The reason [sklearn tracks params](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/base.py#L212-L316) is aforementioned discipline. Looking forward, deffo not something we want to be bound by at all costs.

> I am also unsure of why the model somehow needs the parameters...

So it knows what set of learner params led to it. So it can be reproduced. In theory.

pavlin-policar:
> So it knows what set of learner params led to it. So it can be reproduced. In theory.

Well, yes, but so then why do only sklearn learners have this? And also wouldn't that be an argument to make for the learners to also have their own `params`?

I'd like to just remove `params` from the public API altogether and let sklearners use them internally, because really the only place where this is actually used is in one or two tests that can be changed.

PR #2129, opened 2017-03-22
[FIX] Predictions widget: handle similar but different domains

jerneju:
##### Issue
Widget throws an error similar to "index 1 is out of bounds for axis 0 with size 1". That occurs when predictions which are sent to Predictions widget are made from a data with different domain than a domain which is also sent to that widget. For instance, predictions predict only two possible target values but the data has also a third value.
https://sentry.io/biolab/orange3/issues/223474227/
https://sentry.io/biolab/orange3/issues/216151441/

##### Description of changes
Predictions table is extended to match the number of the data target's values.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -313,11 +313,14 @@ class OWPredictions(OWWidget):
             slots = self._valid_predictors()
             results = []
             class_var = self.class_var
-            nvalues = len(class_var.values)
             for p in slots:
                 values, prob = p.results
-                prob = numpy.c_[prob, numpy.zeros((prob.shape[0], nvalues - prob.shape[1]))]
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/223474227/

jerneju:
https://sentry.io/biolab/orange3/issues/216151441/

janezd:
Could you consider this PR in the light of #2183, which slightly changes the meaning of the widget's attribute `class_var` to represent the (common!) class of predictors? Are the two PRs in any conflict (not in terms of code, but does merging of one break the other?).

If it does, I'd prefer merging #2183 first, since it fixes a more critical bug.

markotoplak:
It seems that with changes from #2183 the problem does not appear anymore.

Issue #2036, opened 2017-02-21
Target selection in Mosaic


astaric:
Currently, mosaic plot uses class variable to color the visualization.

Add a new option to the control area that would allow the user to change the variable used for coloring. It should default to class_var (when available).

jerneju:
Merged #2133.

PR #2133, opened 2017-03-23
[ENH] Mosaic: Selectable color variable

jerneju:
##### Issue
Mosaic plot uses only class variable to color the visualization.
#2036 

##### Description of changes
A new option added to the control area that allows the user to change the variable used for coloring. It defaults to class_var when it is available.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
 from collections import defaultdict
 from functools import reduce
 from itertools import product, chain
@@ -10,7 +11,7 @@ from AnyQt.QtCore import Qt, QSize, pyqtSignal as Signal
 from AnyQt.QtGui import QColor, QPainter, QPen, QStandardItem
 from AnyQt.QtWidgets import QGraphicsScene, QGraphicsLineItem
 
-from Orange.data import Table, filter, Variable
+from Orange.data import Table, filter, Variable, Domain
 from Orange.data.sql.table import SqlTable, LARGE_TABLE, DEFAULT_SAMPLE_TIME
 from Orange.preprocess import Discretize
 from Orange.preprocess.discretize import EqualFreq
@@ -23,6 +24,7 @@ from Orange.widgets.settings import (
 from Orange.widgets.utils import to_html, get_variable_values_sorted
 from Orange.widgets.utils.annotated_data import (create_annotated_table,
                                                  ANNOTATED_DATA_SIGNAL_NAME)
+from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.visualize.utils import (
     CanvasText, CanvasRectangle, ViewWithPress, VizRankDialog)
 from Orange.widgets.widget import OWWidget, Default, Msg
@@ -118,7 +120,7 @@ class MosaicVizRank(VizRankDialog, OWComponent):
     def compute_attr_order(self):
         """
         Order attributes by Relief if there is a target variable. In case of
-        ties or without target, other by name.
+        ties or without target, order by name.
 
         Add the class variable at the beginning when not coloring by class
         distribution.
@@ -250,7 +252,7 @@ class MosaicVizRank(VizRankDialog, OWComponent):
     def row_for_state(self, score, state):
         """The row consists of attributes sorted by name; class is at the
         beginning, if present, so it's on the x-axis and not lost somewhere."""
-        class_var = self.master.data.domain.class_var
+        class_var = self.master.color_data.domain.class_var
         attrs = tuple(
             sorted((self.attr_ordering[x] for x in state),
                    key=lambda attr: (1 - (attr is class_var), attr.name)))
@@ -271,8 +273,6 @@ class OWMosaicDisplay(OWWidget):
                (ANNOTATED_DATA_SIGNAL_NAME, Table)]
 
     PEARSON, CLASS_DISTRIBUTION = 0, 1
-    interior_coloring_opts = ["Pearson residuals",
-                              "Class distribution"]
 
     settingsHandler = DomainContextHandler()
     use_boxes = Setting(True)
@@ -281,6 +281,7 @@ class OWMosaicDisplay(OWWidget):
     variable2 = ContextSetting("", exclude_metas=False)
     variable3 = ContextSetting("", exclude_metas=False)
     variable4 = ContextSetting("", exclude_metas=False)
+    variable_color = ContextSetting("", exclude_metas=False)
     selection = ContextSetting(set())
 
     BAR_WIDTH = 5
@@ -310,6 +311,8 @@ class OWMosaicDisplay(OWWidget):
         self.unprocessed_subset_data = None
         self.subset_data = None
 
+        self.color_data = None
+
         self.areas = []
 
         self.canvas = QGraphicsScene()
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owmosaic.py
```diff
@@ -518,6 +533,30 @@ def send_selection(self):
...
 
+    def get_values(self, data):
+        attr_list, variable_color = self.get_attr_list()
+        if variable_color == '' or None:
+            if len(data.domain.class_vars) > 0:
+                color_var = data.domain.class_vars[0]
+            elif data.domain.attributes[0]:
+                color_var = data.domain.attributes[0]
+            else:
+                color_var = data.domian.metas[0]
```
janezd: `domian`? Add tests for this.

janezd: What if the first meta attribute is a string?

Orange/widgets/visualize/owmosaic.py
```diff
@@ -281,8 +281,11 @@ class OWMosaicDisplay(OWWidget):
     variable2 = ContextSetting("", exclude_metas=False)
     variable3 = ContextSetting("", exclude_metas=False)
     variable4 = ContextSetting("", exclude_metas=False)
+    variable_color = ContextSetting("", exclude_metas=False)
     selection = ContextSetting(set())
 
+    color_var = None
```
janezd: This does work, but we define instance attributes in `__init__`. Settings are an exception.

janezd: ... if this line is needed at all. I may be stupid, by I don't see what `color_var` is doing and whether it's a string or a list of strings.

jerneju: It is not needed, I thought that I needed it, but then I have forgotten to erased it. Any yes, Lint reports that this variable is unused.

Orange/widgets/visualize/owmosaic.py
jerneju: If the first meta attribute is a string, nothing happens. But I am still testing.

jerneju:
Vizrank has to be changed. At the moment it still uses class variable and not color.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -410,13 +440,29 @@ def showEvent(self, ev):
...
+        if variable_color is None:
+            color_var = None
+        else:
+            color_var = self.data.domain[variable_color]
+        class_vars = color_var
+        attributes = [v for v in self.data.domain if v != color_var]
+        metas = [v for v in self.data.domain.metas if v != color_var]
+        domain = Domain(attributes, class_vars, metas)
+        data = self.data.from_table(domain, self.data)
+        self.set_data(data)
```
janezd: Do I see correctly that the combo calls `update_colors`, which calls `change_domain`, which in turn calls `set_data`? And `set_data` is a callback for signal Data, which ... basically resets the widget, including closing the context and so forth?

I think widgets should never call signal handlers like `set_data`. If `set_data` does some things that need to be done when changing the combo (like stopping the vizrank,  updating the graph), extract those into a separate function and call it from `set_data` and from `update_colors` (or `change_domain`).

*Besides*, remove `change_domain` altogether. The widget should output the same domain. Changing the coloring must not replace the class attribute, it should only affect the visual appearance.

janezd: I think that changing the color should retain the selection.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -331,9 +332,27 @@ def __init__(self):
...
+            box2, self, "variable_color",
+            callback=self.update_colors,
+            model=self.color_model, **common_options)
+        """
+        self.cb_attr_color = gui.comboBox(
+            box2, self, value="variable_color",
+            orientation=Qt.Horizontal, contentsLength=12,
+            callback=self.update_colors,
+            sendSelectedValue=True, valueType=str)
+
```
janezd: The changed GUI is confusing.
- If the combo is set to None, the two radio buttons below have no effect.
- If the radio button is set to "Pearson", then the combo has no effect.

1. Replace the None option in the combo with *Pearson residuals*
2. You can then eliminate the radio buttons. 
3. The checkbox should then be active (only if) combo is not Pearson residuals. If you use DomainModel, you will check for this with, simply, `self.variable_color is not None`.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -281,6 +281,7 @@ class OWMosaicDisplay(OWWidget):
     variable2 = ContextSetting("", exclude_metas=False)
     variable3 = ContextSetting("", exclude_metas=False)
     variable4 = ContextSetting("", exclude_metas=False)
+    variable_color = ContextSetting("(No class)", exclude_metas=False)
```
janezd: If you use a model (see bellow), change the default to `None`.

janezd:
Also, please rebase.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -383,19 +382,34 @@ def init_combos(self, data):
...
+        idx = 1 if data.domain.class_var else 0
+        self.variable_color = self.cb_attr_color.itemText(idx)
 
     def get_attr_list(self):
         return [
             a for a in [self.variable1, self.variable2,
                         self.variable3, self.variable4]
-            if a and a != "(None)"]
+            if a and a != "(None)"], \
+               self.variable_color if self.variable_color != "(Pearson residuals)" else None
```
lanzagar: It looks to me like this method is now only called in one additional place (L484), where you discard the first list of attributes and only use the color variable.
Why don't you just use `self.variable_color` there, and you even have an `if` already (just change "is None" to "== Pearson")

I would, however, prefer to put the string "(Pearson residuals)" in a constant at the top and use that. You currently already use the hard-coded string 7 times in this PR.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -383,19 +382,34 @@ def init_combos(self, data):
             if attr.is_discrete or attr.is_continuous:
                 for combo in self.attr_combos:
                     combo.addItem(icons[attr], attr.name)
+        self.cb_attr_color.addItem("(Pearson residuals)")
+        for attr in chain(data.domain.class_vars, data.domain.attributes, data.domain.metas):
```
lanzagar: IIRC we wanted to transition to a new order in dropdowns of: class_vars, metas, attributes?
Since some data can have thousands of attributes and scrolling to the end can be a pain.
To be double-checked.

Even better - use DomainModel. I think it should be possible to add a custom string at the beginning.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -383,19 +382,34 @@ def init_combos(self, data):
...
+            if attr.is_discrete or attr.is_continuous:
```
lanzagar: `attr.is_primitive`

Orange/widgets/visualize/owmosaic.py
```diff
@@ -468,6 +478,28 @@ def reset_graph(self):
...
+        _, variable_color = self.get_attr_list()
+        if variable_color is None:
+            color_var = None
+            self.interior_coloring = self.PEARSON
+            self.bar_button.setEnabled(False)
+        else:
+            color_var = self.data.domain[variable_color]
+            self.interior_coloring = self.CLASS_DISTRIBUTION
+            self.bar_button.setEnabled(True)
+        class_vars = color_var
```
lanzagar: just use `color_var`?

Orange/widgets/visualize/owmosaic.py
jerneju: Yes, DomainModel enables us to add a custom string at the beginning. Now class variables, attributes and metas are added into a combo in an alphabetical order.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -331,13 +334,18 @@ def __init__(self):
         self.vizrank, self.vizrank_button = MosaicVizRank.add_vizrank(
             box, self, "Find Informative Mosaics", self.set_attr)
 
-        self.rb_colors = gui.radioButtonsInBox(
-            self.controlArea, self, "interior_coloring",
-            self.interior_coloring_opts, box="Interior Coloring",
-            callback=self.coloring_changed)
+        box2 = gui.vBox(self.controlArea, box="Interior Coloring")
+        dmod = DomainModel
```
kernc: No.

jerneju: Please tell more.

Issue #2038, opened 2017-02-21
Autocommit on all widgets


astaric:
All widgets should have autocommit/autoapply/autosend set when they are used for the first time.

jerneju:
Should Python Script widget have auto execute set as well when it is used for the first time? @BlazZupan

PR #2136, opened 2017-03-24
[FIX] All widgets are set to auto* when they are used for the first time

jerneju:
##### Description of changes
All widgets are set to autocommit/autoapply/autosend/autoexecute when they are used for the first time.
Python Script test changed.
GH-2038

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpythonscript.py
+++ b/Orange/widgets/data/owpythonscript.py
@@ -376,7 +376,7 @@ class OWPythonScript(widget.OWWidget):
         Setting([Script("Hello world", "print('Hello world')\n")])
     currentScriptIndex = Setting(0)
     splitterState = Setting(None)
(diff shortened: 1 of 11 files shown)


Orange/widgets/data/owpythonscript.py
```diff
@@ -386,7 +386,7 @@ def __init__(self):
         self.in_learner = None
         self.in_classifier = None
         self.in_object = None
-        self.auto_execute = False
```
astaric: This is a bad idea. If it is supposed to be a setting, changing it to false/true in __init__ overrides the saved value.

jerneju: Fixed.

PR #2145, opened 2017-03-28
[ENH] OWTestLearners: Cross validation by feature

VesnaT:
##### Issue

Add type of CV where number of feature values determines number of folds.
Use only discrete meta features.

##### Description of changes
Add radio button option to select this type of CV.
Add combo box to select feature to give nr. of folds.
Documentation comes when gui is approved.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -14,7 +14,7 @@ from Orange.util import OrangeWarning
 from Orange.data import Table, Domain, ContinuousVariable, DiscreteVariable
 
 __all__ = ["Results", "CrossValidation", "LeaveOneOut", "TestOnTrainingData",
-           "ShuffleSplit", "TestOnTestData", "sample"]
+           "ShuffleSplit", "TestOnTestData", "sample", "CrossValidationFeature"]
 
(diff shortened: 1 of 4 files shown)


janezd:
I know this will come as a totally unusual surprise, but I'm disapproving the GUI. :)

The feature, if I read the code correctly, doesn't determine the number of folds, but the folds themselves. This looks like a useful feature, but I wouldn't expect it in cross-validation, and particularly not in the "Number of folds" combo. "From feature" also doesn't fit into an unexpanded combo.

The "from feature" option fundamentally changes the sampling procedure in the sense that it is no longer random.

As a minor argument against this GUI, we also discourage show/hide, and prefer enabling/disabling instead, where this could be sensibly avoided. 

One solution would be to add a new radio button "Cross validation by feature" below "Cross validation". With this, this functionality would be more prominent and we'd no longer need hiding/showing of parts of the interface.

The radio button can be disable if there are no meta attributes.

If you do this, don't forget to provide appropriate migration of settings.

markotoplak:
The main functionality is not tested (the corresponding radio button is never clicked). :)

When cross validation by feature is used for a file, which is then disconnected and connected back, the setting are reset.

Issue #2143, opened 2017-03-27
Edit Domain should now allow us to write duplicate names


jerneju:
##### Orange version
3.5dev

##### Expected behavior
Edit Domain enables us to write duplicate names. That should not be allowed. User may be prevented to write a duplicate name and warned or a number may be added automatically next to a duplicate.

##### Actual behavior
![screenshot_20170327_155607](https://cloud.githubusercontent.com/assets/22157215/24359981/3779f4f8-1306-11e7-83fc-72a2a1294572.png)

![screenshot_20170327_155640](https://cloud.githubusercontent.com/assets/22157215/24359990/3cff9d88-1306-11e7-9fe6-58b00b53b2ef.png)

##### Steps to reproduce the behavior
Put File and Edit Domain widgets and then connect them. Next, open Edit Domain widget and change features' names.

PR #2146, opened 2017-03-28
[FIX] Edit Domain: Prevent duplicate variable names

jerneju:
##### Issue
Edit Domain enables us to write duplicate names and that should be allowed. However user should be informed about that.
#2143

##### Description of changes
User is warned when writing a duplicate name. Also, empty spaces are removed in the beginning and in the end of a new name.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/oweditdomain.py
+++ b/Orange/widgets/data/oweditdomain.py
@@ -213,7 +213,7 @@ class VariableEditor(QWidget):
     def get_data(self):
         """Retrieve the modified variable.
         """
-        name = str(self.name_edit.text())
+        name = str(self.name_edit.text()).strip()
         labels = self.labels_model.get_dict()
 
         # Is the variable actually changed.
@@ -226,12 +226,15 @@ class VariableEditor(QWidget):
 
         return var
 
+    def is_legal(self):
+        name = str(self.name_edit.text()).strip()
+        return not len(name) == 0
+
     def is_same(self):
         """Is the current model state the same as the input.
         """
-        name = str(self.name_edit.text())
+        name = str(self.name_edit.text()).strip()
         labels = self.labels_model.get_dict()
-
         return (self.var is not None and name == self.var.name and
                 labels == self.var.attributes)
 
@@ -243,7 +246,7 @@ class VariableEditor(QWidget):
         self.labels_model.set_dict({})
 
     def maybe_commit(self):
-        if not self.is_same():
+        if not self.is_same() and self.is_legal():
             self.commit()
 
     def commit(self):
@@ -317,7 +320,7 @@ class DiscreteVariableEditor(VariableEditor):
     def get_data(self):
         """Retrieve the modified variable
         """
-        name = str(self.name_edit.text())
+        name = str(self.name_edit.text()).strip()
         labels = self.labels_model.get_dict()
         values = map(str, self.values_model)
 
@@ -401,6 +404,10 @@ class OWEditDomain(widget.OWWidget):
         self.editor_stack.addWidget(VariableEditor())
 
         box.layout().addWidget(self.editor_stack)
+
+        self.Error.add_message(
+            "duplicate_var_name",
+            "A variable name is duplicated.")
 
     @check_sql_input
     def set_data(self, data):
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/oweditdomain.py
```diff
@@ -196,11 +197,12 @@ def _setup_gui_labels(self):
 
         self.main_form.addRow("Labels:", vlayout)
 
-    def set_data(self, var):
+    def set_data(self, var, var_names=None):
```
janezd: Using `var_names=()` as default would simplify `is_same`, wouldn't it?

Orange/widgets/data/oweditdomain.py
```diff
@@ -308,16 +316,16 @@ def _setup_gui_values(self):
         self.values_model.dataChanged.connect(self.on_values_changed)
         self.main_form.addRow("Values:", self.values_edit)
 
-    def set_data(self, var):
+    def set_data(self, var, var_names=None):
```
janezd: Also here, you can probably replace `None` with `()`.

Orange/widgets/data/oweditdomain.py
```diff
@@ -505,7 +518,7 @@ def clear_editor(self):
             current.variable_changed.disconnect(self._on_variable_changed)
         except Exception:
             pass
-        current.set_data(None)
+        current.set_data(None, None)
```
janezd: And then `()` (or nothing) instead of the second `None`, too.

Orange/widgets/data/oweditdomain.py
```diff
@@ -229,11 +231,17 @@ def get_data(self):
...
-
-        return (self.var is not None and name == self.var.name and
-                labels == self.var.attributes)
+        if self.var_names is not None:
+            temp_var_names = list(self.var_names)
+            if self.var is not None and self.var.name in temp_var_names:
+                temp_var_names.remove(self.var.name)
+        else:
+            temp_var_names = []
+        return ((name in temp_var_names)
```
janezd: This function checks whether it is *necessary* to change the domain. You shouldn't abuse it to check whether it's *legal* to change the domain. (See the general comment.)

Orange/widgets/data/oweditdomain.py
```diff
@@ -402,6 +406,10 @@ def __init__(self):
 
         box.layout().addWidget(self.editor_stack)
 
+        self.Warning.add_message(
```
lanzagar: I think two variables with the same name should produce an error and require a fix before proceeding (instead of displaying a warning and working on)

Orange/widgets/data/oweditdomain.py
```diff
@@ -514,10 +522,15 @@ def _on_variable_changed(self):
...
         # Replace the variable in the 'Domain Features' view/model
         old_var = self.input_vars[self.selected_index]
-        new_var = editor.get_data().copy(compute_value=Orange.preprocess.transformation.Identity(old_var))
+        new_var = editor.get_data().copy(
+            compute_value=Orange.preprocess.transformation.Identity(old_var)
+        )
         self.domain_model[self.selected_index] = new_var
-
-
+        var_names = [vn.name for vn in self.domain_model]
```
lanzagar: This is now checked only on variable changes. If I see the warning and reset a variable to the original, the warning does not disappear (the reset functions don't reach this code).
Maybe the check could be moved to commit:
- if everything is ok, remove the error and proceed
- if not, show the error and send None

lanzagar:
Also, editing a variable name and pressing enter does not work... Maybe that's a separate issue, but if it is easy to fix it can perhaps be added here?

Orange/widgets/data/oweditdomain.py
jerneju: Well, @janezd has different opinion (see above).

lanzagar: I think @janezd didn't like that the change was ignored without a message. I don't see anything about whether the message should be a warning or an error.
We do not want to allow duplicate names (which is not just a cosmetic guideline, it can cause serious problems in the code, because we can index by names etc), so the widget should not output incorrect data. In that case an error is the only appropriate type of message.

jerneju:
@lanzagar: Yes, pressing enter is another issue.

PR #2149, opened 2017-03-29
[FIX] Correspondence: Prevent crashing when cont attr has one value

jerneju:
##### Issue
Widget crashes on discrete attributes with only one value.
https://sentry.io/biolab/orange3/issues/243133261/

##### Description of changes
Check if there is a second value.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owcorrespondence.py
+++ b/Orange/widgets/unsupervised/owcorrespondence.py
@@ -2,7 +2,7 @@ import sys
 
 from collections import namedtuple, OrderedDict
 
-import numpy as np
+import numpy
 
 from AnyQt.QtWidgets import QListView, QApplication
 from AnyQt.QtGui import QBrush, QColor, QPainter
@@ -203,7 +203,7 @@ class OWCorrespondenceAnalysis(widget.OWWidget):
 
         p_axes = self._p_axes()
 
-        if points is None:
+        if points == None:
             return
 
         if len(variables) == 2:
@@ -213,7 +213,7 @@ class OWCorrespondenceAnalysis(widget.OWWidget):
         else:
             points = self.ca.row_factors[:, p_axes]
             counts = [len(var.values) for var in variables]
-            range_indices = np.cumsum([0] + counts)
+            range_indices = numpy.cumsum([0] + counts)
             ranges = zip(range_indices, range_indices[1:])
             points = [points[s:e] for s, e in ranges]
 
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/243133261/

Orange/widgets/unsupervised/tests/test_owcorrespondence.py
```diff
@@ -37,3 +37,18 @@ def test_data_values_in_column(self):
...
+        GH-2149
+        """
+        table = Table(
+            Domain(
+                [DiscreteVariable("a", values=["0"])]
+            ),
+            list(zip(
+                ["0", "0", "0"]
+            )))
+        self.send_signal("Data", table)
```
janezd: Cool trick. You will get the same results with `list(zip("000"))`.

However, `[('0',)]*3` or simply `[(0,), (0,), (0,)]` or `np.zeros((3, 1))` would be more reader-friendly.

Orange/widgets/unsupervised/owcorrespondence.py
```diff
@@ -109,7 +109,9 @@ def set_data(self, data):
             self.varlist[:] = [var for var in data.domain.variables
                                if var.is_discrete]
             self.selected_var_indices = [0, 1][:len(self.varlist)]
-            self.component_x, self.component_y = 0, 1
+            self.component_x = 0
+            self.component_y = 1 if \
+                len(self.varlist[self.selected_var_indices[-1]].values) > 1 else 0
```
janezd: You don't need `len` here.

(I'd actually write `self.component_y = bool(self.varlist[self.selected_var_indices[-1]].values)`, but basically just to annoy @astaric. :)

lanzagar: He does need len, because he wants to check if it is >1 and not >0.
It is also harder to annoy @astaric, but the `if ... else` is unnecessary, so:
`self.component_y = len(self.varlist[self.selected_var_indices[-1]].values) > 1`
or if someone is the strict type:
`int(len(self.varlist[self.selected_var_indices[-1]].values) > 1)`

Orange/widgets/unsupervised/owcorrespondence.py
```diff
@@ -350,6 +351,9 @@ def gsvd(M, Wu, Wv):
     F = numpy.c_[D_r] * U * D
     G = numpy.c_[D_c] * V.T * D
 
+    if F.shape == (1, 1) and F == numpy.array([[0]]):
```
janezd: I had to read this twice... Why not `if F.shape == (1, 1) and F[0, 0] == 0`? Simpler and less work for all. Also, F[0, 0] = 1 below. Sorry for overlooking this before.

jerneju: Fixed.

jerneju:
Fixed.

lanzagar:
Seems to be broken.
Try titanic data set and show first two features (compare with master).

jerneju:
https://sentry.io/biolab/orange3/issues/238149545/

PR #2152, opened 2017-03-29
[FIX] Scatter Plot: dealing with scipy sparse matrix

### LINKED ISSUE #2157: Scatter plot fails on sparse
##### Orange version

3.5.dev

##### Expected behavior
Scatter Plot works with sparse.

##### Actual behavior
Scatter Plot fails on sparse.

Connect Corpus to Scatter Plot.

jerneju:
##### Issue
Orange add-on "Text mining" sends "Corpus" instead of "Table". "Corpus" uses sparse matrices and numpy hstack does not know how to handle them. That is why scipy sparse matrix is converted into numpy array.
https://sentry.io/biolab/orange3/issues/243775748/
Fixes #2157.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -1,4 +1,5 @@
 import numpy as np
+import scipy.sparse as sp
 
 from AnyQt.QtCore import Qt, QTimer
 from AnyQt.QtGui import (
@@ -394,7 +395,8 @@ class OWScatterPlot(OWWidget):
 
     # called when all signals are received, so the graph is updated only once
     def handleNewSignals(self):
-        self.graph.new_data(self.data_metas_X, self.subset_data)
+        self.graph.new_data(self.sparse_to_dense(self.data_metas_X),
+                            self.sparse_to_dense(self.subset_data))
         if self.attribute_selection_list and \
                 all(attr in self.graph.domain
                         for attr in self.attribute_selection_list):
@@ -406,6 +408,37 @@ class OWScatterPlot(OWWidget):
         self.cb_reg_line.setEnabled(self.graph.can_draw_regresssion_line())
         self.apply_selection()
         self.unconditional_commit()
+
+    def prepare_data(self):
+        """
+        Only when dealing with sparse matrices.
+        GH-2152
+        """
+        self.graph.new_data(self.sparse_to_dense(self.data_metas_X),
+                            self.sparse_to_dense(self.subset_data),
+                            new=False)
+
+    def sparse_to_dense(self, input_data=None):
+        self.vizrank_button.setEnabled(not (self.data and self.data.is_sparse()))
+        if input_data is None or not input_data.is_sparse():
+            return input_data
+        keys = []
+        attrs = {self.attr_x,
+                 self.attr_y,
+                 self.graph.attr_color,
+                 self.graph.attr_shape,
+                 self.graph.attr_size,
+                 self.graph.attr_label}
+        for i, attr in enumerate(input_data.domain):
+            if attr in attrs:
+                keys.append(i)
+        new_domain = input_data.domain.select_columns(keys)
+        dmx = Table.from_table(new_domain, input_data)
+        dmx.X = dmx.X.toarray()
+        # TODO: remove once we make sure Y is always dense.
+        if sp.issparse(dmx.Y):
+            dmx.Y = dmx.Y.toarray()
+        return dmx
 
     def apply_selection(self):
         """Apply selection saved in workflow."""
@@ -441,12 +474,14 @@ class OWScatterPlot(OWWidget):
         self.update_attr()
 
     def update_attr(self):
+        self.prepare_data()
         self.update_graph()
         self.cb_class_density.setEnabled(self.graph.can_draw_density())
         self.cb_reg_line.setEnabled(self.graph.can_draw_regresssion_line())
         self.send_features()
 
     def update_colors(self):
+        self.prepare_data()
         self.cb_class_density.setEnabled(self.graph.can_draw_density())
 
     def update_density(self):
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -591,17 +591,18 @@ class OWScatterPlotGraph(gui.OWComponent, ScaleScatterPlotData):
         text = self.tiptexts.get(int(modifiers), self.tiptexts[0])
         self.tip_textitem.setHtml(text)
 
-    def new_data(self, data, subset_data=None, **args):
-        self.plot_widget.clear()
-        self.remove_legend()
+    def new_data(self, data, subset_data=None, new=True, **args):
+        if new:
+            self.plot_widget.clear()
+            self.remove_legend()
 
-        self.density_img = None
-        self.scatterplot_item = None
-        self.scatterplot_item_sel = None
-        self.reg_line_item = None
-        self.labels = []
-        self.selection = None
-        self.valid_data = None
+            self.density_img = None
+            self.scatterplot_item = None
+            self.scatterplot_item_sel = None
+            self.reg_line_item = None
+            self.labels = []
+            self.selection = None
+            self.valid_data = None
 
         self.subset_indices = set(e.id for e in subset_data) if subset_data else None
 
@@ -776,12 +777,14 @@ class OWScatterPlotGraph(gui.OWComponent, ScaleScatterPlotData):
         return size_data
 
     def update_sizes(self):
+        self.master.prepare_data()
+        self.update_point_size()
+
+    def update_point_size(self):
         if self.scatterplot_item:
             size_data = self.compute_sizes()
             self.scatterplot_item.setSize(size_data)
             self.scatterplot_item_sel.setSize(size_data + SELECTION_WIDTH)
-
-    update_point_size = update_sizes
 
     def get_color_index(self):
         if self.attr_color is None:
@@ -907,6 +910,9 @@ class OWScatterPlotGraph(gui.OWComponent, ScaleScatterPlotData):
 
     def update_colors(self, keep_colors=False):
         self.master.update_colors()
+        self.update_alpha_value(keep_colors)
+
+    def update_alpha_value(self, keep_colors=False):
         if self.scatterplot_item:
             pen_data, brush_data = self.compute_colors(keep_colors)
             pen_data_sel, brush_data_sel = self.compute_colors_sel(keep_colors)
@@ -922,8 +928,6 @@ class OWScatterPlotGraph(gui.OWComponent, ScaleScatterPlotData):
                 elif self.density_img:
                     self.plot_widget.removeItem(self.density_img)
 
-    update_alpha_value = update_colors
-
     def create_labels(self):
         for x, y in zip(*self.scatterplot_item.getData()):
             ti = TextItem()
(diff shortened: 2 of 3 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/243775748/

Orange/widgets/utils/scaling.py
```diff
@@ -47,7 +47,9 @@ def _compute_scaled_data(self):
             return
 
         Y = data.Y if data.Y.ndim == 2 else np.atleast_2d(data.Y).T
-        self.original_data = np.hstack((data.X, Y)).T
+        self.original_data = np.hstack((data.X if "csr_matrix" not in str(type(data.X))
+                                        else data.X.toarray(),
```
nikicc: We shouldn't transform sparse matrix to dense here but should rather adapt the method to work on sparse matrices.

Orange/widgets/utils/scaling.py
```diff
@@ -47,7 +47,9 @@ def _compute_scaled_data(self):
             return
 
         Y = data.Y if data.Y.ndim == 2 else np.atleast_2d(data.Y).T
-        self.original_data = np.hstack((data.X, Y)).T
+        self.original_data = np.hstack((data.X if "csr_matrix" not in str(type(data.X))
```
nikicc: To check the type it's better to use `issinstance` than checking if string is inside type's string representation. However, in this case checking for `csr_matrix` is not sufficient since this doesn't work for other sparse matrix formats (like `csc` or `dok`).

The best way to check for sparsity is to use scipy's `issparse` For example:
```
import scipy as sp
sp.issparse(data.X)
```

Orange/widgets/utils/scaling.py
jerneju: Yes, firstly I used scipy's hstack. But as you warned me, then there are problems with `self.original_data`.

nikicc:
@jerneju I haven't checked much of the code but this still fails for me on sparse:
screen shot 2017-04-07 at 10 19 42

jerneju:
@nikicc: I will write the code again and that time I am planning to use new methods added in _util.py_.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -419,17 +420,36 @@ def set_subset_data(self, subset_data):
...
         self.attribute_selection_list = None
-        self.update_graph()
+        if not sp.issparse(self.data.X):
+            self.update_graph()
         self.cb_class_density.setEnabled(self.graph.can_draw_density())
         self.cb_reg_line.setEnabled(self.graph.can_draw_regresssion_line())
         self.unconditional_commit()
+        self.corpus_to_table()
+
+    def corpus_to_table(self):
```
nikicc: Could you rename this to something else? In theory we could have sparse table also in Orange, nut just in Text.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -394,18 +395,40 @@ def set_subset_data(self, subset_data):
 
     # called when all signals are received, so the graph is updated only once
     def handleNewSignals(self):
-        self.graph.new_data(self.data_metas_X, self.subset_data)
+        if self.data is None or (self.data and not sp.issparse(self.data.X)):#  table.issparse() !!!
```
nikicc: Couldn't this simply be `if self.data is None or not sp.issparse(self.data.X)`?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -394,18 +395,40 @@ def set_subset_data(self, subset_data):
...
+            self.graph.new_data(self.data_metas_X, self.subset_data)
         if self.attribute_selection_list and \
                 all(attr in self.graph.domain
                         for attr in self.attribute_selection_list):
             self.attr_x = self.attribute_selection_list[0]
             self.attr_y = self.attribute_selection_list[1]
         self.attribute_selection_list = None
+        #if not sp.issparse(self.data.X):
```
nikicc: This is probably leftover from some debugging. Can you please remove it?

Orange/widgets/visualize/owscatterplot.py
nikicc: Or maybe even better:
```
if self.data is not None and self.data.is_sparse():
    self.sparse_to_dense()
else:
    self.graph.new_data(self.data_metas_X, self.subset_data)
```

Doing so you also don't need that if at the beginning of `sparse_to_dense` method.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -394,18 +395,40 @@ def set_subset_data(self, subset_data):
...
         self.apply_selection()
         self.unconditional_commit()
+        self.sparse_to_dense()
+
+    def sparse_to_dense(self):
+        if self.data is None or not self.data.is_sparse():
+            return
+        keys = []
+        for i, attr in enumerate(self.data.domain):
+            if attr in set([self.attr_x, self.attr_y, self.graph.attr_color]):
```
nikicc: I'm probably nitpicking here, but could you move this outside the loop into a variable so you do not create a new set in each iteration?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -394,18 +395,40 @@ def set_subset_data(self, subset_data):
...
         self.cb_reg_line.setEnabled(self.graph.can_draw_regresssion_line())
         self.apply_selection()
         self.unconditional_commit()
+        self.sparse_to_dense()
+
+    def sparse_to_dense(self):
+        if self.data is None or not self.data.is_sparse():
+            return
+        keys = []
+        for i, attr in enumerate(self.data.domain):
```
nikicc: This only iterates over features and class attributes, but not metas. What happens if we color by discrete variable inside metas?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -394,18 +395,40 @@ def set_subset_data(self, subset_data):
...
+            if attr in set([self.attr_x, self.attr_y, self.graph.attr_color]):
+                keys.append(i)
+        new_domain = self.data.domain.select_columns(keys)
+        dmx = Table.from_table(new_domain, self.data)
+        dmx = self.move_primitive_metas_to_X(dmx)
+        dmx.X = dmx.X.toarray()
+        if sp.issparse(dmx.Y):
```
nikicc: Can you please add a todo here? Something like:
#TODO: remove once we make sure Y is always dense.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -394,18 +395,40 @@ def set_subset_data(self, subset_data):
...
+            dmx.Y = dmx.Y.toarray()
+        self.subset_data = None
```
nikicc: What if someone provide sparse subset data?

jerneju:
Code updated.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -407,6 +410,38 @@ def handleNewSignals(self):
...
 
+    def prepare_data(self):
+        self.graph.new_data(self.sparse_to_dense(self.data_metas_X),
+                            self.sparse_to_dense(self.subset_data),
+                            new=False
+                           )
+        self.update_graph()
+
+    def sparse_to_dense(self, input_data=None):
+        if input_data is None:
```
nikicc: This can be simplified to:
```
if input_data is None or not input_data.is_sparse():
    return input_data
```

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -394,7 +395,9 @@ def set_subset_data(self, subset_data):
 
     # called when all signals are received, so the graph is updated only once
     def handleNewSignals(self):
-        self.graph.new_data(self.data_metas_X, self.subset_data)
+        self.graph.new_data(self.sparse_to_dense(self.data_metas_X),
+                            self.sparse_to_dense(self.subset_data)
+                           )
```
nikicc: Can you move this `)` at the end of previous line?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -407,6 +410,38 @@ def handleNewSignals(self):
         self.apply_selection()
         self.unconditional_commit()
 
+    def prepare_data(self):
+        self.graph.new_data(self.sparse_to_dense(self.data_metas_X),
+                            self.sparse_to_dense(self.subset_data),
+                            new=False
+                           )
```
nikicc: Can you move this `)` at the end of previous line?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -407,6 +410,38 @@ def handleNewSignals(self):
...
+        self.update_graph()
```
nikicc: Is this update graph really needed?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -407,6 +410,38 @@ def handleNewSignals(self):
...
+
+    def sparse_to_dense(self, input_data=None):
+        if input_data is None:
+            return None
+        if not input_data.is_sparse():
+            return input_data
+        self.vizrank_button.setVisible(False)  # not for subset data?
+        keys = []
+        attrs = set([self.attr_x,
```
nikicc: My PyCharm suggests: `function call can be replaces with set literal`.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -407,6 +410,38 @@ def handleNewSignals(self):
...
+                            new=False
+                           )
+        self.update_graph()
+
+    def sparse_to_dense(self, input_data=None):
+        if input_data is None:
+            return None
+        if not input_data.is_sparse():
+            return input_data
+        self.vizrank_button.setVisible(False)  # not for subset data?
```
nikicc: Maybe it's better to use `setEnabled` here.

nikicc: Also, once this gets disabled it is not re-enabled if dense data comes in. I suggest we add something like:
```
self.vizrank_button.setEnabled(not (self.data and self.data.is_sparse()))
```

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -972,6 +979,7 @@ def compute_symbols(self):
         return shape_data
 
     def update_shapes(self):
+        self.master.prepare_data()
```
nikicc: Could we also add something like `if self.attr_shape not in self.data.domain`?

Orange/widgets/visualize/tests/test_owscatterplot.py
```diff
@@ -268,6 +269,23 @@ def test_set_strings_settings(self):
...
+        GH-2152
+        GH-2157
+        """
+        table = Table("iris")
+        table.X = sp.csr_matrix(table.X)
+        self.assertTrue(sp.issparse(table.X))
+        table.Y = sp.csr_matrix(table._Y)  # pylint: disable=protected-access
+        self.assertTrue(sp.issparse(table.Y))
+        self.send_signal("Data", table)
+        self.widget.set_subset_data(table[0:30])
```
nikicc: This can be just `table[:30]`

Orange/widgets/visualize/owscatterplot.py
jerneju: Probably not.

jerneju:
Fixed.

PR #2172, opened 2017-03-31
[FIX] KMeans: should not crash when there is less data rows than k

jerneju:
##### Issue
When there is less rows in a data than selected (from - to, second option) the widget crashes. 
https://sentry.io/biolab/orange3/issues/235248479/

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owkmeans.py
+++ b/Orange/widgets/unsupervised/owkmeans.py
@@ -209,12 +209,12 @@ class OWKMeans(widget.OWWidget):
         try:
             self.controlArea.setDisabled(True)
             if not self.check_data_size(self.k_from, self.Error):
-                return
+                return False
             self.check_data_size(self.k_to, self.Warning)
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/235248479/

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -243,9 +244,9 @@ def apply(self):
         self.clear_messages()
         if self.data is not None:
             if self.optimize_k:
-                self.run_optimization()
-                self.mainArea.show()
-                self.update_results()
+                if self.run_optimization():
```
janezd: Say that `mainArea` is hidden. If optimization fails, main area won't be shown and updated. This is OK.

What if `mainArea` is currently shown and contains some data? If optimization fails, it won't be hidden, but it won't update either...

PR #2174, opened 2017-03-31
WidgetOutputsTestMixin: Make it work for OWFreeViz

VesnaT:
##### Issue

The mixin could not be used on OWFreeViz unittest.

##### Description of changes
Add an additional commit for OWFreeViz

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -631,10 +631,12 @@ class WidgetOutputsTestMixin:
     def test_outputs(self):
         self.send_signal(self.signal_name, self.signal_data)
 
-        # only needed in TestOWMDS
+        # only needed in widgets with event loop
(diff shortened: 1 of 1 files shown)


janezd:
Can you change the comment `# only needed in TestOWMDS` to `# only needed in widgets that <insert reason here>`?

Is there a way to avoid these special cases without changing the functionality of widgets? If there are other such widgets --- if there are many such widgets, should we add a class attribute (something like) `COMMIT_BEFORE_TEST` that is `False` by default, such widgets would set it to `True`?

VesnaT:
I think we should leave it as it is and not change the widgets to pass tests, since there are only two of them.

astaric:
Will be fixed in a different way (in prototypes add-on).

PR #2176, opened 2017-03-31
[FIX] Select Rows and Table: Filtering string values

jerneju:
##### Issue
Table throws error when trying to compare string values. A string value written in Select Rows widgets and a numerical value which is set as string.
https://sentry.io/biolab/orange3/issues/244859655/

Put _File_ and _Select Rows_ widgets and connect them. Then open _iris_. Set one of four features to _string_ and _meta_. Then choose that feature in _Select Rows_ and choose filter "_is before_" or some other one.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1199,6 +1199,7 @@ class Table(MutableSequence, Storage):
                 sel = np.vectorize(f)(col)
             elif isinstance(f, (data_filter.FilterContinuous,
                                 data_filter.FilterString)):
+                col = col.astype(str) if isinstance(f, data_filter.FilterString) else col
                 if (isinstance(f, data_filter.FilterString) and
                         not f.case_sensitive):
                     # noinspection PyTypeChecker
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1133,10 +1133,26 @@ class Table(MutableSequence, Storage):
             sel = np.logical_not(sel)
         return self.from_table_rows(self, sel)
 
-    def _filter_values_indicators(self, filter):
-        from Orange.data import filter as data_filter
+    def _filter_values(self, filter):
+        selection = self._filter_values_to_indicator(filter)
+        return self.from_table(self.domain, self, selection)
 
-        if isinstance(filter, data_filter.Values):
+    def _filter_values_to_indicator(self, filter):
+        """Return selection of rows matching the filter conditions
+
+        Handles conjunction/disjunction and negate modifiers
+
+        Parameters
+        ----------
+        filter: Values object containing the conditions
+
+        Returns
+        -------
+        A 1d bool array. len(result) == len(self)
+        """
+        from Orange.data.filter import Values
+
+        if isinstance(filter, Values):
             conditions = filter.conditions
             conjunction = filter.conjunction
         else:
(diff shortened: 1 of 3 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/244859655/

Orange/data/table.py
```diff
@@ -1202,6 +1205,7 @@ def _filter_values_indicators(self, filter):
                 sel = np.vectorize(f)(col)
             elif isinstance(f, (data_filter.FilterContinuous,
                                 data_filter.FilterString)):
+                col = get_str_column(col, isinstance(f, data_filter.FilterString))
```
janezd: Why calling a (trivial) function instead of doing it right here? It further obscures the logic of this code, which is already too contrived.

jerneju: Radon and cyclomatic complexity.

janezd: We measure cyclomatic complexity to make the code more, not less readable. :)

Find another way to decrease the complexity. Cheating with dictionaries typically makes the code slower or more complex, but in this case dictionaries are natural and fast. You can prepare a dictionary (as a class attribute, not a local variable)

```
    operators = {
        f.Equal: operator.eq,
        f.NotEqual: operator.neq,
        f.Less: operator.le,
        ...
    }
```

Then you can replace most if-s with a single `col = ops[f.oper](col, fmin)`.

Perhaps this does not work for some reason. In this case, find another solution to simplify the function. Extracting individual ifs to local functions is not a proper solution to high cyclomatic complexity (except when it makes the code actually better).

You can also extract the code the is within the for loop into a separate function. This would make sense since this is the code that applies the filter to a selection. If this is still not enough, make a separate function for each filter type (elifs in this block of code), and extract the function that prepares the initial `sel` column.

Orange/data/table.py
```diff
@@ -1154,95 +1247,7 @@ def _filter_values_indicators(self, filter):
...
-                                      dtype=bool)
-                else:
-                    raise TypeError("Invalid operator")
-                if conjunction:
-                    sel *= col
-                else:
-                    sel += col
-            else:
-                raise TypeError("Invalid filter")
+            sel = self._filter_by_condition(f, conjunction, sel, data_filter)
```
janezd: This is better than before, but not by much. If there's a simple and obvious a way to make `_filter_by_condition` more legible, do so. (I confess: that's my old code.) If there's not, forget it.

Orange/data/table.py
```diff
@@ -1133,6 +1133,99 @@ def _filter_same_value(self, column, value, negate=False):
             sel = np.logical_not(sel)
         return self.from_table_rows(self, sel)
 
+    def _filter_by_condition(self, f, conjunction, sel, data_filter):  # pragma: no cover
```
astaric: Please stop abusing pragma: no cover.

This method should be tested, along with every other line of code in the table.py module. If codecov complains that the code you have moved around (but not changed) is not covered, make sure that the refactoring is done in a separate commit (so that the change can be inspected separately) and state that in the comment. Or write some tests. Basically any other way to approach it is better than marking actual code with "no cover".

jerneju:
https://sentry.io/biolab/orange3/issues/260544377/

jerneju:
@astaric: Added a description how to reproduce the problem.

astaric:
Thank you.

I have refactored the test a bit and split the _filter_values_indicator into multiple smaller functions (undoing the 383bab2 refactoring commit in the process).

The old _filter_values_indicators tried to handle both conjunctions/disjunctions and specific filters at the same time. I have split that code into two functions. Furthermore, evaluation of the most complicated filters (Discrete, Continuous, String) has been extracted to separate methods.

astaric:
@janezd can you take a look at the new refactoring? I am almost sad that this is going to be thrown away once we move to pandas :)

kernc:
> I am almost sad that this is going to be thrown away once we move to pandas

FWIW I totally don't mind if you do the throwing. :smile:

astaric:
A restart worked as well :)

Issue #2156, opened 2017-03-30
Transpose should work with sparse


ajdapretnar:
##### Orange version

3.5.dev

##### Expected behavior
Transpose should work with sparse.

##### Actual behavior
Transpose fails on sparse.

##### Steps to reproduce the behavior
Connect Corpus to Transpose and Transpose to Data Table.

PR #2177, opened 2017-03-31
[FIX] stats: Open Corpus in OWDataTable after transposing it

VesnaT:
##### Issue

Fixes: #2156

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -210,7 +210,7 @@ def stats(X, weights=None, compute_variance=False):
             np.nanvar(X, axis=0) if compute_variance else np.zeros(X.shape[1]),
             nans,
             X.shape[0] - nans))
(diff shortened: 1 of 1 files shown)


Orange/statistics/util.py
```diff
@@ -193,24 +193,23 @@ def stats(X, weights=None, compute_variance=False):
     is_sparse = sp.issparse(X)
     weighted = weights is not None and X.dtype != object
 
-    if weighted:
-        weights = np.c_[weights] / sum(weights)
```
janezd: I'd still compute `weights` instead of doing it in two places later. Just a suggestion, feel free to ignore.

VesnaT: radon...

PR #2180, opened 2017-03-31
OWManifoldLearning: Show error msg when data is sparse

### LINKED ISSUE #2158: Manifold Learning should work on sparse
##### Orange version

3.5.dev

##### Expected behavior
Manifold Learning works on sparse.

##### Actual behavior
Manifold Learning fails on sparse.

Corpus - Bag of Words - Manifold Learning.

VesnaT:
##### Issue

Related to #2158.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmanifoldlearning.py
+++ b/Orange/widgets/unsupervised/owmanifoldlearning.py
@@ -179,6 +179,7 @@ class OWManifoldLearning(OWWidget):
     class Error(OWWidget.Error):
         n_neighbors_too_small = Msg("Neighbors must be greater than {}.")
         manifold_error = Msg("{}")
+        sparse_not_supported = Msg("Sparse data is not supported.")
 
     def __init__(self):
         self.data = None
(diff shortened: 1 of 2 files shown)


ajdapretnar:
😱 Can this not work on sparse? 😢

Orange/widgets/unsupervised/owmanifoldlearning.py
```diff
@@ -246,8 +247,11 @@ def apply(self):
                 projector = method(**self.get_method_parameters())
                 try:
                     self.progressBarSet(20)
-                    X = projector(self.data).embedding_
-                    data = Table(domain, X, self.data.Y, self.data.metas)
+                    if self.data.is_sparse():
```
nikicc: Could we move this check into `set_data` and not even call apply?
Currently if I click around in the widget progress bar starts, gets set to 20% and then we show the error.

nikicc:
@ajdapretnar I'm afraid not without some major changes. I suggest we accept this since it's better than the current state. But we should probably keep the issue open until someone check exactly which methods can work on sparse and adopts the widget accordingly.

Issue #2148, opened 2017-03-28
Predictions: Should predict a missing class (Predict on test data)


Pelonza:
##### Orange version

3.4.*

##### Expected behavior

For most machine learning applications, you want to be able to use a trained predictor/classifier or other algorithm to NEW data, which doesn't have an outcome already determined.

##### Actual behavior

Right now, to use the "predict" widget, the documentation recommends taking a random sample of the input data, then training the models on one sub-set, and predicting the other set.

##### Steps to reproduce the behavior

Follow your example model to produce the regular behavior...
I was trying to do the Titanic Kaggle.com challenge, where you build a classifier on training data, then predict survive/die on NEW test data... this doesn't work with importing Kaggle.com's base data-sets. I've been struggling to find a way to make it work at all....

##### Additional info (worksheets, data, screenshots, ...)

I think the key here would be to add functionality to the "predict" module that detects if there is new data missing the target column (as defined in the trained classifiers/models). Then, if that column is missing, attach a "blank" column with the same attributes and populate it with the outcome from the trained input models (similar to what's going on already with data that already HAS a target column, but being a bit smarter about detecting it missing).

janezd:
We thought we fixed this bug already. This is critical.

jerneju:
It works in 3.3.12 but not in 3.4.0.

janezd:
Is this related to merging of classifiers and regressors? @pavlin-policar?

pavlin-policar:
I really don't think so... My changes don't touch the models at all, or in how they are created, which is where this seems to fail. My changes don't even really touch the "learning" process much.

pavlin-policar:
I've isolated the the issue to a commit, `e977e8b` (#2026). The issue here, this should be helpful.

I've reproduced the error by splitting a dataset, then removing the target class column from the test set, then sending a trained model to and the test data to the Predictions widget.

janezd:
Ooops, that was my fault! 😟 

Thanks, @pavlin-policar. Now I have a task for Friday.

Pelonza:
@pavlin-policar I wonder if this is al

PR #2183, opened 2017-03-31
[FIX] OWPredictions: Allow classification when data has no target column

janezd:
##### Issue

`OWPredictions` reported mismatching targets if the data had not target column (#2148).

This occurred because the data's `class_var` was `None`, and it was thus not the same as the target of predictors.

##### Description of changes

The widget's `self.class_var` is now set from the target of the models. All models must have the same target. Also, if the data has a target, it has to be the same as the models' target.

This required several other changes in the widget to cope with situations that could not occur previously.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -56,7 +56,12 @@ class OWPredictions(OWWidget):
         empty_data = Msg("Empty data set")
 
     class Error(OWWidget.Error):
-        predictor_failed = Msg("One or more predictors failed (see more...)\n{}")
+        predictor_failed = \
+            Msg("One or more predictors failed (see more...)\n{}")
+        predictors_target_mismatch = \
+            Msg("Predictors do not have the same target.")
+        data_target_mismatch = \
+            Msg("Data does not have the same target as predictors.")
 
     settingsHandler = settings.ClassValuesContextHandler()
     #: Display the full input dataset or only the target variable columns (if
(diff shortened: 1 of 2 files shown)


ajdapretnar:
@lanzagar Could you please just check the code as well and merge. I only tested it on cases and it works.

Orange/widgets/evaluate/owpredictions.py
```diff
@@ -285,12 +300,16 @@ def _invalidate_predictions(self):
...
+                self.data is not None:
+            return [p for p in self.predictors.values()
+                    if p.results is not None and not isinstance(p.results, str)]
+        else:
+            return []
 
     def _update_predictions_model(self):
         """Update the prediction view model."""
-        if self.data is not None:
+        if self.data is not None and self.class_var is not None:
```
lanzagar: Doesn't the second part (class_var not None) also imply the first (data not None).

(The same `if` appears in a couple of places)

Orange/widgets/evaluate/owpredictions.py
```diff
@@ -442,15 +461,15 @@ def _commit_predictions(self):
...
+        if self.class_var and self.class_var.is_discrete:
             newmetas, newcolumns = self._classification_output_columns()
         else:
             newmetas, newcolumns = self._regression_output_columns()
 
         attrs = list(self.data.domain.attributes) if self.output_attrs else []
         metas = list(self.data.domain.metas) + newmetas
-        domain = Orange.data.Domain(attrs, class_var, metas=metas)
+        domain = \
+            Orange.data.Domain(attrs, self.data.domain.class_var, metas=metas)
```
lanzagar: If we assure at the beginning that `self.data.domain.class_var` must be equal to `self.class_var`, then the second (shorter/direct) version would look nicer here. And the line break would probably not be necessary :)

Orange/widgets/evaluate/owpredictions.py
janezd: `self.class_var` comes from predictors and can be `not None` even if no data is present. See https://github.com/janezd/orange3/blob/21ba4679bf19efde4498363a56c69bf3798ea0f9/Orange/widgets/evaluate/owpredictions.py#L224.

Orange/widgets/evaluate/owpredictions.py
janezd: Now I recalled: I did have `self.class_var` and it fit into a single line. But this fails exactly in the case that this PR is fixing: if `self.data.domain.class_var` is `None`, the output data should have no class, either. Changing this to `self.class_var` can add a column of unknown values if the original data has no class.

PR #2194, opened 2017-04-05
[FIX] Test & Score: crashing prevented when learner disconnects

jerneju:
##### Issue
When Test & Score performs evaluation sb might disconnect connection from some learner. Widget then crashes.
https://sentry.io/biolab/orange3/issues/227231658/

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -466,6 +466,15 @@ class OWTestLearners(OWWidget):
             else:
                 self.error()
 
+        self.puts_results(learners, results, class_var)
+
+        self.setStatusMessage("")
+
+    def puts_results(self, learners, results, class_var):
+        """
+        Called by _update_results. This method prepares calculated results and
+        put them into self.learners.
+        """
         learner_key = {slot.learner: key for key, slot in self.learners.items()}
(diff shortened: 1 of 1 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/227231658/

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -381,6 +381,29 @@ def shuffle_split_changed(self):
     def _param_changed(self):
         self._invalidate()
 
+    def push_results(self, learners, results, class_var):
```
astaric: I would put this method below _update_results and pick another name. A docstring might be a good idea as well, as it is not that obvious what the method does.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -466,6 +466,15 @@ def update_progress(finished):
...
                 self.error()
 
+        self.fill_results(learners, results, class_var)
+
+        self.setStatusMessage("")
+
+    def fill_results(self, learners, results, class_var):
+        """
+        Called by _update_results. This method prepares calculated results and
+        put them into self.learners.
```
lanzagar: puts

Issue #2165, opened 2017-03-30
Nomogram fails on sparse


ajdapretnar:
##### Orange version

3.5.dev

##### Expected behavior
Nomogram works on sparse.

##### Actual behavior
Nomogram fails on sparse.

##### Steps to reproduce the behavior
Corpus - Bag of Words - Nomogram (with Logistic Regression)

##### Additional info (worksheets, data, screenshots, ...)
<img width="446" alt="screen shot 2017-03-30 at 11 28 27" src="https://cloud.githubusercontent.com/assets/12524972/24497530/06a5770c-153c-11e7-8384-12ef74699ac5.png">

nikicc:
Maybe the problem is only in SharedComputeValues https://github.com/biolab/orange3-text/issues/224?

nikicc:
Fixed with https://github.com/biolab/orange3/pull/2197.

PR #2197, opened 2017-04-06
[ENH] Nomogram: Support for sparse data

nikicc:
##### Issue

#2165

##### Description of changes

* Statistics.utils: `nanmin`, `nanmax`, `average`, `unique` equivalents of numpy's that support sparse or dense matrices.
* SharedComputeValue: add variable attribute.
* Support nomogram on sparse data.
* Some speedups for nomogram. Time measured on a small BoW data set of shape (140, 3000):
  *  `reconstruct_domain` method: 3.0s -> 0.03 s
  * `calculate_log_reg_coefficients` method: TLDW (minutes+)  -> 1.5 s

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -231,3 +231,81 @@ def stats(X, weights=None, compute_variance=False):
             np.zeros(X.shape[1]),
             nans,
             X.shape[0] - nans))
+
+
+def _sparse_has_zeros(x):
+    """ Check if sparse matrix contains any implicit zeros. """
+    return np.prod(x.shape) != x.nnz
+
+
+def _nan_min_max(x, func, axis=0):
+    if not sp.issparse(x):
+        return func(x, axis=axis)
+    if axis is None:
+        extreme = func(x.data, axis=axis) if x.nnz else float('nan')
+        if _sparse_has_zeros(x):
+            extreme = func([0, extreme])
+        return extreme
+    if axis == 0:
+        x = x.T
+    else:
+        assert axis == 1
+
+    # TODO check & transform to correct format
+    r = []
+    for row in x:
+        values = row.data
+        extreme = func(values) if values.size else float('nan')
+        if _sparse_has_zeros(row):
+            extreme = func([0, extreme])
+        r.append(extreme)
+    return np.array(r)
+
+
+def nanmin(x, axis=None):
+    """ Equivalent of np.nammin that supports sparse or dense matrices. """
+    return _nan_min_max(x, np.nanmin, axis)
+
+
+def nanmax(x, axis=None):
+    """ Equivalent of np.nammax that supports sparse or dense matrices. """
+    return _nan_min_max(x, np.nanmax, axis)
+
+
+def mean(x):
+    """ Equivalent of np.mean that supports sparse or dense matrices. """
+    if not sp.issparse(x):
+        return np.mean(x)
+
+    n_values = np.prod(x.shape)
+    return np.sum(x.data) / n_values
+
+
+def nanmean(x):
+    """ Equivalent of np.nanmean that supports sparse or dense matrices. """
+    if not sp.issparse(x):
+        return np.nanmean(x)
+
+    n_values = np.prod(x.shape) - np.sum(np.isnan(x.data))
+    return np.nansum(x.data) / n_values
+
+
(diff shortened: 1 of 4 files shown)


Orange/statistics/util.py
```diff
@@ -231,3 +231,70 @@ def weighted_mean():
...
+            return np.nansum(x.data) / n_values
+
+
+def unique(x, return_counts=True):
+    """ Equivalent of np.unique that supports sparse or dense matrices. """
+    if not sp.issparse(x):
+        return np.unique(x, return_counts=return_counts)
+    else:
+        n_zeros = np.prod(x.shape) - x.data.size
+        r = np.unique(x.data, return_counts=return_counts)
```
lanzagar: `x.data` can contain explicit zeros right? E.g. make a csr matrix and set a non-zero element to 0.
In this case you need to be careful about inserting another 0 below...

Orange/statistics/util.py
```diff
@@ -231,3 +231,70 @@ def weighted_mean():
...
+        else:
+            n_values = np.prod(x.shape) - n_nans
+            return np.nansum(x.data) / n_values
+
+
+def unique(x, return_counts=True):
+    """ Equivalent of np.unique that supports sparse or dense matrices. """
+    if not sp.issparse(x):
+        return np.unique(x, return_counts=return_counts)
+    else:
```
lanzagar: `else` is unnecessary here and in other functions, which first check if x is not sparse and return something.
It just adds an extra indentation to all of the actual function body.

Orange/statistics/util.py
```diff
@@ -231,3 +231,70 @@ def weighted_mean():
             np.zeros(X.shape[1]),
             nans,
             X.shape[0] - nans))
+
+
+def _sparse_has_zeros(x):
+    """ Check if sparse matrix contains any implicit zeros. """
+    return np.prod(x.shape) != x.data.size
```
lanzagar: It is probably better to use `x.nnz` instead of `x.data.size` everywhere.
Looks like the `spmatrix` base class has `nnz` so every type should have it, while e.g. `dok_matrix` does not have `.data`

nikicc: Corrected. Though, methods still won't work for for `dox_matrix` since we rely on `x.data` elsewhere.

Orange/statistics/util.py
```diff
@@ -231,3 +231,70 @@ def weighted_mean():
...
+        extreme = func(x.data, axis=axis) if x.data.size else float('nan')
+        if _sparse_has_zeros(x):
+            extreme = func([0, extreme])
+        return extreme
+    else:
+        r = []
+        if axis == 0:
+            x = x.T
+
+        # TODO check & transform to correct format
```
nikicc: X is usually in `csr` and hence when one calls this with `axis=0` x becomes `csc` (due to transposing), which is isn't efficient for row slicing.

Orange/statistics/util.py
```diff
@@ -231,3 +231,70 @@ def weighted_mean():
...
             nans,
             X.shape[0] - nans))
+
+
+def _sparse_has_zeros(x):
+    """ Check if sparse matrix contains any implicit zeros. """
+    return np.prod(x.shape) != x.data.size
+
+
+def _nan_min_max(x, axis=0, func=None):
```
nikicc: Fixed.

Orange/statistics/util.py
```diff
@@ -231,3 +231,70 @@ def weighted_mean():
...
+def average(x):
+    """ Equivalent of np.average that supports sparse or dense matrices. """
+    if not sp.issparse(x):
+        return np.average(x)
+    else:
+        n_nans = np.sum(np.isnan(x.data))
+        if n_nans:
+            return float('nan')
+        else:
+            n_values = np.prod(x.shape) - n_nans
```
nikicc: It is 🙈

Orange/statistics/util.py
nikicc: Fixed.

Orange/statistics/util.py
nikicc: Fixed.

nikicc:
@lanzagar I think all issues are addressed now. Please, check again.

Issue #2187, opened 2017-04-04
Scatter Plot: left (axis y) margin is not adapting


jerneju:
##### Orange version
master (3.5dev)

##### Expected behavior
Left axis margin should adapt to a new maximum label length. It does not adapt even if there is a new data send to the widget.

##### Actual behavior
Firstly:
![screenshot_20170404_100658](https://cloud.githubusercontent.com/assets/22157215/24647195/349497b0-191f-11e7-87c8-1653ff3ed49b.png)

And lastly:
![screenshot_20170404_100747](https://cloud.githubusercontent.com/assets/22157215/24647194/348fe0b2-191f-11e7-95c7-d41bbe7a7b7d.png)

##### Steps to reproduce the behavior
See above.

ajdapretnar:
Can confirm the second reported bug. Happened to me yesterday.

jerneju:
Fixed #2200.

PR #2200, opened 2017-04-07
[FIX] Scatter Plot: left margin (axis y) is now adapting

jerneju:
##### Issue
See #2187

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
Does not fix the bug for me...

PR #2205, opened 2017-04-07
[ENH] Add all available methods to feature selection preprocessor

lopezco:
##### Issue
Add missing feature selection methods to OWPreprocess

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpreprocess.py
+++ b/Orange/widgets/data/owpreprocess.py
@@ -471,7 +471,11 @@ class FeatureSelectEditor(BaseEditor):
         ("Gain ratio", preprocess.score.GainRatio),
         ("Gini index", preprocess.score.Gini),
         ("ReliefF", preprocess.score.ReliefF),
-        ("Fast Correlation Based Filter", preprocess.score.FCBF)
+        ("Fast Correlation Based Filter", preprocess.score.FCBF),
+        ("ANOVA", preprocess.score.ANOVA),
+        ("Chi2", preprocess.score.Chi2),
+        ("RReliefF", preprocess.score.RReliefF),
+        ("Univariate Linear Regression", preprocess.score.UnivariateLinearRegression)
     ]
 
     def __init__(self, parent=None):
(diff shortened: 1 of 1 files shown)


astaric:
@janezd looks good to me. What do you think?

Orange/widgets/data/owpreprocess.py
```diff
@@ -488,7 +492,11 @@ def __init__(self, parent=None):
              {"text": "Gain Ratio"},
              {"text": "Gini Index"},
              {"text": "ReliefF"},
-             {"text": "Fast Correlation Based Filter"}
+             {"text": "Fast Correlation Based Filter"},
+             {"text": "ANOVA"},
+             {"text": "Chi2"},
+             {"text": "RReliefF"},
+             {"text": "UnivariateLinearRegression"}
```
janezd: Use spaces: this is shown in a combo.

janezd:
I'm not opposed to merging, but I'd be much happier if error messages were improved.

Loading heart_disease and selecting Univariate linear regression gives an error "Scoring method WrapperMeta requires a class variable of type ContinuousVariable". The end-user should be given messages from the scripting-level exceptions only as a last resort. Here, the user doesn't know what is ContinuousVariable and, especially, what is WrapperMeta. This error is not much better than a crash.

To be fair, the existing scorers also refer to DiscreteVariable. But at least they don't have WrapperMeta. The problem is that the additional scorers seem less robust, so the user get more cryptic messages. Running, for instance, ANOVA on heart_disease gives "Only ContinuousVariables are supported", without explaining that this holds only for ANOVA.

The question (@astaric, @BlazZupan) is whether we allow this widget to be somewhat unfriendly (knowing that this PR is not the cause, it just increases it by some amount) or do we, for instance, add checks before calling feature selectors and issue friendly error messages instead of showing exception messages.

astaric:
@janezd, Cryptic error messages should also be changed in the scripting part, which is IMHO not a part of this PR. (Hearing about WrapperMeta is never useful)

I'll go ahead and merge this and we can improve error messages as a part of another PR.

janezd:
Agreed. I'm assigning this to myself.

PR #2208, opened 2017-04-07
Remove deprecation warning

lopezco:
Remove deprecation warning when creating feature selection method's instances

##### Description of changes
Create an instance of the Class without data and then call the the instance with the data.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/fss.py
+++ b/Orange/preprocess/fss.py
@@ -50,9 +50,8 @@ class SelectBestFeatures(Reprable):
         self.decreasing = decreasing
 
     def __call__(self, data):
-        method = self.method
         # select default method according to the provided data
-        if method is None:
+        if self.method is None:
             autoMethod = True
             discr_ratio = (sum(a.is_discrete
                                for a in data.domain.attributes)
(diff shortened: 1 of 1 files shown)


janezd:
This breaks compatibility with the existing code.

The original function accepted either an instance of a scorer (e.g. `ANOVA()`) or a type (`ANOVA`). The latter case was handled by the constructor provided by `_RefuseDataInConstructor`.

This PR changes the function's signature so that is accepts only types and not instances. I think we should do the opposite; scorers (e.g. ReliefF) may have constructors with meaningful arguments, while the PR calls `model` with default arguments.

The proper solution is thus to not derive scorers from `_RefuseDataInConstructor` and fix any code that calls scorers with types instead of with instances.

I'm not closing this PR yet; @astaric and @kernc, please take a look, too, and decide.

kernc:
Indeed. The tests and docs have been adapted to avoid these deprecation warnings, but the use in code remained. I guess a quick way to remove the warning is to create a scorer _instance_ in [_owpreprocess.py_, i.e. suffix line 508](https://github.com/kernc/orange3/blob/f5e4a3945182cc6b13b53b6bd853124b499af787/Orange/widgets/data/owpreprocess.py#L508) with `()`.

PR #2213, opened 2017-04-10
[FIX] Hierarchical Clustering: Enumeration - metas

jerneju:
##### Description of changes
First Meta set as default if no class variable

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
@jerneju Thank you very much for this quick solution! It is not, however, what we should put in the master (regarding the defaults).
Closed in favour of #2214.
Thanks nevertheless! :)

PR #2217, opened 2017-04-11
[ENH] Reimplement macOS application (.app) build scripts

ales-erjavec:
##### Issue

Update macOS application to use a newer Python (3.6) and Qt (5.8)

##### Description of changes

Replace the old build scripts. There is no prebuild bundle template anymore. The python framework is downloaded/extracted from the official python.org macOS pkg installers, laid out into the .app directory and made relocatable (using install_name_tool to change .dylib' library load paths relative).

All other dependencies are then installed using pip/wheel

However with Python 3.6 we are now responsible for distributing a ssl certificate store.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


astaric:
When I build an app using this script and run it, Orange menu lists 
- About __main__.py
- Hide __main__.py

Would it be possible to rename those entries?

ales-erjavec:
> Would it be possible to rename those entries?

Should be fixed now.

astaric:
Looks ok to me. Can I merge it or will it break the daily builds?

janezd:
I have one more question: can we decrease the bundle size? A year or two ago I noticed that we distribute a lot of stuff, perhaps even QtDesigner, in the bundle.

ales-erjavec:
> Can I merge it or will it break the daily builds?

The daily builds are already broken, so this cannot hurt much.

> can we decrease the bundle size?

I do not think so. This one is about 20MB lighter then before (122MB).

PR #2221, opened 2017-04-11
owlegend: avoid computation in sizeHint

astaric:
PyQt5 tests have been segfaulting occasionally. The culprit was sizeHint method, which crased from some reason. I cannot reproduce the exact problem, but I can move the code that computes the sizehint to __init__, where it will not segfault if it crashes.

### DIFF
--- a/Orange/widgets/visualize/utils/owlegend.py
+++ b/Orange/widgets/visualize/utils/owlegend.py
@@ -223,6 +223,7 @@ class LegendItemSquare(ColorIndicator):
     """
 
     SIZE = QSizeF(12, 12)
+    _size_hint = SIZE
 
     def __init__(self, color, parent):
         super().__init__(parent)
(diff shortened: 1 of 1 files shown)


Orange/widgets/visualize/utils/owlegend.py
```diff
@@ -232,9 +233,10 @@ def __init__(self, color, parent):
         self.__square.setBrush(QBrush(color))
         self.__square.setPen(QPen(QColor(0, 0, 0, 0)))
         self.__square.setParentItem(self)
```
kernc: Might also work to call `super().__init__()` _after_ `self.__square` (or equivalent below) is defined.

astaric: pyqt tends to disagree.

RuntimeError: super-class __init__() of type LegendItemSquare was never called

kernc: Sorry, after `self.__square` is defined, but _before_ `self` is set as its parent.

PR #2224, opened 2017-04-12
[FIX] Tree Viewer: check if there is selected class value

jerneju:
##### Issue
Tree Viewer crashes when Target Class combo selected index number is higher than actual number of class values.
https://sentry.io/biolab/orange3/issues/205702715/

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owtreeviewer.py
+++ b/Orange/widgets/visualize/owtreeviewer.py
@@ -263,6 +263,7 @@ class OWTreeGraph(OWTreeViewer2D):
         self.color_combo.clear()
         self.closeContext()
         self.model = model
+        self.target_class_index = 0
         if model is None:
             self.info.setText('No tree.')
             self.root_node = None
@@ -440,7 +441,7 @@ class OWTreeGraph(OWTreeViewer2D):
         return TreeAdapter(model)
 
 
-def test():
+def main():
     """Standalone test"""
     import sys
     from AnyQt.QtWidgets import QApplication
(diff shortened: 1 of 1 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/205702715/

janezd:
This is a patch, not a fix. :)

The combo should not have had an invalid value in the first place. I guess you should discover why this occurred and fix the problem there.

jerneju:
Now I fixed it in an another way.

lanzagar:
Does not fix the problem for me.
Also, there are now two consecutive calls to openContext.

janezd:
> Now I fixed it in another way.

Where is this other fix? Did you forget to push?

jerneju:
> Where is this other fix? Did you forget to push?
Changed it back.

janezd:
I don't get it. :) What is the state of this PR now? Should we just close it since it doesn't fix the problem in the right way and it doesn't contain any code that would be useful in the actual fix?

If we merge this as a "temporary fix", we nobody will remember to remove it after the actual fix  ... and we have too much of such dead code already. :)

janezd:
I thought that you would have to call `self.color_combo.setCurrentIndex(self.target_class_index)` after `openContext`, but `setup_scene` calls `activate_loaded_settings` which in turn calls the combo's `setCurrentIndex`. Knowing this, even the existing call `self.color_combo.setCurrentIndex(self.target_class_index)` is actually redundant ... but leave it as it is.

PR #2225, opened 2017-04-12
[FIX] Report: handle PermissionError when trying to save

jerneju:
##### Issue
https://sentry.io/biolab/orange3/issues/251030241/

##### Description of changes
Now shows a windows with error message and logs error as well.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/report/owreport.py
+++ b/Orange/canvas/report/owreport.py
@@ -1,4 +1,5 @@
 import os
+import logging
 
 import pkg_resources
 import pickle
@@ -21,6 +22,7 @@ from Orange.widgets.utils.webview import WebviewWidget
 from Orange.canvas.application.canvasmain import CanvasMainWindow
 from Orange.canvas.gui.utils import message_critical
 
+log = logging.getLogger(__name__)
 
 class Column(IntEnum):
     item = 0
@@ -310,8 +312,11 @@ class OWReport(OWWidget):
             self.save(filename)
         else:
             def save_html(contents):
-                with open(filename, "w", encoding="utf-8") as f:
-                    f.write(contents)
+                try:
+                    with open(filename, "w", encoding="utf-8") as f:
+                        f.write(contents)
+                except PermissionError:
+                    self.permission_error(filename)
 
             save_html(self.report_view.html())
         self.report_changed = False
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/251030241/

Orange/canvas/report/owreport.py
```diff
@@ -337,14 +340,14 @@ def open_report(self):
 
         try:
             report = self.load(filename)
-        except (IOError, AttributeError) as e:
+        except (IOError, AttributeError):
```
janezd: I tried to open a file with extension .report, but which was not actually a report file. It crashes with `UnpicklingError`. Perhaps you should add it to list, or even catch the general `Exception` here.

Orange/canvas/report/owreport.py
```diff
@@ -337,14 +340,14 @@ def open_report(self):
...
             message_critical(
-                 self.tr("Could not load an Orange Report file"),
-                 title=self.tr("Error"),
-                 informative_text=self.tr("An unexpected error occurred "
-                                          "while loading '%s'.") % filename,
-                 exc_info=True,
-                 parent=self)
+                self.tr("Could not load an Orange Report file"),
+                title=self.tr("Error"),
+                informative_text=self.tr("An unexpected error occurred "
```
janezd: Remove *unexpected* (all errors are unexpected) and probably also the article, hence "Error occurred while loading '%s'.

Orange/canvas/report/owreport.py
```diff
@@ -337,14 +340,14 @@ def open_report(self):
...
+                                         "while loading '%s'.") % filename,
```
janezd: Use formatting with `{}` and `format`, not the C-style `%`. I would say the the old-style is permitted in Orange only for consistency in the modules that (still) use the old style.

Orange/canvas/report/owreport.py
```diff
@@ -162,13 +162,13 @@ def _setup_ui_(self):
 
         class PyBridge(QObject):
             @pyqtSlot(str)
-            def _select_item(myself, item_id):
+            def _select_item(self, item_id):
```
jerneju: Shame on me. I blindly followed lint's complaints.

PR #2228, opened 2017-04-12
[FIX] Manifold Learning: handling numpy LinAlgError

jerneju:
##### Issue
https://sentry.io/biolab/orange3/issues/210202559/

##### Description of changes
Sometimes singular matrix is calculated from input data and that is now handled.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmanifoldlearning.py
+++ b/Orange/widgets/unsupervised/owmanifoldlearning.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from AnyQt.QtWidgets import QWidget, QVBoxLayout
 from AnyQt.QtCore import Qt
 
@@ -260,6 +262,8 @@ class OWManifoldLearning(OWWidget):
                             self.Error.n_neighbors_too_small("{}".format(n))
                         else:
                             self.Error.manifold_error(e.args[0])
+                    except np.linalg.linalg.LinAlgError as e:
+                        self.Error.manifold_error(str(e))
         self.send("Transformed data", data)
 
     def get_method_parameters(self):
--- a/Orange/widgets/unsupervised/tests/test_owmanifoldlearning.py
+++ b/Orange/widgets/unsupervised/tests/test_owmanifoldlearning.py
@@ -3,7 +3,7 @@
 import numpy as np
 from scipy import sparse
 
-from Orange.data import Table
+from Orange.data import Table, Domain, ContinuousVariable, DiscreteVariable
 from Orange.widgets.unsupervised.owmanifoldlearning import OWManifoldLearning
 from Orange.widgets.tests.base import WidgetTest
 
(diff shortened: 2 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/210202559/

Orange/widgets/unsupervised/owmanifoldlearning.py
```diff
@@ -260,6 +262,8 @@ def apply(self):
                             self.Error.n_neighbors_too_small("{}".format(n))
                         else:
                             self.Error.manifold_error(e.args[0])
+                    except np.linalg.linalg.LinAlgError as e:
+                        self.Error.manifold_error(e.args[0])
```
janezd: Use `str(e)` instead of `e.args[0]`.
 Calling `str()` on an instance of `BaseException` will give [something at least as good as `e.args[0]](https://docs.python.org/3/library/exceptions.html#BaseException). 

Actually, even `self.Error.manifold_error(e)` but I'd prefer being explicit.

Orange/widgets/unsupervised/tests/test_owmanifoldlearning.py
```diff
@@ -73,3 +73,25 @@ def test_sparse_data(self):
         self.send_signal("Data", None)
         self.widget.apply_button.button.click()
         self.assertFalse(self.widget.Error.sparse_not_supported.is_shown())
+
+    def test_data_which_makes_singular(self):
+        """
+        Sometimes singular matrix is calculated from input data and
+        that should be handled.
```
janezd: Rephrase: Handle singular matrices

Orange/widgets/unsupervised/tests/test_owmanifoldlearning.py
```diff
@@ -73,3 +73,25 @@ def test_sparse_data(self):
         self.send_signal("Data", None)
         self.widget.apply_button.button.click()
         self.assertFalse(self.widget.Error.sparse_not_supported.is_shown())
+
+    def test_data_which_makes_singular(self):
```
janezd: `test_singular_matrices` or something like that?

Orange/widgets/unsupervised/tests/test_owmanifoldlearning.py
```diff
@@ -73,3 +73,25 @@ def test_sparse_data(self):
...
+        GH-2228
+        """
+        table = Table(
+            Domain(
+                [ContinuousVariable("a"), ContinuousVariable("b")],
+                class_vars=DiscreteVariable("c", values=["0", "1"])),
+            list(zip(
+                [1, 1, 1],
+                [0, 1, 2],
+                [0, 1, 1]))
```
janezd: Would a simpler, more obviously singular matrix, like all zeros, work here as well?

jerneju: Yes, matrix with all zeros would work here as well. But I just wanted to point out that regular matrix can crash the code as well.

Orange/widgets/unsupervised/tests/test_owmanifoldlearning.py
jerneju: Fixed.

Orange/widgets/unsupervised/tests/test_owmanifoldlearning.py
jerneju: Fixed.

Orange/widgets/unsupervised/owmanifoldlearning.py
jerneju: Fixed.

PR #2232, opened 2017-04-13
[FIX] File: Raise and handle Exc. when file bad pickle

jerneju:
##### Issue
https://sentry.io/biolab/orange3/issues/208584693/

##### Description of changes
Raise TypeError when PickleReader reads a pickle file without a table (and it suppose to be there).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -78,7 +78,8 @@ def detect_encoding(filename):
                     if encoding in (b'utf-8', b'us-ascii', b'iso-8859-1',
                                     b'utf-7', b'utf-16le', b'utf-16be', b'ebcdic'):
                         return encoding.decode('us-ascii')
-        except OSError: pass  # windoze
+        except OSError:
+            pass  # windoze
 
     # file not available or unable to guess the encoding, have chardet do it
     detector = UniversalDetector()
@@ -163,7 +164,7 @@ def sanitize_variable(valuemap, values, orig_values, coltype, coltype_kwargs,
                 offset = len(new_order)
                 column = values if data.ndim > 1 else data
                 column += offset
-                for i, val in enumerate(var.values):
+                for _, val in enumerate(var.values):
                     try:
                         oldval = old_order.index(val)
                     except ValueError:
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/208584693/

Orange/data/io.py
```diff
@@ -807,7 +807,11 @@ class PickleReader(FileFormat):
 
     def read(self):
         with open(self.filename, 'rb') as f:
-            return pickle.load(f)
+            table = pickle.load(f)
+            if not isinstance(table, Table):
+                raise Exception("Not a table.")
```
janezd: Don't raise instances of `Exception`; `Exception` is just the base class for (most) other exceptions. I guess that `TypeError` would be the appropriate type of exception for this.

You can't even catch `Exception` without pylint complaining (unless silenced).

Orange/widgets/data/tests/test_owfile.py
```diff
@@ -195,3 +195,21 @@ def test_check_datetime_disabled(self):
             for i in range(4):
                 vartype_delegate.setEditorData(combo, idx(i))
                 self.assertEqual(combo.count(), counts[i])
+
+    def test_open_bad_pickle(self):
```
janezd: This is not the proper test for this code (and not the proper place for the test function itself). You modified `Orange.data.io.PickleReader`, so you should test  `Orange.data.io.PickleReader` and not a widget. You should always aim to test as little code as possible. The test should not rely on the way in which a widget that uses the reader handles exceptions.

Second, simplify this test. You don't need temporary files. Instead of writing a temporary pickle file, just patch `open` and `pickle.load` to return `None`. Then call `self.assertRaises(TypeError, PickleReader.read, "foo")`. You'll do everything in three simple lines.

Orange/data/io.py
janezd: Change the message to `"file does not contain a data table"`. It's more informative and, like most Python exceptions, starts with a lower case and has no period.

jerneju: Fixed.

Orange/widgets/data/tests/test_owfile.py
jerneju: Fixed.

jerneju:
https://sentry.io/biolab/orange3/issues/255295607/

jerneju:
https://sentry.io/biolab/orange3/issues/269531089/

jerneju:
https://sentry.io/biolab/orange3/issues/271286718/

Issue #2240, opened 2017-04-18
Example workflows not working


ajdapretnar:
Some recent changes in Scatter Plot caused Example workflows to fail. Also Classification Tree / Tree is still not compatible with the current workflow.

Reproduce: load workflows from Help - Examples.

@jerneju Could some recent changes you've made cause Scatter Plot to fail with these workflows?
@pavlin-policar Could you please check why Classification Tree workflow is still not working? I thought we fixed the channel a long time ago...

Thanks!

jerneju:
That does not work since 3424ef94c09c31c2d3a16adef585d7a58efd99b1.

pavlin-policar:
Yes, I thought this worked at some point... I think #2243 should fix the problem with the workflow, but does this problem also occur with the regression widget? I don't know if we could test this, since there don't seem to be any workflow with that as well. It would probably be a good idea to add that as well with `Orange.widgets.regression.owregressiontree.OWTreeLearner`.

ajdapretnar:
This now works. I have tested all the core examples, so I'm closing this.

PR #2241, opened 2017-04-18
[FIX] Scatterplot crashes when loaded from old workflow

jerneju:
##### Issue
#2240

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

PR #2242, opened 2017-04-18
Scheme: test if scheme examples can be opened

jerneju:
##### Issue
#2240 

##### Includes
- [ ] Code changes
- [X] Tests
- [ ] Documentation


jerneju:
@astaric : Tests _Orange/canvas/scheme/tests_ are turned off. This test should fail now but it will probably pass when #2241 and #2243 are merged.

astaric:
Make sure you add new paths that should be included in distribution to MANIFEST and setup.py (package_data).

setup.py
```diff
@@ -3,6 +3,7 @@
 import os
 import sys
 import subprocess
+from distutils.command.build_ext import build_ext
```
astaric: Please revert this.

setup.py
```diff
@@ -228,7 +229,7 @@ def run(self):
 
 
 
-class build_ext_error(build_ext):
+class BuildExtError(build_ext):
```
astaric: and this.

Orange/widgets/tests/test_workflows.py
```diff
@@ -0,0 +1,33 @@
...
+from Orange.widgets import tests
+
+class TestWorkflows(WidgetTest):
+    def test_scheme_examples(self):
+        """
+        Test if Orange workflow examples can be opened. Examples in canvas
+        and also those placed "workflows" subfolder.
+        GH-2240
+        """
+        def test(filename):
```
astaric: convert this function to a (static) method on the class. name it something along the lines of "assertCanBeLoaded". 

Add some error handling code to the test that would catch the error and fail the test with the name of the workflow that could not be loaded. When the tests fails, the name of the problematic workflow will be much more useful than the actual exception (as it will allow us to run and debug the problematic workflow).

Orange/widgets/tests/test_workflows.py
```diff
@@ -0,0 +1,33 @@
...
+            new_scheme = widgetsscheme.WidgetsScheme()
+            with open(filename, "rb") as f:
+                scheme_load(new_scheme, f)
+
+        tutors = workflows.example_workflows()
```
astaric: Move the specification of the workflows to be tested outside of the tests. I define them in a constant on a module level that would look like this:
```
TEST_WORKFLOWS=itertools.chain(
    workflows.example_workflows(),
    discover_workflows(dirname(__file__))
)
```
(I assume that you write a function discover_workflows that gets a path to a dir and yields all .ows files it finds by traversing the directory) This would simplify the test to the point where the checks in test() function could be written in the body of the for loop that iterates through the TEST_WORKFLOWS.

Orange/widgets/tests/test_workflows.py
```diff
@@ -0,0 +1,55 @@
...
+
+class TestWorkflows(WidgetTest):
+    @staticmethod
+    def assertCanBeLoaded(filename):
+        new_scheme = widgetsscheme.WidgetsScheme()
+        with open(filename, "rb") as f:
+            try:
+                scheme_load(new_scheme, f)
+            except UnknownWidgetDefinition as e:
+                log.error("Unknown widget '%s' while loading '%s'.",
```
astaric: use self.fail instead of logging and then raising.

Orange/widgets/tests/test_workflows.py
```diff
@@ -0,0 +1,55 @@
...
+                          str(e), filename,
+                          exc_info=True)
+                raise
+            except Exception:
+                log.error("Error occurred while loading '%s'.",
```
astaric: same here

Orange/widgets/tests/test_workflows.py
```diff
@@ -0,0 +1,55 @@
...
+                raise
+
+    def test_scheme_examples(self):
+        """
+        Test if Orange workflow examples can be opened. Examples in canvas
+        and also those placed "workflows" subfolder.
+        GH-2240
+        """
+        for ows_file in TEST_WORKFLOWS:
+            self.assertCanBeLoaded(ows_file)
```
astaric: as this test is pretty short now, what about moving the content of the assert method back to the test?

jerneju:
@astaric: Should I write another test to increase codecov/patch diff hit?

PR #2243, opened 2017-04-18
Tree: compatibility when loading old schemes

jerneju:
##### Issue
#2240

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


pavlin-policar:
If the widget was called `Orange.widgets.classify.owclassificationtree.OWTreeLearner`, perhaps then `Orange.widgets.regression.owregressiontree.OWTreeLearner` also existed at some point or another. I can't be sure, but it sure can't hurt to add that as well.

jerneju:
Fixed.

PR #2246, opened 2017-04-19
PCA: Show Error Message for Sparse Data

### LINKED ISSUE #2159: PCA should work on sparse
##### Orange version

3.5.dev

##### Expected behavior
PCA works on sparse.

##### Actual behavior
PCA fails on sparse.

Corpus - Bag of Words - PCA.

nikicc:
##### Issue

Fixes #2159.

##### Description of changes
Show error message when sparse data comes into the PCA widget.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owpca.py
+++ b/Orange/widgets/unsupervised/owpca.py
@@ -55,6 +55,7 @@ class OWPCA(widget.OWWidget):
     class Error(widget.OWWidget.Error):
         no_features = widget.Msg("At least 1 feature is required")
         no_instances = widget.Msg("At least 1 data instance is required")
+        sparse_data = widget.Msg("Sparse data is not supported")
 
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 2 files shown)


Orange/widgets/unsupervised/owpca.py
```diff
@@ -192,6 +193,9 @@ def fit(self):
         self.start_button.setEnabled(False)
         if self.data is None:
             return
+        if self.data.is_sparse():
```
lanzagar: Wouldn't this check be more appropriate in set_data.

lanzagar:
Also a test that asserts setting a sparse data set raises the error would probably make all the checks green ;)

Orange/widgets/unsupervised/owpca.py
nikicc: It was placed here since `self.data is None` check is here and this is somehow related. Also all other checks, like `len(data.domain.attributes) == 0` and `len(data) == 0` are here.

Though I think this fits better into `set_data`, I put it here to be consistent with the current state of the code. I can move it into `set_data` if you think that will be more clear.

nikicc:
@lanzagar test added 😄

PR #2219, opened 2017-04-11
[FIX] Test & Score: Instantly recognize new input

jerneju:
##### Issue
#1842 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

Issue #1842, opened 2016-12-28
Test&Score: Instantly recognize new input


ajdapretnar:
##### Orange version
3.3.10-dev

##### Expected behavior
When connecting a new instance of test data to Test&Score, the widget immediately reports 'Test data present, but unused'.

##### Actual behavior
Test&Score stays silent and doesn't trigger a warning.

##### Steps to reproduce the behavior
File - Select Rows - connect two inputs to Test&Score. By default the second input is 'Matching Data - Test Data'. Change it to 'Remaining Data - Test Data'. The widget still doesn't trigger a warning. When Test data is selected and then unselected, the warning appears.

SilentFlame:
@ajdapretnar  are we connecting the two inputs to **Test&Score** from **Select Rows** only?
or something else is the case you are talking about?

Please let me know, as would like to work on this.

ajdapretnar:
![workflow](https://cloud.githubusercontent.com/assets/12524972/21839410/f1af2614-d7d7-11e6-9e4b-b569a579b5c5.PNG)

This is the mentioned workflow. Basically, the warning you see above Test & Score should trigger immediately upon connecting a channel to 'Test Data'.

![test](https://cloud.githubusercontent.com/assets/12524972/21839446/1e301392-d7d8-11e6-89ef-9e4bb583bc29.PNG)

Now, only when I change from cross validation to any other method does the warning trigger.

SilentFlame:
@ajdapretnar still unable to understand the problem.

as in my case no warning triggers.
![1](https://cloud.githubusercontent.com/assets/9693795/21842688/ea6d2bcc-d80d-11e6-8bd3-0fea693bc20e.png)

please provide some more insights on the issue.

ajdapretnar:
The reason for the above issue is that your 'Unmatched Data - Data' has no data as indicated by the dashed line. This probably means your 'Matching Data' subset is also the entire data set. You need to select a subset smaller than the entire data and construct the same workflow.

Now, when I connect the link to 'Test Data' (from either Matching Data or Unmatched, doesn't matter), the warning should trigger to notify the user that he or she should switch to 'Test Data' method. Now it doesn't initially. It's just for greater user friendliness.

To replicate, you need:
1. Train data as Data input.
2. Learner (e.g. LogReg).
3. First the model should be trained with cross validation in T&S.
4. Now connect test data to Test Data input.
5. **The warning should trigger, notifying the user that it has test data on the input but is not using it.**
[...]

PR #2247, opened 2017-04-19
[FIX] Test & Score: Instantly recognize new input

jerneju:
##### Issue
#1842
#2219

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -438,17 +438,16 @@ class OWTestLearners(OWWidget):
             elif self.test_data.domain.class_var != class_var:
                 self.Error.class_inconsistent()
                 return
+        elif self.test_data is not None:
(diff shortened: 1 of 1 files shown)


Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -443,13 +443,14 @@ def _update_results(self):
...
                  if slot.results is None]
         learners = [slot.learner for _, slot in items]
-        if len(items) == 0:
-            return
 
         if self.test_data is not None and \
                 self.resampling != OWTestLearners.TestOnTest:
             self.Warning.test_data_unused()
 
+        if len(items) == 0:
```
lanzagar: Instead of moving the check on `items` away from its creation, how about moving the check for `TestOnTest` up, together with other similar checks. The `if`s can even be merged.

Otherwise this seems to fix the problem :)

PR #2249, opened 2017-04-20
Edit Domain: enter key accepts a name

jerneju:
##### Issue
Enter key resets the names and does not accept a new name.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


jerneju:
Codecov is complaining. Should we care about this in this PR?

janezd:
I don't think so. This is difficult to test. @astaric?

astaric:
I would not bother. If it works, it can be merged as is.

PR #2252, opened 2017-04-21
Avoid random tests stalling

astaric:
##### Issue
Tests sometimes freeze, restart usually helps.

##### Description of changes
Add timeouts to blocking operations, raise TimeoutError after a predetermined amount of time passes. This way the test results will show which tests are problematic and we can consider changing, skipping, or removing them.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -90,6 +90,10 @@ class WidgetTest(GuiTest):
         cls.widgets.append(report)
         OWReport.get_instance = lambda: report
 
+    def tearDown(self):
+        """Process any pending events before the next test is executed."""
+        self.process_events()
+
     def create_widget(self, cls, stored_settings=None, reset_default_settings=True):
         """Create a widget instance using mock signal_manager.
 
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -1,12 +1,12 @@
+from contextlib import contextmanager
 import os
 import time
-from contextlib import contextmanager
 import unittest
 from unittest.mock import Mock
 
+import numpy as np
 import sip
 
-import numpy as np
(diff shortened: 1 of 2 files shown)


Orange/widgets/tests/base.py
```diff
@@ -169,6 +179,8 @@ def process_events(until: callable=None):
                     return result
             except Exception:  # until can fail with anything; pylint: disable=broad-except
                 pass
+            if (datetime.now() - started).total_seconds() * 1000 > timeout:
```
kernc: I'd use `time.clock()` over `datetime.now()`. It's already in seconds.

Orange/widgets/utils/webview.py
```diff
@@ -475,6 +476,21 @@ def remove(self, id):
...
+
+        Parameters
+        ----------
+        until: callable that returns True when condition is satisfied
+        timeout: number of milliseconds to wait until TimeoutError is raised
+        """
+        started = datetime.now()
+        while not until():
+            qApp.processEvents(QEventLoop.ExcludeUserInputEvents)
+            if (datetime.now() - started).total_seconds() * 1000 > timeout:
```
kernc: Likewise.

Orange/widgets/utils/webview.py
```diff
@@ -534,23 +549,20 @@ def __init__(self, parent=None, bridge=None, *, debug=False, **kwargs):
             self._results = IdStore()
 
         def _evalJS(self, code):
-            while not self._jsobject_channel.is_all_exposed():
-                qApp.processEvents(QEventLoop.ExcludeUserInputEvents)
+            wait(until=lambda: self._jsobject_channel.is_all_exposed())
```
kernc: Can pass call back directly without lambda.

PR #2254, opened 2017-04-21
[FIX] owtestlearners: Fix migration of old settings

lanzagar:
##### Issue

Old settings were not migrated correctly (`context_settings` might not be saved/exist)

##### Description of changes
Drop old/incompatible `context_settings` instead of modifying them.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


astaric:
@jerneju can you check if the latest fix solves the problems you have encountered? If it does, please update your review accordingly.

PR #2257, opened 2017-04-21
[ENH] Test and Score: Allow choosing columns

janezd:
We may want to add more scorers in the future, but we'd like to keep the widget simple. This PR allows the user to choose the columns by right-clicking in the top header.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -7,7 +7,7 @@ from collections import OrderedDict, namedtuple
 import numpy as np
 
 from AnyQt import QtGui
-from AnyQt.QtWidgets import QHeaderView, QStyledItemDelegate
+from AnyQt.QtWidgets import QHeaderView, QStyledItemDelegate, QMenu
 from AnyQt.QtGui import QStandardItemModel, QStandardItem
 from AnyQt.QtCore import Qt, QSize
 
@@ -170,6 +170,10 @@ class OWTestLearners(OWWidget):
     TARGET_AVERAGE = "(Average over classes)"
     class_selection = settings.ContextSetting(TARGET_AVERAGE)
 
+    shown_scores = \
+        settings.Setting({"AUC", "CA", "F1", "Precision", "Recall",
+                          "MSE", "RMSE", "MAE", "R2"})
+
     class Error(OWWidget.Error):
         train_data_empty = Msg("Train data set is empty.")
         test_data_empty = Msg("Test data set is empty.")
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I like this. It's a good solution to the 'kitchen sink' problem.
However, when running Test&Score, I get the following warning in the terminal:
```
/Users/ajda/orange/orange3/Orange/widgets/settings.py:378: UserWarning: Could not read defaults for widget <class 'Orange.widgets.evaluate.owtestlearners.OWTestLearners'>
The following error occurred:

'context_settings'
  .format(self.widget_class, ex))
```

It is not related to this PR, I get it on the latest master as well, but it seems to be happening in Test&Score and perhaps it can be patched here. Or in another PR... Oh, and it doesn't break anything obvious.

janezd:
I've been told by @astaric that the aforementioned error has already been patched by @lanzagar in another PR.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -660,6 +676,27 @@ def _invalidate(self, which=None):
 
         self.commit()
 
+    def show_column_chooser(self, pos):
+        # pylint doesn't know that self.shown_scores is a set, not a Setting
+        # pylint: disable=unsupported-membership-test
+        def update(col_name, state):
```
astaric: state -> visible?

Orange/widgets/evaluate/tests/test_owtestlearners.py
```diff
@@ -70,6 +75,58 @@ def test_CrossValidationByFeature(self):
...
+        actions = collections.OrderedDict()
+        menu_add_action = QMenu.addAction
+
+        def addAction(menu, a):
+            action = menu_add_action(menu, a)
+            actions[a] = action
+            return action
+
+        def execmenu(*_):
+            self.assertEqual(list(actions), list(all)[1:])
```
astaric: A comment explaining why complication with execmenu is needed would be nice

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -170,6 +170,10 @@ class OWTestLearners(OWWidget):
     TARGET_AVERAGE = "(Average over classes)"
     class_selection = settings.ContextSetting(TARGET_AVERAGE)
 
+    shown_scores = \
+        settings.Setting({"AUC", "CA", "F1", "Precision", "Recall",
```
astaric: Non-primitive settings cannot be stored to workflow in json format resulting in base64 encoded pickled blobs. Not sure if we care, since the widget uses context settings, which also cannot be serialized with json.

janezd: **If** list can be stored in json and **if** we care(d) I can replace the set with a list. No big changes in the code, except that conceptually this is a set, not a list.

Can it, do we, should I?

Orange/widgets/evaluate/owtestlearners.py
janezd: Perhaps misleading since it would imply it is (was) already visible. I changed it to `checked`.

Orange/widgets/evaluate/tests/test_owtestlearners.py
janezd: Fixed.

Orange/widgets/evaluate/owtestlearners.py
astaric: yes, probably not, no

If this change would "break" json serialization, I would say yes, but since it is already broken, I would say leave it like it is.

PR #2258, opened 2017-04-24
[FIX] Settings: check for attributes before comparing

jerneju:
##### Issue
https://sentry.io/biolab/orange3/issues/256593225/

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -1171,7 +1171,7 @@ class PerfectDomainContextHandler(DomainContextHandler):
         """Context only matches when domains are the same"""
 
         return (self.PERFECT_MATCH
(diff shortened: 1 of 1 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/256593225/

jerneju:
https://sentry.io/biolab/orange3/issues/242062399/

astaric:
Why would there be a context that does not have attributes?

astaric:
Context handlers should be allowed to assume that the context is of the correct kind. The right way to approach this should be fixing the offending widget, which is (although at the moment incorrectly) being done in #2254. Let's continue the discussion there.

Issue #2160, opened 2017-03-30
Sieve Diagram should work on sparse


ajdapretnar:
##### Orange version

3.5.dev

##### Expected behavior
Sieve Diagram works on sparse.

##### Actual behavior
Sieve Diagram fails on sparse.

##### Steps to reproduce the behavior
Corpus - Bag of Words - Sieve Diagram.

jerneju:
https://sentry.io/biolab/orange3/issues/244854556/

PR #2260, opened 2017-04-24
Sieve Diagram: Sparse Support

jerneju:
##### Issue
#2160

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owsieve.py
+++ b/Orange/widgets/visualize/owsieve.py
@@ -168,13 +168,6 @@ class OWSieveDiagram(OWWidget):
             self.domain_model.set_domain(None)
         else:
             self.domain_model.set_domain(data.domain)
-            if any(attr.is_continuous for attr in chain(data.domain, data.domain.metas)):
-                discretizer = Discretize(
-                    method=EqualFreq(n=4), remove_const=False,
-                    discretize_classes=True, discretize_metas=True)
-                self.discrete_data = discretizer(data)
-            else:
-                self.discrete_data = data
         self.attrs = [x for x in self.domain_model if isinstance(x, Variable)]
         if self.attrs:
             self.attr_x = self.attrs[0]
@@ -184,6 +177,8 @@ class OWSieveDiagram(OWWidget):
             self.areas = []
             self.selection = set()
         self.openContext(self.data)
+        if self.data:
+            self.discrete_data = self.sparse_to_dense(data, True)
         self.resolve_shown_attributes()
         self.update_graph()
         self.update_selection()
@@ -191,7 +186,7 @@ class OWSieveDiagram(OWWidget):
         self.vizrank.initialize()
         self.vizrank_button.setEnabled(
             self.data is not None and len(self.data) > 1 and
-            len(self.data.domain.attributes) > 1)
+            len(self.data.domain.attributes) > 1 and not self.data.is_sparse())
 
     def set_attr(self, attr_x, attr_y):
         self.attr_x, self.attr_y = attr_x, attr_y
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owsieve.py
```diff
@@ -192,6 +187,7 @@ def set_data(self, data):
         self.vizrank_button.setEnabled(
             self.data is not None and len(self.data) > 1 and
             len(self.data.domain.attributes) > 1)
+        self.vizrank_button.setVisible(self.data is not None and not self.data.is_sparse())
```
nikicc: I prefer `setEnabled` than `setVisible`. Currently, when no data is present, we disable it not hide it. Let's do the same when the data is sparse.

Orange/widgets/visualize/owsieve.py
```diff
@@ -183,6 +176,8 @@ def set_data(self, data):
             self.attr_x = self.attr_y = None
             self.areas = []
             self.selection = set()
+        if self.data:
```
nikicc: I would move these lines a bit up, to the place where it originally was. Hence, you also can remove one `if`.

Orange/widgets/visualize/owsieve.py
```diff
@@ -200,9 +196,33 @@ def set_attr(self, attr_x, attr_y):
     def update_attr(self):
         """Update the graph and selection."""
         self.selection = set()
+        self.discrete_data = self.sparse_to_dense(self.data)
         self.update_graph()
         self.update_selection()
 
+    def sparse_to_dense(self, data, init=False):
```
nikicc: Can you add a simple description of the method?

Orange/widgets/visualize/tests/test_owsieve.py
```diff
@@ -86,3 +87,21 @@ def test_metadata(self):
...
+
+    def test_sparse_data(self):
+        """
+        Sparse support.
+        GH-2160
+        GH-2260
+        """
+        table = Table("iris")
+        self.send_signal("Data", table)
+        self.assertEqual(len(self.widget.discrete_data.domain), 5)
```
nikicc: Maybe here it's better to use `len(table.domain)` than 5.

Orange/widgets/visualize/owsieve.py
jerneju: `sparse_to_dense` function needs `attr_x` and `attr_y`.

Orange/widgets/visualize/owsieve.py
```diff
@@ -200,9 +196,39 @@ def set_attr(self, attr_x, attr_y):
...
+        """
+        def discretizer(data):
+            if any(attr.is_continuous for attr in chain(data.domain, data.domain.metas)):
+                discretize = Discretize(
+                    method=EqualFreq(n=4), remove_const=False,
+                    discretize_classes=True, discretize_metas=True)
+                return discretize(data)
+            return data
+
+        if not data.is_sparse() and init:
```
nikicc: Codeclimate is complaining about complexity. Will this code also work and hopefully reduce complexity?

```
if not data.is_sparse() and not init:
    return self.discrete_data
if data.is_sparse():
    attrs = {self.attr_x,
             self.attr_y}
    keys = []
    for attr in chain(self.data.domain, data.domain.metas):
        if attr in attrs:
            keys.append(attr.name)
    new_domain = data.domain.select_columns(keys)
    data = Table.from_table(new_domain, data)
    data.X = data.X.toarray()
return discretizer(data)
```

Orange/widgets/visualize/owsieve.py
```diff
@@ -200,9 +196,39 @@ def set_attr(self, attr_x, attr_y):
...
+            return discretizer(data)
+        elif not data.is_sparse() and not init:
+            return self.discrete_data
+        elif data.is_sparse():
+            attrs = {self.attr_x,
+                     self.attr_y}
+            keys = []
```
astaric: What is the following code (filling of the keys list) supposed to do?

jerneju: Extract one or two columns from a sparse table.

Orange/widgets/visualize/owsieve.py
jerneju: Fixed.

Orange/widgets/visualize/owsieve.py
nikicc: @astaric iterating over the domain was removed now. I think this is what you had in mind :)

Orange/widgets/visualize/owsieve.py
```diff
@@ -192,6 +187,7 @@ def set_data(self, data):
         self.vizrank_button.setEnabled(
             self.data is not None and len(self.data) > 1 and
             len(self.data.domain.attributes) > 1)
+        self.vizrank_button.setEnabled(self.data is not None and not self.data.is_sparse())
```
nikicc: This defeats checking for `len(self.data) > 1` and `len(self.data.domain.attributes) > 1` in the previous line since we just check for sparsity.

Please add `and not self.data.is_sparse()` to the condition in the line above.

PR #2262, opened 2017-04-25
Highcharts: Fix Bring Selection to Front

nikicc:
##### Issue

Some highcharts objects don't have `this.graphics` property and for them selection was broken. 

##### Description of changes
Bring to front in hopefully more robust way.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/_highcharts/orange-selection.js
+++ b/Orange/widgets/_highcharts/orange-selection.js
@@ -33,14 +33,7 @@ Highcharts.wrap(Highcharts.Point.prototype, 'select', function (proceed) {
     proceed.apply(this, Array.prototype.slice.call(arguments, 1));
 
     // Raise selected point to front
-    if (this.selected) {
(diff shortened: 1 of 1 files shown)


nikicc:
@kernc this looks equally good to me and doesn't break some of my widgets. What do you think?

kernc:
This looks like it fronts the whole series and not necessarily individual points within the series (consider a scatter plot). In both cases, we're actually purging Highcharts from this repo (coming right up). Any chance you could instead patch your own copy?

astaric:
As @kernc said, highcharts will be removed from Orange in the next major release (3.5).

Any add-ons that use it should include a local copy of the highcharts helper code (this was already done in [educational](https://github.com/biolab/orange3-educational/pull/32) and [timeseries](https://github.com/biolab/orange3-timeseries/pull/12) add-ons).

PR #2264, opened 2017-04-27
Fix multiple warnings in documentation

astaric:
##### Issue
When building documentation, lots of warnings are shown. Some resulted in badly rendered documentation (see lists in [manifold learning docs](https://docs.orange.biolab.si/3/visual-programming/widgets/unsupervised/manifoldlearning.html))

##### Description of changes
Fix most of the warnings.

What remains are warnings that are caused by "including" scikit-learn descriptions that contain references to labels in their user guide. We need to either inter-sphinx properly or remove those links.

##### Includes
- [x] Documentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ In addition, we add the following guidelines:
 * Only ever `import *` to make objects available in another namespace,
   preferably in *\_\_init\_\_.py*. Everywhere else use explicit object
   imports.
-* Use [Napoleon]-comaptible (e.g. NumPy style) docstrings, preferably with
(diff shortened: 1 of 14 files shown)


astaric:
I have tried and failed as well :)

If you succeed, feel free to add additional commits to this branch (no need to open a separate PR).

janezd:
After like two hours of exploration which included adding `-v -v -v` to the Makefile and a peek into Sphinx source code --- I found it. I'm totally proud about myself now.

This deserves a prize for the "commit of the month". Thanks in advance.

astaric:
Nice catch!

PR #2266, opened 2017-04-28
preprocess.score.Scorer: Improve error messages

janezd:
##### Issue

As we discussed in #2205, features selection in the Preprocess widget shows ugly error messages. This happens since it shows exceptions raised by `preprocess.score.Scorer`, which are, in turn, in need of improvement.

##### Description of changes

This PR makes the exceptions issued by `preprocess.score.Scorer` presentable.

I think we should not do this in general since exceptions are for programmers and error messages in the widgets are for the end-user, so the former have to be exact and technical, including the type names etc, while the latter favour readability over exactness.

A check in `preproces.fss` is removed since it is also done in `preprocess.score.Scorer`.

##### Includes
- [X] Code changes

No additional tests are needed.

### DIFF
--- a/Orange/preprocess/score.py
+++ b/Orange/preprocess/score.py
@@ -1,3 +1,4 @@
+import re
 from collections import defaultdict
 from itertools import chain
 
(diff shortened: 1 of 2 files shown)


Orange/preprocess/score.py
```diff
@@ -30,12 +31,37 @@ class Scorer(_RefuseDataInConstructor, Reprable):
         RemoveNaNClasses()
     ]
 
+    @property
```
janezd: Sadly, Python doesn't provide `@classproperty` (and, no, you can't combine `@classmethod` and `@property`). Implementing it doesn't make sense for just a few (overpedantic) uses. :)

astaric:
As the old errors were cryptic (to programmers and users), this is a welcome improvement.

PR #2269, opened 2017-04-28
Scoring: Implement a registry of Score classes

janezd:
##### Issue

We want to let add-ons add new classes for scoring models. The Test & Score widget should recognize and use them.

##### Description of changes

- PR implements a registry of subclasses of `Score` through meta class `ScoreMetaType`. It is derived from `WrapperMeta` and implements a registry. Multiple inheritance (from `Registry`) can't be used since a class can't have two `__new__` methods.

- Subclasses of `Score` can provide a better name and a tooltip for the column. This is undesirable (the code on this level shouldn't be aware of GUI) but rather inevitable if we want to let add-ons add stuff to widgets.

- Subclasses of `Score` can be declared abstract (see `ClassificationScore`, `RegressionScore`)

- The Test & Score widget takes all non-abstract scores from the registry. Built-in scores are ordered as they were before this PR.

Since the PR didn't pass the coverage test because of reformatting `graph_ranks`, I now call this function from tests. It's difficult to test what it draws, so the tests just ensures it doesn't straight out crash.

##### Includes
- [X] Code changes
- [x] Tests

<strike>I don't know how and what to test here. Testing `owtestlearners._get_score` on patched `scoring.Score.registry` seems ridiculous.</strike>

### DIFF
--- a/Orange/evaluation/scoring.py
+++ b/Orange/evaluation/scoring.py
@@ -11,17 +11,41 @@ Examples
 """
 
 import math
-import sys
 
 import numpy as np
 import sklearn.metrics as skl_metrics
+
+from Orange.data import DiscreteVariable, ContinuousVariable
 from Orange.misc.wrapper_meta import WrapperMeta
 
 __all__ = ["CA", "Precision", "Recall", "F1", "PrecisionRecallFSupport", "AUC",
            "MSE", "RMSE", "MAE", "R2", "compute_CD", "graph_ranks", "LogLoss"]
 
 
-class Score(metaclass=WrapperMeta):
+class ScoreMetaType(WrapperMeta):
+    """
+    Maintain a registry of non-abstract subclasses and assign the default
+    value of `name`.
+
+    The existing meta class Registry cannot be used since a meta class cannot
+    have multiple inherited __new__ methods."""
+    def __new__(mcs, name, bases, dict_, **kwargs):
+        cls = WrapperMeta.__new__(mcs, name, bases, dict_)
+        # Essentially `if cls is not Score`, except that Score may not exist yet
+        if hasattr(cls, "registry"):
+            if not kwargs.get("abstract"):
+                # Don't use inherited names, look into dict_
+                cls.name = dict_.get("name", name)
+                cls.registry[name] = cls
+        else:
+            cls.registry = {}
+        return cls
+
+    def __init__(cls, *args, **_):
+        WrapperMeta.__init__(cls, *args)
+
+
+class Score(metaclass=ScoreMetaType):
     """
     ${sklpar}
     Parameters
@@ -29,8 +53,17 @@ class Score(metaclass=WrapperMeta):
     results : Orange.evaluation.Results
         Stored predictions and actual data in model testing.
     """
+    __wraps__ = None
+
     separate_folds = False
     is_scalar = True
+    is_binary = False  #: If true, compute_score accepts `target` and `average`
+    #: If the class doesn't explicitly contain `abstract=True`, it is not
+    #: abstract; essentially, this attribute is not inherited
+    abstract = True
+    class_types = ()
+    name = None
+    long_name = None  #: A short user-readable name (e.g. a few words)
 
     def __new__(cls, results=None, **kwargs):
         self = super().__new__(cls)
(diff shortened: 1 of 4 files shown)


Orange/evaluation/scoring.py
```diff
@@ -287,46 +286,35 @@ def compute_score(self, results, eps=1e-15, normalize=True,
...
-
-
 # pylint: disable=invalid-name
-class R2(Score):
+class R2(RegressionScore):
     __wraps__ = skl_metrics.r2_score
-    class_types = (ContinuousVariable, )
     long_name = "Coefficient of determination"
 
     def compute_score(self, results):
```
janezd: This method is removed in the next comment. (Adding by chunks didn't work properly. :))

astaric:
You can test the MetaClass magic:
 - passing name as a keyword argument
 - name on inherited classes
And widget
 - is "DummyScore" defined in a test visible in Test & Score columns

Also, please rebase, as I have merged the other Test & Score PR.

Orange/evaluation/scoring.py
```diff
@@ -11,26 +11,67 @@
...
+
+    def __init__(cls, *args, **_):
+        WrapperMeta.__init__(cls, *args)
+
+
+# Derived classes may include a keyword argument name with a short readable
+# name of the score, and an attribute long_name with a few word description.
+#
+# class TGOGScore(Score, name="Thingamagog"):
+#    long_name = "Magic Thingamagog score"
```
astaric: TypeError: `__new__()` got multiple values for argument 'name'

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -538,19 +548,18 @@ def puts_results(self, learners, results, class_var):
...
-            else:
-                headers.extend(regression_stats.headers)
-
-        # remove possible extra columns from the model.
-        for i in reversed(range(len(headers),
-                                self.result_model.columnCount())):
-            self.result_model.takeColumn(i)
-
-        self.result_model.setHorizontalHeaderLabels(headers)
+            func = classification_stats if self.data.domain.has_discrete_class \
```
astaric: I would prefer

```
if self.data.domain.has_discrete_class:
    scores = classification_stats.scores
else:
    scores = regression_stats.scores

File: Orange/widgets/evaluate/owtestlearners.py [hunk 4]
```diff
@@ -538,19 +548,18 @@ def puts_results(self, learners, results, class_var):
...
+                else regression_stats
+            scores = func.scores
+        else:
+            scores = []
+        model = self.result_model
+        model.setColumnCount(1 + len(scores))
+        for col_1, score in enumerate(scores):
```
astaric: what does col_1 stand for? Would name index/idx suffice?

kernc:
> I don't know how and what to test here.

You can use the registry to test all scorers in a single place with something meaningful and/or edge-cases.

Orange/evaluation/scoring.py
janezd: The comment referred to an older, broken code. `name` is now just a class attribute. I removed the comment.

Orange/widgets/evaluate/owtestlearners.py
janezd: Fixed: this code is gone altogether.

Orange/widgets/evaluate/owtestlearners.py
janezd: It stood for `col - 1`. Apparently only I use this convention, so I removed `_1`.

janezd:
> You can test the MetaClass magic:
> - passing name as a keyword argument

It wasn't intended to work like that. The comment about it remained from an older, broken version.

> name on inherited classes

Check.

> And widget: is "DummyScore" defined in a test visible in Test & Score columns

It was once again a good idea to force people (=me) to test their code. :)

A list of scores was constructed when module was loaded, so it would appear only if the add-on was loaded before the widget. And it could never work in tests. In the process of fixing this, I removed functions `classification_stats` and `regression_stats` that were never called, but were used for storing attribute `scores`. The code affected by their removal is now simpler than before.

janezd:
@kernc 

> You can use the registry to test all scorers in a single place with something meaningful and/or edge-cases.

This is a good idea. Also, the widget could use some additional tests. But not in this PR.

astaric:
@janezd,this PR requires a non-trivial rebase onto the modified test&score widget. Please mention me when done so I can merge.

PR #2272, opened 2017-05-02
Remove .metadata file if meta data is removed

markotoplak:
Fixes a bug when meta data is removed, data saved. In this case the old meta
data file still exists and will be loaded the next time the data is opened.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -13,7 +13,7 @@ from functools import lru_cache
 from itertools import chain, repeat
 from math import isnan
 from numbers import Number
-from os import path, unlink
+from os import path, remove
 from tempfile import NamedTemporaryFile
 from urllib.parse import urlparse, urlsplit, urlunsplit, unquote as urlunquote
 from urllib.request import urlopen, Request
@@ -395,7 +395,7 @@ class FileFormat(metaclass=FileFormatMeta):
             if getattr(data, 'attributes', None):
                 write_file(metafile)
             elif path.exists(metafile):
-                unlink(metafile)
+                remove(metafile)
 
     @classmethod
     def set_table_metadata(cls, filename, table):
@@ -941,7 +941,7 @@ class UrlReader(FileFormat):
 
             reader = self.get_reader(f.name)
             data = reader.read()
-            unlink(f.name)
+            remove(f.name)
         # Override name set in from_file() to avoid holding the temp prefix
         data.name = path.splitext(name)[0]
         data.origin = self.filename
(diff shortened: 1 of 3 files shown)


Orange/data/io.py
```diff
@@ -379,16 +379,24 @@ def write(cls, filename, data):
...
-                with open(filename + '.metadata', 'wb') as f:
+                with open(fn, 'wb') as f:
                     pickle.dump(data.attributes, f, pickle.HIGHEST_PROTOCOL)
 
+        if isinstance(filename, str):
+            metafile = filename + '.metadata'
+            if path.exists(metafile) and not getattr(data, 'attributes', None):
+                unlink(metafile)
+            if getattr(data, 'attributes', None):
+                write_file(metafile)
```
janezd: For lines 395-398, I would find

```
if getattr(data, 'attribute' , None):
    write_file(metafile)
elif path.exists(metafile):
    unlink(metafile)
```

easier to read.

Orange/tests/test_tab_reader.py
```diff
@@ -226,3 +225,15 @@ def test_no_metadata(self):
...
+        tempdir = tempfile.mkdtemp()
+        table = Table("titanic")
+        table.attributes["a"] = "aa"
+        fname = path.join(tempdir, "out.tab")
+        TabReader.write_table_metadata(fname, table)
+        self.assertTrue(path.isfile(fname + ".metadata"))
+        del table.attributes["a"]
+        TabReader.write_table_metadata(fname, table)
+        self.assertFalse(path.isfile(fname + ".metadata"))
+        shutil.rmtree(tempdir)
```
janezd: I guess you should do this in `finally` (of a `try` put under `tempdir = ...`) to ensure that the directory is removed even when the test fails?

Orange/data/io.py
```diff
@@ -379,16 +379,24 @@ def write(cls, filename, data):
...
                                       for kv in data.attributes.items()))
             else:
-                with open(filename + '.metadata', 'wb') as f:
+                with open(fn, 'wb') as f:
                     pickle.dump(data.attributes, f, pickle.HIGHEST_PROTOCOL)
 
+        if isinstance(filename, str):
+            metafile = filename + '.metadata'
+            if path.exists(metafile) and not getattr(data, 'attributes', None):
+                unlink(metafile)
```
janezd: I confess that I didn't know what `unlink` does. If it's not too dear to you, would you change it to `remove`? :) [Python documentation for unlink](https://docs.python.org/3/library/os.html#os.unlink) just redirects the reader to `remove`.

janezd:
Thanks for also fixing the other tests. :)

PR #2273, opened 2017-05-03
[ENH] Check if updates are available upon startup

kernc:
##### Issue
Too many users using stale Orange versions.

##### Description of changes
Upon application startup, check if newer software version is available.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -7,6 +7,7 @@ import os
 import sys
 import gc
 import re
+import time
 import logging
 from logging.handlers import RotatingFileHandler
 import optparse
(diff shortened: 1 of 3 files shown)


Orange/canvas/__main__.py
```diff
@@ -344,14 +345,42 @@ def show_message(message):
...
-                    QUrl("http://orange.biolab.si/survey/short.html"));
+                    QUrl("https://orange.biolab.si/survey/short.html"));
                 settings.setValue("startup/show-survey", not success)
             else:
                 settings.setValue("startup/show-survey", result != QMessageBox.No)
 
         question.finished.connect(handle_response)
         question.show()
 
+    check_updates = settings.value('startup/check-updates', True, type=bool)
```
astaric: Could this code be put in a separate function, called here with settings as parameter?

main is already too long and I am not fond of making it even longer

PR #2274, opened 2017-05-04
[FIX] Distributions: make sure cvar values are strings and not ints

jerneju:
##### Issue
Test & Score sends data where number of Folds is int and Distributions widget cannot handle that. ComboBox accepts only strings and there are some other issues as well.

https://sentry.io/biolab/orange3/issues/242009555/

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owdistributions.py
+++ b/Orange/widgets/visualize/owdistributions.py
@@ -283,6 +283,7 @@ class OWDistributions(widget.OWWidget):
             self.var = self.varmodel[varidx]
         if self.groupvar_idx > 0:
             self.cvar = self.groupvarmodel[self.groupvar_idx]
+            self.cvar.values = list(map(str, self.cvar.values))
             self.cb_prob.clear()
             self.cb_prob.addItem("(None)")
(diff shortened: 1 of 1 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/242009555/

jerneju:
https://sentry.io/biolab/orange3/issues/242009555/

markotoplak:
Perhaps we could enforce str limitation in DiscreteVariable and we would quickly catch all similar bugs and fix them at the source.

janezd:
> Test & Score sends data where number of Folds is int and Distributions widget cannot handle that. 

If Test & Score makes a mess, other widgets don't have to learn to live with it. This bug must be fixed in Test & Score. God knows how many other widgets (correctly) assume that values are strings.

janezd:
I closed this PR and opened a RFC PR #2277, which also includes the fix for this issue. If #2277 is not accepted, the fix can be cherry picked (d2438637209015f68541e5d363273d2699f712b6).

PR #2277, opened 2017-05-04
DiscreteVariable: Raise exception on non-string values

janezd:
##### Issue

Test & Score created discrete variables with non-string values, which cause problems in Distributions. In #2274 we discussed that we should prohibit this and raise the exception early.

##### Description of changes

- Raise an exception in `DiscreteVariable`'s constructor and in `add_value` if the value is not a string.
- Add a test
- Remove a test that assigned `[1, 2, 3]` to values.
- Fix Results.get_augmented_data that used in int's instead of str's and broke Test & Score.

There are other ways to add a non-string value. But if one want to be destructive, she can easily assign a dict to `values`, and we're not preventing this either.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/tests/test_variable.py
+++ b/Orange/tests/test_variable.py
@@ -247,6 +247,11 @@ class TestDiscreteVariable(VariableTest):
         self.assertEqual(len(var.colors), 4)
         np.testing.assert_array_equal(var.colors[2], user_defined)
 
+    def test_no_nonstringvalues(self):
+        self.assertRaises(TypeError, DiscreteVariable, "foo", values=["a", 42])
+        a = DiscreteVariable("foo", values=["a", "b", "c"])
+        self.assertRaises(TypeError, a.add_value, 42)
+
 
 @variabletest(ContinuousVariable)
 class TestContinuousVariable(VariableTest):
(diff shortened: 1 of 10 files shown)


janezd:
This PR already fixed (but was not merged!) what was later fixed in #2348. I removed the commit that fixed the actual bug, but the commit I kept prevents similar bugs from appearing in the future by prohibiting non-string values (that is, by raising an exception early).

Orange/tests/test_variable.py
```diff
@@ -247,6 +247,11 @@ def test_colors(self):
         self.assertEqual(len(var.colors), 4)
         np.testing.assert_array_equal(var.colors[2], user_defined)
 
+    def test_no_nonstringvalues(self):
+        self.assertRaises(TypeError, DiscreteVariable, "foo", values=["a", 42])
+        a = DiscreteVariable("foo", values=["a", "b", "c"])
+        self.assertRaises(TypeError, DiscreteVariable.add_value, 42)
```
lanzagar: This was probably intended to be `self.assertRaises(TypeError, a.add_value, 42)`

jerneju:
https://sentry.io/biolab/orange3/issues/242009555/

PR #2278, opened 2017-05-05
[FIX]Silhouette Plot: elide hover labels if labels are long

### LINKED ISSUE #2162: Silhouette Plot: elide hover labels if labels longer than n characters
##### Orange version

3.5.dev

##### Expected behavior
Labels are legible.

##### Actual behavior
Labels, especially if coming from Corpus, extend across the entire screen.

Corpus - Silhouette Plot. Hover on the plot.

Labels should be truncated to a sensible length.

jerneju:
##### Issue
Fixes #2162

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owsilhouetteplot.py
+++ b/Orange/widgets/visualize/owsilhouetteplot.py
@@ -35,6 +35,9 @@ from Orange.widgets.unsupervised.owhierarchicalclustering import \
 from Orange.widgets.widget import Msg
 
 
+ROW_NAMES_WIDTH = 200
+
+
 class OWSilhouettePlot(widget.OWWidget):
     name = "Silhouette Plot"
     description = "Visually assess cluster quality and " \
@@ -507,7 +510,10 @@ class SilhouettePlot(QGraphicsWidget):
             item = layout.itemAt(i + 1, 3)
 
             if grp.rownames is not None:
-                item.setItems(grp.rownames)
+                metrics = QFontMetrics(self.font())
+                rownames = [metrics.elidedText(rowname, Qt.ElideRight, ROW_NAMES_WIDTH)
+                            for rowname in grp.rownames]
+                item.setItems(rownames)
                 item.setVisible(self.__rowNamesVisible)
             else:
                 item.setItems([])
@@ -907,8 +913,7 @@ class BarPlotItem(QGraphicsWidget):
         return super().event(event)
 
     def sizeHint(self, which, constraint=QSizeF()):
-        spacing = max(self.__spacing * (self.count() - 1), 0)
-        return QSizeF(300, self.__barsize * self.count() + spacing)
+        return QSizeF(300, (self.__barsize + self.__spacing) * self.count())
 
     def setPreferredBarSize(self, size):
         if self.__barsize != size:
@@ -917,11 +922,6 @@ class BarPlotItem(QGraphicsWidget):
 
     def spacing(self):
         return self.__spacing
-
-    def setSpacing(self, spacing):
-        if self.__spacing != spacing:
-            self.__spacing = spacing
-            self.updateGeometry()
 
     def setPen(self, pen):
         pen = QPen(pen)


Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -507,7 +504,10 @@ def setRowNames(self, names):
             item = layout.itemAt(i + 1, 3)
 
             if grp.rownames is not None:
-                item.setItems(grp.rownames)
+                rownames = [rowname[0:12] + "..." if
```
kernc: Use ellipsis symbol instead: `…`

jerneju: Done!

janezd:
Is this still WIP?

Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -507,7 +504,10 @@ def setRowNames(self, names):
             item = layout.itemAt(i + 1, 3)
 
             if grp.rownames is not None:
-                item.setItems(grp.rownames)
+                rownames = [rowname[0:12] + "…" if
```
janezd: `rowname[:12]` instead of `rownames[0:12]`

janezd: However, this is not the proper way to do it since it disregards the actual widths of characters. See if you can truncate the text with http://doc.qt.io/qt-4.8/qfontmetrics.html#elidedText instead. For this, you will have to decide for (instead of determining) the sensible maximal label width in advance.

Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -507,7 +504,10 @@ def setRowNames(self, names):
...
+                            len(rowname) > 15 else rowname
+                            for rowname in grp.rownames]
+                item.setItems(rownames)
```
janezd: Can you put the untruncated (or less truncated) label in the tooltip?

Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -1047,8 +1047,8 @@ def __naturalsh(self):
         fm = QFontMetrics(self.font())
         spacing = self.__spacing
         N = len(self.__items)
-        width = max((fm.width(text) for text in self.__items),
-                    default=0)
+        width = 1.5 * max((fm.width(text) for text in self.__items),
```
janezd: Why 1.5?

Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -1111,8 +1111,10 @@ def remove(items, scene):
         self.__textitems = []
 
 
-def main(argv=sys.argv):
+def main(argv=None):
```
janezd: I'd just skip the argument and always give an empty list to `QApplication`. Nobody calls this `main` except for this module itself.

Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -494,6 +494,7 @@ def setScores(self, scores, labels, values, colors, rownames=None):
         self.__setup()
 
     def setRowNames(self, names):
+        ROW_NAMES_WIDTH = 200
```
janezd: Introducing a constant is OK since you explain what `200` is. But having it here is almost as if it just above the line where it's used. What about moving this constant to module- or class-level (https://www.python.org/dev/peps/pep-0008/#constants)?

Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -907,8 +911,8 @@ def event(self, event):
         return super().event(event)
 
     def sizeHint(self, which, constraint=QSizeF()):
-        spacing = max(self.__spacing * (self.count() - 1), 0)
-        return QSizeF(300, self.__barsize * self.count() + spacing)
+        spacing = max(self.__spacing, 0)
```
janezd: Why `max(self.__spacing, 0)`? Can `self.__spacing` be negative? I suspect this code was here because of `(self.count() - 1)`.

jerneju:
@janezd 
_cyber-security-breaches.tab_

Orange/widgets/visualize/owsilhouetteplot.py
kernc: So it's better to have code removed further away from where it's used, in this case in one single, particular place?

The ridicule is rhetorical.

janezd: There are some widgets that have all their settings (like widths, various colors...) collected as constants at the top, which is convenient when you want to improve its design. Moving a single constant is the first step in this direction. :)

PR #2281, opened 2017-05-05
[ENH] OWManifold: support t-SNE on sparse

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/2158

##### Description of changes
Support sparse with method=t-SNE and tweaked params.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


ajdapretnar:
This is good, but I get errors upon selecting Jaccard, Chebyshev and Mahalanobis distances. Could we:
- disable/gray out unavailable methods for sparse (not so urgent since we have the error message, but would be nice)
- disable error-inducing distance measures (quite urgent)

Everything else seems to work fine.

kernc:
Now errors on invalid distance.

@astaric How do I accommodate radon?

Issue #2189, opened 2017-04-04
Classifiers Shouldn't Store Whole Data Set?


nikicc:
##### Orange version

3.5.dev

##### Expected behavior
When one save a classifier I would expect that whole data set is not saved along into the pickle file.

##### Actual behavior
Orange saves whole data set into classifiers's attribute `original_data`.

##### Issues
Saving whole data set becomes problematic on larger data sets. For example, say you want to train logistic regression on bag of words data, save a classifier and apply it on new data set. Since whole training set is saved and loaded along with the classifier once quickly runs into memory issues.

##### Widgets affected
If we decide to remove `original_data` from the classifiers, Nomogram should be adopted to work without it.

nikicc:
#2282 solved the problem for storing & reloading classifiers, hence I'm closing the issue.

PR #2282, opened 2017-05-05
Model: Do not pickle stored data

lanzagar:
##### Issue

Models store a pointer to the original training set. This can be huge and causes problems when pickling/unpickling.

##### Description of changes
Skip `original_data` when pickling `Model`s.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


lanzagar:
This is the first solution to #2189. We still need to consider not storing data sets with models at all (only nomogram and tree viewer use that).

Issue #2167, opened 2017-03-30
Speed up distributions on sparse


ajdapretnar:
##### Orange version

3.5.dev

##### Expected behavior
Distributions should be computed faster.

##### Actual behavior
Distributions are painfully slow.

##### Steps to reproduce the behavior
Corpus - Bag of Words - Select Rows (or Impute or Naive Bayes).

markotoplak:
Fixed in #2283

PR #2284, opened 2017-05-05
[WIP] compute_contigency: convert X to CSC

markotoplak:
Speedups if a user successively computes contingencies for single columns. See #2283

Fixes https://github.com/biolab/orange3/issues/2167.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1465,6 +1465,8 @@ class Table(MutableSequence, Storage):
             arr_indi = [e for e, ind in enumerate(col_indi) if f_cond(ind)]
 
             vars = [(e, f_ind(col_indi[e]), col_desc[e]) for e in arr_indi]
+            if vars and arr is self.X and sp.issparse(self.X):
+                self.X = self.X.tocsc()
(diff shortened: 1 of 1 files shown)


Orange/data/table.py
```diff
@@ -1465,6 +1465,8 @@ def _compute_contingency(self, col_vars=None, row_var=None):
             arr_indi = [e for e, ind in enumerate(col_indi) if f_cond(ind)]
 
             vars = [(e, f_ind(col_indi[e]), col_desc[e]) for e in arr_indi]
+            if vars and arr is self.X and sp.issparse(self.X):
+                self.X = self.X.tocsc()
```
lanzagar: This changes `self.X`, but `arr` still points to the old object right?
Could probably be solved by moving the `if` from L1462 after this and changing to `arr = self.X = self.X.tocsc()`

markotoplak: I noticed further problems. If there are unknowns, there is also row selection in L1463, which does not work fast with csc.

astaric:
@markotoplak should this be finished and merged or closed?

PR #2285, opened 2017-05-05
Prevent changing outputs to type Table

### LINKED ISSUE #2169: Prevent transforming Corpus to Table in widgets
##### Orange version

3.5.dev

##### Expected behavior
If a widget receives Corpus on the input, it should send Corpus to the output.

##### Actual behavior
Widgets transform Corpus to Table, making the output useless in native Text widgets.

Corpus - Data Sampler - Word Cloud. (cannot connect to Word Cloud)

**Important!** Make sure preprocessing and other Corpus attributes (e.g. corpus.tokens, corpus.pos_tags) are retained!

janezd:
##### Issue

Fixes #2169.

##### Description of changes

- `Table` has  a new method `transform`, which provides a nicer way to transform a table into a table with a different domain but with the same type, ids and attributes.
- In widgets, calls to `Table.from_table(domain, data)` are changed to `data.transform(domain)`.
- The alternative, `data.from_table(domain, data)` is kept in places where the call has an additional argument, row indices, which is not supported (and won't be supported, although it would be trivial) by `transform`.
- Calls `data.from_table(domain, data)` are simplified to `data.transform(domain)`.

The two widgets that are not fixed are Concatenate and Venn diagram. The latter is similar to Concatenate, so I leave it to @astaric.

Widget's outputs are not `Dynamic`. We have to decide whether to make them such by default.
 
**e19e3f5ab23113d6ac6232ad52ccfde624637527 must be excluded before merging**

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -395,6 +395,31 @@ class Table(MutableSequence, Storage):
                 if new_cache:
                     _conversion_cache = None
 
+    def transform(self, domain):
+        """
+        Construct a table with a different domain.
+
+        The new table keeps the row ids and other information. If the table
+        is a subclass of :obj:`Table`, the resulting table will be of the same
+        type.
+
+        In a typical scenario, an existing table is augmented with a new
+        column by ::
+
+            domain = Domain(old_domain.attributes + [new_attribute],
+                            old_domain.class_vars,
+                            old_domain.metas)
+            table = data.transform(domain)
+            table[:, new_attribute] = new_column
+
+        Args:
+            domain (Domain): new domain
+
+        Returns:
+            A new table
+        """
+        return type(self).from_table(domain, self)
+
     @classmethod
     def from_table_rows(cls, source, row_indices):
         """
--- a/Orange/widgets/data/owoutliers.py
+++ b/Orange/widgets/data/owoutliers.py
@@ -153,9 +153,8 @@ class OWOutliers(widget.OWWidget):
                 else:
                     inliers_ind = np.where(y_pred == 1)[0]
                     outliers_ind = np.where(y_pred == -1)[0]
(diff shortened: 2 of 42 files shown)


Orange/widgets/data/owoutliers.py
```diff
@@ -153,9 +153,8 @@ def commit(self):
                 else:
                     inliers_ind = np.where(y_pred == 1)[0]
                     outliers_ind = np.where(y_pred == -1)[0]
-                    inliers = Table(self.new_domain, self.new_data, inliers_ind)
-                    outliers = Table(self.new_domain,
-                                     self.new_data, outliers_ind)
+                    inliers = self.new_data[inliers_ind]
+                    outliers = self.new_data[outliers_ind]
```
janezd: `new_data` is already in `new_domain`, so we just select the rows here.

Orange/data/table.py
```diff
@@ -398,20 +398,20 @@ def match_type(x):
...
-        
+
         The new table keeps the row ids and other information. If the table
         is a subclass of :obj:`Table`, the resulting table will be of the same
         type.
-        
+
         In a typical scenario, an existing table is augmented with a new
-        column by :: 
-
```
kernc: Fixup this commit into the commit that introduced the changes https://github.com/biolab/orange3/pull/2285/commits/e0488978bf6668fc7a113297037ca62df3dbb96b.

Set editor to Editor>General: Other: Strip trailing on Save = Modified; Always keep trailing on carrot = no.

lanzagar:
Looks good to me after reviewing changes (only tried and tested a few things).
Can merge after the last commit is removed...

janezd:
The last commit is removed.

PR #2286, opened 2017-05-05
Sparse concatenate

### LINKED ISSUE #2154: Concatenate should work with sparse
##### Orange version

3.5.dev

##### Expected behavior
Concatenate works with sparse tables.

##### Actual behavior
Concatenate fails on sparse.

Connect Corpus to Concatenate.

astaric:
##### Issue
Add support for sparse tables to Table.concatenate. Fixes #2154.

##### Description of changes
Modify Table.extend to use vstack to concatenate compatible table, use scipy.sparse.vstack if some of the arrays are sparse.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -18,7 +18,7 @@ from Orange.data import (
     Domain, Variable, Storage, StringVariable, Unknown, Value, Instance,
     ContinuousVariable, DiscreteVariable, MISSING_VALUES
 )
-from Orange.data.util import SharedComputeValue
+from Orange.data.util import SharedComputeValue, vstack, hstack
 from Orange.statistics.util import bincount, countnans, contingency, stats as fast_stats
 from Orange.util import flatten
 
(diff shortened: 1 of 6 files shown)


astaric:
@lanzagar the [(randomly) failing test](https://ci.appveyor.com/project/astaric/orange3-75xll/build/1.0.1741/job/826utu9twsdygpwx#L6021) has nothing to do with this PR. I have restarted the build, hopefully, it will pass this time :)

I am tagging you since you have merged the latest changes to the mosaic (#2133). Can you take a look?

nikicc:
I tested it on text and it doesn't crash any more. Though, the same strange behaviour as described in #2304 happens on BoW and prevents to properly test.

astaric:
Union and intersection are computed according to the "sameness" of the variable, not equality. Since BoW features have compute_value set, variables from different BoWs are considered different, since there is no way to tell if both compute values are the same.

If we modify the widget to only compare type, name, (and values for discrete variables), what do we set for compute_value on the output?

Orange/data/table.py
```diff
@@ -796,17 +796,17 @@ def __setitem__(self, key, value):
...
                 value = Unknown
 
-            if not isinstance(value, Real) and \
+            if not isinstance(value, (Real, np.ndarray)) and \
                     (len(attr_cols) or len(class_cols)):
                 raise TypeError(
                     "Ordinary attributes can only have primitive values")
             if len(attr_cols):
-                if len(attr_cols) == 1:
+                if sp.issparse(self.X) and len(attr_cols) == 1:
```
nikicc: Why do we need this? Scipy seems to be able to assign 2d arrays to columns regardless if `attr_cols = 2` or `attr_cols = [2]`. Is this here just for efficiency?
```py
>>> import numpy as np
>>> import scipy.sparse as sp
>>> x = sp.csr_matrix(np.eye(2))
>>> x.toarray()
array([[ 1.,  0.],
       [ 0.,  1.]])
>>> x[:, [1]] = np.array([[2], [2]])
/Users/Niko/anaconda/envs/orange3/lib/python3.5/site-packages/scipy/sparse/compressed.py:774: SparseEfficiencyWarning: Changing the sparsity structure of a csr_matrix is expensive. lil_matrix is more efficient.
  SparseEfficiencyWarning)
>>> x.toarray()
array([[ 1.,  2.],
       [ 0.,  2.]])
>>> x[:, 1] = np.array([[3], [3]])
>>> x.toarray()
array([[ 1.,  3.],
       [ 0.,  3.]])
```

Orange/widgets/utils/annotated_data.py
```diff
@@ -6,6 +6,30 @@
 ANNOTATED_DATA_FEATURE_NAME = "Selected"
 
 
+def add_columns(domain, attributes=(), class_vars=(), metas=()):
```
nikicc: Should this maybe be put in `domain.py`?

Orange/data/table.py
astaric: The original check has been here for a long time :)

I have removed the if completely.

PR #2287, opened 2017-05-05
codeclimate: relax radon check

astaric:
##### Issue
codeclimate is stricter when checking radon complexity that our existing script

##### Description of changes
Allow both A and B methods in radon check.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -8,6 +8,8 @@ engines:
    enabled: false
  radon:
    enabled: true
+   config:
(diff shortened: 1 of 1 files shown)


kernc:
Enable also eslint, shellcheck, csslint?

astaric:
I have tried, but the addition of those checks introduces about 300 new issues no one is going to fix in the near future. If someone plans to work on those, he/she can enable the engine locally, fix at least some of the existing issues and then propose the change.

PR #2288, opened 2017-05-05
[ENH, PERF] Auto n_jobs=-1 in widgets

kernc:
##### Issue
Slow computation on larger datasets, e.g. MDS.

##### Description of changes
Have algos in widgets use all CPU cores by passing `n_jobs=-1`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -349,6 +349,11 @@ class SklLearner(_ReprableWithParams, Learner, metaclass=WrapperMeta):
 
     def fit(self, X, Y, W=None):
         clf = self.__wraps__(**self.params)
+
+        # If problem "small enough", running it on a single core is probably faster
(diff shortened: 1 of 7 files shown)


thocevar:
Why is there the override for small problems? To keep the system more responsive or is the slowdown that big? It's hard to define the size of the problem for various algorithms in one way. E.g. a significant factor for RandomForest is the number of trees. "Small" datasets with a lot of trees will be executed on a single core and might take a long time.

kernc:
If the problem is _small_, and the number of trees _contained_, the single-threadedness should not play a huge role, give or shave a second.
Regardless, I can no longer stand by this, unaware how well it plays with the new concurrence pattern (and different OS configurations in general).

PR #2290, opened 2017-05-05
Make all signals dynamic by default

janezd:
##### Issue

Following #2285, we would need to make almost all outputs of type `Table` dynamic to allow passing instances of `Corpus` (and potential future other subclasses of `Table`). There seem to be no reason not to make all signals dynamic by default.

##### Description of changes

Flag `Dynamic` is on by default and declared as deprecated. A new flag `Strict` can be used to prevent dynamic signals.

@ales-erjavec, have I overlooked something important?

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/canvas/registry/description.py
+++ b/Orange/canvas/registry/description.py
@@ -9,6 +9,7 @@ import copy
 import warnings
 
 # Exceptions
+from Orange.util import OrangeDeprecationWarning
 
 
 class DescriptionError(Exception):
@@ -44,9 +45,11 @@ NonDefault = 16
 # Explicit - only connected if specifically requested or the only possibility
 Explicit = 32
 
-# Dynamic type output signal
+# (Deprecated) Dynamic type output signal
 Dynamic = 64
 
+# This output can only be connected to inputs of the same class or superclass
+Strict = 128
 
 # Input/output signal (channel) description
 
@@ -156,10 +159,13 @@ class OutputSignal(object):
         if not (flags & Default or flags & NonDefault):
             flags += NonDefault
 
+        if flags & Dynamic:
+            warnings.warn("all outputs are dynamic; flag Dynamic is deprecated",
+                          OrangeDeprecationWarning)
         self.single = flags & Single
         self.default = flags & Default
         self.explicit = flags & Explicit
-        self.dynamic = flags & Dynamic
+        self.dynamic = not (flags & Strict)
         self.flags = flags
 
         if self.dynamic and not self.single:
--- a/Orange/tests/test_table.py
+++ b/Orange/tests/test_table.py
@@ -1832,7 +1832,7 @@ class CreateTableWithDomainAndTable(TableTests):
         new_domain = data.domain.Domain([], iris.domain.class_vars,
                                         iris.domain.attributes, source=iris.domain)
         new_iris = data.Table.from_table(new_domain, iris)
-        
+
         self.assertTrue(sp.issparse(new_iris.X))
         self.assertTrue(sp.issparse(new_iris.metas))
         self.assertEqual(new_iris.X.shape, (len(iris), 0))
--- a/Orange/widgets/widget.py
+++ b/Orange/widgets/widget.py
@@ -814,12 +814,17 @@ Multiple = widget_description.Multiple
 #: Only connected if specifically requested (in a dedicated "Links" dialog)
 #: or it is the only possible connection.
 Explicit = widget_description.Explicit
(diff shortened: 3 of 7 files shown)


Orange/canvas/registry/description.py
```diff
@@ -156,10 +159,13 @@ def __init__(self, name, type, flags=Single + NonDefault,
         if not (flags & Default or flags & NonDefault):
             flags += NonDefault
 
+        if flags & Dynamic:
+            warnings.warn("all outputs are dynamic; flag Dynamic is deprecated",
+                          OrangeDeprecationWarning)
```
kernc: Add `stacklevel=n` so we know where the warning is from?

Orange/canvas/registry/description.py
```diff
@@ -156,10 +159,13 @@ def __init__(self, name, type, flags=Single + NonDefault,
...
         self.single = flags & Single
         self.default = flags & Default
         self.explicit = flags & Explicit
-        self.dynamic = flags & Dynamic
+        self.dynamic = not flags & Strict
```
kernc: In formal logic, negation binds closer. Avoid doubt with parenthesis.

Orange/tests/test_table.py
```diff
@@ -619,6 +619,7 @@ def test_extend(self):
             self.assertEqual(d[i], d[-5 + i])
 
         x = d[:5]
+        d.extend(x)
```
kernc: This?

ales-erjavec:
1. Change the VERSION_HEX in [Orange/canvas/registry/base.py](https://github.com/biolab/orange3/blob/master/Orange/canvas/registry/base.py#L18) to invalidate existing registry caches.

2. Hopefully nobody still uses the 'Multiple' flag on the output, because this will break it (i am not even sure it still works).

Orange/tests/test_table.py
janezd: Shame on me.

Orange/canvas/registry/description.py
janezd: You're right. Done.

Orange/canvas/registry/description.py
janezd: `stacklevel` wouldn't help. This code is called from widget discovery and it doesn't know to which widget it refers, nor is this information visible from the stack trace without observing local variables a few levels higher.

It doesn't matter much, though: our tests treat these warnings as errors, hence if we add a new widget with dynamic signals, we'll know what caused it.

janezd:
@ales-erjavec, I searched for `Multiple` and found it only in inputs. Should I change `self.dynamic = not (flags & Strict)` to `self.dynamic = not (flags & (Strict | Multiple))`? Or would it be better to treat all outputs as `Single` and raise a warning if they are specified as `Multiple`?

janezd:
Actually, I see that the check that the signal is not dynamic and multiple at the same time is already there. So this should be safe.

doc/development/source/widget.rst
```diff
@@ -106,13 +106,18 @@ Input/Output flags:
    Multiple signal (more then one input on the channel). Input with this
    flag receive a second parameter `id`.
 
-.. attribute:: Dynamic
+.. attribute:: Strict
 
-   Only applies to output. Specifies that the instances on the output
-   will in general be subtypes of the declared type and that the output
+   Only applies to output. Specifies that this output can be connected only
+   to inputs that expect this type of its supertypes. Otherwise, the output
```
astaric: or*

Orange/widgets/widget.py
```diff
@@ -814,12 +814,17 @@ def __init__(self, text, persistent_id, icon=None, moreurl=None):
...
 Explicit = widget_description.Explicit
-#: Dynamic output type.
+#: Dynamic output type (obsolete and ignored).
 #: Specifies that the instances on the output will in general be
 #: subtypes of the declared type and that the output can be connected
 #: to any input signal which can accept a subtype of the declared output
 #: type.
 Dynamic = widget_description.Dynamic
+#: Only applies to output. Specifies that this output can be connected only
+#: to inputs that expect this type of its supertypes. Otherwise, the output
```
astaric: type **or** its supertypes. **When not set**, the output can be connected...

janezd: Change to:

> Without explicitly setting this flag, the output can also be connected to any input which accepts a subtype of the declared output type.

janezd:
Closed in favour of #2312.

PR #2292, opened 2017-05-05
[RFC] Signals as objects

janezd:
This is something I wanted to try for a long time. After doing it I'm no longer so sure, but I'd like to see your comments (@ales-erjavec, @astaric, @lanzagar, @kernc).

1. I like the new classes for Error/Warning/Info messages that treat messages as objects that can be activated/deactivated.
2. I don't like using string literals.
3. I like the new-style PyQt's signals: you emit a signal by calling emit on a signal object.

First, see the changes I made to [owtestlearners.py](https://github.com/biolab/orange3/pull/2292/files#diff-bfe1378d373a089b1906fa7c0e28e9a8).

**Output signals** can be declared and emitted in a similar fashion as PyQt's signals, like

    predictions_signal = OutputSignal("Predictions", Table)
    evalutions_results_signal = OutputSignal("Evaluation Results", Results)

and emitted by

    self.predictions_signal.send(predictions)

We can no longer make a mistake of emitting a signal with a wrong name. On the other hand, this hasn't been much of a problem lately. Disadvantage: I haven't been able to come up with good names for objects representing output signals. `evalutions_results_signal` is not too convincing.

**Input signals** are declared by decorating handlers, like in

     @InputSignal("Data", Table, flags=widget.Default)

This is convenient since we don't have to write handlers' names (as string literals); any errors doing this were however already catched in the widget's metaclass. Disadvantage: inputs are not longer listed out in the beginning. Advantage: the sole reason for looking at those lists was to find handler names (and then searching for the corresponding method). Now you can just search for "InputSignal".

Implementation is not complicated. A bit of the work is done in the meta class, which catches the declaration, and part is in the signal mixin's `__init__`, which binds signals to widgets. Two widget's methods are moved into a mixin, which is good. Some other code from widget.py is moved, too. Moving classes `InputSignal` and `OutputSignal` from canvas registry modules to widgets requires that canvas imports some modules from widgets. If this is undesirable, they can be moved back. A more likely direction, though, will be to move the widget base class into canvas, IMHO.

The change is compatible with the existing signal declaration; it in fact moves the new-style declaration into old lists internally.

Feel free to criticize at will or reject. :) Or propose improvements.

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -23,10 +23,10 @@ from Orange.preprocess.preprocess import Preprocess
 from Orange.preprocess import RemoveNaNClasses
 from Orange.widgets import widget, gui, settings
 from Orange.widgets.utils.itemmodels import DomainModel
-from Orange.widgets.widget import OWWidget, Msg
+from Orange.widgets.widget import OWWidget, Msg, Input, Output
 
-Input = namedtuple(
-    "Input",
+InputLearner = namedtuple(
+    "InputLearner",
     ["learner",  # :: Orange.base.Learner
      "results",  # :: Option[Try[Orange.evaluation.Results]]
      "stats"]    # :: Option[Sequence[Try[float]]]
@@ -130,13 +130,16 @@ class OWTestLearners(OWWidget):
     icon = "icons/TestLearners1.svg"
     priority = 100
 
-    inputs = [("Learner", Learner, "set_learner", widget.Multiple),
-              ("Data", Table, "set_train_data", widget.Default),
-              ("Test Data", Table, "set_test_data"),
-              ("Preprocessor", Preprocess, "set_preprocessor")]
+    class Inputs:
+        train_data = Input("Data", Table, flags=widget.Default)
+        test_data = Input("Test Data", Table)
+        learner = Input("Learner", Learner, flags=widget.Multiple)
+        preprocessor = Input("Preprocessor", Preprocess)
 
-    outputs = [("Predictions", Table),
-               ("Evaluation Results", Results)]
+    class Outputs:
+        predictions = Output("Predictions", Table)
+        evalutions_results = Output("Evaluation Results", Results)
+
 
     settings_version = 3
     settingsHandler = settings.PerfectDomainContextHandler(metas_in_res=True)
@@ -293,6 +296,7 @@ class OWTestLearners(OWWidget):
         if self.resampling == OWTestLearners.FeatureFold and not enabled:
             self.resampling = OWTestLearners.KFold
 
+    @Inputs.learner
     def set_learner(self, learner, key):
         """
         Set the input `learner` for `key`.
@@ -301,9 +305,10 @@ class OWTestLearners(OWWidget):
             # Removed
             del self.learners[key]
         else:
-            self.learners[key] = Input(learner, None, None)
+            self.learners[key] = InputLearner(learner, None, None)
             self._invalidate([key])
 
+    @Inputs.train_data
     def set_train_data(self, data):
         """
         Set the input training dataset.
@@ -351,6 +356,7 @@ class OWTestLearners(OWWidget):
                 self.resampling = OWTestLearners.FeatureFold
         self._invalidate()
 
+    @Inputs.test_data
     def set_test_data(self, data):
         """
         Set the input separate testing dataset.
@@ -394,6 +400,7 @@ class OWTestLearners(OWWidget):
                 (False, True): " test "}[(self.train_data_missing_vals,
                                           self.test_data_missing_vals)]
 
+    @Inputs.preprocessor
     def set_preprocessor(self, preproc):
         """
         Set the input preprocessor to apply on the training data.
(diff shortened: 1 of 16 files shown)


astaric:
I kind of like the Output definition (and usage). Should they be grouped in a class the way errors/warnings are?
Not so sure about inputs, I am not so fond of having the declarations scattered throughout the code.

I would drop the name Signal though. widgets have Inputs and Outputs.

janezd:
I like dropping `Signal`. I also don't like scattering input signal declarations throughout the code. There is no technical reason for grouping (for messages, groups are used to distinguish between message types), but grouping would solve the problem of naming and improve readability.

Following all your suggestions (see, I'm in a politician mode now :)) results in

```
class OWTestLearners:
    ...
    class Inputs:
        train_data = Input("Data", Table, flags=widget.Default)
        test_data = Input("Test Data", Table)
        learner = Input("Learner", Learner, flags=widget.Multiple)
        preprocessor = Input("Preprocessor", Preprocess)

    class Outputs:
        predictions = Output("Predictions", Table)
        evaluations_results = Output("Evaluation Results", Results)
```

and then declaring input handlers with

    @Inputs.train_data

and sending outputs with

    Outputs.predictions.send(predictions)

Also, implementation is cleaner than before, with less magic in the meta class. Meta class would include a sanity check that every `Input` is used as a decorator exactly once. Meta class would also translate the existing lists `inputs` and `outputs` to these classes instead of the other way around, as it does in my current prototype.

I like really it now.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -23,10 +23,10 @@
 from Orange.preprocess import RemoveNaNClasses
 from Orange.widgets import widget, gui, settings
 from Orange.widgets.utils.itemmodels import DomainModel
-from Orange.widgets.widget import OWWidget, Msg
+from Orange.widgets.widget import OWWidget, Msg, Input, Output
 
-Input = namedtuple(
-    "Input",
+InputLearner = namedtuple(
```
janezd: Unrelated to the change --- needed just to avoid a name collision with the new class `Input`.

janezd:
Implemented as described above. See https://github.com/biolab/orange3/pull/2292/files#diff-bfe1378d373a089b1906fa7c0e28e9a8 and https://github.com/biolab/orange3/pull/2292/files#diff-a2cbec3457e61cdb3228e0e4b7b95585.

Metamagic is now needed just for compatibility and sanity checks.

ales-erjavec:
>Moving classes `InputSignal` and `OutputSignal` from canvas registry modules to widgets requires that canvas imports some modules from widgets. If this is undesirable, they can be moved back.

This is undesirable. In fact it is unfortunate that they got out of canvas (were imported from the widget code) in the first place. They were intended as simple capability descriptions for the canvas and independent of any concrete 'widget' implementation.

Basically the way this should work is: OWWidget declares IO/meta info (in whatever way it feels like this year)-> Widget Discovery reads/parses this and creates the appropriate abstract capability descriptions that are then used by the rest of the canvas.

janezd:
I realized that signals were meant as descriptors, but I thought upgrading them to something more was a good idea. If you don't agree, I will trust your judgment.

My idea was to make signals more like error messages in widgets. That is, I like the idea that signal descriptors wouldn't be used just by the canvas but also by the widget itself to describe, say, which signal is sent by the `send` method.

Three options.

**First.**
- I remove the `send` method from signals, but widgets send signals with `self.send(self.Outputs.annotated_data, data)`? This requires just a small change in the widget's `send` method and can easily be backwards compatible. But this call is more cumbersome than what we have now and what I proposed in the PR. Not nice.
- Input handlers are marked by decorators, as in my current proposal.
- Inputs and outputs are declared with a class, like in this PR.

With this, signal descriptors can be moved back to canvas, and they have only `__init__` and `__call__`; `__call__` is only needed for the decorator, to store the name of the handler.

**Second.** Like above, but signals have the `send` method to make the call shorter. This means that signals must be bound to some objects, but these objects do not have to be of any specific type. Essentially, this is like my current proposal except that the code is moved back to canvas and that they are no longer aware of widgets. In this way, widgets and canvas are still separate, and signals descriptors are the common representation of signals that is shared between widgets and canvas. Canvas does not import or assume (too) much about implementation of widgets.

I understand that signal description were intended to be just passive descriptions, but is there a reason to make them a bit more active?

**Third:** we just close this pull request. You wrote "*OWWidget declares IO/meta info (in whatever way it feels like this year)*". I think that widget error messages, which we reimplemented in a similar fashion, were changed for the better, but some other changes are fluctuating back and forth, and this could be one of them.

astaric:
**Four** We leave InputSignal and OutputSignal in canvas and create *new* Input and Output classes in widget as they are in this PR. No backward compatibility though, a widget can either have old-style signals and use self.send or new-style signals and use those. No magic in meta class either.

Support for converting new-style signals is added to WidgetDescription.from_module, or that can call some predefined method on widget[mixin] that produces the appropriate description for widget?

janezd:
I like it in principle, but the new classes would contain mostly the same data (a large subset) as the existing, so we'd duplicate most of it. Magic can be easily moved to `from_module`.

janezd:
Not accepted in its current form, but may be reimplemented in #2312.

PR #2295, opened 2017-05-08
Impute: remove nan values in sparse

### LINKED ISSUE #2178: Remove Nans from Sparse Data
##### Orange version

3.5.dev

##### Expected behavior

Logistic Regression works on sparse data that contains nan values.

##### Actual behavior
It crashes with `Input contains Nans, ...`

Load `Election-2016-Tweets.tab` in Corpus. Use just first 100 rows for speed. Pass corpus through bag of words into test & score and use 10-fold cross validation. Add logistic regression on input and test & score crashes.

jerneju:
##### Issue
Fixes #2178.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/impute.py
+++ b/Orange/preprocess/impute.py
@@ -4,7 +4,7 @@ from scipy.sparse import issparse
 import Orange.data
 from Orange.statistics import distribution, basic_stats
 from Orange.util import Reprable
-from .transformation import Transformation, Lookup
+from .transformation import Transformation, Lookup as BaseLookup
 
 __all__ = ["ReplaceUnknowns", "Average", "DoNotImpute", "DropInstances",
            "Model", "AsValue", "Random", "Default"]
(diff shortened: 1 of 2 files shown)


Orange/preprocess/impute.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, variable, value=0):
 
     def transform(self, c):
         if issparse(c):     # sparse does not have unknown values
```
nikicc: Can you please remove the comment `# sparse does not have unknown values` that doesn't apply any more?

Orange/tests/test_impute.py
```diff
@@ -30,6 +30,18 @@ def test_sparse(self):
         rm = preprocess.ReplaceUnknowns(None, value=42).transform(m)
         self.assertEqual((m!=rm).nnz, 0)
 
+    def test_sparse_nans(self):
+        """
+        Remove nans from sparse matrix.
+        GH-2295
```
nikicc: Why reference to the PR in which the code is introduced?

jerneju: I always write a reference to a PR.

Issue #2163, opened 2017-03-30
Paint Data should work on sparse


ajdapretnar:
##### Orange version

3.5.dev

##### Expected behavior
Paint Data works on sparse (at least doesn't fail).

##### Actual behavior
Paint Data fails on sparse.

##### Steps to reproduce the behavior
Corpus - Bag of Words - Paint Data.

janezd:
I think 'doesn't fail' could suffice here. Or do you have a reasonable example of sparse data for Paint?

ajdapretnar:
Couldn't think of any, really. Left it open for interpretation intentionally (perhaps for someone with a wilder imagination).
But in essence, yes, **should not fail on sparse**.

nikicc:
Fixed by #2298.

PR #2298, opened 2017-05-08
Paint Data: Show error msg when data is sparse

jerneju:
##### Issue
#2163

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpaintdata.py
+++ b/Orange/widgets/data/owpaintdata.py
@@ -765,6 +765,7 @@ class OWPaintData(OWWidget):
     class Warning(OWWidget.Warning):
         no_input_variables = Msg("Input data has no variables")
         continuous_target = Msg("Continuous target value can not be used.")
+        sparse_not_supported = Msg("Sparse data is ignored.")
 
     class Information(OWWidget.Information):
         use_first_two = \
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owpaintdata.py
```diff
@@ -979,9 +982,11 @@ def _check_and_set_data(data):
                     data = None
                 elif len(data.domain.attributes) > 2:
                     self.Information.use_first_two()
+            if data and data.is_sparse():
```
nikicc: Currently, if I pass sparse data with more than two attributes, I see an error `Sparse data is not supported` as well as the info message that only `Paint Data uses data from the first two attributes.` The second message is irrelevant until sparse data is supported.

I suggest we move the the check for sparsity at the very beginning. Something like:
```
        def _check_and_set_data(data):
            self.clear_messages()
            if data and data.is_sparse():
                self.Error.sparse_not_supported()
                return False
            ....
```

Orange/widgets/data/owpaintdata.py
```diff
@@ -762,6 +762,9 @@ class OWPaintData(OWWidget):
 
     graph_name = "plot"
 
+    class Error(OWWidget.Error):
+        sparse_not_supported = Msg("Sparse data is not supported.")
```
nikicc: Also, this can probably be just Warning (e.g. `Sparse input data is ignored. `) since the widget can function without it.

Issue #2182, opened 2017-03-31
MDS doesn't store selection in Settings


ajdapretnar:
##### Orange version

3.5.dev

##### Expected behavior
MDS stores the selection and shows it upon reloading the workflow.

##### Actual behavior
MDS doesn't store the selection. Even worse, it it rerun upon reloading the workflow.

##### Steps to reproduce the behavior
File (iris) - MDS. Save the workflow and reopen it.

##### Additional info (worksheets, data, screenshots, ...)
MDS should be rerun with the same seed (checkbox for random, not sure about the options for PCA).
It should also save the selection.

PR #2301, opened 2017-05-09
[ENH] OWMDS, OWLinearProjection: save selection in workflow

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/2182

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -113,6 +113,8 @@ class OWMDS(OWWidget):
     output_embedding_role = settings.Setting(2)
     autocommit = settings.Setting(True)
 
+    selection_indices = settings.Setting(None, schema_only=True)
+
     color_value = settings.ContextSetting("")
     shape_value = settings.ContextSetting("")
     size_value = settings.ContextSetting("")
@@ -523,6 +525,7 @@ class OWMDS(OWWidget):
 
         self.update_controls()
         self.openContext(self.data)
+        self.select_indices(self.selection_indices)
 
     def _toggle_run(self):
         if self.__state == OWMDS.Running:
@@ -1013,8 +1016,9 @@ class OWMDS(OWWidget):
         if output is not None and self._selection_mask is not None and \
                 numpy.any(self._selection_mask):
             subset = output[self._selection_mask]
+            self.selection_indices = numpy.flatnonzero(self._selection_mask)
         else:
-            subset = None
+            self.selection_indices = subset = None
         self.Outputs.selected_data.send(subset)
         self.Outputs.annotated_data.send(create_annotated_table(output, self._selection_mask))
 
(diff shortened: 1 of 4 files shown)


astaric:
Selections are not really context settings as they depend not only on the columns in the dataset but on the number of rows as well. Have you considered using schema only settings?

kernc:
The problem I have with schema_only settings is that there's additional boilerplate associated with using them:
```py
    def closeContext(self):
        if self.current_context is not None:
            self.selection = None
        super().closeContext()
```
Short, but very intricate.

janezd:
File -> Sample -> MDS. Select all points in MDS. Decrease the sample size in Sample. MDS crashes due to selection indices out of range. Linear projection probably as well.

A saved context setting is reused when "useful", e.g. if the setting includes a variable, the variable has to appear in the domain for the context to be useful. In this sense, the context here is not the domain but the sample size (e.g. the saved selection is "useful" if there is enough data for these indices). However, this context would be too vaguely defined, so checking the domain (`DomainContextHandler`) is OK.

One quick solution would be to discard the selection (after `openContext`) if `max(selection_indices) >= len(data)`. A slightly better solution is to have `DomainAndDataSizeContextHandler`, but it can still lead to weird selections. The safest solution would be `DataHashContextHandler`, where the has is computed over `X`, `Y` and `metas`. The latter would also make sense for some other widgets.

kernc:
Right. How is this?

Orange/widgets/widget.py
```diff
@@ -508,6 +508,14 @@ def closeContext(self):
         reinitializing the user interface (e.g. combo boxes) with the new
         data.
         """
+        # Clear schema-only settings when *closing* the context
+        # FIXME: Sadly, schema-only settings aren't cleared when non-contextual
```
astaric: Actually, they are.

As non-context-aware widgets only load settings when a new instance of the widget is put on the canvas, this is achieved by not updating default values. See https://github.com/biolab/orange3/blob/master/Orange/widgets/settings.py#L508

Orange/widgets/widget.py
```diff
@@ -508,6 +508,14 @@ def closeContext(self):
...
+        # SettingsHandler is used
+        if self.current_context is not None:
```
astaric: With this change, https://github.com/biolab/orange3/blob/master/Orange/widgets/visualize/owscatterplot.py#L581 is now redundant.

Orange/widgets/widget.py
kernc: But what if the _same instance_ of widget but passed different data, as described [here](https://github.com/biolab/orange3/pull/2301/files#diff-2281ecd412ed7edd749aca4486d1f338R779)?

astaric:
This looks good to me, rebase when possible so we can merge.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -103,7 +103,7 @@ class Outputs:
 
     settingsHandler = settings.DomainContextHandler()
 
-    max_iter = settings.Setting(300)
+    max_iter = settings.Setting(10)
```
lanzagar: What is this change doing in a commit "save selection in workflow" ?!?

Did you change this while testing and accidentally committed it?
Or decided to change some defaults and make it less obvious by not mentioning it at all and/or putting it in its own commit.

Issue #2161, opened 2017-03-30
CN2 Rule Induction should work on sparse


ajdapretnar:
##### Orange version

3.5.dev

##### Expected behavior
CN2 Rule Induction works on sparse.

##### Actual behavior
CN2 Rule Induction fails on sparse.

##### Steps to reproduce the behavior
Corpus - Bag of Words - CN2 Rule Induction

nikicc:
Fixed by #2303.

PR #2303, opened 2017-05-09
CN2 Rule Induction: Show error message when data is sparse

rokgomiscek:
##### Issue
#2161

##### Description of changes
Added error message when data is sparse.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owrules.py
+++ b/Orange/widgets/model/owrules.py
@@ -218,6 +218,7 @@ class OWRuleLearner(OWBaseLearner):
     auto_apply = Setting(True)
 
     LEARNER = CustomRuleLearner
+    supports_sparse = False
 
     storage_orders = ["ordered", "unordered"]
     storage_covers = ["exclusive", "weighted"]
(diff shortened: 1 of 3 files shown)


nikicc:
Should we rather move this logic to `OWBaseLearner`? Perhaps we should have something like `supports_sparse_data` attribute on the base learner and error raising could be implemented there. Then `OWRuleLearner` would just have to set `supports_sparse_data = False`.

Orange/widgets/model/owrules.py
```diff
@@ -323,14 +323,18 @@ def update_model(self):
...
         if self.check_data():
-            with self.progressBar() as progress:
-                self.learner.set_progress_advance_callback(progress.advance)
-                self.model = self.learner(self.data)
-                self.learner.clear_progress_advance_callback()
-            self.model.name = self.learner_name
-            self.model.instances = self.data
-            self.valid_data = True
+            if self.data.is_sparse():
+                self.Error.sparse_not_supported()
```
kernc: The idea was to have Learner (subtypes) define `supports_sparse` such as they have (and possibly use) [`supports_multiclass` and `supports_weights`](https://github.com/kernc/orange3/blob/2b1398fb5b1b2e15bb83ef7eefa9212f8f336b1e/Orange/base.py#L78-L79). The learners that don't support sparse data then define `supports_sparse = False` (or vice versa), and the handling of state where sparse data is erroneously passed to a unsupporting learner widget is handled in a single place in `OWBaseLearner`.

PR #2308, opened 2017-05-11
Python Script: Check for types of output variables

jerneju:
##### Issue
Various errors reported by sentry. A user can write a code which sends data in incompatible or inappropriate variables. For instance it can send `String` instead of `Table` and that causes some other widget to crash.

##### Description of changes
Only proper variables are sent.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpythonscript.py
+++ b/Orange/widgets/data/owpythonscript.py
@@ -121,7 +121,7 @@ class PythonScriptEditor(QPlainTextEdit):
             text = self.lastLine()
             if text and not text.strip():
                 cursor = self.textCursor()
-                for i in range(min(self.INDENT, len(text))):
+                for _ in range(min(self.INDENT, len(text))):
                     cursor.deletePreviousChar()
             else:
                 super().keyPressEvent(event)
@@ -359,8 +359,8 @@ class OWPythonScript(widget.OWWidget):
 
     inputs = [("in_data", Orange.data.Table, "setExampleTable",
                widget.Default),
-#               ("in_distance", Orange.misc.SymMatrix, "setDistanceMatrix",
-#                widget.Default),
+              # ("in_distance", Orange.misc.SymMatrix, "setDistanceMatrix",
+              # widget.Default),
               ("in_learner", Learner, "setLearner",
                widget.Default),
               ("in_classifier", Model, "setClassifier",
@@ -368,7 +368,7 @@ class OWPythonScript(widget.OWWidget):
               ("in_object", object, "setObject")]
 
     outputs = [("out_data", Orange.data.Table, ),
-#                ("out_distance", Orange.misc.SymMatrix, ),
+               # ("out_distance", Orange.misc.SymMatrix, ),
                ("out_learner", Learner, ),
                ("out_classifier", Model, widget.Dynamic),
                ("out_object", object, widget.Dynamic)]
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/243929444/

jerneju:
https://sentry.io/biolab/orange3/issues/271062603/

Orange/widgets/data/owpythonscript.py
```diff
@@ -655,7 +660,11 @@ def commit(self):
         self.console.new_prompt(sys.ps1)
         for out in self.outputs:
             signal = out.name
-            self.send(signal, self.console.locals.get(signal, None))
+            out_var = self.console.locals.get(signal, None)
+            if not isinstance(out_var, out.type) and out_var is not None:
+                self.Error.invalid_type(signal, out.type)
+                out_var = None
```
janezd: The intention here is to output `None` and inform the user of that. The problem is that if there are multiple errors, the user will only be informed of the last one although all outputs are `None`.

One option is to list all errors in a single error messages, but this would become too long.

A better option is to have separate error messages. You can add them using `MessageGroup.add_message`. You can probably do the following in the class definition (that is, outside any method, in place of the existing declaration of class `Error`).

    class Error(OWWidget.Error):
        pass

    for varname, vartype in outputs:
        Error.add_message("Variable '{}' has to be of type
        invalid_type = Msg(varname, "Variable '{}' has to be of type '{}'.".format(varname, vartype.__name__)

and then raise them by something like

    getattr(self.Error, out_var)()

jerneju: Fixed.

Orange/widgets/data/owpythonscript.py
```diff
@@ -655,8 +660,15 @@ def commit(self):
         self.console.new_prompt(sys.ps1)
         for out in self.outputs:
             signal = out.name
-            self.send(signal, self.console.locals.get(signal, None))
-
+            out_var = self.console.locals.get(signal, None)
+            if not isinstance(out_var, out.type) and out_var is not None:
+                self.Error.add_message(signal, msg=self.error_msg(signal, out.type))
```
janezd: I'd prefer `out.type.__class__` here.

Replace "of type" with "an instance of". 

I don't see much sense in having a separate method for constructing `Msg`. I'd write

    self.Error.add_message(
        signal,
        "Variable '{}' has to be an instance of '{}'.".format(signal, out.type.__name__))

Orange/widgets/data/owpythonscript.py
```diff
@@ -655,8 +660,15 @@ def commit(self):
...
+                getattr(self.Error, signal)()
+                out_var = None
+            self.send(signal, out_var)
+
+    def error_msg(self, a, b):
```
janezd: I suggest removing this method. However, don't use `a` and `b` for names of arguments.

Orange/widgets/data/owpythonscript.py
```diff
@@ -655,8 +660,15 @@ def commit(self):
...
+        return Msg("Variable '{}' has to be of type '{}'.".format(a, b))
```
janezd: No need to construct a `Msg`. `add_message` accepts a string.

jerneju:
Fixed.

jerneju:
https://sentry.io/biolab/orange3/issues/288220340/

jerneju:
https://sentry.io/biolab/orange3/issues/244647726/

jerneju:
https://sentry.io/biolab/orange3/issues/292657971/

PR #2312, opened 2017-05-12
Move creation of widget description to widgets; Reimplement signal definition lists

janezd:
A different take on #2292 and #2290.

## Widget discovery

Widget discovery code used to search for classes with metatype `WidgetMetaClass`, and then it extracted information contained in class attributes into `WidgetDescription`. The canvas thus had to import `WidgetMetaClass` and dug into the widget's attributes.

In this PR, the widget must describe itself. Discovery searches for any class that implements `get_widget_description`, which returns a `WidgetDescription`. Signals are described with `InputSignal` and `OutputSignal`; the widget can represent them internally in any way it wants.

The idea here is that canvas knows less about widget implementation. The entire API would be (more or less) that canvas exposes descriptive classes (`WidgetDescription`, `InputSignal`, `OutputSignal`) and requires that objects driven by the canvas implement a single method (`get_widget_description`). Ideally, canvas would no longer import any widget modules or assume anything else about widgets. This would be a step towards entirely separating the canvas from widgets.

## New-style Signals

Signals are changed as already proposed in #2292 but without touching canvas. Signals can be declared as follows.

```
class OWTestLearners:
    ...
    class Inputs:
        train_data = Input("Data", Table, default=True)
        test_data = Input("Test Data", Table)
        learner = Input("Learner", Learner, multiple=True)
        preprocessor = Input("Preprocessor", Preprocess)

    class Outputs:
        predictions = Output("Predictions", Table)
        evaluations_results = Output("Evaluation Results", Results)
```

Input handlers are than marked with

    @Inputs.train_data

and outputs are sent by

    Outputs.predictions.send(predictions)

This is convenient since we don't have to write handlers' names (as string literals); any errors doing this were however already catched in the widget's metaclass. Disadvantage: inputs are not longer listed out in the beginning. Advantage: the sole reason for looking at those lists was to find handler names (and then searching for the corresponding method). Now you can just search for "InputSignal".

The code is compatible with the existing code.

##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation

### DIFF
--- a/Orange/widgets/utils/signals.py
+++ b/Orange/widgets/utils/signals.py
@@ -105,6 +105,7 @@ class Output(OutputSignal, _Signal):
                  default=False, explicit=False, dynamic=True):
         flags = self.get_flags(False, default, explicit, dynamic)
         super().__init__(name, type, flags, id, doc, replaces or [])
+        self.widget = None
 
     def bound_signal(self, widget):
         """
@@ -189,11 +190,17 @@ class WidgetSignalsMixin:
 
     @classmethod
     def _check_input_handlers(cls):
-        unbound = [signal.name for signal in cls.inputs
-                   if not signal.handler]
+        unbound = [signal.name for signal in cls.Inputs.__dict__.values()
+                   if isinstance(signal, Input) and not signal.handler]
         if unbound:
             raise ValueError("unbound signal(s) in {}: {}".
                              format(cls.__name__, ", ".join(unbound)))
+
+        missing_handlers = [signal.handler for signal in cls.inputs
+                            if not hasattr(cls, signal.handler)]
+        if missing_handlers:
+            raise ValueError("missing handlers in {}: {}".
+                             format(cls.__name__, ", ".join(missing_handlers)))
 
     @classmethod
     def get_signals(cls, direction):
--- a/Orange/widgets/utils/signals.py
+++ b/Orange/widgets/utils/signals.py
@@ -37,19 +37,25 @@ class Input(InputSignal, _Signal):
 
     Parameters
     ----------
-    name (str): signal name
-    type (type): signal type
-    id (str): a unique id of the signal
-    doc (str, optional): signal documentation
-    replaces (list of str): a list with names of signals replaced by this signal
-    multiple (bool, optional): if set, multiple signals can be connected
-        to this output (default: `False`)
-    default (bool, optional): when the widget accepts multiple signals of the
-        same type, one of them can set this flag to act as the default
+    name (str):
+        signal name
+    type (type):
+        signal type
+    id (str):
+        a unique id of the signal
+    doc (str, optional):
+        signal documentation
+    replaces (list of str):
+        a list with names of signals replaced by this signal
+    multiple (bool, optional):
+        if set, multiple signals can be connected to this output
         (default: `False`)
-    explicit (bool, optional): if set, this signal is only used when it is
-        the only option or when explicitly connected in the dialog
-        (default: `False`)
+    default (bool, optional):
+        when the widget accepts multiple signals of the same type, one of them
+        can set this flag to act as the default (default: `False`)
+    explicit (bool, optional):
+        if set, this signal is only used when it is the only option or when
+        explicitly connected in the dialog (default: `False`)
     """
     def __init__(self, name, type, id=None, doc=None, replaces=None, *,
                  multiple=False, default=False, explicit=False):
@@ -85,21 +91,27 @@ class Output(OutputSignal, _Signal):
 
     Parameters
     ----------
-    name (str): signal name
-    type (type): signal type
-    id (str): a unique id of the signal
-    doc (str, optional): signal documentation
-    replaces (list of str): a list with names of signals replaced by this signal
-    default (bool, optional): when the widget outputs multiple signals of the
-        same type, one of them can set this flag to act as the default
-        (default: `False`)
-    explicit (bool, optional): if set, this signal is only used when it is
-        the only option or when explicitly connected in the dialog
-        (default: `False`)
-    dynamic (bool, optional): Specifies that the instances on the output will
-        in general be subtypes of the declared type and that the output can
-        be connected to any input signal which can accept a subtype of the
-        declared output type.
+    name (str):
+        signal name
+    type (type):
+        signal type
+    id (str):
+        a unique id of the signal
+    doc (str, optional):
+        signal documentation
+    replaces (list of str):
+        a list with names of signals replaced by this signal
+    default (bool, optional):
+        when the widget accepts multiple signals of the same type, one of them
+        can set this flag to act as the default (default: `False`)
+    explicit (bool, optional):
+        if set, this signal is only used when it is the only option or when
+        explicitly connected in the dialog (default: `False`)
+    dynamic (bool, optional):
+        Specifies that the instances on the output will in general be subtypes
+        of the declared type and that the output can be connected to any input
+        signal which can accept a subtype of the declared output type
+        (default: `True`)
     """
     def __init__(self, name, type, id=None, doc=None, replaces=None, *,
                  default=False, explicit=False, dynamic=True):
--- a/Orange/widgets/utils/signals.py
+++ b/Orange/widgets/utils/signals.py
@@ -126,7 +126,7 @@ class WidgetSignalsMixin:
             return [copy.copy(signal) for signal in old_style]
         signal_class = getattr(cls, direction.title())
         return [signal for signal in signal_class.__dict__.values()
-                if isinstance(signal, (InputSignal, OutputSignal))]
+                if isinstance(signal, (Input, Output))]
 
 
 class AttributeList(list):
--- a/Orange/widgets/utils/signals.py
+++ b/Orange/widgets/utils/signals.py
@@ -1,0 +1,133 @@
+import copy
+
+from Orange.canvas.registry.description import \
+    InputSignal, OutputSignal, \
+    Single, Multiple, Default, NonDefault, Explicit, Dynamic
+
+
+class Input(InputSignal):
+    def __init__(self, name, type, flags=None, id=None, doc=None, replaces=[],
+                 single=True, multiple=False, default=False, explicit=False,
+                 dynamic=True):
+        if flags is None:
+            flags = (single and Single) | \
+                    (multiple and Multiple) | \
+                    (default and Default or NonDefault) | \
+                    (explicit and Explicit) | \
+                    (dynamic and Dynamic)
+        super().__init__(name, type, "", flags, id, doc, replaces)
+
+    def __call__(self, method):
+        """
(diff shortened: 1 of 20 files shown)


Orange/widgets/widget.py
```diff
@@ -46,21 +48,10 @@ class WidgetMetaClass(type(QDialog)):
...
     def __new__(mcs, name, bases, kwargs):
-        from Orange.canvas.registry.description import (
-            input_channel_from_args, output_channel_from_args)
-
         cls = super().__new__(mcs, name, bases, kwargs)
         if not cls.name: # not a widget
             return cls
 
-        cls.inputs = [input_channel_from_args(inp) for inp in cls.inputs]
-        cls.outputs = [output_channel_from_args(outp) for outp in cls.outputs]
```
ales-erjavec: It might be better to still leave these lines. From memory, I think I saw some widgets that self inspect their own .{inputs,outputs}. They might expect that they are 'normalized' as {Input,Output}Signal. `get_widget_description` might not be called if running widget standalone, or when the descriptions are loaded from cache.

In any case changing them in `get_widget_descriptions` will trip someone up sooner or later.

janezd: You're right. Also, this is preparation of the class, so it belongs to the metaclass, not to `get_widget_description`.

astaric:
@ales-erjavec, can you check if your comments have been addressed and update your review accordingly?

Orange/widgets/widget.py
```diff
@@ -233,6 +226,21 @@ def __init__(self, *args, **kwargs):
...
 
     @classmethod
+    def get_widget_description(cls):
+        if not cls.name:
+            return
+        properties = {name: getattr(cls, name) for name in
+                      ("name", "icon", "description", "priority", "keywords",
+                       "help", "help_ref", "url",
+                       "version", "background", "replaces")}
+        return WidgetDescription(
```
kernc: How do you feel about returning a simple dict here and instantiating it in `WidgetDescription.from_module()`: `cls(**w.get_widget_description())`? This entangles the implementation even less and lets canvas itself fill its internal structures.

doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py
```diff
@@ -3,16 +3,20 @@
...
     icon = "icons/DataSamplerA.svg"
     priority = 10
 
-    inputs = [("Data", Orange.data.Table, "set_data")]
-    outputs = [("Sampled Data", Orange.data.Table)]
+    class Inputs:
+        data = Input("Data", Orange.data.Table)
+
+    class Outputs:
+        sample = Output("Sampled Data", Orange.data.Table)
```
kernc: Understandably, it's a little late for severe API change proposals, but how do we feel about:
* `Inputs` and `Outputs` identifiers being fixed, and
* in OWWidget metaclass interpret `(attr, value)` properties of these classes such that:
  * if `value` is an `Input`, no change;
  * if `value` is a `dict`, fill `Input(**d)`;
  * else, set value to `Input(name=attr.title(), type=value)`.

The simpler definitions could then be:
```py
class Inputs:
    data = Table
class Outputs:
    sampled_data = Table
```

Deriving the idea from `OWWidget.Error` where messages must be instances of `Msg()` whereas simple strings would (with some help from the metaclass) suffice.

Orange/widgets/utils/signals.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class Input(InputSignal, _Signal):
+    """
+    Description of an input signal.
+
+    The class is used to declare input signals for a widget as follows
+    (the example is taken from the widget Test & Score)::
+
+        class Inputs:
```
kernc: Could we name these `Input` and `Output` (singular) as we have `Error` and `Warning`, that is, for eventual consistency?

doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py
kernc: Poo. Decorating methods with `@Inputs.sampled_data` wouldn't work. 😢

Orange/widgets/utils/signals.py
astaric: Wouldn't that clash with the Input and Output class used for signal definition?

doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py
kernc: For the record, for decorators to work, we could possibly _use_ metaclass' `__prepare__`: https://www.python.org/dev/peps/pep-3115/#example
```py
class WidgetMeta:
    @classmethod
    def __prepare__(mcs, name, bases, **kwargs):
        return mcs.Dict()

    def __new__(mcs, name, bases, attrs):
        ...

    class Dict(dict):
        def __setitem__(self, item, value):
            if item in ('Input', 'Output'):
                # Transform Input/Output class so that their
                # attributes are of Signal type
                value = ...
            super().__setitem__(item, value)

Comment (janezd) [inline doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py]: Sure, but this would be a meta-class of Inputs and Outputs, which are currently not derived from any other class - which I like, since its simpler to define them. Plus, as much as I like writing magic, I hate reading it, even if it's mine.

I suggest we don't include it in this PR, but, as we discussed today, perhaps make the same change for signals and messages, if we decide so.

Comment (kernc) [inline doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py]: Correction, Inputs/Outpus need not extend anything. This would be OWWidget metaclass whose namespace dict reinterprets what it really means to be Inputs. Indeed, magic. (But the interface must be there for a reason.)

File: Orange/widgets/utils/signals.py (hunk 4 above)
Comment (janezd) [inline Orange/widgets/utils/signals.py]: I don't think it would clash, it's a different namespace.

I think the plural looks way better. We're listing outputs here. Having singular in messages was a mistake, I think. I used it so that raising messages looked better (`self.Error.data_mismatch()`), but I often make a mistake of declaring the class as `class Errors` instead of `class Error`.

Now, do we
- make the same mistake here, for consistency,
- use plural here, ignoring the consistency (but, heck, is plural indeed so much better?),
- change messages to plural (but see the comment above)?

Comment (kernc) [inline Orange/widgets/utils/signals.py]: https://en.wikipedia.org/wiki/Enumerated_type#C.2B.2B

File: doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py (hunk 3 above)
Comment (janezd) [inline doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py]: You're right. I wasn't paying attention and I thought that you manipulate the `Inputs`'s dictionary. Still, let's keep this out of this PR and decide later if we want it.

File: Orange/widgets/utils/signals.py (hunk 4 above)
Comment (janezd) [inline Orange/widgets/utils/signals.py]: This was why I use singular with messages. I'm kind of undecided, you vote for singular (yes?)... @astaric?

Comment (kernc) [inline Orange/widgets/utils/signals.py]: [Types are singular.](https://en.wikipedia.org/wiki/Enumerated_type#C.2B.2B), lists are plural. I don't vote.

Comment (astaric) [inline Orange/widgets/utils/signals.py]: Inputs and Outputs are not used as types, but as namespaces (We could drop the classes, put the signal defintions directly onto the widget and nothing would change - functionality wise). I do not consider them to be enums (links @kernc posted) but simply a way to group related attributes.

Plural looks better to me.

PS.: If this would not lead to even more discussion, I might even suggest using lowercase name (outputs, inputs), as the same name is used when it gets instantiated and one would expect lowercase name for instance variable.

File: doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py (hunk 3 above)
Comment (astaric) [inline doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py]: [Zen of python, line 2](https://www.python.org/dev/peps/pep-0020/#id3).

The way the code is written now, someone might even be able to figure out what is going on. If we us the "magic" way, PyCharm and PyLint would complain, which usually means that it is not that obvious what is going on ("you say in the class definition that Outputs.data is Table, but now you call Outputs.data.send()??? Table has no send????)

File: Orange/widgets/widget.py (hunk 2 above)
Comment (astaric) [inline Orange/widgets/widget.py]: I kind of like the idea

File: doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py (hunk 3 above)
Comment (kernc) [inline doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py]: I was actually agreeing with your counter example. Then I remembered we use Settings in exactly the same way.

Comment (astaric) [inline doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py]: Its actually the opposite way (explicit declaration magic usage, while your suggestion has magic declaration and explicit usage).

But settings could (should?) be made more explicit. Maybe by extending them from properties and dropping the magic in `__init__`? To better align with the rest, we could also introduce a new "Settings" class inside a widget that would contain all settings (the way we do now for messages and signals). Not sure if this is top priority though.

Comment (janezd) [inline doc/development/source/orange-demo/orangedemo/OWDataSamplerA.py]: .. and then write `self.Settings.variable` instead of `self.variable` everywhere in the widget? I would really hate that.

Having classes like `Error` or `Inputs` requires some magic, which I hate (although it's fun to write it). The reason for having them was to use identifiers instead of literal constants. The `Settings` class would have entirely different purpose, it would be about making the programmer aware that `self.variable` is saved as a setting. I don't see so much value in this; surely not enough to warrant another piece of magic.

As for properties... Generic example is

```
class P:
    def __init__(self,x):
        self.__x = x

    def get_x(self):
        return self.__x

    def set_x(self, x):
        self.__x = x

    x = property(get_x, set_x)
```

We don't want to write `set_x` and `get_x`. However, when the property is created (e.g. with `x = property(...)`) the name to which it is bound is not yet knownm, so we can't create the setter and getter. Metaclasses to the rescue again.

```
class Setting(property):
    def __init__(self, default, fget=None, fset=None):
        super().__init__(fget, fset)
        self.default = default

class WidgetMeta(type):
    def __new__(mcs, name, bases, kwargs):
        cls = super().__new__(mcs, name, bases, kwargs)

        for name, setting in cls.__dict__.items():
            if isinstance(setting, Setting):
                uname = "_" + name
                default = setting.default  # needed for closure; setting.default wouldn't work in getter
                def setter(widget, value):
                    setattr(widget, uname, value)
                def getter(widget):
                    return getattr(widget, uname, default)
                # fget and fset are read-only attributes of property, hence we must
                # create a new Setting
                setattr(cls, name, Setting(default, getter, setter))

        return cls

class Widget(metaclass=WidgetMeta):
    x = Setting(42)
    y = Setting("foo")

w = Widget()
print(w.x)
w.x = 50
print(w.x)
```

This is ugly and it doesn't have any benefits over the existing `Setting` - it still involves metamagic and PyCharm still doesn't know the types of `x` and `y`.

PR #2313, opened 2017-05-12
[ENH] PCA on sparse data.

### LINKED ISSUE #2255: SVD Widget
##### Feature Request
Since PCA cannot not work on sparse data due data centering, it would be nice to have an SVD widget. The functionality should be more or less the same as of PCA widget.

acopar:
##### Issue
Fixes #2255.
Use TruncatedSVD (LSA) on sparse data instead of PCA, since PCA does not work on sparse data. The difference is that TruncatedSVD truncates the vectors and does not center the data before calling svd. For dense data, functionality remains the same.

##### Description of changes
Workflow:
Corpus (Select bookexcerpts.tab) -> Bag of Words -> PCA
pca

- widgets/unsupervised/owpca.py: Select decomposition with radio buttons. 
- projection/pca.py: TruncatedSVD model
- tests: test_pca.py

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/pca.py
+++ b/Orange/projection/pca.py
@@ -33,7 +33,7 @@ class _FeatureScorerMixin(LearnerScorer):
 
 class PCA(SklProjector, _FeatureScorerMixin):
     __wraps__ = skl_decomposition.PCA
-    name = 'pca'
+    name = 'PCA'
     supports_sparse = False
 
     def __init__(self, n_components=None, copy=True, whiten=False,
@@ -61,7 +61,7 @@ class PCA(SklProjector, _FeatureScorerMixin):
 
 class SparsePCA(SklProjector):
     __wraps__ = skl_decomposition.SparsePCA
-    name = 'sparse pca'
+    name = 'Sparse PCA'
     supports_sparse = False
 
     def __init__(self, n_components=None, alpha=1, ridge_alpha=0.01,
@@ -126,7 +126,7 @@ class PCAModel(Projection, metaclass=WrapperMeta):
 
 class IncrementalPCA(SklProjector):
     __wraps__ = skl_decomposition.IncrementalPCA
-    name = 'incremental pca'
+    name = 'Incremental PCA'
     supports_sparse = False
 
     def __init__(self, n_components=None, whiten=False, copy=True,
@@ -157,7 +157,7 @@ class IncrementalPCAModel(PCAModel):
 
 class TruncatedSVD(SklProjector, _FeatureScorerMixin):
     __wraps__ = skl_decomposition.TruncatedSVD
-    name = 'truncated svd'
+    name = 'Truncated SVD'
     supports_sparse = True
 
     def __init__(self, n_components=None, algorithm='randomized', n_iter=5,
--- a/Orange/projection/pca.py
+++ b/Orange/projection/pca.py
@@ -38,22 +38,14 @@ class PCA(SklProjector, _FeatureScorerMixin):
 
     def __init__(self, n_components=None, copy=True, whiten=False,
                  svd_solver='auto', tol=0.0, iterated_power='auto',
-                 random_state=None, preprocessors=None, max_components=None):
+                 random_state=None, preprocessors=None):
         super().__init__(preprocessors=preprocessors)
-        if n_components is not None and max_components is not None:
-            raise ValueError("n_components and max_components can not both be defined.")
-        # max_components limits the number of PCA components if the minimum
-        # shape of the X matrix (after preprocessing) is higher than
-        # max_components, so that sklearn does not always compute the full
-        # transform, which is faster and uses less memory for big data.
-        self.max_components = max_components
         self.params = vars()
 
     def fit(self, X, Y=None):
         params = self.params.copy()
-        if params["n_components"] is None and self.max_components is not None:
-            # shape of X after preprocessing
-            params["n_components"] = min(min(X.shape), self.max_components)
+        if params["n_components"] is not None:
+            params["n_components"] = min(min(X.shape), params["n_components"])
         proj = self.__wraps__(**params)
         proj = proj.fit(X, Y)
         return PCAModel(proj, self.domain)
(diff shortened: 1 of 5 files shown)


Orange/projection/pca.py
```diff
@@ -152,6 +152,35 @@ def partial_fit(self, data):
...
+        self.max_components = max_components
+        self.params = vars()
+
+    def fit(self, X, Y=None):
+        params = self.params.copy()
+        if params["n_components"] is None and self.max_components is not None:
+            # shape of X after preprocessing
+            # Add -1 to avoid error in scikit fit_transform
+            # ValueError: n_components must be < n_features (strict)
+            params["n_components"] = min(min(X.shape) - 1, self.max_components)
```
kernc: Why, then, instead of `n_components` and `max_components` doing about the same thing not have just `n_components` and override it if shape too small?

E.g. what happens if `n_components=1e12` is passed and `X.shape == (3, 3)`?

Orange/tests/test_pca.py
```diff
@@ -84,6 +84,22 @@ def __ipca_test_helper(self, data, n_com, min_xpl_var):
         pc1_ipca = pca_model.partial_fit(data[1::2]).components_[0]
         self.assertAlmostEqual(abs(pc1_ipca.dot(pc1_pca)), 1, 4)
 
+    def test_truncated_svd(self):
+        data = self.ionosphere
+        self.__truncated_svd_test_helper(data, n_com=3, min_xpl_var=0.5)
+        self.__truncated_svd_test_helper(data, n_com=10, min_xpl_var=0.7)
+        self.__truncated_svd_test_helper(data, n_com=31, min_xpl_var=0.99)
+
+    def __truncated_svd_test_helper(self, data, n_com, min_xpl_var):
```
kernc: It's actually ok to use full, _expressive_ variable names where common abbreviations don't exist (e.g. `n_` and `min_` are ok).

Orange/projection/pca.py
acopar: I agree, max_components was in there because of similarities with PCA, but is in fact not needed for TruncatedSVD model. To answer your question, n_components is bound by max_components from inside the widget, so in this particular case it would not make any difference :) Will fix it.

Orange/widgets/unsupervised/owpca.py
```diff
@@ -375,6 +386,26 @@ def _update_normalize(self):
...
+            self._pca_projector.component = self.ncomponents
+            self._pca_preprocessors = PCA.preprocessors
+        else:
+            # Equivalent to PCA, but without data centering
+            self._pca_projector = TruncatedSVD(max_components=MAX_COMPONENTS)
+            self._pca_projector.component = self.ncomponents
+            self._pca_preprocessors = TruncatedSVD.preprocessors
+
+    def _update_center(self):
+        if self.center and self.data is not None and self.data.is_sparse():
```
nikicc: This code is currently unreachable, if I am not mistaken. You are setting `self.center = False` if data is sparse in `set_data`. It is here as a sanity check?

Orange/widgets/unsupervised/owpca.py
```diff
@@ -196,9 +200,16 @@ def set_data(self, data):
             self.sampling_box.setVisible(False)
         if isinstance(data, Table):
             if data.is_sparse():
-                self.Error.sparse_data()
-                self.clear_outputs()
-                return
+                # PCA does not support sparse data
+                # Falling back to TruncatedSVD aka LSA
+                self.center = False
```
nikicc: Besides `self.center` also `self.normalize` shoud be disabled and set to False for sparse data sets.

Default normalization cannot handle sparse data set. It might have worked on BoW data (for which all features are marked to be skipped during normalization), but it does not work for sparse data sets in general.

To reproduce the problem, transform Iris to sparse in Python script, pass it to PCA and check Normalize data.

Orange/widgets/unsupervised/owpca.py
```diff
@@ -125,6 +126,8 @@ def __init__(self):
         self.options_box = gui.vBox(self.controlArea, "Options")
         gui.checkBox(self.options_box, self, "normalize", "Normalize data",
                      callback=self._update_normalize)
+        self.center_box = gui.checkBox(self.options_box, self, "center", 
+            "Center data", callback=self._update_center)
```
nikicc: I'm not sure `Center data` is the right name for this checkbox. The problem as I see it, is that we also have `Normalize data` option above, which can be misleading since data normalization by default implicates data centering. If I select Normalize data and don't select Center data I would expect that the data is only scales, but this is not what happens.

 In essence this checkbox is selecting between PCA and TruncatedSVD. Should we maybe put them to a dropdown?

acopar:
> Besides comments above, I also found one functional glitch. If I pass Iris to PCA then clicking on and off to Center data causes number of components to jump from 3 to 4. Is this expected?

This is a limitation of TruncatedSVD. It needs the number of components to be strictly less than the number of features. See n_components parameter in [TruncatedSVD](http://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html)

> In essence this checkbox is selecting between PCA and TruncatedSVD. Should we maybe put them to a dropdown?

I converted this to radio buttons instead.

All other requests should be fulfilled in this commit. Refer to the file changes and commit message for details.

Orange/widgets/unsupervised/owpca.py
```diff
@@ -194,11 +226,21 @@ def set_data(self, data):
...
             self.sampling_box.setVisible(False)
+
+        self.openContext(data)
         if isinstance(data, Table):
             if data.is_sparse():
-                self.Error.sparse_data()
-                self.clear_outputs()
-                return
+                # PCA does not support sparse data
+                # Falling back to TruncatedSVD aka LSA
```
nikicc: This comment doesn't apply any more, does it?

Orange/widgets/unsupervised/owpca.py
```diff
@@ -160,6 +174,23 @@ def update_model(self):
         else:
             self.__timer.stop()
 
+    def update_buttons_sparse(self):
```
nikicc: Instead of this I would simply put:
```py
def toggle_buttons(self, sparse_data=False):
    buttons = self.decomposition_box.buttons
    for i, cls in enumerate(DECOMPOSITIONS):
        buttons[i].setDisabled(sparse_data and not cls.supports_sparse)
```

and then call with `self.toggle_buttons(data is not None and data.is_sparse())`

Orange/widgets/unsupervised/owpca.py
```diff
@@ -194,11 +226,21 @@ def set_data(self, data):
                 self.start_button.setEnabled(True)
         if not isinstance(data, SqlTable):
             self.sampling_box.setVisible(False)
+
+        self.openContext(data)
         if isinstance(data, Table):
             if data.is_sparse():
```
nikicc: I would put this:
```py
    self.openContext(data)
    if data.is_sparse():
        self.normalize = False
        self.normalize_box.setEnabled(False)
    else:
        self.normalize_box.setEnabled(True)
    self._update_decomposition()

self.toggle_buttons(data is not None and data.is_sparse())
self.data = data
self.fit()
```
after checks for errors.

acopar:
QRadioButton has no attribute isDisabled(), so I must use isEnabled function in comparisons. Other than that, all other requests should be fixed.

Orange/projection/pca.py
```diff
@@ -152,6 +155,38 @@ def partial_fit(self, data):
...
+    def __init__(self, n_components=None, algorithm='randomized', n_iter=5,
+                 random_state=None, tol=0.0, preprocessors=None, max_components=None):
+        super().__init__(preprocessors=preprocessors)
+        if n_components is not None and max_components is not None:
+            raise ValueError("n_components and max_components can not both be defined.")
+        # max_components limits the number of SVD components if the minimum
+        # shape of the X matrix (after preprocessing) is higher than
+        # max_components, so that sklearn does not always compute the full
+        # transform, which is faster and uses less memory for big data.
+        self.max_components = max_components
```
kernc: I'd have `max_components` removed (here and in `PCA` above). No good reason for it. None.

nikicc: +1 for removing `max_components `.

acopar:
max_components is removed from PCA and TruncatedSVD models. Instead, n_components is used to perform the same task.

nikicc:
@markotoplak  we removed `max_components` argument from PCA that you recently introduced in #2234 and propose to simply use `n_components` instead. Please, check if all is OK, especially commit 99e9f969db88abb095053db2474b1ddafc5355c5.

markotoplak:
@nikicc What removing max_components changes is that the number of components on the output can now be smaller than n_components (before an error was raised). This changes nothing in the widgets, but the script users should take care to check the output. I think this is OK.

jerneju:
https://sentry.io/biolab/orange3/issues/287354884/

PR #2314, opened 2017-05-15
[FIX] Paint Data: now stores data in list and not np.array

jerneju:
##### Issue
https://sentry.io/biolab/orange3/issues/240133185/

One way to reproduce this behavior:
Use _Paint Data_ widget and draw some points. Then save it. Close the Orange. Open the Orange and then load the file and close the Orange again. Error should be thrown.

These steps were tested in Orange versions _master_, _3.4.2_, _3.4.1_ and _3.3.12_. In all those versions error occurred.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpaintdata.py
+++ b/Orange/widgets/data/owpaintdata.py
@@ -798,8 +798,13 @@ class OWPaintData(OWWidget):
         self.class_model.rowsRemoved.connect(self._class_count_changed)
 
         if self.data is None:
-            self.data = np.zeros((0, 3))
-        self.__buffer = self.data.copy()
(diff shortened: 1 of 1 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/240133185/

astaric:
How can the error be reproduced by user? (= not in code)?

jerneju:
@astaric : Now added above.

jerneju:
Changed _Paint Data_. Now stores points in `List`. Therefore settings are saved in _literal_ format and not _pickle_.

PR #2321, opened 2017-05-17
[MNT] Travis: Stop running radon in favor of codeclimate

nikicc:
##### Issue

Radon requirements occasionally cannot be met while a pass on Travis is required for all PRs.

##### Description of changes
Stop running radon on Travis and check codeclimate instead.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.travis/stage_script.sh
+++ b/.travis/stage_script.sh
@@ -2,6 +2,7 @@
 if [ "$RUN_PYLINT" ]; then
     cd $TRAVIS_BUILD_DIR
     foldable pip install -r requirements-dev.txt
+    foldable pip uninstall -y radon     # disable radon in favor of codeclimate
     cp pylintrc ~/.pylintrc
     .travis/check_pylint_diff
     exit $?


astaric:
What about simply uninstalling radon before running the script?

IIUC, the scripts only runs pylint checks if radon is not available.

janezd:
What do we get from code climate? I get lost in its GUI, while functionality-wise it is radon + code duplication. As for the latter, it's overly sensitive and mostly useless. As for the former, I reading the radon's output was easier. I may be really wrong. Can we wait for Friday, so you can convince me?

nikicc:
IMHO if we are not 100% strict about radon, then it probably shouldn't be among required checks. Using codeclimate instead would minimize the number of times we have to merge PR that don't have all required checks. But let's discuss this tomorrow.

nikicc:
@astaric, @janezd I removed radon by uninstalling it before running the script. Please check. Note that commit d0a487b1416f087196d34dd8927a27b3851ba734 should be removed before merge.

Issue #2310, opened 2017-05-11
Want to install parallel coordinates in Orange3


SamGG:
Hi,
I would like to use the parallel coordinates widget. This was present in V2.7 (if I remember well), but I didn't find it in V3.4. I found the code in GitHub. Could you help me in adding this visualization widget in V3.4 under Windows?
Best.
https://github.com/biolab/orange3/blob/7dca99338f1eaca7f8d6f4925dd79944a8aa426d/Orange/widgets/visualize/owparallelcoordinates.py
I already asked on stackexchange as advised, but no answer yet.
https://datascience.stackexchange.com/questions/18816/
I did successfully install network tools using tar.gz, pip install and MVS compiler.
Best.

##### Orange version
Orange is 3.4.2, git revision f3076cb, Python 3.4.4 (v3.4.4:737efcadf5a6, Dec 20 2015, 19:28:18) [MSC v.1600 32 bit (Intel)] on win32
running on Win7 Pro 64 bits

##### Expected behavior
I want to view data using parallel coordinates

##### Actual behavior
Not available, although the widget is present on GitHub

##### Steps to reproduce the behavior
NA

##### Additional info (worksheets, data, screenshots, ...)
NA

kernc:
You can checkout https://github.com/biolab/orange3/pull/2327 and see if it works for your use case. Thanks!

SamGG:
Thanks.

lukaseller:
I am also wondering how to get "Parallel Coordinates" up and running. 
Already installed Prototypes 0.14.2 for Orange 3.27 on MacOs, but i cannot find it. 
Was it moved to another package? 
Thanks in advance

rawiz-coder:
I was hoping to use this too but also can't see it in Mac OS Orange v3.39.0 and Prototypes v0.21.2, is it not there anymore or somewhere else or just obsolete (any plans to bring it back)? This kind of visualisation is really great for say eyeballing what kind of factory conditions lead to particular things. Maybe some other visualisation is suggested instead though?

wvdvegte:
Parallel coordinates plot is a very common visualization in machine learning. I don't understand why it's (currently) not available in Orange.

rawiz-coder:
Interestingly it does appear to be in KNIME, although ORANGE is supposed to be easier to use of course.

wvdvegte:
[...]

PR #2327, opened 2017-05-19
[ENH] OWParallelCoordinates: Plotly-based

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/2310

##### Description of changes
Remove defunct parallel coordinates widget; add [Plotly JS](https://plot.ly/python/) graphing library; use it for parallel coordinates widget.

##### Includes
- [X] Code changes
- [ ] Tests
- [x] Documentation

screenshot - 05192017 - 06 54 01 pm


astaric:
I have played around with the widget. It behaves strangely.

* When I click on an attribute in the list, selection keeps changing as I move the mouse around (as if it did not register the mouse-up event).
* The color combo also sometimes selects an item that is different from the One O clicked.
* When I add a new Parallel Coordinates widget to the workflow, connect it to data and open it, it performs 1-2 resizes before the plot changes to correct size.
* I haven't found a way to remove the selection.
* If I click on Optimize Selected Dimension and the plot is displaying more than 9 variables, nothing happens.

I have stopped at this point :)

astaric:
While I would really like to see this merged, It will not make it to the 3.7 release.

@BlazZupan, @janezd, could you take a look and comment on this new widget? If major changes are required, I would rather have it included in prototypes first and then promoted to the core when ready.

kernc:
Was worth a try. 😄

kernc:
For the time being, obsoleted by https://github.com/biolab/orange3-prototypes/pull/99.

PR #2335, opened 2017-05-23
[FIX] MDS: Support distances without domain information

jerneju:
##### Issue
screenshot_20170523_124611

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -1,5 +1,4 @@
 import sys
-import warnings
 
 from xml.sax.saxutils import escape
 from itertools import chain
@@ -36,11 +35,11 @@ from Orange.widgets.utils.annotated_data import (create_annotated_table,
                                                  ANNOTATED_DATA_SIGNAL_NAME)
 
 
-def stress(X, D):
-    assert X.shape[0] == D.shape[0] == D.shape[1]
+def stress(X, distD):
+    assert X.shape[0] == distD.shape[0] == distD.shape[1]
     D1_c = scipy.spatial.distance.pdist(X, metric="euclidean")
     D1 = scipy.spatial.distance.squareform(D1_c, checks=False)
-    delta = D1 - D
+    delta = D1 - distD
     delta_sq = numpy.square(delta, out=delta)
     return delta_sq.sum(axis=0) / 2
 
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/256594464/

jerneju:
https://sentry.io/biolab/orange3/issues/274315543/

Orange/widgets/unsupervised/owmds.py
```diff
@@ -451,7 +451,7 @@ def _clear_plot(self):
             self._legend_item = None
 
     def update_controls(self):
-        if self.data is None and getattr(self.matrix, 'axis', 1) == 0:
+        if self.data is None and getattr(self.matrix, 'axis', 1) in [0, 1]:
```
janezd: `axis` is always either 0 or 1!
 
   .. attribute:: axis

        If axis=1 we calculate distances between rows,
        if axis=0 we calculate distances between columns.

This check is then redundant.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -451,7 +451,7 @@ def _clear_plot(self):
...
             # Column-wise distances
             attr = "Attribute names"
             self.labelvar_model[:] = ["No labels", attr]
```
janezd: ... and if it's 1, the distance is computed over rows, so the widget shouldn't offer *Attribute names* as labels. The option should be called something like "Index". Offer it only for labels, but not for colors, shapes and sizes since it makes no sense there.

While your at it, fix old typeo: solor -> color.

jerneju:
https://sentry.io/biolab/orange3/issues/294176024/

Orange/widgets/unsupervised/owmds.py
```diff
@@ -448,15 +448,14 @@ def _clear_plot(self):
...
-                self.color_value = attr
-                self.shape_value = attr
+            axis = getattr(self.matrix, 'axis', 1)
+            attr = ["Column labels", "Row labels"][axis]
+            self.labelvar_model[:] = ["No labels", attr]
+            self.shapevar_model[:] = ["Same shape"]
+            self.colorvar_model[:] = ["Same color"]
+
+            self.color_value = "Same shape"
+            self.shape_value = "Same color"
```
lanzagar: These two are switched :)

PR #2354, opened 2017-05-30
[FIX] Credentials: handling password credentials error

jerneju:
##### Issue
https://github.com/biolab/orange3-text/pull/253

##### Description of changes
try except block

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/credentials.py
+++ b/Orange/widgets/credentials.py
@@ -1,6 +1,10 @@
+import logging
+
 import keyring
 
 SERVICE_NAME = 'Orange3 - {}'
+
+log = logging.getLogger(__name__)
 
 
 class CredentialManager:
@@ -13,13 +17,11 @@ class CredentialManager:
 
     Examples:
         >>> cm = CredentialManager('Widget Name')
-        >>> cm.username = 'Orange'      # store username
-        >>> cm.password = 'Secret'      # store password
-
-        >>> cm.username                 # get username
-        'Orange'
-        >>> cm.password                 # get password
-        'Secret'
+        >>> cm.some_secret = 'api-key-1234'
+        >>> cm.some_secret
+        'api-key-1234'
+        >>> del cm.some_secret
+        >>> cm.some_secret
     """
     def __init__(self, service_name):
         self.__dict__['__service_name'] = SERVICE_NAME.format(service_name)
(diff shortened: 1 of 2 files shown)


Orange/widgets/credentials.py
```diff
@@ -32,7 +32,10 @@ def __setattr__(self, key, value):
         keyring.set_password(self.service_name, key, value)
 
     def __getattr__(self, item):
-        return keyring.get_password(self.service_name, item)
+        try:
+            return keyring.get_password(self.service_name, item)
+        except Exception:
+            return ""
```
kernc: Shouldn't we default to None as is returned when the key is not yet set?
```py
>>> import keyring
>>> print(keyring.get_password('Orange3 - random test', 'nonexistent key')
None
```

Orange/widgets/credentials.py
```diff
@@ -32,7 +32,10 @@ def __setattr__(self, key, value):
         keyring.set_password(self.service_name, key, value)
```
kernc: Can the setter also raise e.g. if the keyring isn't unlocked?

Orange/widgets/credentials.py
jerneju: OK.

Orange/widgets/credentials.py
jerneju: Yes it does.

Orange/widgets/credentials.py
```diff
@@ -29,10 +33,19 @@ def service_name(self):
...
-        return keyring.get_password(self.service_name, item)
+        try:
+            return keyring.get_password(self.service_name, item)
+        except Exception:
+            log.error("Failed to get password for '{}'.".format(self.service_name),
+                      exc_info=True)
+            return None
 
     def __delattr__(self, item):
         keyring.delete_password(self.service_name, item)
```
kernc: Can the deleter, then, also raise?

jerneju: Yes, sure it can. For instance:
```python
# keyring.errors.PasswordDeleteError: No such password!
# RuntimeError:
```

jerneju:
https://sentry.io/biolab/orange3-text-bl/issues/282422265/

Issue #1916, opened 2017-01-13
FCBC Score segfaults


ales-erjavec:
##### Orange version

3.3.9, 3.3.10.dev0, 3.3.10rc1

##### Expected behavior
When the input data (X columns) to the 'Rank' widget contains all unknown values the FCBC score behaves like the rest and outputs a 'nan' score

##### Actual behavior
FCBC score causes an application segfaults.

The following is printed to the console
```
Exception ignored in: 'Orange.preprocess._relieff._contingency_table'
ValueError: negative dimensions are not allowed
Segmentation fault: 11
```
And this is the full stack trace
```
Thread 0 Crashed:: CrBrowserMain  Dispatch queue: com.apple.main-thread
0   _relieff.cpython-35m-darwin.so	0x000000012a0f980b __pyx_memoryview_fromslice(__Pyx_memviewslice, int, _object* (*)(char*), int (*)(char*, _object*), int) + 267 (_relieff.cpp:1072)
1   _relieff.cpython-35m-darwin.so	0x000000012a1129c7 __pyx_pw_6Orange_10preprocess_8_relieff_1contingency_table(_object*, _object*, _object*) + 2919 (_relieff.cpp:4711)
2   org.python.python             	0x0000000108324039 PyCFunction_Call + 60
3   org.python.python             	0x00000001083844ab PyEval_EvalFrameEx + 22999
4   org.python.python             	0x00000001083882ad fast_function + 220
5   org.python.python             	0x000000010838432b PyEval_EvalFrameEx + 22615
6   org.python.python             	0x00000001083879fd _PyEval_EvalCodeWithName + 1886
7   org.python.python             	0x000000010838832d fast_function + 348
8   org.python.python             	0x000000010838432b PyEval_EvalFrameEx + 22615
9   org.python.python             	0x00000001083879fd _PyEval_EvalCodeWithName + 1886
10  org.python.python             	0x000000010837eabf PyEval_EvalCodeEx + 78
11  org.python.python             	0x000000010830d760 function_call + 369
12  org.python.python             	0x00000001082edcb5 PyObject_Call + 101
13  org.python.python             	0x00000001082fdeb5 method_call + 140
14  org.python.python             	0x00000001082edcb5 PyObject_Call + 101
15  org.python.python             	0x00000001083358f3 slot_tp_call + 54
16  org.python.python             	0x00000001082edcb5 PyObject_Call + 101
[...]

PR #2355, opened 2017-05-30
[FIX] score.FCBF: don't segfault on continuous variables w/ <0 values

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/1916

##### Description of changes
Previously, FCBF always computed contingency as if the variable were discrete. Now uses Orange's Discrete contingency, which discretizes as necessary.
This should, hopefully, also fix segfault on OSX when all values are missing.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/tests/test_score_feature.py
+++ b/Orange/tests/test_score_feature.py
@@ -5,7 +5,7 @@ import unittest
 
 import numpy as np
 
-from Orange.data import Table, Domain, DiscreteVariable
+from Orange.data import Table, Domain, DiscreteVariable, ContinuousVariable
 from Orange import preprocess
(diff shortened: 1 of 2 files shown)


Orange/tests/test_score_feature.py
```diff
@@ -129,5 +129,13 @@ def test_fcbf(self):
         scorer = FCBF()
         weights = scorer(self.zoo, None)
         found = [self.zoo.domain[attr].name for attr in reversed(weights.argsort()[-5:])]
-        reference = ['legs', 'backbone', 'toothed', 'hair', 'aquatic']
+        reference = ['legs', 'milk', 'toothed', 'feathers', 'backbone']
```
kernc: This better now. 😉 👍

PR #2356, opened 2017-05-30
[FIX] OWMap: minor fixes

kernc:
##### Issue
Geo Map crashed when data input, then subset input, then data input cleared.
Geo Map showed warning when setting DomainModel-controlled master value attributes to None.

##### Description of changes
Neither of above should happen.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -7,6 +7,7 @@ import os
 import re
 import itertools
 import warnings
+import logging
 from types import LambdaType
 from collections import defaultdict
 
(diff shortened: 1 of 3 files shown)


kernc:
https://sentry.io/biolab/orange3/issues/282989002/

ajdapretnar:
This works for the workflow we talked about. However, the widget crashes when I connect it to a data set without any numerical features (say, bookexcerpts.tab), open GeoMap and press 'zoom in' function. Also it is constantly printing out all the tokens from the 'Unable to set ComboBox' error. Perhaps just something to keep the console clean?
The rest is good.

kernc:
Thanks. Please see again.

ajdapretnar:
This is good! I like the changes.

But again, what to do regarding the checks?

PR #2357, opened 2017-05-30
[FIX] Impute: sparse

### LINKED ISSUE #2349: Imputation Doesn't Work on Some Sparse Data Sets
##### Orange version

3.5.0.dev0+554b885

##### Expected behavior
Imputation fills missing values.

##### Actual behavior
It doesn't fill missing values on some data sets.

Most easily reproducible on Text. Load Election-2016 tweets, pass them through BoW and Impute with Average. The resulting matrix still has some missing values.

screen shot 2017-05-29 at 15 02 17

jerneju:
##### Issue
Fixes #2349.

##### Description of changes
Work in progress.

##### Includes
- [x] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/impute.py
+++ b/Orange/preprocess/impute.py
@@ -296,7 +296,10 @@ class ReplaceUnknownsRandom(Transformation):
             self.sample_prob = np.ones_like(counts) / len(counts)
 
     def transform(self, c):
-        c = np.array(c, copy=True)
+        if not sp.issparse(c):
+            c = np.array(c, copy=True)
+        else:
+            c = c.toarray().ravel()
         nanindices = np.flatnonzero(np.isnan(c))
 
         if self.variable.is_discrete:
--- a/Orange/preprocess/impute.py
+++ b/Orange/preprocess/impute.py
@@ -1,5 +1,5 @@
-import numpy
-from scipy.sparse import issparse
+import numpy as np
+import scipy.sparse as sp
 
 import Orange.data
 from Orange.statistics import distribution, basic_stats
@@ -26,11 +26,11 @@ class ReplaceUnknowns(Transformation):
         self.value = value
 
     def transform(self, c):
-        if issparse(c):
-            c.data = numpy.where(numpy.isnan(c.data), self.value, c.data)
+        if sp.issparse(c):
+            c.data = np.where(np.isnan(c.data), self.value, c.data)
             return c
         else:
-            return numpy.where(numpy.isnan(c), self.value, c)
+            return np.where(np.isnan(c), self.value, c)
 
 
 class BaseImputeMethod(Reprable):
@@ -83,7 +83,7 @@ class DropInstances(BaseImputeMethod):
 
     def __call__(self, data, variable):
         index = data.domain.index(variable)
-        return numpy.isnan(data[:, index]).reshape(-1)
+        return np.isnan(data[:, index]).reshape(-1)
 
 
 class Average(BaseImputeMethod):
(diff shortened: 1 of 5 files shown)


nikicc:
@jerneju I already did some debugging about this on Friday and IMO the problem is in the file `Orange/statistics/until.py`, method `stats` which for sparse returns `X.min` and `X.max`.  I think `X.min` and `X.max` don't handle missing values and return `np.nan` when some values are missing, while we probably should return minimum and maximum among defined values only. They should probably just be replaced with `nanmin` and `nanmax` methods (L256–272 in the same file).

jerneju:
https://sentry.io/biolab/orange3/issues/284690104/

jerneju:
https://sentry.io/biolab/orange3/issues/284690041/

jerneju:
Well, additional issue:
screenshot_20170601_135914

Orange/tests/test_util.py
```diff
@@ -106,3 +107,12 @@ def assertCorrectArrayType(self, array, shape, sparsity):
...
             warnings.warn('foo', OrangeDeprecationWarning)
+
+    def test_stats_sparse(self):
+        """
+        Stats should not fail when trying to calculate mean on sparse data.
+        GH-2357
+        """
+        x = np.array([[0], [np.nan], [9]])
+        x = sp.csr_matrix(x)
+        self.assertEqual(stats(x)[0][2], 3.)
```
nikicc: This should be 4.5 not 3.

Orange/statistics/util.py
```diff
@@ -281,13 +281,13 @@ def mean(x):
...
-def nanmean(x):
+def nanmean(x, axis=None):
     """ Equivalent of np.nanmean that supports sparse or dense matrices. """
     if not sp.issparse(x):
-        return np.nanmean(x)
+        return np.nanmean(x, axis=axis)
 
-    n_values = np.prod(x.shape) - np.sum(np.isnan(x.data))
-    return np.nansum(x.data) / n_values
+    x.data = np.nan_to_num(x.data)
```
nikicc: `nan_to_num` converts `np.nans` to zeros, which causes `mean` to also treat them as zeros. E.g. for the sparse array of `[np.nan, np.nan, 1]` this implementation returns 0.33 instead of 1. 

What's wrong with the previous implementation?

Orange/preprocess/impute.py
```diff
@@ -311,7 +311,11 @@ def __init__(self, variable, distribution):
             self.sample_prob = np.ones_like(counts) / len(counts)
 
     def transform(self, c):
-        c = np.array(c, copy=True)
+        if not sp.issparse(c):
+            c = np.array(c, copy=True)
+        else:
+            c = c.copy()
```
nikicc: Why do we need a copy? Doesn't `toarray()` already takes care of this?

Orange/preprocess/impute.py
```diff
@@ -311,7 +311,11 @@ def __init__(self, variable, distribution):
...
+            c = c.toarray().flatten()
```
nikicc: Should we use `ravel` instead that doesn't necessarily make an other copy?

Orange/statistics/util.py
```diff
@@ -281,14 +281,32 @@ def mean(x):
     return np.sum(x.data) / n_values
 
 
-def nanmean(x):
+def nanmean(x, axis=None):
```
nikicc: What about:
```py
def nanmean(x, axis=None):
    """ Equivalent of np.nanmean that supports sparse or dense matrices. """
    def nanmean_sparse(x):
        n_values = np.prod(x.shape) - np.sum(np.isnan(x.data))
        return np.nansum(x.data) / n_values

    if not sp.issparse(x):
        return np.nanmean(x, axis=axis)
    if axis is None:
        return nanmean_sparse(x)
    if axis in [0, 1]:
        arr = x if axis == 1 else x.T
        return np.array([nanmean_sparse(row) for row in arr])
    else:
        raise NotImplementedError
```

Orange/preprocess/impute.py
```diff
@@ -296,7 +296,10 @@ def __init__(self, variable, distribution):
             self.sample_prob = np.ones_like(counts) / len(counts)
 
     def transform(self, c):
-        c = np.array(c, copy=True)
+        if not sp.issparse(c):
+            c = np.array(c, copy=True)
+        else:
+            c = c.toarray().ravel()
```
nikicc: What about if we take only `c.data` here and we would need to density the whole column? Consequently, we would need to set only `c.data` in L314.

Orange/statistics/util.py
jerneju: Well, I did some speed testing. The results are interesting and are listed below:

> Ratio for axis  0 :  1.558
> Ratio for axis  1 :  0.664

PR #2368, opened 2017-06-02
DiscreteVariable: make 'values' a read-only property

janezd:
Make `Discrete.values` a read-only property.

Methods for adding to `values` are made private and they are only used in `find_compatible`, which is called by `make`. This is currently unavoidable.

This is related to #2277, but deserves a separate PR since it is more general.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/canvas/scheme/scheme.py
+++ b/Orange/canvas/scheme/scheme.py
@@ -19,7 +19,7 @@ from .node import SchemeNode
 from .link import SchemeLink, compatible_channels
 from .annotations import BaseSchemeAnnotation
 
-from ..utils import check_arg, check_type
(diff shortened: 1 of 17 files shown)


markotoplak:
Widget tests fail.

janezd:
This is not doable. To be read-only, `values` have to be a tuple instead of a list. This breaks, for instance, all widgets that use `combo.setItems(["Same color"] + var.values)`. While we can fix this in the core, it will still break all add-ons.

Without making `values` a tuple, making it a property makes no sense.

I suggest abandoning this PR and keep just *acting* that `values` are read-only (except when we need to change them). @astaric?

janezd:
OK, let's put this PR out of misery.

PR #2371, opened 2017-06-02
[FIX] owmanifoldlearning: Remove `n_jobs=-2` parameter

ales-erjavec:
##### Issue

Manifold Learning frequently deadlocks with the forking multiprocessing backend and a
threaded blas library.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


kernc:
Is it likely that setting the default multiprocessing start method to [forkserver](https://github.com/biolab/orange3/pull/2288/commits/c0cf126a7eb6017d9ed20ffd1a36db3056a05d4b) should mitigate this?

ales-erjavec:
> Is it likely that setting the default multiprocessing start method to forkserver should mitigate this?

Probably. But I would still prefer not to.

PR #2377, opened 2017-06-05
[FIX] Silhouette Plot: now setting axis range properly

jerneju:
##### Issue
Handling `nan` values.
https://sentry.io/biolab/orange3/issues/220334718/
screenshot_20170605_161318

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owsilhouetteplot.py
+++ b/Orange/widgets/visualize/owsilhouetteplot.py
@@ -389,7 +389,7 @@ class OWSilhouettePlot(widget.OWWidget):
 
             if self._mask is not None:
                 scores = np.full(shape=selectedmask.shape,
-                                    fill_value=np.nan)
+                                 fill_value=np.nan)
                 scores[~self._mask] = self._silhouette
             else:
                 scores = self._silhouette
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -574,11 +574,11 @@ def clear(self):
 
     def __setup(self):
         # Setup the subwidgets/groups/layout
-        smax = max((numpy.max(g.scores) for g in self.__groups
+        smax = max((numpy.max(numpy.nan_to_num(g.scores)) for g in self.__groups
```
janezd: Why not `numpy.nanmax` (here) and `numpy.nanmin` (below)? Plus, set it to 1 or -1 if the length is zero or all values are NaN.

janezd:
Is this still WIP, as the title suggests, or can we merge it?

As a sidenote, the snapshot again provides no information (except that the bug can be reproduced by connecting Silhouette to File), so one has to check the code to see what the PR is about. No need to fix this here, but please provide meaningful descriptions in the future.

janezd:
You can add `# pylint disable=protected-access` at the top of the tests. Tests are allowed to access anything. This will decrease the lint error count.

jerneju:
Ok, fixed.

Orange/widgets/visualize/tests/test_owsilhouetteplot.py
```diff
@@ -1,3 +1,4 @@
+# pylint disable=protected-access
```
janezd: Sorry for my typeo: this has no effect due to a missing a colon after pylint (see below). You can also add a comment `# Test methods are allowed to access private members` above.

jerneju: Well, my mistake, I should have known by now that there is a colon after `# pylint`.

PR #2379, opened 2017-06-05
[FIX] OWColor: Fix propagating changes to the output

### LINKED ISSUE #2378: OWColors Widget Doesn't Propagate Changes
##### Orange version

3.5.0.dev0+971a83a

##### Expected behavior
After one changes some colors, the changes should be propageted to say ScatterPlot.

##### Actual behavior
Changes of colors are not propagated.

Load `iris`, pass it though Color widget and then to ScatterPlot

screen shot 2017-06-05 at 17 03 45

nikicc:
##### Issue

Fixes #2378 

##### Description of changes
Assure that when domains are equal, `from_tables` sets the resulting table's domain to the instance that was passed on the input.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -349,7 +349,10 @@ class Table(MutableSequence, Storage):
                     if cached:
                         return cached
                 if domain == source.domain:
-                    return cls.from_table_rows(source, row_indices)
+                    table = cls.from_table_rows(source, row_indices)
+                    # assure resulting domain is the instance passed on input
+                    table.domain = domain
+                    return table
 
                 if isinstance(row_indices, slice):
                     start, stop, stride = row_indices.indices(source.X.shape[0])
--- a/Orange/tests/test_table.py
+++ b/Orange/tests/test_table.py
@@ -1,6 +1,7 @@
 # Test methods with long descriptive names can omit docstrings
 # pylint: disable=missing-docstring
 
+import copy
 import os
 import unittest
 from itertools import chain
@@ -1734,6 +1735,12 @@ class CreateTableWithDomainAndTable(TableTests):
         self.assertIsInstance(new_table, MyTableClass)
         self.assertIsNot(table, new_table)
         self.assertIs(new_table.domain, new_domain)
+
+    def test_transform_same_domain(self):
+        iris = data.Table("iris")
+        new_domain = copy.copy(iris.domain)
+        new_data = iris.transform(new_domain)
+        self.assertIs(new_data.domain, new_domain)
 
     def test_can_copy_table(self):
         new_table = data.Table.from_table(self.domain, self.table)
--- a/Orange/tests/test_table.py
+++ b/Orange/tests/test_table.py
@@ -3,14 +3,14 @@
 
 import copy
 import os
+import random
 import unittest
 from itertools import chain
 from math import isnan
-import random
-
 from unittest.mock import Mock, MagicMock, patch
-import scipy.sparse as sp
+
 import numpy as np
+import scipy.sparse as sp
 
 from Orange import data
 from Orange.data import (filter, Unknown, Variable, Table, DiscreteVariable,


Orange/widgets/data/owcolor.py
```diff
@@ -350,6 +350,7 @@ def set_data(self, data):
                                         self._create_proxies(domain.metas))
             self.openContext(data)
             self.data = data.transform(domain)
+            self.data.domain = domain
```
janezd: I guess we need a comment here, at the very least.

But there may be a more general problem here. One would expect that after `self.data = data.transform(domain)`, we would have `self.data.domain is domain`, while `transform` only ensures `self.data.domain == domain`. We should either change `transform` to (always) compare domains with `is`, not `==`, or we can add a flag to specify how to compare domains.

@lanzagar, @astaric?

astaric: I would change transform to compare with is.

nikicc: CMIIW, but changing `Table.from_table` to compare domains with `is` instead of `==` doesn't look like a trivial change. Also it by itself doesn't make `OWColor` work — the problem then arises in DomainConversion, which doesn't seem to know how to do the transformation. As a result, we get a data table with all values missing.

Before, as we checked with `==` if domains matched we called `from_table_rows` which just makes a selection of rows and keeps the provided domain. Now, if `domain1 == domain2` but `domain1 is not domain2` we need to make a domain conversion.

It seems like the core of the problem with all values missing is in how DomainConversion checks whether some variable is inside a domain. The relevant part of the code from `domain.py`:
```
    def __contains__(self, item):
        """
        Return `True` if the item (`str`, `int`, :class:`Variable`) is
        in the domain.
        """
        return item in self._indices
```
It seems like `item in self._indices == False` while `item in list(self._indices.keys()) == True`. Is this a result of how we compare variables?

Should we maybe just revert to `self.data.domain = domain` as it was before @janezd refactoring? If so, @janezd can you provide an explanation, since I'm not quite sure how variable's proxies work?

astaric: The first problem sounds like a broken hash on Variable.

As for the fix, I would rather move the data.domain = domain to `transform`method than have every caller do this explicitly.

nikicc: @astaric but shouldn't then we fix `Table.from_table` as well to prevent `from_table` and `transform` to behave differently? Isn't `data.transform` currently just a syntactic sugar?

astaric: That is also possible. As we have changed all occurrences of Table.from_table with data.transform, I kind of consider it deprecated, but yeah, we can put the change there.

janezd: ... and risk janezd changing it back to `data.transform` the next time he sees this code? If `data.transform` is syntactic sugar it has to have the same behaviour as the thing that this sugar is coating.

nikicc: > ... and risk @janezd changing it back to `data.transform` the next time he sees this code? 

@janezd change to `data.transform` didn't break anything (`transform` and `to_table` behave the same way). Removing of the extra `self.data.domain = domain` is the change that broke things.

> If data.transform is syntactic sugar it has to have the same behaviour as the thing that this sugar is coating.

Agree. Hence, if we are going to fix this in the core we need to do it in `from_table`. And while I'm all for changing `to_table` to compare domains with `is` instead of `==`,  I'm afraid this might cause other issues all over the place. As it has revealed some problems in domain's `__contains__` which I'm not sure how to solve. Any help appreciated 🙈

nikicc: And if anyone wishes to play around, here is the minimal example of the problem, after the change introduced in this PR:
```py
import Orange
from Orange.data import Table


def _create_proxies(variables):
    part_vars = []
    for var in variables:
        if var.is_discrete or var.is_continuous:
            var = var.make_proxy()
            if var.is_discrete:
                var.values = var.values[:]
        part_vars.append(var)
    return part_vars


data = Table('iris')
domain = data.domain
domain = Orange.data.Domain(_create_proxies(domain.attributes),
                            _create_proxies(domain.class_vars),
                            _create_proxies(domain.metas))

new_data = data.transform(domain)
print(new_data)
```

The `new_data` contains all missing values since variables from the destination domain are not in the source domain and compute values are not set.

nikicc:
@astaric  this should work now, please check.

PR #2380, opened 2017-06-06
[FIX] Distances: prevent inf numbers

jerneju:
##### Issue
screenshot_20170606_112022

> ---------------------------- ValueError Exception -----------------------------
> Traceback (most recent call last):
>   File "pyqtgraph/graphicsItems/AxisItem.py", line 525, in paint specs = self.generateDrawSpecs(painter)
>   File "pyqtgraph/graphicsItems/AxisItem.py", line 819, in generateDrawSpecs tickLevels = self.tickValues(self.range[0], self.range[1], lengthInPixels)
>   File "pyqtgraph/graphicsItems/AxisItem.py", line 691, in tickValues num = int((maxVal-start) / spacing) + 1
> ValueError: cannot convert float NaN to integer
> -------------------------------------------------------------------------------

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -34,7 +34,7 @@ from Orange.widgets import widget, gui, settings
 from Orange.widgets.utils import colorpalette, itemmodels
 from Orange.widgets.utils.annotated_data import (create_annotated_table,
                                                  ANNOTATED_DATA_SIGNAL_NAME)
-from Orange.widgets.widget import Input, Output
+from Orange.widgets.widget import Input, Output, Msg
 
 __all__ = ["OWHierarchicalClustering"]
 
@@ -763,6 +763,9 @@ class OWHierarchicalClustering(widget.OWWidget):
     cluster_roles = ["Attribute", "Class variable", "Meta variable"]
     basic_annotations = ["None", "Enumeration"]
 
+    class Error(widget.OWWidget.Error):
+        not_finite_distances = Msg("Some distances are infinite")
+
     def __init__(self):
         super().__init__()
 
@@ -962,10 +965,15 @@ class OWHierarchicalClustering(widget.OWWidget):
     @Inputs.distances
     def set_distances(self, matrix):
         self.error()
+        self.Error.clear()
         if matrix is not None:
             N, _ = matrix.shape
             if N < 2:
                 self.error("Empty distance matrix")
+                matrix = None
+        if matrix is not None:
+            if not np.all(np.isfinite(matrix)):
+                self.Error.not_finite_distances()
                 matrix = None
 
         self.matrix = matrix
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -5,7 +5,7 @@ from collections import namedtuple, OrderedDict
 from itertools import chain
 from contextlib import contextmanager
 
-import numpy
+import numpy as np
 
 from AnyQt.QtWidgets import (
     QGraphicsWidget, QGraphicsObject, QGraphicsLinearLayout, QGraphicsPathItem,
@@ -605,7 +605,7 @@ class DendrogramWidget(QGraphicsWidget):
         else:
             drect = QSizeF(leaf_count, self._root.value.height)
 
-        eps = numpy.finfo(numpy.float64).eps
+        eps = np.finfo(np.float64).eps
 
         if abs(drect.width()) < eps:
             sx = 1.0
@@ -1135,7 +1135,7 @@ class OWHierarchicalClustering(widget.OWWidget):
 
         if isinstance(items, Orange.data.Table) and self.matrix.axis == 1:
             # Select rows
-            c = numpy.zeros(self.matrix.shape[0])
+            c = np.zeros(self.matrix.shape[0])
 
             for i, indices in enumerate(maps):
                 c[indices] = i
--- a/Orange/widgets/unsupervised/tests/test_owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/tests/test_owhierarchicalclustering.py
@@ -2,6 +2,7 @@
 # pylint: disable=missing-docstring
 import numpy as np
 import Orange.misc
+from Orange.data import Table, Domain, ContinuousVariable, DiscreteVariable
 from Orange.distance import Euclidean
 from Orange.widgets.tests.base import WidgetTest, WidgetOutputsTestMixin
 from Orange.widgets.unsupervised.owhierarchicalclustering import \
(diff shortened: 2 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/280206866/

jerneju:
Rebased.

Orange/widgets/unsupervised/owdistances.py
```diff
@@ -160,6 +161,9 @@ def checks(metric, data):
         except MemoryError:
             self.Error.distances_memory_error()
             return
+
+        if np.isinf(met).any():
+            met = np.nan_to_num(met)
```
kernc: Without the conditional, this would make one linear pass instead of three.

Orange/widgets/unsupervised/owdistances.py
```diff
@@ -160,6 +161,8 @@ def checks(metric, data):
         except MemoryError:
             self.Error.distances_memory_error()
             return
+
+        met = np.nan_to_num(met)
```
janezd: No. This is just a quick local patch, not a solution. 

It reminds me of a guy (whom we all know and) who - years ago - programmed some machine learning algorithm. He occasionally got probabilities above 1, so he fixed the problem by adding

    if (p > 1) {
        p = 1;
    }

If we get an error report, we shouldn't find the simplest way to just fix it somewhere. If some distances are infinite, they are infinite. We can't just set them to 42 and pretend everything is cool. If we believe that distance matrix should never contain infinite values, we should raise an exception. If we decide that infinite values can arise and are legal, other functions/widgets should be able to handle those.

In this case, I'd say that infinite values are allowed and hierarchical clustering should do something about them. Replacing them with large values (perhaps `len(data) * np.nanmax(data)`) could make sense within the hierarchical clustering. Or it may be even more complex since it depends upon linkage.

So: if there are infinite distances, we need to think what would we like the clustering to do with them, not what is the easiest way to get rid of them.

nikicc: > perhaps len(data) * np.nanmax(data)

If you opt for this suggestion, please, just don't use `np.nanmax(data)`. We should never use `np.nanmax(data)` since it doesn't work on sparse data! We have our own implementation of `nanmax` in `Orange/statistics/util.py`.

lanzagar: I was about to write that while I totally agree with @janezd, most solutions for inf distances will probably be equivalent to substitution with large numbers. In which case maybe we should consider if we want every widget to duplicate this substitution.
But there is always the possibility that somewhere inf numbers would be illegal (and we would want a warning). And even worse - this substitution will probably not allow downstream methods to ignore the problem, since you still cannot multiply the new "large values", or add two of them etc. which could happen in average linkage and so on.

jerneju: @janezd, **Distance Transformation** widget could have a check box where user could select to impute infinite distances. @lanzagar suggested something like that.

janezd: This sounds more like clipping. With lower and upper range?

Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -763,6 +763,9 @@ class Outputs:
     cluster_roles = ["Attribute", "Class variable", "Meta variable"]
     basic_annotations = ["None", "Enumeration"]
 
+    class Error(widget.OWWidget.Error):
+        not_finite_distances = Msg("Not all distances are finite")
```
janezd: Rephrase: Some distances are infinite.

Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -970,6 +974,9 @@ def set_distances(self, matrix):
 
         self.matrix = matrix
         if matrix is not None:
+            if not np.all(np.isfinite(matrix)):
+                self.Error.not_finite_distances()
+                return
```
janezd: I may have overlooked something, but I think this is wrong: it aborts the function without reseting the items (and related models). Note that the check above sets `matrix` to `None`. In line with this, you should probably have something like:

        if matrix is not None:
            if not np.all(np.isfinite(matrix)):
                self.Error.not_finite_distances()
                matrix = None

        self.matrix = matrix
        if matrix is not None:
            self._set_items(matrix.row_items, matrix.axis)
        else:
            self._set_items(None)

The line `self.matrix = matrix` is moved from above to the place after the check for infinite values.

jerneju: Ok, fixed.

Orange/widgets/unsupervised/owhierarchicalclustering.py
jerneju: Why some error messages have dot at the end and some have not?

janezd: You're right, we have to decide. @BlazZupan, @astaric, @lanzagar, @ajdapretnar? I don't have a clear preference.

Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -962,11 +965,15 @@ def axis_view(orientation):
     @Inputs.distances
     def set_distances(self, matrix):
         self.error()
+        self.Error.clear()
         if matrix is not None:
             N, _ = matrix.shape
             if N < 2:
                 self.error("Empty distance matrix")
                 matrix = None
+            if not np.all(np.isfinite(matrix)):
```
janezd: No, not like that. If it comes here, `matrix` can be set to `None` in the line above. Another `if matrix is not None` is needed here.

Orange/widgets/unsupervised/owhierarchicalclustering.py
ajdapretnar: I vote for dots in sentences (verb present).

janezd: Good point. No verb, no dot. Verb, dot. Period. :)

However, some error messages are not sentences, like "Out of memory" or "Error during optimization". What about those?

Message without verb bad, message with verb good? Makes them longer, though. Sometimes long not good, when widget narrow.

ajdapretnar: Oh no, I would keep those dot-less. Short, too.

Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -1082,6 +1090,8 @@ def _update_labels(self):
         self.labels.setMinimumWidth(1 if labels else -1)
 
     def _invalidate_clustering(self):
+        if self.matrix is not None and not np.all(np.isfinite(self.matrix)):
+            return
```
janezd: First, this check may be redundant since `np.all(np.isfinite(self.matrix))` is already checked above and `self.matrix` is set to None in this case. Hence, the situation in which this conjunction is true cannot arise, I guess.

Second -- if this condition can be fulfilled -- is skipping these calls really OK?

- `_update()` clears the plot and then performs further cleanup if `self.matrix is None`. I guess this must be done if the widget receives invalid data.
- `_update_labels()` checks for `self.root ` (which is set to `None` in `_update()`) and works correctly when `matrix` is `None`.
- `_invalidate_output()` has to be called pass `None` to outputs.

I believe that these two lines have to be removed. But please check whether there are any other side-effects.

jerneju: These two lines are now removed.

Orange/widgets/unsupervised/owhierarchicalclustering.py
janezd: This is Ok now.

PR #2381, opened 2017-06-06
[FIX] Scatter Plot: two minor errors

jerneju:
##### Issue
screenshot_20170606_120406

Put _Scatter Plot_ widget on the canvas. Then try to change shape. Select **(Same shape)**.

screenshot_20170606_120429

Put _Scatter Plot_ widget on the canvas.  Then do something. For instance, try to change color. **Find Informative Projections** button enables and it crashes when an user starts vizrank.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -323,8 +323,9 @@ class OWScatterPlot(OWWidget):
         self.vizrank.initialize()
         self.vizrank.attrs = self.data.domain.attributes if self.data is not None else []
         self.vizrank_button.setEnabled(
-            self.data is not None and self.data.domain.class_var is not None
-            and len(self.data.domain.attributes) > 1 and len(self.data) > 1)
+            self.data is not None and not self.data.is_sparse() and
+            self.data.domain.class_var is not None and
+            len(self.data.domain.attributes) > 1 and len(self.data) > 1)
         if self.data is not None and self.data.domain.class_var is None \
             and len(self.data.domain.attributes) > 1 and len(self.data) > 1:
             self.vizrank_button.setToolTip(
@@ -423,7 +424,6 @@ class OWScatterPlot(OWWidget):
                             new=False)
 
     def sparse_to_dense(self, input_data=None):
-        self.vizrank_button.setEnabled(not (self.data and self.data.is_sparse()))
         if input_data is None or not input_data.is_sparse():
             return input_data
         keys = []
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/288041149/

jerneju:
https://sentry.io/biolab/orange3/issues/288034131/

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -989,7 +989,7 @@ def update_shapes(self):
         self.make_legend()
 
     def assure_attribute_present(self, attr):
-        if attr not in self.data.domain:
+        if self.data and attr not in self.data.domain:
```
janezd: Use `self.data is not None`.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -322,9 +322,7 @@ def set_data(self, data):
             self.init_attr_values()
         self.vizrank.initialize()
         self.vizrank.attrs = self.data.domain.attributes if self.data is not None else []
-        self.vizrank_button.setEnabled(
-            self.data is not None and self.data.domain.class_var is not None
-            and len(self.data.domain.attributes) > 1 and len(self.data) > 1)
+        self._vizrank_button_enabled()
```
janezd: I suppose that the bug that you are fixing here is that the button was not enabled when it should have been. This shouldn't be "fixed" so that the button is always enabled. These conditions were correct, so the bug -- the reason that this line was not executed -- must be somewhere else.

You should find the actual error and fix it instead of removing the code that (I suppose) works.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -423,7 +421,7 @@ def prepare_data(self):
                             new=False)
 
     def sparse_to_dense(self, input_data=None):
-        self.vizrank_button.setEnabled(not (self.data and self.data.is_sparse()))
+        self._vizrank_button_enabled()
```
janezd: Same here.

Orange/widgets/visualize/owscatterplot.py
jerneju: Well, I moved code somewhere else. Now I have put it back here and added check for sparse.

Orange/widgets/visualize/owscatterplot.py
jerneju: Fixed.

Orange/widgets/visualize/owscatterplotgraph.py
jerneju: Fixed.

janezd:
What should I do to reproduce the error? Please provide actual descriptions; these screenshots are not informative.

jerneju:
@janezd : Instructions to reproduce these two errors are now added above.

jerneju:
https://sentry.io/biolab/orange3/issues/300025410/

PR #2342, opened 2017-05-25
Core widgets: new signals

jerneju:
##### Issue
Cannot connect core Orange widgets to text add-on widgets which accept corpus.

##### Description of changes
Something similar as in #2312 .

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #2382, opened 2017-06-06
Widget: new signals: Part II

jerneju:
##### Issue
#2342 

##### Description of changes
- [X] Paint Data
- [X] Data Info
- [X] Concatenate
- [X] Preprocess
- [X] Impute
- [X] Outliers
- [x] Python Script
- [x] Image Viewer
- [x] Color
- [x] Continuize
- [x] Discretize
- [x] Feature Constructor
- [x] Purge Domain
- [x] Save Data
- [x] File
- [x] SQL Table
- [x] CN2 Rule Viewer
- [x] Nomogram
- [x] Linear Regression
- [x] Logistic Regression
- [x] Save Model
- [x] Load Model
- [x] AdaBoost
- [x] ROC Analysis
- [x] Lift Curve
- [x] Calibration Plot
- [x] Distance File
- [x] PCA
- [x] Correspondence Analysis
- [x] Distances
- [x] Distance Transformation
- [x] Save Distance Matrix

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owconcatenate.py
+++ b/Orange/widgets/data/owconcatenate.py
@@ -19,6 +19,7 @@ from Orange.widgets import widget, gui, settings
 from Orange.widgets.settings import Setting
 from Orange.widgets.utils.annotated_data import add_columns
 from Orange.widgets.utils.sql import check_sql_input
+from Orange.widgets.widget import Input, Output
 
 
 class OWConcatenate(widget.OWWidget):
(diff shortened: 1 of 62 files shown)


Orange/widgets/data/owconcatenate.py
```diff
@@ -126,10 +131,12 @@ def __init__(self):
         box.layout().insertSpacing(1, 20)
 
     @check_sql_input
+    @Inputs.primary_data
```
jerneju: If I reverse the order of decorators, the widget crashes.

janezd: Gosh, this is so annoying. The order in the PR, while essentially incorrect, works because `@Inputs.<name>` decorator does not really wrap the function but just stores its name. On the other hand, the (essentially) correct order does not work because the `check_sql_input` decorator changes the method's `__name__` to `new_f`, and so `@Inputs` stores `new_f` as the signal handler.

Please change the order anyway, so we're prepared for potential future changes in signal handling. To prevent crashes, fix the `check_sql_input` decorator. In the definition of the decorator, decorate `new_f` with `wraps`, like this:

    @wraps(f)
    def new_f(widget, data, *args, **kwargs):

This will rename `new_f` to the name of the wrapped method, so `@Inputs` will see the correct name.

jerneju: Ok.

lanzagar:
I don't know what is up with AppVeyor again, but the test (last commit) looks good to me. @janezd vouches for the rest so I guess this can be merged :)

PR #2390, opened 2017-06-08
[WIP] [FIX] New signals: default output

jerneju:
##### Issue
screenshot_20170608_152918
Selected Data should be default instead of Features.

screenshot_20170608_153216
Data should be default instead of Features.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectcolumns.py
+++ b/Orange/widgets/data/owselectcolumns.py
@@ -252,7 +252,7 @@ class OWSelectAttributes(widget.OWWidget):
 
     class Outputs:
         data = Output("Data", Table)
-        features = Output("Features", widget.AttributeList, dynamic=False)
+        features = Output("Features", widget.AttributeList)
 
     want_main_area = False
     want_control_area = True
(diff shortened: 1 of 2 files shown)


janezd:
As far as I recall, we said that the problem here is that Data became a dynamic signal, and static signals have precedence over dynamic.

@ales-erjavec, I had an idea that we could change the signal connection policy so that when the signal is dynamic, it still has the same priority as static if the types actually match. That is, if Table is connected to Table (not to Corpus) it remains the default signal although it's dynamic (because it would match even if it was static). I'm not sure I still like it; it's messy. Do you have a better idea?

Another option is to eliminate non-dynamic signals. There are currently almost no static signals. @astaric?

At any rate, while this modification fixes the problem, I don't think it fixes it correctly. If we accept it, we should at least add a comment why the change was necessary.

ales-erjavec:
> @ales-erjavec, I had an idea that we could change the signal connection policy so that when the signal is dynamic, it still has the same priority as static if the types actually match. That is, if Table is connected to Table (not to Corpus) it remains the default signal although it's dynamic (because it would match even if it was static). I'm not sure I still like it; it's messy. Do you have a better idea?

I think this is the best solution. Please see #2431

janezd:
Close in favor of #2431.

PR #2391, opened 2017-06-08
OWFile: dialog formats not set in class

markotoplak:
If dlg_formats were set before an object was created, new file formats
from the add-ons were not detected unless OWFile was already cached in
the widget registry.

##### Issue

Fixes https://github.com/markotoplak/orange-infrared/issues/74

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -17,7 +17,7 @@ from Orange.widgets.settings import Setting, ContextSetting, \
     PerfectDomainContextHandler, SettingProvider
 from Orange.widgets.utils.domaineditor import DomainEditor
 from Orange.widgets.utils.itemmodels import PyListModel
-from Orange.widgets.utils.filedialogs import RecentPathsWComboMixin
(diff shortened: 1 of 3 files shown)


janezd:
@markotoplak, using `property` was meant as a quick fix, just to avoid having to change any other code.

I think `dlg_formats` should be a private method with `lru_cache`, or a private class method without `lru_cache`. AFAIK, it is used only in a single place in the widget.

markotoplak:
@janezd Agree. Does it have to be private? I think it is an useful public method because it allows me to make a similar (multiple) file opener.

janezd:
I'm not too keen on widgets importing other widgets. Sieve and Mosaic were coupled in this way, and it was a mess until we extracted the common functions into a separate module.

If we have a use case for this, I would suggest extracting the whole method into a separate module. If we don't have it yet, let's make it a private method until we do. :)

markotoplak:
I have a use case in infrared multiple file reader, so I added it to utils.filedialogs.

I did not cache anything as calling this function is fast.

PR #2392, opened 2017-06-09
[FIX] Logistic Regression: impute

jerneju:
##### Issue
screenshot_20170609_115223

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -117,6 +117,7 @@ class Learner(_ReprableWithPreprocessors):
 
         self.domain = data.domain
         model = self._fit_model(data)
+        model.used_vals = [np.unique(y) for y in data.Y[:, None].T]
         model.domain = data.domain
         model.supports_multiclass = self.supports_multiclass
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/265497692/

Orange/base.py
```diff
@@ -114,9 +114,9 @@ def __call__(self, data):
         if len(data.domain.class_vars) > 1 and not self.supports_multiclass:
             raise TypeError("%s doesn't support multiple class variables" %
                             self.__class__.__name__)
-
         self.domain = data.domain
         model = self._fit_model(data)
+        model.data = data
```
lanzagar: Can we avoid this. We already have `model.original_data` below which we would like to get rid of.
Every model should not store the complete data set.

jerneju: Now fixed in a different way.

Issue #2388, opened 2017-06-08
URL text box does not add "http://" or "https://" automatically


jerneju:
##### Orange version
master

##### Expected behavior
A user only writes URL address without _http://_ or _https://_ and that prefix is later added automatically.

##### Actual behavior
File widget throws an error: 

> unknown url type

PR #2393, opened 2017-06-09
io, OWFile: if URL without scheme, assume http by default

kernc:
##### Issue
https://github.com/biolab/orange3/issues/2388

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -923,6 +923,12 @@ class DotReader(FileFormat):
 
 
 class UrlReader(FileFormat):
+    def __init__(self, filename):
+        filename = filename.strip()
+        if not urlparse(filename).scheme:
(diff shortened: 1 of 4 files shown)


jerneju:
screenshot_20170609_122356

kernc:
~~Should~~Might work now.

lanzagar:
Tests fail. Please fix.

Test with more urls.

Why lstrip?

kernc:
> Tests fail. Please fix.

I don't know why the test fails. Please help.

> Test with more urls.

What other URLs are reasonable to test with?

astaric:
The failing test on python3.4 was caused by a call to a method that has been added in Qt5. After replacing the call the test passes (locally), let's see what happens on travis.

PR #2394, opened 2017-06-09
[FIX] ROC Analysis: color support for more than 9 evaluation learners

jerneju:
##### Issue
screenshot_20170609_123758

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owrocanalysis.py
+++ b/Orange/widgets/evaluate/owrocanalysis.py
@@ -448,7 +448,10 @@ class OWROCAnalysis(widget.OWWidget):
             names = ["#{}".format(i + 1)
                      for i in range(len(results.predicted))]
 
-        self.colors = colorpalette.ColorPaletteGenerator(len(names))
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/210623933/

Orange/widgets/evaluate/owrocanalysis.py
```diff
@@ -448,8 +448,7 @@ def _initialize(self, results):
             names = ["#{}".format(i + 1)
                      for i in range(len(results.predicted))]
 
-        self.colors = colorpalette.ColorPaletteGenerator(
-            len(names), colorbrewer.colorSchemes["qualitative"]["Dark2"])
+        self.colors = colorpalette.ColorPaletteGenerator(len(names))
```
janezd: The palette that we had before could have been chosen for a reason. It should be used if possible; switch to a more generic palette only if needed.

PR #2395, opened 2017-06-09
[ENH] Impute widget: Parallel execution in the background

ales-erjavec:
##### Issue

Imputer widget blocks the main GUI thread.

##### Description of changes

Move the column imputer fitting into background thread[s].

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
Great, thanks!

Model based imputation on adult is now 30s instead of 55s on my machine (one variable takes more time than others). 

Average imputation of GDS2808 is now 140s instead of 400s + computation can be interrupted.

PR #2397, opened 2017-06-12
[FIX] CN2 Rule Induction: handling memory error

jerneju:
##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owrules.py
+++ b/Orange/widgets/model/owrules.py
@@ -322,13 +322,17 @@ class OWRuleLearner(OWBaseLearner):
         """
         Reimplemented from OWBaseLearner.
         """
+        self.Error.out_of_memory.clear()
+        self.model = None
         if self.check_data():
-            self.model = self.learner(self.data)
-            self.model.name = self.learner_name
-            self.model.instances = self.data
-            self.valid_data = True
-        else:
-            self.model = None
+            try:
+                self.model = self.learner(self.data)
+            except MemoryError:
+                self.Error.out_of_memory()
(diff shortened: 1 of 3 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/292584416/

Orange/widgets/model/owrules.py
```diff
@@ -322,13 +322,18 @@ def update_model(self):
...
         if self.check_data():
-            self.model = self.learner(self.data)
-            self.model.name = self.learner_name
-            self.model.instances = self.data
-            self.valid_data = True
-        else:
-            self.model = None
+            try:
+                self.model = self.learner(self.data)
+                print(type(self.learner))
```
lanzagar: This is probably a leftover debug print.

janezd:
@jerneju, can you perhaps finish this? That is, discover how CN2 runs out of memory?

If you don't succeed, let's merge this PR.

jerneju:
I did some tests using `memory_profiler`. I used data from _adult.tab_. Here are the results:
Number of used nominal features   Memory consumption
1                                                      0.0 MiB
2                                                      4.9 MiB
3                                                      14.7 MiB
4                                                      85.2 MiB

janezd:
It will always be possible to cause MemoryError in CN2 even with reasonably sized data by using unreasonable settings. Even if CN2 can be optimized, the checks in this PR won't hurt.

We're thus merging this, and leave any optimization of CN2 as further work.

PR #2401, opened 2017-06-13
[FIX] Update checker: LooseVersion doesn't handle str parts

kernc:
##### Issue
```py
>>> from distutils.version import LooseVersion
>>> LooseVersion('3.4.dev') <= LooseVersion('3.4.1')
--------------------------------------------------------------------------------
TypeError: unorderable types: str() < int()
```

##### Description of changes
Use newer `package_resources.SetuptoolsVersion` which doesn't exhibit this problem.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -124,7 +124,6 @@ def check_for_updates():
         settings.setValue('startup/last-update-check-time', int(time.time()))
 
         from urllib.request import urlopen
-        from distutils.version import LooseVersion
         from Orange.version import version as current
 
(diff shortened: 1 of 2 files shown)


kernc:
https://sentry.io/biolab/orange3/issues/293383178/

astaric:
Are you sure this one handles everything? Would it be possilble to add some tests?

astaric:
PS. Could you also add a trailing / to the URL used to check the version (to avoid unnecessary redirects)?

PR #2403, opened 2017-06-14
[ENH] SQL: save user name and password via credentials manager

jerneju:
##### Issue
User name and password should not be saved in Settings.

##### Description of changes
Use credentials manager.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owsql.py
+++ b/Orange/widgets/data/owsql.py
@@ -264,9 +264,12 @@ class OWSql(OWWidget):
                         "Specify a table name to materialize the query")
                     return
                 try:
-                    with self.backend.execute_sql_query("DROP TABLE IF EXISTS " + self.materialize_table_name):
+                    with self.backend.execute_sql_query("DROP TABLE IF EXISTS " +
+                                                        self.materialize_table_name):
                         pass
-                    with self.backend.execute_sql_query("CREATE TABLE " + self.materialize_table_name + " AS " + self.table):
+                    with self.backend.execute_sql_query("CREATE TABLE " +
+                                                        self.materialize_table_name +
+                                                        " AS " + self.table):
                         pass
                     with self.backend.execute_sql_query("ANALYZE " + self.materialize_table_name):
                         pass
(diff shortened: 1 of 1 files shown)


astaric:
Not sure how this is supposed to work since username/password depend on the server/database.

To reproduce, create a workflow that connects to two databases using different credentials. Save it, reload it and see what happens.

jerneju:
@astaric: Well, besides username and password also server name/address/other could be saved. And not just one, more of them.

Orange/widgets/data/owsql.py
```diff
@@ -50,12 +51,15 @@ class OWSql(OWWidget):
     want_main_area = False
     resizing_enabled = False
 
+    cm_sql_username = CredentialManager("SQL username")
+    cm_sql_password = CredentialManager("SQL password")
```
kernc: Could this be, as per CredentialManager's docstring, a single `cm = CredentialManager('SQL Table')` line, further below used as `self.cm.username`, `self.cm.password`?

kernc:
In that case, `cm` would need to be constructed per input values: `cm = CredentialManager('SQL - Table - {hostname}:{port}'.format(...))`. Credentials wouldn't propagate with saved workflows.

What's the issue with Settings again?

Orange/widgets/data/owsql.py
jerneju: What about `"SQL Table: {}:{}".format(self.host, self.port)`

jerneju:
@kernc 
> What's the issue with Settings again?
Error reporting sends workflow with that data.

astaric:
Looks ok to me. Could you also add a migration that moves credentials from widget to Credentials Manager (it will be used when loading an old workflow).

Bonus points if you add some tests :)

Orange/widgets/data/owsql.py
```diff
@@ -55,8 +55,8 @@ class Outputs:
     port = Setting(None)
     database = Setting(None)
     schema = Setting(None)
-    username = ""
-    password = ""
+    username = Setting(None)
```
astaric: What about adding a migrate_settings (class) method and storing the username and password to credentials manager there?

jerneju: Ok, I will do that.

jerneju: Fixed.

PR #2367, opened 2017-06-02
clustering/hierarchical: Check the input linkage matrix

ales-erjavec:
##### Description of changes

Validate the the input linkage matrix and raise an error if invalid

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

PR #2338, opened 2017-05-24
[ENH] New windows installer scripts

ales-erjavec:
##### Issue

Improve windows installers.

##### Description of changes

There are now two installer flavors. One for standard python.org distribution and one for (Ana|Mini)conda distribution. They both create/install into a separate environment.

General improvements:
* Better (un)installer logging (log is left in $TEMP)
* Validate target directory
* Helper launchers to execute application attached to the console with max log level
* Helper launchers to open Command Prompt with activated env

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #2408, opened 2017-06-16
Requirements: scipy version bump to 0.16.1

markotoplak:
Scipy 0.16.1 is installed with the 32-bit Windows installer. Bumping
as we do not test with older scipy anyway.

### DIFF
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -1,5 +1,5 @@
 numpy>=1.9.0
-scipy>=0.11.0
+scipy>=0.16.1
 scikit-learn>=0.18.1
 bottleneck>=1.0.0
 # Reading Excel files


lanzagar:
So we will still not require 0.19? If we are supporting 0.16+, that means #2367 needs to catch the errors possible in versions <0.19?

markotoplak:
0.16.1 is also OK and there are no 0.19.0 installers that would still work in Windows XP.

See @ales-erjavec comment in #2338.

lanzagar:
Well @ales-erjavec commented that 0.16.1 is not affected, but 0.17 and 0.18 are.
The windows installer is fine (uses ==0.16.1), but the changes here require >=0.16.1 so somebody could have 0.18 and satisfy the dep. while still getting the bug.

PR #2417, opened 2017-06-21
[FIX] Feature Constructor: no fail when no values

jerneju:
##### Issue
Error occurs when _Values_ box is empty.
screenshot_20170621_114851

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -244,7 +244,7 @@ class DiscreteFeatureEditor(FeatureEditor):
     def editorData(self):
         values = self.valuesedit.text()
         values = re.split(r"(?<!\\),", values)
-        values = tuple(v.replace(r"\,", ",").strip() for v in values)
+        values = tuple(filter(None, [v.replace(r"\,", ",").strip() for v in values]))
         return DiscreteDescriptor(
             name=self.nameedit.text(),
             values=values,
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/279049068/

Orange/widgets/data/owfeatureconstructor.py
```diff
@@ -244,7 +244,7 @@ def setEditorData(self, data, domain):
     def editorData(self):
         values = self.valuesedit.text()
         values = re.split(r"(?<!\\),", values)
-        values = tuple(v.replace(r"\,", ",").strip() for v in values)
+        values = tuple(filter(None, [v.replace(r"\,", ",").strip() for v in values]))
```
janezd: You can use generators instead of list comprehension, that is

values = tuple(filter(None, (v.replace(r"\,", ",").strip() for v in values)))

Orange/widgets/data/owfeatureconstructor.py
```diff
@@ -292,7 +292,7 @@ def data(self, index, role=Qt.DisplayRole):
             return super().data(index, role)
 
 
-class FeatureConstructorSettingsHandler(DomainContextHandler):
+class FeatureConstructorHandler(DomainContextHandler):
```
janezd: I preferred the original name. In fact, I'd call it `FeatureConstructorContextHandler` and disable the pylint's warning. The current name suggest that it handles the feature constructor, not its contexts.

PR #2418, opened 2017-06-21
[ENH] Python Script: Multiple inputs

### LINKED ISSUE #2324: Multiple inputs/outputs for "Python Script"
##### Orange version 

3.4.dev0+1b1813c

##### Expected behavior
Often times it would be extremely useful to be able to connect several inputs to the Python Script widget and be able to do operations with them to generate one or multiple outputs.

Would it be possible to have _in_data_ and _out_data_ as an array and handle arbitrary number of inputs/outputs?  (Or at least for the input part, because I see how outputting an array can confuse other widget functionality)

##### Actual behavior
Python Script accepts only one input and provides one output channel.

jerneju:
##### Issue
Fixes #2324 .

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpythonscript.py
+++ b/Orange/widgets/data/owpythonscript.py
@@ -357,14 +357,14 @@ class OWPythonScript(widget.OWWidget):
     icon = "icons/PythonScript.svg"
     priority = 3150
 
-    inputs_names = ["in_data", "in_learner", "in_classifier", "in_object"]
+    inputs_names = ("in_data", "in_learner", "in_classifier", "in_object")
 
     class Inputs:
-        in_data = Input("in_data", Orange.data.Table, default=True)
-        # in_distance = Input("in_distance", Orange.misc.SymMatrix, default=True)
-        in_learner = Input("in_learner", Learner, default=True)
-        in_classifier = Input("in_classifier", Model, default=True)
-        in_object = Input("in_object", object)
+        in_data = Input("in_data", Orange.data.Table, default=True, multiple=True)
+        # in_distance = Input("in_distance", Orange.misc.SymMatrix, default=True, multiple=True)
+        in_learner = Input("in_learner", Learner, default=True, multiple=True)
+        in_classifier = Input("in_classifier", Model, default=True, multiple=True)
+        in_object = Input("in_object", object, multiple=True)
 
     outputs_names = ["out_data", "out_learner", "out_classifier", "out_object"]
 
@@ -387,11 +387,11 @@ class OWPythonScript(widget.OWWidget):
     def __init__(self):
         super().__init__()
 
-        self.in_data = None
-        self.in_distance = None
-        self.in_learner = None
-        self.in_classifier = None
-        self.in_object = None
+        self.in_datas = dict()
+        self.in_distances = dict()
+        self.in_learners = dict()
+        self.in_classifiers = dict()
+        self.in_objects = dict()
 
         for s in self.libraryListSource:
             s.flags = 0
@@ -402,7 +402,7 @@ class OWPythonScript(widget.OWWidget):
         gui.label(
             self.infoBox, self,
             "<p>Execute python script.</p><p>Input variables:<ul><li> " + \
-            "<li>".join(name for name in self.inputs_names) + \
+            "<li>".join(name + ', ' + name + 's' for name in self.inputs_names) + \
             "</ul></p><p>Output variables:<ul><li>" + \
             "<li>".join(name for name in self.outputs_names) + \
             "</ul></p>"
@@ -513,25 +513,32 @@ class OWPythonScript(widget.OWWidget):
         self.controlArea.layout().addStretch(1)
         self.resize(800, 600)
 
+    def _set_dict(self, dic, input, id):
+        id = id[0] if isinstance(id, tuple) else id
+        if input is not None:
+            dic[id] = input
+        elif id in dic.keys():
+            del dic[id]
+
     @Inputs.in_data
-    def setExampleTable(self, et):
-        self.in_data = et
+    def setExampleTable(self, et, id=None):
+        self._set_dict(self.in_datas, et, id)
 
     # @Inputs.in_distance
-    def setDistanceMatrix(self, dm):
-        self.in_distance = dm
+    def setDistanceMatrix(self, dm, id=None):
+        self._set_dict(self.in_distances, dm, id)
 
     @Inputs.in_learner
-    def setLearner(self, learner):
-        self.in_learner = learner
+    def setLearner(self, learner, id=None):
+        self._set_dict(self.in_learners, learner, id)
 
     @Inputs.in_classifier
-    def setClassifier(self, classifier):
-        self.in_classifier = classifier
+    def setClassifier(self, classifier, id=None):
+        self._set_dict(self.in_classifiers, classifier, id)
 
     @Inputs.in_object
-    def setObject(self, obj):
-        self.in_object = obj
+    def setObject(self, obj, id=None):
+        self._set_dict(self.in_objects, obj, id)
 
     def handleNewSignals(self):
         self.unconditional_commit()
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owpythonscript.py
```diff
@@ -645,7 +667,7 @@ def saveScript(self):
             f.close()
 
     def initial_locals_state(self):
-        d = dict([(i.name, getattr(self, i.name, None)) for i in self.inputs])
+        d = dict([(name, getattr(self, name, None)) for name in self.inputs_names])
```
janezd: I think it'd be better to pass lists, not dicts. The script doesn't know the meaning of id's, so it would always use just the values.

Orange/widgets/data/owpythonscript.py
```diff
@@ -386,10 +393,14 @@ def __init__(self):
         super().__init__()
 
         self.in_data = None
+        self.in_datas = OrderedDict()
```
janezd: Any reason for using ordered dicts here?

jerneju: No.

Orange/widgets/data/owpythonscript.py
jerneju: I completely forgot about this. Well, lists cannot be passed. There needs to be a dictionary since some other code (`class InteractiveInterpreter`) does not accept lists.

Orange/widgets/data/owpythonscript.py
```diff
@@ -513,25 +520,36 @@ def __init__(self):
         self.controlArea.layout().addStretch(1)
         self.resize(800, 600)
 
+    def _set_dict(self, dic, input, id):
+        id = id[0] if isinstance(id, dict) else id
```
janezd: Can id be a dict? (And why would this dict have a key `0`?) As I can see, it can be a tuple and it's OK if it remains a tuple.

Orange/widgets/data/owpythonscript.py
janezd: Where is this code? I can't find `InteractiveInterpreter` (I may be blind).

If it stays a dictionary, use dict comprehension, i.e. `{name: getattr(self, name, None) for name in self.inputs_names}`.

jerneju: https://docs.python.org/3/library/code.html

> This class deals with parsing and interpreter state (the user’s namespace); ... The optional locals argument specifies the **dictionary** in which code will be executed;

Orange/widgets/data/owpythonscript.py
```diff
@@ -513,25 +520,36 @@ def __init__(self):
...
+        if input is not None:
+            dic[id] = input
+        elif id in dic.keys():
+            del dic[id]
+
     @Inputs.in_data
-    def setExampleTable(self, et):
+    def setExampleTable(self, et, id=None):
         self.in_data = et
```
janezd: This is not OK: `self.in_data` is going to contain the last received signal. If I recall correctly, we said that `in_data` is going to contain some data if exactly one data is present. If there are more, `in_data` should be `None`; it there are multiple inputs, the content of `in_data` would be picked essentially random and thus useless.

Perhaps the cheapest solution is to make it a property. Remove this line and `self.in_data = None`, and add

@property
def in_data(self):
    if len(self.in_datas) == 1:
        return next(self.in_datas.itervalues())

Or, cleaner, fix `initial_locals_state` to set `in_data` to `None` if there are multiple signals.

Same for `in_distance` etc.

Orange/widgets/data/owpythonscript.py
janezd: No, no, `in_data` has to be in a dictionary, but doesn't itself have to be a dictionary. `locals` is a dict with local variables (name bindings, actually), that is, keys are variable names and values are their values. That is, `d` has to be a dict, but its values can be anything.

Here:

change the `inputs_names` back to just

        inputs_names = ("in_data", "in_learner", "in_classifier", "in_object")

and remove singulars in `__init__`:

        self.in_data = {}
        self.in_learner = {}
        self.in_classifier = {}
        self.in_object = {}

Change signal handlers to

    def setExampleTable(self, et, id=None):
        self._set_dict(self.in_data, et, id)

    def setDistanceMatrix(self, dm):
        self.in_distance = dm

    def setLearner(self, learner, id=None):
        self._set_dict(self.in_learner, learner, id)

    def setClassifier(self, classifier, id=None):
        self._set_dict(self.in_classifier, classifier, id)

    def setObject(self, obj, id=None):
        self._set_dict(self.in_object, obj, id)

and construct locals with

    def initial_locals_state(self):
        d = dict.fromkeys(o.name for o in self.outputs)
        for name in self.inputs_names:
            value = getattr(self, name, None)
            all_values = list(value.values()) if value is not None else []
            one_value = all_values[0] if len(all_values) == 1 else None
            d[name + 's'] = all_values
            d[name] = one_value
        return d

If I haven't overlooked something, this gives the behaviour I wanted. Please check, and also see whether `id` can be a dict (my other comment).

janezd: Ooops

    def setDistanceMatrix(self, dm):
        self._set_dict(self.in_distance, dm, id)

obviously.

jerneju: I think that `id` used to be a dict but now it is definitely not anymore.

Orange/widgets/data/owpythonscript.py
jerneju: I see. So `in_data` or etc. is `None` when there are more tables or etc.

Orange/widgets/data/owpythonscript.py
jerneju: And I see that you want to add one additional input: `DistanceMatrix`. That used to be commented out, I have no idea why.

janezd: Actually not, I just changed the (uncommented) code for `DistanceMatrix`. You can keep it commented.

Orange/widgets/data/owpythonscript.py
```diff
@@ -513,25 +513,32 @@ def __init__(self):
         self.controlArea.layout().addStretch(1)
         self.resize(800, 600)
 
+    def _set_dict(self, dic, input, id):
+        id = id[0] if isinstance(id, tuple) else id
```
janezd: Isn't this *always* a tuple?

Orange/widgets/data/owpythonscript.py
```diff
@@ -653,7 +659,12 @@ def saveScript(self):
 
     def initial_locals_state(self):
         d = dict([(name, getattr(self, name, None)) for name in self.inputs_names])
```
janezd: I suppose the dict should be initialized as empty, `d = {}`.

Orange/widgets/data/tests/test_owpythonscript.py
```diff
@@ -68,3 +68,23 @@ def test_wrong_outputs(self):
             self.assertTrue(getattr(self.widget.Error, _output).is_shown())
             self.send_signal(_input, data)
             self.assertFalse(getattr(self.widget.Error, _output).is_shown())
+
+    def test_multiple_data_inputs(self):
```
janezd: Could the test be expanded to test more of widget's functionality, like what happens with these multiple objects?

janezd:
No need to close this - #2506 is just an alternative. :)

It this PR is accepted, please cherry-pick #2506's commit  f0d4a885b29e2253667a604a432988b4fdb88529, which contains the tests useful for this PR as well.

janezd:
#2506 now does the same thing + few other minor fixes.

borondics:
This is a very nice change, thanks! I actually used it today, but had to come back all the way here to find out how it works. Nobody likes to write help - myself included!

borondics:
Ah, I just saw that it is indicated on the left panel of the widget.

PR #2422, opened 2017-06-22
[ENH] Canvas Annotations: Text markup editing

ales-erjavec:
##### Issue

Implements inline text markup editing for canvas annotations. Html, rst and markdown are supported.

Use alt-right click on the annotation to change how the content should be interpreted.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


astaric:
IIUC, the Windows build fails because CommonMark is not available on conda?

ales-erjavec:
It failed because CommonMark was not listed in conda  recipe.

PR #2428, opened 2017-06-23
[FIX] Distributions: show probabilities upon selection

### LINKED ISSUE #2420: Distributions: show probabilities upon selection
##### Orange version

3.5.dev

##### Expected behavior
When selecting Show probabilities the probability line(s) are immediately displayed.

##### Actual behavior
The lines are not immediately displayed. One needs to switch the visualization to see probabilities. Re-drawing the plot should be run upon selection.

File - Distributions. Press Show probabilities.

jerneju:
##### Issue
Fixes #2420 .

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owdistributions.py
+++ b/Orange/widgets/visualize/owdistributions.py
@@ -220,6 +220,8 @@ class OWDistributions(widget.OWWidget):
         self.clear()
         self.warning()
         self.data = data
+        self.distributions = None
+        self.contingencies = None
         if self.data is not None:
             if not self.data:
                 self.warning("Empty input data cannot be visualized")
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/296415020/

Orange/widgets/visualize/owdistributions.py
```diff
@@ -371,8 +371,6 @@ def display_distribution(self):
                 self.tooltip_items.append((self.plot, item))
 
     def _on_relative_freq_changed(self):
-        if not self.distributions:
-            return
```
janezd: I think you should keep these two lines, too.

Connect File -> Distributions to give some data to Distributions. Now remove the data by deleting the connection. Click Show relative frequencies and the widget will crash. These two lines prevent this.

jerneju: No. If we really wanted to keep this check, we would write `if self.distributions is None or not len(self.distributions)`.

However this check should not be here. `self.distributions` can be empty or `None`, but `self.contingencies` may be filled with values and that is the reason that the code below should be executed.

Crash is now prevented in a different way.

PR #2437, opened 2017-06-27
[FIX] Test and Score: Show default columns

janezd:
##### Issue

Test & Score does not show any columns by default.

##### Description of changes

An extensive list of changes (see below) makes sure that the settings are initially filled with the list of columns showed by default.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
AppVeyor failed with sth about credentials manager, which is unrelated to this PR. Help?

lanzagar:
I don't think the failure is because of credentials manager stuff. The tests pass (despite some complaints during the run).
At the end it says `conda-script.py` error. Still looks to be unrelated to this PR though.

PR #2443, opened 2017-06-29
[WIP] [ENH] Orange: telemetry

jerneju:
##### Issue
No data is collected. For example, which widgets and add-ons are used most etc. That could help developers to choose which widgets have higher priority.

##### Description of changes
**Work in progress.**

Now it can be collected:
- user location (country, city, approximate coordinates)
- user id, operating system, platform, python version, modules, etc.
- widget used in each specific scheme
- time between starting a new scheme and closing it
- time when user starts and closes the application

All data are sent only when application is being closed.

Data is sent only if user agrees (TODO).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -116,6 +116,26 @@ def show_survey():
         return question
 
 
+def user_telemetry_agreement():
+    settings = QSettings()
+    user_telemetry_agree = settings.value("startup/user-telemetry-agree", False, type=bool)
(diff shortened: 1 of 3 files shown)


janezd:
Thanks for the PR, but after discussing it with @BlazZupan, I am (temporarily?) closing it. The topic of collecting user data is very sensitive, and thus needs to be discussed and agreed upon within the core group before starting the implementation.

PR #2448, opened 2017-07-03
MDS: use Scatter Plot Graph instead of (Mds)plotutils

jerneju:
##### Issue
Too much duplicated code.

##### Description of changes
MDS now uses Scatter Plot Graph instead of (Mds)plotutils.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -1,38 +1,28 @@
 import sys
 
-from xml.sax.saxutils import escape
-from itertools import chain
-
-import pkg_resources
-
-import numpy
+import numpy as np
 import scipy.spatial.distance
 
-from AnyQt.QtWidgets import (
-    QFormLayout, QHBoxLayout, QGroupBox, QToolButton, QActionGroup, QAction,
-    QApplication
-)
-from AnyQt.QtGui import (
-    QColor, QPen, QBrush, QPainter, QKeySequence, QCursor, QIcon
-)
+from AnyQt.QtWidgets import QFormLayout, QApplication
+from AnyQt.QtGui import QPainter
 from AnyQt.QtCore import Qt, QTimer
 
 import pyqtgraph as pg
-import pyqtgraph.graphicsItems.ScatterPlotItem
 
 import Orange.data
+from Orange.data import Domain, Table, ContinuousVariable
 import Orange.projection
 from Orange.projection.manifold import torgerson
 import Orange.distance
-from Orange.data.domain import filter_visible
 import Orange.misc
-from Orange.widgets import widget, gui, settings
-from Orange.widgets.utils import colorpalette, itemmodels
+from Orange.widgets import gui, settings
+from Orange.widgets.settings import SettingProvider
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.canvas import report
+from Orange.widgets.visualize.owscatterplotgraph import OWScatterPlotGraph, InteractiveViewBox
 from Orange.widgets.widget import Msg, OWWidget, Input, Output
 from Orange.widgets.utils.annotated_data import (create_annotated_table,
-                                                 ANNOTATED_DATA_SIGNAL_NAME)
+                                                 ANNOTATED_DATA_SIGNAL_NAME, get_unique_names)
 
 
 def stress(X, distD):
@@ -40,27 +30,56 @@ def stress(X, distD):
     D1_c = scipy.spatial.distance.pdist(X, metric="euclidean")
     D1 = scipy.spatial.distance.squareform(D1_c, checks=False)
     delta = D1 - distD
-    delta_sq = numpy.square(delta, out=delta)
+    delta_sq = np.square(delta, out=delta)
     return delta_sq.sum(axis=0) / 2
 
 
-def make_pen(color, width=1.5, style=Qt.SolidLine, cosmetic=False):
-    pen = QPen(color, width, style)
-    pen.setCosmetic(cosmetic)
-    return pen
+class MDSInteractiveViewBox(InteractiveViewBox):
+    def _dragtip_pos(self):
+        return 10, 10
 
 
-class ScatterPlotItem(pg.ScatterPlotItem):
-    Symbols = pyqtgraph.graphicsItems.ScatterPlotItem.Symbols
+class OWMDSGraph(OWScatterPlotGraph):
+    jitter_size = settings.Setting(0)
 
-    def paint(self, painter, option, widget=None):
-        if self.opts["pxMode"]:
-            painter.setRenderHint(QPainter.SmoothPixmapTransform, True)
+    def update_data(self, attr_x, attr_y, reset_view=True):
+        super().update_data(attr_x, attr_y, reset_view=reset_view)
+        for axis in ["left", "bottom"]:
+            self.plot_widget.hideAxis(axis)
+        self.plot_widget.setAspectLocked(True, 1)
 
-        if self.opts["antialias"]:
-            painter.setRenderHint(QPainter.Antialiasing, True)
+    def get_size_index(self):
+        if self.attr_size == "Stress":
+            return -2
+        return super().get_size_index()
 
-        super().paint(painter, option, widget)
+    def compute_sizes(self):
+        def scale(a):
+            dmin, dmax = np.nanmin(a), np.nanmax(a)
+            if dmax - dmin > 0:
+                return (a - dmin) / (dmax - dmin)
+            else:
+                return np.zeros_like(a)
+
+        self.master.Information.missing_size.clear()
+        size_index = self.get_size_index()
+        if size_index == -1:
+            size_data = np.full((self.n_points,), self.point_width,
+                                dtype=float)
+        elif size_index == -2:
+            size_data = scale(stress(self.master.embedding, self.master.effective_matrix))
+            size_data = self.MinShapeSize + size_data * self.point_width
+        else:
+            size_data = \
+                self.MinShapeSize + \
+                self.scaled_data[size_index, self.valid_data] * \
+                self.point_width
+        nans = np.isnan(size_data)
+        if np.any(nans):
+            size_data[nans] = self.MinShapeSize - 2
+            self.master.Information.missing_size(self.attr_size)
+        return size_data
+
 
 #: Maximum number of displayed closest pairs.
 MAX_N_PAIRS = 10000
@@ -80,6 +99,8 @@ class OWMDS(OWWidget):
         selected_data = Output("Selected Data", Orange.data.Table, default=True)
         annotated_data = Output(ANNOTATED_DATA_SIGNAL_NAME, Orange.data.Table)
 
+    settings_version = 2
+
     #: Initialization type
     PCA, Random = 0, 1
 
@@ -93,13 +114,6 @@ class OWMDS(OWWidget):
         ("None", -1)
     ]
 
-    JitterAmount = [
-        ("None", 0),
-        ("0.1 %", 0.1),
-        ("0.5 %", 0.5),
-        ("1 %", 1.0),
-        ("2 %", 2.0)
-    ]
     #: Runtime state
     Running, Finished, Waiting = 1, 2, 3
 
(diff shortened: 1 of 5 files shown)


janezd:
Is this still WIP? Perhaps it would be better to finish the open PRs before starting new ones. Or do you have some open questions here? (If so, post them.)

jerneju:
Not WIP anymore.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -326,40 +249,49 @@ def button(action):
...
+
+    def update_graph(self, reset_view=True, **_):
+        self.graph.zoomStack = []
+        if self.graph.data is None:
+            return
+        self.graph.update_data(self.variable_x, self.variable_y, reset_view)
+
+    def selection_changed(self):
+        self.commit()
+
```
janezd: Doesn't all this same code appear also in the scatter plot widget? @astaric suggested that this code belongs to `OWScatterPlotGraph`.

janezd:
I approve the idea, but some code should probably be moved to `OWScatterPlot`. Also, see whether there will be any conflict with #2536.

Orange/widgets/unsupervised/owmds.py
jerneju: Yes. The same code appears there. So can I make a super class from which MDS and Scatter Plot are inherited?

jerneju:
@janezd: There is no conflict with that PR.

janezd:
Now that #2541 is merged, can you remove some of the code in this widget?

If you do, you can then squash the commits.

jerneju:
@janezd

Orange/widgets/unsupervised/owmds.py
```diff
@@ -157,19 +156,14 @@ def __init__(self):
...
-        self._brush_data = None
-        self._shape_data = None
-        self._size_data = None
-        self._label_data = None
         self._similar_pairs = None
-        self._scatter_item = None
-        self._legend_item = None
-        self._selection_mask = None
-        self._subset_mask = None  # type: Optional[numpy.ndarray]
+        self._subset_mask = []  # type: Optional[np.ndarray]
```
janezd: This used to be `self._subset_mask = None`. Why change it to `[]`? If this is to avoid having to test for `None` in [line 649](https://github.com/biolab/orange3/pull/2448/files#diff-33a6f21b7026aae1b62fbdfca1d53b86R649), you shouldn't set it to `None` in line 341. It is set to `np.ndarray` in 575.

The type hint, `# type: Optional[np.ndarray]`, suggests that this can be `None` or `ndarray`. Not an (empty) list.

This is a mess. Please revert it to how it was before: `None` and `ndarray`, not a list.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -187,135 +181,51 @@ def __init__(self):
         )
 
         form.addRow("Max iterations:",
-                    gui.spin(box, self, "max_iter", 10, 10 ** 4, step=1))
+                    gui.spin(box_optimization, self, "max_iter", 10, 10 ** 4, step=1))
```
janezd: You can break this like this:

        form.addRow(
            "Max iterations:",
            gui.spin(box_optimization, self, "max_iter", 10, 10 ** 4, step=1))

to keep the lines after 80 characters. Do the same in the following lines for consistency.  This format makes the code more readable *in this particular case*, IMHO.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -187,135 +181,51 @@ def __init__(self):
...
 
         form.addRow("Initialization:",
-                    gui.radioButtons(box, self, "initialization",
+                    gui.radioButtons(box_optimization, self, "initialization",
```
janezd: Why? IMHO, such temporary boxes can keep a short name `box`. We all understand it's a temporary into which a "paragraph" of a code puts the controls it's constructing. Making it more "descriptive" does not make the code cleaner, it just makes lines longer.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -651,7 +509,7 @@ def __next_step(self):
             self.unconditional_commit()
             self.__draw_similar_pairs = True
             self._update_plot()
-            self.plot.autoRange(padding=0.1, items=[self._scatter_item])
+            #.plot.autoRange(padding=0.1, items=[self._scatter_item])
```
janezd: Why is this commented? If it is no longer needed, remove. If it needs to be replaced with something, replace.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -664,7 +522,7 @@ def __next_step(self):
             self.progressBarSet(100.0 * progress, processEvents=None)
             self.embedding = embedding
             self._update_plot()
-            self.plot.autoRange(padding=0.1, items=[self._scatter_item])
+            # self.plot.autoRange(padding=0.1, items=[self._scatter_item])
```
janezd: Same.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -684,10 +542,10 @@ def __invalidate_embedding(self):
         if self.initialization == OWMDS.PCA:
             self.embedding = torgerson(X)
         else:
-            self.embedding = numpy.random.rand(len(X), 2)
+            self.embedding = np.random.rand(len(X), 2)
 
         self._update_plot()
-        self.plot.autoRange(padding=0.1, items=[self._scatter_item])
+        # self.plot.autoRange(padding=0.1, items=[self._scatter_item])
```
janezd: Same.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -714,31 +572,15 @@ def handleNewSignals(self):
 
         if self._subset_mask is None and self.subset_data is not None and \
                 self.data is not None:
-            self._subset_mask = numpy.in1d(self.data.ids, self.subset_data.ids)
+            self._subset_mask = np.in1d(self.data.ids, self.subset_data.ids)
 
         self._update_plot()
-        self.plot.autoRange(padding=0.1)
+        # self.plot.autoRange(padding=0.1)
```
janezd: Same.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -748,150 +590,16 @@ def _update_plot(self):
...
-            self._label_data = label_items
-
         emb_x, emb_y = self.embedding[:, 0], self.embedding[:, 1]
-        if self.jitter > 0:
-            _, jitter_factor = self.JitterAmount[self.jitter]
-            emb_x = jitter(emb_x, jitter_factor, rstate=42)
-            emb_y = jitter(emb_y, jitter_factor, rstate=667)
 
-        if self.connected_pairs and self.__draw_similar_pairs:
+        def connected_pairs():
```
janezd: Rename to `connect_pairs`. Perhaps move the initial `if` out of the function, so that `if` controls whether the function `connect_pairs` is called or not.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -1017,336 +689,37 @@ def commit(self):
...
         caption = report.render_items_vert((
-            ("Color", self.color_value != "Same color" and self.color_value),
-            ("Shape", self.shape_value != "Same shape" and self.shape_value),
-            ("Size", self.size_value != "Same size" and self.size_value),
-            ("Labels", self.label_value != "No labels" and self.label_value)))
+            ("Color", name(self.graph.attr_color)),
+            ("Label", name(self.graph.attr_label)),
+            ("Shape", name(self.graph.attr_shape)),
+            ("Size", name(self.graph.attr_size)),
+            ("Jittering", self.graph.jitter_size)))
```
janezd: Color, label, shape and size are not reported when set to None (which is OK). Do the same for jittering. Don't report it when set to 0; add ` %`, when non-zero. That is, do something like:

```
("Jittering", self.graph.jitter_size != 0 and "{} %".format(self.graph.jitter_size))
```

janezd:
Context settings do not work. Give it Iris, set some attribute for Labels, change to Zoo and back to Iris. The setting for labels is lost.

janezd:
Remove coordinate axes.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -157,19 +156,14 @@ def __init__(self):
...
-        self._scatter_item = None
-        self._legend_item = None
-        self._selection_mask = None
-        self._subset_mask = None  # type: Optional[numpy.ndarray]
+        self._subset_mask = []  # type: Optional[np.ndarray]
         self._invalidated = False
         self._effective_matrix = None
 
+        self.variable_x = ContinuousVariable("x")
+        self.variable_y = ContinuousVariable("y")
```
janezd: Rename these to mds-x, mds-y.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -920,69 +628,33 @@ def jitter(x, factor=1, rstate=None):
...
+        connected_pairs()
+        self._update_data(self.variable_x, self.variable_y, True)
+
+    def _update_data(self, variable_x, variable_y, reset_view):
+        self.graph.update_data(variable_x, variable_y, reset_view=reset_view)
+        for axis in ["left", "bottom"]:
+            self.plot.hideAxis(axis)
+        self.plot.setAspectLocked(True, 1)
 
     def commit(self):
```
janezd: This is not a new problem, but this PR is a good place to solve it since this code will be touched anyway.

What if the data already contains attributes "x" and "y" (or "mds-x", "mds-y", as suggested in another comment)? I think the function needs to be changed so that in this case it creates two new attributes with unique names (see  `Orange.widgets.utils.annotated_data.get_next_name`).

Orange/widgets/unsupervised/owmds.py
```diff
@@ -187,135 +181,51 @@ def __init__(self):
...
-        box.layout().addWidget(button(action_pan))
-        box.layout().addSpacing(4)
-        box.layout().addWidget(button(action_reset_zoom))
-        box.layout().addStretch()
-
-        self.controlArea.layout().addWidget(box)
+        self.graph.box_zoom_select(self.controlArea)
 
         box = gui.vBox(self.controlArea, "Output")
         self.output_combo = gui.comboBox(
```
janezd: We just discussed this with @BlazZupan. There is no universal place for such columns. E.g. in PCA, the components should appear as attributes, and the original attributes as metas. Here, however, x and y can be meta attributes. So: in the intereset of space, remove this combo box and always put coordinates as meta attributes.

Meta attributes are first-class citizens in all visualization widgets. In rare cases, *Select columns* can be used to move them to features.

jerneju: Removed. However, many widgets has an option where one can select an output.

Orange/widgets/unsupervised/owmds.py
jerneju: Ok.

Orange/widgets/unsupervised/owmds.py
jerneju: Fixed.

Orange/widgets/utils/annotated_data.py
```diff
@@ -39,13 +44,21 @@ def get_next_name(names, name):
...
         return name
     return "{} ({})".format(name, max(indexes, default=1) + 1)
 
 
+def get_unique_names(names, proposed):
+    if len([name for name in proposed if name in names]):
+        max_index = max([max(get_indexes(names, name),
+                             default=1) for name in proposed], default=1)
+        for i, name in enumerate(proposed):
+            proposed[i] = "{} ({})".format(name, max_index + 1)
```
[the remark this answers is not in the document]
jerneju: Now smallest possible index.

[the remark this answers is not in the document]
janezd: @lanzagar, max index makes sense. If I already have variables `foo`, `foo (1)`, `foo (3)` and `foo (4)`, I want the new one to be `foo (5)`, not `foo (2)`.

You're right that it won't matter much, but I'd follow this logic. `get_next_name` already [does so](https://github.com/jerneju/orange3/blob/2b317d59ff6a0dd1af42da92c2f5a7d94b481c50/Orange/widgets/utils/annotated_data.py#L47). @jerneju, sorry to push you back and forth. :|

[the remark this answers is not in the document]
jerneju: OK, I have reverted it back.

jerneju:
Well, it is true, but this is a separate issue. It happens in a original version too.

lanzagar:
We can leave the fix of similar pairs disappearing for a later PR.
Squash the commits please, and I will merge.

Issue #2280, opened 2017-05-05
Nomogram slow


ajdapretnar:
##### Orange version

3.5.dev

##### Expected behavior
Nomogram computes the results and displays it interactively.

##### Actual behavior
Nomogram recomputes the results on each window resize, which is a pain when dealing with large data tables (e.g. sparse, Corpus).

##### Steps to reproduce the behavior
Corpus (bookexcerpts) - Bag of Words - Logistic Regression - Nomogram.

##### Additional info (worksheets, data, screenshots, ...)
Can we figure out a way to make Nomogram responsive? Incremental?

PR #2449, opened 2017-07-04
Update Nomogram widget

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/2280

##### Description of changes
Fix slowness by only recomputing what is selected/drawn.
Also split main view into three parts.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/tests/test_ownomogram.py
+++ b/Orange/widgets/visualize/tests/test_ownomogram.py
@@ -23,18 +23,19 @@ class TestOWNomogram(WidgetTest):
         cls.lenses = Table("lenses")
 
     def setUp(self):
-        self.widget = self.create_widget(OWNomogram)
-        self.widget.repaint = True
+        self.widget = self.create_widget(OWNomogram)  # type: OWNomogram
 
     def test_input_nb_cls(self):
         """Check naive bayes classifier on input"""
         self.send_signal(self.widget.Inputs.classifier, self.nb_cls)
         self.assertEqual(len([item for item in self.widget.scene.items() if
                               isinstance(item, DiscreteFeatureItem)]),
-                         len([a for a in self.data.domain.attributes]))
+                         min(self.widget.n_attributes,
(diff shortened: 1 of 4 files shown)


kernc:
Thanks! Hopefully resolved by going with this second approach?

Orange/widgets/visualize/tests/test_ownomogram.py
```diff
@@ -23,8 +23,7 @@ def setUpClass(cls):
         cls.lenses = Table("lenses")
 
     def setUp(self):
-        self.widget = self.create_widget(OWNomogram)
-        self.widget.repaint = True
+        self.widget = self.create_widget(OWNomogram)  # type: OWNomogram
```
kernc: The two length-asserting test failures can be mitigated by setting in `setUp()`:
```py
self.widget.display_index = 0  # i.e. display all attributes
```
Anybody has an idea of an alternative solution, and perhaps something for the third failing test as well?

ajdapretnar:
TODO:  check split window since it results in duplication.

ajdapretnar:
The view now seems better than at the last check, but if I can nitpick, it would benefit from an additional 3 mm of extra space above Total. :)

screen shot 2017-07-10 at 09 22 52

However, a more serious issue is that the visualization fails when trying to resize.

  File "/Users/ajda/orange/orange3/Orange/widgets/visualize/ownomogram.py", line 673, in resizeEvent
    self.w.update_scene()
  File "/Users/ajda/orange/orange3/Orange/widgets/visualize/ownomogram.py", line 878, in update_scene
    self.set_feature_marker_values(old_marker_values)
  File "/Users/ajda/orange/orange3/Orange/widgets/visualize/ownomogram.py", line 1069, in set_feature_marker_values
    item.dot.move_to_val(marker_values[i])
IndexError: list index out of range

ajdapretnar:
Now, as far as I'm concerned, it's great!

PR #2450, opened 2017-07-05
[FIX] preprocess: Fix RemoveNaNClasses / Use existing HasClass

lanzagar:
##### Issue

SklLearner methods do not work on SqlTable data because of the RemoveNaNClasses preprocessor.

##### Description of changes
Use an existing (more general) implementation of the RemoveNaNClasses preprocessor - the HasClass filter.
Why do we even have both (functionality looks exactly the same)?

Before merging, we should check whether the RemoveNaNClasses preprocessor can be removed completely and HasClass filter used in its place.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/preprocess.py
+++ b/Orange/preprocess/preprocess.py
@@ -12,7 +12,7 @@ import Orange.data
 from Orange.data.filter import HasClass
 from Orange.preprocess.util import _RefuseDataInConstructor
 from Orange.statistics import distribution
-from Orange.util import Reprable, Enum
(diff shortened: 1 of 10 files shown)


Orange/preprocess/preprocess.py
```diff
@@ -197,6 +198,7 @@ def __call__(self, data):
         return data.transform(domain)
 
 
+@deprecated("Orange.data.filter.HasClas")
```
lanzagar: HasClas -> HasClass

PR #2457, opened 2017-07-08
[FIX] Table: Fix printing data with sparse Y

pavlin-policar:
##### Issue
Table `__str__` method crashed when the Y was sparse.

_Steps to reproduce:_
```
data = Table('iris')
data.X = sp.csr_matrix(data.X)
# Transpose is needed due to the way sparse converts 1d arrays into sparse matrices
data.Y = sp.csr_matrix(data.Y).T
print(data)
```

Also, printing any Table would be missing the final closing `]` e.g.
```
[[0, 1],
 [1, 0]
```
I've fixed this to
```
[[0, 1],
 [1, 0]]
```

##### Description of changes
The `__str__` method no longer crashes on a sparse Y. Also, I've made it so `var_name=value` is printed for any non-zero value and all discrete values. The `var_name` is now also hidden for any target variable (the reasoning being that there are usually not that many target variables, and showing the variable name wouldn't add any clarity to the output).

@nikicc could you please check if this is ok?

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -189,6 +189,8 @@ class Table(MutableSequence, Storage):
     def Y(self, value):
         if len(value.shape) == 1:
             value = value[:, None]
+        if sp.issparse(value) and len(self) != value.shape[0]:
+            value = value.T
         self._Y = value
 
     def __new__(cls, *args, **kwargs):
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -115,15 +115,24 @@ class RowInstance(Instance):
         def sp_values(matrix, variables):
             if not sp.issparse(matrix):
                 return Instance.str_values(matrix[row], variables, limit)
-            begptr, endptr = matrix.indptr[row:row + 2]
-            rendptr = endptr if not limit else min(endptr, begptr + 5)
-            variables = [variables[var]
-                         for var in matrix.indices[begptr:rendptr]]
-            s = ", ".join(
-                "{}={}".format(var.name, var.str_val(val))
-                for var, val in zip(variables, matrix.data[begptr:rendptr]))
-            if limit and rendptr != endptr:
+
+            row_entries, idx = [], 0
+            while idx < len(variables):
+                # Make sure to stop printing variables if we limit the output
+                if limit and len(row_entries) >= 5:
+                    break
+
+                var = variables[idx]
+                if var.is_discrete or matrix[row, idx]:
+                    row_entries.append("%s=%s" % (var.name, var.str_val(matrix[row, idx])))
+
+                idx += 1
+
+            s = ", ".join(row_entries)
+
+            if limit and idx < len(variables):
                 s += ", ..."
+
             return s
 
         table = self.table
@@ -825,7 +834,7 @@ class Table(MutableSequence, Storage):
         return self.X.shape[0]
 
     def __str__(self):
-        return "[" + ",\n ".join(str(ex) for ex in self)
+        return "[" + ",\n ".join(str(ex) for ex in self) + "]"
 
     def __repr__(self):
         head = 5
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -114,6 +114,8 @@ class RowInstance(Instance):
     def _str(self, limit):
         def sp_values(matrix, variables):
             if not sp.issparse(matrix):
+                if matrix.ndim == 1:
+                    matrix = matrix[:, np.newaxis]
                 return Instance.str_values(matrix[row], variables, limit)
 
             row_entries, idx = [], 0
@@ -140,7 +142,7 @@ class RowInstance(Instance):
         row = self.row_index
         s = "[" + sp_values(table.X, domain.attributes)
         if domain.class_vars:
-            s += " | " + sp_values(table._Y, domain.class_vars)
+            s += " | " + sp_values(table.Y, domain.class_vars)
         s += "]"
         if self._domain.metas:
             s += " {" + sp_values(table.metas, domain.metas) + "}"
@@ -1507,10 +1509,8 @@ class Table(MutableSequence, Storage):
 
                 for col_i, arr_i, _ in cont_vars:
                     if sp.issparse(arr):
-                        col_data = arr.data[
-                                   arr.indptr[arr_i]:arr.indptr[arr_i + 1]]
-                        rows = arr.indices[
-                               arr.indptr[arr_i]:arr.indptr[arr_i + 1]]
+                        col_data = arr.data[arr.indptr[arr_i]:arr.indptr[arr_i + 1]]
+                        rows = arr.indices[arr.indptr[arr_i]:arr.indptr[arr_i + 1]]
                         W_ = None if W is None else W[rows]
                         classes_ = classes[rows]
                     else:
--- a/Orange/tests/test_sparse_table.py
+++ b/Orange/tests/test_sparse_table.py
@@ -48,3 +48,29 @@ class InterfaceTest(tabletests.InterfaceTest):
         iris = Table('iris')
         iris.X, iris.Y = csr_matrix(iris.X), csr_matrix(iris.Y)
         str(iris)
+
+    def test_Y_setter_1d(self):
+        iris = Table('iris')
+        assert iris.Y.shape == (150,)
+        iris.Y = csr_matrix(iris.Y)
+        # We expect the Y shape to match the X shape, which is (150, 4) in iris
+        self.assertEqual(iris.Y.shape, (150, 1))
+
+    def test_Y_setter_2d(self):
+        iris = Table('iris')
+        assert iris.Y.shape == (150,)
+        # Convert iris.Y to (150, 1) shape
+        new_y = iris.Y[:, np.newaxis]
+        iris.Y = np.hstack((new_y, new_y))
+        iris.Y = csr_matrix(iris.Y)
+        # We expect the Y shape to match the X shape, which is (150, 4) in iris
+        self.assertEqual(iris.Y.shape, (150, 2))
+
+    def test_Y_setter_2d_single_instance(self):
+        iris = Table('iris')[:1]
+        # Convert iris.Y to (1, 1) shape
+        new_y = iris.Y[:, np.newaxis]
+        iris.Y = np.hstack((new_y, new_y))
+        iris.Y = csr_matrix(iris.Y)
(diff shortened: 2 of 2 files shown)


Orange/data/table.py
```diff
@@ -112,26 +112,32 @@ def __setitem__(self, key, value):
...
+
             return s
 
         table = self.table
         domain = table.domain
         row = self.row_index
         s = "[" + sp_values(table.X, domain.attributes)
         if domain.class_vars:
-            s += " | " + sp_values(table._Y, domain.class_vars)
+            s += " | " + sp_values(table._Y, domain.class_vars, False)
```
nikicc: IMO this isn't OK if there are multiple variables. For example, pass some text data through BoW and move all features to class variables. Then in the print, I only get values, and there is no way to know which value corresponds to which variable.

screen shot 2017-07-10 at 16 21 58


I agree that we can probably skip the label if there is only one sparse variable set as class, but we cannot skip it when there are multiple.

Orange/tests/test_table.py
```diff
@@ -2195,6 +2195,12 @@ class _ExtendedTable(data.Table):
         self.assertIsInstance(data_file, _ExtendedTable)
         self.assertIsInstance(data_url, _ExtendedTable)
 
+    def test_str(self):
+        try:
```
nikicc: Why is this better than simply `str(self.table)` without the try-except block?

Orange/data/table.py
```diff
@@ -112,26 +112,32 @@ def __setitem__(self, key, value):
...
+            max_idx, has_more_columns = min(5, columns), columns > 5
+
             s = ", ".join(
-                "{}={}".format(var.name, var.str_val(val))
-                for var, val in zip(variables, matrix.data[begptr:rendptr]))
-            if limit and rendptr != endptr:
+                ("%s=" % var.name if (var.is_discrete or matrix[row, idx]) and
+                                     show_labels else "")
+                + var.str_val(matrix[row, idx])
+                for var, idx in zip(variables, range(max_idx)))
```
nikicc: What about `for idx, var in enumerate(variable[:max_idx])`?

pavlin-policar: Much better, thanks!

Orange/tests/test_table.py
pavlin-policar: It isn't.

Orange/data/table.py
pavlin-policar: Do we want to show all the class labels, even if they are 0?

Orange/tests/test_table.py
astaric: If I add this test (without changing the problematic code) to master, all tests pass. Could you add a test that fails until the error is fixed?

Orange/data/table.py
nikicc: For continuous variables probably not; Table also doesn't show those.

Orange/tests/test_table.py
pavlin-policar: Fixed. Tbh I have no idea what's the real issue with the old code, it seems to be some complicated indexing issue back from when sparse matrices didn't support numpy style indexing.

Orange/data/table.py
```diff
@@ -112,26 +112,33 @@ def __setitem__(self, key, value):
...
+            # Sparse matrices can't handle slices where an index is out of
+            # bounds like numpy arrays can, so we must determine largest index
+            _, columns = matrix.shape
+            max_idx, has_more_columns = min(5, columns), columns > 5
+
             s = ", ".join(
-                "{}={}".format(var.name, var.str_val(val))
-                for var, val in zip(variables, matrix.data[begptr:rendptr]))
-            if limit and rendptr != endptr:
+                ("%s=" % var.name if (var.is_discrete or matrix[row, idx]) and
```
nikicc: Not sure how I only spotted this now, but this only conditions the variable name on the value `matrix[row, idx]` being non-zero; while the value itself is still printed. Check this:
screen shot 2017-07-11 at 14 15 05

pavlin-policar: Is this undesirable? Couldn't the values be different from 1, and we'd want to show that?

nikicc: We probably misunderstood. I wasn't objecting to the `interface=1.000` part. The problem is in the following `0.000` part for which I do not know which feature it represents. Also, we should probably skip showing zeros in sparse data altogether (which we do not in the above example).

I suggest we stick to the format that is used in Table. That is, for all values that are non-zero we show `<feature>=<value>`. We skip printing features with zero values altogether, except for discrete features when we print `<feature>=<value>` pair regardless of the value.

pavlin-policar:
I've now also changed the Table.Y property setter to properly handle a sparse row. This means we can now do
```
iris.Y = sp.csr_matrix(iris.Y)
```
instead of having to do
```
iris.Y = sp.csr_matrix(iris.Y).T
```

The reason this was happening is that converting a 1d numpy array to a sparse matrix would produce a matrix with shape (1, 150) for iris, whereas we really want a shape of (150, 1).

This is also consistent with the way a (150, 2) array is cast to sparse, which does produce a sparse matrix with shape (150, 2).

Orange/data/table.py
pavlin-policar: Is it ok now?

Orange/data/table.py
```diff
@@ -178,6 +192,8 @@ def Y(self):
     def Y(self, value):
         if len(value.shape) == 1:
             value = value[:, None]
+        if sp.issparse(value) and value.shape[0] == 1:
```
nikicc: Hmmm, this might be a bit more complicated. Currently, this doesn't work for tables with only one instance but multiple targets. 
```py
>>> data = Table('iris')[:1]

>>> data.X = sp.csr_matrix(data.X)
>>> multiple_targets = np.hstack((data.Y[:, None], data.Y[:, None]))
>>> data.Y = sp.csr_matrix(multiple_targets)

>>> print(data.X.shape)
(1, 4)
>>> print(multiple_targets.shape)
(1, 2)
>>> print(data.Y.shape)
(2, 1)
```
Maybe something like this will do? 
`sp.issparse(value) and len(self) != value.shape[0] and value.shape[1] == len(self)`?

Orange/data/table.py
```diff
@@ -112,26 +112,40 @@ def __setitem__(self, key, value):
...
+            _, columns = matrix.shape
+            max_idx, has_more_columns = min(5, columns), columns > 5
+
+            row_entries = []
+            for idx, var in enumerate(variables[:max_idx]):
+                s = ""
+                if var.is_discrete or matrix[row, idx]:
+                    if show_labels:
+                        s += "%s=" % var.name
+                    s += var.str_val(matrix[row, idx])
```
nikicc: `s = "%s=" % var.name if show_labels else ''`? Though this might change due to comments above.

pavlin-policar: I think this only makes it less readable and also if we don't need to display labels, we skip a slow string concatenation.

Orange/data/table.py
```diff
@@ -112,26 +112,40 @@ def __setitem__(self, key, value):
...
             return s
 
         table = self.table
         domain = table.domain
         row = self.row_index
         s = "[" + sp_values(table.X, domain.attributes)
         if domain.class_vars:
-            s += " | " + sp_values(table._Y, domain.class_vars)
+            n_targets = table.Y.shape[-1]
+            s += " | " + sp_values(table.Y, domain.class_vars, n_targets > 1)
```
pavlin-policar: Sounds good.

Orange/data/table.py
pavlin-policar: `sp.issparse(value) and len(self) != value.shape[0]` appears to cover it.

Orange/data/table.py
```diff
@@ -114,24 +114,34 @@ def __setitem__(self, key, value):
...
-                "{}={}".format(var.name, var.str_val(val))
-                for var, val in zip(variables, matrix.data[begptr:rendptr]))
-            if limit and rendptr != endptr:
+
+            # Sparse matrices can't handle slices where an index is out of
+            # bounds like numpy arrays can, so we must determine largest index
+            _, columns = matrix.shape
+
+            row_entries, idx = [], 0
+            while len(row_entries) < 5 and idx < len(variables):
```
nikicc: This doesn't consider the `limit` argument. I.e. only 5 values are printed regardless whether the limit is set to either true or false.

screen shot 2017-07-15 at 17 19 19

Orange/tests/test_sparse_table.py
```diff
@@ -41,3 +43,34 @@ def test_row_assignment(self):
...
+        # Convert iris.Y to (150, 1) shape
+        new_y = iris.Y[:, np.newaxis]
+        iris.Y = np.hstack((new_y, new_y))
+        iris.Y = csr_matrix(iris.Y)
+        # We expect the Y shape to match the X shape, which is (150, 4) in iris
+        self.assertEqual(iris.Y.shape, (150, 2))
+
+    def test_Y_setter_2d_single_instance(self):
+        iris = Table('iris')[:1]
+        # Convert iris.Y to (150, 1) shape
```
nikicc: `# Convert iris.Y to (1, 1) shape`

Orange/tests/test_sparse_table.py
```diff
@@ -41,3 +43,34 @@ def test_row_assignment(self):
...
+        new_y = iris.Y[:, np.newaxis]
+        iris.Y = np.hstack((new_y, new_y))
+        iris.Y = csr_matrix(iris.Y)
+        # We expect the Y shape to match the X shape, which is (150, 4) in iris
```
nikicc: `# We expect the Y shape to match the X shape, which is (1, 4)`

Orange/tests/test_table.py
```diff
@@ -1254,7 +1254,7 @@ def test_is_sparse(self):
         self.assertTrue(table.is_sparse())
 
     def test_repr_sparse_with_one_row(self):
-        table = data.Table("iris")[:1]
+        table = data.Table("iris")[::50]
```
nikicc: Why this change? If we change this, then the name `test_repr_sparse_with_one_row` does not fit anymore since we have three instances.

Orange/data/table.py
pavlin-policar: Sorry for taking so long to fixing this, but this should be alright now.

PR #2458, opened 2017-07-08
Statistics.util: Implement digitize

pavlin-policar:
##### Issue
The statistics module did not have an implementation of [`digitize`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.digitize.html).

##### Description of changes
Add implementation of `digitize` which supports dense and sparse data.

@nikicc Would it perhaps make sense to return a sparse matrix if a sparse matrix is passed? Assuming that a large sparse matrix would be passed, and that all the "0"s would belong to the "0" bin, then it would be more efficient to return a sparse matrix instead of a dense numpy array. However there is absolutely no guarantee that the "0"s would indeed belong to the "0" bin, so this may not make sense, as it would just complicate the API. Is this of any concern, or is the current implementation ok?

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -316,3 +316,53 @@ def unique(x, return_counts=False):
         if explicit_zeros:
             return r
         return np.insert(r, 0, 0)
+
+
+def digitize(x, bins, right=False):
+    """Equivalent of np.digitize that supports sparse and dense matrices.
+
+    If a sparse matrix is provided and the '0's belong to the '0'th bin, then
+    a sparse matrix is returned.
+
+    Because this can return both sparse and dense matrices, we must keep the
+    return shape consistent. Since sparse matrices don't support 1d matrices,
+    we reshape any returned 1d numpy array to a 2d matrix, with the first
+    dimension shape being 1. This is equivalent to the behaviour of sparse
+    matrices.
+
+    Parameters
+    ----------
+    x : Union[np.ndarry, sp.csr_matrix, sp.csc_matrix]
+    bins : np.ndarray
+    right : Optional[bool]
+
+    Returns
+    -------
+    Union[np.ndarray, sp.csr_matrix]
+
+    """
+    if not sp.issparse(x):
+        result = np.digitize(x, bins, right)
+        # In order to keep the return shape consistent, and sparse matrices
+        # don't support 1d matrices, make sure to convert 1d to 2d matrices
+        if result.ndim == 1:
+            result = result.reshape(((1,) + result.shape))
+        return result
+
+    # Find the bin where zeros belong, depending on the `right` parameter
+    zero_bin = np.searchsorted(bins, 0, side=['right', 'left'][right])
+
+    if zero_bin == 0:
+        r = sp.lil_matrix(x.shape, dtype=np.int64)
+    else:
+        r = zero_bin * np.ones(x.shape, dtype=np.int64)
+
+    for idx, row in enumerate(x.tocsr()):
+        r[idx, row.indices] = np.digitize(row.data, bins, right)
+
+    # Orange mainly deals with `csr_matrix`, but `lil_matrix` is more efficient
+    # for incremental building
+    if sp.issparse(r):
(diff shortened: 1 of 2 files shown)


Orange/statistics/util.py
```diff
@@ -316,3 +316,21 @@ def unique(x, return_counts=False):
...
+def digitize(x, bins, right=False):
+    """ Equivalent of np.digitize that supports sparse and dense matrices. """
+    if not sp.issparse(x):
+        return np.digitize(x, bins, right)
+
+    # Find the bin where zeros belong, depending on the `right` parameter
+    zero_bin = np.searchsorted(bins, 0, side=['right', 'left'][right])
+
+    r = zero_bin * np.ones(x.shape, dtype=np.int64)
+    for idx, row in enumerate(x):
```
kernc: Perhaps cast `x.tocsr()` before row-wise iteration.

nikicc:
> @nikicc Would it perhaps make sense to return a sparse matrix if a sparse matrix is passed? Assuming that a large sparse matrix would be passed, and that all the "0"s would belong to the "0" bin, then it would be more efficient to return a sparse matrix instead of a dense numpy array.

Yes, I believe so. Currently, our only real use case for sparse data is bag of words, where 0 is minimal value and hence I suppose it should go into the 0-th bin. 

> However there is absolutely no guarantee that the "0"s would indeed belong to the "0" bin, so this may not make sense, as it would just complicate the API. Is this of any concern, or is the current implementation ok?

AFAIK currently passing a BoW data essentially densifies all the data which is undesirable.

Maybe we can do similarly as we did for normalization? There we only support zero offsets for normalization of sparse data sets (class `Normalizer`, file `Orange/preprocess/transformation.py`). Here we could check if the `zero_bin == 0` and return sparse matrix in such cases. If not, we could either raise an error or return a dense matrix. Doing so, we should cover BoW case in the OWFeatureStatistics without the need to densify sparse data.

pavlin-policar:
How's this? If the '0' belongs to the '0'th bin, then we return a sparse matrix, which will cover BoW, and if it doesn't we return a dense matrix. I don't think it makes sense to raise an exception if the 0s don't line up, since then this will only be useful for this particular use case.

pavlin-policar:
However, returning sparse matrices messes up the return shape, since sparse matrices cannot be 1d. To keep the return shapes consistent for all cases, we reshape any returned 1d numpy array to be 2d, so that the shapes are consistent with sparse matrices. This is different from the numpy implementation, which returns a 1d array if a 1d array is provided on input.

Orange/tests/test_statistics.py
```diff
@@ -158,3 +158,62 @@ def test_nanmean(self):
...
+
+    def test_digitize(self):
+        for x in self.data:
+            x_sparse = csr_matrix(x)
+            bins = np.arange(-2, 2)
+
+            np.testing.assert_array_equal(
+                np.digitize(x, bins),
+                digitize(x, bins),
+                'Digitize fails on non-sparse data'
```
nikicc: `Digitize fails on dense data`?

Orange/statistics/util.py
```diff
@@ -316,3 +316,53 @@ def unique(x, return_counts=False):
...
+    -------
+    Union[np.ndarray, np.csrmatrix]
+
+    """
+    if not sp.issparse(x):
+        result = np.digitize(x, bins, right)
+        # In order to keep the return shape consistent, and sparse matrices
+        # don't support 1d matrices, make sure to convert 1d to 2d matrices
+        if result.ndim == 1:
+            result = result.reshape(((1,) + result.shape))
```
nikicc: Not sure if I like this, but then again I don't have a better suggestion.

pavlin-policar: I agree, I don't like this at all, but I we can't return two different matrix shapes depending on which bin the '0's fall into.

nikicc:
Looks good to me. I'm merging since codecov only complains about unrelated problems.

PR #2463, opened 2017-07-11
Statistics.util.var: Implement variance

pavlin-policar:
##### Issue
The statistics module did not have an implementation of [`var`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.var.html) to compute the variance.

##### Description of changes
Add implementation of `var` which supports dense and sparse data.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -375,3 +375,13 @@ def digitize(x, bins, right=False):
         r = r.tocsr()
 
     return r
+
+
+def var(x, axis=None):
+    """ Equivalent of np.var that supports sparse and dense matrices. """
+    if not sp.issparse(x):
+        return np.var(x, axis)
+
+    result = x.multiply(x).mean(axis) - np.square(x.mean(axis))
+    result = np.squeeze(np.asarray(result))
+    return result
--- a/Orange/tests/test_statistics.py
+++ b/Orange/tests/test_statistics.py
@@ -1,11 +1,13 @@
 import unittest
 import warnings
+from itertools import chain
+
 import numpy as np
 import scipy as sp
 from scipy.sparse import csr_matrix, issparse
 
 from Orange.statistics.util import bincount, countnans, contingency, stats, \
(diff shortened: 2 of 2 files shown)


Orange/statistics/util.py
```diff
@@ -316,3 +316,16 @@ def unique(x, return_counts=False):
...
+
+def var(x, axis=0):
+    """ Equivalent of np.var that supports sparse and dense matrices. """
+    if not sp.issparse(x):
+        return np.var(x, axis)
+
+    x_sq = x.copy()
+    x_sq.data = x_sq.data ** 2
+
+    result = np.mean(x_sq, axis) - np.square(np.mean(x, axis))
```
nikicc: What about `np.mean(x.multiply(x), axis) - np.square(np.mean(x, axis))`?

pavlin-policar: Much better, thanks!

Orange/tests/test_statistics.py
```diff
@@ -158,3 +158,12 @@ def test_nanmean(self):
             np.testing.assert_array_equal(
                 nanmean(X_sparse),
                 np.nanmean(X))
+
+    def test_var(self):
+        for data in self.data:
+            for axis in range(len(data.shape)):
```
nikicc: Should we add `axis=None` to also test calculating variance on the whole matrix?

Orange/statistics/util.py
```diff
@@ -316,3 +316,13 @@ def unique(x, return_counts=False):
         if explicit_zeros:
             return r
         return np.insert(r, 0, 0)
+
+
+def var(x, axis=0):
```
nikicc: To be consistent with numpy and the methods above, I suggest we set the default value for axis as `None` instead of `0`.

pavlin-policar: Thanks, for some reason I thought the default value for numpy was 0...

pavlin-policar:
Fixed.

pavlin-policar:
Done

PR #2464, opened 2017-07-11
[ENH] Widget status bar

ales-erjavec:
##### Issue

The OWWidget's 'message_bar' is placed in an inappropriate position above the rest of the GUI causing frequent re-layouts which both flicker the GUI and change the position of controls under the mouse from the unsuspecting user.

##### Description of changes

* Replace the top message area with a summary in a (permanent) status bar at the bottom.
* Also add a progress bar widget, mirroring the `progressBarValue` state.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/messages.py
+++ b/Orange/widgets/utils/messages.py
@@ -32,9 +32,10 @@ from operator import attrgetter
 from warnings import warn
 from inspect import getattr_static
 
-from AnyQt.QtWidgets import QApplication, QStyle, QSizePolicy
+from AnyQt.QtWidgets import QStyle, QSizePolicy
 
 from Orange.widgets import gui
+from Orange.widgets.utils.messagewidget import MessagesWidget
 
 
 class UnboundMsg(str):
@@ -311,7 +312,7 @@ class WidgetMessagesMixin(MessagesMixin):
 
     def __init__(self):
         super().__init__()
-        self.message_bar = self.message_label = self.message_icon = None
+        self.message_bar = None
         self.messageActivated.connect(self.update_message_state)
         self.messageDeactivated.connect(self.update_message_state)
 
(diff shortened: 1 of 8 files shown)


Orange/widgets/utils/messages.py
```diff
@@ -326,73 +327,33 @@ def update_message_state(self):
...
-                format(msg.group.bar_background, font_size,
-                       str(msg).replace("\n", "<br/>&nbsp;&nbsp;&nbsp; "))
-                for msg in messages)
-            self._set_message_bar(group, text, tooltip)
+
+        self.message_bar.clear()
+        if messages:
+            self.message_bar.setMessages(
+                (m, msg(m)) for i, m in enumerate(messages)
+            )
```
janezd: Why `enumerate` and `i`?

Orange/widgets/utils/messagewidget.py
```diff
@@ -0,0 +1,561 @@
...
+    datauri : str
+    """
+    pm = QPixmap(pm)
+    device = QBuffer()
+    assert device.open(QBuffer.ReadWrite)
+    pm.save(device, b'png')
+    device.close()
+    data = bytes(device.data())
+    payload = base64.b64encode(data).decode("ascii")
+    return "data:image/png;base64," + payload
```
janezd: This could be also useful elsewhere. Do we have a suitable place for such functions? (If not, ignore.)

Orange/widgets/utils/messagewidget.py
```diff
@@ -0,0 +1,561 @@
...
+        """
+        Is this message instance empty (has no text or icon)
+        """
+        return (not self.text and self.icon.isNull() and
+                not self.informativeText and not self.detailedText)
+
+
+def standard_pixmap(severity):
+    # type: (Severity) -> QStyle.StandardPixmap
+    maping = {
```
janezd: map**p**ing?

janezd:
I'm eager to merge it, except for one (maybe two) small things.

Before, the "summary" in the message bar showed the text up to the first line break, and the entire message was shown in the tooltip.

screen shot 2017-07-21 at 10 49 06

Now, it shows the entire message.

screen shot 2017-07-21 at 10 51 46

I'd prefer if the text in the bar was limited to a single line.

You may not need to fix this: as I understand, the message can now include a short text, a more detailed text ... This is certainly better than splitting by \n --- except that the `Msg` class doesn't yet accept this attributes. Instead of reimplementing the old functionality (which is used only by a few widgets), you could add the necessary arguments to `UnboundMsg.__new__` (and `_BoundMsg.__call__` to allow string formatting).

The other thing - also visible in these screenshots - is the width of the tooltip, which used to resize to content.

Orange/widgets/utils/messages.py
ales-erjavec: Leftovers from intermediate implementation

Orange/widgets/utils/messagewidget.py
ales-erjavec: I don't know. `Orange.widgets.utils`??

Orange/widgets/utils/messagewidget.py
ales-erjavec: Yes

ales-erjavec:
I restored the previous line splitting behavior, and tooltip wrapping.
`_BoundMsg.__call__` also has a new `exc_info` parameter to include exception traceback in the message.

Issue #2297, opened 2017-05-08
PDF export in Hierarchical Clustering


ajdapretnar:
##### Orange version

3.5.dev

##### Expected behavior
Pdf exports nicely in Hierarchical Clustering.

##### Actual behavior
Pdf is all blue.

##### Steps to reproduce the behavior
Iris - Distances - HierClust. Export visualization in pdf.

##### Additional info (worksheets, data, screenshots, ...)
[iris.pdf](https://github.com/biolab/orange3/files/982729/iris.pdf)

ajdapretnar:
I'm not sure where to put this, but I found another bug in HierClust. Has to do with visualization of... I don't know, reaaaaaaally small values I think?

<img width="963" alt="screen shot 2017-05-09 at 11 46 56" src="https://cloud.githubusercontent.com/assets/12524972/25845247/7cfa5c34-34ad-11e7-81af-8f6512147f6c.png">

PR #2465, opened 2017-07-12
[FIX] owhierarchicalclustering: Explicit geometry transform

ales-erjavec:
##### Issue

Fixes gh-2297

The cause of the broken PDF export was a non aspect preserving implicit dendrogram scaling and use of cosmetic pen to implement fast resize in response to user interaction. QPDFWriter however cannot handle cosmetic pens.

##### Description of changes

Explicit transform/scale the dendrogram geometry to scene display coordinates and avoid use of cosmetic pens.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -4,6 +4,7 @@ import sys
 from collections import namedtuple, OrderedDict
 from itertools import chain
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I tested this and it works, but codeclimate is globally misbehaving. Should I ignore it and just merge or wait for a pass?

janezd:
I haven't seen anything useful from Codeclimate so far, so I'm ignoring it.

PR #2473, opened 2017-07-14
[TST] WidgetTest.send_signal: assert signal value of correct type

kernc:
##### Issue
`send_signal` didn't check the type of sent value. Sending around a Learner class instead of an instance led me to a behavior that took much too long to debug. In general, widget implementors aren't expected to validate types as canvas does it for them. So should the tests.

##### Description of changes
Assert passed signal value is of correct type.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/preprocess.py
+++ b/Orange/preprocess/preprocess.py
@@ -488,7 +488,7 @@ class Scale(Preprocess):
         return data.transform(domain)
 
 
-class PreprocessorList(Reprable):
+class PreprocessorList(Preprocess):
     """
     Store a list of preprocessors and on call apply them to the data set.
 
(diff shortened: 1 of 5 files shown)


astaric:
FYI, you can use
```
python -m unittest Orange.tests Orange.widgets.tests
```
to run tests locally :)

janezd:
+1 for this PR.

I added another assertion: `get_output` used to return `None` if the output with the given name did not exist. I spent half an hour debugging why the Scatter plot didn't output a certain (misspelled) signal.

janezd:
@kernc, can you rebase this?

kernc:
So you'll be able to fix tests, I assume? 😃

janezd:
You read me like an open book. I'll do it -- though most of the 42 come from your assert, not mine. :)

PR #2477, opened 2017-07-17
[FIX] Change Variable Icons: Discrete -> Categorical, Continuous -> Numeric

lanzagar:
##### Issue

Variable icons do not match names used in the File widget, which is usually where people set/check variable types now. Many places in Orange use the same icons and use various names inconsistently.

**Proposal**
We use:
- "categorical" marked with a "C" icon
- "numeric" marked with an "N" icon

##### Description of changes
Moving towards consistency will be a long (Sisyphean) struggle.
This PR makes the file widget consistent, using categorical (C) and numeric (N). It also updates the icons' tooltips. The same icons are used across all Orange.
If/When other labels and descriptions are found using different terms, they should be adapted to use the same names.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/classification/majority.py
+++ b/Orange/classification/majority.py
@@ -22,7 +22,7 @@ class MajorityLearner(Learner):
     def fit_storage(self, dat):
         if not dat.domain.has_discrete_class:
             raise ValueError("classification.MajorityLearner expects a domain "
-                             "with a (single) discrete variable")
+                             "with a (single) categorical variable")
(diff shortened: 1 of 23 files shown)


kernc:
grep `([Dd]iscrete|[Cc]ontinuous) (attribute|variable|feature)` **.py

janezd:
How did I miss those? Thanks...

PR #2480, opened 2017-07-19
[ENH] Radviz: new widget

jerneju:
##### Issue
Radviz widget wanted. Radviz widget is available in Orange 2 but not in Orange 3.

##### Description of changes
New widget in Orange 3. This is much simpler widget than one in Orange 2. Radviz in Orange2 has too many settings.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation


janezd:
### Crashes and errors

- The widget crashes if given a domain without continuous attributes

- If no attributes are shown (e.g. after clicking `None`), the widget crashes when the mouse is moved into the plot.

- If I try to display only a single attribute, I get a warning "Points with missing 'x' or 'y' are not displayed". The user can't know what this means. However, points are shown, but at the coordinate of the sole anchor. Instead, they should be spread between (0, 0) and the anchor. (Since the coordinates are computed as sums over anchors, it's strange that is doesn't work like this already).

- Vizrank searches only for projections of the first three attributes (or, say, first seven if we look for projections with seven attributes).

### Appearance

- The distances between the two arrows for moving the anchor should be independent of the graph size. They shouldn't be squashed together if the graph is small or moved too far from each other if the graph is large or zoomed in.

- Make the circle black or dark grey, possibly a bit thicker.

- Don't show x and y coordinates in tooltips.

- Related to the above point and elsewhere: the user should never been told about some 'x' and 'y' (this is the widget's internal thing) -- except in the output signal. 

### Usability

- 'Show all data on mouse hover' does not appear to have any effect.

- Show class density does not work if the color is set by continuous attribute. This may be OK, but disable the checkbox in this case.

- Moving anchors clears the selection. It shouldn't. Also, keep the selection when changing the set of displayed attributes.

- If the data already includes attributes 'x' or 'y', the output data should rename the columns. For instance, if we plot a radviz of the 'glass' data set and output the projection into a table, we get two attributes named y. Perhaps the attributes should be named `radviz-x` and `radviz-y` (or radviz-x-1, radviz-y-1 and so forth if radviz-x already exists). I believe we already have a method for smart renaming of attributes in such cases.

janezd:
This bug is probably more general and also appears in other widgets: set color to a different attribute than shape. You get this.

screen shot 2017-09-09 at 17 27 30

jerneju:
@janezd: Thank you for all the comments.

I think that color shape error was fixed in #2487 and this PR has not been rebased yet.

lanzagar:
- Class should not be in displayed axes by default (use first 5 features?, and class as color)
- Class density has similar problems as in Linear Projection (not scaled after change, moving anchors refreshes correctly)
- Why limit vizrank to 4+ features? Make the minimum 3. And...
- ... change vizrank to evaluate all combinations UP TO the selected number of vars
- At least 2 features HAS->HAVE to be chosen
- Discrete color variable has to be selected -> Categorical
- The circle is a bit pixelated

For next PR:
- we should support regression data in vizrank
- we should think about improving vizrank by first ranking 3-var combinations and then using that for constructing and evaluationg 4-var comb. etc

doc/visual-programming/source/widgets/visualize/radviz.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Just like all point-based visualizations, this widget includes tools for
+intelligent data visualization (VizRank, see [3]_) and interface for explorative
+data analysis - selection of data points in visualization. Just like in
+:ref:`Scatter Plot` widget, intelligent visualization can be used to find a set
+of attributes that would result in an interesting visualization. For now, this
+works only with class-labeled data set, where interesting visualizations are
+those that well separate data instances of different class. Radviz graph above
+is according to this definition an example of a very good visualization,
+while the one below - where we show an VizRank's interface (:obj:`VizRank` button in
```
lanzagar: the one below is not.
The extra sentence can IMO be omitted, if not add it separately (otherwise hard to read).

doc/visual-programming/source/widgets/visualize/radviz.rst
```diff
@@ -0,0 +1,94 @@
...
+
+The snapshot shown below shows a Radviz widget with a visualization of the
+data set from functional genomics ([2]_). In this particular
+visualization the data instances are colored according to the corresponding
+class, and the visualization space is colored according to the computed class
+probability. Notice that the particular visualization very nicely separates
+the data instances of the different class, making the visualization interesting
+and potentially informative.
+
+.. image:: images/Radviz-Brown.png
```
lanzagar: This image is not added in this PR!

doc/visual-programming/source/widgets/visualize/radviz.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Just like all point-based visualizations, this widget includes tools for
+intelligent data visualization (VizRank, see [3]_) and an interface for explorative
+data analysis - selection of data points in visualization. Just like the
+:ref:`Scatter Plot` widget, it can be used to find a set
+of variables that would result in an interesting visualization. For now, this
+works only with a class-labeled data set, where interesting visualizations are
+those that well separate data instances of different classes. The Radviz graph above
+is according to this definition an example of a very good visualization,
+while the one below is not - where we show an VizRank's interface (:obj:`VizRank` button in
```
lanzagar: Above, below... there is only one image!?

lanzagar:
Merged.
For next PR: vizrank needs a minor fix (here and in linear projection) - when running, changing the number of vars should be disabled (like in mosaic).

PR #2481, opened 2017-07-19
[FIX] Add-ons: Fix Installation of Official Add-ons Through Drag & Drop

nikicc:
##### Issue

Installing official add-ons through drag & drop doesn't work from cleaning the add-ons names onwards (3.4.0 onwards?). The mismatch occurs when we compare cleaned addon name with the zip file name. Since the official addon name was cleaned it doesn't match the uncleaned zip filename.

##### Description of changes
Clean the drag & drop file name in the same manner as official addons.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/addons.py
+++ b/Orange/canvas/application/addons.py
@@ -588,7 +588,7 @@ class AddonManagerDialog(QDialog):
             if path.endswith(self.ADDON_EXTENSIONS):
                 name, vers, summary, descr = (get_meta_from_archive(path) or
                                               (os.path.basename(path), '', '', ''))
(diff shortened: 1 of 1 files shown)


kernc:
If you also change
```py
future.set_result((AddonManagerDialog._packages or []) + packages)
```
to
```py
future.set_result(packages + (AddonManagerDialog._packages or []))
```
a few lines below, so as to sort the dropped package first, _this_ will be the one found and selected when multiple packages with the same name are on the list (i.e. a package listed from PyPi and the a more recent version of it dropped).

PR #2484, opened 2017-07-21
[FIX] Mosaic Vizrank: compute_attr_order is called every step

jerneju:
##### Issue
`compute_attr_order` is called every step. It does not need to be.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -137,12 +137,6 @@ class MosaicVizRank(VizRankDialog, OWComponent):
                 attrs = sorted(zip(weights, data.domain.attributes),
                                key=lambda x: (-x[0], x[1].name))
                 self.attr_ordering = [a for _, a in attrs]
-        if class_var is not None:
-            if self._compute_class_dists():
-                if self.attr_ordering[0] is class_var:
-                    del self.attr_ordering[0]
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owmosaic.py
```diff
@@ -152,7 +152,6 @@ def state_count(self):
         Return the number of combinations, starting with a single attribute
         if Mosaic is colored by class distributions, and two if by Pearson
         """
-        self.compute_attr_order()
```
lanzagar: I agree it was not very good to compute the order every time one needed the count.
But that line was there because of the next one, which sets `n_attrs` using `attr_ordering`. Removing just the first and still relying on `len(self.attr_ordering)` is not the best.
Can you change the next line to actually compute the number of attributes? Should be trivial using len of the domain, just be careful to (not) count the class appropriately.
Then the extra calls in tests will not be needed as well.

jerneju: Fixed. I also removed some dead code. It has not been used since changing color (class) was introduced.

lanzagar:
Seems there are still some issues.
Try iris and move the class to metas. Then run vizrank.

lanzagar:
There are issues with subset data, but unrelated with the changes and should be fixed in another PR.

Issue #2478, opened 2017-07-18
Rank widget takes ages to compute for moderate-sized data sets


BlazZupan:
##### Orange version
3.4.3

##### Expected behavior
Univariate scores such as Information Gain or Gain Ratio have low complexity and should compute fast. Handling larger binary data sets with thousand of features and instances should not be a problem.

##### Actual behavior
Rank widget takes several minutes even for a relatively small data set with, say, 1000 instances and 300 features.

##### Steps to reproduce the behavior

Use File -> Rank on a data set generated by the following code.

```
from Orange.data import Domain, DiscreteVariable, Table
import numpy as np

n, p = 1000, 100  # instances, features
X = np.random.randint(2, size=n*p).reshape((n, p))
Y = np.random.randint(2, size=n)
attributes = [DiscreteVariable("x%d" % i, values=["0", "1"]) for i in range(p)]
class_var = DiscreteVariable("c", values=["0", "1"])
domain = Domain(attributes, class_var)
data = Table(domain, X, Y)
data.save("full-binary.tab")
```

##### Additional info (worksheets, data, screenshots, ...)

I have selected only Information Gain and Gain Ratio in Rank widget. Other scores should not be (pre)computed and should be computed only if selected. I fear that the widget computes a non-univariate measure such as Relief and that this takes time.

PR #2494, opened 2017-07-25
[FIX] Slow Rank

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/2478

##### Description of changes
Updated owrank  to use PyTableModel and its faster sorting, but then I kinda broke something. Then I kinda fixed it by refactoring the whole widget into something I could understand. And it still somewhat works.

Old selection settings won't restore correctly as they were saved as indices of _sorted_ values. Suggestions welcome.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -225,6 +225,7 @@ class OWRank(OWWidget):
             for method in scoring_methods:
                 box.layout().addWidget(QCheckBox(
                     method.name, self,
+                    objectName=method.shortname,  # To be easily found in tests
                     checked=method.name in self.selected_methods,
                     stateChanged=partial(self.methodSelectionChanged, method_name=method.name)))
             gui.rubber(box)
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -11,14 +11,14 @@ from collections import namedtuple
 import numpy as np
 from scipy.sparse import issparse
 
+from AnyQt.QtGui import QFontMetrics
 from AnyQt.QtWidgets import (
     QTableView, QRadioButton, QButtonGroup, QGridLayout, QSizePolicy,
-    QStackedLayout, QStackedWidget, QWidget
+    QStackedLayout, QStackedWidget, QWidget, QHeaderView,
 )
-from AnyQt.QtGui import QStandardItemModel, QStandardItem
 from AnyQt.QtCore import (
     Qt, QItemSelection, QItemSelectionRange, QItemSelectionModel,
-    QSortFilterProxyModel
+    QSize,
 )
 
 from Orange.base import Learner
@@ -29,6 +29,7 @@ from Orange.canvas import report
 from Orange.widgets import gui
 from Orange.widgets.settings import (DomainContextHandler, Setting,
                                      ContextSetting)
+from Orange.widgets.utils.itemmodels import PyTableModel
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.widget import OWWidget, Msg, Input, Output
 
@@ -36,7 +37,7 @@ from Orange.widgets.widget import OWWidget, Msg, Input, Output
 def table(shape, fill=None):
     """ Return a 2D table with shape filed with ``fill``
     """
-    return [[fill for j in range(shape[1])] for i in range(shape[0])]
+    return np.full(shape, fill).tolist()
 
 
 ScoreMeta = namedtuple("score_meta", ["name", "shortname", "score"])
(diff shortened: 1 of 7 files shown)


Orange/widgets/utils/itemmodels.py
```diff
@@ -63,7 +64,135 @@ def _as_contiguous_range(the_slice, length):
...
+
+    def sortColumn(self):
+        """The column currently used for sorting (-1 if no sorting is applied)"""
+        return self.__sortColumn
+
+    def sortOrder(self):
+        """The current sort order"""
+        return self.__sortOrder
+
+    def mapToSource(self, rows: int or numpy.ndarray or Ellipsis):
```
ales-erjavec: This only evaluates to `def mapToSource(self, rows: int):`

`Union[int, Sequence[int], type(...)]` would be better.

However I would drop the Ellipsis unless you assure the returned array is is a copy or a read only view. Otherwise it makes little sense to keep `__sortInd` private.

I am also not too thrilled about the name. `mapToSourceRows` would be better (
I purposefully did not chose this name in TableModel in order to  differentiate from `QAbstractProxyModel.map{To,From}Source`)

Orange/widgets/utils/itemmodels.py
```diff
@@ -819,7 +959,7 @@ def addAction(self, action, *args):
         return self.insertAction(-1, action, *args)
 
 
-class TableModel(QAbstractTableModel):
+class TableModel(AbstractSortTableModel, QAbstractTableModel):
```
ales-erjavec: AbstractSortTableModel is already an QAbstractTableModel.

Orange/widgets/data/owrank.py
```diff
@@ -6,52 +6,146 @@
...
+                         alternatingRowColors=True,
+                         **kwargs)
+        self.setItemDelegate(gui.ColoredBarItemDelegate(self))
+        self.setItemDelegateForColumn(0, QItemDelegate())
+
+        header = self.verticalHeader()
+        header.setSectionResizeMode(header.Fixed)
+        header.setFixedWidth(50)
+        header.setDefaultSectionSize(22)
+        header.setTextElideMode(Qt.ElideMiddle)  # QT-BUG
```
kernc: Thanks. :)

Orange/widgets/data/owrank.py
```diff
@@ -573,71 +458,31 @@ def autoSelection(self):
...
             if len(self.selected_rows):
-                selection = QItemSelection()
-                for row in self.selected_rows:
+                for row in model.mapFromSource(self.selected_rows):
                     selection.append(QItemSelectionRange(
                         model.index(row, 0), model.index(row, columnCount - 1)))
 
         selModel.select(selection, QItemSelectionModel.ClearAndSelect)
 
+        self.setFocus(Qt.OtherFocusReason)  # Prevent graying out scores?
```
kernc: Similar was there before: https://github.com/kernc/orange3/blob/177ce2a524d592849ee32463455dc2de6d7b2e6e/Orange/widgets/data/owrank.py#L527-L528
I don't quite get what it was supposed to fix. @thocevar ?

Orange/widgets/utils/itemmodels.py
```diff
@@ -63,7 +64,135 @@ def _as_contiguous_range(the_slice, length):
...
+            return self.__sortInd[rows]
+        return rows
+
+    def mapFromSource(self, rows: int or numpy.ndarray or Ellipsis):
+        """Return array of row indices in the model for given source table rows"""
+        if self.__sortIndInv is not None:
+            return self.__sortIndInv[rows]
+        return rows
+
+    def invalidate(self):
```
kernc: I'd rather keep it as I get to override it in Rank's model.

Orange/widgets/data/owrank.py
```diff
@@ -158,413 +253,203 @@ def button(text, buttonid, toolTip=None):
...
+        self.ranksModel.clear()
+        self.ranksModel.resetSorting(True)
+
+        self.get_method_scores.cache_clear()
+        self.get_scorer_scores.cache_clear()
+
+        self.Error.clear()
+        self.Information.clear()
+        self.Information.missings_imputed(
+            shown=self.data is not None and self.data.has_missing())
```
ales-erjavec: <s>self.data</s> -> data
self.data is still the old dataset here.

Orange/widgets/data/tests/test_owrank.py
```diff
@@ -238,12 +245,13 @@ def test_data_which_make_scorer_nan(self):
...
             )))
-        self.widget.score_checks[3].setChecked(True) #ANOVA
+        self.widget.selected_methods.add('ANOVA')
         self.send_signal(self.widget.Inputs.data, table)
 
-    def test_setting_migration_extends_header_state(self):
+    def test_setting_migration_fixes_header_state(self):
         # Settings as of version 3.3.5
         settings = {
+            '__version__': 1,
```
astaric: Settings in 3.3.5 did not have versions yet. Therefore, __version__ was not present in the dict.

Orange/widgets/data/owrank.py
```diff
@@ -698,66 +537,38 @@ def create_scores_table(self, labels):
...
-        right_data = right.data(role)
-        if not self.comparable(right_data):
-            right_data = float('-inf')
-        try:
-            return left_data < right_data
-        except TypeError:
-            return str(left_data) < str(right_data)
+        # If older settings, restore sort header to default
+        # Saved selected_rows will likely be incorrect
+        if version is None or version < 2:
```
astaric: Why check for None?
When settings dict does not have the __version__ key, [0 is passed as version](https://github.com/biolab/orange3/blob/master/Orange/widgets/settings.py#L461)

kernc: Backwards compatibility. https://github.com/biolab/orange3/commit/642e446aeb709c4da448d82ed62b68820c5a19ff 😎

Issue #2370, opened 2017-06-02
Tree Doesn't Work With Sparse Data


nikicc:
##### Orange version

3.5.0.dev0+ccbdbba

##### Expected behavior
Tree works with sparse data.

##### Actual behavior
It doesn't due to `np.isnan`

##### Steps to reproduce the behavior
Pass any sparse data to Tree.

##### Additional info (worksheets, data, screenshots, ...)
![screen shot 2017-06-02 at 14 15 43](https://cloud.githubusercontent.com/assets/713026/26725268/fccae160-479d-11e7-9e67-e518adf6a6fe.png)

PR #2496, opened 2017-07-26
[FIX] Trees: Fix classification from sparse data

janezd:
##### Issue

Tree prediction for sparse data does not work.

##### Description of changes

Added a corresponding function in Cython.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/classification/_tree_scorers.c
+++ b/Orange/classification/_tree_scorers.c
@@ -615,7 +615,7 @@ typedef volatile __pyx_atomic_int_type __pyx_atomic_int;
 #endif
 
 
-/* "../../virtual/orange3/lib/python3.5/site-packages/Cython/Includes/numpy/__init__.pxd":725
+/* "../../env/o3/lib/python3.5/site-packages/Cython/Includes/numpy/__init__.pxd":725
  * # in Cython to enable them only on the right systems.
  * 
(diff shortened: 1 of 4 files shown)


ajdapretnar:
Dela. 🌈 🥇 💫

janezd:
Oops, this expects csr. @nikicc, it would be trivial to add a function for csc, too. Or do I just call to_csr?

nikicc:
@janezd I would just call `to_csr`. Do you have any idea what is going on with the Mosaic test on AppVeyor?

janezd:
I added a separate function for csr; it's almost the same, just with different indexing.

I've no idea about Mosaic. Let's see if it fails again.

kernc:
It's tremendous satisfaction working in a team that agrees on what to optimize for.
```py
>>> import scipy.sparse
>>> csr = scipy.sparse.random(1e5, 1e4, format='csr')
>>> %timeit csr.tocsc()
417 ms ± 2.65 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

janezd:
Half a second. If the code that I added is executed ten (OK, twenty) times (on the data of this size), the time saved by adding the code will roughly equal the time I spent by adding it.

kernc:
Thanks, that's _exactly_ what I meant. 😃

nikicc:
Related to #2370.

PR #2497, opened 2017-07-26
[FIX] Regression Tree Sparse Support

jerneju:
##### Issue
Regression Tree Sparse Support

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/regression/tree.py
+++ b/Orange/regression/tree.py
@@ -1,6 +1,7 @@
 """Tree inducers: SKL and Orange's own inducer"""
 
 import numpy as np
+import scipy.sparse as sp
 import sklearn.tree as skl_tree
 
 from Orange.base import TreeModel as TreeModelInterface
@@ -96,7 +97,7 @@ class TreeLearner(Learner):
             if score == 0:
                 return REJECT_ATTRIBUTE
             mapping, branches = MappedDiscreteNode.branches_from_mapping(
-                data.X[:, attr_no], mapping, len(attr.values))
+                col_x, mapping, len(attr.values))
             node = MappedDiscreteNode(attr, attr_no, mapping, None)
             return score, node, branches, 2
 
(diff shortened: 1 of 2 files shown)


nikicc:
Continuation of #2370.

Orange/regression/tree.py
```diff
@@ -95,8 +96,11 @@ def _score_disc_bin():
             # we don't do it here
             if score == 0:
                 return REJECT_ATTRIBUTE
+            datax = data.X[:, attr_no]
+            if is_sparse:
+                datax = datax.toarray()
             mapping, branches = MappedDiscreteNode.branches_from_mapping(
-                data.X[:, attr_no], mapping, len(attr.values))
+                datax, mapping, len(attr.values))
```
nikicc: Couldn't we use `col_x` here?

Orange/widgets/model/tests/test_tree.py
```diff
@@ -41,17 +41,32 @@ def test_parameters_unchecked(self):
...
+        """
+        table1 = Table("housing")
+        self.send_signal("Data", table1)
+        model_dense = self.get_output("Model")
+        table2 = Table("housing")
+        table2.X = sp.csr_matrix(table2.X)
+        self.send_signal("Data", table2)
         model_sparse = self.get_output("Model")
         self.assertTrue(np.array_equal(model_dense._code, model_sparse._code))
-        self.assertTrue(np.array_equal(model_dense._thresholds, model_sparse._thresholds))
```
nikicc: Why is this removed from `test_sparse_data_classification `? Is this not relevant for regression?

jerneju: Because it fails.

jerneju: And I think this is fixed in #2496.

nikicc: But it didn't fail before for classification, if it was among tests, right?

jerneju: Before did not fail since the test did not run properly. Sparse table had never been sent to the widget.

nikicc: I see.

PR #2499, opened 2017-07-26
[FIX] OWBaseLearner: Use new-style signals for base widget

kernc:
Fix merging subclass' signals with OWBaseLearner's defaults.

##### Issue
Fixes #2466

##### Description of changes
Install new-style signals on OWBaseLearner, and merge defaults with addendums so that the unions persist.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #2503, opened 2017-07-26
New style signals for learner widgets

### LINKED ISSUE #2466: Logistic regression and linear regression widgets lost Coefficients outputs
Logistic regression and linear regression widgets lost Coefficients outputs (they are still in the code, but ineffective).

##### Orange version
master

##### Expected behavior
The widgets would have the output as they do in 3.4.4.

### LINKED ISSUE #2492: AdaBoost lost its Learner input
##### Orange version

3.5.dev

##### Expected behavior
AdaBoost accepts Learner as an input to override Tree base estimator.

##### Actual behavior
AdaBoost no longer has a Learner input. Probably lost in the input/output redesign.

File - AdaBoost. Try to connect SVM to AdaBoost.

astaric:
##### Issue
Fixes #2492, fixes #2466, closes #2499

##### Description of changes
Update base learner widget to new-style signals. Update base learner metaclass to "fix" output types to the exact class of the outputted learner/model.
Replace querying the `__dict__` to get the signals with getmembers, the following "sees" the signals on parent classes.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -1,106 +1,55 @@
+from unittest.mock import Mock
```
janezd: Why?

Orange/widgets/utils/signals.py
```diff
@@ -6,6 +6,8 @@
 # increasing counter for ensuring the order of Input/Output definitions
 # is preserved when going through the unordered class namespace of
 # WidgetSignalsMixin.Inputs/Outputs.
+from Orange.widgets.utils import getmembers
```
janezd: Move this line above the comment.

Orange/widgets/utils/owlearnerwidget.py
astaric: leftovers. removed

Orange/widgets/utils/signals.py
astaric: done

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -1,106 +1,53 @@
...
-
-
-class OWBaseLearnerMeta(DefaultWidgetChannelsMetaClass):
-    """Metaclass that adds default inputs (table, preprocess) and
-    outputs (learner, model) for learner widgets.
+    This metaclass ensures that each of the subclasses gets
+    its own Outputs class with output that match the corresponding
+    learner.
     """
+    def __new__(cls, name, bases, attributes):
```
kernc: mcs

kernc:
Can you also patch `OWWidget.send()` to accept old-style `self.send()` syntax. This chunk: https://github.com/biolab/orange3/pull/2499/files#diff-832b90f318ce8a2c9220070cb22a7fa0R181

Orange/widgets/model/owrules.py
```diff
@@ -333,7 +333,7 @@ def update_model(self):
                 self.model.name = self.learner_name
                 self.model.instances = self.data
                 self.valid_data = True
-        self.send(self.OUTPUT_MODEL_NAME, self.model)
+            self.Outputs.model.send(self.model)
```
kernc: This fails the test. I guess it's ok to reset output if input data is foo.

Orange/widgets/utils/owlearnerwidget.py
astaric: As far as I know, first argument to methods should be either self or cls if it is a class method (which __new__ is).

https://www.python.org/dev/peps/pep-0008/#id46

Orange/widgets/model/owrules.py
astaric: Thanks.

astaric:
Regarding [patching the send](https://github.com/biolab/orange3/pull/2503#issuecomment-318192309), what about adding the `@deprecated` decorator to the send method instead?

kernc:
I'm 👍 on _both_. Old-style baselearner subclasses fail on this change without that patch.

astaric:
I am not sure I like this backward compatibility all over the place ...

For send backward compatibility I would add this method to the OWBaseLearner class
```
    def send(self, signalName, value, id=None):
        # A subclass might still use the old syntax to send outputs
        # defined on this class
        for _, output in getmembers(self.Outputs, Output):
            if output.name == signalName:
                output.send(value, id=id)
                return

        super().send(signalName, value, id)
```

For the rest, I am not sure, as I do not know what the problem was that warranted the fix. Maybe we could use a test for this backward compatibility?

astaric:
To be precise, I kind of like the separation that the widget either uses old or new style signals, but not both.

If OWBaseLearner subclasses cause problems, the backward compatibility code should be (IMHO) put on the OWBaseLearner class.

kernc:
I forgot to remove this little snippet. The validity of signals is checked already by the signalManager. I don't think all the necessary changes can be made on OWBaseLearner alone, at least not in a clean manner. (Though I'd have you be the judge of that.)

This now works with e.g. Polynomial Regression / Classification.

@nikicc says this PR is for the release, which is urgent, he says. I just don't want to break old add-on code before someone else gets to fix it.

janezd:
>  widget either uses old or new style signals, but not both.

This was a conscious decision when implementing new signals. I think we shouldn't allow mixing the two styles. If we do, we can be sure to soon have widgets that will mix them for no reason.

astaric:
This fixes the Polynomial Classification/Regression for me and touches only OWBaseLearner (and includes the tests :P)

If there was anything else that was not working, let me know.
(your commits are still available here: https://github.com/astaric/orange3/tree/new-style-signals-kernc)

Orange/widgets/utils/owlearnerwidget.py
kernc: Indeed, appears to be an old pylint thing.

PR #2510, opened 2017-07-29
[ENH] Add parameters and similarity measures to tSNE

pavlin-policar:
##### Issue
The Manifold Learning widget did provide tSNE, yet lacked all the important parameters with which to tune the embedding. Most notably, I find `learning_rate` to be the most important parameter and has a huge effect on the result.

##### Description of changes
Add relevant parameters (perplexity, early exaggeration, learning rate, max iterations and initialization).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmanifoldlearning.py
+++ b/Orange/widgets/unsupervised/owmanifoldlearning.py
@@ -93,11 +93,29 @@ class TSNEParametersEditor(ManifoldParametersEditor):
     # rename l2 to Euclidean
     metric_values = [("l2", "Euclidean")] + metric_values
 
+    perplexity = Setting(30)
+    early_exaggeration = Setting(4)
+    learning_rate = Setting(1000)
+    n_iter = Setting(1000)
+
+    init_index = Setting(0)
+    init_values = [("random", "Random"), ("pca", "PCA")]
+
     def __init__(self, parent):
         super().__init__(parent)
+
         self.metric_combo = self._create_combo_parameter(
             "metric", "Metric:")
-        self.parameters["init"] = "pca"
+        self.perplexity_spin = self._create_spin_parameter(
+            "perplexity", 0, 1000, "Perplexity:")
+        self.early_exaggeration_spin = self._create_spin_parameter(
+            "early_exaggeration", 1, 1000, "Early exaggeration:")
+        self.lr_spin = self._create_spin_parameter(
+            "learning_rate", 1, 1000, "Learning rate:")
+        self.n_iter_spin = self._create_spin_parameter(
+            "n_iter", 0, 1e5, "Max iterations:")
+        self.init_radio = self._create_radio_parameter(
+            "init", "Initialization:")
 
 
 class MDSParametersEditor(ManifoldParametersEditor):
@@ -276,14 +294,14 @@ class OWManifoldLearning(OWWidget):
                 self.Error.out_of_memory()
             except np.linalg.linalg.LinAlgError as e:
                 self.Error.manifold_error(str(e))
+
         self.Outputs.transformed_data.send(out)
 
     def get_method_parameters(self, data, method):
         parameters = dict(n_components=self.n_components)
         parameters.update(self.params_widget.parameters)
         if data is not None and data.is_sparse() and method == TSNE:
-            parameters.update(method='exact',
-                              init='random')
+            parameters.update(method='exact', init='random')
         return parameters
 
     def send_report(self):
--- a/Orange/widgets/unsupervised/owmanifoldlearning.py
+++ b/Orange/widgets/unsupervised/owmanifoldlearning.py
@@ -94,12 +94,12 @@ class TSNEParametersEditor(ManifoldParametersEditor):
     metric_values = [("l2", "Euclidean")] + metric_values
 
     perplexity = Setting(30)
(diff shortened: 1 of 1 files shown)


pavlin-policar:
@lanzagar I've added the missing parameters to tSNE. This can probably be merged easily. 

I decided to leave out the Orange metrics because that would require many more changes e.g. setting precomputed parameters, computing PCA manually on the widget, all of which is doable, but should go into a separate PR.

Orange/widgets/unsupervised/owmanifoldlearning.py
```diff
@@ -93,11 +93,29 @@ class TSNEParametersEditor(ManifoldParametersEditor):
     # rename l2 to Euclidean
     metric_values = [("l2", "Euclidean")] + metric_values
 
+    perplexity = Setting(30)
+    early_exaggeration = Setting(4)
+    learning_rate = Setting(1000)
+    n_iter = Setting(1000)
+
+    init_index = Setting(0)
+    init_values = [("random", "Random"), ("pca", "PCA")]
```
lanzagar: Can we leave the default init to be pca.
I would also switch their positions to have pca as the first (default) option and then random.
This would then also make it consistent with mds (pca = first & default)

Orange/widgets/unsupervised/owmanifoldlearning.py
```diff
@@ -93,11 +93,29 @@ class TSNEParametersEditor(ManifoldParametersEditor):
...
             "metric", "Metric:")
-        self.parameters["init"] = "pca"
+        self.perplexity_spin = self._create_spin_parameter(
+            "perplexity", 0, 1000, "Perplexity:")
+        self.early_exaggeration_spin = self._create_spin_parameter(
+            "early_exaggeration", 1, 1000, "Early exaggeration:")
+        self.lr_spin = self._create_spin_parameter(
+            "learning_rate", 1, 1000, "Learning rate:")
+        self.n_iter_spin = self._create_spin_parameter(
+            "n_iter", 0, 1e5, "Max iterations:")
```
lanzagar: Max iterations spin allows values as low as 0. But running it with <250 results in an error. Change min value of spin control to 250.

Orange/widgets/unsupervised/owmanifoldlearning.py
```diff
@@ -93,11 +93,29 @@ class TSNEParametersEditor(ManifoldParametersEditor):
...
+    init_values = [("random", "Random"), ("pca", "PCA")]
+
     def __init__(self, parent):
         super().__init__(parent)
+
         self.metric_combo = self._create_combo_parameter(
             "metric", "Metric:")
-        self.parameters["init"] = "pca"
+        self.perplexity_spin = self._create_spin_parameter(
+            "perplexity", 0, 1000, "Perplexity:")
```
lanzagar: I am not sure about this one, but sklearn doc says to consider values 5-50. I think spin limits of 1-100 should be fine?

Orange/widgets/unsupervised/owmanifoldlearning.py
```diff
@@ -93,11 +93,29 @@ class TSNEParametersEditor(ManifoldParametersEditor):
...
+        self.early_exaggeration_spin = self._create_spin_parameter(
+            "early_exaggeration", 1, 1000, "Early exaggeration:")
```
lanzagar: Not sure about this one either, but 1000 sounds really high. Probably 1-100 should be fine?

Orange/widgets/unsupervised/owmanifoldlearning.py
```diff
@@ -93,11 +93,29 @@ class TSNEParametersEditor(ManifoldParametersEditor):
     # rename l2 to Euclidean
     metric_values = [("l2", "Euclidean")] + metric_values
 
+    perplexity = Setting(30)
+    early_exaggeration = Setting(4)
+    learning_rate = Setting(1000)
+    n_iter = Setting(1000)
```
lanzagar: I suggest we have the same defaults as the latest sklearn version (they have been changing them recently)
perplexity=30, e.e.=12, lr=200, iter=1000

Orange/widgets/unsupervised/owmanifoldlearning.py
lanzagar: I know this goes against the sklearn defaults, but we already had pca as the only option before and I think it makes more sense for users (random can show a completely different figure each time). The consistency with mds is an added bonus.

Orange/widgets/unsupervised/owmanifoldlearning.py
pavlin-policar: I didn't realize they had changed them. Fixed now.

Orange/widgets/unsupervised/owmanifoldlearning.py
pavlin-policar: As far as I can tell, both the MDS widget and manifold MDS control pane have the option to initialize randomly. I have, however, made PCA the default value, since that is what MDS does in both cases. Is there a third MDS widget that I missed?

lanzagar: This is fine now. That is what I meant - just change the order of pca and random, so that pca is default (but we still have both options, like in mds)

PR #2513, opened 2017-08-01
Updated Data info icon design.

larazupan:


markotoplak:
I find the old icon clearer.

In the chosen font the dot on the "i" is very far. The this font the letter "i" seems more line 1 with a dot somewhere over it to me.

janezd:
I agree with Marko.

PR #2514, opened 2017-08-03
[ENH] Widget status bar buttons

ales-erjavec:
##### Issue

Move standard Report/Save Image ... buttons to the status bar

##### Description of changes

* Provide a standard buttons for common actions in the OWWidget statusbar.
* <s>Deprecate `buttons_area_orientation` and the `buttonsArea`</s>. The old QPushButtons are still created but are not inserted into any layout or parent hierarchy. Any subclass that inserted them into other layouts (there are plenty examples of this) should still work.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Perhaps I should go through all widgets and remove the manually rearranged push buttons? And, perhaps, also through all other widgets and see whether their layout after this PR is still OK? I'd then push the corresponding commits into this PR, so we can finally merge it?

@ales-erjavec, would this be OK?

ales-erjavec:
I would rather not do that in this pull request. Keeping and using the back-compatibility guarantee at least for a little while is the best test that it works.

janezd:
I didn't mean to remove the back-compatibility. It's going to stay there since add-ons may shuffle the buttons, too. I just wanted to remove the extra report buttons appearing in the widgets, and fix any broken layouts in the widgets from which the button was removed.

I went through all core widgets now, and the layout is mostly fine, so ... merging.

PR #2517, opened 2017-08-04
Scripts: prepare signed macos binary

astaric:
This is a script I use to prepare a signed macos release. I had to modify the mac build process a bit in order to be able to sign the resulting app.

The script could (should?) be rewritten in a way that does not hardcode version and paths, but that would require some bash programming skills I do not possess.

### DIFF
--- a/scripts/macos/sign-app.sh
+++ b/scripts/macos/sign-app.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Build app
+rm -rf ~/dev/orange3/dist/Orange3.app
+./build-macos-app.sh ~/dev/orange3/dist/Orange3.app
+# Missing symlink Current messes with code signing
+ln -s 3.6 ../../dist/Orange3.app/Contents/Frameworks/Python.framework/Versions/Current
+# sign bundle
+codesign -s "Developer ID" /Users/anze/dev/orange3/dist/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.6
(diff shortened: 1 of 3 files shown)


scripts/macos/sign-app.sh
```diff
@@ -0,0 +1,16 @@
...
+ln -s 3.6 ../../dist/Orange3.app/Contents/Frameworks/Python.framework/Versions/Current
+# sign bundle
+codesign -s "Developer ID" /Users/anze/dev/orange3/dist/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.6
+codesign -s "Developer ID" /Users/anze/dev/orange3/dist/Orange3.app/Contents/MacOS/pip 
+codesign -s "Developer ID" /Users/anze/dev/orange3/dist/Orange3.app
+
+# Create disk image
+./create-dmg-installer.sh --app ../../dist/Orange3.app ../../dist/Orange3-3.4.5-signed.dmg
+# Sign disk image
+codesign -s "Developer ID" /Users/anze/dev/orange3/dist/Orange3-3.4.5-signed.dmg
```
kernc: Something like this might do the trick?
```bash
orange_version="$(../../dist/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.6/bin/python -c 'from Orange.version import short_version; print(short_version)')"

./create-dmg-installer [...] ../../dist/Orange3-$orange_version-signed.dmg
[...]
```

Issue #2423, opened 2017-06-22
File widget: change attribute type from numeric to string


brunap:
##### Orange version
3.4.4.

##### Expected behavior
If attribute type change from numeric to string, the number representation shall be without decimal point in case all values are integers.

##### Actual behavior
If attribute type change from numeric to string, the number representation include decimal point although all attribute values are integers.

##### Steps to reproduce the behaviour
open attached data sheet (test.xlsx) with file widget
change attribute type for the id attribute from numeric to string
see the id values in the data table widget

##### Additional info (worksheets, data, screenshots, ...)
data file:

[test.xlsx](https://github.com/biolab/orange3/files/1095108/test.xlsx)

before attribute type change

![one](https://user-images.githubusercontent.com/11367976/27439424-33fa6410-5768-11e7-8e62-0ae1e91bb95f.png)

after attribute type change

![numeric-string](https://user-images.githubusercontent.com/11367976/27439459-48fdbb28-5768-11e7-8570-1467003737eb.png)

PR #2518, opened 2017-08-04
FileWidget: DomainEditor update of type conversions.

### LINKED ISSUE #2424: File widget: change attribute type from nominal to numeric
##### Orange version
3.4.4.

##### Expected behavior
If attribute type change from nominal to numeric, the number values shall remain the same. E.g. if there were two nominal values 1 and 2, the same two values should remain after the attribute type change.

##### Actual behavior
If attribute type change from nominal to numeric, the number representation change. E.g. if there were two nominal values 1 and 2, after the attribute type change we have 0 and 1. This happens only if values are 1 and 2, if the values are 22 and 33 (or other) this does not happen.

open attached data sheet (test.xlsx) with file widget
change attribute type for '2nd-code' from nominal to numeric
see '2nd-code' values in the data table widget

data file

[test.xlsx](https://github.com/biolab/orange3/files/1095164/test.xlsx)

before attribute type change

one

after attribute type change

nominal-numeric

thocevar:
##### Issue
Fixes #2424 and #2423.

##### Description of changes
Conversion nominal -> numeric now retains the actual values (bugfix).
Conversion numeric -> * drops decimal places (e.g. 18.0 -> 18) if all entries are integer.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owfile.py
+++ b/Orange/widgets/data/tests/test_owfile.py
@@ -247,6 +247,30 @@ a
         formats = dialog_formats()
         self.assertTrue(".123" in formats)
 
-
-class TestDomainEditor(TestCase):
-    pass
+    def test_domain_editor_conversions(self):
+        dat = """V0\tV1\tV2\tV3\tV4\tV5
+                 c\tc\td\td\tc\td
+                  \t \t \t \t \t
+                 3.0\t1.0\t4\ta\t0.0\tx
+                 1.0\t2.0\t4\tb\t0.0\ty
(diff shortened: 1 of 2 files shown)


brunap:
This does not solve issue #2424 for the test.xlsx dataset I attached to the issue. After changing datatype for the attribute '2nd-code' from categorical to numeric, I have the same error (the number representation change).

thocevar:
It works with .tab and .csv files as in the screenshot but not with .xlsx for some reason. I'll look into it.

PR #2520, opened 2017-08-07
[FIX] Fix asynchronous widget tests

ales-erjavec:
##### Issue

Asynchronous tests for owmds and owtestlearners set inputs to their respective widgets when they should not (a per `OWWidget.isBlocking`).
 
##### Description of changes

* Raise RuntimeError in `WidgetTest.send_signal` if the target widget is not ready to receive inputs.
* Add `wait` parameter to `WidgetTest.get_output`.
* Fix test_owmds and test_owtestlearners.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -3,6 +3,8 @@ import os
 import time
 import unittest
 from unittest.mock import Mock
+# pylint: disable=unused-import
+from typing import List, Optional, Type, TypeVar
 
 import numpy as np
 import sip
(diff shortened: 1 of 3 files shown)


Orange/widgets/tests/base.py
```diff
@@ -215,31 +221,42 @@ def send_signal(self, input, value, *args, widget=None, wait=-1):
             else:
                 raise ValueError("'{}' is not an input name for widget {}"
                                  .format(input, type(widget).__name__))
+        if widget.isBlocking():
```
markotoplak: Of the widget is blocking the code will not get past this check and therefore the "wait" parameter is not used. Was this intended?

ales-erjavec: Yes. The (existing) wait is done after the inputs are set and `handleNewSignals` is called. It is somewhat misplaced , but I do not want to change it.

PR #2522, opened 2017-08-09
[FIX] Another color palette when too many colors needed

jerneju:
##### Issue
Similar as in #2394. This time fixed in widgets **Calibration Plot** and **Lift Curve**.

Additional info: https://sentry.io/biolab/orange3/issues/210623933/

##### Description of changes
Same solution as in #2394.

Three tests use the same code and that is why that code is moved to a separate file.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owliftcurve.py
+++ b/Orange/widgets/evaluate/owliftcurve.py
@@ -155,8 +155,10 @@ class OWLiftCurve(widget.OWWidget):
         if names is None:
             names = ["#{}".format(i + 1) for i in range(N)]
 
-        self.colors = colorpalette.ColorPaletteGenerator(
-            N, colorbrewer.colorSchemes["qualitative"]["Dark2"])
+        scheme = colorbrewer.colorSchemes["qualitative"]["Dark2"]
+        if N > len(scheme):
+            scheme = colorpalette.DefaultRGBColors
+        self.colors = colorpalette.ColorPaletteGenerator(N, scheme)
 
         self.classifier_names = names
         self.selected_classifiers = list(range(N))
--- a/Orange/widgets/evaluate/owliftcurve.py
+++ b/Orange/widgets/evaluate/owliftcurve.py
@@ -5,7 +5,7 @@ Lift Curve Widget
 """
 from collections import namedtuple
 
-import numpy
+import numpy as np
 import sklearn.metrics as skl_metrics
 
 from AnyQt import QtWidgets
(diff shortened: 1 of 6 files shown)


Orange/widgets/evaluate/owliftcurve.py
```diff
@@ -155,8 +155,11 @@ def _initialize(self, results):
         if names is None:
             names = ["#{}".format(i + 1) for i in range(N)]
 
+        scheme = colorbrewer.colorSchemes["qualitative"]["Dark2"]
+        if N > len(scheme):
+            scheme = colorpalette.DefaultRGBColors
         self.colors = colorpalette.ColorPaletteGenerator(
-            N, colorbrewer.colorSchemes["qualitative"]["Dark2"])
+            N, scheme)
```
janezd: Why line break here?

Orange/widgets/evaluate/tests/common.py
```diff
@@ -0,0 +1,23 @@
...
+        classification.TreeLearner(),
+        classification.SVMLearner(),
+        classification.KNNLearner(),
+        classification.CN2Learner(),
+        classification.SGDClassificationLearner(),
+        classification.RandomForestLearner(),
+        classification.NaiveBayesLearner(),
+        classification.SGDClassificationLearner()
+    ]
+    return evaluation.CrossValidation(data, learners, k=2, store_data=True)
```
janezd: Rename `common.py` to `base.py`. Instead of function `color_palette_test`, `base.py` should include a class derived from `WidgetTest` with a method `test_many_evaluation_results`. `TestOWROCAnalysis`, `TestOWCalibrationPlot` and `TestOWLiftCurve` are then derived from this class.

Orange/widgets/evaluate/tests/test_owcalibrationplot.py
```diff
@@ -49,3 +50,11 @@ def test_nan_input(self):
         self.assertTrue(self.widget.Error.invalid_results.is_shown())
         self.send_signal(self.widget.Inputs.evaluation_results, None)
         self.assertFalse(self.widget.Error.invalid_results.is_shown())
+
+    def test_many_evaluation_results(self):
```
janezd: Remove this method (inherit it from the base class).

Orange/widgets/evaluate/tests/test_owliftcurve.py
```diff
@@ -51,3 +52,11 @@ def test_nan_input(self):
         self.assertTrue(self.widget.Error.invalid_results.is_shown())
         self.send_signal(self.widget.Inputs.evaluation_results, None)
         self.assertFalse(self.widget.Error.invalid_results.is_shown())
+
+    def test_many_evaluation_results(self):
```
janezd: Remove this method (inherit it from the base class).

Orange/widgets/evaluate/tests/common.py
jerneju: Done.

PR #2526, opened 2017-08-11
Add Show Settings to widget context menu

astaric:
##### Issue
It is hard to see widget's settings

##### Description of changes
Add an action to the widget's context menu that is shown when canvas is run in debug mode (debug log level), which prints the current widget's settings to the log.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/canvas/document/schemeedit.py
+++ b/Orange/canvas/document/schemeedit.py
@@ -858,8 +858,8 @@ class SchemeEditWidget(QWidget):
         """
         Print widget settings to log.
         """
-        import pprint
-        pp = pprint.PrettyPrinter(indent=4)
+        from Orange.widgets.settings import SettingsPrinter
+        pp = SettingsPrinter(indent=4)
         selected = self.scene().selectedItems()
         for item in selected:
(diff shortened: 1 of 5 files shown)


astaric:
@ales-erjavec is there a better way to add an item to the context menu?

I dislike the import of widget.settings in the canvas code (this is why I made it a local import), but since the __widgetMenu is "really" private, I do not want to touch it from the outside.

ales-erjavec:
> @ales-erjavec is there a better way to add an item to the context menu?

In orange-canvas-core and orange-widgets-core there is a 'side channel' by which widgets can add their own actions into the canvas context menu (something like https://github.com/ales-erjavec/orange3/commit/45519a75ab6c145f7d1382c4d900a0be1dfeb4a1) But I am not certain that is what you have in mind.

> I dislike the import of widget.settings in the canvas code (this is why I made it a local import)

Maybe dispatch the request to workflow model  instance (e.g. `Scheme.dump_settings(node: SchemeNode)`) and have `WidgetsScheme` implement it.

PR #2528, opened 2017-08-11
[FIX] Mosaic Display: subset data

### LINKED ISSUE #2515: Mosaic: crash on data and subset data
##### Orange version
From at least 3.3.8 onward

##### Expected behavior
Should not crash. Subset data properly interpreted.

##### Actual behavior
screenshot_20170803_122307

https://sentry.io/biolab/orange3/issues/252105951/

jerneju:
##### Issue
Fixes #2515 .

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -303,8 +303,8 @@ class OWMosaicDisplay(OWWidget):
 
         self.data = None
         self.discrete_data = None
-        self.unprocessed_subset_data = None
         self.subset_data = None
+        self.subset_indices = None
 
         self.color_data = None
 
@@ -379,6 +379,7 @@ class OWMosaicDisplay(OWWidget):
         if data is None:
             self.color_model.set_domain(None)
             return
+        self.color_model.set_domain(self.data.domain)
         for combo in self.attr_combos[1:]:
             combo.addItem("(None)")
 
@@ -438,33 +439,28 @@ class OWMosaicDisplay(OWWidget):
             self.init_combos(None)
             return
 
-        self.color_model.set_domain(self.data.domain)
         self.init_combos(self.data)
 
         self.openContext(self.data)
 
-        # if we first received subset we now call setSubsetData to process it
-        if self.unprocessed_subset_data:
-            self.set_subset_data(self.unprocessed_subset_data)
-            self.unprocessed_subset_data = None
-
-        self.set_color_data()
-
     @Inputs.data_subset
     def set_subset_data(self, data):
-        self.Warning.incompatible_subset.clear()
-        if self.data is None:
-            self.unprocessed_subset_data = data
-            return
-        try:
-            self.subset_data = data.transform(self.data.domain)
-        except:
-            self.subset_data = None
-            self.Warning.incompatible_subset(shown=data is not None)
+        self.subset_data = data
 
     # this is called by widget after setData and setSubsetData are called.
     # this way the graph is updated only once
     def handleNewSignals(self):
+        self.Warning.incompatible_subset.clear()
+        self.subset_indices = None
+        if self.data is not None and self.subset_data:
+            try:
+                indices = {e.id for e in self.subset_data.transform(self.data.domain)}
+            except:
+                self.Warning.incompatible_subset(shown=self.subset_indices is not None)
+            finally:
+                self.subset_indices = [ex.id in indices for ex in self.data]
+
+        self.set_color_data()
         self.reset_graph()
 
     def clear_selection(self):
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/252105951/

janezd:
Feed a dataset and a subset from File to Mosaic. Then change the data set in File. The widget crashes. 

The problem is that `set_data` calls `set_color_data` which in turn calls `update_graph`. `update_graph` should be called from `handleNewSignals`, after both signals are processed. Combo `cb_attr_color` should call another method that calls `set_color_data` and then `update_graph`.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -837,7 +841,7 @@ def line(x1, y1, x2, y2):
                         if sum(counts) == 1:
```
janezd: Marking the whole rectangle is ugly, unlikely and non-obvious (despite being here for at least three years). I suggest removing this `if` and its contents altogether.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -461,6 +462,9 @@ def set_subset_data(self, data):
         except:
             self.subset_data = None
             self.Warning.incompatible_subset(shown=data is not None)
+        finally:
+            indices = list(set(e.id for e in self.subset_data))
```
janezd: `indices = {e.id for e in self.subset_data}`

That is, use set comprehension and, more importantly, `indices` should be a set.

janezd: I would set `self.subset_indices` to `None` if `self.data` is `None`. Alternatively, this twisted code (not by your fault!) could be improved. I think that `set_subset_data` should be only:

    @Inputs.data_subset
    def set_subset_data(self, data):
        self.subset_data = data

The rest of this function belongs to `handleNewSignals` (or a separate method called by `handleNewSignals`). This is precisely what `handleNewSignals` is meant for. (It's possible that this widget was originally written before we had `handleNewSignals`.)

jerneju: Fixed.

Orange/widgets/visualize/owmosaic.py
jerneju: Removed.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -438,34 +439,29 @@ def set_data(self, data):
...
     # this way the graph is updated only once
     def handleNewSignals(self):
-        self.reset_graph()
+        self.Warning.incompatible_subset.clear()
+        self.subset_indices = None
+        if self.data is not None and self.subset_data:
+            try:
+                indices = {e.id for e in self.subset_data.transform(self.data.domain)}
+            except:
+                self.Warning.incompatible_subset(shown=self.subset_indices is not None)
```
janezd: Have I overlooked something, or is `self.subset_indices` always `None` here? Originally, this was `data is not None`, but here even this is not needed since data is checked to be non-None.

Second, where can this `try` clause fail at all?

Perhaps add a test that feeds iris as data and titanic as subset, and check that the warning is displayed. (Currently it's not - I tried.)

A general problem: how do we defined an "incompatible subset"? I guess a subset is incompatible if none of the attributes can be transformed, that is, if `transformed = self.subset_data.transform(self.data.domain)` and `np.all(np.isnan(transformed.X)) and np.all(np.isnan(transformed.Y))`.

janezd: Maybe `Table.transform` should have an additional argument, a flag that would tell it to raise an exception if no attribute can be transformed. `Table.transform` can probably easily detect this (I haven't checked), it's faster than `np.all(np.isnan(..))` and, most importantly, does not report an incompatible domain when the domain is compatible (maybe even the same), but all data is missing.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -438,34 +440,29 @@ def set_data(self, data):
...
+        self.subset_indices = indices = None
+        if self.data is not None and self.subset_data:
+            transformed = self.subset_data.transform(self.data.domain)
+            if np.all(np.isnan(transformed.X)) and np.all(np.isnan(transformed.Y)):
+                self.Warning.incompatible_subset()
+            else:
+                indices = {e.id for e in transformed}
+                self.subset_indices = [ex.id in indices for ex in self.data]
+
+        self.clear_selection()
```
janezd: This is new. Is it necessary -- and desired?

Selection is already cleared upon receiving new data. This would also clear it upon receiving a subset, which is unnecessary and undesired, I guess. Have I overlooked something?

Orange/widgets/visualize/owmosaic.py
```diff
@@ -482,6 +479,7 @@ def reset_graph(self):
 
     def set_color_data(self):
         if self.data is None or len(self.data) < 2 or len(self.data.domain.attributes) < 1:
+            self.coloring_changed()
```
janezd: Is this necessary? It would stop vizrank, which can't be running in this case anyway, and it updates graph, which is also unnecessary.

jerneju: Removed.

janezd:
Everything looks ok, I'd merge ... except that appveyor stopped. @astaric?

jerneju:
@janezd : Appveyor succeeded.

PR #2541, opened 2017-08-22
Move common things from Scatter Plot to SP Graph and Plot GUI

### LINKED ISSUE #2540: Scatter Plot: cannot reduce the window height
##### Orange version
master

##### Expected behavior
One should be able to make widget smaller.

##### Actual behavior
screenshot_20170822_131217

jerneju:
##### Issue
Duplicated code.

##### Description of changes
1) Scatter Plot Graph can accept custom **ViewBox**. Some widgets need to use custom ones because they use special mouse events.

2) Some common things were moved to Scatter Plot Graph and Plot GUI. These common things can also be used in other widgets (Freeviz, Linear Projection, MDS, Radviz).

3) Static method `create_groups_table` moved to `annotated_data.py` as a function. Can be used by many other widgets as well.

Does not add scrollbars anymore. It is highly not recommended to use scrollbars in Orange.

screenshot_20170822_131736

##### Includes
- [x] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/plot/owplotgui.py
+++ b/Orange/widgets/utils/plot/owplotgui.py
@@ -257,6 +257,10 @@ class OWPlotGUI:
         :type icon_name: str
 
     '''
+
+    JITTER_SIZES = [0, 0.1, 0.5, 1, 2, 3, 4, 5, 7, 10]
+
+
     def __init__(self, plot):
         self._plot = plot
         self.color_model = DomainModel(placeholder="(Same color)",
@@ -288,6 +292,11 @@ class OWPlotGUI:
     ZoomSelection = 15
     ZoomReset = 16
 
+    ToolTipShowsAll = 17
+    ClassDensity = 18
+    RegressionLine = 19
+    LabelOnlySelected = 20
+
     SelectionAdd = 21
     SelectionRemove = 22
     SelectionToggle = 23
@@ -308,6 +317,9 @@ class OWPlotGUI:
     AntialiasLines = 45
     DisableAnimationsThreshold = 48
     AutoAdjustPerformance = 49
+
+    JitterSizeSlider = 51
+    JitterNumericValues = 52
 
     UserButton = 100
 
@@ -365,7 +377,7 @@ class OWPlotGUI:
             When the checkbox is toggled, the attribute ``value`` of the plot object is set to the checkbox' check state,
             and the callback ``cb_name`` is called.
         '''
-        gui.checkBox(widget, self._plot, value, label, callback=self._get_callback(cb_name))
+        return gui.checkBox(widget, self._plot, value, label, callback=self._get_callback(cb_name))
 
     def antialiasing_check_box(self, widget):
         '''
@@ -373,11 +385,43 @@ class OWPlotGUI:
         '''
         self._check_box(widget, 'use_antialiasing', 'Use antialiasing', 'update_antialiasing')
 
+    def jitter_size_slider(self, widget):
+        values = getattr(self._plot.master, "jitter_sizes", self.JITTER_SIZES)
+        gui.valueSlider(
+            widget=widget, master=self._plot, value='jitter_size', label="Jittering: ",
+            values=values, callback=self._plot.master.reset_graph_data,
+            labelFormat=lambda x: "None" if x == 0 else ("%.1f %%" if x < 1 else "%d %%") % x)
+
+    def jitter_numeric_check_box(self, widget):
+        gui.checkBox(
+            widget=gui.indentedBox(widget=widget), master=self._plot, value="jitter_continuous",
+            label="Jitter numeric values", callback=self._plot.master.reset_graph_data)
+
     def show_legend_check_box(self, widget):
         '''
             Creates a check box that shows and hides the plot legend
         '''
         self._check_box(widget, 'show_legend', 'Show legend', 'update_legend')
+
+    def tooltip_shows_all_check_box(self, widget):
+        self._check_box(widget=widget, value="tooltip_shows_all",
+                        label='Show all data on mouse hover', cb_name="cb_tooltip_shows_all")
+
+    def class_density_check_box(self, widget):
+        self._plot.master.cb_class_density = \
+            self._check_box(widget=widget, value="class_density", label="Show class density",
+                            cb_name=self._plot.master.update_density)
+
+    def regression_line_check_box(self, widget):
+        self._plot.master.cb_reg_line = \
+            self._check_box(widget=widget, value="show_reg_line",
+                            label="Show regression line",
+                            cb_name=self._plot.master.update_regression_line)
+
+    def label_only_selected_check_box(self, widget):
+        self._check_box(widget=widget, value="label_only_selected",
+                        label="Label only selected points",
+                        cb_name=self._plot.master.graph.update_labels)
 
     def filled_symbols_check_box(self, widget):
         self._check_box(widget, 'show_filled_symbols', 'Show filled symbols', 'update_filled_symbols')
@@ -449,6 +493,19 @@ class OWPlotGUI:
             self.AlphaValue
             ], widget, box, "Points")
 
+    def plot_properties_box(self, widget, box=None):
+        """
+        Create a box with controls for common plot settings
+        """
+        return self.create_box([
+            self.ToolTipShowsAll,
+            self.ShowGridLines,
+            self.ShowLegend,
+            self.ClassDensity,
+            self.RegressionLine,
+            self.LabelOnlySelected
+            ], widget, box, "Plot Properties")
+
     def plot_settings_box(self, widget, box=None):
         '''
             Creates a box with controls for common plot settings
@@ -460,9 +517,15 @@ class OWPlotGUI:
             ], widget, box, "Plot settings")
 
     _functions = {
-        ShowLegend: show_legend_check_box,
         ShowFilledSymbols: filled_symbols_check_box,
+        JitterSizeSlider: jitter_size_slider,
+        JitterNumericValues: jitter_numeric_check_box,
+        ShowLegend: show_legend_check_box,
         ShowGridLines: grid_lines_check_box,
+        ToolTipShowsAll: tooltip_shows_all_check_box,
+        ClassDensity: class_density_check_box,
+        RegressionLine: regression_line_check_box,
+        LabelOnlySelected: label_only_selected_check_box,
         PointSize: point_size_slider,
         AlphaValue: alpha_value_slider,
         Color: color_value_combo,
--- a/Orange/widgets/utils/plot/owplotgui.py
+++ b/Orange/widgets/utils/plot/owplotgui.py
@@ -31,11 +31,12 @@ from Orange.data import ContinuousVariable, DiscreteVariable
 from Orange.widgets import gui
 from Orange.widgets.utils.itemmodels import DomainModel
 
-from .owconstants import *
-
-from AnyQt.QtWidgets import QWidget, QToolButton, QGroupBox, QVBoxLayout, QHBoxLayout, QMenu, QAction
+from AnyQt.QtWidgets import QWidget, QToolButton, QVBoxLayout, QHBoxLayout, QMenu, QAction
 from AnyQt.QtGui import QIcon
-from AnyQt.QtCore import Qt, pyqtSignal, QObject
+from AnyQt.QtCore import Qt, pyqtSignal
+
+from .owconstants import NOTHING, ZOOMING, SELECT, SELECT_POLYGON, PANNING, SELECTION_ADD,\
+    SELECTION_REMOVE, SELECTION_TOGGLE, SELECTION_REPLACE
 
 
 class OrientedWidget(QWidget):
@@ -69,7 +70,7 @@ class OWToolbar(OrientedWidget):
         :param parent: The toolbar's parent widget
         :type parent: :obj:`.QWidget`
     '''
-    def __init__(self, gui, text, orientation, buttons, parent, nomargin = False):
+    def __init__(self, gui, text, orientation, buttons, parent, nomargin=False):
         OrientedWidget.__init__(self, orientation, parent)
         self.buttons = {}
         self.groups = {}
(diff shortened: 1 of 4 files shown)


Orange/widgets/visualize/utils/plotutils.py
```diff
@@ -0,0 +1,151 @@
...
+from AnyQt.QtCore import Qt
+from AnyQt.QtGui import QKeySequence
+
+from Orange.widgets import gui
+
+
+JITTER_SIZES = [0, 0.1, 0.5, 1, 2, 3, 4, 5, 7, 10]
+
+
+def create_scroll_area(master):
```
janezd: This does not belong to plotutils. Feels like `gui.py`, though I'm not sure it's general enough. But if it is, it's not related to plots only.

Orange/widgets/visualize/utils/plotutils.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+        zoom_fit = QAction(
+            "Fit in view", self.master,
+            shortcut=QKeySequence(Qt.ControlModifier | Qt.Key_0),
+            triggered=fit_to_view
+        )
+        self.addActions([zoom_in, zoom_out, zoom_fit])
+
+
+def set_size_policy(groupbox):
```
janezd: Shouldn't this be a method of `Box`?

Orange/widgets/visualize/utils/plotutils.py
```diff
@@ -0,0 +1,151 @@
...
+            labelFormat=lambda x:
+            "None" if x == 0 else ("%.1f %%" if x < 1 else "%d %%") % x)
+        if is_cb_jitter_cont:
+            gui.checkBox(
+                widget=gui.indentedBox(widget=self.box_plot),
+                master=self.master, value="graph.jitter_continuous",
+                label="Jitter numeric values", callback=self.master.reset_graph_data)
+
+
+class Box:
```
janezd: I'd call this `_Box`.

Orange/widgets/visualize/utils/plotutils.py
```diff
@@ -0,0 +1,151 @@
...
+    def __init__(self, master, groupbox=None):
+        self.master = master
+        self.groupbox = groupbox
+
+
+class BoxAxisData(Box):
```
janezd: ... on the other hand, I'm not sure these are proper classes. Following the current (perhaps suboptimal) practice from `gui.py`, this would be functions. They don't have any attributes except for `master` and `groupbox` that are essentially used only for passing the data to the `__call__`.

Their use is also weird: the class is constructed just to be called and then thrown away, e.g. `plotutils.BoxAxisData(self, groupbox)()`. I'd prefer `plotutils.box_axis_data(self, groupbox)`

Orange/widgets/visualize/utils/plotutils.py
```diff
@@ -0,0 +1,151 @@
...
+            "None" if x == 0 else ("%.1f %%" if x < 1 else "%d %%") % x)
+        if is_cb_jitter_cont:
+            gui.checkBox(
+                widget=gui.indentedBox(widget=self.box_plot),
+                master=self.master, value="graph.jitter_continuous",
+                label="Jitter numeric values", callback=self.master.reset_graph_data)
+
+
+class Box:
+    def __init__(self, master, groupbox=None):
```
janezd: `groupbox` has a default (`None`), but must never be `None`. Hence, it shouldn't have the default. The name is also misleading: it's not the box of the group (groupbox) but the box into which the group is put. `parent` may be a better name.

Orange/widgets/visualize/utils/plotutils.py
```diff
@@ -0,0 +1,151 @@
...
+
+class BoxAxisData(Box):
+    def __call__(self):
+        common_options = dict(
+            labelWidth=50, orientation=Qt.Horizontal, sendSelectedValue=True,
+            valueType=str)
+        box_axis_data = gui.vBox(self.groupbox, "Axis Data")
+        self.master.cb_attr_x = gui.comboBox(
+            box_axis_data, self.master, "attr_x", label="Axis x:", callback=self.master.update_attr,
+            model=self.master.xy_model, **common_options)
```
janezd: These classes suppose a number of master's method names, e.g. `update_attr` and attributes, e.g. `xy_model`. This does not look right -- but perhaps I'd do the same. Not sure.

janezd:
Avoid the horizontal scrollbar. It should never be needed.

I'm not even sure we should have the vertical scrollbar. If the widget does not fit on a reasonable screen, it should be either simplified or, worse case, use tabs. For a start, removing box titles may already reduce the height. The new status bar will eventually remove the bottom two buttons.

jerneju:
Sure, I can remove both scrollbars. However, I would keep common stuff in a separated file since they can be used by various widgets.

ajdapretnar:
I agree with @janezd here. I didn't want to instantly put down this PR, but I don't think widgets should be shrinkable beyond control area. It bothers me, too, especially with Heat Map and Box Plot. But I also agree that widgets should be simplified and/or use tabs.

However, what I was thinking about was whether we can make the windows smaller **only** when control area is disabled. Now cont.ar. is prevent the widget to be shrunk further, which somewhat goes against the intuition behind hiding it in the first place.

Orange/widgets/visualize/utils/plotutils.py
jerneju: Can this `plotutils.py` be renamed to `plotgui.py`?

jerneju:
On the other hand some widgets really take a lot of space:

screenshot_20170823_132248

ajdapretnar:
I think there's no need for the first two boxes to be so large. In Box Plot you can squish them very well. We should do the same for other widgets, e.g. LinProj.

Orange/widgets/visualize/utils/plotutils.py
janezd: Sounds reasonable.

janezd:
I believe we also have collapsable boxes. `gui.collapsableWidgetBox` may or may not be useful, and may or may not work. I'd use this as the last resort, though. At the moment, no widget uses this.

Orange/widgets/visualize/utils/plotutils.py
jerneju: I have replaced classes with functions.

Orange/widgets/visualize/utils/plotgui.py
```diff
@@ -0,0 +1,113 @@
...
+        box_zoom_select, nomargin=True,
+        buttons=[g.StateButtonsBegin, g.SimpleSelect, g.Pan, g.Zoom,
+                 g.StateButtonsEnd, g.ZoomReset]
+    )
+    buttons = zoom_select_toolbar.buttons
+    buttons[g.Zoom].clicked.connect(graph.zoom_button_clicked)
+    buttons[g.Pan].clicked.connect(graph.pan_button_clicked)
+    buttons[g.SimpleSelect].clicked.connect(graph.select_button_clicked)
+    buttons[g.ZoomReset].clicked.connect(graph.reset_button_clicked)
+    return box_zoom_select
```
janezd: Move to `OWScatterPlotGraph`.

Orange/widgets/visualize/utils/plotgui.py
```diff
@@ -0,0 +1,113 @@
...
+    gui.valueSlider(
+        widget=parent, master=master, value='graph.jitter_size',
+        label="Jittering: ", values=jitter_sizes, callback=master.reset_graph_data,
+        labelFormat=lambda x:
+        "None" if x == 0 else ("%.1f %%" if x < 1 else "%d %%") % x)
+    if is_cb_jitter_cont:
+        gui.checkBox(
+            widget=gui.indentedBox(widget=parent),
+            master=master, value="graph.jitter_continuous",
+            label="Jitter numeric values", callback=master.reset_graph_data)
```
janezd: Move to `OWScatterPlotGraph`.

Orange/widgets/visualize/utils/plotgui.py
```diff
@@ -0,0 +1,113 @@
...
+
+def box_radius(master, parent):
+    box_radius = gui.widgetBox(parent, "Hide")
+    master.rslider = gui.hSlider(
+        widget=box_radius, master=master, value="radius", minValue=0, maxValue=100,
+        step=5, label="Radius", createLabel=False, ticks=True,
+        callback=master.update_radius)
+    master.rslider.setTickInterval(0)
+    master.rslider.setPageStep(10)
+    return box_radius
```
janezd: This one is specific for free viz and doesn't belong here.

Orange/widgets/visualize/utils/plotgui.py
```diff
@@ -0,0 +1,113 @@
...
+        widget=box_plot_prop, master=master, value="graph.class_density",
+        label='Show class density',
+        callback=master.update_density)
+    master.cb_reg_line = gui.checkBox(
+        widget=box_plot_prop, master=master, value="graph.show_reg_line",
+        label="Show regression line", callback=master.update_regression_line)
+    gui.checkBox(
+        widget=box_plot_prop, master=master, value="graph.label_only_selected",
+        label="Label only selected points", callback=master.graph.update_labels)
+    return box_plot_prop
```
janezd: Move to `OWScatterPlotGraph`. Regression line, which is specific to scatter plot, can be added by the widget.

Orange/widgets/visualize/utils/plotgui.py
```diff
@@ -0,0 +1,113 @@
...
+
+def box_points(master, parent):
+    g = master.graph.gui
+    box_pp = g.point_properties_box(parent)
+    master.models = [master.xy_model] + g.points_models
+    return box_pp
```
janezd: Should this be a function? How general is it?

Orange/widgets/visualize/utils/plotgui.py
```diff
@@ -0,0 +1,113 @@
...
+        labelWidth=50, orientation=Qt.Horizontal, sendSelectedValue=True,
+        valueType=str)
+    box_axis_data = gui.vBox(parent, "Axis Data")
+    master.cb_attr_x = gui.comboBox(
+        box_axis_data, master, "attr_x", label="Axis x:", callback=master.update_attr,
+        model=master.xy_model, **common_options)
+    master.cb_attr_y = gui.comboBox(
+        box_axis_data, master, "attr_y", label="Axis y:", callback=master.update_attr,
+        model=master.xy_model, **common_options)
+    return box_axis_data
```
janezd: Isn't this specific to scatter plot? It can stay in the scatter plot's `__init__`.

Orange/widgets/visualize/utils/plotgui.py
```diff
@@ -0,0 +1,113 @@
...
+    g = master.graph.gui
+    g.add_widgets([g.ShowLegend, g.ShowGridLines], box_plot_prop)
+    gui.checkBox(
+        widget=box_plot_prop, master=master, value="graph.tooltip_shows_all",
+        label="Show all data on mouse hover")
+    master.cb_class_density = gui.checkBox(
+        widget=box_plot_prop, master=master, value="graph.class_density",
+        label='Show class density',
+        callback=master.update_density)
+    master.cb_reg_line = gui.checkBox(
```
janezd: Put these attributes on the box, e.g. `box_plot_prop.cb_reg_lin = gui.checkBox(...`.

Even better, don't store it at all: *explore whether* checkbox can be accessed through `self.controls.graph.class_density` and through `self.graph.controls.class_density`.

Orange/widgets/visualize/utils/plotgui.py
```diff
@@ -0,0 +1,113 @@
...
+    zoom_out = QAction(
+        "Zoom out", master, shortcut=QKeySequence.ZoomOut,
+        triggered=lambda: zoom(1 / 1.25)
+    )
+    zoom_fit = QAction(
+        "Fit in view", master,
+        shortcut=QKeySequence(Qt.ControlModifier | Qt.Key_0),
+        triggered=fit_to_view
+    )
+    master.addActions([zoom_in, zoom_out, zoom_fit])
```
janezd: This can be probably moved to `OWScatterPlotGraph`.

Orange/widgets/visualize/utils/plotgui.py
jerneju: Done.

Orange/widgets/visualize/utils/plotgui.py
jerneju: Yes, probably.

Orange/widgets/visualize/utils/plotgui.py
jerneju: Well, method `draw_regression_line` is already in `OWScatterPlotGraph`.

jerneju:
@janezd: Please review this.

Orange/widgets/utils/plot/owplotgui.py
```diff
@@ -365,20 +377,53 @@ def _check_box(self, widget, value, label, cb_name):
...
+        return gui.checkBox(widget, self._plot, value, label, callback=self._get_callback(cb_name))
 
     def antialiasing_check_box(self, widget):
         '''
             Creates a check box that toggles the Antialiasing of the plot
         '''
         self._check_box(widget, 'use_antialiasing', 'Use antialiasing', 'update_antialiasing')
 
+    def jitter_size_slider(self, widget):
+        values = self._plot.master.jitter_sizes if hasattr(self._plot.master, "jitter_sizes") \
```
janezd: This could be `values = getattr(self._plot.master, "jitter_sizes", self.JITTER_SIZES)`. 

Is it not `"JITTER_SIZES"`, not `"jitter_sizes"`? I suppose https://github.com/biolab/orange3/pull/2541/commits/49cf76d6006d060f170315e5dff06b41aa9dee97#diff-d6ec84862acc9707da06d662d6752cbdR489 should define this sizes, right?

However, I'd prefer passing jitter sizes as an argument to this function instead of forcing the widget to specify it in a completely different location with a prescribed name. E.g.
`def jitter_size_slider(self, widget, jitter_sizes=JITTER_SIZES):`. (For this to work without warnings, `JITTER_SIZES` would have to be a tuple, not a list.)

Orange/widgets/utils/plot/owplotgui.py
```diff
@@ -31,12 +31,12 @@
 from Orange.widgets import gui
 from Orange.widgets.utils.itemmodels import DomainModel
 
-from .owconstants import *
```
janezd: Have you checked that no other code relies on having those constants in `owplotgui`'s namespace? Just for the case...

Orange/widgets/utils/plot/owplotgui.py
```diff
@@ -572,7 +576,8 @@ def tool_button(self, id, widget):
         '''
         id, name, attr_name, attr_value, callback, icon_name = self._expand_id(id)
         if id == OWPlotGUI.Select:
-            b = self.menu_button(self.Select, [self.SelectionOne, self.SelectionAdd, self.SelectionRemove, self.SelectionToggle], widget)
+            b = self.menu_button(self.Select, [self.SelectionOne, self.SelectionAdd,
+                                               self.SelectionRemove, self.SelectionToggle], widget)
```
janezd: This call has three arguments. They used to be in a single line that was too long. Now they are in two lines; one and a half in the first line and one and a half in the second. None of this is optimal, but I preferred the original. Or

            b = self.menu_button(
                self.Select,
                [self.SelectionOne, self.SelectionAdd,
                 self.SelectionRemove, self.SelectionToggle],
                widget)

Don't break the lines that are not broken. :)

jerneju: Fixed.

Orange/widgets/utils/plot/owplotgui.py
jerneju: Everything needed by `ScatterPlotGraph` is imported. **Parallel Coordinates** may need some other constants but anyway that widget is a dead code. However I will revert this code back just in case.

janezd: If nobody who currently imports owplotgui uses these names, it's better to remove the import. I just asked whether you checked. If confirmed, please re-remove it.

janezd:
I suppose that the line 49cf76d#diff-d6ec84862acc9707da06d662d6752cbdR489 is now redundant. I removed it and will merge if tests pass.

Orange/widgets/utils/plot/owplotgui.py
jerneju: OK, I reverted it back and also did some style changes. That is why I now imported even more constants, now they are used instead of numbers (see https://github.com/biolab/orange3/pull/2541/files#diff-665cab46a232703eb5e7bb8b696303ecR105 ).

jerneju:
@janezd : Tests passed.

janezd:
You'll need to rebase some other PRs now, and perhaps you can also clean up some that do not necessarily require rebasing. Please ping me on the PRs that are ready for review.

PR #2543, opened 2017-08-22
[WIP] DataHashContextHandler

jerneju:
##### Issue
See 
- https://github.com/biolab/orange3/pull/2301#discussion_r118804152
- https://github.com/biolab/orange3/pull/2301#issuecomment-304228325

##### Description of changes
Work in progress.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -862,6 +862,29 @@ class ContextHandler(SettingsHandler):
         return value
 
 
+class DataHashContextHandler(ContextHandler):
+    NO_MATCH = 0
+    MATCH = 1
+    def new_context(self, data):
+        """Create a new context."""
+        context = super().new_context()
+        if data and data.domain:
+            context.data_hash = data.checksum()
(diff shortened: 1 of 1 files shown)


janezd:
@astaric, what did we decide about data hash context handlers? Do I recall correctly that you argued against them?

astaric:
I have argued that they do not make sense for storing selections, as selections are not really context settings. (you usually do no want to restore the selection when a new widget instance sees the data it has seen before) For those cases, schema-only settings make more sense.

I someone knows of another case where DataHashContextHandler would be useful, I can argue against that as well :)

janezd:
Do we merge things we don't (yet?) need? Or do we close the PR and hope we remember it in case we eventually need it?

astaric:
I vote for closing.

kernc:
> Do we merge things we don't (yet?) need?

Never a question. [YAGNI](http://wiki.c2.com/?YouArentGonnaNeedIt).

PR #2549, opened 2017-08-25
DomainContextSettings: Encode variables in lists

nikicc:
##### Issue

When lists of variables were stored in ContextSettings, reloading them did not work if Variables in the input data domain weren't made with `Variable.make`.

##### Description of changes
Lists in ContextSettings are also encoded/decoded as settings.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -869,6 +869,9 @@ class DomainContextHandler(ContextHandler):
 
     MATCH_VALUES_NONE, MATCH_VALUES_CLASS, MATCH_VALUES_ALL = range(3)
 
+    ENCODED_VARIABLE = 100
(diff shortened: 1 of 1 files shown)


jerneju:
Please rebase.

jerneju:
@nikicc : What is the status of this PR?

nikicc:
I think there was some problem with this approach, but I'm not really sure. IMO the issue still persist, perhaps we should solve it differently, or just open an issue and close this PR.

astaric:
The problem with this PR is that supporting both old style (variable stored in settings by names) and new style (variables stored as variables) is a mess. For the moment, widgets that store variables in lists should implement their own context handler that handles this.

PR #2558, opened 2017-09-04
Statistics.countnans: Fix sparse implementation and add axis support

pavlin-policar:
##### Issue
The statistics module implementation of `countnans` did not support the `axis` argument. Moreover, it appeared to have been computing number of NaNs incorrectly on sparse data in general.

##### Description of changes
Add implementation of `countnans` which correctly counts NaNs and supports the `axis` keyword.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -44,6 +44,22 @@ def sparse_has_zeros(x):
     return np.prod(x.shape) != x.nnz
 
 
+def sparse_zero_weights(x, weights):
+    """ Extract the weight values of all zeros in a sparse matrix. """
+    if not sp.issparse(x):
+        raise TypeError('The matrix provided was not sparse.')
+
+    if weights.ndim == 1:
+        n_items = np.prod(x.shape)
+        zero_indices = np.setdiff1d(np.arange(n_items), x.indices, assume_unique=True)
+        return weights[zero_indices]
+    else:
+        # Can easily be implemented using a coo_matrix
+        raise NotImplemented(
+            'Computing zero weights on ndimensinal weight matrix is not implemented'
+        )
+
+
 def bincount(x, weights=None, max_val=None, minlength=None):
     """Return counts of values in array X.
 
@@ -83,11 +99,8 @@ def bincount(x, weights=None, max_val=None, minlength=None):
     # Store the original matrix before any manipulation to check for sparse
     x_original = x
     if sp.issparse(x):
-        n_items = np.prod(x_original.shape)
-        zero_indices = np.setdiff1d(np.arange(n_items), x.indices, assume_unique=True)
-
         if weights is not None:
-            zero_weights = weights[zero_indices].sum()
+            zero_weights = sparse_zero_weights(x, weights).sum()
             weights = weights[x.indices]
         else:
             zero_weights = sparse_count_zeros(x)
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -31,10 +31,17 @@ def _count_nans_per_row_sparse(X, weights, dtype=None):
 
 
 def sparse_count_zeros(x):
-    """ Count the number of zeros in a sparse matrix. """
+    """ Count the number of implicit zeros in a sparse matrix. """
     if not sp.issparse(x):
         raise TypeError('The matrix provided was not sparse.')
     return np.prod(x.shape) - x.nnz
+
+
+def sparse_has_zeros(x):
+    """ Check if sparse matrix contains any implicit zeros. """
+    if not sp.issparse(x):
+        raise TypeError('The matrix provided was not sparse.')
+    return np.prod(x.shape) != x.nnz
 
 
 def bincount(x, weights=None, max_val=None, minlength=None):
@@ -315,17 +322,12 @@ def stats(X, weights=None, compute_variance=False):
             X.shape[0] - nans))
 
 
-def _sparse_has_zeros(x):
-    """ Check if sparse matrix contains any implicit zeros. """
-    return np.prod(x.shape) != x.nnz
-
-
 def _nan_min_max(x, func, axis=0):
     if not sp.issparse(x):
         return func(x, axis=axis)
     if axis is None:
         extreme = func(x.data, axis=axis) if x.nnz else float('nan')
-        if _sparse_has_zeros(x):
+        if sparse_has_zeros(x):
             extreme = func([0, extreme])
         return extreme
     if axis == 0:
@@ -338,7 +340,7 @@ def _nan_min_max(x, func, axis=0):
     for row in x:
         values = row.data
         extreme = func(values) if values.size else float('nan')
-        if _sparse_has_zeros(row):
+        if sparse_has_zeros(row):
             extreme = func([0, extreme])
         r.append(extreme)
     return np.array(r)
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -25,14 +25,18 @@ def bincount(X, max_val=None, weights=None, minlength=None):
 
     Works kind of like np.bincount(), except that it also supports floating
     arrays with nans.
+
+    Returns
+    -------
+    Tuple[np.ndarray, int]
+        Returns the bincounts and the number of NaN values.
+
     """
+    # Store the original matrix before any manipulation to check for sparse
+    X_ = X
     if sp.issparse(X):
-        minlength = max_val + 1
-        bin_weights = weights[X.indices] if weights is not None else None
-        return (np.bincount(X.data.astype(int),
-                            weights=bin_weights,
-                            minlength=minlength, ),
-                _count_nans_per_row_sparse(X, weights))
+        weights = weights[X.indices] if weights is not None else weights
+        X = X.data
 
     X = np.asanyarray(X)
     if X.dtype.kind == 'f' and bn.anynan(X):
@@ -45,11 +49,21 @@ def bincount(X, max_val=None, weights=None, minlength=None):
             nans = (~nonnan).sum(axis=0)
     else:
         nans = 0. if X.ndim == 1 else np.zeros(X.shape[1], dtype=float)
+
     if minlength is None and max_val is not None:
         minlength = max_val + 1
-    bc = np.array([]) if minlength is not None and minlength <= 0 else \
-        np.bincount(X.astype(np.int32, copy=False),
-                    weights=weights, minlength=minlength).astype(float)
+
+    if minlength is not None and minlength <= 0:
+        bc = np.array([])
+    else:
+        bc = np.bincount(
+            X.astype(np.int32, copy=False), weights=weights, minlength=minlength
+        ).astype(float)
+        # Since `csr_matrix.values` only contain non-zero values, we must count
+        # those separately and set the appropriate bin
+        if sp.issparse(X_):
+            bc[0] = np.prod(X_.shape) - X_.nnz
+
     return bc, nans
 
 
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -10,7 +10,7 @@ import scipy.sparse as sp
 import bottleneck as bn
 
 
-def _count_nans_per_row_sparse(X, weights):
+def _count_nans_per_row_sparse(X, weights, dtype=None):
     """ Count the number of nans (undefined) values per row. """
     if weights is not None:
         X = X.tocoo(copy=False)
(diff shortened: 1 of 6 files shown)


pavlin-policar:
@nikicc It appears that the existing code was not counting the number of NaNs, but rather the number of zero elements.

```
>>> x = np.array([1, 1, 0, 0, 2, np.nan, 0, 0]).reshape(2, -1)
array([[  1.,   1.,   0.,   0.],
       [  2.,  nan,   0.,   0.]])
>>> _count_nans_per_row_sparse(csr_matrix(x), None)
4.0
```

Apparently, this functionality is used in [`Table._compute_distributions`](https://github.com/biolab/orange3/blob/master/Orange/data/table.py#L1416) (hence my failing test). I haven't looked any further than that, but it appears that either the distributions in the table must be being computed incorrectly, or the function name is wrong.

Orange/statistics/util.py
```diff
@@ -12,28 +12,32 @@
...
+
+    return counts
 
 
 def bincount(X, max_val=None, weights=None, minlength=None):
     """Return counts of values in array X.
 
     Works kind of like np.bincount(), except that it also supports floating
     arrays with nans.
+
```
nikicc: Would you be willing to also document the `max_val` argument? This doesn't seem to be a numpy argument, and it's not obvious what it is for.

Orange/statistics/util.py
```diff
@@ -46,44 +50,72 @@ def bincount(X, max_val=None, weights=None, minlength=None):
...
+    if minlength is not None and minlength <= 0:
+        bc = np.array([])
+    else:
+        bc = np.bincount(
+            X.astype(np.int32, copy=False), weights=weights, minlength=minlength
+        ).astype(float)
+        # Since `csr_matrix.values` only contain non-zero values, we must count
+        # those separately and set the appropriate bin
+        if sp.issparse(X_):
+            bc[0] = np.prod(X_.shape) - X_.nnz
```
nikicc: Should we maybe make `np.prod(X_.shape) - X_.nnz` as a separate function, something like `sparse_count_zeros`? We already have `_sparse_has_zeros` and we will need to add the number of zero elements to the Table widget sooner or later.

Orange/tests/test_statistics.py
```diff
@@ -241,3 +228,94 @@ def test_var(self):
...
+class TestBincount(unittest.TestCase):
+    def test_bincount_counts_nans(self):
+        dense = [0, 0, 0, 1, np.nan, 1]
+        sparse = csr_matrix(dense)
+        expected = [3, 2], 1
+
+        np.testing.assert_equal(bincount(dense), expected)
+        np.testing.assert_equal(bincount(sparse), expected)
+
+        hist, n_nans = bincount([0., 1., 3], max_val=3)
```
nikicc: What exactly does `max_val=3` does here?

Orange/tests/test_statistics.py
```diff
@@ -241,3 +228,94 @@ def test_var(self):
...
+                      [2, 2]])
+        np.testing.assert_equal(countnans(x, weights=w, axis=0), [0, 3])
+        np.testing.assert_equal(countnans(x, weights=w, axis=1), [1, 2])
+
+        w = np.array([1, 2])
+        np.testing.assert_equal(countnans(x, weights=w, axis=0), [0, 4])
+        np.testing.assert_equal(countnans(x, weights=w, axis=1), [1, 2])
+
+
+class TestBincount(unittest.TestCase):
```
nikicc: Perhaps add one test with weights?

Orange/statistics/util.py
```diff
@@ -46,44 +50,72 @@ def bincount(X, max_val=None, weights=None, minlength=None):
...
+
     """
     if not sp.issparse(X):
         X = np.asanyarray(X)
         isnan = np.isnan(X)
         if weights is not None and weights.shape == X.shape:
             isnan = isnan * weights
+
+        # In order to keep return types consistent with sparse vectors, we will
+        # handle `axis=1` given a regular 1d numpy array equivallently as
```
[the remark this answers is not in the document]
pavlin-policar: I agree, this does seem to be the most sensible solution. Most of all, we should be consistent with the behaviour on sparse and dense matrices, otherwise much confusion is inevitable.

Orange/tests/test_statistics.py
```diff
@@ -241,3 +228,94 @@ def test_var(self):
...
+        x = [[1, np.nan],
+             [2, np.nan]]
+        np.testing.assert_equal(
+            countnans(x, axis=1), [1, 1],
+            'Countnans fails on dense data with `axis=1`')
+        np.testing.assert_equal(
+            countnans(csr_matrix(x), axis=1), [1, 1],
+            'Countnans fails on sparse data with `axis=1`')
+
+    def test_countnans_with_2d_weights(self):
```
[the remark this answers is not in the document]
pavlin-policar: This didn't work before and I wasn't sure whether or not it was even worth the hassle, since I assume we only use sparse matrices when the data is very big, and forming a dense 2d weight matrix of that size kind of defeats the purpose...

However, I've implemented it now. In any case sparse weight matrices weren't supported before and aren't supported now. I don't know if it's a requirement, but it would require more work.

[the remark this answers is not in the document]
nikicc: You're right, we shouldn't worry about weights on sparse matrices.

Orange/statistics/util.py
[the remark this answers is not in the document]
nikicc: > Most of all, we should be consistent with the behaviour on sparse and dense matrices, otherwise much confusion is inevitable.

But on the other hand, changing user-provided input of `axis=1` to `axis=0` just because `axis=1` would result in an error is IMO worse than some slight incompatibility between sparse and dense. If I could come up with at least one use-case where this incompatibility would really hurt us I might be less against it.

IMO if you want to stick to 100% same behavior, I would prefer if we also raise an error when a sparse row is provided and user passes axis=1. Hence we could essentially treat a sparse matrix of shape (1, X) as 1D array and get exactly the same behaviour as with 1D dense matrices.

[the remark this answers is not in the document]
pavlin-policar: Yes, I agree, changing the user input was a bad idea. The error is much better.

Orange/statistics/util.py
pavlin-policar: I've added thorough documentation.

Orange/statistics/util.py
pavlin-policar: I think this is a good idea.

Perhaps it would also make sense to implement `sparse_num_elements` which would essentially be `np.prod(x.shape)` since it's also used a lot. This would make the functionality much clearer, but I'm hesitant because it seems far too excessive.

Orange/tests/test_statistics.py
pavlin-policar: This was a remnant of the old tests. I've removed it and added an appropriate test.

Orange/tests/test_statistics.py
pavlin-policar: Done.

pavlin-policar:
At first I was puzzled why zeros were counted as NaNs in sparse matrices. This has no apparent advantages at all and can be confusing when first dealing with it...

This probably came from pandas, where the default fill for sparse matrices is NaN (but can be changed to anything). This doesn't make much sense to me, because if we separate the two, we can easily determine which data is missing and which is actually zero. Pandas does differentiate between the explicit and implicit NaNs, but then again, pandas is far more flexible than plain sparse matrices.

nikicc:
> This probably came from pandas, where the default fill for sparse matrices is NaN (but can be changed to anything).

Not sure where this came from, but it was incorrect. The current assumption is that values not stored in the sparse matrix correspond to zeros and not missing values! Missing values are explicitly stored in sparse matrices as `np.nan`. If there still exist functions that don't obey this rule, they should be corrected.

Orange/tests/test_distribution.py
```diff
@@ -338,93 +338,82 @@ def test_get_distributions(self):
...
-        X = sp.csr_matrix((sdata, indices, indptr), shape=(5, 20))
+        X = sp.csr_matrix(
+            # 0  1  2  3       4       5       6  7  8  9 10 11 12   13 14 15 16      17 18 19
+            # --------------------------------------------------------------------------------
+            [[0, 2, 0, 2,      1,      1,      2, 0, 0, 1, 0, 0, 0,   1, 1, 0, 2, np.nan, 2, 0],
+             [0, 0, 1, 1, np.nan, np.nan,      1, 0, 2, 0, 0, 0, 0,   0, 2, 0, 1, np.nan, 0, 0],
+             [0, 0, 0, 1,      0,      2, np.nan, 0, 0, 0, 0, 0, 0,   0, 0, 0, 0,      0, 0, 0],
+             [0, 0, 0, 0,      0,      0,      0, 0, 0, 0, 0, 0, 0,   0, 0, 0, 0,      0, 0, 0],
+             [0, 0, 2, 0,      0,      0,      1, 0, 0, 0, 0, 0, 0, 1.1, 0, 0, 0,      0, 0, 0]]
+        )
```
nikicc: This test doesn't quite work for explicit zeros. E.g. if you add `X[0,0] = 0` here, the test should pass and the results should be the same — but it fails. IMO the problem is in the `bincount`, check the comment above.

Orange/statistics/util.py
```diff
@@ -30,35 +30,67 @@ def _count_nans_per_row_sparse(X, weights):
...
-        nonnan = ~np.isnan(X)
-        X = X[nonnan]
+    x_original = x
+    if sp.issparse(x):
+        n_items = np.prod(x_original.shape)
+        zero_indices = np.setdiff1d(np.arange(n_items), x.indices, assume_unique=True)
+
+        if weights is not None:
+            zero_weights = weights[zero_indices].sum()
+            weights = weights[x.indices]
```
nikicc: Does `x.indices` works here for both `csc` and `csr`? If not, please cast it to whatever (csc/csr) is required beforhand.

Orange/statistics/util.py
```diff
@@ -44,6 +44,22 @@ def sparse_has_zeros(x):
...
 
 
+def sparse_zero_weights(x, weights):
+    """ Extract the weight values of all zeros in a sparse matrix. """
+    if not sp.issparse(x):
+        raise TypeError('The matrix provided was not sparse.')
+
+    if weights.ndim == 1:
+        n_items = np.prod(x.shape)
+        zero_indices = np.setdiff1d(np.arange(n_items), x.indices, assume_unique=True)
```
nikicc: Does `x.indices` works here for both `csc` and `csr`? If not, please cast it to whatever (csc/csr) is required beforhand.

Orange/statistics/util.py
```diff
@@ -10,80 +10,181 @@
...
+    if minlength is not None and minlength <= 0:
+        bc = np.array([])
+    else:
+        bc = np.bincount(
+            x.astype(np.int32, copy=False), weights=weights, minlength=minlength
+        ).astype(float)
+        # Since `csr_matrix.values` only contain non-zero values, we must count
+        # those separately and set the appropriate bin
+        if sp.issparse(x_original):
+            bc[0] = zero_weights
```
nikicc: This should probably be `bc[0] = bc[0] + zero_weights` to account for explicit zeros stored in `x.data`.

Orange/statistics/util.py
```diff
@@ -10,80 +10,181 @@
...
+
+        w = sp.coo_matrix((data_weights, (nan_rows, nan_cols)), shape=X.shape)
+        w = w.tocsr()
+
+        return np.fromiter((np.sum(row.data) for row in w), dtype=dtype)
+
+    return np.fromiter((np.isnan(row.data).sum() for row in X), dtype=dtype)
+
+
+def sparse_count_zeros(x):
```
nikicc: Should we perhaps rename this to `sparse_count_implicit_zeros` to make sure explicit zeros aren't counted?

Orange/statistics/util.py
```diff
@@ -10,80 +10,181 @@
...
+    """ Count the number of implicit zeros in a sparse matrix. """
+    if not sp.issparse(x):
+        raise TypeError('The matrix provided was not sparse.')
+    return np.prod(x.shape) - x.nnz
+
+
+def sparse_has_zeros(x):
```
nikicc: Should we perhaps rename this to `sparse_has_implicit_zeros ` to make sure explicit zeros aren't considered?

Orange/tests/test_distribution.py
nikicc: Also, once this is corrected, please add some explicit zeros to this example.

Orange/statistics/util.py
```diff
@@ -10,80 +10,181 @@
...
+    """ Check if sparse matrix contains any implicit zeros. """
+    if not sp.issparse(x):
+        raise TypeError('The matrix provided was not sparse.')
+    return np.prod(x.shape) != x.nnz
 
 
-def bincount(X, max_val=None, weights=None, minlength=None):
+def sparse_zero_weights(x, weights):
```
nikicc: Is this only meant to be used when both `x` and `weights` are of the same shape, i.e. one dimensional?

pavlin-policar: Adding a check here causes more problems than it's worth, because this can be called in two ways that are valid:
1. `x` has shape `(1, n)`
2. `x` has shape `(n, 1)`
These are basically equivalent since they are both "1d", but this would make the check more complicated.

nikicc:
These commits were moved to PR #2698.

PR #2561, opened 2017-09-05
[ENH] Add-on installation with Conda

astaric:
##### Issue
Some add-ons have dependencies that are a pain to install on windows. Conda makes this easier

##### Description of changes
Add-on installation/upgrading/uninstallation calls conda commands before pip. If conda command succeeds, pip says that everything is up to date/uninstalled and does nothing.

Needs testing on win/linux, with different configurations (orange installed in conda root/conda environment, orange installed beside conda, ...)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/addons.py
+++ b/Orange/canvas/application/addons.py
@@ -38,8 +38,8 @@ from AnyQt.QtGui import (
 
 from AnyQt.QtCore import (
     QSortFilterProxyModel, QItemSelectionModel,
-    Qt, QObject, QMetaObject, QEvent, QSize, QTimer, QThread, Q_ARG
-)
+    Qt, QObject, QMetaObject, QEvent, QSize, QTimer, QThread, Q_ARG,
+    QSettings)
 from AnyQt.QtCore import pyqtSignal as Signal, pyqtSlot as Slot
 
 from ..gui.utils import message_warning, message_information, \
@@ -861,7 +861,12 @@ class PipInstaller:
 
 class CondaInstaller:
     def __init__(self):
-        self.conda = self._find_conda()
+        enabled = QSettings().value('add-ons/allow-conda-experimental',
+                                    False, type=bool)
+        if enabled:
+            self.conda = self._find_conda()
+        else:
+            self.conda = None
 
     def _find_conda(self):
         executable = sys.executable
--- a/Orange/canvas/application/addons.py
+++ b/Orange/canvas/application/addons.py
@@ -942,16 +942,6 @@ def python_process(args, script_name=None, **kwargs):
         # Don't run the script with a 'gui' (detached) process.
         dirname = os.path.dirname(executable)
         executable = os.path.join(dirname, "python.exe")
-        # by default a new console window would show up when executing the
-        # script
-        startupinfo = subprocess.STARTUPINFO()
-        if hasattr(subprocess, "STARTF_USESHOWWINDOW"):
-            startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
-        else:
-            # This flag was missing in inital releases of 2.7
-            startupinfo.dwFlags |= subprocess._subprocess.STARTF_USESHOWWINDOW
-
-        kwargs["startupinfo"] = startupinfo
 
     if script_name is not None:
         script = script_name
@@ -965,6 +955,12 @@ def python_process(args, script_name=None, **kwargs):
 
 
 def create_process(cmd, executable=None, **kwargs):
+    if hasattr(subprocess, "STARTUPINFO"):
+        # do not open a new console window for command on windows
+        startupinfo = subprocess.STARTUPINFO()
+        startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
+        kwargs["startupinfo"] = startupinfo
+
     return subprocess.Popen(
         cmd,
         executable=executable,
--- a/Orange/canvas/application/addons.py
+++ b/Orange/canvas/application/addons.py
@@ -756,6 +756,7 @@ class Installer(QObject):
         self.__interupt = False
         self.__queue = deque(steps)
         self.pip = PipInstaller(user_install)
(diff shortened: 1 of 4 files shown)


Orange/canvas/application/addons.py
```diff
@@ -760,127 +770,156 @@ def setStatusMessage(self, message):
...
-                        os.unlink(match)
-
-            if retcode != 0:
-                self.error.emit(fmt_cmd(cmd), pkg, retcode, output)
-                return
+        try:
+            if command == Install:
+                self.setStatusMessage(
+                    "Installing {}".format(pkg.installable.name))
+                if self.conda:
```
kernc: If `self.conda` is an instance of `CondaInstaller()`, shouldn't this check be `self.conda.conda`?

astaric: Technically yes, but since CondaInstaller [reimplements `__bool__`](https://github.com/biolab/orange3/pull/2561/files#diff-55a6a7f617395a75195203c03b442985R881), this is the same thing

ajdapretnar:
Not sure I tested it right, but I've made a new environment and got the branch from Github. Ran Orange and tried to install the add-ons via the dialog. Got this:

  File "c:\orange\orange3\Orange\canvas\application\addons.py", line 780, in _next
    self.pip.install(pkg.installable)
  File "c:\orange\orange3\Orange\canvas\application\addons.py", line 816, in install
    run_command(cmd)
  File "c:\orange\orange3\Orange\canvas\application\addons.py", line 917, in run_command
    raise CommandFailed(command, process.returncode, output)
  File "c:\orange\orange3\Orange\canvas\application\addons.py", line 742, in __init__
    cmd = " ".join(map(shlex.quote, cmd.cmd))
AttributeError: 'list' object has no attribute 'cmd'

Windows10.

astaric:
Fixed, can you try again?

ajdapretnar:
Um, now I don't get this error, but another issue occurs. Potentially not related to this PR, but I tried to installed all the add-ons and some crash normally (Geo, Recommendation, Timeseries), while the rest seem to get installed. However, when I re-run Orange, the entire program crashes. Ran with -l 4, I see that the issue is in Spark and Textable add-ons, most likely.

markotoplak:
Work on linux without conda. Works just as well as installing with pip directly (succeeds if pip succeeds, fails with an error of it).

kernc:
Even if a single addon is selected and the installation fails, the information dialog encouraging a restart still shows.
screenshot_2017-09-06_16-52-02

astaric:
Blocked by #2616

Until we install conda command and conda-forge channel with our miniconda installer, this will not work

astaric:
@kernc, @ajdapretnar, @markotoplak 

I think I am done. Could you try it out again?

Orange/canvas/application/addons.py
```diff
@@ -760,127 +770,170 @@ def setStatusMessage(self, message):
...
+            USER_SITE, re.sub('[^\w]', '_', dist.project_name))
+        pip_record = next(iglob(install_path + '*.dist-info/RECORD'),
+                          None)
+        if pip_record:
+            with open(pip_record) as f:
+                files = [line.rsplit(',', 2)[0] for line in f]
+        else:
+            files = [os.path.join(
+                USER_SITE, 'orangecontrib',
+                dist.project_name.split('-')[-1].lower()), ]
```
kernc: This assumes all user-installed add-ons are in `orangecontrib` package, but this is just an internal convention, with the real add-on being anything that provides an `orange3.addon` entry point.

kernc:
Installing e.g. Orange3-Timeseries
```text
Running conda install --yes Orange3-Timeseries
Fetching package metadata ...............
Solving package specifications: .

Package plan for installation in environment /home/jk/Downloads/miniconda3:

The following NEW packages will be INSTALLED:

    anyqt:              0.0.8-py35_1 conda-forge
    bottleneck:         1.2.1-py35_1 conda-forge
    dill:               0.2.6-py35_0 conda-forge
    keyring:            9.0-py35_0   conda-forge
    orange3:            3.4.1-py35_0 anaconda   
    orange3-timeseries: 0.3.1-py35_0 conda-forge
```
Orange3 (lower version) gets reinstalled.

astaric:
@kernc what are your conda channel priorities?

kernc:
The order of channels in .condarc:
```text
channels:
  - conda-forge
  - anaconda
  - anaconda-fusion
  - defaults
```

Orange/canvas/application/addons.py
astaric: So I should wrap this in try: except: pass?

This is the same code we are (using at the moment)[https://github.com/biolab/orange3/blame/master/Orange/canvas/application/addons.py#L844], I have just moved it to a separate class.

kernc: I know, and I wrote it, but it's wrong in its assumptions.

No idea how to make it correct.

HTH

ajdapretnar:
When installing Orange3-Network, installation seems to stall forever. It reaches 'Orange3-Network successfully installed' message, but it never exits the process. I also can't quit the dialog window and have to force quit in the console.

Also, installations prints every increment in the process, e.g. orange3-networks 3%, orange3-networks 4%, orange3-networks 5%... etc. Is there a silent mode we can turn on? :)

astaric:
I have added the -q flag to conda calls. The stalling is not related to conda, since it already happens sometimes with the master version of Orange.

astaric:
I have disabled the conda installation of add-ons by default (it can be enabled with a checkbox in settings), hidden the download progressbar (--quiet option is passed to conda).

I have updated the order of channels in the environment created by our miniconda installer, as having defaults on top downgrades orange whenever you install an add-on. Pip uninstall now only runs when conda fails to uninstall. And all package names passed to conda are now lowercase (heuristic, but works :)).

If there is anything that needs to be fixed, please let me know. Otherwise I consider this PR complete.

PR #2512, opened 2017-07-31
[ENH] FreeViz: new widget

jerneju:
##### Issue
No FreeViz widget.

##### Description of changes
1) Icon needed (Issue #2519 ). Maybe we can use Linear Projection's icon.
2) Tests coverage so far: 82% (Scatter Plot has 74%) Missing GUI tests!
3) Scripting part in a separate PR (need to discuss with somebody).
4) Now can handle data sets with many attributes.
5) Highly recommended that PR #2541 is accepted and merged before this PR. DONE

##### Includes
- [x] Code changes
- [x] Tests
- [ ] Documentation

PR #2563, opened 2017-09-07
[ENH] FreeViz script

jerneju:
##### Issue
No FreeViz projection.

##### Description of changes
FreeViz projection scripting part similar as **MDS**, **PCA**.

```python
from Orange.projection.freeviz import FreeViz
from Orange.data import Table
data = Table("iris")
freeviz = FreeViz()
model = freeviz(data)
new_data = model(data)
```
Code coverage is 98%.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation

### DIFF
--- a/Orange/projection/base.py
+++ b/Orange/projection/base.py
@@ -6,7 +6,21 @@ from Orange.base import _ReprableWithPreprocessors
 from Orange.misc.wrapper_meta import WrapperMeta
 import Orange.preprocess
 
-__all__ = ["Projector", "Projection", "SklProjector"]
+__all__ = ["LinearCombinationSql", "Projector", "Projection", "SklProjector"]
+
+
+class LinearCombinationSql:
+    def __init__(self, attrs, weights, mean=None):
+        self.attrs = attrs
+        self.weights = weights
+        self.mean = mean
+
+    def __call__(self):
+        if self.mean is None:
+            return ' + '.join('{} * {}'.format(w, a.to_sql())
+                              for a, w in zip(self.attrs, self.weights))
+        return ' + '.join('{} * ({} - {})'.format(w, a.to_sql(), m, w)
+                          for a, m, w in zip(self.attrs, self.mean, self.weights))
 
 
 class Projector(_ReprableWithPreprocessors):
--- a/Orange/projection/freeviz.py
+++ b/Orange/projection/freeviz.py
@@ -1,0 +1,348 @@
+
+import numpy
+import scipy.spatial
+
+
+def squareform(d):
+    """
+    Parameters
+    ----------
+    d : (N * (N - 1) // 2, ) ndarray
+        A hollow symmetric square array in condensed form
+
+    Returns
+    -------
+    D : (N, N) ndarray
+        A symmetric square array in redundant form.
+
+    See also
+    --------
+    scipy.spatial.distance.squareform
+    """
+    assert d.ndim == 1
+    return scipy.spatial.distance.squareform(d, checks=False)
+
+
+def row_v(a):
+    """
+    Return a view of `a` as a row vector.
+    """
+    return a.reshape((1, -1))
+
+
+def col_v(a):
+    """
+    Return a view of `a` as a column vector.
+    """
+    return a.reshape((-1, 1))
+
+
+def allclose(a, b, rtol=1e-5, atol=1e-8, equal_nan=False):
+    # same as numpy.allclose in numpy==1.10
+    return numpy.all(numpy.isclose(a, b, rtol, atol, equal_nan=equal_nan))
+
+
+def forces_regression(distances, y, p=1):
+    y = numpy.asarray(y)
+    ydist = scipy.spatial.distance.pdist(y.reshape(-1, 1), "sqeuclidean")
+    mask = distances > numpy.finfo(distances.dtype).eps * 100
+    F = ydist
+    if p == 1:
+        F[mask] /= distances[mask]
+    else:
+        F[mask] /= distances[mask] ** p
+    return F
+
+
+def forces_classification(distances, y, p=1):
+    diffclass = scipy.spatial.distance.pdist(y.reshape(-1, 1), "hamming") != 0
+    # handle attractive force
+    if p == 1:
+        F = -distances
+    else:
+        F = -(distances ** p)
+
+    # handle repulsive force
+    mask = (diffclass &
+            (distances > numpy.finfo(distances.dtype).eps * 100))
+    assert mask.shape == F.shape and mask.dtype == numpy.bool
+    if p == 1:
+        F[mask] = 1 / distances[mask]
+    else:
+        F[mask] = 1 / (distances[mask] ** p)
+    return F
+
+
+def gradient(X, embeddings, forces, embedding_dist=None, weights=None):
+    X = numpy.asarray(X)
+    embeddings = numpy.asarray(embeddings)
+
+    if weights is not None:
+        weights = numpy.asarray(weights)
+        if weights.ndim != 1:
+            raise ValueError("weights.ndim != 1 ({})".format(weights.ndim))
+
+    N, P = X.shape
+    _, dim = embeddings.shape
+
+    if not N == embeddings.shape[0]:
+        raise ValueError("X and embeddings must have the same length ({}!={})"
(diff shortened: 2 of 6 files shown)


jerneju:
@janezd

janezd:
Could you rebase #2512 to this, so I can observe this PR in actual action? You'll have to do it some day anyway.

#2512 does not have to be finished, I just want to see how the projection function works.

jerneju:
@janezd: Rebased.

jerneju:
@janezd : Is this ok to merge?

Orange/projection/freeviz.py
```diff
@@ -0,0 +1,467 @@
...
+
+
+class FreeViz(Projector):
+    name = 'FreeViz'
+    supports_sparse = False
+
+    def __init__(self, weights=None, center=True, scale=True, dim=2, p=1,
+                 initial=None, maxiter=500, alpha=0.1, atol=1e-5, preprocessors=None):
+        super().__init__(preprocessors=preprocessors)
+        self.params = vars()
```
janezd: We use this monstrosity only in SKL wrappers. I think `FreeViz` does not need `params`.

Orange/projection/freeviz.py
```diff
@@ -0,0 +1,467 @@
...
+        self.p = p
+        self.initial = initial
+        self.maxiter = maxiter
+        self.alpha = alpha
+        self.atol = atol
+        self.is_class_discrete = False
+
+    def __call__(self, data):
+        if data and data.domain.class_var.is_discrete:
+            self.is_class_discrete = True
```
janezd: The same instance of `FreeViz` can be called multiple times and with different data sets, hence it can't count that `is_class_discrete` is `False` (as set in the constructor`.

```
if data is not None:
    self.is_class_discrete = data.domain.class_var.is_discrete
```

Orange/projection/freeviz.py
```diff
@@ -0,0 +1,467 @@
...
+                                      scale=self.scale, dim=self.dim, p=self.p,
+                                      initial=self.initial, maxiter=self.maxiter, alpha=self.alpha,
+                                      atol=self.atol, is_class_discrete=self.is_class_discrete)
+            self.components_ = a
+            return FreeVizModel(self, self.domain)
+
+    def transform(self, X):
+        EX = np.dot(X, self.components_)
+        return EX
+
```
janezd: @markotoplak, the code from the beginning of the file and up to here is adapted from PCA. I know the idea, but you're an expert here. Can you take a quick look, please?

Orange/projection/freeviz.py
```diff
@@ -0,0 +1,467 @@
...
+    def __init__(self, freeviz):
+        self.freeviz = freeviz
+
+    def __call__(self, data):
+        if data.domain != self.freeviz.pre_domain:
+            data = data.transform(self.freeviz.pre_domain)
+        return self.freeviz.transform(data.X)
+
+
+class _LinearCombination:
```
janezd: @lanzagar, could `Orange.projection.pca._LinearCombination` be moved to Orange.projection.base.LinearCombination`, so it could be reused here?

Also, `PCAModel` is just a specific case of linear model. Could the common functionality be extracted into a base class? Could you and @jerneju discuss this? I don't insist it should be done, I'm just suggesting considering it. And it doesn't have to be in this PR, either.

Orange/projection/freeviz.py
jerneju: Fixed.

Orange/projection/freeviz.py
```diff
@@ -0,0 +1,467 @@
...
+    """Transform into a given FreeViz component."""
+
+    def __init__(self, projection, feature, freeviz_transform):
+        super().__init__(freeviz_transform)
+        self.projection = projection
+        self.feature = feature
+        self.transformed = None
+
+    def compute(self, data, freeviz_space):
+        if self.feature >= freeviz_space.shape[1]:
```
markotoplak: How could this happen at all?

Orange/projection/freeviz.py
markotoplak: Seems OK.

I think it would be useful to add a test, where we would
1. open data, apply some preprocessor, splits the data into two parts, use FreeViz on the first part, and then transform the second part,
2. open data, split into two parts, apply the same preprocessor and FreeViz only on the first part, and then transform the second part,

where the transformed second part in (1) and (2) has to be the same.

Orange/projection/freeviz.py
jerneju: Thanks. Fixed.

Orange/projection/freeviz.py
lanzagar: 1) Could and should.
2) Could. Probably should? Definitely should be considered.

We'll take a look.

Orange/projection/freeviz.py
```diff
@@ -0,0 +1,463 @@
+
+import numpy as np
+import scipy.spatial
+
+from Orange.data import ContinuousVariable, Domain
+from Orange.data.util import SharedComputeValue
+from Orange.projection import Projector, Projection
+
+
```
lanzagar: Define `__all__` here and pull `FreeViz` up into Orange.projection.
Tests and documentation should then import from there.

Orange/projection/base.py
```diff
@@ -6,7 +6,21 @@
 from Orange.misc.wrapper_meta import WrapperMeta
 import Orange.preprocess
 
-__all__ = ["Projector", "Projection", "SklProjector"]
+__all__ = ["LinearCombinationSql", "Projector", "Projection", "SklProjector"]
+
+
+class LinearCombinationSql:
```
janezd: I don't know what this actually does, but I suppose that this is a general linear combination, and it also supports SQL -- and not a linear combination that requires SQL, right? If so, the name is misleading. @lanzagar, could you help here?

astaric: This is the sql equivalent of the compute_value. It only works for sql, as in-memory data is processed with a different object - Projector.

janezd: Ah, OK, I see. Thanks.

Orange/tests/test_freeviz.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def test_initial(self):
+        FreeViz.init_radial(1)
+        FreeViz.init_radial(2)
+        FreeViz.init_radial(3)
+        FreeViz.init_random(2, 4, 5)
+
+    def test_transform_changed_domain(self):
+        """
+        Thest where:
```
janezd: Thest -> test

Orange/tests/test_freeviz.py
```diff
@@ -0,0 +1,101 @@
...
+        ndata = normalize(data)
+        model = freeviz(ndata[:100])
+        result_1 = model(ndata[100:])
+
+        # normalize only the "training" part
+        ndata = normalize(data[:100])
+        model = freeviz(ndata)
+        result_2 = model(data[100:])
+
+        self.assertTrue(np.allclose(result_1, result_2))
```
janezd: Why not `np.testing.assert_almost_equal`?

@markotoplak, the test is otherwise OK, right?

jerneju: > ======================================================================
> ERROR: test_transform_changed_domain (Orange.tests.test_freeviz.TestFreeviz)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/travis/build/biolab/orange3/build/travis-test/Orange/tests/test_freeviz.py", line 101, in test_transform_changed_domain
>     np.testing.assert_almost_equal(result_1, result_2)
>   File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/numpy/testing/utils.py", line 568, in assert_almost_equal
>     if not (gisfinite(desired) and gisfinite(actual)):
> ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
>

jerneju:
```python
numpy.testing.assert_allclose
```

markotoplak: The test is almost OK.

To make results repeatable, init_random should be fixed so that it uses rstate = np.random.RandomState(0) if not given. Then, inititalization should be removed from the test.

The comment on the test is superfluous - the code and two in code comments say it all.

Issue #2404, opened 2017-06-14
"field larger than field limit (131072)" error


pdc-kaminaga:
I tried File / CSV File Load.
But I got a "field larger than field limit (131072)" error.
Please tell me how to solve this problem.

kernc:
`cut -c -200 in > out`

More seriously, if your file doesn't contain single fields of over 130K of characters, then you might have a delimiter mismatch. Can you paste an example line from the file?

-----
While using `csv`, we might want to [increase field_size_limit](https://stackoverflow.com/questions/15063936/csv-error-field-larger-than-field-limit-131072).

pdc-kaminaga:
It was as you pointed out.
I update the data and try again.
Thank you.

X4:
Reopening, because I run into the same issue, like many others who may have long texts in their dataset.
What's the apparent solution here? Becaues adding a python widget into the pane with the stackoverflow fix doesn't solve it.

kernc:
Fixed in https://github.com/biolab/orange3/pull/2564.

PR #2564, opened 2017-09-08
Increase field limit

ajdapretnar:
##### Issue

Attempt at solving #2404. 

##### Description of changes
Increase csv field limit.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -31,6 +31,9 @@ from Orange.data import (
 from Orange.util import Registry, flatten, namegen
 
 
+# Support values longer than 128K (i.e. text contents features)
+csv.field_size_limit(100*1024*1024)
+
 __all__ = ["Flags", "FileFormat"]
 
 


thocevar:
There seems to be a problem with sys.maxsize and underlying C types on some 64-bit systems (OverflowError: Python int too large to convert to C long). Consider replacing with 100*1024*1024. 100 MB should be more than enough. Besides, other components might choke on larger fields anyway.

ajdapretnar:
So the argument should be 100*1024*1024 instead of sys?

ajdapretnar:
Fixed.

Orange/data/io.py
```diff
@@ -749,6 +749,10 @@ class CSVReader(FileFormat):
     SUPPORT_SPARSE_DATA = False
     PRIORITY = 20
 
+    def __init__(self, filename):
+        super().__init__(filename)
+        csv.field_size_limit(100*1024*1024)
```
kernc: This is a global side-effect. Put the line on top in a global namespace, somewhere shorty below the `csv` import. Precede by a comment:
> Support values longer than 128K (i.e. text contents features)

ajdapretnar:
@kernc @thocevar I hope this is what you had in mind.

PR #2575, opened 2017-09-11
[FIX] widgets/tests: Compatibility with Python 3.5.{0,1}

ales-erjavec:
##### Issue

typing.Type was introduced only in Python 3.5.2 

##### Description of changes

try-except the import.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
I'm stupid -- of course we don't need to import it if it's not there; it appears just in the comments. :)

PR #2577, opened 2017-09-11
preprocess.randomize: Don't shuffle entire rows

janezd:
### Issue

9b5d6c4d57b0e47627b2e185d85e19b606131bef has broken shuffling of attributes in Randomize widget: `skl_shuffle` shuffles whole rows, so the effect of shuffling attributes is the same as of shuffling the class.

### Fix

Dense data is handled as before 9b5d6c4d57b0e47627b2e185d85e19b606131bef.

Sparse data is converted to csc and indices are replaced with random indices. These indices are unique within each row.

The new method for randomization is now tested. I haven't added any other tests for this (otherwise untested) class.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/preprocess/preprocess.py
+++ b/Orange/preprocess/preprocess.py
@@ -4,8 +4,8 @@ Preprocess
 
 """
 import numpy as np
+import scipy.sparse as sp
 import sklearn.preprocessing as skl_preprocessing
-from sklearn.utils import shuffle as skl_shuffle
 import bottleneck as bn
 
 import Orange.data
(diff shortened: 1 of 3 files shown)


Orange/preprocess/preprocess.py
```diff
@@ -363,7 +363,22 @@ def __call__(self, data):
         return new_data
 
     def randomize(self, table):
-        return skl_shuffle(table, random_state=self.rand_seed)
+        np.random.seed(self.rand_seed)
```
markotoplak: Why don't you rather initialize a RandomState with that seed and then use that?

PR #2578, opened 2017-09-12
BoxPlot: Reorder the attributes to put the class at the top

janezd:
##### Issue

The class is selected as the default grouping attribute, yet it's hidden on the bottom of the list.

Furthermore, we have pushed classed to the top in other widgets.

##### Description of changes

Reorder to put the class at the top.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -204,6 +204,7 @@ class OWBoxPlot(widget.OWWidget):
             order=order,
             placeholder="None",
             valid_types=Orange.data.DiscreteVariable)
+        self.group_vars.clear()  # Remove 'None' from the list view
         view = gui.listView(
             self.controlArea, self, "group_var", box="Subgroups",
             model=self.group_vars, callback=self.grouping_changed)
@@ -345,6 +346,7 @@ class OWBoxPlot(widget.OWWidget):
         self.infot1.setText("")
         self.attrs.set_domain(None)
         self.group_vars.set_domain(None)
+        self.group_vars.clear()  # Remove 'None' from the list view
         self.is_continuous = False
         self.update_display_box()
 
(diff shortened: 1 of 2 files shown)


astaric:
Could we use DomainModel here? That is what we used in other widgets when we were changing ordering.

janezd:
`DomainModel` in other widgets is only used in combos.

I tried it now and it works, except that it can't include separators.

astaric:
There is a failing boxplot test on travis. Can you take a look?

janezd:
`DomainModel` keeps "None" as an option even when there is no domain. This is needed in combo boxes, where `DomainModel` is mostly used. In lists, it's ugly:

screen shot 2017-09-13 at 20 21 17

"Clean" solution: add yet another optional flag to `DomainModel`'s constructor that would tell it not to show "None" if there is no domain. This would be ambiguous since the model can also include strings. What happens with them when there is no domain? Besides, I don't want to multiply the number of arguments.

Implemented solution: calling model's `clear()` after setting the domain to `None`. In separate commit for easier reversal if we decide against it.

Orange/widgets/visualize/owboxplot.py
```diff
@@ -270,14 +277,13 @@ def set_data(self, dataset):
...
-            self.attrs[:] = chain(domain.variables,
-                                  (a for a in domain.metas if a.is_primitive()))
+            self.group_vars.set_domain(domain)
+            self.attrs.set_domain(domain)
             if self.attrs:
-                self.attribute = self.attrs[0]
+                # Select the first non-class variable; if all are classes,
+                # select the first class
+                self.attribute = \
+                    self.attrs[len(domain.class_vars) % len(self.attrs)]
```
astaric: I don't get the formula (`len(domain.class_vars) % len(self.attrs)`)

How does this "select the first non-class variable; if all are classes, select the first class"

janezd: Let's say that there are 5 classes and 2 other attributes (or metas). Then `len(domain.class_vars)` is 5 and `len(self.attrs)` is 7, so `5 % 7` is 5 and you get the first non-class variable.

If there are 5 classes and 0 other attributes, `len(domain.class_vars)` and `len(self.attrs)` equal 5, so `5 % 5`  = 0 and you get the first class.

PR #2579, opened 2017-09-12
[ENH] DomainEditor: Add horizontal header

janezd:
##### Issue

DomainEditor (e.g. in the File Widget) does not have horizontal header and hence doesn't allow the user to resize the columns. This is a problem when the features have longer names.

##### Description of changes

I added the header. While it is better in terms of functionality, the previous design may have been less cluttered. @BlazZupan, what do you think?

screen shot 2017-09-11 at 21 57 27

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/utils/domaineditor.py
+++ b/Orange/widgets/utils/domaineditor.py
@@ -97,6 +97,13 @@ class VarTableModel(QAbstractTableModel):
             self.dataChanged.emit(index.sibling(row, 0), index.sibling(row, 3))
             return True
 
+    def headerData(self, i, orientation, role=Qt.DisplayRole):
+        if orientation == Qt.Horizontal and role == Qt.DisplayRole and i < 4:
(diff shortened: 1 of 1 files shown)


ajdapretnar:
Oh yes, I hated this, too!

janezd:
Codecov/patch failed since tests don't seem to go into `return Qt.AlignLeft` (although the view in fact uses it) in

    if role == Qt.TextAlignmentRole:
        return Qt.AlignLeft

I can obviously write a test that covers this line, but ... won't since there's nothing to test here.

ajdapretnar:
This caused other problems. Now, it is impossible to resize the widget, even when all data are continuous (Values field is empty). Could you give it a look?

screen shot 2017-09-13 at 13 41 24

janezd:
I can't replicate this -- I can resize it.

ajdapretnar:
Updated the master and now I can, too. Sorry for the panic. :O

PR #2580, opened 2017-09-12
Replace string with text as variable type (in GUI only)

janezd:
##### Issue

Many muggles don't understand what's a *string*. They know what's *text*, though.

##### Description of changes

It seems that this appears only in Domain editor, and in icons. Fixing the former is trivial. <strike>For icons, "T" is already used for time variables. However, I think it's better to have the same letter (in an icon of a different color) then having the cryptic "string".

This follows the earlier change in which discrete and continuous were replaced by categorical and numeric.</strike> Icons are not changed to avoid confusion.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/data/tests/test_owfile.py
+++ b/Orange/widgets/data/tests/test_owfile.py
@@ -95,7 +95,7 @@ class TestOWFile(WidgetTest):
 
         self.open_dataset("iris")
         idx = self.widget.domain_editor.model().createIndex(4, 1)
-        self.widget.domain_editor.model().setData(idx, "string", Qt.EditRole)
+        self.widget.domain_editor.model().setData(idx, "text", Qt.EditRole)
         self.widget.apply_button.click()
         data = self.get_output(self.widget.Outputs.data)
         self.assertIsInstance(data.domain["iris"], StringVariable)
@@ -263,7 +263,7 @@ a
                 self.assertEqual(str(a), model.data(model.createIndex(i, 0), Qt.DisplayRole))
             # make conversions
             model.setData(model.createIndex(0, 1), "categorical", Qt.EditRole)
-            model.setData(model.createIndex(1, 1), "string", Qt.EditRole)
+            model.setData(model.createIndex(1, 1), "text", Qt.EditRole)
             model.setData(model.createIndex(2, 1), "numeric", Qt.EditRole)
             model.setData(model.createIndex(3, 1), "numeric", Qt.EditRole)
             model.setData(model.createIndex(6, 1), "numeric", Qt.EditRole)
--- a/Orange/widgets/utils/domaineditor.py
+++ b/Orange/widgets/utils/domaineditor.py
@@ -35,7 +35,7 @@ class VarTableModel(QAbstractTableModel):
     DISCRETE_VALUE_DISPLAY_LIMIT = 20
 
     places = "feature", "target", "meta", "skip"
-    typenames = "categorical", "numeric", "string", "datetime"
+    typenames = "categorical", "numeric", "text", "datetime"
     vartypes = DiscreteVariable, ContinuousVariable, StringVariable, TimeVariable
     name2type = dict(zip(typenames, vartypes))
     type2name = dict(zip(vartypes, typenames))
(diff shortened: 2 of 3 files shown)


astaric:
While I do not completely disagree with the change, it introduces some inconsistencies.

In header of the datafile (second row), you can now write c, d, s or t which correspond to Numeric, Categorical, Text and Time in gui. Sholud we allow new names there as well?

ajdapretnar:
I agree, it makes more sense for lay user. Not a fan of two Ts, but you're right, it's the lesser evil. Hence I vote 👍.

And now me being evil: would you please update the documentation, too? 😈 Or force the person who did the previous change (continuous - numeric) to do it. 😁

kernc:
T in time could also be a D as in datetime, since the latter letter is now unused.

janezd:
@astaric: Besides 's' and 'string', the attribute can already also be described as 'text'. I'd promote this form. Besides that, we already have a huger inconsistency: 'c' is used to describe numeric attributes in files.

@kernc: I thought about using 'D'. But it increases the mess because 'd' in files is now used for categorical attributes (which are presented as C in widgets). I don't totally dislike the idea of using D, but since icons have different colors, I'd prefer the two-T solution.

Bottom line: we should undocument single letter descriptions in data files.

ajdapretnar:
I believe icons would be even more confusing. Current letters might be strange at first, but when you get used to them, they are straightforward and easy to recognize. I don't think any new user was particularly confused about it... (my personal perception during workshops)

kernc:
Icons would still need to be unicolor, uniform style. For a number, a curve of a continuous bell-shaped function; for categorical, a histogram.

janezd:
@ajdapretnar Why do you think icons would be more confusing - if they are well chosen?

@kernc I thought about curves and histograms, too. But I didn't like them because not all numeric variables are perceived as curve-like, and the connection between categories and histograms is also indirect. Curves and histograms are not general symbols for numbers and categories. This presentation is related just to a particular visualization and may not be obvious on the first glance. On the other hand, letters are even less obvious.

Changing this affects screenshots in throughout documentation and teaching materials hence I wouldn't change N and C until (if ever) we have a suggestion we totally like.

For now I suggest:

- @ajdapretnar decides whether we change S -> T and consequently T -> D.
- we undocument single-letter type specifications since they do not match icons and user-visible type names.

ajdapretnar:
> @ajdapretnar Why do you think icons would be more confusing - if they are well chosen?

You almost answered it yourself 😀  

> @kernc I thought about curves and histograms, too. But I didn't like them because not all numeric variables are perceived as curve-like, and the connection between categories and histograms is also indirect. Curves and histograms are not general symbols for numbers and categories. This presentation is related just to a particular visualization and may not be obvious on the first glance. On the other hand, letters are even less obvious.

I think we are causing more confusion than solving. I'm afraid icons would be so small they would become illegible and hence confusing. Also, if there are no obvious symbols (as it seems there aren't), we can't expect the users to _get it_ right away either.

My opinion: if the current system is well-documented, then people can get used to it quickly. Personally, I love the color coded squares. It makes it so easy to recognize variable type!

> @ajdapretnar decides whether we change S -> T and consequently T -> D.

Fine. It would be interesting to know, what is the level of confusion this solves and/or introduces. :)

janezd:
> Fine. It would be interesting to know, what is the level of confusion this solves and/or introduces. :)

Maybe 0 and 0. :)

How do you explain, in your workshops, why S represents text? If this is not an issue at all, we can keep S.

ajdapretnar:
That's the thing - nobody has ever asked that at any of the workshops!

Btw, I have asked my friends about this (strictly non-CS and all with MAs) and one said 'numeric', 'categorical' and 'text' seems clearer to her, but she wouldn't mind at all to go with the rest if they are explained somewhere. :D

janezd:
As suspected. :)

I changed the PR so that it only replaces "string" with "text" in the File widget (and related tests) and in the tooltip for the "S" icon. The icon doesn't match the type name (visible by the user) but the icons are just decorative anyway.

markotoplak:
@janezd , @ajdapretnar , @kernc  I do not think icons need to represent the concepts literally, they just need to be memorable and show differences. Letters in a box defeat that purpose. I vote for any graphical metaphor - curves versus histrograms, color gradient vs sharply separated colors, even coffee cups vs a waterfall (only slightly joking - still better than letters for me).

Issue #1581, opened 2016-09-20
Orange no longer remembers the number of decimal places


janezd:
Orange used to keep track of the number of decimal places it should print for a specific variable. For instance, the "ST by exercise" has only a single decimal in the heart_disease.tab file, and the "age" and the "number of vessels colored" have none, obviously. Orange used to record this when loading the data. Now all numeric values have three decimals.

PR #2583, opened 2017-09-13
io: Compute number of decimals for numeric variables

janezd:
##### Issue

screen shot 2017-09-12 at 20 52 10

screen shot 2017-09-12 at 20 51 32

##### Description of changes

screen shot 2017-09-12 at 20 54 58

screen shot 2017-09-12 at 20 49 52

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -133,6 +133,27 @@ def guess_data_type(orig_values):
 
 def sanitize_variable(valuemap, values, orig_values, coltype, coltype_kwargs,
                       domain_vars, existing_var, new_var_name, data=None):
+    def reorder_values():
+        new_order, old_order = \
+            var.values, coltype_kwargs.get('values', var.values)
+        if new_order != old_order:
+            offset = len(new_order)
+            column = values if data.ndim > 1 else data
+            column += offset
+            for _, val in enumerate(var.values):
+                try:
+                    oldval = old_order.index(val)
+                except ValueError:
+                    continue
+                bn.replace(column, offset + oldval, new_order.index(val))
+
+    def get_number_of_decimals():
+        ndecimals = max((len(value) - value.find(".")
+                         for value in orig_values
+                         if isinstance(value, str) and "." in value),
+                        default=1)
+        return ndecimals - 1
+
     if valuemap:
         # Map discrete data to ints
         def valuemap_index(val):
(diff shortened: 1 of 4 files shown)


janezd:
Fixes #1581.

Orange/data/io.py
```diff
@@ -156,20 +177,17 @@ def valuemap_index(val):
...
-                        oldval = old_order.index(val)
-                    except ValueError:
-                        continue
-                    bn.replace(column, offset + oldval, new_order.index(val))
+            # Reorder discrete values to match existing variable
+            reorder_values()
+
+    if isinstance(var, ContinuousVariable) and var.adjust_decimals:
+        ndecimals = get_number_of_decimals()
+        if var.adjust_decimals == 2 or ndecimals > var.number_of_decimals:
```
astaric: I do not understand what `var.adjust_decimals == 2` means

Orange/data/io.py
```diff
@@ -133,6 +133,27 @@ def guess_data_type(orig_values):
 
 def sanitize_variable(valuemap, values, orig_values, coltype, coltype_kwargs,
                       domain_vars, existing_var, new_var_name, data=None):
+    def reorder_values():
```
astaric: I would rather have this in a separate commit. I assume you just moved the code into a separate method to make code more readable (and radon happy) but I might be wrong.

Orange/data/io.py
janezd: See here: https://github.com/biolab/orange3/blob/master/Orange/data/variable.py#L439

Orange/data/io.py
janezd: I agree and I'd usually do so. But if you don't insist I wouldn't waste time tearing this apart. It's highly unlikely that anybody would ever again see this. Splitting this PR into two commits would be useful if somebody would want to revert just a part of this PR. Probably won't happen.

astaric: Leave it as it is. I commented after a spent a could of minutes figuring out what you changed,only to realise it was just refactoring.

Orange/data/io.py
astaric: It would be nice to have named constants instead of magic values (for instance: 0=NEVER, 1=MORE_PRECISE, 2=ALWAYS).
A comment why this is needed here would be nice as well, as the documentation on the variable states that this behaviour is what ContinuousVariable does by itself.

The code also differs a bit from the documented behaviour of the ContinuousVariable: adjust_decimals is supposed to be set to 1 after number_of_decimals is set for the first time.

Orange/data/io.py
astaric: Leave it as it is. I wrote the comment after I spent a few minutes figuring out what changed in the code, only to find out that it was just some refactoring.

Orange/data/io.py
janezd: I added a comment and changed the code so that it sets `adjust_decimals` to 1. I removed a redundant part where this couldn't be done.

I agree about the constants, but not in this PR. In particular because `ContinuousVariable.to_val` no longer adjusts the number of decimals, so other fixes are needed there. I'll open an issue to discuss this.

astaric:
I'll merge this once the tests pass. Refactoring/cleanup can be done as a part of #2593

PR #2584, opened 2017-09-13
[FIX] Fix saving reports on Python 3.6

janezd:
##### Issue

This is urgent. The current release is broken: Reports cannot be saved because `ReportItems` (derived from `QStandardItem`) can't be pickled.

##### Description of changes

Add `ReportItem.__getnewargs__`.

I don't have Python 3.6 env on this machine so I ran tests from a bundle. Can somebody with "real" Python 3.6 check this?

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/canvas/report/owreport.py
+++ b/Orange/canvas/report/owreport.py
@@ -56,6 +56,10 @@ class ReportItem(QStandardItem):
         self.id = id(icon)
         super().__init__(icon, name)
 
+    def __getnewargs__(self):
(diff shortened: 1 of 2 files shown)


janezd:
The problem is probably caused by this change in Python: https://bugs.python.org/issue22995.

janezd:
Codecov/patch coverage failed in files unrelated to this PR.

astaric:
I have tested on "normal" python 3.6. Test passes, reports can be saved and loaded.

PR #2591, opened 2017-09-13
[FIX] canvas: Fix link runtime state modeling

ales-erjavec:
##### Issue

Links in workflow previews are rendered as dotted lines. This is due to workflow mode's inability to account for the possibility that it has no associated runtime execution state.

##### Description of changes

Add (and use as default) a `NoState` flag to account for no runtime execution state. Style the graphical presentation accordingly.
 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/scheme/link.py
+++ b/Orange/canvas/scheme/link.py
@@ -83,15 +83,17 @@ class SchemeLink(QObject):
         """
         Flags indicating the runtime state of a link
         """
+        #: The link has no associated state.
+        NoState = 0
         #: A link is empty when it has no value on it
-        Empty = 0
+        Empty = 1
         #: A link is active when the source node provides a value on output
-        Active = 1
+        Active = 2
         #: A link is pending when it's sink node has not yet been notified
         #: of a change (note that Empty|Pending is a valid state)
-        Pending = 2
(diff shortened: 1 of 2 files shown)


Orange/canvas/scheme/link.py
```diff
@@ -83,15 +83,17 @@ class State(enum.IntEnum):
...
         #: A link is empty when it has no value on it
-        Empty = 0
+        Empty = 1
         #: A link is active when the source node provides a value on output
-        Active = 1
+        Active = 2
         #: A link is pending when it's sink node has not yet been notified
         #: of a change (note that Empty|Pending is a valid state)
-        Pending = 2
+        Pending = 4
```
astaric: Is Pending a valid state or is it always Empty|Pending or Active|Pending?

ales-erjavec: Let's say it is valid.

astaric: I asked because it used to be marked with a dashed line (it was not active), but it is marked with a solid line in this PR (it is not empty). If the state does not occur normally by itself, it is not a problem.

PR #2592, opened 2017-09-13
travis: Add python3.6 to build matrix

astaric:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


astaric:
The decrease in overall coverage is caused by mssql backend no longer being covered (as the pymssql library is no longer installed) Since there were no actual tests for that code, I do not think this is a problem.

PR #2595, opened 2017-09-18
[FIX] OWBoxPlot: Faster selection

lanzagar:
##### Issue

Selection of instances in Box Plot had complexity O(n_all * n_selected), making it too slow / freeze for data with a couple thousand instances.

##### Description of changes
Use faster membership testing.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -885,10 +885,11 @@ class OWBoxPlot(widget.OWWidget):
         selected, selection = None, []
         if self.conditions:
             selected = Values(self.conditions, conjunction=False)(self.dataset)
-            selection = [i for i, inst in enumerate(self.dataset)
-                         if inst in selected]
+            selection = np.in1d(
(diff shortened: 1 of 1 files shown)


Orange/widgets/visualize/owboxplot.py
```diff
@@ -885,10 +885,12 @@ def commit(self):
         selected, selection = None, []
         if self.conditions:
             selected = Values(self.conditions, conjunction=False)(self.dataset)
-            selection = [i for i, inst in enumerate(self.dataset)
-                         if inst in selected]
+            selected_ids = set(selected.ids)
+            selection = [i for i, id in enumerate(self.dataset.ids)
+                         if id in selected_ids]
```
kernc: How about:
```py
selection = np.in1d(self.dataset.ids, selected.ids).nonzero()[0]
```

lanzagar: Even better 👍

PR #2598, opened 2017-09-19
[FIX] Saving of multiple selections in ScatterPlot

markotoplak:
##### Issue
Selection saving did not work.

##### Description of changes
Selection saving needed only a small fix. To also save different selection groups I changed the format of saved selections.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -117,6 +117,7 @@ class OWScatterPlot(OWWidget):
         annotated_data = Output(ANNOTATED_DATA_SIGNAL_NAME, Table)
         features = Output("Features", Table, dynamic=False)
 
+    settings_version = 2
     settingsHandler = DomainContextHandler()
 
     auto_send_selection = Setting(True)
@@ -125,7 +126,7 @@ class OWScatterPlot(OWWidget):
 
     attr_x = ContextSetting(None)
     attr_y = ContextSetting(None)
-    selection = Setting(None, schema_only=True)
+    selection_group = Setting(None, schema_only=True)
 
     graph = SettingProvider(OWScatterPlotGraph)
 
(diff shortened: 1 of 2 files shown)


markotoplak:
Potential problem: the save files are incompatible with older versions. As far as I tested there is no mechanism that would notify users that they are trying to open a file that is too new for their Orange installation.

markotoplak:
How to properly test saving of schema_only settings? test_points_selection in test_owscatterplot only tested loading of settings - and that is why we did not catch this earlier.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -518,6 +522,13 @@ def onDeleteWidget(self):
         self.graph.plot_widget.getViewBox().deleteLater()
         self.graph.plot_widget.clear()
 
+    @classmethod
+    def migrate_settings(cls, settings, version):
+        if "selection" in settings \
```
astaric: I would prefer
```
if version < 2:
```

astaric:
For testing saving of settings, you can use something like this:
https://github.com/biolab/orange3/blob/master/Orange/widgets/visualize/tests/test_owscatterplot.py#L256

Current widget settings can be packed
```
settings = self.widget.settingsHandler.pack_data(self.widget)
```
and restored to a new setting instance
```
w = self.create_widget(OWScatterPlot, stored_settings=settings)
```
The new widget gets the settings if would get if the first widget was saved to a file.

astaric:
As for the compatibility with older version, you could use a new setting name. This way, when opening a workflow with an old version of Orange, the selection would be lost, but there would be no crash.

markotoplak:
I added saving tests to the first commit and also renamed the setting.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -518,6 +522,11 @@ def onDeleteWidget(self):
         self.graph.plot_widget.getViewBox().deleteLater()
         self.graph.plot_widget.clear()
 
+    @classmethod
+    def migrate_settings(cls, settings, version):
+        if version < 2 and "selection" in settings and settings["selection"]:
```
astaric:
```
if version < 2 and settings.get(selection):
```

PR #2599, opened 2017-09-19
[MNT] Cleanup domain context handler

astaric:
##### Issue
DomainContextHandler used to distinguish between attribute and metas, allowing the programmer to create settings that could contain only attributes, metas, or both. The widgets have since been modified to work with variables regardless of their position in domain, but the flags in DomainContextHandler remained.

Another feature that died were reservoirs - a way to collect a list of all variables present in domain that were not used by other settings. 

##### Description of changes
Remove the code that is not used anymore

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


Orange/widgets/settings.py
```diff
@@ -949,21 +950,12 @@ def filter_value(self, setting, data, domain, attrs, metas):
...
-        for setting, data, instance in \
-                self.provider.traverse_settings(data=context.values, instance=widget):
-            if not isinstance(setting, ContextSetting) or setting.name not in data:
-                continue
-
-            value = self.decode_setting(setting, data[setting.name], domain)
-            setattr(instance, setting.name, value)
-            if hasattr(setting, "selected") and setting.selected in data:
-                setattr(instance, setting.selected, data[setting.selected])
+        # our decode_setting needs domain to be able to restore
```
astaric: @janezd, @lanzagar 
What do you think? While I admit that this is a bit hackish, it avoids having a copying the whole function to change a single line.

janezd: I can't believe this is coming from the guy who overruled my excellent modulo trick a few days ago. :)

Is the line in question this line?

``` 
value = self.decode_setting(setting, data[setting.name])
```

Shouldn't it actually be

``` 
value = self.decode_setting(setting, data[setting.name], *args)
```

`ContextHandler.settings_to_widget` accepts `*args` but doesn't use them. I think they were meant to be used exactly in this line.

Would this fix the problem?

Orange/widgets/settings.py
```diff
@@ -791,22 +790,25 @@ def close_context(self, widget):
         self.settings_from_widget(widget)
         widget.current_context = None
 
-    def settings_to_widget(self, widget, *args):
+    def settings_to_widget(self, widget, *args, decode_setting=None):
```
janezd: I find this barely better than patching.

janezd:
The most important comment (referring to patch) is marked as outdated. :) Here it is again.

----------

I can't believe this is coming from the guy who overruled my excellent modulo trick a few days ago. :)

Is the line in question this line?

``` 
value = self.decode_setting(setting, data[setting.name])
```

Shouldn't it actually be?

``` 
value = self.decode_setting(setting, data[setting.name], *args)
```

`ContextHandler.settings_to_widget` accepts `*args` but doesn't use them. I think they were meant to be used exactly in this line.

Would this fix the problem?

Orange/widgets/settings.py
```diff
@@ -877,19 +877,11 @@ def __init__(self, max_vars_to_pickle=100, match_values=0,
         self.attributes_in_res = attributes_in_res
         self.metas_in_res = metas_in_res
 
-        self.has_ordinary_attributes = attributes_in_res
-        self.has_meta_attributes = metas_in_res
+        self.has_ordinary_attributes = True
+        self.has_meta_attributes = True
```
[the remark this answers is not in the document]
astaric: Tha same question applies to all but one other attributes set in `__init__`. The only one that is actually used it match_values, the rest are either not used anymore or set in `super.__init__`. I'll clean them up.

Orange/widgets/settings.py
astaric: but it is more functional :)

astaric:
This would fix the problem. And break all existing Context Handlers that override decode_setting. We can do this at a later date, for this PR I will just revert the two commits that "simplify" the settings_to_widget method.

I guess I'll revert both commits and keep the duplicated function.

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -467,14 +467,10 @@ def _define_symbols():
...
 class OWScatterPlotGraph(gui.OWComponent, ScaleScatterPlotData):
-    attr_color = ContextSetting(
-        None, ContextSetting.OPTIONAL, exclude_metas=False)
-    attr_label = ContextSetting(
-        None, ContextSetting.OPTIONAL, exclude_metas=False)
-    attr_shape = ContextSetting(
-        None, ContextSetting.OPTIONAL, exclude_metas=False)
-    attr_size = ContextSetting(
-        None, ContextSetting.OPTIONAL, exclude_metas=False)
+    attr_color = ContextSetting(None, required=ContextSetting.OPTIONAL)
```
astaric: ContextSetting.OPTIONAL was passed to the not_attibute parameter, although the constant is meant for the required parameter.

janezd:
Not noticing a bug shouldn't be a reason to call it a feature. :)

If a Scatter plot is connected to a widget that may or may not emit the class, and the class is used for coloring, you still want the domains to match. Grouped selections in all projections (scatter, mds, linear...), for instance, emit a class with a different name than the classical single selection.

I vote for keeping (+ fixing).

astaric:
Fixed. I spent way longer on writing a test for this than I did on writing the fix.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -124,8 +124,8 @@ class Outputs:
     auto_sample = Setting(True)
     toolbar_selection = Setting(0)
 
-    attr_x = ContextSetting(None)
-    attr_y = ContextSetting(None)
+    attr_x = ContextSetting(None, required=ContextSetting.REQUIRED)
+    attr_y = ContextSetting(None, required=ContextSetting.REQUIRED)
```
janezd: I'm puzzled here. I thought that the default for `required` was `ContextSetting.REQUIRED`. I implemented widgets under this assumption. Now I see it's actually not. So, for instance, Sieve diagram, which does not mark its attributes as required, can get a context without these attributes? It reuses invalid contexts and does not crash just because it can keep the attributes from before the context was opened? Could you take a look?

We may want to comb through all widgets and check the settings. I would make `REQUIRED` the default, and only use `OPTIONAL` in the few places like colors and shapes in the scatter plot. Not necessarily in this pull request; we can open this as an issue.

astaric: Until this PR is merged, everything was marked as OPTIONAL, but treated as required.
I will amend the last commit and change the default to required.

Issue #2600, opened 2017-09-19
Errors installing add-ons


janezd:
I got this in a version checked out from github (I use Python 3.5) and in the macOs bundle. In the former case, I was installing Education and Bioinformatics, in bundle I tried installing Infrared.

<img width="420" alt="screen shot 2017-09-19 at 14 34 37" src="https://user-images.githubusercontent.com/2387315/30611747-de1d1ee0-9d47-11e7-91fc-0f55850ffcac.png">

In the bundle, `pip3 install --upgrade setuptools` solved the problem by upgrading setuptools from 28.8.0 to 36.5.0.

In the checked out version, even pip would no longer execute, so I had to manually patch it with [this](https://github.com/pypa/setuptools/issues/885#issuecomment-307696027).

I'm not sure how to reproduce this. Some of students also got the same error. 

Can we, for a start, upgrade the setuptools in the bundle?

markotoplak:
I got the same error when I installed the Educational add-on from the canvas into my checked-out version (all the other add-ons were installed with pip install -e .). Pip did not execute. When I deleted all traces of the Educational add-on from my site-packages, pip and other python modules started working again.

Then I checked out that add-on and installed it with "pip install -e ." and it worked.

PR #2602, opened 2017-09-20
[FIX] setup: Increase required setuptools version

ales-erjavec:
##### Issue

Fixes gh-2600

##### Description of changes

Increase required setuptools version to 36.3 

This is required for installing namespace packages from wheels
build with newer versions of setuptools (see setuptools issue 885)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -12,4 +12,4 @@ joblib>=0.9.4
 keyring
 keyrings.alt    # for alternative keyring implementations
 dill
-setuptools>=11.3
+setuptools>=36.3


astaric:
@ales-erjavec Do we also need to update the conda-spec file?
https://github.com/biolab/orange3/blob/master/scripts/windows/specs/conda-spec.txt

ales-erjavec:
It would be better to update [conda-recipe/meta.yaml](https://github.com/biolab/orange3/blob/master/conda-recipe/meta.yaml) as  well as the conda-forge one.

The conda-spec.txt is (an example of) a generated file. The env spec is otherwise rebuild when the installers get created.

astaric:
Ok. I'll update the conda-forge recipe when we do the 3.6 release (probably this friday).

PR #2611, opened 2017-09-22
Feature Constructor: Add some builtin functions

janezd:
##### Issue

Feature Constructor widget does not offer `abs` in the combo.

##### Description of changes

I added `abs` and, while at it, also some other builtin function that could be useful.

##### Includes
- [X] Code changes
- [X] Tests


astaric:
Could you also provide some tests?

janezd:
I was like -- test what? Then I tested whether `abs` is there --- and it wasn't! :)

Fixed and done.

PR #2616, opened 2017-09-24
conda_installer: Include conda in environment

astaric:
##### Issue
Standalone Miniconda installer (offline) does not include conda "executable" and does not use the conda-forge channel.

To reproduce use current miniconda installer (on a clean windows installation) to install orange.
Navigate to %LOCALAPPDATA%/Orange/Scripts
conda.bat is not present

##### Description of changes
Create conda.bat and .condarc files manually

### DIFF
--- a/scripts/windows/condainstall.bat
+++ b/scripts/windows/condainstall.bat
@@ -23,6 +23,18 @@ if not exist "%PREFIX%\python.exe" (
         "%CONDA%" install --yes --copy --quiet --prefix "%PREFIX%" "%CD%\%%f" ^
             || exit /b !ERRORLEVEL!
     )
+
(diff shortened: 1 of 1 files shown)


astaric:
Another thing that is missing from the current installers is the addition of the conda-forge channel.
The easiest way to do this (without breaking any existing environments) is by creating a .condarc file in the root of the environment ("C:\Users\x\AppData\Local\Orange") with the following content:
```
channels: 
  - defaults
  - conda-forge
```

astaric:
@ales-erjavec I think I'm done. Can you take a look?

ales-erjavec:
Seems to work OK with conda 4.1, 4.2 and 4.3.

Issue #2617, opened 2017-09-25
Rank: incorrect ranking


ajdapretnar:
##### Orange version

3.6.dev

##### Expected behavior
Rank sorts by the first score by default.

##### Actual behavior
Rank sorts by #. This causes ugly visualization (not enough space for sorting triangle) and confuses the user.

##### Steps to reproduce the behavior
File (heart-disease) - Rank.

##### Additional info (worksheets, data, screenshots, ...)
<img width="653" alt="screen shot 2017-09-25 at 10 12 06" src="https://user-images.githubusercontent.com/12524972/30798476-209be152-a1da-11e7-8966-1a74f0dce052.png">

Ranking by # should be disabled. I can't think of any good reason why anyone would want to rank by #.

ajdapretnar:
Also, when sorting by an attribute with NA score, NA gets ranked first. It should be ranked last.

<img width="765" alt="screen shot 2017-09-25 at 10 20 35" src="https://user-images.githubusercontent.com/12524972/30798849-46e7258c-a1db-11e7-964e-3b33b630fa11.png">

PR #2618, opened 2017-09-25
OWRank: sort NaNs last; fix sort indicator

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/2617

##### Description of changes
Always sort NaNs last. Implemented by exposing reimplementable `AbstractSortTableModel._argsortData()`. Also fix header indicator by sorting via `view.horizontalHeader().setSortIdicator()` instead of its model.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -549,7 +549,7 @@ class OWRank(OWWidget):
         # Saved selected_rows will likely be incorrect
         if version is None or version < 2:
             column, order = 0, Qt.DescendingOrder
-            headerState = settings.pop("headerState")
+            headerState = settings.pop("headerState", None)
 
             # Lacking knowledge of last problemType, use discrete ranks view's ordering
             if isinstance(headerState, (tuple, list)):
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -147,6 +147,13 @@ class TableModel(PyTableModel):
         if yes_reset:
             super().resetSorting()
 
+    def _argsortData(self, data, order):
+        """Always sort NaNs last"""
+        indices = np.argsort(data, kind='mergesort')
+        if order == Qt.DescendingOrder:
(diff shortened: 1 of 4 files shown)


ajdapretnar:
This look but.

Would you care to check this error, too?

orange/orange3/Orange/widgets/settings.py:380: UserWarning: Could not read defaults for widget <class 'Orange.widgets.data.owrank.OWRank'>
The following error occurred:

'headerState'
  .format(self.widget_class, ex))

ajdapretnar:
Still some comments.
1) When I sort by, say, Info. gain and chance the data set, the sorting doesn't pertain. I have to re-sort again even though there's an arrow indicating the sorted column.
2) Could we turn off sorting by # or is this not possible with the model?

kernc:
> 1.

Fixed.

> Could we turn off sorting by # or is this not possible with the model?

Sure it's possible. But one may wish to split scores by variable type, or to keep discrete variables with the number of values within a certain threshold. My vote is against.

ajdapretnar:
> But one may wish to split scores by variable type, or to keep discrete variables with the number of values within a certain threshold. My vote is against.

Ok. It is just a little ugly when sorting by #. Perhaps add a tad of extra width then?

screen shot 2017-09-26 at 15 29 58

Orange/widgets/data/owrank.py
```diff
@@ -417,13 +424,14 @@ def updateScores(self):
 
         self.ranksModel.wrap(model_array.tolist())
         self.ranksModel.setHorizontalHeaderLabels(('#',) + labels)
-        self.ranksView.setColumnWidth(0, 30)
+        self.ranksView.setColumnWidth(0, 40)
```
kernc: @ajdapretnar Wide enough?

ajdapretnar: Yes, but now it looks annoyingly wide. 😆 I'm such a brat. I'll merge this as it is a great solution! 👌

PR #2624, opened 2017-09-26
OWKMeans: Add a column with silhouette score

janezd:
##### Issue

KMeans computes silhouette scores, but does not output them.

##### Description of changes

Add a column with silhouette scores.

The widget used to computes a silhouette on a sample if the number of instances was larger than 5000. It still behaves like this; if the number of instances is too large, it outputs a column of NaNs.

Silhouette scores are transformed by arctan(x) / pi + 0.5 to put it into [0, 1] (or, usually, [0.5, 1]).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -70,6 +72,10 @@ class Outputs:
     class Error(widget.OWWidget.Error):
         failed = widget.Msg("Clustering failed\nError: {}")
 
+    class Warning(widget.OWWidget.Error):
```
jerneju: `Warning` or `Error`?

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -286,14 +292,13 @@ def selected_row(self):
     def select_row(self):
         self.send_data()
 
-    def _get_var_name(self):
+    def _get_var_name(self, name):
```
jerneju: This method is very similar to a function in `annotated_data.py`.

janezd:
Done.

While at it, I fixed numbering: If `x` already exists, the name for the next `x` should be `x (1)`, not `x (2)`. This may, but likely won't break any saved workflows.

jerneju:
Ok. 

BTW I usually move files using `git mv` in order to preserve editing history.

PR #2625, opened 2017-09-26
Rename Selection Group to Selected

janezd:
##### Issue

If Scatter plot's Data is connected to another widget, the column name in the output signal depends upon whether the user selects more groups or only one. If it is, for instance, connected to another scatter plot, the user has to change the name of the variable used for coloring.

##### Description of changes

Rename *Selection group* to *Selected*.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
What if Scatter plot was also modified to always output create_groups_table data?

janezd:
This would replace No and Yes with Unselected an G1. Scatterplot is usually used with a single selection, so — no.

PR #2627, opened 2017-09-27
[FIX] Scatter Plot: remove dead and commented code, tests

jerneju:
##### Issue
Dead code. Commented code. Low coverage.

##### Description of changes
Some code removed.
Coverage increased from 74% to 83%.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -2,7 +2,7 @@ import numpy as np
 import scipy.sparse as sp
 
 from AnyQt.QtCore import Qt, QTimer
-from AnyQt.QtGui import QPen, QFont, QFontInfo, QPalette
+from AnyQt.QtGui import QPen, QPalette
 from AnyQt.QtWidgets import QApplication
 
 from sklearn.neighbors import NearestNeighbors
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owscatterplot.py
```diff
@@ -414,17 +385,14 @@ def set_shown_attributes(self, attributes):
...
-
     def init_attr_values(self):
         domain = self.data and self.data.domain
         for model in self.models:
             model.set_domain(domain)
         self.attr_x = self.xy_model[0] if self.xy_model else None
         self.attr_y = self.xy_model[1] if len(self.xy_model) >= 2 \
             else self.attr_x
-        self.graph.attr_color = domain and self.data.domain.class_var or None
+        self.graph.attr_color = self.data.domain.class_var if domain else None
```
janezd: The previous expression `domain and self.data.domain.class_var or None` is equivalent to `self.data.domain.class_var if (domain and self.data.domain.class_var) else None`, which is what I actually meant here. This is not the same as `self.data.domain.class_var if domain else None`, which assumes that `self.data.domain.class_var` is either a `Variable` or `None` (which, in fact, it is :) ). I'm just nitpicking; the previous expression was more explicit, but too unreadable.

PR #2629, opened 2017-09-27
allows_cycle flag

payno:
##### Issue
Can't create loop inside a workflow

##### Description of changes
Add the 'allows_cycle' flag into `OWWidget`.

- If `False` won't change the current behavior.
- If `True` then the widget will allow connection bringing cycle in the workflow. This is why if set to True, the widget/user has to know how to handle cycling on his workflow.

##### Includes
- [X] Code changes
- [x] add unit test
- [x] check for 8 cycle
- [x] check for sub cycle

### DIFF
--- a/Orange/canvas/registry/description.py
+++ b/Orange/canvas/registry/description.py
@@ -195,6 +195,7 @@ class WidgetDescription(object):
         Widget's background color (in the canvas GUI).
     replaces : list-of-str, optional
         A list of `id`s this widget replaces (optional).
+    allows_cycle: True if the widget allows cycle creation on her
 
     """
     def __init__(self, name, id, category=None, version=None,
@@ -204,7 +205,8 @@ class WidgetDescription(object):
                  help=None, help_ref=None, url=None, keywords=None,
                  priority=sys.maxsize,
                  icon=None, background=None,
-                 replaces=None):
+                 replaces=None,
+                 allows_cycle=False):
 
         if not qualified_name:
             # TODO: Should also check that the name is real.
@@ -228,6 +230,7 @@ class WidgetDescription(object):
         self.icon = icon
         self.background = background
         self.replaces = replaces
+        self.allows_cycle = allows_cycle
 
     def __str__(self):
         return ("WidgetDescription(name=%(name)r, id=%(id)r), "
--- a/Orange/canvas/scheme/node.py
+++ b/Orange/canvas/scheme/node.py
@@ -70,6 +70,9 @@ class SchemeNode(QObject):
         self.__state_messages = {}
         self.properties = properties or {}
 
+    def allows_cycle(self):
+        return self.description.allows_cycle
+
     def input_channels(self):
         """
         Return a list of input channels (:class:`InputSignal`) for the node.
--- a/Orange/canvas/scheme/scheme.py
+++ b/Orange/canvas/scheme/scheme.py
@@ -345,8 +345,9 @@ class Scheme(QObject):
         """
         check_type(link, SchemeLink)
 
-        if self.creates_cycle(link):
-            raise SchemeCycleError("Cannot create cycles in the scheme")
+        if not (link.source_node.allows_cycle() or link.sink_node.allows_cycle()):
+            if self.creates_cycle(link):
+                raise SchemeCycleError("Cannot create cycles in the scheme")
 
         if not self.compatible_channels(link):
             raise IncompatibleChannelTypeError(
@@ -385,6 +386,8 @@ class Scheme(QObject):
         """
         check_type(link, SchemeLink)
         source_node, sink_node = link.source_node, link.sink_node
+        if source_node.allows_cycle() or sink_node.allows_cycle():
+            return False
         upstream = self.upstream_nodes(source_node)
         upstream.add(source_node)
         return sink_node in upstream
@@ -527,11 +530,12 @@ class Scheme(QObject):
         .. note:: This can depend on the links already in the scheme.
 
         """
-        if source_node is sink_node or \
-                self.is_ancestor(sink_node, source_node):
-            # Cyclic connections are not possible.
-            return []
-
+        if not (source_node.allows_cycle() or sink_node.allows_cycle()):
+            if source_node is sink_node or \
+                    self.is_ancestor(sink_node, source_node):
+                # Cyclic connections are not possible. Except if tagged
+                # 'allows_cycle'
+                return []
         outputs = source_node.output_channels()
         inputs = sink_node.input_channels()
 
--- a/Orange/canvas/scheme/scheme.py
+++ b/Orange/canvas/scheme/scheme.py
@@ -345,9 +345,7 @@ class Scheme(QObject):
         """
         check_type(link, SchemeLink)
 
-        if not (link.source_node.allows_cycle() or link.sink_node.allows_cycle()):
-            if self.creates_cycle(link):
-                raise SchemeCycleError("Cannot create cycles in the scheme")
+        self.check_cycles_will_be_valid(link)
 
         if not self.compatible_channels(link):
             raise IncompatibleChannelTypeError(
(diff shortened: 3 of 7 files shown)


payno:
If this PR is taking into consideration and if you think this need some unit test I can elaborate some if needed.

jerneju:
This PR does not work. And why would you need a loop inside a workflow?

janezd:
Loops are prevented by default since there is nothing to stop the cycling.

This functionality is enabled in a project based on Orange, [Oasys](https://www.elettra.trieste.it/oasys.html), which uses just the canvas while the widgets computes some simulations, and looping makes sense and is controlled.

Orange in this repository will not have it unless you can present an extremely strong case.

kernc:
> there is nothing to stop the cycling

The widget that explicitly sets `allow_cycle=True` prevents cycling by not resending the output.

@payno What is the use case for this?

payno:
First of all thank you for your comments.
I think I will give you some context, it will help.

So first we are currently developing an Orange3 add-on called 'tomwer' (https://gitlab.esrf.fr/payno/tomwer).
The idea is to help scientists to control acquisition on beamlines and prepare the post-processing.

Here is a screen shot of the Orange add-on and a use case of a loop.

tomwer_setup

In some cases user want to:
  - have a processing stream A (it can be an addition of n 'boxes'. In this case 'ftseries reconstruction' box is stream A) 
  - check result of the stream A action (in this case the box called 'data validator'). This is where the 'expertise' is needed to simplify (a lot).
  - If the result is good according to the user he will simply continue to the next box pushing the 'validate button'.
  - If not then this mean that processing stream A have to be executed again with different parameters that user will change manually. So this will be done by pushing an other button (change reconstruction parameters in this case).
  - **The stream A can be complex and associate different boxes**.
  - The action can be repeated as many time as the user feel necessary. 

So this is why at this point we would need cycling.
With this modification (the one we have on our fork) the two widgets cycling have the `allow_cycle` parameter set to True.

The idea was to add a flag to keep the default behavior and let only user/developer set this flag. Maybe having it hidden would be better. Or maybe this is not an ' extremely strong case' for you to consider cycling.

For now we have a fork of your project with only small modifications (at east on the code source) but as you can imagine we would love to revoke it and use directly the main one.

Last point, 'tomwer' is developed by the ESRF (M. Sanchez del Rio is also from the same facility - and we went with him and L. Rebuffi last year).
So we are aware about Oasys but we would like to stay as close as possible from the original Orange project to keep up to date and access other add-on that might interest users.

jerneju:
@payno 

screenshot_20171002_141939

kernc:
@jerneju rerun Orange with `--force-discovery`.

jerneju:
This does not work.

janezd:
Could you be more specific here? :)

jerneju:
@janezd: It is not backwards compatible.

jerneju:
I had some problems. Now it does work, sorry for inconvenience.

janezd:
The ultimate authority (BDFL?) for canvas is @ales-erjavec. Aleš, I believe there are legitimate uses for cycles in Orange. Not with our widgets, but in, for instance, the context of Oasys where a simulation is run for a fixed number of times, or until convergence. What is your position on allowing cycles?

I see a problem in the current implementation. https://github.com/biolab/orange3/pull/2629/files#diff-e90504d40fd28531e83fd6933d15c6d1R533 checks that either source or sink must allow cycles. This enforces the order in which the widgets are put onto the canvas. I think that it would suffice if the cycle that is being created contains at least one "cycle breaker" - which doesn't have to be one of the two widgets that actually close it.

`allow_cycles` doesn't sound like a property of the widget. It's more like `cycle_breaker` (but I don't like this name either). Can we find a better name?

payno:
Thanks for the review.
the `cycle breaker` (`cycle controller`?) sounds to be a better option than the one proposed. More flexible from my point of view.

The only point I see is that as the paradigm of the scheme will change (allowing n cycle in the stream ) some code would have to be added to tag/register or detect each cycle. Something that is not done for now.
So inside each cycle check for a `cycle breaker`. Which might be done a million of different ways, some probably you would like to avoid.

If you are in for this I can propose modifications to have at least a first draw of it.
But you might have to tell me the best way from your point of view to do it. (checking the full stream each time or not ...)
This way it might be less time consuming for you.
Also as it is quiet 'core' functionalities I am not sure you would prefer a non core developer do it.

ales-erjavec:
> What is your position on allowing cycles?

Undecided.

* They generate a lot of ambiguity in the workflow, at least without further semantic and structural restrictions.
* They would be useful.

Some prior art:
* LabView has explicit structured looping constructs.
* YAWL has no restrictions on cycles, but has a much richer language to express execution semantics.


> I think that it would suffice if the cycle that is being created contains at least one "cycle breaker" - which doesn't have to be one of the two widgets that actually close it.

A single connection can create multiple cycles in a strongly connected component. It is not clear that 'cycle breaker' could actually break cycles in the general case. Should the cycles be limited to simple cycles?

@payno Note that the signal propagation (SignalManager) also need to change. It is not currently prepared to deal with cyclic dependencies (can deadlock). If you are using this please also consider applying https://github.com/ales-erjavec/orange-canvas/commit/b23bed8555c7a04133db6485aa60bcb4604b8ef6

janezd:
> They generate a lot of ambiguity in the workflow, at least without further semantic and structural restrictions.

Oh, oh, yes. Signal propagation order suddenly becomes important.

You made me think about visual ambiguities. Workflow would be confusing if it is not immediately clear which widget is the breaker. Maybe the canvas should present breaker widgets differently? For instance with two half circles, like `(| |)`. And the signals would be connected in the opposite direction to avoid crossing the line in the schema - in which case the graphical presentation would need to contain an arrow within the widget icon? We need to think about visual aspects.

@payno, there's a bit of a problem here: there's a bunch of things to think about. There is no clear promotor in the core group, so nobody will be pushing for it. Luckily there is no clear opposition either, so if you keep pushing and solving problems, you may have a good chance -- if this problems can be solved. So what about the problem @ales-erjavec exposed: if you have a workflow in a shape of 8 - with two breakers, one at the top and one at the bottom, and the two loops of course contain a different number of widgets :), what is the order of signal processing? To make things worse for you, the rule has to be kind of obvious, say directly, visually from the workflow. Plus, the same rule has to work for any weird graph with loops.

payno:
I see tricky problem...

first @ales-erjavec thanks for the link.

@ales-erjavec : is avoiding cases A and B equivalent to `simple cycles` idea ?
   - 8 shape
8shape
   - sub cycles
subcycles

Limiting to `simple cycles` is a good idea to start I think.

An other way would be to go for defining clearly the signal propagation as suggested.
For example getting the execution order defined by the user when creating the connections.
And allowing the user to change this order (index tag on the link between boxes - 1, 2 ... )

indexes1

A second option would be for widgets enabling cycling that all inputs and output would be tagged with indexes following connection order.
indexes2


Internally this order would be managed by the SignalManager. If I remember correctly this might not be a huge modification of the SignalManager.
This added with an 'infinite' loop check might work to deal with the cases A and B. At least user would have a clear view of the order of execution. And with this information it would be up to them to deal with it. Which sound fair. But this is my point of view.
My concern here would be about the amount of modifications to bring to the existing code. Ordering might be a huge modification. What do you think ?

payno:
I hope to find some time to work on this issue.
But I would like to avoid the proposal of a solution going in the wrong direction.

So for you @ales-erjavec  and @janezd  should I rather propose a solution:
- only allowing `simple cycles` (with the cases explained previously)
- allowing all cycles using an execution ordering based on link between nodes. This order will be initially the same as the order of creation of those links.
  But will also be editable by the user. But this modification might bring large modifications of the `SignalManager`

?

janezd:
I don't really understand what are 'simple cycles'. Moreover, I think we haven't yet arrived to a clean solution - there was a proposal, we identified a problem there, and a fix is to limit it to a certain types of cycles or to essentially ignore the problem by following the order in which the connections are added, although this is not indicated visually, and letting the user edit it somehow...

I wouldn't be happy with adding a hack that would satisfy the current needs, but would cause headaches down the road. We must find an obvious, clean solution to include loops. We may be thinking about the problem in the wrong way.

This is just my opinion, but I'm not the one to make decisions here. @ales-erjavec is.

ales-erjavec:
>  but I'm not the one to make decisions here. @ales-erjavec is.

ales-erjavec:
I don't know. The more I think of this the less enthusiastic I am. I think that unless the full YAWL like CPN backed (or equivalently powerful) modeling is implemented then this is doomed.

janezd:
> but  I'm not the one to make decisions here. @ales-erjavec is.

What I meant to say is that I (as many others) know a lot about widgets and can make decisions about them, but you are the only one who knows the canvas. I can express my opinion whether this would be desirable, but only you can say whether it's feasible (without turning everything around), both in terms of code and of philosophy. If you think it's not, it's not.

payno:
> I don't really understand what are 'simple cycles'. 

this is allowing cycle connection except when we found cases of 8 shape or sub cycle from my point of view.

> essentially ignore the problem by following the order in which the connections are added, although this is not indicated visually, and letting the user edit it somehow...

This would be indicated visually by labeling the links between nodes. (and the `label` would be editable)

> I think that unless the full YAWL like CPN backed (or equivalently powerful) modeling is implemented then this is doomed.

Is this true for the both cases or in the case of the 'simple cycles' we cam avoid that ?

ales-erjavec:
> Is this true for the both cases or in the case of the 'simple cycles' we cam avoid that ?

If you restrict to single entry and exit and no other incoming edges to the loop and limit the user interaction to single focus widget in the loop (i.e no widget on the loop can 'spontaneously' update the outputs out of turn), then yes (probably).

payno:
Sorry for this PR a little bit 'complicated' to understand I can imagine.

The last commits include the management of a 'simple cycle' as discussed before.

I added some unit test in test_scheme. But it looks like those tests are not runned anymore.
So I am not sure this is the correct place to add them ? 
But the `TestSchemeCycleDetection` class should pass.

The cycle detection is quiet 'raw' for now but I guess orange canvas do not intend to manage thousand of node.
This is using basic graph theory.
   - we remove all nodes not part of a cycle (`clean_nodes_without_successor`)
   - then we deduce all existing cycles and making sure of their uniqueness (using set structure, etc ...)
Then we check for two information on cycles to validate the construction:
   - making sure they have at least on widget with the 'allows_cycle' attributes at True
   - making sure that cycles are not 'connected' by making sure cycle are not including shared `SchemeNode`

Finally some unit test have been added to check:
   - 8 cycle is refused
   - sub cycles are refused
   - simple distinct cycles are accepted
   - cycle with no controller are refused

payno:
@ales-erjavec and @janezd  is this implementation for 'simple' cycle adequate for you ?

markotoplak:
@payno, I am sorry this is stuck here for so long... I have seen that Tomwer users need to install your branch of Orange. Is it because of the cycles or are there any other features that you needed?

payno:
Yes, the main reason came from our need for creating cycles.
We also want to skip the signal compression between "Orange Widgets" and we created a widget variable for this too. I can propose a PR if you want.
Finally we have some 'quick and dirty' modifications concerning the logs. We want to save times and keep a trace of the last 10 Orange executions.

markotoplak:
@payno, thank you. I do not understand what do you mean with "signal compression". Could you explain?

payno:
Sure, this refer to the 'SignalManager' class and to the call of the compress_signals function that I don't want to process in our workflows because I want to make sure all signal is treated (even if triggered two times almost at the same time).
So for some of my 'Orange widget' I require to skip this process.

markotoplak:
@payno, thank you for this contribution. We were discussing cycles this week and decided that both this proposal and cycles as implemented in Oasis would introduce potential problems later on. 

Perhaps it would be best if we try find a solution that would satisfy your use-case but would avoid cycles.

Issue #2279, opened 2017-05-05
Basket format from Orange v2 not fully supported in Orange v3


BlazZupan:
##### Orange version
3.4

##### Expected behavior
Reads old .basket files. When duplicate terms in a single row are encountered, the term counts should add up. In the data below the `surprise`  for second data instance should have a value of 6, and `and` for the third instance a value of 2.

##### Actual behavior
Fails on duplicate terms and emits ValueError.

##### Steps to reproduce the behavior
Store the following data into a tmp.basket file and read with `Orange.data.Table("tmp.basket")`.

```
nobody, expects, the, Spanish, Inquisition=5
our, chief, weapon, is, surprise=3, surprise=2, and, fear,fear, and, surprise
our, two, weapons, are, fear, and, surprise, and, ruthless, efficiency
to, the, Pope, and, nice, red, uniforms, oh damn
```

PR #2632, opened 2017-09-27
Basket term summation

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/2279

##### Description of changes
Constructs COO instead of CSR for minimal performance impact.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
Can you benchmark this on a very large file and compare it to the previous version?

Previous version compiled to almost clean C except for storing value names (difficult to avoid), raising exceptions (doesn't matter) and some bounds (could be disabled). Now it converts C doubles to Python floats and appends them to a Python list.

Going to coo is a good idea (I didn't know that it handles duplicates in this way!), but you could still use pre-scan to determine the upper bounds for array sizes and use arrays instead of lists. The code is already there.

Don't throw away perfectly cool finite automata! :)

kernc:
260 ms before vs. 290 ms now on 4.3 MB ((62560, 126), 4%) foodmart.basket.

janezd:
+11 %. Acceptable.
(It's interesting why it isn't much slower than arrays: check `__Pyx_PyList_Append`, it's a nice shortcut for `list.append`. The caps lock functions it calls are actually fast macros.)

Still, if it isn't too much of a hassle, you could keep the arrays -- or not?

Because guys, I see where this is going. You're secretly plotting to eliminate all my beautiful finite automata. One by one. You always pretend the PR is about fixing some minor inconsequential bug and, bang, one automaton less. All my beautiful automata I so carefully put together. One by one. Shame on you!

ajdapretnar:
This does support sparse for Associate add-on. However, it doesn't seem to solve #2279. I still get ValueError.

kernc:
@ajdapretnar It supplies new C source files which need to be compiled. If you have a compiler, after checkout, pip install.

> Still, if it isn't too much of a hassle, you could keep the arrays -- or not?

@janezd If I go with arrays, to what size do I preallocate them?

ajdapretnar:
> @ajdapretnar It supplies new C source files which need to be compiled. If you have a compiler, after checkout, pip install.

Oooo. 😲 Such beauty!

janezd:
@kernc, can't you do it like it was before: rough count and decreasing the size later if needed? I wouldn't bother for a minimal speed gain, but since the code is (was) already there...

kernc:
Rough count, what does that mean? Before it was a CSR with number of rows / max columns known. Now it's a COO with similar items stacking together. Sorry, but other than `n_lines*max_columns` (i.e. dense), I don't see it. But happy to have you do it. 😃

janezd:
(At least) one of us is overlooking something. Isn't `len(data)` the same for csr and coo?

https://github.com/biolab/orange3/pull/2632/files#diff-677ca243c794fab1cb043a76541c7028L82 computes the number of elements for X, Y and meta. For coo you just no longer need `n_lines`. What am I missing?

kernc:
Oh, _that_'s what that is. 👍 

Despite my best efforts, this is now an additional 40 ms slower. I left it in a standalone commit, hoping perhaps you might not as approve of any slower and certainly more complex, error-prone code.

kernc:
Using CPP std::vector for X and Y, no change compared to simple Python lists. The former is consistently a bit faster.

janezd:
Codecov/patch is apparently stuck. Can you push again?

janezd:
We have problems with CI today. The same AppVeyor error (in another widget) happened to me in #2669.

astaric:
@janezd could you copy the errors to [Random-coverage](https://github.com/biolab/orange3/wiki/Random-coverage) on wiki? After you restart the build, the errors are lost. I have a feeling there are only a couple of tests that need to be skipped/rewritten and our "CI problems" will go away.

PR #2637, opened 2017-09-28
Allow the widget height to be smaller than control area when the control area is hidden

janezd:
##### Issue

The widget height can't be smaller than control area even when the control area is hidden.

##### Description of changes

- Override splitter's `sizeHint` and `minimumSize` to ignore the control area when hidden.
- Call `updateGeometry` on splitter when hiding/showing the control area.

screen shot 2017-09-28 at 21 50 05

screen shot 2017-09-28 at 21 49 57

##### Includes
- [X] Code changes
- I see no reasonable non-trivial way to test this.

### DIFF
--- a/Orange/widgets/widget.py
+++ b/Orange/widgets/widget.py
@@ -4,6 +4,7 @@
 import sys
 import os
 import types
+from operator import attrgetter
 
 from AnyQt.QtWidgets import (
     QWidget, QDialog, QVBoxLayout, QSizePolicy, QApplication, QStyle,
@@ -221,6 +222,12 @@ class OWWidget(QDialog, OWComponent, Report, ProgressBarMixin,
         if self.controlArea is not None:
             # Otherwise, the first control has focus
             self.controlArea.setFocus(Qt.ActiveWindowFocusReason)
+
+        if self.splitter is not None:
+            sc = QShortcut(
+                QKeySequence(Qt.ControlModifier | Qt.ShiftModifier | Qt.Key_D),
+                self)
+            sc.activated.connect(self.splitter.flip)
         return self
 
     # pylint: disable=super-init-not-called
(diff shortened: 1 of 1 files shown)


janezd:
- I fixed the bug for widgets without control area.
- I added the space key is a shortcut for hiding and showing the control area [does not work properly yet!].
- Signals are no longer needed, but I kept and improved them to allow widgets to adapt to showing/hiding the control area by, say, putting some controls on the graph when control area is hidden.

janezd:
> I added the space key as a shortcut for hiding and showing the control area [does not work properly yet!].

Apparently it does. The shortcut was - weirdly - processed by a wrong widget (e.g. box plot, when the focus was on scatter plot). I dont' get it why it happened, and it no longer happens.

kernc:
> I added the space key as a shortcut for

-2 👎 Spacebar is a common a11y key for activating the current focused element.

janezd:
@kernc, you're right. Space bar is really practical, though, because it's a bar and so it's easy to hit. I'd love to have this when lecturing, so I can just set up the widget and hit space. (*If you need more space, hit a space.* :) ) Would Alt-Space be OK? All shortcuts around Space are dangerous, though; this one may trigger changing the keyboard language in some systems, I guess.

We generally avoid (or even never use?) default buttons in widgets. Activating other controls with spacebar will cause no harm. What about going with *practicality beats purity*?

ales-erjavec:
**Most definitely** not space. This is not even up for discussion, it completely breaks keyboard interaction and control in subcomponents.
Canvas already uses <kbd>control + shift + D</kbd> for similar purpose. The D stands for tool **d**ock, which does not quite apply here, but might be worth considering for consistency.

janezd:
Huh, I understood that when a subcomponent handles an event, the event doesn't propagate to the containing widget, so everything works as before. At least in what I tested nothing seemed broken (e.g. line edits still worked).

But you're an expert in this. @kernc's objection is valid, too. I changed the shortcut to <kbd>control + shift + D</kbd>.

ales-erjavec:
No. Shortcuts are matched in the input buffer before any key press events are dispatched. When a shortcut matches, the current focused widget gets delivered an QEvent.ShortcutOverride event. If the widget accepts this event the shortcut is suppressed and the regular key press events are delivered to the widget, otherwise the shortcut triggers and eats the input. All full text input widgets do handle the ShortcutOverride events, but buttons, combo box, ... do not.

janezd:
How do you know all this stuff?!

If you think it's OK now, you can merge. I didn't plan to add any tests.

PR #2639, opened 2017-09-29
TableModel: Add old style mappings as deprecated

nikicc:
##### Issue

The methods for mapping from/to source table rows in `TableModel` were renamed from _mapFrom**Table**Rows_ to _mapFrom**Source**Rows_ as of https://github.com/biolab/orange3/pull/2494. Hence mapping for all models extending `TableModel` in addons now crash with:
```
AttributeError: '*TableModel' object has no attribute 'mapToTableRows'
```

##### Description of changes
Add old style methods names and mark them as deprecated.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


nikicc:
@astaric I also added a re-implemented `columnSortKeyData` and added a deprecation warning. I think all should be fine now :)

PR #2641, opened 2017-09-29
Select Columns: move filter to a separate file

jerneju:
##### Issue
Duplicated code.

##### Description of changes
Code which is also used by some other widgets (Linear Projection, Radviz) is moved from Select Columns to a separate file.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


jerneju:
Those commits cannot be squashed. So many commits are needed in order to preserve code history.

Orange/widgets/data/owselectcolumns.py
```diff
@@ -263,53 +100,24 @@ class Outputs:
 
     def __init__(self):
         super().__init__()
+        VariablesFilter().__init__()
```
janezd: Don't call a constructor of a non-base class!

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,244 @@
...
+                current = completer.currentRow()
+                current += diff
+                completer.setCurrentRow(current % completer.completionCount())
+                completer.complete()
+            return True
+        else:
+            return False
+
+
+class VariablesFilter(QObject):
```
janezd: I like the idea of this PR. Implementation is messy, though. It is a `QObject` (why?) that constructs a line edit, a view and a model, and stores all of them as attributes that are later extracted. What does this object represent?

If you take a closer look -- one uses it by constructing it, taking the three objects, and then discarding the object. This is then not a class but rather a function that returns two objects, line edit and a view. Model is accessible through a view or, even better, given as an argument to the function, which lets the widget construct a model it needs (`DomainModel`, `VariableListModel`...). The two update methods become enclosed functions, and the `add_*` methods are removed. All attributes (that are needed) become just local variables stored in closures.

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,244 @@
...
+    def __init__(self, variable_list_model=VariableListModel(enable_dnd=True)):
```
janezd: You should never use mutable values as default arguments (without having a very good reason for doing so, and being very careful).

What happens here is that *all* instances of `VariablesFilter` are gonna share the same instance of `VariableListModel`. Open canvas, connect two `SelectColumnsWidget` to `File` and have it visible at the same time. Move attributes around in one of them and observe how the other changes, too!

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,244 @@
...
+            items = [prefix + " " + item
+                     for item in self.original_completer_items]
+        else:
+            items = self.original_completer_items
+        old = list(map(str, self.completer_model.stringList()))
+
+        if set(old) != set(items):
+            self.completer_model.setStringList(items)
+
+    def add_filter(self):
```
janezd: Why is a method that *returns* a filter called `add_filter`? And a method that returns a model is called `add_model`? Qt calls such getters simply `filter`, `model`...

These methods won't be needed anyway if the class is changed into a function.

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,244 @@
...
+        return self.filter_edit
+
+    def add_model(self):
+        return self.available_attrs_proxy
+
+    def add_available_attrs(self):
```
janezd: `add_available_attrs` returns a view and a model?

Orange/widgets/data/owselectcolumns.py
```diff
@@ -263,53 +100,24 @@ class Outputs:
...
-        self.available_attrs_view.setModel(self.available_attrs_proxy)
-
-        aa = self.available_attrs
-        aa.dataChanged.connect(self.update_completer_model)
-        aa.rowsInserted.connect(self.update_completer_model)
-        aa.rowsRemoved.connect(self.update_completer_model)
-
-        self.available_attrs_view.selectionModel().selectionChanged.connect(
-            partial(self.update_interface_state, self.available_attrs_view))
-        self.available_attrs_view.dragDropActionDidComplete.connect(dropcompleted)
```
janezd: This line is removed from the widget, but not replaced in the helper function. Dragging from available attributes to the right does not trigger commit.

Orange/widgets/data/owselectcolumns.py
jerneju: Well, thanks, forgot to remove it. Firstly I wanted to implement it in a different way.

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,236 @@
...
+    completer_navigator = CompleterNavigator(parent)
+    filter_edit.installEventFilter(completer_navigator)
+
+    proxy = VariableFilterProxyModel()
+    proxy.setSourceModel(model)
+    view = VariablesListItemView(
+        acceptedType=Orange.data.Variable)
+    view.setModel(proxy)
+
+    aa = model
```
janezd: Why `aa`? Now that the `model` is not buried within some object, just use `model` instead of `aa`.

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,236 @@
...
+        QCompleter.CaseSensitivelySortedModel)
+
+    filter_edit.setCompleter(completer)
+    completer_navigator = CompleterNavigator(parent)
+    filter_edit.installEventFilter(completer_navigator)
+
+    proxy = VariableFilterProxyModel()
+    proxy.setSourceModel(model)
+    view = VariablesListItemView(
+        acceptedType=Orange.data.Variable)
```
janezd: No need for line break here (any more).

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,236 @@
...
+        old = list(map(str, completer_model.stringList()))
+
+        if set(old) != set(items):
+            completer_model.setStringList(items)
+
+    original_completer_items = []
+
+    filter_edit = QLineEdit()
+    filter_edit.setToolTip("Filter the list of available variables.")
+    if hasattr(filter_edit, "setPlaceholderText"):
```
janezd: This `if` can be removed. I suppose we don't support Qt < 4.8, right (@astaric, @ales-erjavec )?

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,236 @@
...
+                current = completer.currentRow()
+                current += diff
+                completer.setCurrentRow(current % completer.completionCount())
+                completer.complete()
+            return True
+        else:
+            return False
+
+
+def variables_filter(parent=None, model=None):
```
janezd: Maybe a docstring here?

janezd: `model` is no longer optional. Reorder the arguments (put `model` first) and remove the default.

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,236 @@
...
+    def update_completer_model():
+        """ This gets called when the model for available attributes changes
+        through either drag/drop or the left/right button actions.
+
+        """
+        nonlocal original_completer_items
+        vars = list(model)
```
janezd: Is this needed? Iteration over `model` should work anyway.

Orange/widgets/utils/listfilter.py
```diff
@@ -0,0 +1,236 @@
...
+    filter_edit.setToolTip("Filter the list of available variables.")
+    if hasattr(filter_edit, "setPlaceholderText"):
+        filter_edit.setPlaceholderText("Filter")
+
+    completer = QCompleter()
+    completer.setCompletionMode(QCompleter.InlineCompletion)
+    completer_model = QStringListModel()
+    completer.setModel(completer_model)
+    completer.setModelSorting(
+        QCompleter.CaseSensitivelySortedModel)
```
janezd: Model can be given as an argument to `QCompleter`. Check whether `completionMode` and `modelSorting` can also be given as keyword arguments.

janezd:
Rebase to master.

Orange/widgets/utils/listfilter.py
jerneju: No, they cannot be given.

kernc: Can you confirm?

https://github.com/biolab/orange3-geo/blob/28396b83bf713b3bbe7b84ef94b3c8decf5d1bb8/orangecontrib/geo/widgets/owgeocoding.py#L175-L177

Orange/widgets/utils/listfilter.py
janezd: I checked: Orange requires Qt 4.8 since `setPlaceholderText` is used in multiple places. You can remove the `if` statement and just always set the placeholder.

Orange/widgets/utils/listfilter.py
janezd: It accepts the keyword alright, but, interestingly, complains about the type:

```
completer = QCompleter(completer_model,
                       modelSorting=QCompleter.CaseSensitivelySortedModel)
```

gives `unable to convert a Python 'ModelSorting' object to a C++ 'ModelSorting' instance`.

Orange/widgets/utils/listfilter.py
janezd: I added a commit with this, so we can merge then. :)

PR #2647, opened 2017-10-03
Outlier multi-class minor fix.

mstrazar:
##### Issue
Multi-class check fails when data.Y is of dimension (n, 1) or (n, 0).

##### Description of changes
Small fix of Outliers widget checking for multi-class data.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owoutliers.py
+++ b/Orange/widgets/data/owoutliers.py
@@ -170,7 +170,7 @@ class OWOutliers(widget.OWWidget):
         inliers = outliers = None
         self.n_inliers = self.n_outliers = None
         if self.data is not None and len(self.data) > 0:
-            if self.data.Y.ndim > 1:
+            if self.data.Y.ndim > 1 and self.data.Y.shape[1] > 1:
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owoutliers.py
```diff
@@ -170,7 +170,7 @@ def commit(self):
         inliers = outliers = None
         self.n_inliers = self.n_outliers = None
         if self.data is not None and len(self.data) > 0:
-            if self.data.Y.ndim > 1:
+            if self.data.Y.ndim > 1 and self.data.Y.shape[1] > 1:
```
lanzagar: a) Can we use: `if len(self.data.domain.class_vars) > 1`
b) As discussed, we should find out how you managed to get a Y with shape (n,0)

jerneju: Try this: *FIle* (open Iris) -> *Transpose* -> See in *Python Script*.

PR #2649, opened 2017-10-04
[FIX] Scatter Plot: Scatter Plot automatically sends selection

### LINKED ISSUE #2646: Scatter Plot: always automatically sends selection
##### Orange version
since 3.3.7 (or maybe since one of previous versions)

##### Expected behavior
Scatter Plot automatically sends selection only when the checkbox **Send automatically** is checked.

##### Actual behavior
It always sends selection.

screenshot_20171003_110437

jerneju:
##### Issue
Fixes #2646 

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -425,7 +425,7 @@ class OWScatterPlot(OWWidget):
         self.graph.update_data(self.attr_x, self.attr_y, reset_view)
 
     def selection_changed(self):
-        self.send_data()
+        self.commit()
 
     def send_data(self):
         selected = None
--- a/Orange/widgets/visualize/tests/test_owscatterplot.py
+++ b/Orange/widgets/visualize/tests/test_owscatterplot.py
@@ -339,6 +339,21 @@ class TestOWScatterPlot(WidgetTest, WidgetOutputsTestMixin):
         states = [state for state in vizrank.iterate_states(None)]
         self.assertIsNotNone(vizrank.compute_score(states[0]))
 
+    def test_auto_send_selection(self):
+        """
+        Scatter Plot automatically sends selection only when the checkbox Send automatically
+        is checked.
+        GH-2649
+        GH-2646
+        """
+        data = Table("iris")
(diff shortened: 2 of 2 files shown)


Orange/widgets/visualize/owscatterplot.py
```diff
@@ -425,7 +425,8 @@ def update_graph(self, reset_view=True, **_):
         self.graph.update_data(self.attr_x, self.attr_y, reset_view)
 
     def selection_changed(self):
-        self.send_data()
+        if self.auto_send_selection:
+            self.send_data()
```
janezd: This half-reimplements the logic of `Orange.widgets.gui.auto_commit`. This does not work:

- disable auto commit,
- select some points,
- enable auto commit.

Selected points are not sent.

The `auto_commit` function has a local variable `dirty`, which you would have to sent, but can't access from here.

A solution that would work would be to simply call `self.commit` instead of directly calling `self.send_data`. A (slight) problem with this is that it would re-send features when the selection is changed. This could be prevented by storing the last sent features and not sending them when not needed. Not most elegant solution, though.

Orange/widgets/visualize/tests/test_owscatterplot.py
```diff
@@ -339,6 +339,24 @@ def test_vizrank(self):
...
+            self.widget.controls.auto_send_selection.setChecked(is_checked)
+            self.assertEqual(is_checked, self.widget.controls.auto_send_selection.isChecked())
+            self._select_data()
+
+        data = Table("iris")
+        self.send_signal(self.widget.Inputs.data, data)
+        check_select(False)
+        self.assertIsNone(self.get_output(self.widget.Outputs.selected_data))
+        check_select(True)
+        self.assertIsInstance(self.get_output(self.widget.Outputs.selected_data), Table)
```
janezd: Add the test for the scenario I described above.

Orange/widgets/visualize/owscatterplot.py
jerneju: Fixed.

PR #2651, opened 2017-10-04
OWFile: make loading large datasets faster

kernc:
##### Issue
Loading single-cell/merged-small.pickle with shape (400, 50k) was considered too slow.

##### Description of changes
Vectorize float-is_integer-checking calls, and make them apply only when relevant.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/utils/domaineditor.py
```diff
@@ -265,8 +265,11 @@ def get_domain(self, domain, data):
             col_data = self._get_column(data, orig_var, orig_plc)
             is_sparse = sp.issparse(col_data)
 
-            cont_ints = type(orig_var) == ContinuousVariable and \
-                        all(x.is_integer() for x in self._iter_vals(col_data) if not np.isnan(x))
+            def cont_ints():
+                if type(orig_var) == ContinuousVariable:
+                    data = col_data.data[np.isnan(col_data.data)]
+                    return (data == data.astype(np.int16)).all()
```
jerneju: That returns the same result?

jerneju:
```python
data = np.array([0, 1, 2, np.nan, 3])
new_data = data[np.isnan(data)]
(new_data == new_data.astype(np.int16)).all()
False
all([x.is_integer() for x in data if not np.isnan(x)])
True
```

kernc: Missing negation. Thanks.

Orange/widgets/utils/domaineditor.py
```diff
@@ -265,8 +265,11 @@ def get_domain(self, domain, data):
             col_data = self._get_column(data, orig_var, orig_plc)
             is_sparse = sp.issparse(col_data)
 
-            cont_ints = type(orig_var) == ContinuousVariable and \
-                        all(x.is_integer() for x in self._iter_vals(col_data) if not np.isnan(x))
+            def cont_ints():
+                if type(orig_var) == ContinuousVariable:
+                    data = col_data.data[np.isfinite(col_data.data)]
+                    return (data == data.astype(np.int16)).all()
```
janezd: Maybe `np.int64`? As I understand, these number-like strings can be arbitrarily large, so 64-bit ints are needed to cover anything that can be stored in `float64`. (If the file contains a number-like string that is larger than that, it is already lost before arriving here.)

Orange/widgets/utils/domaineditor.py
```diff
@@ -288,7 +291,7 @@ def get_domain(self, domain, data):
                     col_data = [orig_var.repr_val(x) if not np.isnan(x) else ""
                                 for x in self._iter_vals(col_data)]
                 elif type(orig_var) == ContinuousVariable:
-                    col_data = [str(int(x)) if cont_ints else orig_var.repr_val(x)
+                    col_data = [str(int(x)) if cont_ints() else orig_var.repr_val(x)
```
lanzagar: Why did you change cont_ints to a function? It is now executed many times here...

PR #2654, opened 2017-10-05
[ENH] Boxplot: Allow hiding labels

### LINKED ISSUE #2652: boxplot: Hide labels
Boxplot should include an option that can be used to not show the label.

Reasoning:
screen shot 2017-10-05 at 00 05 09

janezd:
##### Issue

Fixes #2652.

##### Description of changes

Add a checkbox.

I would like to discuss the change (perhaps live) before merging. We will have many controls that we'll want to add to the widgets and for each we need to think whether it can be avoided, lest we end up where we were in Orange 2.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -139,6 +139,7 @@ class OWBoxPlot(widget.OWWidget):
     stattest = Setting(0)
     sig_threshold = Setting(0.05)
     stretched = Setting(True)
+    show_labels = Setting(True)
     auto_commit = Setting(True)
 
     _sorting_criteria_attrs = {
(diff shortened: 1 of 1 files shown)


ajdapretnar:
I have been thinking about this for a long time. One option that comes to mind is eliding the labels and placing them diagonally. The other option is a legend, but again probably with eliding (tooltips can be helpful here).

To discuss.

astaric:
Let's add back tabs! :)

(or a little gear icon at the bottom of a widget which opens "advanced settings")

janezd:
It's OK here, since the checkbox does not increase the height (in fact, numeric variables already had such an option, it's called "Annotate" there). But we'll probably eventually need a general approach for this.

Eliding mostly won't help, and diagonal positioning is excelishly ugly (imho).

PR #2655, opened 2017-10-06
Update chardet requirement

mat-cor:
chardet>=3.0.2 is required by requests library


kernc:
`chardet>=2.3.0` is not incompatible with `>=3.0.2`?

astaric:
If you do
```
pip install chardet==2.3.*
```
followed by
```
python setup.py develop
```
you get an error:
```
error: chardet 2.3.0 is installed but chardet<3.1.0,>=3.0.2 is required by {'requests'}
```

PR #2657, opened 2017-10-06
[RFC] Merge data context handler

astaric:
##### Issue
MergeData widget does not remember setting/forgets when inputs change.

##### Description of changes
Implement a context handler that is aware that the data comes from two data source.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owmergedata.py
+++ b/Orange/widgets/data/owmergedata.py
@@ -10,6 +10,7 @@ import Orange
 from Orange.data import StringVariable, ContinuousVariable
 from Orange.data.util import hstack
 from Orange.widgets import widget, gui, settings
+from Orange.widgets.settings import DomainContextHandler, ContextHandler
 from Orange.widgets.utils import itemmodels
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.widget import Input, Output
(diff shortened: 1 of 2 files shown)


janezd:
This context handler is semantically a subclass of `ContextHandler`, but derived from `DomainContextHandler` to be able to (ab)use its non-static methods for encoding variables. This is a cheap solution; deriving the class from `ContextHandler` would require rewriting lots of code already in `DomainContextHandler`.

Its overriding of several methods to change `attributes` and `metas` from dicts of variables to dicts of datasets (with dicts of variables as values) may cause maintainability problems down the road. The class currently has to override `open_context`, `_var_exists` and `decode_settings`. Other inherited methods miraculously work as they are because they happen to not touch `attributes` and `metas` in a wrong way. If somebody changes some method in `DomainContextHandler`, e.g. `new_context` to peek into `attributes` and `metas`, (s)he will also have to add a hacked version of this method to `MultiDomainContextHandler`. This is not OK: `DomainContextHandler` rightfully expects that `attributes` and `metas` are lists, and derived classes should not mess with this in such a way that legitimate changes would require fixing derived classes.

Well, that's my humble opinion, anyway.

Possible solutions:

1. Merge this and keep fingers crossed

2. Derive `MultiDomainContextHandler` from `ContextHandler`. If `DomainContextHandler` contains some code that can be used by other classes as well, this code should be moved out of `DomainContextHandler`.

3. Is `MultiDomainContextHandler` meant to be general or specific for this widget? If the former, move it to `Orange.widget.settings` and goto solution 1 or 2. If the latter, it can be renamed to `MergeDataContextHandler` and perhaps specialized for this widget, if this simplifies it -- that is, if it no longer requires larger chunks of code from `DomainContextHandler`. Perhaps.

astaric:
I am not sure if we need this at all, as we could always implement better heuristics for merge data widget which could solve the problem. The saving to workflow should work already.

janezd:
Saving to workflow should work, but in a schema it doesn't, does it? Heuristic would only improve the initial configuration.

astaric:
Can be reopened when needed

PR #2659, opened 2017-10-08
[FIX] OWSql does not save selected table/query

ceprio:
##### Issue
Added functionality to the widget:

##### Description of changes
- The widget now saves the settings of the selected table name and SQL Query 
- Restoring of table name and SQL query is also now working when opening a workflow
- Connection to the database is done earlier to allow the tables list to be fetched
- Table name is saved and restored as text (or name of table) instead of Id (as previously)
- Empty username and password are now saved as '' (empty string) instead of None. This allows the use of SQL Server's Windows authentication (use of the current user credentials if both fields are empty).
- Placing 'Custom SQL' in second in the table list in order to see it faster when the list of tables is long.
- Execute SQL query only if minimum number of chars entered
Note also:
- A bug was found within SQL Server 2012 (and earlier versions it seem) that had to be dealt with in the backend mssql.py. I added the fix in this pull request so the widget can be tested correctly even on this SQL server version. More details in the source code.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/sql/backend/mssql.py
+++ b/Orange/data/sql/backend/mssql.py
@@ -117,17 +117,26 @@ class PymssqlBackend(Backend):
     EST_ROWS_RE = re.compile(r'StatementEstRows="(\d+)"')
 
     def count_approx(self, query):
-        try:
-            with self.connection.cursor() as cur:
+        with self.connection.cursor() as cur:
+            try:
                 cur.execute("SET SHOWPLAN_XML ON")
                 try:
                     cur.execute(query)
                     result = cur.fetchone()
                     return int(self.EST_ROWS_RE.search(result[0]).group(1))
+                except AttributeError: 
+                    # This is to catch a bug from SQL Server 2012
+                    # (the StatementEstRows=float instead of an int)
+                    pass
                 finally:
                     cur.execute("SET SHOWPLAN_XML OFF")
-        except pymssql.Error as ex:
-            if "SHOWPLAN permission denied" in str(ex):
-                warnings.warn("SHOWPLAN permission denied, count approximates will not be used")
-                return None
-            raise BackendError(str(ex)) from ex
+            except pymssql.Error as ex:
+                if "SHOWPLAN permission denied" in str(ex):
+                    warnings.warn("SHOWPLAN permission denied, count approximates will not be used")
+                    return None
+                raise BackendError(str(ex)) from ex
+            # In case of an AttributeError, give a second chance:
+            # Use the long method for counting (or upgrade SQL Server version :) ) 
+            cur.execute("SELECT count(*) FROM ( {} ) x".format(query))
+            result = cur.fetchone()
+            return result[0]
--- a/Orange/data/sql/backend/mssql.py
+++ b/Orange/data/sql/backend/mssql.py
@@ -127,10 +127,10 @@ class PymssqlBackend(Backend):
                     if not match:
                     # Either StatementEstRows was not found or
                     # a float is received.
-                    # If it is a float then it is most probable 
+                    # If it is a float then it is most probable
                     # that the server's statistics are out of date
                     # and the result is false. In that case
-                    # it is preferable to return None so 
+                    # it is preferable to return None so
                     # an exact count be used.
                         return None
                     return int(match.group(1))
--- a/Orange/data/sql/backend/mssql.py
+++ b/Orange/data/sql/backend/mssql.py
@@ -125,8 +125,8 @@ class PymssqlBackend(Backend):
                     result = cur.fetchone()
                     return int(self.EST_ROWS_RE.search(result[0]).group(1))
                 except AttributeError:
-                # This is to catch a bug from SQL Server 2012
-                # resulting StatementEstRows is float instead of int
+                # This is to catch a float received in StatementEstRows
+                # a float is received when the server's statistics are out of date.
                     pass
                 finally:
                     cur.execute("SET SHOWPLAN_XML OFF")
@@ -136,7 +136,7 @@ class PymssqlBackend(Backend):
                     return None
                 raise BackendError(str(ex)) from ex
             # In case of an AttributeError, give a second chance:
-            # Use the long method for counting (or upgrade SQL Server version :) ) 
+            # Use the long method for counting 
             cur.execute("SELECT count(*) FROM ( {} ) x".format(query))
             result = cur.fetchone()
             return result[0]
--- a/Orange/data/sql/backend/mssql.py
+++ b/Orange/data/sql/backend/mssql.py
@@ -124,9 +124,9 @@ class PymssqlBackend(Backend):
                     cur.execute(query)
                     result = cur.fetchone()
                     return int(self.EST_ROWS_RE.search(result[0]).group(1))
-                except AttributeError: 
-                    # This is to catch a bug from SQL Server 2012
-                    # (the StatementEstRows=float instead of an int)
+                except AttributeError:
+                # This is to catch a bug from SQL Server 2012
+                # resulting StatementEstRows is float instead of int
                     pass
                 finally:
                     cur.execute("SET SHOWPLAN_XML OFF")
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owsql.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/python
```
astaric: As widgets are mostly used from Canvas, we usually do not include this. For testing, widget can be run with python widget_module.py.

Orange/widgets/data/owsql.py
```diff
@@ -20,15 +22,15 @@
...
+class TableModels(PyListModel):
     def data(self, index, role=Qt.DisplayRole):
         row = index.row()
         if role == Qt.DisplayRole:
             return str(self[row])
         return super().data(index, role)
 
 
-class BackendModel(PyListModel):
+class BackendModels(PyListModel):
```
astaric: [Model](http://doc.qt.io/qt-5/model-view-programming.html) is a part of the Qt's Model/View paradigm and should remain singular.

Orange/widgets/data/owsql.py
```diff
@@ -85,10 +87,10 @@ def __init__(self):
         vbox = gui.vBox(self.controlArea, "Server", addSpace=True)
         box = gui.vBox(vbox)
 
-        self.backendmodel = BackendModel(Backend.available_backends())
+        self.backendmodels = BackendModels(Backend.available_backends())
```
astaric: Same as above, model should remain singular. If you prefer plural, I would simply use
self.backends = BackendModel(...)

Orange/widgets/data/owsql.py
```diff
@@ -124,17 +126,23 @@ def __init__(self):
...
+        self.tablecombo.setModel(self.tablemodels)
         self.tablecombo.setToolTip('table')
         tables.layout().addWidget(self.tablecombo)
-        self.tablecombo.activated[int].connect(self.select_table)
+        self.connect()
+
+        index = self.tablecombo.findText(str(self.table))
+        if index != -1:
+            self.tablecombo.setCurrentIndex(index)
+        self.tablecombo.activated[int].connect(self.select_table) # set up the callback to select_table in case of selection change
```
astaric: Could you move the commend in a line before the command? We try to limit the lines in source code to maximum of 100 characters (though some go even further and limit it to 79)

Orange/widgets/data/owsql.py
```diff
@@ -232,25 +241,25 @@ def connect(self):
...
 
-        self.tablemodel.append("Select a table")
-        self.tablemodel.extend(self.backend.list_tables(self.schema))
-        self.tablemodel.append("Custom SQL")
+        self.tablemodels.append("Select a table")
+        self.tablemodels.append("Custom SQL")
+        self.tablemodels.extend(self.backend.list_tables(self.schema))
 
     def select_table(self):
+        "Called on tablecombo selection change"
```
astaric: """ is usually used for docstrings

Orange/widgets/data/owsql.py
```diff
@@ -347,6 +360,7 @@ def get_table(self):
                 domain = s.get_domain(inspect_values=True)
                 self.Information.data_sampled()
             else:
+#
```
astaric: Please remove

Orange/widgets/data/owsql.py
ceprio: Oh, ok I'll do that.

Orange/data/sql/backend/mssql.py
```diff
@@ -117,17 +117,26 @@ def _guess_variable(self, field_name, field_metadata, inspect_table):
...
-            with self.connection.cursor() as cur:
+        with self.connection.cursor() as cur:
+            try:
                 cur.execute("SET SHOWPLAN_XML ON")
                 try:
                     cur.execute(query)
                     result = cur.fetchone()
                     return int(self.EST_ROWS_RE.search(result[0]).group(1))
+                except AttributeError:
+                # This is to catch a bug from SQL Server 2012
```
astaric: If I understand correctly, the query plan in SQL Server 2012 produces an XML, which does not match the EST_ROWS_RE regexp as it represents the number of rows as float? If this is the case, why not update the regular expression to match both?

Instead of
```
StatementEstRows="(\d+)"
```
which expects an integer between the parentheses, we could use something like this
```
r'StatementEstRows="(\d+)(\.\d*)?"'
```
the second regular expressions allow the number to end with a decimal point and some numbers, but only parses the whole part

ceprio: Not exacty, after searching it seem that our SQL server statistics are not maintained often enough and this results in bad values being returned in form of a float thus the AttributeError. I'm not sure if a well maintained server would return a float: I have a second database where statistics are maintained and I never received a float from the Plan query.
It would require more samples to figure out what to do. This code works in all situations, the drawback is the execution time if a float is returned. Maybe there should be a warning issued like those deprecated warnings found in some libraries saying that "statistics are out of date, SELECT count(*) used." but I do not know how to throw such a message in Orange.

Orange/widgets/data/owsql.py
```diff
@@ -20,7 +21,7 @@
 MAX_DL_LIMIT = 1000000
 
 
-class TableModel(PyListModel):
+class TableModels(PyListModel):
```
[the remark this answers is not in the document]
ceprio: Should have anticipated this one.

Orange/data/sql/backend/mssql.py
ceprio: I modified the code comments to explain the reason of the AttributeError. If someone ever comes back with a float with the right estimated result, the code will have to be updated as you suggested and if possible with a sanity check on the statistics (so un-maintained statistics would be caught).

ceprio:
The code has been updated, please check comments (especially mssql.py).
Thx

Orange/data/sql/backend/mssql.py
```diff
@@ -117,17 +117,26 @@ def _guess_variable(self, field_name, field_metadata, inspect_table):
...
+            try:
                 cur.execute("SET SHOWPLAN_XML ON")
                 try:
                     cur.execute(query)
                     result = cur.fetchone()
                     return int(self.EST_ROWS_RE.search(result[0]).group(1))
+                except AttributeError:
+                # This is to catch a float received in StatementEstRows
+                # a float is received when the server's statistics are out of date.
+                    pass
```
astaric: I have read your comment and I agree, if the float is a result of a reallly bad estimate, it should not be used. I still do not like the flow of this function though. I would put it like this:
```
match = self.EST_ROWS_RE.search(result[0])
if not match:
    # Comment explaining why this happens
    return None
return int(match.group(1))
```

Orange/data/sql/backend/mssql.py
```diff
@@ -117,17 +117,26 @@ def _guess_variable(self, field_name, field_metadata, inspect_table):
...
-                return None
-            raise BackendError(str(ex)) from ex
+            except pymssql.Error as ex:
+                if "SHOWPLAN permission denied" in str(ex):
+                    warnings.warn("SHOWPLAN permission denied, count approximates will not be used")
+                    return None
+                raise BackendError(str(ex)) from ex
+            # In case of an AttributeError, give a second chance:
+            # Use the long method for counting
+            cur.execute("SELECT count(*) FROM ( {} ) x".format(query))
```
astaric: This should not be necessary. When count_approx returns None, an exact count is already performed (see [SqlTable.count_approx](https://github.com/biolab/orange3/blob/master/Orange/data/sql/table.py#L274))

ceprio:
Your proposal makes much sense, I had not seen SqlTable.count_approx. Thank you for that.
Tested with success and your way also catches the non-existance of StatementEstRows in the result if that ever appends.
Thank you,

astaric:
Thank you for your contribution. It will be included in the next release (planned for the last week of October)

Issue #1592, opened 2016-09-23
Variable reuse does not take Variable.attributes into account


ales-erjavec:
##### Orange version

3.3.*
##### Expected behavior

Loading the following two datasets in the same session

```
image
s
type=image origin=https://example.org/
image.png
```

```
image
s
type=image origin=/some/local/path
image.png
```

should produce two distinct variables with different `v.attributes` dictionary
##### Actual behavior

Only one variable instance is created (depending on which is loaded first) and then reused.

astaric:
@janezd, could we solve this by returning variable proxies in make?

janezd:
No, I guess. `make` must return the same variable, otherwise it won't match in any dicts we use for various fast mappings. Besides, sometimes you want to copy attributes.

Do we have two kinds of `Variable.attributes`? Those that are unique (related to images), so they shouldn't get copied and have to match; and those that are shared, like colors?

kernc:
Couldn't it be a question of:

1. redefining `__eq__` and `__hash__` into something sensible:
    ```py
    def __eq__(self, other):
        return self.name == other.name

    def __hash__(self):
        return hash(self.name)
```
2. not using `Variable.make`,
3. totally defining the vartype → vartype transformation matrix.

Let me know if no. 3 needs any explaining. Currently, we are inconsistent anyway (cf. `__eq__`, `__hash__`).

ajdapretnar:
@ales-erjavec What is the status of this issue? From what I could reproduce (noob skills), the problem remains, but we have two merged PRs as reference...

ales-erjavec:
It seems to be fixed.

PR #2667, opened 2017-10-10
[FIX] Variable.make returns proxies

astaric:
##### Issue
gh-1592

##### Description of changes
make returns proxies

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/data/tests/test_variable.py
```diff
@@ -441,6 +443,8 @@ def test_make_proxy_disc(self):
         self.assertEqual(abc, abc1)
         self.assertEqual(abc, abc2)
         self.assertEqual(abc1, abc2)
+        self.assertEqual(hash(abc), hash(abc1))
+        self.assertEqual(hash(abc1), hash(abc2))
```
janezd: Maybe add a test for non-equal hashes? I can imagine a scenario in which we mess the code so that all variables have the same hash. :)

Orange/data/tests/test_variable.py
```diff
@@ -467,6 +471,37 @@ def test_make_proxy_cont(self):
...
+        image1 = image.make_proxy()
+        image2 = image1.make_proxy()
+
+        image.attributes["origin"] = "a"
+        image1.attributes["origin"] = "b"
+        image2.attributes["origin"] = "c"
+
+        self.assertEqual(image.attributes["origin"], "a")
+        self.assertEqual(image1.attributes["origin"], "b")
+        self.assertEqual(image2.attributes["origin"], "c")
```
janezd: You could also add a simpler and more explicit test `self.assertIsNot(image.attributes, image1.attributes)`, and same for the other pairs.

PR #2668, opened 2017-10-10
[FIX] Report Table: Make Table Headers Bold

nikicc:
##### Issue

When reporting a view (or model?) through `report_table` header column and row is printed in the same format as the data values. Although there is a code that mark header columns as `<th>` (and not as `<td>`) this is invalidated by the `font-weight` attribute.

 E.g. in `Test & Score` before:

screen shot 2017-10-10 at 16 33 43
and after:
screen shot 2017-10-10 at 16 33 07

##### Description of changes
Set `font-weight` to `bold` when inside header so `<th>` is not invalidated.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/report/index.html
+++ b/Orange/canvas/report/index.html
@@ -70,6 +70,10 @@
         padding: 2px 6px 2px 6px;
     }
 
+    th {
+        font-weight: bold;
+    }
+
     @media screen {
         td, th {
             white-space: nowrap;
--- a/Orange/canvas/report/report.py
+++ b/Orange/canvas/report/report.py
@@ -241,7 +241,7 @@ class Report:
                     bgcolor = 'transparent'
 
                 font = data(Qt.FontRole)
-                weight = 'bold' if font and font.bold() else 'normal'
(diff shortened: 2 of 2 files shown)


janezd:
I like the change. But what if we changed the style definition for the `th` tag instead? The tag is used in several places and I suppose that we'd always like to have it bold.

nikicc:
> I like the change. But what if we changed the style definition for the th tag instead? The tag is used in several places and I suppose that we'd always like to have it bold.

I agree. Please, check again.

janezd:
Could we avoid `!important`? I think the culprit is (report.py, line 244)

`weight = 'bold' if font and font.bold() else 'normal'`.

What if we had 

`weight = 'font-weight: bold;' if font and font.bold() else ''`

and then replace `'font-weight:{weight};'` with `{weight}`? That is, skip `font-weight` property when it is not bold?

Another benefit from this would be that our HTML would look a small tad less like it was generated by Word. :)

nikicc:
Fixed, please check again.

PR #2669, opened 2017-10-10
Remove explicit Report button, improve widget layouts

janezd:
##### Issue

#2514 moved Report button to the status bar.

- Widgets that explicitly positioned the button still show it.
- The layout of widgets without the Report button is changed, and sometimes doesn't look good.

##### Description of changes

Remove Report button, check and fix layouts.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owimpute.py
+++ b/Orange/widgets/data/owimpute.py
@@ -201,9 +201,10 @@ class OWImpute(OWWidget):
 
         box = gui.auto_commit(
             self.controlArea, self, "autocommit", "Apply",
-            orientation=Qt.Horizontal, checkbox_label="Apply automatically")
-        box.layout().insertSpacing(0, 80)
-        box.layout().insertWidget(0, self.report_button)
+            orientation=Qt.Horizontal,
+            checkbox_label="Apply automatically")
+        box.button.setFixedWidth(180)
+        box.layout().insertStretch(0)
 
         self.data = None
         self.learner = None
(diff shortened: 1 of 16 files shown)


Orange/widgets/data/owimpute.py
```diff
@@ -202,8 +202,8 @@ def __init__(self):
         box = gui.auto_commit(
             self.controlArea, self, "autocommit", "Apply",
             orientation=Qt.Horizontal, checkbox_label="Apply automatically")
-        box.layout().insertSpacing(0, 80)
-        box.layout().insertWidget(0, self.report_button)
+        box.button.setFixedWidth(180)
```
jerneju: Why is here this line and not also in **Impute** widget (https://github.com/biolab/orange3/pull/2669/files#diff-a2cbec3457e61cdb3228e0e4b7b95585R233)?

janezd: You mean Discretize? Because I didn't like any of them and I couldn't decide which I disliked less. If a widget is wide, there is no good position for the Apply button.

Now I how changed (unified) the layout of Discretize, Color, Concatenate and Impute to look the same and reasonably nice.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -269,11 +269,9 @@ def add_learner_name_widget(self):
             orientation=Qt.Horizontal, callback=self.learner_name_changed)
 
     def add_bottom_buttons(self):
-        box = gui.hBox(self.controlArea, True)
-        box.layout().addWidget(self.report_button)
-        gui.separator(box, 15)
-        self.apply_button = gui.auto_commit(box, self, 'auto_apply', '&Apply',
-                                            box=False, commit=self.apply)
+        self.apply_button = gui.auto_commit(
```
jerneju: Neither this class nor class's children use attribute `apply_button`. Only the tests use it.

PR #2670, opened 2017-10-10
[FIX] owhierarchicalclustering: Fix performance on deselection

ales-erjavec:
##### Issue

The widget is slow to respond when the selection is changed in such a way that the number of groups diminishes significantly.

Example:

File (vehicle.tab) -> Distances (Euclidean) -> Hierarchical clustering (Average)

In the *Selection* box select and set *Height ratio: 0.1%*. Switch between that and *Top N: 3* Going from Height ratio to top N is significantly slower then the other way.

##### Description of changes

The code (`set_selected_items`) had quadratic time complexity in the number of deselected items.
Fixed by a more judicious placement of calls to `_re_enumerate_selections`

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -439,6 +439,7 @@ class DendrogramWidget(QGraphicsWidget):
         if item in self._selection:
             if state == False:
                 self._remove_selection(item)
+                self._re_enumerate_selections()
                 self.selectionChanged.emit()
         else:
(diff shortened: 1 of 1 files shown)


ajdapretnar:
Hmm, it is still very slow (I can't really tell the difference in time, it just _feels_ very long). You can say it can't be done, that's fine. I just noticed this when working on a 33180 x 149 data set.

jerneju:
@ales-erjavec Please restart Travis and AppVeyor.

PR #2671, opened 2017-10-11
[FIX] Scatter Plot Graph: reset view when zooming too much

jerneju:
##### Issue
Zooming out and then zooming in crashes the widget which uses Scatter Plot Graph.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


jerneju:
https://sentry.io/biolab/orange3/issues/300247018/

PR #2675, opened 2017-10-12
[FIX] MDS: No optimization when subset data

jerneju:
##### Issue
~~1. MDS cannot handle sparse data.~~ That will be fixed in #2690 .
2. Optimization runs when subset data is sent to the widget. This is not necessary.
3. Plot resets when _Show similar pairs_ slider is moved.
4. Similar pairs are not removed when _Show similar pairs_ slider is set to 0. And they should be removed.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -350,7 +350,6 @@ class OWMDS(OWWidget):
         # invalidate the pen/brush when the subset is changed
         self._subset_mask = None  # type: Optional[np.ndarray]
         self.controls.graph.alpha_value.setEnabled(subset_data is None)
-        self._invalidated = True
 
     def _clear(self):
         self._similar_pairs = None
@@ -574,10 +573,10 @@ class OWMDS(OWWidget):
 
     def handleNewSignals(self):
         if self._invalidated:
+            self.__draw_similar_pairs = False
             self._invalidated = False
             self._initialize()
             self.start()
-        self.__draw_similar_pairs = False
 
         if self._subset_mask is None and self.subset_data is not None and \
                 self.data is not None:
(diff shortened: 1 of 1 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/367233902/

janezd:
Add tests for sparse data.

ajdapretnar:
This now works!
But sending Data Subset to MDS does not do anything. I mean it does not hollow out the points as it should... Am I missing something?

Orange/widgets/unsupervised/owmds.py
```diff
@@ -651,18 +652,21 @@ def _setup_plot(self, new=False):
 
         data = self.data
 
-        primitive_metas = tuple(a for a in data.domain.metas if a.is_primitive())
-        keys = [k for k, a in enumerate(data.domain.metas) if a.is_primitive()]
-        data_metas = data.metas[:, keys]
-
-        attributes = self.data.domain.attributes + (self.variable_x, self.variable_y) + \
-                     primitive_metas
+        if self.data.is_sparse():
```
nikicc: Why do we need a different logic for sparse here?

nikicc:
Also, on hover doesn't show all data for sparse. Is there a special reason for this?

Orange/widgets/unsupervised/owmds.py
```diff
@@ -271,6 +271,8 @@ def update_regression_line(self):
 
     def init_attr_values(self):
         domain = self.data and len(self.data) and self.data.domain or None
+        if domain is not None and self.data.is_sparse():
+            domain = Domain(attributes=[], class_vars=domain.class_vars)
```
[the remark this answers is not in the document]
nikicc: #2690 should probably solve this issue.

Orange/widgets/unsupervised/owmds.py
jerneju: #2693

PR #2677, opened 2017-10-13
[FIX] Outliers widget no longer checks classes and doesn't crash on singular covariances matrices

### LINKED ISSUE #2588: Outliers: Multiple class value error instead of No class value error
##### Orange version

3.5.dev

##### Expected behavior
Outliers returns a correct error message. If there are no class values in the data, it says 'Data missing a class variable'.

##### Actual behavior
Outliers returns 'Multiple class data is not supported.'

File (iris) - Select Columns (move iris to metas) - Outliers.

janezd:
##### Issue

- Outliers widget does not work with multiple classes.
- ... and crashes when variance matrices is singular, instead of reporting an error.
- `EllipticEnvelopeLearner` removes instances without classes.

Fixes #2588.

##### Description of changes

- Remove tests for classes
- Fix the crash for singular data
- Set proper preprocessors for `EllipticEnvelopeLearner`

Tests do not add much new; they passed before, too. The PR actually removes tests that checked for what shouldn't had been checked.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/classification/elliptic_envelope.py
+++ b/Orange/classification/elliptic_envelope.py
@@ -2,6 +2,7 @@ import sklearn.covariance as skl_covariance
 
 from Orange.base import SklLearner, SklModel
 from Orange.data import Table
+from Orange.preprocess import Continuize, RemoveNaNColumns, SklImpute
 
 __all__ = ["EllipticEnvelopeLearner"]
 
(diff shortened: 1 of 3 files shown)


Orange/classification/elliptic_envelope.py
```diff
@@ -27,6 +28,7 @@ def mahalanobis(self, observations):
 class EllipticEnvelopeLearner(SklLearner):
     __wraps__ = skl_covariance.EllipticEnvelope
     __returns__ = EllipticEnvelopeClassifier
+    preprocessors = [Continuize(), RemoveNaNColumns(), SklImpute()]
```
jerneju: Where are these preprocessors used?

kernc: https://github.com/kernc/orange3/blob/8ee65c0f5905cf6c9bd7297d5dd256a274a8d303/Orange/base.py#L132-L143

PR #2678, opened 2017-10-13
[ENH] Add Groups column to Selected Data in Scatter plot output

janezd:
##### Issue

When selecting multiple groups in scatter plot, the Selected Data output contains no indication about the group.

##### Description of changes

- Add a Groups column and corresponding tests
- Refactoring of `send_data`.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -21,9 +21,8 @@ from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.visualize.owscatterplotgraph import OWScatterPlotGraph
 from Orange.widgets.visualize.utils import VizRankDialogAttrPair
 from Orange.widgets.widget import OWWidget, AttributeList, Msg, Input, Output
-from Orange.widgets.utils.annotated_data import (create_annotated_table,
-                                                 ANNOTATED_DATA_SIGNAL_NAME,
-                                                 create_groups_table)
+from Orange.widgets.utils.annotated_data import (
+    create_annotated_table, create_groups_table, ANNOTATED_DATA_SIGNAL_NAME)
 
 
 class ScatterPlotVizRank(VizRankDialogAttrPair):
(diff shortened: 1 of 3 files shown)


Orange/widgets/visualize/owscatterplot.py
```diff
@@ -428,25 +428,35 @@ def selection_changed(self):
...
-            if len(selection) > 0:
-                selected = self.data[selection]
-        if graph.selection is not None and np.max(graph.selection) > 1:
-            annotated = create_groups_table(self.data, graph.selection)
-        else:
-            annotated = create_annotated_table(self.data, selection)
-        self.Outputs.selected_data.send(selected)
-        self.Outputs.annotated_data.send(annotated)
+        data = self.data
+        domain = data.domain
```
jerneju: There is no check if there is data or not.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -428,25 +428,35 @@ def selection_changed(self):
...
+            groups = graph.selection[selection, np.newaxis]
+            group_var = DiscreteVariable(
+                name=get_next_name(domain, "Group"),
+                values=["G{}".format(i + 1) for i in range(np.max(groups))])
+            sel_domain = Domain(
+                domain.attributes, domain.class_vars,
+                domain.metas + (group_var, ))
+            metas = np.hstack((self.data.metas[selection], groups - 1))
+            return Table(
+                sel_domain, data.X[selection], data.Y[selection], metas)
```
kernc: Wondering, could you somehow reuse/extend `create_annotated_table()` for this?

Orange/widgets/visualize/owscatterplot.py
janezd: Yet it survived the tests. Not good.

janezd: Oh, it didn't, I just didn't run all tests. Travis caught me. That's good.

Orange/widgets/visualize/owscatterplot.py
janezd: I started by filtering the annotated data, but it was messy. Now I've changed the `create_groups_table` and it's way simpler. Thanks for the suggestion.

Orange/widgets/visualize/tests/test_owscatterplot.py
```diff
@@ -18,13 +18,17 @@ class TestOWScatterPlot(WidgetTest, WidgetOutputsTestMixin):
...
+        cls.same_input_output_domain = False
 
         cls.signal_name = "Data"
         cls.signal_data = cls.data
 
     def setUp(self):
         self.widget = self.create_widget(OWScatterPlot)
 
+    def _compare_selected_annotated_domains(self, selected, annotated):
+        pass
```
kernc: What is this?

janezd: This disables an inapplicable test from the base class.

Hierarchical clustering, which has a similar situation, [replaces this](https://github.com/biolab/orange3/blob/master/Orange/widgets/unsupervised/tests/test_owhierarchicalclustering.py#L32) with checking that the domains have same variables and that the annotation appears only in one domain ... but I don't see much sense in testing this here, so I just disabled the test.

I now added a short comment explaining why the method is overridden.

kernc: > I now added a short comment explaining why the method is overridden.

Thanks.

janezd:
As suggested by @BlazZupan, the annotation column is now added as a class if the data had no class before. Any arguments against the change?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -428,25 +427,26 @@ def selection_changed(self):
...
     def send_data(self):
-        selected = None
-        selection = None
         # TODO: Implement selection for sql data
+        def _get_selected():
+            if not len(selection):
+                return None
+            return create_groups_table(data, graph.selection, False, "Group")
+
+        def _get_annotated():
```
jerneju: Let's consider moving this somewhere else. Code in this function is used in many widgets.

janezd:
You mean `_get_annotated`? It is useful only within `send_data`. We'd better consider if this `send_data` is common to all projection widgets (Scatter plot, MDS, Linear, FreeViz). It barely interact with the widget; maybe it belong to the graph -- not with this name, of course -- as a method that prepares the data sent by the widget. (I said maybe -- I haven't really looked into it.)

PR #2679, opened 2017-10-14
Distributions: Return new distribution object on normalize

pavlin-policar:
##### Issue
Normalizing a distribution would normalize inplace, leaving no way to revert the operation.

##### Description of changes
When normalizing a distribution, return a new distribution object, keeping the old one intact.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
If we do this, I'd rename the method to `normalized`. It's like `sort` and `sorted` for lists. (On the other hand, strings have `replace`, but I suppose it's because `replaced` wouldn't really work.)

Why do you need this? In a normal process, you would compute a distribution, normalize it ... and forget about the non-normalized, wouldn't you? Or, if you got the distribution as an argument and thus wouldn't dare to change it in place, you'd just copy it before calling `normalize`.

We could also keep `normalize` as it is, and add `normalized` that would copy the distribution, call its `normalize` and return it.

pavlin-policar:
I like `normalize` better because it's a verb, and most methods have a verb name. I think the only exception are some python builtin methods, like `sorted`. If we followed the principle of least astonishment, I think it's more typical to have verbs as methods and attributes/properties as nouns. That said, a property on the distribution class would be okay too.

Tbh, I don't know if I'll end up using this, but I saw it and thought it could be improved. Following the same principle, inplace operations are rare in Python, `np.sort` being the only one that comes to mind. As such, they are "surprising". And even in the sort case, it makes sense because copying can be expensive, this is not the case in distributions, and also, since it would happen very rarely, any speed or memory inefficiency would be negligible.

Also as you pointed out, if we wanted to keep around the original distribution, we'd have to copy it then normalize, but that requires us to know how to copy a distribution object, complete with the `unknowns`. While this is very easy, it requires us to know how to copy it properly, and that seems like more work than necessary.

Again, if you feel this is completely unnecessary, that's fine, but I think this would make working with distributions slightly simpler.

PR #2680, opened 2017-10-14
[FIX] Update distributions

pavlin-policar:
##### Issue
Builds upon #2679. `class_distributions` crashed when given data with multiple target variables.

##### Description of changes
- Fixes the crash
- I implemented `min` and `max` for ordered discrete variables
- Better implementation of `random`, renamed to `sample` which can now return an arbitrarily sized array sampled from the distribution
- The code was very old and formatted very strangely in some places, making it difficult to read, so I attempted to rectify some of it.

If my changes to the code formatting are undesired (since they do mess up the history a bit), the important commit is https://github.com/biolab/orange3/commit/ac27b734c76ce23e507cb1ed4820206e84ae4cb4, which fixes the crash. 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


kernc:
Leave a comment what _resolution_ you are closing with.

pavlin-policar:
The PRs I closed were very old, and the issues I was trying to solve with them were solved in a different way. While the code in this PR could be useful, it is currently unnecessary and as far as I'm aware, will be obsolete when we move to pandas. Perhaps the only truly useful commit was ac27b73 where I fixed a bug that lay undiscovered for several years, most other commits were refactorings.

kernc:
But https://github.com/biolab/orange3/commit/ac27b734c76ce23e507cb1ed4820206e84ae4cb4 (https://github.com/biolab/orange3/pull/2680/commits/6af19064a0e0e257c5b7bfcd32d13f70bf39fb86) is not yet applied?

pavlin-policar:
Looking over the commits, I realize that only 7d675ba was problematic, the others are all valid improvements. I'll drop that commit and reopen.

PR #2685, opened 2017-10-16
[FIX] OWRank: Fix autocommit

lanzagar:
##### Issue

Selecting rows was sending new selections even when auto-send was unchecked.


janezd:
Could you add some tests?

Actually, I added them. (You're welcome. ;) )

lanzagar:
(Thank you :))

PR #2688, opened 2017-10-16
[RFC][WIP] OWWidget: Add a menu bar

ales-erjavec:
##### Issue

Menu bar is a more appropriate place for the  Help / Report / Save image actions.

##### Description of changes

Move the Help/Report/Save actions to the menu bar. Extend with other useful actions.

Subclasses can/should also use the menu to add their own actions. This is done for 'Data Table' and 'Paint Data'

Other widgets that should use the menu:
- Scatter Plot: Zoom In/ Out, Zoom reset
- Silhouette Plot, Hierarchical Clustering (all with s zoom slider/actions) Zoom In/Out or Increase/Decrease Font size.
...

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/scheme/widgetsscheme.py
+++ b/Orange/canvas/scheme/widgetsscheme.py
@@ -23,10 +23,13 @@ import enum
 import itertools
 from collections import namedtuple, deque
 from urllib.parse import urlencode
-from typing import List, Tuple
+from functools import partial
+
+from typing import List, Tuple, Iterator
+
 import sip
 
-from AnyQt.QtWidgets import QWidget, QShortcut, QLabel, QSizePolicy, QAction
+from AnyQt.QtWidgets import QWidget, QLabel, QSizePolicy, QAction, QMenu
 from AnyQt.QtGui import QKeySequence, QWhatsThisClickedEvent
 
 from AnyQt.QtCore import (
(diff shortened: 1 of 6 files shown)


BlazZupan:
Help menu disappears when I press on the report button. I opened Orange, used a File widget, checked that Help menu appears (it does) and then clicked on the report icon. The report appears, but Help menu is now gone. I placed another widget on canvas, opened the widget, and help was missing as well.

lanzagar:
This probably makes much more sense on macs than what we have currently.
On win and linux systems the menu is above every widget, which is either ok or unwanted, depending on the user and personal preferences.
I suggest we have an option in Settings where we can toggle to show the menu bar over all widgets or not.
The default should be to show it, it is also very nice that users can now see the shortcuts for all the actions next to them in the menu. And once they remember the shortcuts for what they are using, they can turn the menu off and have cleaner widgets showing just the contents.

ajdapretnar:
Quite honestly, I don't like this feature. On Macs it is probably invisible, but on Windows this adds yet another line to the window size, making widgets even taller. I don't see how this improves the functionality, that is currently very nicely implemented in the footer of the window. 🤷‍♀️ If you think this is a feature, then at least, in agreement with @lanzagar, let us enable turning it off.

Some examples on Windows:

BlazZupan:
This is working and is implemented exactly as we have intended and planned. The idea behind this implementation was to reduce the size of the widget window by removing the status bar. The status bar, however, became important to show any warnings and hence cannot be removed. While the window size on the Macs does not change with the introduction of the menus, this is not so on windows. I have discussed all these issues with @ales-erjavec and we agreed to temporarily close this PR to currently stay with the status bar as it is, and perhaps think about the menus later if the need arises and we decide to move the warnings somewhere else.

PR #2689, opened 2017-10-16
[RFC][FIX] OWScatterPlot: Change output Feature to AttributeList

### LINKED ISSUE #2687: Scatter Plot: send Features from one SP to another
##### Orange version

3.7.dev

##### Expected behavior
Scatter Plot (1) can be connected to Scatter Plot (2) by Features output as well.

##### Actual behavior
Scatter Plot (2) doesn't accept Features from Scatter Plot (1).

File - Scatter Plot (1) - Scatter Plot (2)

Line in Edit Links cannot be connected.

janezd:
##### Issue

Fixes #2687.

##### Description of changes

Change the output type of `Feature` to `AttributeList`. Alternatively, input could be changed to `Table` but other widget (Sieve, Select columns...) use `AttributeList`; I see little use for `Table` in these contexts.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -9,8 +9,7 @@ from sklearn.neighbors import NearestNeighbors
 from sklearn.metrics import r2_score
 
 import Orange
-from Orange.data import Table, Domain, StringVariable, ContinuousVariable, \
-    DiscreteVariable
+from Orange.data import Table, Domain, ContinuousVariable, DiscreteVariable
 from Orange.canvas import report
 from Orange.data.sql.table import SqlTable, AUTO_DL_LIMIT
 from Orange.preprocess.score import ReliefF, RReliefF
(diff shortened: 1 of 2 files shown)


janezd:
The widget still outputs the signal Feature even when the auto send is disabled (like it did before this PR). I don't see much wrong with it: the auto send setting applies to the data, while features are just always send.

If this is undesired, let @lanzagar contribute a commit with a fix and I'll add a test, as in #2685. :)

lanzagar:
Not sure if it is undesired. It could be a bit confusing, but is also simple enough that I don't think we need to worry about it until we hear a complaint :)

[Side Note] I am not even sure we need an auto-send control here at all. They make sense in widgets where an action could take time. When we just want to avoid updates in downstream widgets, this can be done by pausing the signal. That should probably be the preferred/universal way, and it could be up for (a future) discussion whether another way to do the same thing is good.
Just food for thought, not opening that discussion here and now.

PR #2690, opened 2017-10-17
[ENH] DomainModel: Don't Show Hidden Variables by Default

nikicc:
##### Issue

DomainModel shows all variables, even those, that are marked as hidden. 

##### Description of changes
DomainModel shows only non-hidden variables by default. If `skip_hidden_vars` is set to `False`, all variables are shown.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -1,4 +1,3 @@
-import pickle
 from numbers import Number, Integral
 from math import isnan, isinf
 
@@ -23,6 +22,7 @@ from AnyQt.QtWidgets import (
 import numpy
 
 from Orange.data import Variable, Storage, DiscreteVariable, ContinuousVariable
+from Orange.data.domain import filter_visible
 from Orange.widgets import gui
 from Orange.widgets.utils import datacaching
 from Orange.statistics import basic_stats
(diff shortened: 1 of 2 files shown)


Orange/widgets/utils/itemmodels.py
```diff
@@ -903,6 +904,8 @@ def set_domain(self, domain):
                     *(vars for i, vars in enumerate(
                         (domain.attributes, domain.class_vars, domain.metas))
                       if (1 << i) & section)))
+                if self.skip_hidden_vars:
+                    to_add = filter_visible(to_add)
```
janezd: `to_add = list(filter_visible(to_add))`

This is needed because of `if to_add` [later on](https://github.com/biolab/orange3/pull/2690/files#diff-4e5076ff2a69272284062b73fbddf0faR918), which tests whether there is some more content and a separator needs to be added. Since `filter_visible` returns a generator, the condition is always `True` and separator is added even when the generator would return nothing.

janezd:
Could you add some tests?

Actually, I added them. (You're welcome. :) ).

I haven't added the fix for the bug from the comment. Squash it into your previous commit. :)

Orange/widgets/utils/itemmodels.py
nikicc: It should be fixed now. Though, in the future, we might consider changing `filter_visible` to return a tuple instead of a generator.

PR #2691, opened 2017-10-17
[FIX] settings: Do not clear schema_only settings on close_context

ales-erjavec:
##### Issue

Since gh-2301 schema_only settings are destructively cleared when used in combination with a context handler. This breaks (what I would hope to be) a valid use case

```python
class A(OWWidget):
    ...
    bar = Setting(None, schema_only=True)
    def __init__(self):
        if self.bar is None:
             self.bar = ...  # not loaded from a saved workflow

    def foo(self):
        self.closeContext()
        self.openContext(...)
        assert self.bar is not None
```
This state where bar is None would then be saved in a workflow.
This pattern is used in Textable add-on for issuing uuids to widgets

The same pattern is also used in OWBaseLearner widget for the learner_name but without a context handler.

I am not sure what the intent there was. Presumably the semantics of the selection restore in scatter plot, ... is such that it is only applied once when restored from a saved workflow.

##### Description of changes

* undo changes to settings from gh-2301
* change selection restore in scatter plot and linear projection accordingly

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -787,17 +787,6 @@ class ContextHandler(SettingsHandler):
         if widget.current_context is None:
             return
 
-        # Clear schema-only settings when *closing* the context
-        # FIXME: Sadly, schema-only settings aren't cleared when non-contextual
-        # SettingsHandler is used. Example: widget has non-contextual
-        # SettingsHandler, user loads schema that includes widget, user
-        # passes new data to widget => widget's schema-only settings still
-        # have previous values because equivalent of below reset was never
-        # called (i.e. `close_context()`).
-        for name, setting in self.known_settings.items():
-            if setting.schema_only:
-                setattr(widget, name, setting.default)
-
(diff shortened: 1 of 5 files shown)


kernc:
>  Presumably the semantics of the selection restore in scatter plot, ... is such that it is only applied once when restored from a saved workflow.

What if not only scatter plot; would all widgets need a `self.__did_restore_state`?

ales-erjavec:
> What if not only scatter plot; would all widgets need a self.__did_restore_state?

Presumably, yes.

ales-erjavec:
A kingdom for a saveState slot.

astaric:
@ales-erjavec the addition of sparse support in scatter plot created conflicts with this. Can you rebase?

ales-erjavec:
Ok.

astaric:
Thanks

PR #2693, opened 2017-10-18
MDS: Fixup for Sparse Data

jerneju:
##### Issues
* Scatter Plot Graph cannot handle sparse data.
* As of #2648 MDS did not work for sparse data.

##### Description of changes
* Scatter Plot Graph and Scaling can now handle sparse data and primitive data in metas by themselves.
* Fix MDS to handle sparse data.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -4,6 +4,7 @@ from xml.sax.saxutils import escape
 from math import log10, floor, ceil
 
 import numpy as np
+import scipy.sparse as sp
 from scipy.stats import linregress
 
 from AnyQt.QtCore import Qt, QObject, QEvent, QRectF, QPointF, QSize
@@ -20,7 +21,6 @@ from pyqtgraph.graphicsItems.ScatterPlotItem import ScatterPlotItem
 from pyqtgraph.graphicsItems.TextItem import TextItem
 from pyqtgraph.graphicsItems.InfiniteLine import InfiniteLine
 from pyqtgraph.Point import Point
-
 
 from Orange.widgets import gui
 from Orange.widgets.utils import classdensity, get_variable_values_sorted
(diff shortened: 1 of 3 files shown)


Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -691,7 +720,7 @@ def update_data(self, attr_x, attr_y, reset_view=True):
             self.plot_widget.addItem(self.density_img)
 
         self.data_indices = np.flatnonzero(self.valid_data)
-        if len(self.data_indices) != self.original_data.shape[1]:
+        if len(self.data_indices) != len(self.original_data):
```
nikicc: Why this change?

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -759,21 +788,15 @@ def set_labels(self, axis, labels):
...
-        size_index = self.get_size_index()
-        if size_index == -1:
+        if self.attr_size is None:
             size_data = np.full((self.n_points,), self.point_width,
                                 dtype=float)
         else:
             size_data = \
                 self.MinShapeSize + \
-                self.scaled_data[size_index, self.valid_data] * \
+                self.scaled_data.get_column_view(self.attr_size)[0][self.valid_data] * \
```
nikicc: Is this needed if we densified before?

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -782,7 +811,7 @@ def compute_sizes(self):
         return size_data
 
     def update_sizes(self):
-        self.master.prepare_data()
+        self.sparse_to_dense()
```
nikicc: Don't we also need the `self.set_data` call here? Same for a few lines below.

jerneju: Yes, after `self.set_data(self.sparse_to_dense())` .

jerneju:
https://sentry.io/biolab/orange3/issues/367233902/

PR #2572, opened 2017-09-09
[FIX] Statistics.unique: Fix Sparse Return Order For Negative Numbers

pavlin-policar:
##### Issue
- The statistics module implementation of `unique` returned incorrectly ordered arrays in case of a negative number e.g.
```
>>> unique(csr_matrix([-1, 0, 0, 1, 2]))
array([ 0, -1,  1,  2], dtype=int64)

>>> np.unique([-1, 0, 0, 1, 2])
array([-1,  0,  1,  2])
```

- The statistics module implementation of `nanunique` did not support any parameters that `unique` supported.

##### Description of changes
Fixes above issues. This PR depends on #2698 for the `dense_sparse` testing utility.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

PR #2696, opened 2017-10-20
Statistics.stack: Move to statistics module

pavlin-policar:
##### Issue
`hstack` and `vstack` existed in `Orange.data.util`, however all the other functions that handle sparse and dense data are situated in `Orange.statistics.util`.

##### Description of changes
Move functions and tests to `Orange.statistics.util` module to keep all functions that deal with sparse/dense compatibility in same module.

Based on #2572.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/util.py
+++ b/Orange/data/util.py
@@ -1,9 +1,10 @@
 """
 Data-manipulation utilities.
 """
-import numpy as np
 import bottleneck as bn
-from scipy import sparse as sp
+import numpy as np
+# Backwards compatibility
+from Orange.statistics.util import hstack, vstack  # pylint: disable=unused-import
 
 
 def one_hot(values, dtype=float):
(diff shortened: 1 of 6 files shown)


Orange/data/util.py
```diff
@@ -1,9 +1,8 @@
 """
 Data-manipulation utilities.
 """
-import numpy as np
 import bottleneck as bn
-from scipy import sparse as sp
+import numpy as np
```
kernc: Add
```py
from Orange.statistics.util import hstack, vstack  # back-compat
```

nikicc:
Hmmm ... I like that we have everything together but importing `vstack` and `hstack` from `Orange.statistics.util` seems strange to me. 

Perhaps, should we introduce a separate module for this sparse/dense compatibility methods? Some other methods from `Orange.statistics.util` could be moved there.

pavlin-policar:
I agree, `Orange.statistics.util` is probably not the best place for this. I think `Orange.data.util` would be a much better module, or something new altogether.

pavlin-policar:
This had conflicts and it would be easier to simply start fresh from master, since the changes in this PR were quite trivial.

We had decided that it would be strange to have common functions (such as `vstack` and `hstack` in a module named `statistics`; a different module would have to be made where we move all the methods that are essentially wrappers around sparse and numpy functions. This introduces backwards compatibility issues.

PR #2697, opened 2017-10-20
[ENH] Domain Model: order without separators

jerneju:
##### Issue
Cannot set order and skip separators

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -875,15 +875,14 @@ class DomainModel(VariableListModel):
 
         Parameters
         ----------
-        order: tuple, can be also int
-            How attributes, metas and classes are ordered and where between
-            them are separators.
+        order: tuple or int
+            Order of attributes, metas, classes, separators and other options
         separators: bool
-            Ignore separators when set to False despite being set in order.
+            If False, remove separators from `order`.
         placeholder: str
-            Placeholder is the text that is shown when no variable is selected.
-        valid_types: tuple (Variable, ContinuousVariable, DiscreteVariable, etc.)
-            Continuous, discrete, etc. variables.
+            The text that is shown when no variable is selected
+        valid_types: tuple
+            (Sub)types of `Variable` that are included in the model
         alphabetical: bool
             If true, variables are sorted alphabetically.
         skip_hidden_vars: bool
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -869,8 +869,26 @@ class DomainModel(VariableListModel):
                  ATTRIBUTES)
     PRIMITIVE = (DiscreteVariable, ContinuousVariable)
 
-    def __init__(self, order=SEPARATED, placeholder=None,
+    def __init__(self, order=SEPARATED, separators=True, placeholder=None,
                  valid_types=None, alphabetical=False, skip_hidden_vars=True, **kwargs):
+        """
+
+        Parameters
+        ----------
+        order: tuple, can be also int
+            How attributes, metas and classes are ordered and where between
+            them are separators.
+        separators: bool
+            Ignore separators when set to False despite being set in order.
+        placeholder: str
+            Placeholder is the text that is shown when no variable is selected.
+        valid_types: tuple (Variable, ContinuousVariable, DiscreteVariable, etc.)
+            Continuous, discrete, etc. variables.
+        alphabetical: bool
+            If true, variables are sorted alphabetically.
+        skip_hidden_vars: bool
+            If true, variables marked as "hidden" are skipped.
+        """
         super().__init__(placeholder=placeholder, **kwargs)
         if isinstance(order, int):
             order = (order,)
@@ -880,6 +898,8 @@ class DomainModel(VariableListModel):
             order = (None,) + \
                     (self.Separator, ) * (self.Separator in order) + \
                     order
+        if not separators:
+            order = [e for e in order if e is not self.Separator]
         self.order = order
         self.valid_types = valid_types
         self.alphabetical = alphabetical
--- a/Orange/widgets/utils/tests/test_itemmodels.py
+++ b/Orange/widgets/utils/tests/test_itemmodels.py
@@ -310,3 +310,21 @@ class TestDomainModel(TestCase):
                             skip_hidden_vars=False)
         model.set_domain(Domain(attrs))
         self.assertEqual(list(model), disc)
+
+    def test_no_separators(self):
+        """
+        GH-2697
+        """
+        attrs = [ContinuousVariable(n) for n in "abg"]
+        classes = [ContinuousVariable(n) for n in "deh"]
+        metas = [ContinuousVariable(n) for n in "ijf"]
+
+        model = DomainModel(order=DomainModel.SEPARATED, separators=False)
+        model.set_domain(Domain(attrs, classes, metas))
+        self.assertEqual(list(model), classes + metas + attrs)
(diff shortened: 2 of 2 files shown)


Orange/widgets/utils/itemmodels.py
```diff
@@ -869,8 +869,19 @@ class DomainModel(VariableListModel):
                  ATTRIBUTES)
     PRIMITIVE = (DiscreteVariable, ContinuousVariable)
 
-    def __init__(self, order=SEPARATED, placeholder=None,
+    def __init__(self, order=SEPARATED, separators=True, placeholder=None,
                  valid_types=None, alphabetical=False, skip_hidden_vars=True, **kwargs):
+        """
+
+        :param order: How attributes, metas and classes are ordered and where between
```
janezd: We lately prefer [Napoleon](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

janezd: How attributes (...) -> Order of attributes, metas, classes, separators and other options

Orange/widgets/utils/tests/test_itemmodels.py
```diff
@@ -310,3 +310,12 @@ def test_filtering(self):
...
         self.assertEqual(list(model), disc)
+
+    def test_no_separators(self):
+        attrs = [ContinuousVariable(n) for n in "abg"]
+        classes = [ContinuousVariable(n) for n in "deh"]
+        metas = [ContinuousVariable(n) for n in "ijf"]
+
+        model = DomainModel(order=DomainModel.SEPARATED, separators=False)
+        model.set_domain(Domain(attrs, classes, metas))
+        self.assertEqual(list(model), classes + metas + attrs)
```
janezd: What about checking whether it works with `separators=True`?

Orange/widgets/utils/itemmodels.py
```diff
@@ -869,8 +869,19 @@ class DomainModel(VariableListModel):
...
+                      them are separators.
+        :param separators: Separators are ignored when set to False despite being set in order.
```
janezd: Separators are ignored -> Ignore separators.
When set to False -> When what if set to False?

I tried to reformulate the sentence, but couldn't. I believe it is because I disagree with the commit as a whole.

Orange/widgets/utils/itemmodels.py
```diff
@@ -869,8 +869,19 @@ class DomainModel(VariableListModel):
...
+        :param placeholder: The first element.
```
janezd: Wrong. Placeholder is the text that is shown when no variable is selected (i.e., `None`). It is usually, but not necessarily, the first element.

See: https://github.com/jerneju/orange3/blob/e62c7a6b785dab00f5c619c561d4cb361636bce0/Orange/widgets/utils/itemmodels.py#L888. `None` is inserted at the beginning of `order` only if it's not already there.

Orange/widgets/utils/itemmodels.py
```diff
@@ -869,8 +869,19 @@ class DomainModel(VariableListModel):
...
+        :param valid_types: For instance continuous, discrete, etc. variables.
```
janezd: "For instance" is an example, not a description.

Orange/widgets/utils/itemmodels.py
```diff
@@ -869,8 +869,19 @@ class DomainModel(VariableListModel):
...
+        :param alphabetical: Alphabetical order of attributes or as they are ordered in domain.
```
janezd: If true, variables are sorted alphabetically.

Orange/widgets/utils/itemmodels.py
```diff
@@ -869,8 +869,19 @@ class DomainModel(VariableListModel):
...
+        :param skip_hidden_vars: hidden vars can be hidden or shown
```
janezd: This is not a description of the option. It also does not explain what are hidden vars.

Orange/widgets/utils/itemmodels.py
```diff
@@ -869,8 +869,19 @@ class DomainModel(VariableListModel):
...
+        :param kwargs:
```
janezd: Remove.

Orange/widgets/utils/tests/test_itemmodels.py
jerneju: I actually added this test before you responded.

jerneju:
I removed the second commit very soon after opening this PR. @janezd sorry for wasting your time.

astaric:
@janezd The idea would be to keep "default order", and skip separators.

When you override order, I agree, it does not make sense to skip separators, as you would not include them anyway. But we would prefer to have as many widgets as possible use the default order, which should include separators where they can be displayed and not where they cannot.

janezd:
What is better? `DomainModel(order=DomainModel.SEPARATED, separators=False)` or `DomainModel(order=DomainModel.UNSEPARATED)`? Or is your point that we could use `DomainModel(separator=False)`?

If the latter, I'd prefer

    SEPARATED = <as it is now>
    UNSEPARATED = <same, without separators>

    def __init__(self, order=None, separators=True, placeholder=None,
                   valid_types=None, alphabetical=False, skip_hidden_vars=True, **kwargs):		                   valid_types=None, alphabetical=False, skip_hidden_vars=True, **kwargs):

        ...
        if order is None:
            order = self.SEPARATED is separators else self.UNSEPARATED

(and no further changes in the code) since this implementation corresponds to what the change (if I understand it correctly) is actually about: the default value for `order` depends on whether separators can be shown or not, and it also matches

> When you override order, I agree, it does not make sense to skip separators.

Or, if you think it's better, we keep `order=SEPARATED` as default, but I'd maybe suggest implementing the change by

    if not separators:
        order = [element for element in order if element is not self.SEPARATOR]

It's a bit longer, but more explicit.

astaric:
It was the latter.

I would prefer to keep one default order so I would go with
```
if not separators:
    order = [element for element in order if element is not self.SEPARATOR]
```

janezd:
This is just computing the same thing every time, instead of having it precomputed and stored as a class attribute. There actually *are* two different defaults, it's just the question of where do you compute the second one.

jerneju:
Ok, I changed it as you two suggested.

Orange/widgets/utils/itemmodels.py
```diff
@@ -869,8 +869,25 @@ class DomainModel(VariableListModel):
                  ATTRIBUTES)
     PRIMITIVE = (DiscreteVariable, ContinuousVariable)
 
-    def __init__(self, order=SEPARATED, placeholder=None,
+    def __init__(self, order=SEPARATED, separators=True, placeholder=None,
```
janezd: In general, beware of changing the parameter order or inserting arguments within the existing arguments. If anybody constructed a `DomainModel` by calling `DomainModel(DomainModel.Primitive, "(No color)")` and not `DomainModel(DomainModel.Primitive, placeholder="(No color)")`, this change would break his/her code.

However, we seem to have always given these arguments with keywords.

astaric: We could also add a *, before them to make it explicit that they can only be used as keyword arguments.

janezd: I like it. Not just here.

- It makes it easier to manipulate arguments without breaking backward compatibility.
- It forces is to name arguments in calls, which makes the code more readable and prevents making some mistakes we've made in the past.

We should at least encourage adding a `*` in new functions.

PR #2699, opened 2017-10-20
[ENH] Scatter Plot Graph and Scaling can handle metas

jerneju:
##### Issue
Scatter Plot moves primitive meas to attributes.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -55,11 +55,6 @@ class OWMDSGraph(OWScatterPlotGraph):
             self.plot_widget.hideAxis(axis)
         self.plot_widget.setAspectLocked(True, 1)
 
-    def get_size_index(self):
-        if self.attr_size == "Stress":
-            return -2
-        return super().get_size_index()
-
     def compute_sizes(self):
         def scale(a):
             dmin, dmax = np.nanmin(a), np.nanmax(a)
@@ -69,17 +64,16 @@ class OWMDSGraph(OWScatterPlotGraph):
                 return np.zeros_like(a)
 
         self.master.Information.missing_size.clear()
-        size_index = self.get_size_index()
-        if size_index == -1:
+        if self.attr_size is None:
             size_data = np.full((self.n_points,), self.point_width,
                                 dtype=float)
-        elif size_index == -2:
+        elif self.attr_size == "Stress":
             size_data = scale(stress(self.master.embedding, self.master.effective_matrix))
             size_data = self.MinShapeSize + size_data * self.point_width
         else:
             size_data = \
                 self.MinShapeSize + \
-                self.scaled_data[size_index, self.valid_data] * \
+                self.scaled_data.get_column_view(self.attr_size)[0][self.valid_data] * \
                 self.point_width
         nans = np.isnan(size_data)
         if np.any(nans):
@@ -270,11 +264,6 @@ class OWMDS(OWWidget):
 
     def init_attr_values(self):
         domain = self.data and len(self.data) and self.data.domain or None
-        if domain is not None:
-            domain = Domain(
-                attributes=domain.attributes,
-                class_vars=domain.class_vars,
-                metas=tuple(a for a in domain.metas if a.is_primitive()))
         for model in self.models:
             model.set_domain(domain)
         self.graph.attr_color = self.data.domain.class_var if domain else None
@@ -653,21 +642,12 @@ class OWMDS(OWWidget):
         coords = np.vstack((emb_x, emb_y)).T
 
         data = self.data
-
-        primitive_metas = tuple(a for a in data.domain.metas if a.is_primitive())
-        keys = [k for k, a in enumerate(data.domain.metas) if a.is_primitive()]
-        data_metas = data.metas[:, keys].astype(float)
-
-        attributes = self.data.domain.attributes + (self.variable_x, self.variable_y) + \
-                     primitive_metas
+        attributes = data.domain.attributes + (self.variable_x, self.variable_y)
         domain = Domain(attributes=attributes,
-                        class_vars=self.data.domain.class_vars)
-        if data_metas is not None:
-            data_x = (self.data.X, coords, data_metas)
-        else:
-            data_x = (self.data.X, coords)
-        data = Table.from_numpy(domain, X=hstack(data_x),
-                                Y=self.data.Y)
+                        class_vars=data.domain.class_vars,
+                        metas=data.domain.metas)
+        data = Table.from_numpy(domain, X=hstack((data.X, coords)),
+                                Y=data.Y, metas=data.metas)
         subset_data = data[self._subset_mask] if self._subset_mask is not None else None
         self.graph.new_data(data, subset_data=subset_data, new=new)
         self.graph.update_data(self.variable_x, self.variable_y, True)
--- a/Orange/widgets/utils/scaling.py
+++ b/Orange/widgets/utils/scaling.py
@@ -1,5 +1,8 @@
+from itertools import chain
+
 import numpy as np
 
+from Orange.data import Domain
 from Orange.statistics.basic_stats import DomainBasicStats
 from Orange.widgets.settings import Setting
 from Orange.widgets.utils import checksum
@@ -12,12 +15,11 @@ class ScaleData:
 
     def _reset_data(self):
         self.domain = None
-        self.data = None
-        self.original_data = None  # as numpy array
+        self.data = None  # as Orange Table
         self.scaled_data = None  # in [0, 1]
         self.jittered_data = None
         self.attr_values = {}
-        self.domain_data_stat = []
+        self.domain_data_stat = {}
         self.valid_data_array = None
         self.attribute_flip_info = {}  # dictionary with attr: 0/1 if flipped
         self.jitter_seed = 0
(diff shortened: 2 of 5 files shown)


Orange/widgets/utils/scaling.py
```diff
@@ -30,53 +31,59 @@ def rescale_data(self):
...
+        domain = self.domain
+        for attr in domain.attributes + domain.class_vars + domain.metas:
+            c = self.scaled_data.get_column_view(attr)[0]
             if attr.is_discrete:
-                no_jit[index] *= 2
-                no_jit[index] += 1
-                no_jit[index] /= 2 * len(attr.values)
+                c *= 2
+                c += 1
+                c /= 2 * len(attr.values)
```
janezd: Isn't (2x+1)/(2a) same as (x + 0.5)/a? Aren't the above three lines the same as

```
c += 0.5
c /= len(attr.values)
```

After rewriting it this way, I also understand what the code actually does. :) Moves the data into the middle of intervals and normalizes it to [0, 1].

I know this was here before, but only God knows why it was written like this.

Orange/widgets/utils/scaling.py
```diff
@@ -30,53 +31,59 @@ def rescale_data(self):
...
-                  (self.original_data, self.scaled_data, self.valid_data_array))
+                  (self.data, self.scaled_data, self.valid_data_array))
 
     def _compute_jittered_data(self):
         data = self.data
         self.jittered_data = self.scaled_data.copy()
         random = np.random.RandomState(seed=self.jitter_seed)
-        for index, col in enumerate(self.jittered_data):
+        domain = self.domain
+        for attr in domain.attributes + domain.class_vars + domain.metas:
```
janezd: You can use `chain(domain, domain.metas)` in such situations.

Orange/widgets/utils/scaling.py
```diff
@@ -92,8 +99,14 @@ def set_data(self, data, skip_if_same=False, no_data=False):
         if data is None:
             return
 
-        self.domain = data.domain
-        self.data = data
+        domain = data.domain
+        self.primitives = data.domain.attributes + data.domain.class_vars + \
+                          tuple([v for v in data.domain.metas if v.is_primitive()])
```
janezd: You could use `chain(data.domain, (v for v in data.domain.metas if v.is_primitive()))`.

If you keep it as it is: don't use `tuple([...])`: just use generators, `tuple(...)`, without constructing a list first.

Orange/widgets/utils/scaling.py
```diff
@@ -92,8 +99,14 @@ def set_data(self, data, skip_if_same=False, no_data=False):
...
+        new_domain = Domain(attributes=domain.attributes,
+                            class_vars=domain.class_vars,
+                            metas=tuple([v for v in domain.metas if v.is_primitive()]))
```
janezd: Same - no need to construct a list first.

Orange/widgets/utils/scaling.py
```diff
@@ -103,67 +116,71 @@ def set_data(self, data, skip_if_same=False, no_data=False):
...
         if attr.is_continuous:
             self.attr_values[attr] = [-self.attr_values[attr][1],
                                       -self.attr_values[attr][0]]
-
-        self.jittered_data[index] = 1 - self.jittered_data[index]
-        self.scaled_data[index] = 1 - self.scaled_data[index]
+        col = self.jittered_data.get_column_view(attr)[0]
+        col = 1 - col
+        col = self.scaled_data.get_column_view(attr)[0]
+        col = 1 - col
```
janezd: These four lines have no effect, do they?

Did you mean something like this?

```
col *= -1
col += 1
```

Or (shorter, but maybe slower) `col[:] = 1 - col`.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -653,21 +642,12 @@ def _setup_plot(self, new=False):
...
 
         data = self.data
-
-        primitive_metas = tuple(a for a in data.domain.metas if a.is_primitive())
-        keys = [k for k, a in enumerate(data.domain.metas) if a.is_primitive()]
-        data_metas = data.metas[:, keys].astype(float)
-
-        attributes = self.data.domain.attributes + (self.variable_x, self.variable_y) + \
-                     primitive_metas
+        attributes = data.domain.attributes + (self.variable_x, self.variable_y)
```
janezd: @BlazZupan, do we want the columns that contain x and y coordinates from the MDS to appear as ordinary attributes or as metas?

They used to be in metas, but now that visualization widgets can show metas as well, I would tend to store x and y as metas to not pollute that feature set.

Orange/widgets/utils/scaling.py
```diff
@@ -30,53 +31,59 @@ def rescale_data(self):
 
     def _compute_domain_data_stat(self):
         stt = self.domain_data_stat = \
-            getCached(self.data, DomainBasicStats, (self.data,))
-        for index in range(len(self.domain)):
-            attr = self.domain[index]
+            getCached(self.data, DomainBasicStats, (self.data, True))
+        domain = self.domain
+        for attr in domain.attributes + domain.class_vars + domain.metas:
```
janezd: Maybe `chain(domain, domain.metas)`?

Orange/widgets/utils/scaling.py
```diff
@@ -103,67 +116,71 @@ def set_data(self, data, skip_if_same=False, no_data=False):
...
+    def get_valid_list(self, attrs):
         """
         Get array of 0 and 1 of len = len(self.data). If there is a missing
         value at any attribute in indices return 0 for that instance.
         """
         if self.valid_data_array is None or len(self.valid_data_array) == 0:
             return np.array([], np.bool)
+        domain = self.domain
+        indices = []
+        for index, attr in enumerate(domain.attributes + domain.class_vars + domain.metas):
```
janezd: `enumerate(chain(domain, domain.metas))`?

janezd (commented):
This is so much simpler. All widgets should use models for combos/lists of variables.

Orange/widgets/utils/scaling.py
jerneju: To be sincere, I would prefer `domain.all_variables` or something similar. Like we have `domain.variables = domain.attributes + domain.class_vars` we could have something that includes `attributes`, `class_vars` and `metas`.

janezd: I hate this, too. 

When one gets older, (s)he start adding historic notes to everything. So let me say that in the beginning, like 20 years ago, Orange used to have attributes and classes, and both together were called variables. Then I introduced meta attributes, and they didn't mean the same thing as today: meta attributes were (essentially) sparse, while attributes and classes were dense ("essentially", because there was no numpy, these were C++ structures). Since all code - before the introduction of metas - expected that variables were attributes + classes, I would break everything if I added metas there. It also made more sense than today since variables (=attributes + classes) were the dense part of the data.

The idea that any part can be dense or sparse, and that metas are not so different from attributes and so forth evolved gradually even in Orange 3. Hence we have not initially decided that `variables` will be attributes + class + meta. To make things worse, we (most probably: I) extended this bad design to iteration over `Domain`. @kernc (I think) correctly pointed out that since `var in domain` returns `True` also when `var` is in `domain.metas`, so `for var in domain` should iterate over attributes, classes and also metas. So in all the places where you could use `chain(domain, domain.metas)` in your code, just `domain` should suffice instead. Alas, fixing the iteration over `Domain` would break so much code that we gave up. This horrible design decision stays forever, I fear.

I don't like `all_variables`. `variables` should already be *all*, so `all_variables` sounds almost like `really_all_variables`. Plus, the next thing you'll want will be `primitive_variables`. And maybe also `continuous_variables`. I wouldn't like to multiply the number of domains properties.

janezd: For the sake of brainstorming - and if this is indeed necessary - what about having a method `Domain.all(condition)`. If `condition` is a type derived from `Variable`, `all` returns all variables of that type (and derived types). Hence , `Domain.all(DiscreteVariable)` returns all discrete variables. We then add `PrimitiveVariable` as a new common base class for `DiscreteVariable` and `ContinuousVariable`. With that, `Domain.all(PrimitiveVariable)` returns all primitive variables. Finally, the default value for `condition` is `Variable`, so `Domain.all()` returns attributes + class_vars + metas.

Method `all` could also accept other types of conditions if we can think of any.

Just for brainstorming.

Orange/widgets/utils/scaling.py
jerneju: Yes, the latter one can consume even 4 times more time.

Orange/widgets/utils/scaling.py
jerneju: Great idea, I implemented it.

Orange/data/domain.py
```diff
@@ -211,6 +211,9 @@ def variables(self):
     def metas(self):
         return self._metas
 
+    def all(self, condition=Variable):
+        return [v for v in chain(self._variables, self._metas) if isinstance(v, condition)]
```
kernc: `all()` is not good. The idiom means something completely different in Python/NumPy context. `filter()` would be better, `.where()` would be best.

This change should require re-touching _everywhere_ the former (`[]+[]` or `chain`) idiom was used.

kernc (dismissed):

Orange/data/domain.py
jerneju: Ok.

janezd: I don't like `where`. :)

`all` reads well: I want all discrete variables, hence `domain.all(DiscreteVariable)`. But I agree the name is wrong wrt its usual meaning in Python.

`domain.where(DiscreteVariable)` doesn't read so well and it also means two different things in numpy, one of which is similar to this `Domain.where` (but with different arguments) and one which is completely different.

`domain.filter(DiscreteVariable)` is sort of fine, but it gives an impression that it will return a new instance of `Domain` with just discrete variables.

What about `domain.collect(DiscreteVariable)`? Or `domain.pick(DiscreteVariable)` -- though I prefer `collect`?

Orange/data/domain.py
```diff
@@ -211,6 +211,9 @@ def variables(self):
     def metas(self):
         return self._metas
 
+    def where(self, condition=Variable):
```
astaric: Could you do this in a separate PR?
While I agree that this change is neatly hidden here, I would prefer to have modifications to Variables / Domain in a separate, clearly marked PR (where we can discuss different alternatives).

Orange/widgets/utils/scaling.py
```diff
@@ -30,53 +30,58 @@ def rescale_data(self):
 
     def _compute_domain_data_stat(self):
         stt = self.domain_data_stat = \
-            getCached(self.data, DomainBasicStats, (self.data,))
-        for index in range(len(self.domain)):
-            attr = self.domain[index]
+            getCached(self.data, DomainBasicStats, (self.data, True))
+        domain = self.domain
+        for attr in domain.where():
```
astaric: This looks weird.

janezd: 👍

Orange/data/domain.py
janezd: 👍

jerneju:
https://sentry.io/biolab/orange3/issues/307091588/

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -47,9 +47,12 @@ def iterate_states(self, initial_state):
...
         graph = self.master.graph
-        ind12 = [graph.domain.index(self.attrs[x]) for x in state]
-        valid = graph.get_valid_list(ind12)
-        X = graph.jittered_data[ind12, :][:, valid].T
+        attrs = [self.attrs[x] for x in state]
+        valid = graph.get_valid_list(attrs)
+        cols = []
+        for var in attrs:
+            cols.append(graph.jittered_data.get_column_view(var)[0][valid])
+        X = np.vstack(cols).T
```
lanzagar: `X = np.column_stack(cols)`

Orange/widgets/utils/scaling.py
```diff
@@ -30,53 +32,58 @@ def rescale_data(self):
...
-        self.original_data = np.hstack((data.X, Y)).T
-        self.scaled_data = no_jit = self.original_data.copy()
-        self.valid_data_array = np.isfinite(no_jit)
-        for index in range(len(data.domain)):
-            attr = data.domain[index]
+        if np.any(data.metas):
+            all_data = (data.X, Y, data.metas)
+        else:
+            all_data = (data.X, Y)
+        all_data = np.hstack(all_data).T
```
lanzagar: This can have dtype==object when stacking with metas which will cause np.isfinite to fail 2 lines below

lanzagar (changes requested):

Orange/widgets/unsupervised/owmds.py
lanzagar: Actually this line here is not related to that (it's for constructing data for plotting).
Output data has new coordinates in metas as we want, see commit function below (L671)

Orange/widgets/utils/scaling.py
jerneju: Thanks. Fixed.

jerneju:
https://sentry.io/biolab/orange3/issues/400459027/

Orange/widgets/utils/scaling.py
lanzagar: Fixed here, but I get similar errors elsewhere (see sentry report).
Try fixing all metas dtype problems (converting all primitive metas to float might be enough, maybe something else is also needed). Then please test thoroughly! Use at least the workflow from the sentry report.

jerneju: I found another error caused by Corpus (issue: https://sentry.io/biolab/orange3-text-bl/issues/399242697/ )

lanzagar (approved):

Orange/data/domain.py
lennee: Addressed in 44c14bb97. Expanded the TODO on `Provider.kt:11-18` to explicitly warn against flipping `supported=true` without first wiring `isTdsFirstPartyDeliveryEnabled` into the dispatch/quote/validate entry points (TDS-2421), and noted that the LD kill-switch constant exists (`LaunchDarklyFlags.TDS_FIRST_PARTY_DELIVERY_ENABLED`) but has no runtime call site yet — so the enum flag is the only active guard until TDS-2421 lands.

PR #2706, opened 2017-10-24
[RFC] Domain: return all variables

jerneju:
##### Issue
Due to historical background Orange Data Mining has issues:

1. `domain.variables` are attributes and classes but not metas
2. There is no such thing as `domain.something` where that would be attributes, classes, and metas.
3. Yes, `some_meta in domain` may return `True`, but when one would iterate over `domain` it does not iterate over metas.
4. Iteration over attributes, classes, and metas or over all primitive features including metas would be useful in many widgets.

See also:
https://github.com/biolab/orange3/pull/2699#discussion_r146087213
https://github.com/biolab/orange3/pull/2699#discussion_r146343078

##### Description of changes

- PrimitiveVariable (created only to be used for separating primitive and non primitive variables).

- Domain method which returns list of all features. This method can be called with an argument which specifies which features to return (all by default).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/domain.py
+++ b/Orange/data/domain.py
@@ -211,6 +211,9 @@ class Domain:
     def metas(self):
         return self._metas
 
+    def filter(self, pred):
+        return [v for v in chain(self, self.metas) if pred(v)]
+
     def __len__(self):
         """The number of variables (features and class attributes)."""
         return len(self._variables)
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -12,7 +12,7 @@ from Orange.data import _variable
 from Orange.util import Registry, color_to_hex, hex_to_color, Reprable
 
 __all__ = ["Unknown", "MISSING_VALUES", "make_variable", "is_discrete_values",
-           "Value", "Variable", "ContinuousVariable", "DiscreteVariable",
+           "Value", "Variable", "PrimitiveVariable", "ContinuousVariable", "DiscreteVariable",
            "StringVariable", "TimeVariable"]
 
 
@@ -348,7 +348,7 @@ class Variable(Reprable, metaclass=VariableMeta):
         `True` if the variable's values are stored as floats.
         Non-primitive variables can appear in the data only as meta attributes.
         """
-        return issubclass(cls, (DiscreteVariable, ContinuousVariable))
+        return issubclass(cls, PrimitiveVariable)
 
     @property
     def is_discrete(self):
@@ -439,7 +439,11 @@ class Variable(Reprable, metaclass=VariableMeta):
         return var
 
 
-class ContinuousVariable(Variable):
+class PrimitiveVariable(Variable):
+    TYPE_HEADERS = ()
+
+
+class ContinuousVariable(PrimitiveVariable):
     """
     Descriptor for continuous variables.
 
@@ -536,7 +540,7 @@ class ContinuousVariable(Variable):
         return var
 
 
-class DiscreteVariable(Variable):
+class DiscreteVariable(PrimitiveVariable):
     """
     Descriptor for symbolic, discrete variables. Values of discrete variables
     are stored as floats; the numbers corresponds to indices in the list of


astaric:
Great. Now please update the description of the pull request with the description of the problem (what are we trying to solve) and the proposed way to fix it. If there are alternative ways we could do this, list those as well so we can discuss and find out which of them should be implemented.

astaric:
Which of the issues does this PR solve? If I understand correctly, problems 1. and 3. remain even after the fix.

Furthermore, I do not like the addition of a new "base class". Variable already has method `is_primitive`. Why not use that?

Are there any other ways this could be solved? What are their (dis)advantages? Why should use the fix in this PR and not one of the alternative approaches?

astaric:
domain.all(Variable.is_primitive)?

janezd:
```
class Domain:
    ...
    def all(self, pred):
       return [var for var in chain(self, self.metas) if pred(var)]
```

Interesting. And it can be called `all`.

astaric:
I would still like to know what are the alternatives. The correct way IMHO would be to fix __iter__.

kernc:
👍² for fixing `__iter__`.

The predicate nature of input parameter is precisely why `.filter()` or `.where()` could work.

👎 for `.all()`, which has a well-established meaning throughout.

janezd:
Fixing `__iter__`: I enthusiastically agree, but it will be difficult to find all places that need to be adapted.

`all`: sorry, I meant `filter`. If `__iter__` is fixed, there may be no need for a `filter` method. If `__iter__` stays broken, `filter` method would help.

jerneju:
Yes, fixing `__iter__` would be great. @janezd ?

janezd:
> Yes, fixing __iter__ would be great. @janezd?

Me? No chance. Too risky. I've no idea how to scan through all current iterations through `Domain`. Any place I miss would add metas in a place where only variables and classes are expected, possibly adding non-primitive attributes to a method that deals only with primitive ones.

Besides, add-ons do this, too.

The first step towards this would be adding a warning to the existing `Domain.__iter__` and leaving it there for half a year.

jerneju:
I propose:

1. Adding `filter` method.
2. Changing iteration over `domain` to iteration over `domain.variables` wherever we can find these occurrences.
3. Adding warning to the `__iter__` method.
4. Changing `__iter__` someday in the future.

astaric:
I am not too fond of adding methods we plan to remove in half a year (once `__iter__` starts working the way it is supposed to).

astaric:
@janezd: #2714 replaces all (tested) iterations over domain with iteration over domain.variables

janezd:
I still like `filter`. It's not a temporary solution: `domain.filter(DiscreteVariable)` is so much shorter and simpler than `(var for var in domain if isinstance(var, DiscreteVariable)`.

I don't mind if it's called `domain.get_variables_by_type(DiscreteVariable)` (OK, this is stupid), but I think a function like this would be generally useful.

kernc:
`for var in domain if var.is_discrete`.

To be more generally useful, `.filter()` would need to accept a predicate.

janezd:
`filter(Variable.is_discrete, domain)`? Not too happy, but OK. We close this PR?

kernc:
Why aren't you happy? What's wrong with `for var in domain if var.is_discrete`?

janezd:
You mean `(var for var in domain if var.is_discrete)`?  `domain.filter(DiscreteVariable)` is shorter and more readable (under reasonable assumptions that one can easily guess what `Domain.filter` does).

But I'm OK with (builtin) `filter(Variable.is_discrete, domain)` if you don't want to multiply the number of `Domain`'s methods.

astaric:
I really really really dislike the artificial `PrimitiveVariable` base class. Does `domain.filter` make sense without it?

kernc:
We already have `Variable.is_primitive` ...

janezd:
> I really really really dislike the artificial PrimitiveVariable base class. Does domain.filter make sense without it?

Not much. But it's not so artificial: Variable can be either PrimitiveVariable or NonPrimitiveVariable. Primitive can be Discrete or Continuous. There is a hierarchy. There are just no specific common methods for PrimitiveVariable. Think about PrimitiveVariable as an ABC.

> We already have Variable.is_primitive

... which can be used as a predicate if `filter` also supports predicates -- which it should, anyway.

janezd:
Assuming we will some day fix `Domain.__iter__`, we will be able to use `filter(Variable.is_primitive, domain)`.

With this PR we can already have `domain.filter(Variable.is_primitive)` or `domain.filter(PrimitiveVariable)`.

Without any obvious advantage of the latter over the former, I'm applying Occam here.

Issue #2708, opened 2017-10-24
Saving Report to PDF Crashes


nikicc:
##### Orange version

3.7.0.dev0+445176c

##### Expected behavior
One can save the report to PDF.

##### Actual behavior
Orange crashes with `AttributeError: 'WebviewWidget' object has no attribute 'print_'`.

##### Steps to reproduce the behavior
![screen shot 2017-10-24 at 11 55 52](https://user-images.githubusercontent.com/713026/31936594-4fab86e8-b8b2-11e7-9bf4-2e554d0c6693.png)

jerneju:
What happens when you write this in a Python console:
`from Orange.widgets.utils.webview import WebviewWidget`

nikicc:
```
>>> from Orange.widgets.utils.webview import WebviewWidget
>>> from AnyQt.QtWebKitWidgets import QWebView
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/Niko/PyCharmProjects/virtual-envs/orange3-ve/lib/python3.6/site-packages/AnyQt/QtWebKitWidgets.py", line 14, in <module>
    from PyQt5.QtWebKitWidgets import *
ModuleNotFoundError: No module named 'PyQt5.QtWebKitWidgets'
>>> 
```

PR #2710, opened 2017-10-24
OWReport: fix printing to PDF on WebEngine

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/2708

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/report/owreport.py
+++ b/Orange/canvas/report/owreport.py
@@ -1,5 +1,6 @@
 import os
 import logging
+from collections import OrderedDict
 
 import pkg_resources
 import pickle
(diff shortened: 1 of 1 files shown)


nikicc:
This solves the issue if I click on the Save button. However, if I click on Print and then select save to PDF it still crashes with: 
```
Traceback (most recent call last):  File ".../orange3/Orange/canvas/report/owreport.py", line 357, 
in _print_report self.report_view.print_(printer)
AttributeError: 'WebviewWidget' object has no attribute 'print_'
```

Could this be fixed as well?

astaric:
I get this:

```
Error encountered in widget Report:

  File "/Users/anze/dev/orange3/Orange/canvas/report/owreport.py", line 329, in save_report
    self.report_view.page().printToPdf(filename)
AttributeError: 'QWebEnginePage' object has no attribute 'printToPdf'

During handling of the above exception, another exception occurred:

  File "/Users/anze/dev/orange3/Orange/canvas/report/owreport.py", line 336, in save_report
    self.report_view.print_(printer)
AttributeError: 'WebviewWidget' object has no attribute 'print_'
```

astaric:
This no longer crashes on Qt5.6 with webengine, but the results is not what I would expect. What gets printed is what is currently displayed in the report window, if the content is longer than the window, the image in pdf contains a scrollbar.

astaric:
@ales-erjavec any idea how to solve it on qt 5.6?

astaric:
As there are no obvious solutions for Qt5.6, I am merging this as-is. The printing of reports to pdf does not work for any Qt5 that uses web engine and this fix at least fixes the problem for Qt5.7+.

PR #2714, opened 2017-10-26
domain: Add deprecation warning in __iter__

astaric:
The current behaviour is confusing, as `x in domain` returns True for meta variables, but iter(domain) does not yield them.

This PR adds a deprecation warning, points users to iterate over domain.variables instead.

### DIFF
--- a/Orange/data/sql/table.py
+++ b/Orange/data/sql/table.py
@@ -355,9 +355,9 @@ class SqlTable(Table):
         if columns is not None:
             columns = [self.domain[col] for col in columns]
         else:
-            columns = list(self.domain)
+            columns = self.domain.variables
             if include_metas:
-                columns += list(self.domain.metas)
+                columns += self.domain.metas
         return self._get_stats(columns)
 
     def _get_stats(self, columns):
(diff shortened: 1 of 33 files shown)


kernc:
I dared push something that might pass the tests.

Orange/data/sql/table.py
```diff
@@ -355,7 +355,7 @@ def _compute_basic_stats(self, columns=None,
         if columns is not None:
             columns = [self.domain[col] for col in columns]
         else:
-            columns = list(self.domain)
+            columns = list(self.domain.variables)
```
janezd: Remove casting to `list` here and below?

Orange/data/sql/table.py
```diff
@@ -387,7 +387,7 @@ def _compute_distributions(self, columns=None):
...
```
janezd: And here?

Orange/data/sql/table.py
astaric: wouldn't this break `columns += list(self.domain.metas)` in line 360?

astaric:
Can anyone help me find a change lint is complaining about?

Orange/data/sql/table.py
janezd: I wrote "here and below". :)

jerneju:
@astaric:
> [W0612(unused-variable), TestDomainInit.test_iter] Unused variable 'w'
https://github.com/biolab/orange3/pull/2714/files#diff-5c5b01d8ff0e8c82d64276be2384cec8R270

PR #2715, opened 2017-10-27
Fix PythonScript old schema compatibility

markotoplak:
##### Issue
Old schemas did not open. Someone forgot that "replaces" should be a list.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


jerneju:
Added 3.5.0 workflow with Python Script.

PR #2720, opened 2017-10-27
Variable: Allow using is_* as predicates

### LINKED ISSUE #2719: Make is_discrete, is_continuous (also) useful as predicates
In #2706 we assumed that `filter(Variable.is_discrete, domain)` would return discrete variables from `domain`. It doesn't.

`is_discrete` and `is_continuous` are currently instance properties that return `bool` and as such cannot be used as predicates. To be used as predicates, they would need to be class properties resolving to a function. Python's `Descriptors` actually foresaw [this dual functionality](https://docs.python.org/2/reference/datamodel.html#object.__get__). Implementation would be as follows. 

```
class _IndicatorDescriptor:
    def __init__(self, atype):
        self.check_type = lambda instance: isinstance(instance, atype)

    def __get__(self, instance, owner):
        if instance is not None:
            return self.check_type(instance)
        else:
            return self.check_type

Variable.is_discrete = _IndicatorDescriptor(DiscreteVariable)
Variable.is_continuous = _IndicatorDescriptor(ContinuousVariable)
Variable.is_string = _IndicatorDescriptor(StringVariable)
Variable.is_time = _IndicatorDescriptor(TimeVariable)
```

This requires that `DiscreteVariable` and `ContinuousVariable` are defined before `Variable`, which is impossible, so properties are inserted into `Variable` class at the end of the module. (Needless to say, `Variable` could have a metaclass that would ensure that any derived class would push its corresponding property into the base class `Variable`. Needless to hope, @astaric won't like it. :))

This now allows

```
>>> v = ContinuousVariable("x")
>>> v.is_continuous  # current functionality
True
>>> Variable.is_continuous(v)  # needed for filter predicates
True
>>> v.is_discrete
False
>>> Variable.is_discrete(v)
False
```

This doesn't work for `is_primitive`, which is a class method. A similar trick should work there, however.

Please comment before I continue with `is_primitive`, add tests etc.

janezd:
##### Issue

Fixes #2719.

##### Description of changes

Implement `is_discrete` etc. as descriptors for instance and class attribute access to allow using them in e.g. `filter(Variable.is_discrete, domain)`.

Add an optional argument to `is_primitive`. `is_primitive` is needed a different treatment because it's a class method and not a property.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -343,28 +343,13 @@ class Variable(Reprable, metaclass=VariableMeta):
             cls._clear_cache()
 
     @classmethod
-    def is_primitive(cls):
+    def is_primitive(cls, var=None):
         """
         `True` if the variable's values are stored as floats.
         Non-primitive variables can appear in the data only as meta attributes.
         """
-        return issubclass(cls, (DiscreteVariable, ContinuousVariable))
-
-    @property
-    def is_discrete(self):
-        return isinstance(self, DiscreteVariable)
-
-    @property
-    def is_continuous(self):
-        return isinstance(self, ContinuousVariable)
-
-    @property
-    def is_string(self):
-        return isinstance(self, StringVariable)
-
-    @property
-    def is_time(self):
-        return isinstance(self, TimeVariable)
+        to_check = cls if var is None else type(var)
+        return issubclass(to_check, (DiscreteVariable, ContinuousVariable))
 
     def repr_val(self, val):
         """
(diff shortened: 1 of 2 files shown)


Orange/data/variable.py
```diff
@@ -1019,3 +1004,19 @@ def to_val(self, s):
             return self.parse(s)
         else:
             return super().to_val(s)
+
+
+class _TypeIndicatorDescriptor:
+    def __init__(self, atype):
+        self.check_type = lambda instance: isinstance(instance, atype)
```
kernc: Make a named function, and a __private one.

Orange/data/variable.py
```diff
@@ -1019,3 +1004,19 @@ def to_val(self, s):
             return self.parse(s)
         else:
             return super().to_val(s)
+
+
+class _TypeIndicatorDescriptor:
```
kernc: Docstring why we have this.

janezd: Docstring would indeed help. Done.

Orange/data/variable.py
janezd: PEP 8 recommends `def f(x): return 2 * x` over `f = lambda x: 2 * x`. This doesn't really apply here since I can't assign the proper name to the function (`def self.check_type(instance): ...` ?! :)). But I agree that it's clearer with a named function due to lambda's awful syntax.

kernc: Lambdas are perfectly clean. My only concern was they are not picklable.

janezd: Turning it into a (local) function doesn't help. But this is not a problem here, since this `self` is not an instance of `Variable` but of `_TypeIndicatorDescriptor`, which is a part of the `Variable` class and is created when importing the module. 

That is, you indeed can't pickle `_TypeIndicatorDescriptor` (precisely because of `check_type`),

```
>>> pickle.dumps(DiscreteVariable.is_discrete)
AttributeError: Can't pickle local object '_TypeIndicatorDescriptor.__init__.<locals>.__check_type'
```

but `_TypeIndicatorDescriptor` is never pickled anyway. `Variable` class is still picklable, and hence are its instances.

```
>>> pickle.dumps(DiscreteVariable)
b'\x80\x03cOrange.data.variable\nDiscreteVariable\nq\x00.'
```

PR #2721, opened 2017-10-30
[RFC] [WIP] [ENH] DomainModel instead of VariableListModel

jerneju:
##### Issue
`DomainModel` should be used anywhere possible.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owdiscretize.py
+++ b/Orange/widgets/data/owdiscretize.py
@@ -41,8 +41,7 @@ _dispatch = {
     EqualFreq: lambda m, data, var: disc.EqualFreq(m.k)(data, var),
     EqualWidth: lambda m, data, var: disc.EqualWidth(m.k)(data, var),
     Remove: lambda m, data, var: None,
-    Custom: lambda m, data, var:
-        disc.Discretizer.create_discretized_var(var, m.points)
+    Custom: lambda m, data, var: disc.Discretizer.create_discretized_var(var, m.points)
 }
 
 
@@ -261,7 +260,6 @@ class OWDiscretize(widget.OWWidget):
                  if var.is_continuous]
         self.varmodel[:] = cvars
 
-        class_var = data.domain.class_var
         has_disc_class = data.domain.has_discrete_class
 
         self.default_bbox.buttons[self.MDL - 1].setEnabled(has_disc_class)
--- a/Orange/widgets/data/owdiscretize.py
+++ b/Orange/widgets/data/owdiscretize.py
@@ -7,9 +7,11 @@ from AnyQt.QtCore import Qt
 
 import Orange.data
 import Orange.preprocess.discretize as disc
+from Orange.data import ContinuousVariable
 
 from Orange.widgets import widget, gui, settings
-from Orange.widgets.utils import itemmodels, vartype
+from Orange.widgets.utils import vartype
+from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.widget import Input, Output
 
 __all__ = ["OWDiscretize"]
@@ -204,7 +206,8 @@ class OWDiscretize(widget.OWWidget):
         # List view with all attributes
         self.varview = QListView(selectionMode=QListView.ExtendedSelection)
         self.varview.setItemDelegate(DiscDelegate())
-        self.varmodel = itemmodels.VariableListModel()
+        self.varmodel = DomainModel(order=(DomainModel.CLASSES, DomainModel.ATTRIBUTES),
+                                    valid_types=ContinuousVariable)
         self.varview.setModel(self.varmodel)
         self.varview.selectionModel().selectionChanged.connect(
             self._var_selection_changed
@@ -256,9 +259,7 @@ class OWDiscretize(widget.OWWidget):
     def _initialize(self, data):
         # Initialize the default variable states for new data.
         self.class_var = data.domain.class_var
-        cvars = [var for var in data.domain.variables
-                 if var.is_continuous]
-        self.varmodel[:] = cvars
+        self.varmodel.set_domain(data.domain)
 
         has_disc_class = data.domain.has_discrete_class
 
@@ -300,7 +301,7 @@ class OWDiscretize(widget.OWWidget):
 
     def _clear(self):
         self.data = None
-        self.varmodel[:] = []
+        self.varmodel.set_domain(None)
         self.var_state = {}
         self.saved_var_states = {}
         self.default_bbox.buttons[self.MDL - 1].setEnabled(True)
--- a/Orange/widgets/unsupervised/owdistancemap.py
+++ b/Orange/widgets/unsupervised/owdistancemap.py
@@ -3,11 +3,10 @@ import itertools
 from functools import reduce
 from operator import iadd
 
-import numpy
+import numpy as np
 
 from AnyQt.QtWidgets import (
-    QFormLayout, QGraphicsRectItem, QGraphicsGridLayout, QGraphicsWidget,
-    QApplication
+    QFormLayout, QGraphicsRectItem, QGraphicsGridLayout, QApplication
 )
 from AnyQt.QtGui import (
     QFontMetrics, QPen, QIcon, QPixmap, QLinearGradient, QPainter, QColor,
@@ -28,7 +27,8 @@ from Orange.widgets.utils import itemmodels, colorbrewer
 from Orange.widgets.utils.annotated_data import (create_annotated_table,
                                                  ANNOTATED_DATA_SIGNAL_NAME)
 from Orange.widgets.widget import Input, Output
-from .owhierarchicalclustering import DendrogramWidget, GraphicsSimpleTextList
+from Orange.widgets.unsupervised.owhierarchicalclustering import DendrogramWidget,\
+    GraphicsSimpleTextList
 
 def _remove_item(item):
     item.setParentItem(None)
@@ -181,7 +181,7 @@ class DistanceMapItem(pg.ImageItem):
             return -1, -1
         else:
             h, w = self.image.shape
-            i, j = numpy.floor([pos.y(), pos.x()])
+            i, j = np.floor([pos.y(), pos.x()])
             if 0 <= i < h and 0 <= j < w:
                 return int(i), int(j)
             else:
@@ -193,9 +193,9 @@ class DistanceMapItem(pg.ImageItem):
             return -1, -1
         else:
             h, w = self.image.shape
-            i, j = numpy.floor([pos.y(), pos.x()])
-            i = numpy.clip(i, 0, h - 1)
-            j = numpy.clip(j, 0, w - 1)
+            i, j = np.floor([pos.y(), pos.x()])
+            i = np.clip(i, 0, h - 1)
+            j = np.clip(j, 0, w - 1)
             return int(i), int(j)
 
     def __clearSelections(self):
@@ -552,10 +552,10 @@ class OWDistanceMap(widget.OWWidget):
                 tree = self._ordered_cluster_tree()
 
             leaves = hierarchical.leaves(tree)
-            indices = numpy.array([leaf.value.index for leaf in leaves])
+            indices = np.array([leaf.value.index for leaf in leaves])
             X = self.matrix
-            self._sorted_matrix = X[indices[:, numpy.newaxis],
-                                    indices[numpy.newaxis, :]]
+            self._sorted_matrix = X[indices[:, np.newaxis],
+                                    indices[np.newaxis, :]]
             self._sort_indices = indices
 
     def _invalidate_annotations(self):
@@ -568,8 +568,8 @@ class OWDistanceMap(widget.OWWidget):
         elif self.annotation_idx == 1:  # Enumeration
             labels = [str(i + 1) for i in range(self.matrix.shape[0])]
         elif self.annot_combo.model()[self.annotation_idx] == "Attribute names":
-                attr = self.matrix.row_items.domain.attributes
-                labels = [str(attr[i]) for i in range(self.matrix.shape[0])]
+            attr = self.matrix.row_items.domain.attributes
+            labels = [str(attr[i]) for i in range(self.matrix.shape[0])]
         elif self.annotation_idx == 2 and \
                 isinstance(self.items, widget.AttributeList):
             labels = [v.name for v in self.items]
@@ -597,17 +597,17 @@ class OWDistanceMap(widget.OWWidget):
 
     def _update_color(self):
         if self.matrix_item:
-            name, colors = self.palettes[self.colormap]
+            _, colors = self.palettes[self.colormap]
             n, colors = max(colors.items())
-            colors = numpy.array(colors, dtype=numpy.ubyte)
+            colors = np.array(colors, dtype=np.ubyte)
             low, high = self.color_low * 255, self.color_high * 255
-            points = numpy.linspace(low, high, n)
-            space = numpy.linspace(0, 255, 255)
+            points = np.linspace(low, high, n)
+            space = np.linspace(0, 255, 255)
 
-            r = numpy.interp(space, points, colors[:, 0], left=255, right=0)
-            g = numpy.interp(space, points, colors[:, 1], left=255, right=0)
-            b = numpy.interp(space, points, colors[:, 2], left=255, right=0)
-            colortable = numpy.c_[r, g, b]
+            r = np.interp(space, points, colors[:, 0], left=255, right=0)
+            g = np.interp(space, points, colors[:, 1], left=255, right=0)
+            b = np.interp(space, points, colors[:, 2], left=255, right=0)
+            colortable = np.c_[r, g, b]
             self.matrix_item.setLookupTable(colortable)
 
     def _invalidate_selection(self):
@@ -707,7 +707,7 @@ class TextList(GraphicsSimpleTextList):
 
 def palette_gradient(colors, discrete=False):
     n = len(colors)
-    stops = numpy.linspace(0.0, 1.0, n, endpoint=True)
+    stops = np.linspace(0.0, 1.0, n, endpoint=True)
     gradstops = [(float(stop), color) for stop, color in zip(stops, colors)]
     grad = QLinearGradient(QPointF(0, 0), QPointF(1, 0))
     grad.setStops(gradstops)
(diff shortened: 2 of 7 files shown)


Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -173,8 +176,9 @@ def set_data(self, data):
         warning_msg = ""
         candidatevars = []
         if data is not None:
+            domain = data.domain
             candidatevars = [
-                v for v in data.domain.variables + data.domain.metas
+                v for v in itertools.chain(domain.class_vars, domain.metas, domain.attributes)
                 if v.is_discrete and len(v.values) >= 2]
```
jerneju: @janezd

Orange/widgets/data/owdiscretize.py
```diff
@@ -291,7 +289,7 @@ def _reset(self):
         # restore the individual variable settings back to defaults.
         def_method = self._current_default_method()
         self.var_state = {}
-        for i in range(len(self.varmodel)):
+        for i, _ in enumerate(self.varmodel):
```
janezd: Why not `range(len(...))`? I think it was better. If lint complains, can we fix it elsewhere (or silence it)?

Orange/widgets/unsupervised/owdistancemap.py
```diff
@@ -333,8 +341,8 @@ def __init__(self):
         self.annot_combo = gui.comboBox(
             self.controlArea, self, "annotation_idx", box="Annotations",
             callback=self._invalidate_annotations, contentsLength=12)
-        self.annot_combo.setModel(itemmodels.VariableListModel())
-        self.annot_combo.model()[:] = ["None", "Enumeration"]
+        self.order = (self.Names[self.Enumeration], ) + DomainModel.SEPARATED
```
janezd: Why is this particular order stored and others are not?

Maybe it would be better if you constructed all models in `__init__` and the just use the correct one. Even better, those that are not domain models but have just fixed strings can be class attributes.

Orange/widgets/unsupervised/owdistancemap.py
```diff
@@ -441,21 +449,22 @@ def set_distances(self, matrix):
 
     def set_items(self, items, axis=1):
         self.items = items
-        model = self.annot_combo.model()
+        names = self.Names
         if items is None:
-            model[:] = ["None", "Enumeration"]
+            model = DomainModel(order=(names[self.Enumeration], ))
```
janezd: Why is this a domain model? It doesn't have any variables, let alone domain, does it?

Orange/widgets/unsupervised/owdistancemap.py
```diff
@@ -441,21 +449,22 @@ def set_distances(self, matrix):
...
         elif not axis:
-            model[:] = ["None", "Enumeration", "Attribute names"]
+            model = DomainModel(order=(names[self.Enumeration], names[self.Attr_names]))
```
janezd: Same.

Orange/widgets/unsupervised/owdistancemap.py
```diff
@@ -441,21 +449,22 @@ def set_distances(self, matrix):
...
-            annot_vars = list(filter_visible(items.domain.variables)) + list(items.domain.metas)
-            model[:] = ["None", "Enumeration"] + annot_vars
+            model = DomainModel(order=self.order)
+            model.set_domain(items.domain)
             self.annotation_idx = 0
             self.openContext(items.domain)
         elif isinstance(items, list) and \
                 all(isinstance(item, Orange.data.Variable) for item in items):
-            model[:] = ["None", "Enumeration", "Name"]
+            model = DomainModel(order=(names[self.Enumeration], names[self.Name]))
```
janezd: Same.

Orange/widgets/unsupervised/owdistancemap.py
```diff
@@ -441,21 +449,22 @@ def set_distances(self, matrix):
...
         else:
-            model[:] = ["None", "Enumeration"]
+            model = DomainModel(order=(names[self.Enumeration], ))
```
janezd: Same.

Orange/widgets/unsupervised/owdistancemap.py
jerneju: Because `self.order` is used later in the code.

Orange/widgets/data/oweditdomain.py
```diff
@@ -556,12 +557,12 @@ def commit(self):
...
-                n_attrs = len(input_domain.attributes)
+                n_metas = len(input_domain.metas)
                 n_class_vars = len(input_domain.class_vars)
                 all_new_vars = list(self.domain_model)
-                attrs = all_new_vars[: n_attrs]
-                class_vars = all_new_vars[n_attrs: n_attrs + n_class_vars]
-                new_metas = all_new_vars[n_attrs + n_class_vars:]
+                attrs = all_new_vars[n_class_vars + n_metas : ]
+                class_vars = all_new_vars[: n_class_vars]
+                new_metas = all_new_vars[n_class_vars: n_class_vars + n_metas]
```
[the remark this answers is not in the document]
janezd: Ah, OK, I see.

Orange/widgets/unsupervised/owdistancemap.py
janezd: So could be other orders. This one is not special.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -14,33 +14,34 @@
...
+from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.widget import Input
 from Orange.widgets.visualize.owlinearprojection import LegendItem, ScatterPlotItem
 
 from Orange.widgets.visualize.owscatterplotgraph import HelpEventDelegate
 
 
-def selected_index(view):
-    """Return the selected integer `index` (row) in the view.
+def selected_variable(model, view):
```
janezd: The first argument is not necessary,

Orange/widgets/visualize/owdistributions.py
```diff
@@ -14,33 +14,34 @@
...
 
     `view` must be in single selection mode.
     """
     indices = view.selectedIndexes()
     assert len(indices) < 2, "View must be in single selection mode"
     if indices:
-        return indices[0].row()
-    else:
-        return -1
+        return model.data(indices[0], Qt.EditRole)
```
janezd: since you can change this to `return indices[0].data(Qt.EditRole)`.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -91,9 +92,9 @@ class Inputs:
     settingsHandler = settings.DomainContextHandler(
         match_values=settings.DomainContextHandler.MATCH_VALUES_ALL)
     #: Selected variable index
-    variable_idx = settings.ContextSetting(-1)
+    var = settings.ContextSetting(None)
     #: Selected group variable
-    groupvar_idx = settings.ContextSetting(0)
+    cvar = settings.ContextSetting(None)
```
janezd: While you're refactoring, could you also rename `cvar` to `group_var`?

janezd:
@lanzagar, when is the deadline for this to be reviewed and merged? I can do one final review tomorrow, but on Friday I have unavoidable meetings and won't be around.

Orange/widgets/visualize/owdistributions.py
jerneju: Fixed.

lanzagar:
Tomorrow is much better than Friday. Today is even better ;)
(the release was planned for Friday I think, maybe even pushed to Thursday if possible)

jerneju:
@janezd : Your commit was unnecessary - I just dropped my commit which had renamed that variable / attribute.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -145,15 +145,16 @@ def __init__(self):
...
 
         box = gui.vBox(self.controlArea, "Group by")
         self.icons = gui.attributeIconDict
         self.groupvarview = gui.comboBox(
-            box, self, "groupvar_idx",
-            callback=self._on_groupvar_idx_changed,
-            valueType=str, contentsLength=12)
+            box, self, "group_var",
+            callback=self._on_groupvar_changed,
+            valueType=str, contentsLength=12,
```
janezd: `valueType` and `sendSelectedValue` are ignored when using a model.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -227,34 +228,25 @@ def set_data(self, data):
...
-                    self.groupvarmodel[1:].index(domain.class_var) + 1
+                self.group_var = domain.class_var
             self.openContext(domain)
-            self.variable_idx = min(max(self.variable_idx, 0),
-                                    len(self.varmodel) - 1)
-            self.groupvar_idx = min(max(self.groupvar_idx, 0),
-                                    len(self.groupvarmodel) - 1)
-            itemmodels.select_row(self.varview, self.variable_idx)
+            self.group_var = None if self.group_var not in self.groupvarmodel else self.group_var
+            self.var = self.var if self.var in self.varmodel else None
```
janezd: A call to `self.clear` sets `self.var` to `None`. This line keeps it `None` -> nothing is selected when the widget receives new data.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -227,34 +228,25 @@ def set_data(self, data):
...
+            self.group_var = self.group_var if self.group_var in self.groupvarmodel else None
```
janezd: This line is repeated (with a reversed condition) from above.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -114,11 +115,10 @@ def __init__(self):
 
         self.distributions = None
         self.contingencies = None
-        self.var = self.cvar = None
         varbox = gui.vBox(self.controlArea, "Variable")
 
-        self.varmodel = itemmodels.VariableListModel()
-        self.groupvarmodel = []
+        self.varmodel = DomainModel(separators=False, valid_types=DomainModel.PRIMITIVE)
```
janezd: This is one case where we don't want the class to be at the beginning; we use class for grouping (by default).

Orange/widgets/visualize/owdistributions.py
```diff
@@ -227,34 +228,25 @@ def set_data(self, data):
...
             self.openContext(domain)
-            self.variable_idx = min(max(self.variable_idx, 0),
-                                    len(self.varmodel) - 1)
-            self.groupvar_idx = min(max(self.groupvar_idx, 0),
-                                    len(self.groupvarmodel) - 1)
-            itemmodels.select_row(self.varview, self.variable_idx)
+            self.group_var = None if self.group_var not in self.groupvarmodel else self.group_var
+            self.var = self.var if self.var in self.varmodel else None
+            self.group_var = self.group_var if self.group_var in self.groupvarmodel else None
+            itemmodels.select_row(self.varview, self.var)
```
janezd: AFAIK, the second argument for `select_row` must be an `int`. Anyway, when the data is loaded, a visualization is shown, but nothing is selected in the listview - I suspect that because of this function.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -91,9 +92,9 @@ class Inputs:
     settingsHandler = settings.DomainContextHandler(
         match_values=settings.DomainContextHandler.MATCH_VALUES_ALL)
     #: Selected variable index
-    variable_idx = settings.ContextSetting(-1)
+    var = settings.ContextSetting(None)
     #: Selected group variable
-    groupvar_idx = settings.ContextSetting(0)
+    group_var = settings.ContextSetting(None)
```
janezd: Settings migrations are missing. Any saved workflows will stop workflowing.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -577,18 +559,17 @@ def onDeleteWidget(self):
         super().onDeleteWidget()
 
     def get_widget_name_extension(self):
-        if self.variable_idx >= 0:
-            return self.varmodel[self.variable_idx]
+        if self.var is not None:
```
janezd: If this condition is `False`, the method returns `None`, which incidentally equals `self.var`.

Why is the condition necessary, then?

janezd:
@jerneju, this is from Sieve. Distributions probably needs something similar.

    @classmethod
    def migrate_context(cls, context, version):
        if not version:
            settings.rename_setting(context, "attrX", "attr_x")
            settings.rename_setting(context, "attrY", "attr_y")
            settings.migrate_str_to_variable(context)

janezd:
@lanzagar, unless this fixes a known bug, I recommend against merging this just before the release, even if the test is fixed and the context migration is implemented. I discovered&fixed a few bugs just yesterday, which is a bad sign.

janezd:
Oops, I see you already came to the same conclusion. :)

lanzagar:
Yes, this one is too active to be included at the last moment before the release. Especially since there is no hurry - there is no bug, just refactoring that can be included in the next version.
Already moved from 3.8 to 3.9 milestone.

Orange/widgets/visualize/owdistributions.py
jerneju: No, we cannot migrate context in this case. @astaric confirmed this.

janezd: @astaric, how did this work at all, with indices as attributes? By taking the best matching domain and hoping it contains the needed attribute?!

Can we solve this by calling some migration-like method after opening the context? The domain is known by then, so we can change the index into a variable?

The alternative is to ignore it. In this case, I'd volunteer to go through all widgets and remove all uses of indices instead of variables, so we get rid of these problems asap.

astaric: Only attributes, represented either with name or variable are considered when matching contexts. Indices require a bit of luck and usually work with a bit of code in the widget that makes sure that when an invalid index is restored from a setting, it is changed to a valid value.

Migration can be done manually in widget after open_context returns. The context can be accessed via self.current_context.

+1 for changing all widgets to store variables.

janezd: Exactly, my question was whether we should add a patch after `openContext` to migrate the context after the domain is known, or should we say it was already rather broken before and we aren't migrating broken settings?

astaric: I lean slightly towards making it work - converting it after openContext.

The way I see the conversion could be done is by adding a keyword_only argument `domain` to `migrate_context` on this widget and calling it after openContext.

janezd:
@astaric, please take a look. For this to work, we need to keep the old settings. I hate this.

Without keeping them, the handler would not copy their values into the widget, so the hacked `migrate_context` wouldn't be able to access them. (Actually, it could peek into the current context I suppose, but this is private stuff, it shouldn't go there.)

Renaming the settings (e.g. `settings.rename_setting(context, "variable_idx", "var")`) doesn't work either, because `openContext` then tries to use these settings, which causes `ValueError: Combo box does not contain item 1`.

The current solution works, but I don't like keeping the old settings in the list. One option is to move the declaration of these settings to just before `migrate_context`, so nobody will see them.

janezd:
@jerneju, I accidentally force-pushed (too used to add `-f`! :( ) without fetching first, so I lost your commit in which you fixed the failing tests. **Don't pull from repository.** Also, I'd appreciate if you could add the commit back. (I'll try to fix it now, I hope it's related just to the order of indices in the test.)

janezd:
@jerneju, I'm sorry, could you just re-add your commit on Monday? I don't fancy fixing this if you've already fixed it once. :) Just **dont pull from github** before. And please don't lose my commit. :)

Orange/widgets/visualize/owdistributions.py
```diff
@@ -582,9 +588,28 @@ def send_report(self):
...
+            if domain is not None:  # Called after openContext, not migrations
+                settings.migrate_idx_to_variable(
+                    self, "variable_idx", "var", self.varmodel,
+                    (var for var in chain(domain.variables, domain.metas)
+                     if var.is_primitive()))
+                settings.migrate_idx_to_variable(
+                    self, "groupvar_idx", "group_var", self.groupvarmodel,
+                    (var
+                     for var in chain([None], domain.variables, domain.metas)
+                     if var is None or var.is_primitive()))
```
jerneju: The code in the lines above has a bug or bugs. The widget does not behave correctly and a test fails. Now the widget has **Group by** set to **(None)** by default.

janezd: True, but this is related to what we have to decide about.

It does migrate the settings. However, if we keep the obsolete settings just to make sure they are copied into the widget (and be visible to this method), they will always be present and they'll override the new settings. I thought `if version is None or version < 2:` would prevent this, but it obviously can't -- `version` is always `None` here. :)

I'll see how to finish this after we decide how to proceed with porting these settings. At this point, the widget can't know whether `groupvar_idx` is there from old settings or not.

astaric:
@janezd, I do not like keeping the old settings in the widget either.

I would rather have widget peek into current_context and remove the value after it is migrated than always restoring it to a widget and performing some magic to figure out whether the old value needs to be applied or not.

An alternative would be to extend context handler in a way that knows how to restore this specific setting from both old and new stored value. `decode_settings` has access to the domain so this should be possible.

janezd:
I approved the changes just to get rid of my request for changes (I should have dismissed it, actually).

@astaric, could you review just this commit? 6d63e8dc6d736b82e7f45876502c627802c6cdc7
Then I can go through the rest again, and merge.

jerneju:
@janezd : I have squashed some of my commits.

janezd:
When did you squash? Before or after my last commit? I saw the tests passed after my yesterday's commit (except appveyor being stuck in the last one). Was I wrong or ...?

Is it related or are they failing at random?

jerneju:
Now it should pass the tests.

janezd:
@astaric, the new code no longer abuses `migrate_context`, and does not have a separate method in settings.

Radon doesn't like it. If I extract the code into a local function I'm essentially replicating what we've had before, except that the previous function was in `settings` and available for others, too. Besides, any general function (local or in settings) will involve some ugly `setattr`'s and similar.

BlazZupan:
Jernej is no longer working on this PR, and it has been half a year since it was last touched. There seems to be no advancement in it despite much discussion. I am hence closing it. If this is in error, please reopen.

PR #2724, opened 2017-10-30
[FIX] errorreporting: Remove use of pip internal api

ales-erjavec:
##### Issue

Error reporting uses internal pip api, which will be [moved in Pip 10](https://mail.python.org/pipermail/distutils-sig/2017-October/031642.html).

##### Description of changes

Remove use of pip internal api

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/errorreporting.py
+++ b/Orange/canvas/application/errorreporting.py
@@ -39,6 +39,17 @@ REPORT_POST_URL = 'https://qa.orange.biolab.si/error_report/v1/'
 log = logging.getLogger()
 
 
+def get_installed_distributions():
+    for dist in pkg_resources.working_set:  # type: pkg_resources.Distribution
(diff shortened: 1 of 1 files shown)


ales-erjavec:
Which version of setuptools do you have installed (`pip list --format=freeze | grep setuptools`)

How was orange3 installed and what does `/orange3/Orange3.egg-info/PKG-INFO` contain

jerneju:
`setuptools==36.5.0`

`PKG-INFO`: `Version: 3.8.0.dev0+cb8ab62`

ales-erjavec:
I was unable to duplicate the error, but the error is part of the documented [Distribution.version](http://setuptools.readthedocs.io/en/latest/pkg_resources.html#distribution-attributes) behavior.

PR #2726, opened 2017-10-30
[FIX] Select Rows: None on output when no data

jerneju:
##### Issue
**Rank** widget crashes because receives empty `Table` and empty `Domain` which is created by **Select Rows**.

##### Description of changes
**Select Rows** commits `None`.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -530,8 +530,11 @@ class OWSelectRows(widget.OWWidget):
                 matching_output = remover(matching_output)
                 non_matching_output = remover(non_matching_output)
 
-        self.Outputs.matching_data.send(matching_output)
-        self.Outputs.unmatched_data.send(non_matching_output)
+        def output_filter(output):
+            return output if output is not None and len(output) else None
+
+        self.Outputs.matching_data.send(output_filter(matching_output))
+        self.Outputs.unmatched_data.send(output_filter(non_matching_output))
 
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/386054859/

Orange/widgets/data/owselectrows.py
```diff
@@ -530,8 +530,11 @@ def commit(self):
                 matching_output = remover(matching_output)
                 non_matching_output = remover(non_matching_output)
 
-        self.Outputs.matching_data.send(matching_output)
-        self.Outputs.unmatched_data.send(non_matching_output)
+        def output_filter(output):
+            return output if output is not None and len(output) else None
```
janezd: I here agree with @astaric's usual comment about my code. If-else expressions in Python are mostly unreadable. I'd prefer

```
if output is not None and len(output):
    return output
```

You don't even need `return None` here, but you may add it to make it explicit.

janezd: Oh, besides, I don't like the function at all. I'd prefer having

```
if not len(matching_output):
    matching_output = None
```

and the same for `non_matching_output`.

The same amount of code, but way more explicit.

janezd:
I made the change. Another option is to allow empty tables, but not empty domains. In this case we don't call `remover` if `not len(data)`.

PR #2730, opened 2017-11-02
[FIX] Fix AbstractSortTableModel.mapFromSourceRows for empty list or array

### LINKED ISSUE #2728: Data Table: Index Error
##### Orange version

3.7.dev

##### Expected behavior
Data Table reports no error.

##### Actual behavior
When reordering instances by some attribute, Data Table reports Index Error.

File (iris) - Data Table, reorder by some attribute (say petal length), select some instances from the table, then click outside of the table (as if to deselect the selection).

IndexError: arrays used as indices must be of integer (or boolean) type

janezd:
##### Issue

Fixes #2728.

##### Description of changes

Test whether a list is empty before using it for indexing an array.

##### Includes
- [X] Code changes
- [x] Tests

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -118,7 +118,7 @@ class AbstractSortTableModel(QAbstractTableModel):
 
         Parameters
         ----------
-        rows : int or list of int or numpy.ndarray or Ellipsis
+        rows : int or list of int or numpy.ndarray of dtype=int or Ellipsis
             View (sorted) rows.
 
         Returns
@@ -127,7 +127,10 @@ class AbstractSortTableModel(QAbstractTableModel):
             Source rows matching input rows. If they are the same,
             simply input `rows` is returned.
         """
-        if self.__sortInd is not None:
+        # self.__sortInd[rows] fails if `rows` is an empty list or array
+        if self.__sortInd is not None \
+                and (isinstance(rows, (Integral, type(Ellipsis)))
+                     or len(rows)):
             new_rows = self.__sortInd[rows]
             if rows is Ellipsis:
                 new_rows.setflags(write=False)
@@ -139,7 +142,7 @@ class AbstractSortTableModel(QAbstractTableModel):
 
         Parameters
         ----------
-        rows : int or list of int or numpy.ndarray or Ellipsis
+        rows : int or list of int or numpy.ndarray of dtype=int or Ellipsis
             Source model rows.
 
         Returns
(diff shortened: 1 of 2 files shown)


Orange/widgets/utils/itemmodels.py
```diff
@@ -127,7 +127,10 @@ def mapToSourceRows(self, rows):
             Source rows matching input rows. If they are the same,
             simply input `rows` is returned.
         """
-        if self.__sortInd is not None:
+        # self.__sortInd[rows] fails if `rows` is an empty list or array
+        if self.__sortInd is not None \
+                and (isinstance(rows, (int, type(Ellipsis)))
+                     or len(rows)):
```
janezd: @kernc, what do you prefer:

- `(isinstance(rows, (int, type(Ellipsis))) or len(rows))`,
- `(not isinstance(rows, (list, np.ndarray)) or len(rows))`, or
- `(not isinstance(rows, collections.Sized) or len(rows))`?

If the method is called with the argument types specified in the docstring, all are equivalent. Also, in cases where they are not equivalent (e.g. if `rows` is something stupid), the next line will fail anyway.

kernc: I already see:
```py
>>> arr = np.ones(5)
>>> empty_list = []
>>> arr[empty_list]
array([], dtype=float64)
```
I.e. empty input, empty output.

You don't?

kernc: I would prefer:
```py
if self.__sortInd is not None:
    if rows is not Ellipsis:
        rows = np.asarray(rows, dtype=int)

    new_rows = self.__sortInd[rows]
    ...
```
But this also relies on indexing (_at least_) with an empty _array_ working.

kernc:
I think https://github.com/biolab/orange3/pull/2733 is a more correct fix.

Orange/widgets/utils/itemmodels.py
kernc: _I_ would prefer your first option. Note to also fix `.mapFromSourceRows()` below.

janezd: dtype must not be float; the caller must ensure that. This is similar to `__getitem__` which expects ints, and doesn't cast float to int. I made this clear in the docstring.

I fixed `mapFromSourceRows()`.

janezd:
If I understood your comments (and their removals) correctly, this is complete now. I also improved the tests for the two mappings, so they now actually test something. :)

kernc:
Yes. Even so, I dared push some improvements while the build remains untamed.

janezd:
Feel free to dare. I force-pushed, hope it builds now. (Sorry for removing your name from the commit; it disappeared while resolving merge conflicts(?!).)

astaric:
This looks like a "segfault" to me, no idea why it happens though. It usually happens when an error occurs when destroying objects upon closing of the interpreter.

kernc:
That's what one gets for squandering over sensible modifications.

PR #2732, opened 2017-11-02
[FIX] Error Reporting: send report even when recursion error is raised

jerneju:
##### Issue
Sometimes error reporting crashes when trying to prepare a string with local variables. `RecursionError` is raised. Instead of the actual error this error is sent.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/errorreporting.py
+++ b/Orange/canvas/application/errorreporting.py
@@ -25,6 +25,8 @@ from AnyQt.QtWidgets import (
     QVBoxLayout, QWidget
 )
 
+from Orange.util import try_
+
 try:
     from Orange.widgets.widget import OWWidget
     from Orange.version import full_version as VERSION_STR
@@ -193,7 +195,8 @@ class ErrorReporting(QDialog):
             err_module = '{}:{}'.format(
                 frame.tb_frame.f_globals.get('__name__', frame.tb_frame.f_code.co_filename),
                 frame.tb_lineno)
-            err_locals = pformat(OrderedDict(sorted(frame.tb_frame.f_locals.items())))
+            err_locals = OrderedDict(sorted(frame.tb_frame.f_locals.items()))
+            err_locals = try_(lambda: pformat(err_locals), try_(lambda: str(err_locals)))
 
         def _find_widget_frame(tb):
             while tb:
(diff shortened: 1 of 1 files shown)


Orange/canvas/application/errorreporting.py
```diff
@@ -184,12 +188,15 @@ def _find_last_frame(tb):
                 tb = tb.tb_next
             return tb
 
-        err_locals, err_module, frame = None, None, _find_last_frame(tb)
+        err_module, frame = None, _find_last_frame(tb)
```
kernc: Leave `err_locals=None` default even when `not frame`? Then you can `pass` the exception below.

jerneju: Yes, firstly I did so. Reverting back.

kernc: I think it's safer since `err_locals` is used _unconditionally_ further below.

Orange/canvas/application/errorreporting.py
```diff
@@ -189,7 +193,10 @@ def _find_last_frame(tb):
             err_module = '{}:{}'.format(
                 frame.tb_frame.f_globals.get('__name__', frame.tb_frame.f_code.co_filename),
                 frame.tb_lineno)
-            err_locals = pformat(OrderedDict(sorted(frame.tb_frame.f_locals.items())))
+            try:
+                err_locals = pformat(OrderedDict(sorted(frame.tb_frame.f_locals.items())))
+            except RuntimeError:  # RecursionError is not supported by Python < 3.5
```
kernc: I assume RecursionError here is triggered because _something_ in the above line triggers an exception which unhandled gets passed to our ErrorReporting hook, and the thing reiterates. Could we instead try to catch that something?

jerneju:
https://sentry.io/biolab/orange3/issues/274030284/

Orange/canvas/application/errorreporting.py
jerneju: Added the link despite you've already found it.

kernc: RecursionError is a bit deep. Just catch Exception ...
```py
from Orange.util import try_

err_locals = OrderedDict(sorted(frame.tb_frame.f_locals.items()))
err_locals = try_(lambda: pformat(err_locals), try_(lambda: str(err_locals)))
```

PR #2736, opened 2017-11-03
Variable: Add sparse flag

nikicc:
##### Issue

In order to properly support sparse features, this PR introduces the `Variable.sparse` flag, which indicates that the variable suggests to be stored in sparse format.

##### Description of changes
* Add `sparse` flag to variables.
* When copying variables, copy the sparse flag as well. This is far from correct, but it should do for a start.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -259,6 +259,11 @@ class Variable(Reprable, metaclass=VariableMeta):
         defines a static method `compute_value`, which returns `Unknown`.
         Non-primitive variables must redefine it to return `None`.
 
+    .. attribute:: sparse
+
+        A flag about sparsity of the variable. When set, the variable suggests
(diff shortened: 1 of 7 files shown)


Orange/data/variable.py
```diff
@@ -565,12 +571,13 @@ class DiscreteVariable(Variable):
     _all_vars = collections.defaultdict(list)
     presorted_values = []
 
-    def __init__(self, name="", values=(), ordered=False, base_value=-1, compute_value=None):
+    def __init__(self, name="", values=(), ordered=False, base_value=-1,
+                 compute_value=None, *, sparse=False):
```
astaric: @janezd we might want to move * after values, but not in this PR.

PR #2737, opened 2017-11-03
[ENH] Allow custom (generic) names in Transpose Widget 

### LINKED ISSUE #2729: Allow custom (generic) names in Transpose Widget
When I transpose a dataset with no meta attributes, I end up with variables named Feature 1, Feature 2, Feature 3, ...
It would be nice if I could specify my own prefix (such as User, Transaction, Cell, House, ...).

The text box with the prefix (the default can stay Feature, although I would vote for Row) could be placed below the Generic options, the way variable combo is placed below the From meta attribute option.
screen shot 2017-11-02 at 11 09 31

jerneju:
##### Issue
Fixes #2729 

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtranspose.py
+++ b/Orange/widgets/data/owtranspose.py
@@ -19,11 +19,16 @@ class OWTranspose(OWWidget):
     class Outputs:
         data = Output("Data", Table, dynamic=False)
 
+    GENERIC, FROM_META_ATTR = range(2)
+
     resizing_enabled = False
     want_main_area = False
 
+    DEFAULT_PREFIX = "Feature"
+
     settingsHandler = DomainContextHandler()
-    feature_type = ContextSetting(0)
+    feature_type = ContextSetting(GENERIC)
+    feature_name = ContextSetting("")
     feature_names_column = ContextSetting(None)
     auto_apply = Setting(True)
 
@@ -34,27 +39,40 @@ class OWTranspose(OWWidget):
         super().__init__()
         self.data = None
 
-        # GUI
-        box = gui.vBox(self.controlArea, "Feature names")
-        self.feature_radio = gui.radioButtonsInBox(
-            box, self, "feature_type", callback=lambda: self.apply(),
-            btnLabels=["Generic", "From meta attribute:"])
+        box = gui.radioButtons(
+            self.controlArea, self, "feature_type", box="Feature names",
+            callback=lambda: self.apply())
 
+        button = gui.appendRadioButton(box, "Generic")
+        edit = gui.lineEdit(
+            gui.indentedBox(box, gui.checkButtonOffsetHint(button)), self,
+            "feature_name",
+            placeholderText="Type a prefix ...", toolTip="Custom feature name")
+        edit.editingFinished.connect(self._apply_editing)
+
+        self.meta_button = gui.appendRadioButton(box, "From meta attribute:")
         self.feature_model = DomainModel(
             order=DomainModel.METAS, valid_types=StringVariable,
             alphabetical=True)
         self.feature_combo = gui.comboBox(
-            gui.indentedBox(
-                box, gui.checkButtonOffsetHint(self.feature_radio.buttons[0])),
-            self, "feature_names_column", callback=self._feature_combo_changed,
+            gui.indentedBox(box, gui.checkButtonOffsetHint(button)), self,
+            "feature_names_column", callback=self._feature_combo_changed,
             model=self.feature_model)
 
         self.apply_button = gui.auto_commit(
             self.controlArea, self, "auto_apply", "&Apply",
             box=False, commit=self.apply)
+        self.apply_button.button.setAutoDefault(False)
+
+        self.set_controls()
+
+    def _apply_editing(self):
+        self.feature_type = self.GENERIC
+        self.feature_name = self.feature_name.strip()
+        self.apply()
 
     def _feature_combo_changed(self):
-        self.feature_type = 1
+        self.feature_type = self.FROM_META_ATTR
         self.apply()
 
     @Inputs.data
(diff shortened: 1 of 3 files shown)


janezd:
Not like this. Keep two radio buttons, Generic and From meta attribute. Generic than has a line edit with placeholder Feature.

Indent the line edit so that it is aligned with the combo.

Orange/widgets/data/owtranspose.py
```diff
@@ -7,6 +11,28 @@
 from Orange.widgets.widget import Input, Output
 
 
+def add_radio_buttons(widget, master, value, btn_labels=(), orientation=Qt.Vertical, callback=None):
```
janezd: I don't like this function at all. It is called `add_radio_buttons`, and it appends radio buttons and line edits. And then a combo is added separately since this function handles only line edits. (Well, if it handled everything, it would just replace any placing of components inside control area.)

It looks general, but is called only once and replace what we would otherwise do with a sequence of calls to `gui` method that we usually use for constructing the interface. I would suggest refactoring this code into `__init__`.

jerneju: I wanted to make it general, well then I moved everything into `__init__`.

Orange/widgets/data/owtranspose.py
```diff
@@ -76,25 +106,40 @@ def update_controls(self):
             if self.feature_model:
                 self.feature_names_column = self.feature_model[0]
         enabled = bool(self.feature_model)
-        self.feature_radio.buttons[1].setEnabled(enabled)
+        self.feature_radio.buttons[self.FROM_META_ATTR].setEnabled(enabled)
```
janezd: The previous version allowed changing the radio button, but prevented changing the combo when no data. I don't see much point in disabling the combo and lineedit, and leave the radios on.

On the other hand, is there any harm in letting the user monkey around the controls when there is no data? I'd just leave them enabled.

Orange/widgets/data/owtranspose.py
```diff
@@ -76,25 +106,40 @@ def update_controls(self):
...
         self.feature_combo.setEnabled(enabled)
-        self.feature_type = int(enabled)
+        self.feature_type = self.FROM_META_ATTR if enabled else self.GENERIC
+
+    def _apply_editing(self):
+        self.feature_type = self.GENERIC
+        self.custom_feature_name.setText(self.custom_feature_name.text().strip())
+        self.feature_name = self.custom_feature_name.text()
+        if self.auto_apply:
```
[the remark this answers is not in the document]
janezd: If you added `if` to stop the changes from being applied when auto apply was disabled: the problem is that the auto apply button is a default button, hence Enter triggers it. To prevent this, use `self.apply_button.button.setAutoDefault(False)`.

janezd:
I pushed my take on it. I hope it works. More tests would be cool. I might provide some; currently they don't test the typical crashes expected to occur when user does something (e.g. clicking report) with a widget without data.

Orange/widgets/data/tests/test_owtranspose.py
```diff
@@ -33,7 +33,30 @@ def test_parameters(self):
...
-        self.widget.feature_radio.buttons[0].click()
+        self.widget.feature_type = self.widget.GENERIC
         self.widget.apply()
         self.assertTrue(all(["Feature" in x.name for x
                              in self.get_output(self.widget.Outputs.data).domain.attributes]))
+
+    def test_transpose_custom_name(self):
+        """
+        Custom name can be set.
+        GH-2737
```
janezd: In my opinion -- and I would want to hear what @astaric, @lanzagar and @kernc think about it -- tests should test some functionality and not specific bugs that were discovered, or be related to some specific issue in any way. Hence, I do not like comments like GH-2737. This makes our unit tests look like a bunch of tests for patches. This is irrelevant -- potential bugs must be covered by "normal" tests, but bug-specific tests.

This bothered me for some time already, but now I stumbled upon this again when I wanted to reorganize tests for this widget, and I decided to "dissolve" this test into some other test.

Again, this is just my opinion.

Orange/widgets/data/tests/test_owtranspose.py
```diff
@@ -33,7 +33,30 @@ def test_parameters(self):
...
+        self.assertEqual(w.feature_type, w.GENERIC)
+        CUSTOM_TEXT = "Bravo"
+        w.feature_name = CUSTOM_TEXT
+        w.apply()
+        self.assertTrue(all([CUSTOM_TEXT in v.name for v
+                             in self.get_output(w.Outputs.data).domain.attributes]))
+
+    def test_send_report(self):
+        """
+        GH-2773
```
janezd: Here is an even better example: `test_send_report` should be a test for reports. Why is it marked as "GH-2737"?

Orange/widgets/data/tests/test_owtranspose.py
janezd: but bug-specific tests. -> not bug specific tests.

kernc: Fixing bugs or commonplace stuff like test_report, sure. But in general, it's not necessarily obvious why a certain feature was added in amendments as opposed to initially. It's customary to cross link to the issue(s) where/when the surrounding discussion occurred. Case by case; ymmv.

Orange/widgets/data/owtranspose.py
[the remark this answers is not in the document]
jerneju: Exactly, that was the reason I added that `if`.

Orange/widgets/data/owtranspose.py
```diff
@@ -34,27 +39,41 @@ def __init__(self):
...
-            btnLabels=["Generic", "From meta attribute:"])
+        box = gui.radioButtons(
+            self.controlArea, self, "feature_type", box="Feature names",
+            callback=lambda: self.apply())
+
+        button = gui.appendRadioButton(box, "Generic")
+        edit = gui.lineEdit(
+            gui.indentedBox(box, gui.checkButtonOffsetHint(button)), self,
+            "feature_name",
+            placeholderText="Prefix for column names",
```
lanzagar: Currently, I can not read the whole placeholder text because the widget is too small and not resizable.
We can leave the text as is and just make the widget a bit larger?

Orange/widgets/data/tests/test_owtranspose.py
```diff
@@ -13,27 +15,130 @@ def setUp(self):
...
-                             in self.get_output(self.widget.Outputs.data).domain.attributes]))
+            [a.name for a in output.domain.attributes],
+            [metas[1].to_val(m) for m in data.metas[:, 1]])
+
+        # Switch to generic
+        widget.feature_type = widget.GENERIC
+        widget.apply()
+        output = self.get_output(widget.Outputs.data)
+        self.assertTrue(
+            all(x.name.startswith("Feature ")
```
lanzagar: Wouldn't it be better to test that names start with `OWTranspose.DEFAULT_PREFIX`?
(in case someone changes the default in the future)

Orange/widgets/data/tests/test_owtranspose.py
```diff
@@ -13,27 +15,130 @@ def setUp(self):
...
+        output = self.get_output(widget.Outputs.data)
+        self.assertTrue(
+            all(x.name.startswith("Foo ") for x in output.domain.attributes))
+
+        # Check that the widget uses default when name is not given
+        widget.feature_name = ""
+        widget.apply()
+        output = self.get_output(widget.Outputs.data)
+        self.assertTrue(
+            all(x.name.startswith("Feature ")
```
lanzagar: Wouldn't it be better to test that names start with `OWTranspose.DEFAULT_PREFIX`?
(in case someone changes the default in the future)

PR #2743, opened 2017-11-04
OWWidget: Store control area visibility as setting

janezd:
##### Issue

Widgets store their geometry but not the visibility of `controlArea`.

##### Description of changes

Store the visibility as a setting.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/widget.py
+++ b/Orange/widgets/widget.py
@@ -67,6 +67,7 @@ class WidgetMetaClass(type(QDialog)):
         return cls
 
 
+# pylint: disable=too-many-instance-attributes
 class OWWidget(QDialog, OWComponent, Report, ProgressBarMixin,
                WidgetMessagesMixin, WidgetSignalsMixin,
                metaclass=WidgetMetaClass):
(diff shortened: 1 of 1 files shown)


astaric:
I vote for making this a schema_only setting.

We have also discussed storing which widgets are shown and re-displaying them once the workflow is opened. That might be related.

@ales-erjavec had a nice idea about adding save_state method to the widget and calling it whenever widget's settings are packed. With it, the setting value could be updated just before widget's settings are packed.

janezd:
> I vote for making this a schema_only setting.

Of course. Done.

BlazZupan:
Works as intended according to my tests.

It exposes though some other problem we have when working with sizes of the widgets. Example: put some data in Scatterplot, then hide the control, save the workflow and close. Open the workflow, show Scatterplot. The control is hidden by the widget is higher than when used in the original workflow. Add another Scatterplot: in this one, the control is shown (right, because it is a new widget), but the size of the main area is minimal (and we need to widen the widget).

A solution to above is that we would need to remember the "size" of the main area. When we expand the control, the widget should widen such that the main area remains of the same width. And when we collapse the control area, the widget should shrink to the left to maintain the width of the main area. An exception is a height of the widget: when control is shown, the hight is the max of the current height and the high with controls. When control area is collapsed, the widget should remember what was the last collapsed height and should not only reduce the width.

janezd:
> And when we collapse the control area, the widget should shrink to the left to maintain the width of the main area.

I think it has to shrinking to the right, that is retain the position of the main area.

janezd:
Should work now.

PR #2751, opened 2017-11-07
[FIX] Test & Learn: do not crash on a data with class only nans

jerneju:
##### Issue
Widget **Test & Learn** crashes when receives data with a class which has only `nan` values. This is because all data is removed and `Table` suddenly has length `0`.

##### Description of changes
The data is set to `None` when its length is `0`.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -206,6 +206,7 @@ class OWTestLearners(OWWidget):
         class_inconsistent = Msg("Test and train data sets "
                                  "have different target variables.")
         memory_error = Msg("Not enough memory.")
+        no_class_values = Msg("Target variable has no values.")
         only_one_class_var_value = Msg("Target variable has only one value.")
 
     class Warning(OWWidget.Warning):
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/323337251/

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -388,6 +388,8 @@ def set_train_data(self, data):
             self.Warning.missing_data(self._which_missing_data())
             if data:
                 data = HasClass()(data)
+                if not len(data):
+                    data = None
```
markotoplak: Is the warning still OK in this case? I do not know what it is, but we need to show clearly that nothing can get computed.

jerneju: When there is no target variable it shows an error: "Train data input requires a target variable." and now where there is a target variable with only `nan` values it shows a warning: "Instances with unknown target values were removed from train data.".

What would be the best error / warning message?

jerneju: @markotoplak : Now I fixed it in a different way.

PR #2752, opened 2017-11-07
Add option to adjust symbol size

JakaKokosar:
##### Issue
Unable to adjust symbol size like, for example, in Scatter Plot widget.

##### Description of changes
Add another horizontal slider which controls symbol size.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpaintdata.py
+++ b/Orange/widgets/data/owpaintdata.py
@@ -759,6 +759,8 @@ class OWPaintData(OWWidget):
 
     brushRadius = Setting(75)
     density = Setting(7)
+    symbol_size = Setting(10)
+
     #: current data array (shape=(N, 3)) as presented on the output
     data = Setting(None, schema_only=True)
     labels = Setting(["C1", "C2"], schema_only=True)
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owpaintdata.py
```diff
@@ -759,6 +759,8 @@ class Outputs:
 
     brushRadius = Setting(75)
     density = Setting(7)
+    symbolSize = Setting(10)
```
jerneju: Maybe `symbol_size` reads better?

Orange/widgets/data/owpaintdata.py
```diff
@@ -1223,6 +1237,10 @@ def pen(color):
             pen.setCosmetic(True)
             return pen
 
+        def brush(color):
```
jerneju: I suppose this function is here only because `pen` has its own method. Otherwise this is not needed.

Orange/widgets/data/owpaintdata.py
```diff
@@ -1223,6 +1237,10 @@ def pen(color):
...
+            brush = QBrush(color)
+            return brush
```
jerneju: Just `return QBrush(color)`?

Orange/widgets/data/owpaintdata.py
```diff
@@ -1235,10 +1253,13 @@ def pen(color):
 
         colors = self.colors[self.__buffer[:, 2]]
         pens = [pen(c) for c in colors]
+        brushes = [brush(c) for c in colors]
```
jerneju: If function `brush` is removed:
```python
brushes = [QBrush(c) for c in colors]
```

PR #2756, opened 2017-11-08
[NOMERGE] Revert "itemmodels.AbstractSortTableModel: Fix mapFromSourceRows"

kernc:
5854208ff56968218b13431784cb8fb1eb1b17cb is the first commit where AppVeyor started consistently failing. Submitted against my better judgment, this should wholly exclude that commit being at fault.


kernc:
Finally onto something! 😂

astaric:
The last time we had random segfaults was when we were testing reports (which use some qthread context magic). Have any new reports been added lately?

kernc:
Fixed in https://github.com/biolab/orange3/pull/2759.

PR #2757, opened 2017-11-09
[FIX] Scatter Plot: VizRank disabled when no class vars

jerneju:
##### Issue
**Scatter Plot** VizRank crashes when _class column_ has only `nan` values.

##### Description of changes
VizRank is disabled as it is when there is no `class_var`.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -251,7 +251,7 @@ class OWScatterPlot(OWWidget):
         self.vizrank.attrs = self.data.domain.attributes if self.data is not None else []
         self.vizrank_button.setEnabled(
             self.data is not None and not self.data.is_sparse() and
-            self.data.domain.class_var is not None and
+            self.data.domain.class_var is not None and not np.isnan(self.data.Y).all() and
             len(self.data.domain.attributes) > 1 and len(self.data) > 1)
         if self.data is not None and self.data.domain.class_var is None \
             and len(self.data.domain.attributes) > 1 and len(self.data) > 1:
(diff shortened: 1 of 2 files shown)


jerneju:
https://sentry.io/biolab/orange3/issues/401246621/

Orange/widgets/visualize/tests/test_owscatterplot.py
```diff
@@ -392,6 +392,27 @@ def test_vizrank(self):
...
+            self.send_signal(self.widget.Inputs.data, data)
+            self.assertEqual(is_enabled, self.widget.vizrank_button.isEnabled())
+
+        data1 = Table("iris")[::30]
+        data2 = Table("iris")[::30]
+        data2.Y[:] = np.nan
+        domain = Domain(
+            attributes=data2.domain.attributes[:4], class_vars=DiscreteVariable("iris", values=[]))
+        data2 = Table(domain, data2.X, Y=data2.Y)
+        data3 = Table("iris")[::30]
```
lanzagar: What is the difference between data1 and data3?  (i.e. why this line)

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -252,6 +252,8 @@ def set_data(self, data):
         self.vizrank_button.setEnabled(
             self.data is not None and not self.data.is_sparse() and
             self.data.domain.class_var is not None and
+            (len(self.data.domain.class_var.values)
+             if self.data.domain.class_var.is_discrete else True) and
```
lanzagar: This fixes the case when the class has no values. How about when the class has e.g. 3 values, but all instances have unknown classes (e.g. a test set with the same variable as the training set, but class unknown)

Orange/widgets/visualize/tests/test_owscatterplot.py
jerneju: Now it is the difference.

PR #2758, opened 2017-11-09
owdatasets: Move constants and util functions

lanzagar:
##### Description of changes
Refactoring:  Moved constants with local and remote locations inside the class along with some util functions. This makes the widget more self contained and allows easier reuse / modifications.

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -31,32 +31,17 @@ from Orange.widgets.utils.signals import Output
 from Orange.widgets.widget import Msg
 
 
-INDEX_URL = "http://datasets.orange.biolab.si/"
 log = logging.getLogger(__name__)
 
 
-def local_cache_path():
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owdatasets.py
```diff
@@ -477,6 +459,24 @@ def closeEvent(self, event):
...
         super().closeEvent(event)
 
+    def ensure_local(self, domain, filename, force=False,
+                     progress_advance=None):
+        localfiles = LocalFiles(self.local_cache_path,
+                                serverfiles=ServerFiles(server=self.INDEX_URL))
+        if force:
+            localfiles.download(domain, filename, callback=progress_advance)
+        return localfiles.localpath_download(
+            domain, filename, callback=progress_advance)
```
ales-erjavec: `ensure_local` as such should not be a method. Parameterize it with target local path and index url if necessary.

lanzagar: Changed.

PR #2760, opened 2017-11-09
[FIX] Canvas: Palette propagation

ales-erjavec:
##### Issue

Canvas conflicts with dark theme/styles, e.g:

screen shot 2017-11-09 at 17 51 11

##### Description of changes

* Propagate the palette in the canvas where applicable
* Use palette's colors instead of hardcoded ones
* Add an option to specify --style=fusion:breeze-dark (the palette's colors are taken from KDE's t 'Breeze Dark' theme). **This is mostly just for testing, the fusion style is not a good with dark color themes**
* Suppress the default 'orange' styling in the main window if the application palette is to dark.

<s>If deemed necessary a GUI option in the preferences dialog can also be added to select the style.</s>

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -18,7 +18,7 @@ from unittest.mock import patch
 
 import pkg_resources
 
-from AnyQt.QtGui import QFont, QColor, QDesktopServices
+from AnyQt.QtGui import QFont, QColor, QPalette, QDesktopServices
 from AnyQt.QtWidgets import QMessageBox
 from AnyQt.QtCore import (
     Qt, QDir, QUrl, QSettings, QThread, pyqtSignal, QT_VERSION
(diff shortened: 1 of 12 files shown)


kernc:
❤️ 

Any possibility to also support [`QT_STYLE_OVERRIDE` env variable](http://doc.qt.io/qt-5/qapplication.html#QApplication)?

ales-erjavec:
What do you mean? QT_STYLE_OVERRIDE works the same as always.

kernc:
Hmm, I set QT_STYLE_OVERRIDE, which works great for other Qt5 (non-Python) apps (qBittorrent, Clementine, VirtualBox), but doesn't seem to work with Orange.
```bash
$ env | grep QT_STYLE                                                                                                                                                                                                                               
QT_STYLE_OVERRIDE=adwaita-dark
```

ales-erjavec:
Do you have PyQt5 installed from wheels?

Try using the system provided PyQt5. I.e. use `--system-site-packages` when creating the env or just symlink `/usr/lib/python3/dist-packages/{sip.*so,PyQt5}` into the created env

kernc:
That did the trick. Looks quite good. Except the icons, perhaps, aren't best suited to dark themes.

screenshot_2017-12-10_04-13-50

PR #2764, opened 2017-11-10
owsilhuetteplot: use QSizeF.transpose

markotoplak:
QSizeF.transposed is not available on Qt4.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


astaric:
IIUC, this crashes on PyQt4. Could we have a test that calls this code to make sure it does not crash any more?

markotoplak:
No tests. :)

PR #2767, opened 2017-11-10
OWDatasets: fix row height

markotoplak:
##### Issue
Row height was too small

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -283,6 +283,7 @@ class OWDataSets(widget.OWWidget):
             item1 = QStandardItem()
             item1.setData(" " if datainfo.islocal else "", Qt.DisplayRole)
             item1.setData(datainfo, Qt.UserRole)
+            item1.setData(QSize(26, 26), Qt.SizeHintRole)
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owdatasets.py
```diff
@@ -283,6 +283,7 @@ def info(file_path):
             item1 = QStandardItem()
             item1.setData(" " if datainfo.islocal else "", Qt.DisplayRole)
             item1.setData(datainfo, Qt.UserRole)
+            item1.setData(QSize(26, 26), Qt.SizeHintRole)
```
lanzagar: On my system this looks quite big.
24 still results in higher rows but looks better IMO.

markotoplak:
Why do we need this at all?

kernc:
This is how it looks here. The lines with target pixmaps are higher than the lines without.
screenshot_2017-11-13_13-18-35
While I'd never wish to advocate for this, some padding is in.

ales-erjavec:
@kernc Can you please check gh-2774

janezd:
Closed in favour of #2774.

PR #2769, opened 2017-11-11
DomainModel: Prevent modification except through set_domain

janezd:
##### Issue

`DomainModel` should not allow modifying the underlying list, except through `set_domain`.

##### Description of changes

- Decorate all list methods that would modify the underlying list to raise an exception, unless they are called from within `set_domain`. This also prevents modifying the model through `setData`, `setItemData`, `insertRows` and `removeRows`.
- Fix widgets that directly changed the model: MDS, Box plot
- Add tests for `DomainModel` and `PyListModel`
- Fix detected problems in `PyListModel`:
-- fix `__add__` and `__iadd__` which didn't work at all
-- remove a bug in `_is_index_valid`, refactor it, add checks to detect models broken by misuse of `wrap` (see #2771); current implementation used an additional argument to circumvent the potential bugs
-- remove unnecessary use of `wrap` in Edit domain

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -216,6 +216,8 @@ class OWMDS(OWWidget):
         g = self.graph.gui
         box = g.point_properties_box(self.controlArea)
         self.models = g.points_models
+        self.models[2].order = \
+            self.models[2].order[:1] + ("Stress", ) + self.models[2].order[1:]
 
         gui.hSlider(box, self, "connected_pairs", label="Show similar pairs:", minValue=0,
                     maxValue=20, createLabel=False, callback=self._on_connected_changed)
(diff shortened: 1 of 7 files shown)


Orange/widgets/unsupervised/owmds.py
```diff
@@ -216,6 +216,8 @@ def __init__(self):
         g = self.graph.gui
         box = g.point_properties_box(self.controlArea)
         self.models = g.points_models
+        self.models[2].order = \
+            self.models[2].order[:1] + ("Stress", ) + self.models[2].order[1:]
```
janezd: This modifies a model that is then used by all widgets. Copy the model before modifying it.

janezd:
Travis says: `Ran 2018 tests in 331.382s`

It would appear that we write roughly one test per year.

PR #2774, opened 2017-11-13
OWDatasets: Force uniform row heights in the view

ales-erjavec:
##### Issue

Alternative to gh-2767

##### Description of changes

Use `uniformRowHeights` property on `QTreeView` and reimplement item delegate's `sizeHint` for the first column to always include the icon.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -54,7 +54,23 @@ def format_exception(error):
     return "\n".join(traceback.format_exception_only(type(error), error))
 
 
-class SizeDelegate(QStyledItemDelegate):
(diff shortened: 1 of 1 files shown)


kernc:
Row heights are uniform. Should probably also resize Title column to contents.

ales-erjavec:
'Title' column width and position is restored from settings and I do not want to change it. So instead the title column width is initialized (to some reasonable size) on first instantiation only.

ales-erjavec:
You can test this with:
```bash
python -m Orange.widgets.data.owdatasets -stylesheet /dev/stdin <<<"QTreeView { icon-size:30px; }"
```
(modify accordingly for your chosen shell)

PR #2775, opened 2017-11-14
[FIX] Scatter Plot Graph: crash on metas column with all 0 values

jerneju:
##### Issue
**Scaling** does not stack metas with other arrays when metas array has all values set to zero.

##### Description of changes
Another check for metas array.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/scaling.py
+++ b/Orange/widgets/utils/scaling.py
@@ -49,11 +49,7 @@ class ScaleData:
             return
 
         Y = data.Y if data.Y.ndim == 2 else np.atleast_2d(data.Y).T
-        if np.any(data.metas):
-            all_data = (data.X, Y, data.metas)
-        else:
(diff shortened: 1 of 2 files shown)


markotoplak:
I do not understand why was this needed. As far as I know, numpy hstack works with zeros as well. 

On a related note, what would happen if X had an attribute that was all zeros?

lanzagar:
Exactly, that is why now everything is hstacked and it should work no matter what kind of metas there are.
Before, there was an if condition `if np.any(data.metas):` which not only failed for no metas, but also for all-zero metas, in which case they were not hstacked at all. (Probably the if should have checked `if data.domain.metas`) But no idea why the if was necessary at all...

markotoplak:
My stupidity, I read this patch backwards. :D

PR #2780, opened 2017-11-16
[FIX] FreeViz: 2 issues when no data

jerneju:
##### Issue
Widget crashes when there is no data and
1) Radius slider is moved
2) Class density, etc. is changed

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owfreeviz.py
+++ b/Orange/widgets/visualize/owfreeviz.py
@@ -815,6 +815,8 @@ class OWFreeViz(widget.OWWidget):
         self.graph.update_data(self.variable_x, self.variable_y, reset_view)
 
     def update_density(self):
+        if self.graph.data is None:
+            return
         self._update_graph(reset_view=False)
 
     def selection_changed(self):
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owfreeviz.py
```diff
@@ -810,8 +810,9 @@ def reset_graph_data(self, *_):
             self._update_graph()
 
     def _update_graph(self, reset_view=True, **_):
+        if self.graph.data is None:
```
janezd: `_update_graph` is called from `reset_graph_data` and `update_density`. The former already checks that `graph.data` is not `None`. You could add this check to the latter as well instead of having it in `_update_graph`, and just move the assert in `_update_graph` to the top.

jerneju: Done.

lanzagar:
Not in the scope of this PR (which fixes bugs and we want in the 3.8 release), but maybe for future refactoring
1) `assert not self.plotdata is None` should be `assert self.plotdata is not None`
2) I fail to see why the above asserts are needed at all (plotdata is never None, even after initialization)

Back to the PR  - the fixes work for me and look ok.

PR #2786, opened 2017-11-20
Pullrequest

alemirone:
##### Description of changes

7c2c2b6  allows to reused an existing widget as a base class
and still have its input and outputs so that the new inputs and outputs exists together with the base class ones

57228e8  allows a widget to know the node which owns it.

0228004 enforces uniqueness of the widget names

fe8a0d4  is important for big calculations. It allows the receiving widget to know
   is the signal is triggere by a link or a send. The information may be used 
by the widget to decide it further computing will be triggered or not

ea23057 is a correction for 7c2c2b6

The other commits are not important
except 0ed3d50 which is useful when big calculation may take the currect directory
as root for the paths outputs, but I have little hope that you may like it
Regards
   Alessandro Mirone 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


alemirone:
Hi Everybody,
Alessandro Mirone
alex
operator for beamline

are all the same person (me) and I am going to sign the agreement for alemirone
regards

astaric:
7c2c2b6: this is already possible, see:
https://github.com/biolab/orange3/blob/master/Orange/widgets/model/owlinearregression.py#L28

ajdapretnar:
This PR needs to be cleaned up and renamed, if anything. @astaric @lanzagar Are there any commits that we wish to introduce to Orange?

janezd:
I am closing this due to inactivity, unclear purpose of the PR and because we have to decide which of the proposed changes would we indeed like to see in Orange.

If there are some strong arguments for any of these features, let us discuss them as issues first.

PR #2790, opened 2017-11-22
Add nansum and nanmedian.

mstrazar:
##### Issue
Added NaN sum and median to Orange.statistics.util for consistent computation of medians and sums for dense and sparse matrices, if the data may contain NaNs.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -394,7 +394,7 @@ def mean(x):
 
 
 def _apply_func(x, dense_func, sparse_func, axis=None):
-    """ Equivalent of np.nanmean that supports sparse or dense matrices. """
+    """ General wrapper for a function depending on sparse or dense matrices. """
     if not sp.issparse(x):
         return dense_func(x, axis=axis)
     if axis is None:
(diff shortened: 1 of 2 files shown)


Orange/statistics/util.py
```diff
@@ -427,6 +435,23 @@ def nanvar_sparse(x):
...
+        nz = np.logical_not(np.isnan(x.data))
+        n_zeros = np.prod(x.shape) - len(x.data)
+        n_nz = sum(nz)
+        if n_zeros > n_nz:
+            # Typical case if use of sparse matrices make sense
+            return 0
+        else:
+            # Possibly contains NaNs and
+            # more nz values than zeros, so allocating memory should not be too problematic
+            return np.nanmedian(np.array(x.todense()))
```
nikicc: I think that `x.toarray()` does the same as `np.array(x.todense())`

Orange/tests/test_statistics.py
```diff
@@ -126,6 +126,13 @@ def test_nanmin_nanmax(self):
                     nanmax(X_sparse, axis=axis),
                     np.nanmax(X, axis=axis))
 
+    def test_nansum(self):
+        for X in self.data:
```
nikicc: Can we use '@dense_sparse' decorator for running these tests? It also tests on sparse matrices with explicit zeros.

PR #2801, opened 2017-11-28
[ENH] New signals: Trees, Forest

### LINKED ISSUE #2797: Pythagorean Tree & Tree Viewer Have Static Output Signals
##### Orange version

3.8.0.dev0+a3b4f8b

##### Expected behavior
Pythagorean Tree & Tree Viewer would have dynamic signals as most Orange widgets now do.

##### Actual behavior
Pythagorean Tree & Tree Viewer seem to have static signals which causes problems for addons. For example, in Text that uses Corpuses instead of Tables, the outputs from these two widgets cannot be passes back to widgets accepting `Corpus` on input.

In the scheme below, one cannot connet the output of Pythagorean Tree or Tree Viewer back to Corpus Viewer:
screen shot 2017-11-27 at 17 51 37

jerneju:
##### Issue
Fixes #2797 

##### Description of changes
Tree Viewer
Pythagorean Tree
Pythagorean Forest
~~Geo Map~~

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owpythagorastree.py
+++ b/Orange/widgets/visualize/owpythagorastree.py
@@ -8,11 +8,12 @@ from AnyQt.QtWidgets import QLabel, QSizePolicy
 
 from Orange.base import TreeModel, SklModel
 from Orange.data.table import Table
-from Orange.widgets import gui, settings, widget
+from Orange.widgets import gui, settings
 from Orange.widgets.utils.annotated_data import (
     create_annotated_table,
     ANNOTATED_DATA_SIGNAL_NAME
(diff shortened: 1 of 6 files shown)


nikicc:
@lanzagar I suggest we merge this, if you still allow merging before the release.

lanzagar:
I agree introducing another variable and line of code is often not worth it just to save a few characters in a couple of lines after that. But it is a matter of preference and not a big issue so I would prefer not to wait with merging until the last minutes before a release.

PR #2804, opened 2017-11-28
[ENH] Scatter Plot Graph: max discrete values colors and shape

### LINKED ISSUE #2782: Scatter Plot: does not support many different shapes
##### Orange version
Probably all versions since **3.3.12** or even before.

##### Expected behavior
Maybe should at least show a warning that cannot draw so many shapes.

##### Actual behavior
See a screenshot below. Class variable **Iris10** has 10 different values. They have different colors but the same shape.

screenshot_20171117_135835

jerneju:
##### Issue
Fixes #2782

##### Description of changes
screenshot_20171205_120034

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -1,10 +1,10 @@
+from collections import Counter
 import sys
 import itertools
 from xml.sax.saxutils import escape
 from math import log10, floor, ceil
 
 import numpy as np
-import scipy.sparse as sp
 from scipy.stats import linregress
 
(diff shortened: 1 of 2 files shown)


jerneju:
> I have a much worse color palette (gradient from red to blue with many similar colors).
Color palette is variable dependent. See `variable.py`.

lanzagar:
> Color palette is variable dependent. See variable.py.

Which means your changes here are problematic because they do not handle this...
The colors are determined based on the variable (number of values), but you change the values before plotting. So you can't use the palette for a variable with e.g. 100 values after merging that into 11 values.

PR #2811, opened 2017-12-01
owconcatenate: Use table names for values of source var

lanzagar:
##### Issue

Since loading a workflow doesn't produce a consistent ordering, indices aren't reliable.

##### Description of changes
Use names of the tables as values of the new (optional) source variable.
(Append indices if not unique)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
@mstrazar please check if this solves your problems with saved workflows (for now).
We should consider a more general solution for this type of problems in a future release (probably needs some changes to core mechanics).

PR #2813, opened 2017-12-01
Updated Neural Networks icon

larazupan:


markotoplak:
Would this look better on the canvas if the image was moved slightly to the left? Could you try it, please?

PR #2829, opened 2017-12-15
DomainContextSetting: don't skip meta-only domains

janezd:
##### Issue

If a domain has just meta attributes, `DomainContextHandler` does not open a context.

I encountered this when converting the Distance map widget to use variables instead of ints. I loaded 'conferences.dst' with Distance File, and connected it to Distance map, which failed since the widget had no current context. I no longer have this problem since I had to change the widget for other reasons, but the bug may reappear elsewhere.

I can write a test if @astaric agrees with the change.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -916,7 +916,7 @@ class DomainContextHandler(ContextHandler):
         return context
 
     def open_context(self, widget, domain):
-        if not domain:
+        if domain is None:
             return None, False
 
         if not isinstance(domain, Domain):


astaric:
I would change it to
```
if domain is None:
```

If you see the line(s) below, you will see that domain can also be table (and that is what is usually passed by widgets...).

janezd:
Of course. Yesterday was not my day.

Now I wouldn't even add tests - there's nothing new to test here.

PR #2830, opened 2017-12-16
[FIX] Impute: Fix state serialization/restore

ales-erjavec:
##### Issue

```
File ---------------> Impute
                     ^
Random forest ------/
```

Change some  individual variable imputer to *Model based (Random forest)*

Remove the *Impute* widget. Create a new *Impute* widget and connect it to *File* again (without connecting it to any learner):
```
File ---> Impute
```
with the same file opened in *File*. The *Impute* widget restores the variables' state as *model  (Random forest)* but there is no Random forest input.

The problem is that Impute stores actual constructed `Imputer` instances in its settings.

##### Description of changes

Explicitly model what the widget does and not only its (by) products.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owimpute.py
+++ b/Orange/widgets/data/owimpute.py
@@ -1,10 +1,11 @@
 import sys
 import copy
 import logging
+import enum
 import concurrent.futures
 from concurrent.futures import Future  # pylint: disable=unused-import
-from collections import namedtuple
-from typing import List, Any  # pylint: disable=unused-import
+from collections import namedtuple, OrderedDict
+from typing import List, Any, Dict, Tuple, Type, Optional # pylint: disable=unused-import
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owimpute.py
```diff
@@ -416,38 +468,26 @@ def onDeleteWidget(self):
...
-                return AsDefault, (method.method,)
-            elif isinstance(method, impute.Model):
-                return impute.Model, (method.learner,)
-            elif isinstance(method, impute.Default):
-                return impute.Default, (method.default,)
-            else:
-                return type(method), None
-
-        methods = set(method_key(m) for m in methods)
+        variables = [self.varmodel[i.row()] for i in indexes]
```
lanzagar: This appears to be unused,
and is the same as `selected_vars` below.

PR #2831, opened 2017-12-16
[FIX] Bincount: Fix crash on array with all nans

pavlin-policar:
##### Issue
Utility version of `bincount` crashed on sparse vector of all NaNs. 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


pavlin-policar:
Thanks! I've had to change the `dense_sparse` decorator so it would, in addition to the previous ways it ran tests, also run test for sparse array with no explicit zeros. Otherwise any test I wrote for `bincount` would be pointless, since it only crashed when the only value in the array was NaN.

nikicc:
@pavlin-policar looks good to me.

PR #2836, opened 2017-12-19
ENH: Add pandas_compat.table_from_frame(df)

kernc:
##### Issue
Various third parties work with pandas.DataFrames. Missing was a convenient method of transforming result DataFrame into Orange Table.

##### Description of changes
Add `table_from_frame` utility function. No hard pandas dependencies. Please advise on the best way to run the tests.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/pandas_compat.py
+++ b/Orange/data/pandas_compat.py
@@ -1,0 +1,78 @@
+"""Pandas DataFrame↔Table conversion helpers"""
+import numpy as np
+import pandas as pd
+from pandas.api.types import (
+    is_categorical_dtype, is_object_dtype,
+    is_datetime64_any_dtype, is_numeric_dtype,
+)
+
+from Orange.data import (
+    Table, Domain, DiscreteVariable, StringVariable, TimeVariable,
+    ContinuousVariable,
+)
+
+__all__ = ['table_from_frame']
+
+
+def table_from_frame(df, *, force_nominal=False):
+    """
+    Convert pandas.DataFrame to Orange.data.Table
+
+    Parameters
+    ----------
+    df : pandas.DataFrame
+    force_nominal : boolean
+        If True, interpret ALL string columns as nominal (DiscreteVariable).
+
+    Returns
+    -------
+    Table
+    """
+
+    def _is_discrete(s):
+        return (is_categorical_dtype(s) or
+                is_object_dtype(s) and (force_nominal or
+                                        s.nunique() < s.size**.666))
+
+    def _is_datetime(s):
+        if is_datetime64_any_dtype(s):
+            return True
+        try:
+            if is_object_dtype(s):
+                pd.to_datetime(s, infer_datetime_format=True)
+                return True
+        except Exception:  # pylint: disable=broad-except
+            pass
+        return False
+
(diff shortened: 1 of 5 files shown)


astaric:
AFAIAC pandas can be installed on travis (and appveyor) to run the tests while not being added to the install_requires listed in setup.py.

I assume this is how pandas tests HDFS support as there are tests in their repository and tables listed in optional-requirements, while it does not get installed when you run `pip install pandas`.

Orange/data/tests/test_pandas.py
```diff
@@ -0,0 +1,12 @@
...
+try:
+    import pandas
+except ImportError:
+    pandas = None
+
+@unittest.skipIf(pandas is None, "Missing package 'pandas'")
+class TestPandasCompat(unittest.TestCase):
+    def test_table_from_frame(self):
+        from Orange.data.pandas_compat import Test
+        Test().test_table_from_frame()
```
astaric: Why don't you move the body of the test here?

Orange/data/pandas_compat.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Returns
+    -------
+    Table
+    """
+
+    def _is_discrete(s):
+        return (is_categorical_dtype(s) or
+                is_object_dtype(s) and (force_nominal or
+                                        s.nunique() < s.size**.666))
```
astaric: I guess you do not want to use [is_discrete_values](https://github.com/biolab/orange3/blame/master/Orange/data/variable.py#L33) here as it is too complicated, but you could at least use the same constant :)

Orange/data/pandas_compat.py
```diff
@@ -0,0 +1,81 @@
...
+            s = pd.to_datetime(s, infer_datetime_format=True)
+            X.append(s.astype('str').replace('NaT', np.nan).map(tvar.parse).values)
+        elif is_numeric_dtype(s):
+            attrs.append(ContinuousVariable(name))
+            X.append(s.values)
+        else:
+            metas.append(StringVariable(name))
+            M.append(s.values.astype(object))
+
+    MAX_LENGTH = max(len(X[0]) if X else 0,
```
astaric: Is this ever going to be different than df.shape[0] (or equivalent)?

Orange/data/pandas_compat.py
kernc: 👍

It's the exact same [universal] constant, just computed with more precision. 😛

Orange/data/pandas_compat.py
kernc: Constructing Table objects, one sometimes gets so invested with details that they tend to lose grasp of the whole picture.

Thanks.

astaric:
The code look ok to me. Squash if you intended to and I'll merge.

Since you ticked the documentation checkbox, should this be mentioned in the docs somewhere? I do not have a specific place in mind so the aswer to this could be "no", or "not yet", but I think it is a useful helper more people should know about :)

kernc:
Was referring to reasonable docstring as provided. Guessing reviewers should be the ones to catch it until DataFrames become a more frequently integrated pattern.

PR #2841, opened 2017-12-22
[ENH] Canvas: Add zoom

janezd:
##### Issue

Zooming is currently rather rigid. There is also no zooming out.

##### Description of changes

Add zooming by ⌘+scroll; zooming is continuous on trackpads, and discrete with mouse.
Add zooming by ⌘+ and ⌘- ,and reset by ⌘0.
Remove the zoom toggle button from the toolbar.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/canvas/view.py
+++ b/Orange/canvas/canvas/view.py
@@ -2,10 +2,11 @@
 Canvas Graphics View
 """
 import logging
+from math import copysign
 
 from AnyQt.QtWidgets import QGraphicsView
-from AnyQt.QtGui import QCursor, QIcon
-from AnyQt.QtCore import Qt, QRect, QSize, QRectF, QPoint, QTimer
+from AnyQt.QtGui import QCursor, QIcon, QTransform, QWheelEvent
+from AnyQt.QtCore import QT_VERSION, Qt, QRect, QSize, QRectF, QPoint, QTimer
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


janezd:
@ales-erjavec, can you take a quick look?

Orange/canvas/canvas/view.py
```diff
@@ -2,10 +2,11 @@
 Canvas Graphics View
 """
 import logging
+from math import floor, copysign
```
ales-erjavec: floor is unused

Orange/canvas/canvas/view.py
```diff
@@ -63,6 +66,32 @@ def mouseReleaseEvent(self, event):
...
+    def __set_zoom(self, scale):
+        self.__scale = min(15, max(scale, 3))
+        transform = QTransform()
+        transform.scale(self.__scale / 10, self.__scale / 10)
+        self.setTransform(transform)
+
+    def wheelEvent(self, event: QWheelEvent):
+        # use mouse position as anchor while zooming
+        self.setTransformationAnchor(2)
+        if event.modifiers() & Qt.ControlModifier:
```
ales-erjavec: ` ... and event.buttons() == Qt.NoButtons:`

Orange/canvas/canvas/view.py
```diff
@@ -63,6 +66,32 @@ def mouseReleaseEvent(self, event):
...
+            delta = event.angleDelta().y()
+            if QT_VERSION >= 0x050300 \
+                    and event.source() != Qt.MouseEventNotSynthesized \
```
ales-erjavec: [http://doc.qt.io/qt-5/qwheelevent.html#source](http://doc.qt.io/qt-5/qwheelevent.html#source) says it was introduced in Qt 5.5.

janezd: I changed this to 0x050500. The [documentation for the corresponding enum](http://doc.qt.io/qt-5/qt.html#MouseEventSource-enum) says that the enum exists from 5.3. I don't know how would you get the source then.

PR #2843, opened 2017-12-22
Add a keyboard shortcut to delete a widget with "backspace"

borondics:
CMD-backspace on Mac is a bit awkward to use

### DIFF
--- a/Orange/canvas/document/schemeedit.py
+++ b/Orange/canvas/document/schemeedit.py
@@ -295,13 +295,9 @@ class SchemeEditWidget(QWidget):
                     triggered=self.showSettings,
                     enabled=False)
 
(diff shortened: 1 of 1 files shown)


markotoplak:
I support this. I think Backspace is not used anywhere on canvas so it should be free. 

@BlazZupan , @janezd ?

astaric:
+1

kernc:
Does Backspace delete selected files in Finder?

lanzagar:
OK, with me. Maybe even make backspace default on all platforms and remove the reversing that follows these shortcut definitions.

borondics:
@kernc no, it doesn't, the default there is cmd-backspace.

PR #2852, opened 2018-01-05
[ENH] Distances: Optimize PearsonR/SpearmanR

ales-erjavec:
##### Issue

* PearsonR is poorly implemented by a double for loop in python.
* SpearmanR quadruples it's workload only to throw most of it away. 

##### Description of changes

* Use `numpy.corrcoef` for PearsonR
* Implement a variants of corrcoef/spearmanr which compute correlations between two arrays more efficiently (without also computing all correlations within the two arrays).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/distance/distance.py
+++ b/Orange/distance/distance.py
@@ -425,14 +425,13 @@ class CorrelationDistanceModel(DistanceModel):
             return (1. - rho) / 2.
 
     def compute_correlation(self, x1, x2):
-        pass
+        raise NotImplementedError()
 
 
 class SpearmanModel(CorrelationDistanceModel):
     def compute_correlation(self, x1, x2):
-        n1 = x1.shape[1 - self.axis]
-        n2 = x2.shape[1 - self.axis] if x2 is not None else 0
         if x2 is None:
+            n1 = x1.shape[1 - self.axis]
             if n1 == 2:
                 # Special case to properly fill degenerate self correlations
                 # (nan, inf on the diagonals)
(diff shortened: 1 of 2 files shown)


Orange/distance/distance.py
```diff
@@ -418,25 +418,128 @@ def __init__(self, absolute, axis=1, impute=False):
...
+            n1 = x1.shape[1 - self.axis]
+            if n1 == 2:
+                # Special case to properly fill degenerate self correlations
+                # (nan, inf on the diagonals)
+                rho = stats.spearmanr(x1, x1, axis=self.axis)[0]
+                assert rho.shape == (4, 4)
+                rho = rho[:2, :2].copy()
+            else:
+                # scalar if n1 == 1
+                rho = stats.spearmanr(x1, axis=self.axis)[0]
```
thocevar: Are these two cases (if, else) necessary? At first glance stats.spearmanr seems to (efficiently) handle the case of a missing second attribute.

thocevar: Never mind.

ales-erjavec:
```python
In [1]: import Orange, numpy

In [2]: d = Orange.data.Table(numpy.random.random(size=(200, 200)))

In [3]: %timeit Orange.distance.PearsonR(d)
```
Before
```
1.59 s ± 12.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
After
```
614 µs ± 84.3 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```python
In [1]: import Orange, numpy

In [2]: d = Orange.data.Table(numpy.random.random(size=(400, 200)))

In [3]: %timeit Orange.distance.SpearmanR(d[1:], d[:-1])
```
Before
```
201 ms ± 2.28 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
After
```
47.9 ms ± 1.82 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

PR #2856, opened 2018-01-05
Improve imports

janezd:
##### Issue

Wildcard imports are a bad idea.

##### Description of changes

Import explicitly, or disable warning where such import is appropriate.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -958,28 +958,5 @@ def unique(seq):
     return unique_el
 
 
-def main(argv=None):
-    from AnyQt.QtWidgets import QApplication
-    if argv is None:
-        argv = sys.argv
-    app = QApplication(list(argv))
-    argv = app.arguments()
-    if len(argv) > 1:
-        filename = argv[1]
-    else:
-        filename = "iris"
-
-    w = OWFeatureConstructor()
-    w.show()
-    w.raise_()
-    data = Orange.data.Table(filename)
-    w.setData(data)
-    w.handleNewSignals()
-    app.exec_()
-    w.setData(None)
-    w.handleNewSignals()
-    w.saveSettings()
-    return 0
-
-if __name__ == "__main__":
-    sys.exit(main())
+if __name__ == "__main__":  # pragma: no cover
+    OWFeatureConstructor.test_run(Orange.data.Table("iris"))
--- a/Orange/widgets/data/owselectcolumns.py
+++ b/Orange/widgets/data/owselectcolumns.py
@@ -1,4 +1,3 @@
-import sys
 from functools import partial
 
 from AnyQt.QtWidgets import QWidget, QGridLayout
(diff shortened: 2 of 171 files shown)


janezd:
@astaric, I'm not so sure about this one. See the places where we already disable `wildcard-import` warning: we do this where we pull the names from subpackages into a package, e.g. the learners into `modelling.__init__`. I guess we could allow this, but only in those places, and only if the subpackages specify the list of imported names in `__all__`.

astaric:
I am fine with using * import for pulling members to a package (with __all__ defined in subpackages).

The two test files should be fixed though. And I have no idea what lazy modules are doing in the preprocess.__init__

Orange/widgets/data/owselectcolumns.py
```diff
@@ -418,27 +417,5 @@ def send_report(self):
...
-    from AnyQt.QtWidgets import QApplication
-    if argv is None:
-        argv = sys.argv
-    argv = list(argv)
-    app = QApplication(list(argv))
-
-    if len(argv) > 1:
-        filename = argv[1]
-    else:
-        filename = "brown-selected"
```
janezd: I removed such code since this is a total overkill.

Orange/widgets/data/owfeatureconstructor.py
```diff
@@ -958,28 +958,5 @@ def unique(seq):
...
-
-    w = OWFeatureConstructor()
-    w.show()
-    w.raise_()
-    data = Orange.data.Table(filename)
-    w.setData(data)
-    w.handleNewSignals()
-    app.exec_()
-    w.setData(None)
-    w.handleNewSignals()
```
janezd: The purpose of these `main` functions is to show the widget for visual inspection when they're being implemented or changed. Testing whether the widget survives removing the data belongs to unit tests, not here, hence I removed such code when refactoring.

If somebody feels we should try sending `None` after the widget is (manually) tested, this can be easily reimplemented in `test_run` - but I'd strongly rather not.

astaric:
-1 for cleaning main in widgets. I do not like the assumption that __main__ is only for visually inspecting the widget.

I mostly use them when debugging a problem and that usually includes modifying the code in __main__ (changing order of signals, sending and removing outputs, ... The fact that they are straight forward (although sometimes complicated) is a + for me since I can usually understand what is going on and know how to modify them without much thought.

janezd:
> debugging a problem and that usually includes modifying the code in main (changing order of signals, sending and removing outputs

You should do this in unit tests and keep that code there. We should stop leaving our debug code in widgets. (We also don't leave commented-out prints in the code.) Even if you use `main` for this purpose, you'd have to clean it up afterwards.

I don't want to go through all widgets, move their code into a `main` function, which is going to be exactly the same for all widgets, except for those that retained some code used for debugging some ancient bugs. Just for the case that somebody will perhaps sometimes want to have it.

All this is however beyond the point, you can still run whatever tests with adding and removing signals you want: instead of changing the main function, provide a `_test_run_init` method and send signals and re-send signals there. (Like in https://github.com/biolab/orange3/pull/2856/files#diff-39be72554856e6c399803af6ad541157R881, except that it there sends signals with different id's.)

I don't think there's anything you could do in `main` that you can't do in `_test_run_init`, except for tearing down the widget. If you need this, I can add another method for it.

janezd:
Pylint on my computer wants pkg_resources to be imported *after* AnyQt, so I (hesitantly) moved it. Pylint on Travis now wants pkg_resources *before* AnyQt. Anybody knows what causes the difference?

```
anaconda-client==1.5.5
AnyQt==0.0.8
astroid==1.5.3
pylint==1.8.1
setuptools-git==1.1
```

PR #2865, opened 2018-01-12
Added a simple unittest to Sieve

ajdapretnar:
##### Issue

No test for removing data input.

##### Description of changes
Test added. Also a minor streamlining of tests. Couldn't add 'titanic' to the final one, I reckon file too big.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/tests/test_owsieve.py
+++ b/Orange/widgets/visualize/tests/test_owsieve.py
@@ -2,6 +2,7 @@
 # pylint: disable=missing-docstring
 from math import isnan
 from unittest.mock import patch
+
 import numpy as np
 
 from AnyQt.QtCore import QEvent, QPoint, Qt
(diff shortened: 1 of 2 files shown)


ajdapretnar:
This is ready for another review. @janezd 
I am not sure how to test whether the first two attributes from iris are discrete. The problem is Sieve creates a separate attribute `self.discrete_data` where discretized values are stored. `attr_x` and `attr_y` however, remain continuous.

Perhaps we should consider moving `def discretizer(data)` outside of sparse check? @jerneju

Orange/widgets/visualize/tests/test_owsieve.py
```diff
@@ -22,10 +24,52 @@ def setUpClass(cls):
...
+        self.send_signal(self.widget.Inputs.data, self.titanic)
+        self.assertEqual(self.widget.attr_x, self.titanic.domain.class_var)
+        self.assertEqual(self.widget.attr_y, self.titanic.domain.attributes[0])
+        # Change attributes to two different ones
+        self.widget.attr_x = self.titanic.domain.attributes[1]
+        self.widget.attr_y = self.titanic.domain.attributes[2]
+        # Remove data signal
+        self.send_signal(self.widget.Inputs.data, None)
+        self.assertEqual(self.widget.attr_x, None)
+        self.assertEqual(self.widget.attr_y, None)
```
janezd: Since you are testing `discrete_data` below ... maybe you can add
```
        self.assertIsNone(self.widget.discrete_data)
```

And then you can fix the widget to pass this test. :)

Orange/widgets/visualize/tests/test_owsieve.py
```diff
@@ -22,10 +24,52 @@ def setUpClass(cls):
...
+
+    def test_continuous_data(self):
+        self.send_signal(self.widget.Inputs.data, self.iris)
+        self.assertEqual(self.widget.attr_x, self.iris.domain.class_var)
+        self.assertEqual(self.widget.attr_y, self.iris.domain.attributes[0])
+        self.assertTrue(self.widget.discrete_data)
+
+    def test_few_attributes(self):
+        # Test widget behaviour when data has only a few attributes
+        # Test for 2 attributes
```
janezd: Could you repeat this comment (substituting 1 and 0 for 2) below?

Orange/widgets/visualize/tests/test_owsieve.py
```diff
@@ -22,10 +24,52 @@ def setUpClass(cls):
...
+        # remembered in the settings
+        self.send_signal(self.widget.Inputs.data, self.titanic)
+        self.assertEqual(self.widget.attr_x, self.titanic.domain.attributes[1])
+        self.assertEqual(self.widget.attr_y, self.titanic.domain.attributes[2])
+
+    def test_continuous_data(self):
+        self.send_signal(self.widget.Inputs.data, self.iris)
+        self.assertEqual(self.widget.attr_x, self.iris.domain.class_var)
+        self.assertEqual(self.widget.attr_y, self.iris.domain.attributes[0])
+        self.assertTrue(self.widget.discrete_data)
```
janezd: I'd prefer using `assertIsNotNone` here, since that's what you're actually testing. Yet, this test doesn't do much since `discrete_data` will be non-None whenever `data` is not `None`. Maybe you want to test

        self.assertTrue(self.widget.discrete_data.domain[0].is_discrete)

Orange/widgets/visualize/tests/test_owsieve.py
```diff
@@ -42,13 +86,12 @@ def test_missing_values(self):
         data = Table(Domain(attrs, class_var), X, np.array([np.nan] * 6))
         self.send_signal(self.widget.Inputs.data, data)
 
-    def test_keyerror(self):
+    def test_single_line(self):
```
janezd: +1 for changing the name. You can also remove these GH-xxxx tags. They notify what was fixed when a function or a test was written, but this belongs to commit messages, not here. In particular in the code of the widget, in this case in method `sparse_to_dense`; the fact that the function was written to fix GH-2260 is not relevant for the future.

Orange/widgets/visualize/tests/test_owsieve.py
```diff
@@ -2,6 +2,8 @@
 # pylint: disable=missing-docstring
 from math import isnan
 from unittest.mock import patch
+
+import Orange
```
janezd: Don't import `Orange`. `Domain` and `Table` are already imported directly into this file's namespace (see below).

Also, when you do import `Orange`, it has to come after all third-party imports, in this case, it's after `AnyQt`. Otherwise it will trigger one or two pylint warnings (module import order and ungrouped imports).

Orange/widgets/visualize/tests/test_owsieve.py
```diff
@@ -22,10 +24,52 @@ def setUpClass(cls):
...
+        self.assertEqual(self.widget.attr_x, self.iris.domain.class_var)
+        self.assertEqual(self.widget.attr_y, self.iris.domain.attributes[0])
+        self.assertTrue(self.widget.discrete_data)
+
+    def test_few_attributes(self):
+        # Test widget behaviour when data has only a few attributes
+        # Test for 2 attributes
+        attr2 = self.titanic.domain[:2]
+        domain2 = Orange.data.Domain(attr2)
+        data2 = Orange.data.Table(domain2, self.titanic)
```
janezd: You can use just `Table` and `Domain` here. They are already imported.

ajdapretnar:
Ok, I fix.

Orange/widgets/visualize/tests/test_owsieve.py
janezd: Heh, now I checked what travis/pylint wrote.

```
Orange/widgets/visualize/tests/test_owsieve.py:7: [C0411(wrong-import-order), ] third party import "import numpy as np" should be placed before "import Orange"
Orange/widgets/visualize/tests/test_owsieve.py:12: [C0412(ungrouped-imports), ] Imports from package Orange are not grouped
```

I became a bit of an expert for pylint's messages about this while trying to push a pull request that fixed the imports. :)

ajdapretnar:
Fixed.

PR #2866, opened 2018-01-12
[FIX] Add labels to combo when data comes from distance matrix

### LINKED ISSUE #2863: MDS does not show labels
##### Orange version

3.8

##### Expected behavior

Connect File Distances -> MDS, load a files that has labels, e.g. slovenian-towns.dst.

- MDS allows the user to label the points with the labels. 
- Labels are default choice in the combo (not "None")

##### Actual behavior

Labels can't be selected.

See above. Note that this happens only if the data comes from File Distance, not File -> Distance.

janezd:
Fixes #2863.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -216,8 +216,10 @@ class OWMDS(OWWidget):
         g = self.graph.gui
         box = g.point_properties_box(self.controlArea)
         self.models = g.points_models
-        self.models[2].order = \
-            self.models[2].order[:1] + ("Stress", ) + self.models[2].order[1:]
+        self.size_model = self.models[2]
+        self.label_model = self.models[3]
+        self.size_model.order = \
+            self.size_model.order[:1] + ("Stress", ) + self.models[2].order[1:]
 
         gui.hSlider(box, self, "connected_pairs", label="Show similar pairs:", minValue=0,
                     maxValue=20, createLabel=False, callback=self._on_connected_changed)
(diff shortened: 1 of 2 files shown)


Orange/widgets/unsupervised/owmds.py
```diff
@@ -265,13 +267,18 @@ def update_regression_line(self):
...
             model.set_domain(domain)
         self.graph.attr_color = self.data.domain.class_var if domain else None
         self.graph.attr_shape = None
         self.graph.attr_size = None
         self.graph.attr_label = None
+        if domain is not None:
+            for var in domain.metas:
+                if var.is_string:
+                    self.graph.attr_label = var
+                    break
```
lanzagar: In addition to fixing labels for distance matrices, this PR changes some default behaviour in the same commit.
And I am not so sure we even want to auto-set labels to the first string var. The labels a user would want are just as often in a categorical var. But even if the first string var does contain instance labels, we might not want to show them unless explcitly selected, because of our notorious label-overlapping problems. It will not look nice for any dataset with a couple hundred instances (e.g. brown-selected has short labels, <200 instances, shows the correct thing, and I find it borderline ok)

janezd: I'd guess that showing string labels (as opposed to showing nothing) is going to minimize the expected number of mouse clicks. If I have string labels - and I'll often have a single one, I'd usually want to see them.

For class labels, we already have colors. For attribute labels, we can't guess which one to show and we'll usually miss.

@BlazZupan?

But I agree that we shouldn't show labels by default if there are too many. (You already said this on Friday, but I forgot.) Should I add an `and len(data) <= 200` above?

lanzagar: A popular thing from text mining, which we showed several times in workshops (!), is to load a corpus (use the default book-excerpts), connect to Bag of Words and then to MDS to show a map of documents.
Try that after this PR ;)

I am still in favour of not showing any labels automatically, but if I had to, then in addition to `len(data)<=200` I would add that the max len of all string values from the variable should be <12 or something like that.

janezd:
I removed default labels. I haven't squashed the commits, so we can bring them back (now or in the future).

PR #2867, opened 2018-01-12
Add Domain.empty, add warning to Domain.__bool__

janezd:
**Based on #2866 (1 commit). You may want to merge that request first.**

##### Issue

#2866 fixes an issue with labels in MDS that occured since `self.data.domain` was `False` for a `Domain` with only meta attributes. This occurs since `Domain.__len__` (correctly) returns the number of elements that `Domain.__iter__` iterates across, however, `Domain.__iter__` (incorrectly) ignores metas.

The issue will remain even after we fix `Domain.__iter__`: when checking whether a domain is true, we usually mean whether it's not `None`.

##### Description of changes

`Domain` has a new method, `empty`.

`__bool__` warns that it's ambiguous and suggests using either `empty` or `is None`. The behaviour of `__bool__` is unchanged - it returns `True` if the lenght of the domain is non-zero.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/domain.py
+++ b/Orange/data/domain.py
@@ -1,3 +1,4 @@
+import warnings
 import weakref
 
 from math import log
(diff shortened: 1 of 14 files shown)


lanzagar:
Rebase please.

PR #2868, opened 2018-01-12
Updated documentation + new template

### LINKED ISSUE #1705: Outdated screenshots
This issue is intended for keeping track of GUI modifications in widgets, which require the screenshots to be updated and the documentation to be reviewed.
**Note**: Documentation (text) should be kept up to date and reflect the current features. This is only to avoid frequent and unnecessary retaking of screenshots for a widget that is being changed (perhaps in multiple PRs).

The proposed plan is to do a batch update before each new major release.

ajdapretnar:
##### Issue

Fixes #1705.

##### Description of changes
Documented new functionalities fo Box Plot and Preprocess (#1705).
Added description of preprocessors to Preprocess.
Implemented new templated as discussed with @janezd.
Fixed a minor bug in Radviz doc.

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation


janezd:
I may have been too radical on Friday when simplifying the template. :) It indeed does not work when the number of signals is too large.

What about

screen shot 2018-01-14 at 10 49 02

It's still an improvement from

screen shot 2018-01-14 at 10 52 09

To achieve this in .rst, I used

```
.. |br| raw:: html

   <br />
   
Scatter Plot
============

Scatterplot visualization with explorative analysis and intelligent data
visualization enhancements.

-----------------

**Inputs**:
    **Data**: input data set |br|
    **Data Subset**: subset of instances |br|
    **Features**: list of attributes

**Outputs**:
    **Selected Data**: instances selected from the plot |br|
    **Data**:  data with an additional column showing whether a point is selected

-----------------
```

janezd:
However, why we don't use .md for widget documentation? It's friendlier, it has better tools for editing (e.g. a real instant preview in Visual Studio Code and, I suppose, Sublime), and it can be mixed with html without hassle.

We already have `recommonmark` in our requirements, it's just not enabled in sphinx config.

astaric:
I like the layout in https://github.com/biolab/orange3/pull/2868#issuecomment-357500664, could it be achieved in css instead of writing |br|? Does sphinx support custom css?

Regarding markdow, can it do automatic linking to class/function documentation? I think that was the reason our md files were converted back to rst.

ajdapretnar:
I also like the second suggestion.
And yes, I think there was a reason why we switched to .rst, but I don't remember specifically. Anže might be on it.

ajdapretnar:
P.s. We need to keep in mind widget icons are used for Widget catalog on our website. This, of course, doesn't mean keeping them in the documentation but keeping them in the repository. I think. Can @astaric confirm?

doc/visual-programming/source/widgets/visualize/treeviewer.rst
```diff
@@ -1,31 +1,18 @@
...
+Inputs
+    Tree
+        decision tree
 
-   Data set with an additional attribute for selection labels.
+Outputs
+    Selected Data
+        instances selected from the tree node
+    Data
+        data with an additional column showing whether a point is selected
```
janezd: Consider renaming *point* to *instance* in this case. Also in Silhouette plot. (Point is also OK, as in "data point". You decide.)

janezd:
Copy them to `static` and change links in the catalog?

ajdapretnar:
For now, I am not touching this. This has been broken by merging Data Sets to Datasets PR and now this documentation is practically obsolete. So decide what to do and advise how to fix this mess.

doc/visual-programming/source/widgets/visualize/treeviewer.rst
ajdapretnar: Nope, this was accidentally not fixed when copy-pasting. As said - lots of editing.

astaric:
I have rebased this PR onto the latest master, this fixes the conflicts.

Regarding the icons, I would keep them as they are, unless someone feels like rewriting the [build_widget_catalog script](https://github.com/biolab/orange3/blob/master/doc/visual-programming/build_widget_catalog.py).

ajdapretnar:
Great, thanks @astaric! Probably someone should give it a quick look and merge.

markotoplak:
I saved the branch prior to rebase so until we merge this branch we can still decide to revert the "Data set" to "Dataset" conversion.

astaric:
I have updated (replaced) create_widget_catalog script with one that does not parse icons from documentation, but instead constructs widget registry and gets information from there.

The scripts generates all the files on travis, I'll check if they get uploaded to the website after the merge build completes.

astaric:
Everything works after upload.

I manually restarted the webserver, since widgets.json file seems to be cached, while old images were removed with the upload of the latest documentation.

PR #2877, opened 2018-01-17
[WIP] gui.py: Simplify and refactor into manageable modules

janezd:
##### Issue

gui.py is 3300 lines of mess.

##### Description of changes

This PR will split it into sensible maintainable modules, remove the things that are no longer (or have never been) used, and refactor the code to break its probably compatibility with Python 2.2 (for which it was originally written).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/data/owconcatenate.py
```diff
@@ -110,7 +110,7 @@ def __init__(self):
 
         form.addRow(
             self.tr("Feature name:"),
-            gui.lineEdit(ibox, self, "source_attr_name", valueType=str,
+            gui.lineEdit(ibox, self, "source_attr_name",
```
janezd: This is (was) the default.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -285,7 +285,6 @@ def __init__(self):
         self.cbox = gui.vBox(self.controlArea, "Target Class")
         self.class_selection_combo = gui.comboBox(
             self.cbox, self, "class_selection", items=[],
-            sendSelectedValue=True, valueType=str,
```
janezd: `sendSelectedValue=True` is now detected automatically from the value type.

Orange/widgets/data/owfile.py
```diff
@@ -158,8 +158,7 @@ def __init__(self):
 
         self.sheet_box = gui.hBox(None, addToLayout=False, margin=0)
         self.sheet_combo = gui.comboBox(None, self, "xls_sheet",
-                                        callback=self.select_sheet,
-                                        sendSelectedValue=True,)
```
janezd: `sendSelectedValue=True` is now detected automatically from the value type.

Orange/widgets/gui.py
```diff
@@ -886,10 +887,13 @@ def lineEdit(widget, master, value, label=None, labelWidth=None,
     if validator:
         ledit.setValidator(validator)
     if value:
+        if valueType is None:
+            current_value = getdeepattr(master, value)
+            valueType = str if current_value is None else type(current_value)
```
janezd: Revert to the old default if the `current_value` is `None`. Otherwise, the `valueType` must match the type of `current_value`; if any widget kept the default `valueType` (`str`) but `current_value` was something else (that is, `int` or `float`), the widget was already broken.

Orange/widgets/gui.py
```diff
@@ -886,10 +887,13 @@ def lineEdit(widget, master, value, label=None, labelWidth=None,
...
         ledit.cback = connectControl(
             master, value,
             callbackOnType and callback, ledit.textChanged[str],
-            CallFrontLineEdit(ledit), fvcb=value and valueType)[1]
```
janezd: `value and` is not needed -- see the `if` above.

Orange/widgets/gui.py
```diff
@@ -1545,21 +1549,17 @@ def minimumSizeHint(self):
...
-# TODO comboBox looks overly complicated:
-# - can valueType be anything else than str?
-# - sendSelectedValue is not a great name
 def comboBox(widget, master, value, box=None, label=None, labelWidth=None,
              orientation=Qt.Vertical, items=(), callback=None,
-             sendSelectedValue=False, valueType=str,
-             emptyString=None, editable=False,
+             sendSelectedValue=None, emptyString=None, editable=False,
              contentsLength=None, maximumContentsLength=25,
+             *, model=None,
```
janezd: Better than getting it from `misc`.

Orange/widgets/gui.py
```diff
@@ -1545,21 +1549,17 @@ def minimumSizeHint(self):
         return sh
 
 
-# TODO comboBox looks overly complicated:
-# - can valueType be anything else than str?
-# - sendSelectedValue is not a great name
 def comboBox(widget, master, value, box=None, label=None, labelWidth=None,
              orientation=Qt.Vertical, items=(), callback=None,
-             sendSelectedValue=False, valueType=str,
```
janezd: I didn't find anybody specifying a `valueType` other than `str`. This commit removes this argument from calls (in core widgets). If any other widget specifies `valueType`, it will now be silently ignored in `misc`.

Orange/widgets/gui.py
```diff
@@ -2391,25 +2394,31 @@ def action(self, value):
...
+
+        def action_int():
+            if value < combo.count():
+                combo.setCurrentIndex(value)
+            else:
+                log.warning("Unable to set '{}' to {}; largest index is {}".
+                            format(combo, value, combo.count() - 1))
+
+        combo = self.control
+        if isinstance(value, int):
```
janezd: This allows using both, integers and strings. I'll probably split this class in two in further commits; if `valueType` is `str`, the widget should not set the corresponding attribute to an `int`. The current implementation, however, does not hurt, it just allows broken widgets to work.

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -138,13 +138,13 @@ def __init__(self):
         layout.addWidget(sb, 0, 1)
         gui.lineEdit(
             sb, self, "n_init", controlWidth=60,
-            valueType=int, validator=QIntValidator(), callback=self.invalidate)
```
janezd: `valueType` is now detected automatically.

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -138,13 +138,13 @@ def __init__(self):
...
+            validator=QIntValidator(), callback=self.invalidate)
         layout.addWidget(
             gui.widgetLabel(None, "Maximum iterations: "), 1, 0, Qt.AlignLeft)
         sb = gui.hBox(None, margin=0)
         layout.addWidget(sb, 1, 1)
         gui.lineEdit(
-            sb, self, "max_iterations", controlWidth=60, valueType=int,
```
janezd: `valueType` is now detected automatically.

Orange/widgets/utils/plot/owplotgui.py
```diff
@@ -659,8 +659,8 @@ def alpha_value_slider(self, widget):
     def _combo(self, widget, value, label, cb_name, items=(), model=None):
         gui.comboBox(widget, self._plot, value, label=label, items=items,
                      model=model, callback=self._get_callback(cb_name),
-                     labelWidth=50, orientation=Qt.Horizontal, valueType=str,
-                     sendSelectedValue=True, contentsLength=12)
```
janezd: `valueType` and `sendSelectedValue` is now detected automatically.

Orange/widgets/utils/plot/owplotgui.py
```diff
@@ -853,7 +853,7 @@ def effects_box(self, widget, box=None):
 
     def theme_combo_box(self, widget):
         c = gui.comboBox(widget, self._plot, "theme_name", "Theme",
-                         callback=self._plot.update_theme, sendSelectedValue=1, valueType=str)
```
janezd: `valueType` and `sendSelectedValue` is now detected automatically.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -209,7 +209,7 @@ def __init__(self):
         self.groupvarview = gui.comboBox(
             box, self, "groupvar_idx",
             callback=self._on_groupvar_idx_changed,
-            valueType=str, contentsLength=12)
```
janezd: This `valueType` is wrong, but `gui.comboBox` ignored it anyway since `sendSelectedValue` was left `False`.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -321,8 +321,7 @@ def __init__(self):
             gui.comboBox(
                 box, self, value="variable{}".format(i),
                 orientation=Qt.Horizontal, contentsLength=12,
-                callback=self.reset_graph,
-                sendSelectedValue=True, valueType=str, emptyString="(None)")
```
janezd: `valueType` and `sendSelectedValue` is now detected automatically.

However: this widget must be converted to using models.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -335,8 +334,7 @@ def __init__(self):
         self.cb_attr_color = gui.comboBox(
             box2, self, value="variable_color",
             orientation=Qt.Horizontal, contentsLength=12, labelWidth=50,
-            callback=self.set_color_data,
-            sendSelectedValue=True, model=self.color_model, valueType=str)
```
janezd: `sendSelectedValue` and `valueType` are ignored in presence of a model.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -156,21 +156,18 @@ def __init__(self):
         self.__pending_selection_restore = self.selection_group
         self.selection_group = None
 
-        common_options = dict(
-            labelWidth=50, orientation=Qt.Horizontal, sendSelectedValue=True,
-            valueType=str)
```
janezd: `sendSelectedValue` and `valueType` are ignored in presence of a model.

Orange/widgets/visualize/owsieve.py
```diff
@@ -105,8 +105,7 @@ def __init__(self):
         self.domain_model = DomainModel(valid_types=DomainModel.PRIMITIVE)
         combo_args = dict(
             widget=self.attr_box, master=self, contentsLength=12,
-            callback=self.update_attr, sendSelectedValue=True, valueType=str,
-            model=self.domain_model)
```
janezd: `sendSelectedValue` and `valueType` are ignored in presence of a model.

Orange/widgets/visualize/owtreeviewer2d.py
```diff
@@ -410,7 +410,7 @@ def __init__(self):
             gui.comboBox(box, self, 'max_tree_depth',
                          items=["Unlimited"] + [
                              "{} levels".format(x) for x in range(2, 10)],
-                         addToLayout=False, sendSelectedValue=False,
```
janezd: This was and is the default.

kernc:
So what if, as an option, instead of fixing an obviously unsalvageable mess, one'd left _gui_ stable as it is and instead started work on a better _gui2_ catering to all the insight gathered from the former and over the years?

Orange/widgets/gui/_checkbox.py
```diff
@@ -0,0 +1,68 @@
...
+    cbox = QtWidgets.QCheckBox(label, b)
+
+    if labelWidth:
+        cbox.setFixedSize(labelWidth, cbox.sizeHint().height())
+    cbox.setChecked(getdeepattr(master, value))
+
+    values = [Qt.Unchecked, Qt.Checked, Qt.PartiallyChecked]
+    connectControl(
+        master, value, None, cbox.toggled[bool],
+        cfront=lambda val: val is not None and cbox.setCheckState(values[val]),
```
janezd: Callfront classes were needed just to store all needed variables. Closures are a better place for this. Callfront classes will thus be replaced with simple lambdas like this one here (or with functions, where necessary).

(Another reason for call front classes, besides storing variables, was that they prevented cyclic callbacks. `connectControl` will take care of this by wrapping these lambdas into classes, but we don't need to know about that here -- this belongs to `connectControl`.)

janezd:
> So what if, as an option, instead of fixing an obviously unsalvageable mess, one'd left gui stable as it is and instead started work on a better gui2, catering to all the insight gathered from the former and over the years?

The thought strongly crossed my mind and I'll keep it there.

But I'll try to refactor the existing gui first. I don't think it's unsalvageable. Functions have too many weird and unused arguments, some of which can be removed or obsoleted. Otherwise the code is just bloated and disorganized. And poorly written (often my fault; Blaz and I started this as Python newbies).

astaric:
Could changes to widgets be done in a separate pull request? I would prefer to have the gui changes backward compatible, at least for a while.

(Breaking changes create a mess, old versions of add-ons will not work with new version of Orange and vice versa)

janezd:
All changes will be backward compatible in the sense that they won't break anything that is not already broken. And I'll be careful with the latter, too.

I agree that not making any changes to widgets is a good way to ensure this. I moved the changes that I made so far into another branch.

PR #2842, opened 2017-12-22
[REF] Rename Data Sets to Datasets

jerneju:
##### Issue
Fixes #2762 .

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #2882, opened 2018-01-19
Revert "[REF] Rename Data Sets to Datasets"

janezd:
Reverts biolab/orange3#2842.

There seems no consensus in the community, with native speakers leaning towards "data sets". We thus don't gain anything by changing this, except for outdating screenshot, manuals, tutorials...

We currently consistently write it separately in all materials we publish, so let's keep it this way.

### DIFF
--- a/Orange/canvas/application/workflows/110-file-and-data-table-widget.ows
+++ b/Orange/canvas/application/workflows/110-file-and-data-table-widget.ows
@@ -8,10 +8,10 @@
 		<link enabled="true" id="0" sink_channel="Data" sink_node_id="1" source_channel="Data" source_node_id="0" />
 	</links>
 	<annotations>
-		<text font-family="Helvetica" font-size="14" id="0" rect="(17.0, 21.0, 160.0, 55.0)">A File widget. Double click to open it and select the dataset file.</text>
+		<text font-family="Helvetica" font-size="14" id="0" rect="(17.0, 21.0, 160.0, 55.0)">A File widget. Double click to open it and select the data set file.</text>
 		<text font-family="Helvetica" font-size="14" id="1" rect="(50.0, 229.0, 118.0, 40.0)">The output of the File widget.</text>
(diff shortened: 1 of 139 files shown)


kernc:
_Almost_ consistently. But there is no pain in it. Natural languages are alive and common pairs of noun words have a tendency to, sometimes, first get hyphenated, then merged. Let's wait another decade or so.

jerneju:
Kaggle uses _datasets_: https://www.kaggle.com/datasets .

jerneju:
**Google** uses _datasets_: https://cloud.google.com/public-datasets/ .
**Amazon** uses _datasets_: https://aws.amazon.com/public-datasets/ .

janezd:
Sure does, yes. Consistently. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-public-data-sets.html

Would never ever write it separately on this page: https://cloud.google.com/bigquery/public-data/

Would somebody please fix the Wikipedia page. It's all wrong! Outdated. https://en.wikipedia.org/wiki/Data_set. SAD!

My point is exactly that you can write it like this or like that, thanks. If both are right, there is totally no reason to make Orange inconsistent with itself and all existing materials we have. 

Let me answer what @astaric will now most surely remark now (so he does not have to bother doing it). This renaming makes no sense, it does not achieve anything, except for presumably following the presumed trend. In this respect this is not the same as renaming discrete to categorical and continuous to numeric: I asked my wife. She had no clue about discrete (and confused it with discreet because we talked in Slovene where the word is the same), and continuous variable is a variable that is measure regularly. So: she missed, both term. When asked about numerical and categoric, she described the former as "numbers" and the latter as, say, gender. Correct, both. Continuous and discrete made no sense for most people, and numeric and categorical do, for most people. This is why renaming variable types made sense, although it introduced inconsistence. Renaming data sets does not serve any purpose. Period.

I had a senate meeting on Tuesday. Wasted three hours repeating the same thing for the n-th time (for the m-th month in a row). Left totally pissed; told the dean to finish the session or I'm leaving. And perhaps never coming back.

I'm not wasting any more time on this, either.

PR #2883, opened 2018-01-19
Add-ons: show warning on insufficient permissions

markotoplak:
##### Issue
Installing add-ons as a normal user raises PermissionErrors on Windows 

##### Description of changes
Removed user_install option (pip's --user) because:
- it does not work with conda installers
- possible clashes between different installers
- the user_install check did not work in some cases (system-wide Win7
  conda install)

Now Orange checks if it can write in site-packages folder. If it can not, it shows a warning instead of the add-on dialog.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/canvasmain.py
+++ b/Orange/canvas/application/canvasmain.py
@@ -1595,7 +1595,13 @@ class CanvasMainWindow(QMainWindow):
             self.__update_from_settings()
 
     def open_addons(self):
-        from .addons import AddonManagerDialog
+        from .addons import AddonManagerDialog, have_install_permissions
+        if not have_install_permissions():
+            QMessageBox(QMessageBox.Warning,
+                        "Add-ons: insufficient permissions",
+                        "Insufficient permissions to install add-ons. Try starting Orange "
+                        "as a system administrator or install Orange in user folders.",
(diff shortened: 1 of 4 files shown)


kernc:
👍 on test and warning. 👎 on removed `--user`. Now @lanzagar can no longer install add-ons via the manager.

astaric:
He can, if he manages to run orange as root

Orange/canvas/application/canvasmain.py
```diff
@@ -1595,7 +1595,13 @@ def open_canvas_settings(self):
             self.__update_from_settings()
 
     def open_addons(self):
-        from .addons import AddonManagerDialog
+        from .addons import AddonManagerDialog, have_install_permissions
+        if not have_install_permissions():
+            QMessageBox(QMessageBox.Warning,
+                        "Add-ons: insufficient permissions",
+                        "Insufficient permissions to install add-ons. Try starting Orange" +\
```
lanzagar: The "+" is not needed, but a space in one of the two strings is!

lanzagar:
I wouldn't want to run Orange as root, but adding a parameter in Settings is acceptable :)

PR #2884, opened 2018-01-22
[ENH] Scatter Plot: do not calculate jittered data when not necessary

### LINKED ISSUE #2881: Scatterplot: jittering is too slow
##### Orange version
3.8.0

##### Expected behavior
Jittering should only be calculated if _Jitter numeric values_ is checked.

##### Actual behavior
Each tick on _Jittering_ slider in Scatter Plot takes a lot of time with large datasets even when _Jitter numeric values_ is unchecked.

jerneju:
##### Issue
Fixes #2881 .

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
This is not the correct solution. It solves dragging the jittering slider when only continuous variables are shown, but if a user changed the variables shown, the jittering is out-of-sync with the view (the view will still contain the old "unchanged" jitter.

This should be fixed differently. We have to move jittering out of scaling and it should be only done for currently shown features.

PR #2885, opened 2018-01-23
[ENH] canvas: Open dropped ows files

astaric:
##### Issue
When an ows file is dropped on canvas window, nothing happens.

##### Description of changes
When an ows file is dropped on canvas window, it is opened.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/canvasmain.py
+++ b/Orange/canvas/application/canvasmain.py
@@ -19,7 +19,8 @@ from AnyQt.QtWidgets import (
 from AnyQt.QtGui import QColor, QIcon, QDesktopServices, QKeySequence
 
 from AnyQt.QtCore import (
-    Qt, QEvent, QSize, QUrl, QTimer, QFile, QByteArray, QSettings, QT_VERSION
+    Qt, QEvent, QSize, QUrl, QTimer, QFile, QByteArray, QSettings, QT_VERSION,
+    QObject
 )
 
 try:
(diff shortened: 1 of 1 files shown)


astaric:
@ales-erjavec, is this a proper way to do it?

I considered handling drops in CanvasMainWindows, but then the whole window would be the drop target.

ales-erjavec:
>I considered handling drops in CanvasMainWindows, but then the whole window would be the drop target.

You can:

1) condition the drop on certain sub areas by looking at event.pos() and accepting/rejecting it accordingly
2) use an event filter on the 'target' widget. This is how it was already [done in 2.7 ](https://github.com/biolab/orange/commit/04304dcb1ce570c16005cbb35b72c48de1a0acfe)

astaric:
Thanks, I like the event filter way. I'll update the code accordingly.

astaric:
I ended up using the same code as in 2.7 :)

I did not use the OSX_NSURL_toLocalFile workaround as it is not needed for Qt5 (we started using it because PyQt4 never got official support for "permanent" file urls on macos). Since all our installers ship with PyQt5, this should be ok.

markotoplak:
Works for me on Linux.

One issue though: If I already have something open and chose "Cancel" in the save changes dialog, parts of the dialog remain until (parts of) the canvas is redrawn.

PR #2890, opened 2018-01-29
[FIX] canvasmain: Deal with the case QWebView is None

payno:
##### Issue
Try to call QWebView but can be set to None

##### Description of changes
manage the case where help_view is None

##### Includes
- [X] Code changes


ales-erjavec:
This was already done in c951646aff08fbcf57eca1bc69b469c95a857503

PR #2896, opened 2018-01-31
[FIX] listfilter: Fix filter line edit completion

ales-erjavec:
##### Issue

Inline filter completion in the Select Columns widget does not work.

##### Description of changes

Make QCompleter a child of the filter line edit.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/listfilter.py
+++ b/Orange/widgets/utils/listfilter.py
@@ -211,7 +211,7 @@ def variables_filter(model, parent=None):
     filter_edit.setPlaceholderText("Filter")
 
     completer_model = QStringListModel()
-    completer = QCompleter(completer_model)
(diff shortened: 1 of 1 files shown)


lanzagar:
This results in the filter line auto-completing to the first hit and the box below shows only that one feature...
E.g. Iris with all attributes on the left; when I write "sepal" I see only one feature instead of two.

If the filter box shows the first hit, I would expect to see all matching variables in the box below, until pressing enter when the auto-complete suggestion is accepted.

lanzagar:
Any suggestions on how to implement completion so as not to interfere with regular filtering tasks?
@ales-erjavec or someone else?

astaric:
The completion now only works on variable attributes, so it has no effect on regular filtering.

PR #2897, opened 2018-01-31
LineEditWFocusOut: Pressing enters selects text

markotoplak:
Pressing Enter in line edit should show the user that the data has
been "entered". Selecting the field seems better than clearing the focus
because (1) it allows the field to be easily re-edited (maybe the user
just wanted to see the efects of her change) and because (2) we do not
lose the focus position (moving or field forward or backward is easy).

This is also what QSpinBox does on Enter.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -798,7 +798,10 @@ class LineEditWFocusOut(QtWidgets.QLineEdit):
         self.__changed = True
 
     def returnPressedHandler(self):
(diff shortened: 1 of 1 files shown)


markotoplak:
I think this still achieves the goal of #1634.

lanzagar:
When I input some text, press enter and then tab to move to the next control, the text in the first line edit stays selected.

PR #2899, opened 2018-02-02
[ENH] Add-ons: Option to list only trusted add-ons

markotoplak:
##### Description of changes
Add-on with has a new checkbox which by default hides unofficial add-ons.

Installed add-ons are always shown.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
@lanzagar, I updated this. Could you check?

PR #2901, opened 2018-02-06
OWKMeans: Perform clustering in worker threads

pavlin-policar:
##### Issue
K-means executed everything in the main GUI thread, leading to freezing when processing large datasets. Moreover, the computation started as soon as any data was connected, so connecting large datasets was, again, problematic.

##### Description of changes
Offload all the computation into separate threads, and start clustering only when Apply is clicked (or checkbox is checked).

Also some general refactoring, to improve consistency with other, newer widgets.

As far as I can tell, and the tests were written, the behavior remained the same as before (other than the time of execution).

@ales-erjavec Could you take a look at this please, since you're the thread master? I wasn't sure how to go about testing threaded code, is this method ok?

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owkmeans.py
+++ b/Orange/widgets/unsupervised/owkmeans.py
@@ -299,7 +299,8 @@ class OWKMeans(widget.OWWidget):
 
         if self.optimize_k and all(isinstance(self.clusterings[i], str)
                                    for i in range(self.k_from, self.k_to + 1)):
-            self.Error.failed('All failed')
+            # Show the error of the last clustering
+            self.Error.failed(self.clusterings[self.k_to])
 
         self.send_data()
 
--- a/Orange/widgets/unsupervised/owkmeans.py
+++ b/Orange/widgets/unsupervised/owkmeans.py
@@ -331,8 +331,6 @@ class OWKMeans(widget.OWWidget):
             task.watcher.exceptionReadyAt.disconnect(self.__on_exception)
             task.watcher.doneAll.disconnect(self.__commit_finished)
 
-            concurrent.futures.wait(task.futures)
-            task.watcher.flush()
             self.progressBarFinished()
             self.setBlocking(False)
 
--- a/Orange/widgets/unsupervised/owkmeans.py
+++ b/Orange/widgets/unsupervised/owkmeans.py
@@ -1,6 +1,5 @@
-import concurrent
-from concurrent.futures import Future
-from typing import Optional, Union, List, Dict
+from concurrent.futures import Future  # pylint: disable=unused-import
+from typing import Optional, List, Dict  # pylint: disable=unused-import
 
 import numpy as np
 from AnyQt.QtCore import Qt, QTimer, QAbstractTableModel, QModelIndex, QThread, \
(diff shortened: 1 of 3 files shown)


Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -172,114 +229,185 @@ def adjustSize(self):
...
-            "not_enough_data",
-            "Too few ({}) unique data instances for {} clusters")
-        data_ok = n <= len(self.data)
-        msg_group.not_enough_data(len(self.data), n, shown=not data_ok)
-        return data_ok
+    def enough_data_instances(self, k):
+        """k cannot be larger than the number of data instances."""
+        return len(self.data) >= k
 
     def _compute_clustering(self, k):
```
ales-erjavec: Please refactor `_compute_clustering` into a pure function with explicit max_iteration, init, n_init, data, ... parameters and change the `__launch_tasks` accordingly.

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -172,114 +229,185 @@ def adjustSize(self):
...
+            self.cluster()
+
         QTimer.singleShot(100, self.adjustSize)
-        self.send_data()
 
-    def invalidate(self, force_apply=False):
+    def invalidate(self, force=False):
         self.Error.clear()
         self.Warning.clear()
         self.clusterings = {}
```
ales-erjavec: `cancel` pending tasks here regardless of `auto_commit` state. If any are still in flight and are not cancelled then `self.clusterings` will be updated in `__clustering_complete` with now invalid results.

Orange/widgets/unsupervised/owkmeans.py
pavlin-policar: I'm guessing that when we call `executor.submit`, the actual job function isn't called straight away, so the problem with the non-pure code was that the settings may change after applying but before the `_compute_clustering` was actually called. Am I interpreting this correctly?

I've changed it in any case.

pavlin-policar:
I'm afraid I wasn't able to reproduce this error. Could you perhaps describe your workflow?

This is very strange, because selecting a different 'Initialization' cancels the tasks, so `update_results` should never be called from the old tasks... Should I disconnect the watcher callbacks *before* cancelling the tasks [here](https://github.com/biolab/orange3/pull/2901/files#diff-0913fd017e312578148451e62fe5d98eR331)? Could this be the issue?

Orange/widgets/unsupervised/owkmeans.py
ales-erjavec: > Am I interpreting this correctly?

Yes.

ales-erjavec:
>... because selecting a different 'Initialization' cancels the tasks, so update_results should never be called from the old tasks... 

Only since https://github.com/biolab/orange3/pull/2901/commits/b4430783c757ef798e6f17a3e7f5438603da4f2f, which fixed the error.

ales-erjavec:
There is one issue which was already present before:

run optimization for 2 to (say) 10 clusters, uncheck 'Apply Automatically', increase "From" to 5. At this point the 'Silhouette Scores' table is not updated. Select '2'  from the scores table (or '3' just as long as it causes a send_data to be called). On the output is a data table with 7 clusters. 

Since you already clear the results table on any initialization parameter change, you might also consider doing the same for `k_from` and `k_to`.

pavlin-policar:
> run optimization for 2 to (say) 10 clusters, uncheck 'Apply Automatically', increase "From" to 5. At this point the 'Silhouette Scores' table is not updated. Select '2' from the scores table (or '3' just as long as it causes a send_data to be called). On the output is a data table with 7 clusters.

I'd be happy to fix this, but I think it should go into a separate PR to avoid one all-encompassing refactor.

pavlin-policar:
Nice catch, thanks!

PR #2908, opened 2018-02-16
[ENH] Select Rows: Add annotated data output

BlazZupan:
##### Issue
Widget outputted only selected or unselected data, whereas in a number of cases one would benefit to have data with the Selected column, especially if selection would be performed on a combination or continuous meta attribute(s) and then used to test if we can predict the grouping.

##### Description of changes
Added an output channel (Data).

##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation


ajdapretnar:
This doesn't seem to work as it should.
No matter what I do, the second row is selected. I tried iris, titanic and housing data sets, and they all display erroneous behaviour.

screen shot 2018-02-16 at 09 20 48

PR #2915, opened 2018-02-19
[FIX] KMeans clear results on k change, do not recluster

pavlin-policar:
##### Issue
- An issue pointed out by @ales-erjavec in #2901 :

>run optimization for 2 to (say) 10 clusters, uncheck 'Apply Automatically', increase "From" to 5. At this point the 'Silhouette Scores' table is not updated. Select '2' from the scores table (or '3' just as long as it causes a send_data to be called). On the output is a data table with 7 clusters.

- KMeans reclusters data when metas or class_vars change, even when X doesn't.

##### Description of changes
Fix both issues. Based on #2901. Remove possibility of crash on settings migrations (not that this ever happened)

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owkmeans.py
+++ b/Orange/widgets/unsupervised/owkmeans.py
@@ -469,8 +469,14 @@ class OWKMeans(widget.OWWidget):
     @Inputs.data
     @check_sql_input
     def set_data(self, data):
-        self.data = data
-        self.invalidate()
+        self.data, old_data = data, self.data
+
+        # Do not needlessly recluster the data if X hasn't changed
+        if old_data and self.data and np.array_equal(self.data.X, old_data.X):
+            if self.auto_commit:
+                self.send_data()
+        else:
+            self.invalidate()
 
     def send_report(self):
         # False positives (Setting is not recognized as int)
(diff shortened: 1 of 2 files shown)


ajdapretnar:
This looks ok to me. Just one question. When I change metas with Select Columns and don't have Apply Automatically selected in k-Means, the data still gets transmitted through the widget (the output changes). I would expect the output not to change, unless I press Apply. Correct me, if I am wrong.

p.s. I could not reproduce the mentioned issue, but everything works for me, so no complaints here.

pavlin-policar:
You're right, this is not expected. I'll fix it right away.

The issue was a bit subtle, but you can notice it if you look at the output in a data table and look at how the clusters change.

Open k-means on e.g. iris, disable auto-apply and run optimization from 2-8 clusters. Show the output in a data table, selecting 2 clusters will give values C1 and C2, as expected. Now change the parameter "from" to 5 without pressing "Apply". The results will still show the previous range 2-8, but now selecting 2 will output clusterings for k=5 (values C1-C5), which is wrong (as seen on image).

PR #2918, opened 2018-02-21
[FIX] Test and Score: Fix averaging over classes for binary scores

PrimozGodec:
##### Issue

Binary scores were evaluated with target class when no target class was selected

##### Description of changes
This PR is fixing binary scores to use weighted scoring when no target class is selected

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/tests/test_owtestlearners.py
+++ b/Orange/widgets/evaluate/tests/test_owtestlearners.py
@@ -7,7 +7,7 @@ import numpy as np
 from AnyQt.QtWidgets import QMenu
 from AnyQt.QtCore import QPoint
 
-from Orange.classification import MajorityLearner
+from Orange.classification import MajorityLearner, LogisticRegressionLearner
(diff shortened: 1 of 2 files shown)


Orange/widgets/evaluate/tests/test_owtestlearners.py
```diff
@@ -247,6 +247,38 @@ class NewRegressionScore(RegressionScore):
...
+        w.class_selection = "Iris-versicolor"
+        w._on_target_class_changed()
+        versicolor_auc = float(w.view.model().item(0, 1).text())
+
+        w.class_selection = "Iris-virginica"
+        w._on_target_class_changed()
+        virginica_auc = float(w.view.model().item(0, 1).text())
+
+        self.assertIsNotNone(average_auc)
+        self.assertIsNotNone(versicolor_auc)
```
lanzagar: Why is checking that the aucs are not None necessary?
If they would not be computed, the conversion to float would fail before this anyway.

PR #2919, opened 2018-02-21
contingency: allow weights

astaric:
##### Issue
Contingency did not allow weights since our old bottlechest implementation of bincount (probably) did not support them. As our current bincount handles weights just fine, there is no need for the error anymore

##### Description of changes
Remove weights check in contingency, pass weights to bincount.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -233,7 +233,8 @@ def contingency(X, y, max_X=None, max_y=None, weights=None, mask=None):
         The maximal value in the array
     max_y : int
         The maximal value in `y`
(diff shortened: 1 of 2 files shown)


astaric:
Where is here?

PrimozGodec:
Sory I just wanted to mark the location in docstring where ... are instead of weights definition. Can you just define weights there?

astaric:
Done

PrimozGodec:
Thank you, it looks good now.

PR #2920, opened 2018-02-21
[FIX] sgd: Change deprecated n_iter to max_iter

### LINKED ISSUE #2917: SGDClassificationLearner deprecated parameter (n_iter)
DeprecationWarning: n_iter parameter is deprecated in 0.19 and will be removed in 0.21. Use max_iter and tol instead.

smiks:
##### Issue
SGDClassificationLearner deprecated parameter (n_iter). Fixes #2917.

##### Description of changes
Deprecation fix: changed n_iter to max_iter (SGDClassifier)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owsgd.py
+++ b/Orange/widgets/model/owsgd.py
@@ -306,6 +306,11 @@ class OWSGD(OWBaseLearner):
                 coeffs.name = "coefficients"
         self.Outputs.coefficients.send(coeffs)
 
+    @classmethod
+    def migrate_settings(cls, settings_, version):
+        if version < 2:
+            settings_["max_iter"] = settings_.pop("n_iter", 5)
+
 
 if __name__ == '__main__':
     import sys
--- a/Orange/widgets/model/owsgd.py
+++ b/Orange/widgets/model/owsgd.py
@@ -76,7 +76,7 @@ class OWSGD(OWBaseLearner):
     learning_rate_index = Setting(0)
     eta0 = Setting(.01)
     power_t = Setting(.25)
-    max_iter = Setting(10)
+    max_iter = Setting(5)
 
     def add_main_layout(self):
         self._add_algorithm_to_layout()
--- a/Orange/widgets/model/owsgd.py
+++ b/Orange/widgets/model/owsgd.py
@@ -24,6 +24,8 @@ class OWSGD(OWBaseLearner):
     ]
     priority = 90
 
+    settings_version = 2
+
     LEARNER = SGDLearner
 
     class Outputs(OWBaseLearner.Outputs):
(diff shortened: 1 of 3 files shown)


Orange/widgets/model/owsgd.py
```diff
@@ -76,7 +76,7 @@ class Outputs(OWBaseLearner.Outputs):
     learning_rate_index = Setting(0)
     eta0 = Setting(.01)
     power_t = Setting(.25)
-    n_iter = Setting(5)
+    max_iter = Setting(10)
```
ajdapretnar: Default in sklearn is 5. Is there a particular reason to have it at 10?

pavlin-policar: Please implement a `migrate_settings` method. Without it, old workflows won't use the old `n_iter` value as the `max_iter` value while their semantics are essentially the same.

smiks: No particular reason. Sklearn will change it to 1000 from 0.21. I decided for 10 only to see if I changed everything correctly (testing widget in Orange).

smiks: I don't know how to do it.
Is this correct (added in class OWSGD)?

```
    @classmethod
    def migrate_settings(cls, settings_, version):
        if version < 2:
            settings_["max_iter"] = settings_.get("n_iter", 5)
```

pavlin-policar: That should do the trick. Perhaps delete the setting `n_iter` as well to clean up? You can do this in one line `settings_["max_iter"] = settings_.pop("n_iter", 5)`.

smiks:
I'll try to do both (```tol``` parameter and fix for SGDRegressionLearner).
I only have to fix learner (widget is ok - in case I don't add ```tol``` parameter), right?

lanzagar:
Yes, the SGD widget calls the regression or classification learner depending on the type of data it gets.
It now already uses max_iter, so just the regression SGD class needs to be updated to match it.

lanzagar:
It was hard to fall back to the old functionality (of setting just the number of iterations) without being able to set tol to None - I guess that is why you also removed migrate settings? I added a checkbox for `tol` and brought back migration.
You can double check my commit to make sure everything is still OK, then I think this is ready for merging.

smiks:
I'll check it. It is my first PR for Orange so I still have a lot to learn about Orange.

smiks:
Looks ok. I have also tested it locally with different datasets (iris and my datasets).  Tested different settings to see if both parameters work. Looks like it works, otherwise I wouldn't get this message ```Maximum number of iteration reached before convergence. Consider increasing max_iter to improve the fit.```

smiks:
Thanks everyone. I'm happy to have contributed to Orange.

PR #2924, opened 2018-02-26
[FIX] contingency.pyx: Fix out of bound write

ales-erjavec:
##### Issue

Computing contingencies for more then 127 classes results in an out or bounds write in `contingency_floatarray` possibly leading to a segmentation fault (in the best case) or just wrong results.

```python
import Orange.statistics.contingency, numpy
X = Orange.data.ContinuousVariable("X")
C = Orange.data.DiscreteVariable("C", values=["C{}".format(i + 1) for i in range(1024)])
domain = Orange.data.Domain([X], [C])
d = Orange.data.Table.from_numpy(
    domain,
    numpy.random.uniform(size=(4000, 1)).round(3),
    numpy.random.randint(0, 1024, size=(4000, 1)),
)
c = Orange.statistics.contingency.get_contingency(d, X, C)
```

consistently results in a segfault on macOS Python 3.5 due to memory corruption.

##### Description of changes

* Enable cython bounds check on `contingency_floatarray`.
* Change dtype of classes to intp_t to allow more then 127 'class' values.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


lanzagar:
Nice catch!

PR #2930, opened 2018-03-02
[FIX] Save Image to SVG fixed on Qt5

markotoplak:
##### Issue
SVG export does not work on Qt5.

##### Description of changes
Changed two lines in the pyqtgraph. Including the file here as pyqtgraph rarely releases. This PR included a modified pyqtgraph/exporters/SVGExporter.py from PyQtGraph that included proposed fixes
pyqtgraph/pyqtgraph#641 and pyqtgraph/pyqtgraph#642

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/io.py
+++ b/Orange/widgets/io.py
@@ -158,7 +158,7 @@ class SvgFormat(ImgFormat):
 
     @staticmethod
     def _get_exporter():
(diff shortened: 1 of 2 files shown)


lanzagar:
While exporting to .svg is fixed, .pdf still has some minor (cropping) issues that can hopefully be fixed in another PR.

PR #2936, opened 2018-03-03
Allow oversampling in Data Sampler widget

### LINKED ISSUE #2850: Allow oversampling in Data Sampler widget
When using `Fixed Sample Size` in Data Sampler Widget, `Instances` spinbox is limited to the number of rows in the dataset. It makes sense when we sample without replacement, but when sampling with replacement there is no need for such upper bound.

When `Sample with replacement` checkbox is selected, `Instances` spin should allow entry of arbitrary high numbers.

cemsbr:
##### Issue
Fixes #2850

##### Description of changes
Changes:
1. Set instances' limit when there is data and replacement is chosen;
2. Remove instances' limit otherwise;
3. Warn if sample with replacement is bigger than input.

Refactoring:
1. Extract 2 ** 31 - 1 constant;
2. Clear warnings when errors are cleared.

##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owdatasampler.py
+++ b/Orange/widgets/data/tests/test_owdatasampler.py
@@ -81,6 +81,38 @@ class TestOWDataSampler(WidgetTest):
                     self.assertEqual(len(self.zoo), len(sample) + len(other))
                     self.assertNoIntersection(sample, other)
 
+    def test_bigger_size_with_replacement(self):
+        """Allow bigger output without replacement."""
+        _, out_size, actual_size = self._set_bigger_sample_size(True)
+        self.assertEqual(out_size, actual_size,
+                         'Sample size should not be lowered with replacement')
+
+    def test_bigger_size_without_replacement(self):
+        """Lower output samples to match input's without replacement."""
+        in_size, _, actual_size = self._set_bigger_sample_size(False)
+        self.assertEqual(in_size, actual_size)
+
+    def _set_bigger_sample_size(self, with_replacement):
+        """Load data, set sample size and click to generate samples.
+
+        The sample size is set to out_size which is bigger than the input size
+        in_size.
+
+        Returns integers in_size, out_size and the actual gui sample size.
+        """
+        in_size = 2
+        out_size = in_size + 1
+        data = self.iris[:in_size]
+        self.widget.set_data(data)
+
+        self.select_sampling_type(self.widget.FixedSize)
+        self.widget.controls.replacement.setChecked(with_replacement)
+        self.widget.sampleSizeSpin.setValue(out_size)
+        self.widget.commit()
+
+        actual_size = self.widget.sampleSizeSpin.value()
+        return in_size, out_size, actual_size
+
     def assertNoIntersection(self, sample, other):
         for inst in sample:
             self.assertNotIn(inst, other)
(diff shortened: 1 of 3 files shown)


cemsbr:
I did a grep in the docs and didn't find any text about the maximum sample size. So, I marked the "Documentation" item as done.

Orange/widgets/data/tests/test_owdatasampler.py
```diff
@@ -81,6 +81,42 @@ def test_no_intersection_in_outputs(self):
...
+    def test_bigger_size_without_replacement(self):
+        """Lower output samples to match input's without replacement."""
+        in_size, _, actual_size = self._set_bigger_sample_size(False)
+        self.assertEqual(in_size, actual_size)
+
+    def test_bigger_output_warning(self):
+        self._set_bigger_sample_size(with_replacement=True)
+        self.assertTrue(self.widget.Warning.bigger_sample.is_shown())
+
+    def _set_bigger_sample_size(self, with_replacement):
```
astaric: I would use a bit different helper function. Something along the lines of:
```python
def set_fixed_sample_size(self, sample_size, with_replacement=False):
    self.select_sampling_type(self.widget.FixedSize)
    self.widget.sampleSizeSpin.setValue(sample_size)
    self.widget.controls.replacement.setChecked(with_replacement)
    self.widget.commit()
```

While this introduces a bit more code duplication in the actual tests, it allows us to reuse the helper in other (future) tests. The same way you were able to use the `select_sampling_type` we wrote for one of our previous tests.

Orange/widgets/data/tests/test_owdatasampler.py
```diff
@@ -81,6 +81,42 @@ def test_no_intersection_in_outputs(self):
...
+
+        The sample size is set to out_size which is bigger than the input size
+        in_size.
+
+        Returns integers in_size, out_size and the actual gui sample size.
+        """
+        in_size = 2
+        out_size = in_size + 1
+        data = self.iris[:in_size]
+        self.widget.set_data(data)
```
astaric: We try to avoid calling input handlers directly and use self.send_signal instead.
```
```

cemsbr:
@astaric, I followed your suggestions. However, I have no idea why the test below is failing.
```
FAIL: test_optimization_fails (Orange.widgets.unsupervised.tests.test_owkmeans.TestOWKMeans)
  File "/opt/python/3.5.5/lib/python3.5/unittest/mock.py", line 1159, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/biolab/orange3/build/travis-test/Orange/widgets/unsupervised/tests/test_owkmeans.py", line 223, in test_optimization_fails
    self.send_signal(self.widget.Inputs.data, self.iris, wait=1000)
  File "/home/travis/build/biolab/orange3/build/travis-test/Orange/widgets/tests/base.py", line 245, in send_signal
    self.assertTrue(spy.wait(timeout=wait))
```

astaric:
Thanks.

K-Means clustering is one of the widgets that does the processing in a separate thread, so the test waits a bit and hopes that processing is done. Since the time needed for execution depends on the system load, it can sometimes happen that the processing is not done yet, triggering that error. We usually work around this by re-running tests, which then usually pass :)

PR #2950, opened 2018-03-08
OWDataSets: refactor code, move some functionalities to new methods

JakaKokosar:
##### Issue
https://github.com/biolab/orange3-single-cell/issues/121

##### Description of changes
Split large chunk of code into smaller methods, so we can easily override them.
Moved item model creation in a separate method.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -1,4 +1,3 @@
-import enum
 import logging
 import numbers
 import os
@@ -6,10 +5,11 @@ import sys
 import traceback
 
 from xml.sax.saxutils import escape
-from concurrent.futures import ThreadPoolExecutor, Future
+from concurrent.futures import ThreadPoolExecutor, Future  # pylint: disable=unused-import
 
-from types import SimpleNamespace as namespace
-from typing import Optional, Dict, Tuple
+from types import SimpleNamespace
+from typing import Optional, Dict, Tuple, List  # pylint: disable=unused-import
+from collections import namedtuple
 
 from AnyQt.QtWidgets import (
     QLabel, QLineEdit, QTextBrowser, QSplitter, QTreeView,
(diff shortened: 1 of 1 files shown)


JakaKokosar:
@lanzagar

Orange/widgets/data/owdatasets.py
```diff
@@ -125,6 +123,12 @@ def __init__(self):
         super().__init__()
         self.local_cache_path = os.path.join(data_dir(), self.DATASET_DIR)
 
+        self._header_labels = [header['label'] for _, header in self.HEADER_SCHEMA]
+        self._html_description_tags = ['year', 'source', 'title', 'description', 'seealso', 'references']
+
+        self._header_index = namedtuple('_header_index', [*[info_tag for info_tag, _ in self.HEADER_SCHEMA]])
```
lanzagar: This is a syntax error on older python versions (as reported by failing tests).
But also, I don't see a reason for unpacking and repacking the list...

JakaKokosar: Thanks, forgot to remove this when I was testing something.

Orange/widgets/data/owdatasets.py
```diff
@@ -208,71 +200,64 @@ def __init__(self):
...
+                self.allinfo_remote[file_path].get('version', '') != self.allinfo_local[file_path].get('version', '')
+        )
+        islocal &= not outdated
+
+        prefix = os.path.join('', *file_path[:-1])
+        filename = file_path[-1]
+
+        # add missing variables to the namespace
+        # not needed if all info files have the same JSON structure
+        for tag in self._html_description_tags:
```
lanzagar: Not all variables are in `_html_description_tags`, e.g. one test fails, because 'size' is missing (which is in `_header_index`, but not in `_html_description_tags`.
Also, did you make sure that it is OK to set all missing things to `None`? Previously, some default values were different, e.g. `[]` for 'references'.

PR #2958, opened 2018-03-15
[ENH] Neural network widget that works in a separate thread

markotoplak:
##### Issue
In general, long computation shouldn't leave the GUI unresponsive. But the real issue is that I am not sure I can write threaded widgets, so I tried.

##### Description of changes
Implemented with the help of @ales-erjavec's example from the documentation with minor changes. Callbacks with scikit-learn are very hackish though. Does anybody have a better idea?

Just to test stopping, I also added a cancel button to the widget. Perhaps all interruptible widget should get a stop icon in the status bar and perhaps also some actionable "icon" on the canvas.  

##### Includes
- [X] Code changes
- [X] Fixed tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owneuralnetwork.py
+++ b/Orange/widgets/model/owneuralnetwork.py
@@ -1,8 +1,8 @@
 from functools import partial
+import copy
 import logging
 import re
 import sys
-from unittest.mock import patch
 import concurrent.futures
 
 from AnyQt.QtWidgets import QApplication, qApp
@@ -18,7 +18,6 @@ from Orange.widgets.utils.owlearnerwidget import OWBaseLearner
 from Orange.widgets.utils.concurrent import (
     ThreadExecutor, FutureWatcher, methodinvoke
 )
-
 
 
 class Task:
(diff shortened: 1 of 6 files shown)


markotoplak:
@ales-erjavec Could you check this PR and comment, please?

Please, also see the changes to the widget's cancel(). There I needed to process events so that the progress bars updates worked OK when I used the cancel button.

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -81,6 +124,118 @@ def get_hidden_layers(self):
...
+                find = re.findall(r"Iteration (\d+)", args[0])
+                if find:
+                    iters = int(find[0])
+            callback(iters)
+
+        def build_model(data, learner):
+            if learner.kwargs["solver"] != "lbfgs":
+                # enable verbose printouts within scikit and redirect them
+                with patch.dict(learner.kwargs, {"verbose": True}),\
+                     patch("builtins.print", print_callback):
```
ales-erjavec:
```
Time -->

Thread-1
-|patch|-----|exit|--------

Thread-2
----|patch|--------|exit|--
          ^             ^
         stores         restores
     patched print      the patched print of Thread-1

```

markotoplak: Whoops. Thank you. Did not know that patch mutates state globally....

kernc:
The alternative to your last commit would be to set `max_iter=2` and call [`clf.partial_fit()`](http://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPClassifier.html#sklearn.neural_network.MLPClassifier.partial_fit) `real-max_iter/2`-times, with callbacks in-between.

markotoplak:
@kernc Yes, that would certainly be an option. But then I find automatic stopping due to too tiny updates problematic: we would have to detect it and avoid running partial_fit again.

ales-erjavec:
What about learning rate schedule? From what I can tell the 'adam' solver uses learning rate decay, so any model trained in such manner could be very different then one trained in a single run.

markotoplak:
@ales-erjavec, what do you think about this attempt?

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -81,6 +123,106 @@ def get_hidden_layers(self):
...
+        """
+        if self._task is not None:
+            self._task.cancel()
+            assert self._task.future.done()
+            # disconnect the `_task_finished` slot
+            self._task.watcher.done.disconnect(self._task_finished)
+            self._task = None
+        # threads use signals to run functions in the main thread and some
+        # can still be quoued (perhaps change)
+        qApp.processEvents()
```
ales-erjavec: Remove this. If needed route the progress updates via an intermediary QObject [maybe like this](https://github.com/ales-erjavec/orange3/commit/03c6791c81302989afbfe9c046ba5dac6e41de24)

markotoplak: Thanks! I haven't thought of this.

markotoplak:
I do not like faking n_iter_ either, but haven't thought of anything better here... Feel free to decide that this is too dirty to get into Orange.

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -142,35 +156,42 @@ def __update(self):
...
         # copy to set the callback so that the learner output is not modified
         # (currently we can not pass callbacks to learners __call__)
         learner = copy.copy(self.learner)
         learner.callback = callback
 
         def build_model(data, learner):
-            return learner(data)
+            try:
+                return learner(data)
+            except CancelTaskException:
```
lanzagar: What is the difference between former (no try-except) and this?

ales-erjavec: The later does not capture the exception instance's `__traceback__` in the resulting Future (which we know will be discarded).

Issue #2976, opened 2018-03-27
Can't retrieve package list from Add-ons in Menu


LuckyDao:
##### Orange version
Version 3.11.0

##### Expected behavior
Expected to populate add ons

##### Actual behavior
Throws max() error

##### Steps to reproduce the behavior
enter Orange3, click Options>Add-ons

##### Additional info (worksheets, data, screenshots, ...)

![image](https://user-images.githubusercontent.com/35459931/37984761-c8806b78-31bc-11e8-9397-2d98e56fe859.png)

Varidrix:
Encountering the same problem. Reinstall of doesn't seem to work.

janezd:
MeToo

kernc:
Fixed. For now.

LuckyDao:
Thank you!

PR #2977, opened 2018-03-27
[FIX] Skip add-ons for which PyPI returned no valid versions.

kernc:
##### Issue
Fixes https://github.com/biolab/orange3/issues/2976

##### Description of changes

1. Skip add-ons for which PyPI returned no valid versions.
2. Relabel "trusted add-ons" to "official add-ons".

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/addons.py
+++ b/Orange/canvas/application/addons.py
@@ -706,6 +706,10 @@ def list_pypi_addons():
     releases = multicall()
     multicall = xmlrpc.client.MultiCall(pypi)
     for addon, versions in zip(addons, releases):
+        # Skip unavailable
+        if not versions:
+            continue
+
         # Workaround for PyPI bug of search not returning the latest versions
         # https://bitbucket.org/pypa/pypi/issues/326/my-package-doesnt-appear-in-the-search
         version_ = max(versions, key=version.LooseVersion)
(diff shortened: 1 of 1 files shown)


astaric:
The offending add-on is [Orange3-Bioinformatics](https://pypi.org/project/Orange3-Bioinformatics), but I cannot see anything wrong with the package. Could this be related to pypi moving to the new infrastructure?

lanzagar:
If orange3-bioinformatics is the only one with problems, maybe @JakaKokosar should also take a look at that.

astaric:
I mistyped, Orange-Bioinformatics (the old one)

lanzagar:
In that case #2975 will also help.
The problem is that that PR and this will only make a difference in the next release of Orange. If someone finds out what went wrong and how to fix the old Bioinformatics, that would be even better!

As for the contents of this PR - I approve the actual fix and leave it to @markotoplak to decide if he likes the unrelated renaming of "trusted" -> "official" (the choice was actually a hot topic a while back :))

JakaKokosar:
I've hidden the latest (2.6.25) Orange-Bioinformatics package from PyPI because we don't want to have both versions listed. I didn't take into consideration that Orange-Bioinformatics is still listed as a trusted add-on. 

Should I make this package visible again, just until the next release?

markotoplak:
@JakaKokosar , how did you hide it? I saw astaric's commit removed orange3 add-on from the KEYWORDS. I made a new release and added it back.

markotoplak:
Ok, now, that the latest version is not hidden on pypi anymore, Orange without fixes works again.

kernc:
:+1:

PR #2988, opened 2018-04-06
addons: catch OSError in have_install_permission

astaric:
PermissionError is just one of the possible errors that can be raised from a failed open attempt. When one of the other errors occurs, an error was
shown (as can be seen on sentry)


astaric:
See https://sentry.io/biolab/orange3/issues/469564153

PR #2999, opened 2018-04-15
[ENH] DomainModel: Add match_exact_types option

pavlin-policar:
##### Issue
Domain model always includes subtypes of a variable type. 

In feature statistics, we want to be able to color the histograms with continuous variables, but coloring them with datetime makes little sense (or it may make sense if it was more carefully thought out and implemented properly).

##### Description of changes
Add `match_exact_types` parameter that enables you to show only the exact type you specified.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -885,7 +885,8 @@ class DomainModel(VariableListModel):
     PRIMITIVE = (DiscreteVariable, ContinuousVariable)
 
     def __init__(self, order=SEPARATED, separators=True, placeholder=None,
-                 valid_types=None, alphabetical=False, skip_hidden_vars=True, **kwargs):
+                 valid_types=None, alphabetical=False, skip_hidden_vars=True,
+                 match_exact_types=False, **kwargs):
         """
 
         Parameters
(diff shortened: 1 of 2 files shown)


kernc:
Why would coloring by datetime variable not make sense? :open_mouth: Datetime values are well-ordered and have endpoints; can indicate what came early and what comes late.

How does Feature Statistics differ from any other widget where coloring is possible by any continuous variable, including datetimes?

pavlin-policar:
Well, yes, I guess it does make sense to do this. My reasoning was that it may not be immediately obvious to the user. Also, since we've opted to hide time variables in feature statistics for now, it would be quite confusing to see a variable in the drop down menu but not on the display, which supposedly shows the available variables (even though it really doesn't :smile:).

Giving this further thought, I'll probably put the time variables back in, since treating them as regular continuous variables seems perfectly reasonable.

> How does Feature Statistics differ from any other widget where coloring is possible by any continuous variable, including datetimes?

It doesn't. However, I would argue that the other visualization widgets in Orange were not designed with time variables in mind, since most of them don't display them properly e.g. scatterplot on `cyber-security-breaches`. The unix timestamp appears in almost every visualization, with the exception of Box plot, which appears to display them correctly.
image

In any case, I'll probably re-add time variables, and this change will become unnecessary.

kernc:
Where label strings are off, it's because the labels aren't rendered with the use of `Variable.repr_val()`.

Nothing in Orange was particularly designed, I'm often reassured. :smile:

pavlin-policar:
I've added support for time variables. This is no longer necessary.

PR #3011, opened 2018-04-19
[FIX] boxplot labels overlap

astaric:
##### Issue
When labels are wider than their respective bars in a boxplot, labels overlap:

screen shot 2018-04-19 at 15 39 12

##### Description of changes
Limit  (all but the last) label width to the width of the bar they are labelling. Hide labels for very short bars (unless the whole label can be displayed in available space).

screen shot 2018-04-19 at 20 49 11

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -610,9 +610,6 @@ class OWBoxPlot(widget.OWWidget):
                          y - label.boundingRect().height() - 8)
             label.setMaxWidth(bar_part.boundingRect().width())
             self.box_scene.addItem(label)
-        if label is not None and bar_part.boundingRect().width() >= label.MIN_LABEL_WIDTH:
-            # last label in row can extend beyond its bar
-            label.setMaxWidth(None)
 
     def __draw_bars(self, y, bars):
         """Draw bars
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -10,7 +10,7 @@ from AnyQt.QtWidgets import (
     QGraphicsPathItem, QGraphicsRectItem, QSizePolicy
 )
 from AnyQt.QtGui import QPen, QColor, QBrush, QPainterPath, QPainter, QFont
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owboxplot.py
```diff
@@ -972,6 +1027,50 @@ def send_report(self):
...
+            """Overrides QGraphicsSimpleTextItem.paint
+
+            If text is wider than maxWidth, a white rect is drawn to obscure
+            letters that extend beyond allowed width.
+            """
+            super().paint(painter, option, widget)
+
+            if self.__max_width is None or self.boundingRect().width() < self.__max_width:
+                return
+
```
markotoplak: @ales-erjavec Do you perhaps have any ideas how to limit label painting to some rect? So that we would not need to draw white rects over label parts?

ales-erjavec: * set clip rect on the painter
* actually implement bounded text painting (QPainter has quite a few drawText overloads for bounded text painting, also see QFontMetrics.elidedText)

lanzagar:
Otherwise good, I just noticed that the last label can extend beyond the graph (to the right):
screenshot_18-05-11_11 46 41

Which even cuts off the text when the window size is smaller...
screenshot_18-05-11_11 46 08

Orange/widgets/visualize/owboxplot.py
```diff
@@ -972,6 +1027,44 @@ def send_report(self):
         if text:
             self.report_caption(text)
 
+    class Label(QGraphicsSimpleTextItem):
+        """Boxplot Label with settable maxWidth"""
```
lanzagar: Should we move something like this to the gui module, so it can be reused elsewhere?

astaric: When anyone else needs a label like this, it can be moved and imported from there. Until then, I see no point in further polluting the gui namespace.

astaric:
The label extending to the right was a "feature", but since cropping of the visualization does not account for the label, I have removed it.

ajdapretnar:
Yolo.

Issue #3007, opened 2018-04-19
Addons: Sort packages in a predictable way


astaric:
<img width="470" alt="screen shot 2018-04-19 at 06 58 08" src="https://user-images.githubusercontent.com/588601/38972127-1488eb0a-439f-11e8-8040-e6e0c45ffd10.png">

It does not look like they are sorted by stability, nor alphabetically. If the current order is not deliberate, I would go with alphabetical sort.

markotoplak:
Fixed in #3013

PR #3013, opened 2018-04-20
[FIX] Sorting add-ons in the alphabetical order

PrimozGodec:
##### Issue

Add-ons are in unpredictable order https://github.com/biolab/orange3/issues/3007

##### Description of changes
Sorting add-ons in the alphabetical order

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/addons.py
+++ b/Orange/canvas/application/addons.py
@@ -311,6 +311,7 @@ class AddonManagerWidget(QWidget):
                 self.__view.model().index(0, 0),
                 QItemSelectionModel.Select | QItemSelectionModel.Rows
             )
+        self.__proxy.sort(1)  # sorting list of add-ons in alphabetical order
 
     def item_state(self):
         steps = []


Orange/canvas/application/addons.py
```diff
@@ -300,6 +300,7 @@ def set_items(self, items):
 
             model.appendRow([item1, item2, item3, item4])
 
+        model.sort(1)  # sorting list of add-ons in the alphabetical order
```
markotoplak: It does not work correctly. Have you tested if then the correct add-ons are also installed? If I added "pip" to the list and tried to upgrade it, it upgraded the wrong package.

ales-erjavec: @PrimozGodec Try `self.__proxy.sort(1)`

PrimozGodec: Thank you @markotoplak and @ales-erjavec. I think that now it works.

Issue #2972, opened 2018-03-26
Better control for selecting variables


astaric:
Orange currently uses combo boxes in multiple widgets as a mean to select a single variable. When working with genetic/bow/... data with tens of thousands of feature, the combo box is slow to open and hard to use (it extends over the whole screen). geo-gds360 documentation set can be used with scatter plot to reproduce the problem.

We could make selecting the variable easier by introducing a custom component. When closed, it can look like an ordinary combo box displaying the selected value. When opened, it could display a widget similar to this one:
<img width="204" alt="screen shot 2018-03-26 at 14 20 47" src="https://user-images.githubusercontent.com/588601/37906023-1a2d6172-3101-11e8-886a-109eada0b52f.png">
The list includes all variables currently included in the combo box, sorted in the same way (defined by DomainModel). The height of the list is limited. If there are too many variables, a scrollbar is displayed. Filter field is used to filter the displayed variables.

When control is opened, the currently selected item is selected in the list. up and down arrows should change the selected item to one above/below the current one. If a user starts typing, the filter field should get the content and the list should immediately update. When filtering is changed (but not empty), the first matching row should get selected. When the filter is removed (set to empty), the last selected item should probably remain selected.

When an item in the list is clicked or return is pressed, the control is closed and the selected item from control is passed to the widget, callbacks are called,... If the control is closed with Esc, no changes should be made to the widget.

lanzagar:
A very necessary change in my opinion!
Maybe as a bonus feature we could have tab-completion. E.g. if features have names "AVeryLongFeatureName1", "AVeryLongFeatureName2", ... I can press tab after "AVe" and it completes the remainder that is common to all (currently matching) names.

ajdapretnar:
Yes, please!

PR #3014, opened 2018-04-20
[RFC][ENH] Combo Box Search Filter

ales-erjavec:
##### Issue

Implements gh-2972

##### Description of changes

* Implement a QComboBox, such that the displayed popup list view implements text search.
* Use it in 'Hierarchical Clustering' widget as an example.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


BlazZupan:
Great prototype, works fast! Thanks. 

On mouse over the fields, the selection (on mac, the coloring of the line indicating current position in the list) does not change. Following is an example of how this works for selection of linkage:

a

and how the line for Zzz3 is not marked with blue for selection of annotation:

b

Filter looks like an text entry box. When I first open the menu, I wanted to click somewhere inside the box to activate it and have the caret show, but action closes the pull down menu. I understand that the problem is with deciding which Qt widget is in focus. @markotoplak, @astaric, can you look at this and comment.

ajdapretnar:
This works really well for me. Something we could merge soon?

lanzagar:
It is indeed great.
I suggest we merge this right after the current release and start adding it to widgets so that the next release has a batch of new combo boxes.
Both of @thocevar's comments make sense, I would also like to see home/end move the cursor where I am typing. But tweaks can also be made later, this has been open for too long already, so I don't mind merging even as it is now.

ales-erjavec:
Can you please test Home/End keys again to ensure they correspond to standard platform behavior.

thocevar:
Home/End keys work on windows.

PR #3018, opened 2018-04-25
[ENH] Icon redesign: k-means, clustering, distances

larazupan:


markotoplak:
Thanks. It seems like you submitted the wrong icon for the Distance map widget. Could you fix it, please?

I only a bit unsure about the new Save Distance icon... The proposed diskette is different that the Save data one but I am not sure which to keep (the old one could also get the same label).

unsupervised_before

unsupervised_after

PR #3022, opened 2018-05-04
[ENH] Data Info widget displays data attributes

BlazZupan:
##### Issue
Some data in orange contain meta information about the data table. For instance, bioinformatics add-on passes the data on the organism or location of gene names by adding this information in .attributes field of the Orange data table. While this information is used in practically all bioinformatics widgets, there is no widget that would explicitly display this information.

##### Description of changes
The box Data Attributes was added to the widget.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owdatainfo.py
+++ b/Orange/widgets/data/owdatainfo.py
@@ -37,7 +37,7 @@ class OWDataInfo(widget.OWWidget):
         self.data_set_size = self.features = self.meta_attributes = ""
         self.location = ""
         for box in ("Data Set Size", "Features", "Targets", "Meta Attributes",
-                    "Location"):
+                    "Location", "Data Attributes"):
             name = box.lower().replace(" ", "_")
             bo = gui.vBox(self.controlArea, box,
                           addSpace=False and box != "Meta Attributes")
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/tests/test_owdatainfo.py
```diff
@@ -0,0 +1,26 @@
...
+from Orange.widgets.data.owdatainfo import OWDataInfo
+from Orange.widgets.tests.base import WidgetTest
+
+
+class TestOWDataInfo(WidgetTest):
+    def setUp(self):
+        self.widget = self.create_widget(OWDataInfo)
+
+    def test_data(self):
+        """No crash on empty data"""
```
lanzagar: Wrong docstring (the data in this test is not empty).

Orange/widgets/data/owdatainfo.py
```diff
@@ -179,10 +177,14 @@ def join_if(items):
             ("{} textual", str_metas)
         ))
 
+        if data.attributes:
+            self.data_attributes = pack_table(data.attributes.items())
```
lanzagar:
```py
else:
    self.data_attributes = ""
```
to clear this box for data without attributes.

BlazZupan: Long data attribute labels are now stripped. I decided against wrapping, as this would solve only the problem for shorter labels and or longer labels we would need to implement something entirely different.

PR #3032, opened 2018-05-14
[FIX] Select Columns: Drag/drop

ales-erjavec:
##### Issue

Drag/Drop between *Select Columns* list views is slow (as noticed in gh-3030)

##### Description of changes

Reintroduce efficient drag/drop in *Select Columns*

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


astaric:
Drag and drop works fine now on gds data, while it stalled for more than a minute before.

PR #3038, opened 2018-05-21
[ENH] Display Widgets on Top option

astaric:
##### Issue
When Canvas gets focus, all widgets get hidden behind it

##### Description of changes
View -> Float Widgets on Top flag to show widgets on top of all other windows.

### DIFF
--- a/Orange/canvas/scheme/widgetsscheme.py
+++ b/Orange/canvas/scheme/widgetsscheme.py
@@ -28,7 +28,7 @@ import sip
 from AnyQt.QtWidgets import QWidget, QShortcut, QLabel, QSizePolicy, QAction
 from AnyQt.QtGui import QKeySequence, QWhatsThisClickedEvent
 
-from AnyQt.QtCore import Qt, QObject, QCoreApplication, QTimer, QEvent
+from AnyQt.QtCore import Qt, QObject, QCoreApplication, QTimer, QEvent, QSettings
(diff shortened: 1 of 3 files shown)


ales-erjavec:
On macOS all windows with the `Qt.WindowStaysOnTop` hint stack above other windows including windows belonging to other applications.
Perhaps we can use `QGuiApplication.applicationStateChanged` to unset/set the flag around the state changes.

Orange/canvas/scheme/widgetsscheme.py
```diff
@@ -627,6 +635,15 @@ def widget_processing_state(self, widget):
         """
         return self.__widget_processing_state[widget]
 
+    def show_widgets_on_top_changed(self):
+        """
+        `Float Widgets on Top` menu option has changed.
+
+        Update the flag on existing widgets.
+        """
```
ales-erjavec: Change this to a setter (`def set_show_widgets_on_top(self, state):`) record the  state in an instance attribute, and use that instead of querying QSettings in  `__set_float_on_top`, so the state is local to the workflow and not shared between them.

astaric:
@ales-erjavec , state is now stored as an instance attribute in WidgetManager.

lanzagar:
Does not really achieve the goal completely on my system (using dwm :)), but I could get the desired functionality even before this PR (and can still).
It does bring all widgets to the top when selected so it might be useful for that (if we add a shortcut).

PR #3047, opened 2018-06-01
[ENH] Save data with compression

lanzagar:
1. Saving with compression was prevented due to a file format extension check that used only the last extension (in e.g. file.tab.gz)
2. Compression was not supported when saving and loading in pickle format
3. Changed default extension for PickleReader from .pickle to the shorter .pkl (both were supported before and still are).


lanzagar:
I improved the OSX workaround (compression extensions were repeated),
and used it on windows as well (although .gz worked after another extension, .xz and .bz2 did not).

Should work on all platforms now.

PR #3053, opened 2018-06-05
[FIX] OWScatterOWScatterPlotGraph: Match group colors with marker colors

### LINKED ISSUE #3051: Scatter Plot, MDS: selection of groups in wrong colors
Group selections in scatter diagrams are marked in a wrong color. For instance, take Iris and group-select two subsets (press shift when selecting the second group):

screen shot 2018-06-04 at 13 37 46

Take feed the selected data into another scatterplot, and select Group for color:

screen shot 2018-06-04 at 13 38 44

Notice that in selection red and green colors were used to mark the points of different groups, while the rest of Orange uses different coloring palette for first two discrete classes.

VesnaT:
##### Issue

Fixes #3051

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/annotated_data.py
+++ b/Orange/widgets/utils/annotated_data.py
@@ -117,9 +117,16 @@ def create_groups_table(data, selection,
                         var_name=ANNOTATED_DATA_FEATURE_NAME):
     if data is None:
         return None
-    values = ["G{}".format(i + 1) for i in range(np.max(selection))]
+    max_sel = np.max(selection)
+    values = ["G{}".format(i + 1) for i in range(max_sel)]
     if include_unselected:
-        values.insert(0, "Unselected")
+        # Place Unselected instances in the "last group", so that the group
+        # colors and scatter diagram marker colors will match
+        values.append("Unselected")
+        mask = (selection != 0)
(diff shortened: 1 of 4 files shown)


Orange/widgets/utils/annotated_data.py
```diff
@@ -117,9 +117,16 @@ def create_groups_table(data, selection,
...
+    values = ["G{}".format(i + 1) for i in range(max_sel)]
     if include_unselected:
-        values.insert(0, "Unselected")
+        # Place Unselected instances in the "last group", so that the group
+        # colors and scatter diagram marker colors will match
+        values.append("Unselected")
+        mask = (selection != 0)
+        selection = selection.copy()
+        selection[mask] = selection[mask] - 1
+        selection[~mask] = selection[~mask] = max_sel
```
markotoplak: Why not just subtract 1 from everything and then replace -1 with something else?

VesnaT: Because in some cases array's dtype is "uint8". In that case subtracting 1 from 0 results in 255.
I was thinking about changing dtype, but I think the above solution is more robust.

PR #3061, opened 2018-06-11
[FIX] BSD compatibility patch

yurivict:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/classification/_simple_tree.c
+++ b/Orange/classification/_simple_tree.c
@@ -53,7 +53,7 @@ enum { IntVar, FloatVar };
  *
  */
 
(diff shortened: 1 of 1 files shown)


lanzagar:
Although we don't _really_ support BSD (not included in any of our test suites), I don't mind a few fixes if that is all it takes to make it work. I would not be surprised, however, if there are problems elsewhere too...

Thanks for the patch!

yurivict:
> I would not be surprised, however, if there are problems elsewhere too...

I've created the FreeBSD port for Orange. It launches fine, although I didn't test it in-depth.

Yuri

PR #3064, opened 2018-06-13
[ENH] RecentFiles: Check for missing file in workflow dir

VesnaT:
##### Description of changes
When opening saved workflow and file widget detects missing file, check if the file is placed in the same directory as workflow is saved.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/tests/test_filedialogs.py
+++ b/Orange/widgets/utils/tests/test_filedialogs.py
@@ -1,0 +1,22 @@
+import os
+import unittest
+from tempfile import NamedTemporaryFile
+
+from Orange.widgets.utils.filedialogs import RecentPath
+
+
+class TestRecentPath(unittest.TestCase):
+    def test_resolve(self):
+        temp_file = NamedTemporaryFile(dir=os.getcwd(), delete=False)
+        file_name = temp_file.name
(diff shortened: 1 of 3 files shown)


Orange/widgets/utils/tests/test_filedialogs.py
```diff
@@ -0,0 +1,16 @@
...
+class TestRecentPath(unittest.TestCase):
+    def test_resolve(self):
+        f = open("test.tab", 'w')
+        f.close()
+        recent_path = RecentPath(
+            os.path.join("temp/datasets", "test.tab"), "", "datasets/test.tab"
+        )
+        search_paths = [("basedir", os.getcwd())]
+        self.assertIsNotNone(recent_path.resolve(search_paths))
+        os.remove("test.tab")
```
lanzagar: This looks like a classic trap of removing a file at the end, which might not happen if something goes wrong before the remove statement.
And here it is actually not that unusual since you have an assert above that might fail - try running this test on master.

I think there should be some utility functions for testing on temporary files somewhere... In any case some form of `finally: cleanup` is needed and I suggest using python's TemporaryFiles too.

PR #3084, opened 2018-06-22
[FIX] OWRank: Select Attributes fixes and improvements

thocevar:
- allow change of selection method from All to Manual
- add selection method and selected_rows to context settings

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -182,7 +182,7 @@ class OWRank(OWWidget):
 
     SelectNone, SelectAll, SelectManual, SelectNBest = range(4)
 
-    nSelected = Setting(5)
+    nSelected = ContextSetting(5)
     auto_apply = Setting(True)
 
     sorting = Setting((0, Qt.DescendingOrder))
@@ -339,6 +339,7 @@ class OWRank(OWWidget):
             self.selectionMethod = OWRank.SelectNBest
 
         self.openContext(data)
+        self.selectButtons.button(self.selectionMethod).setChecked(True)
 
     def handleNewSignals(self):
         self.setStatusMessage('Running')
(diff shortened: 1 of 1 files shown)


thocevar:
Loading the manually selected rows from context settings doesn't work. Method autoSelection doesn't seem to successfully select the rows in the view's selection model. As a consequence, on_select overwrites the selected_rows with an empty list.

ales-erjavec:
> Loading the manually selected rows from context settings doesn't work. 

Try
```diff
diff --git a/Orange/widgets/data/owrank.py b/Orange/widgets/data/owrank.py
index 83436d3ae..c0a009ff8 100644
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -339,7 +339,7 @@ class OWRank(OWWidget):
             self.selectionMethod = OWRank.SelectNBest
 
         self.openContext(data)
-        self.setSelectionMethod(self.selectionMethod)
+        self.selectButtons.button(self.selectionMethod).setChecked(True)
 
     def handleNewSignals(self):
         self.setStatusMessage('Running')
```
`setSelectionMethod` calls  `autoSelection`, but in `set_data` the `ranksModel` is still empty.

Orange/widgets/data/owrank.py
```diff
@@ -523,9 +523,6 @@ def commit(self):
         selected_attrs = []
         if self.data is not None:
             attributes = self.data.domain.attributes
-            if len(attributes) == len(self.selected_rows):
```
lanzagar: You can delete the row above as well and use the long version below (it is now only used once)

lanzagar:
While fixing this, could you also change it so that commit is only called once.

PR #3085, opened 2018-06-22
[ENH] Canvas: 'Window Groups' continued

ales-erjavec:
##### Issue

Applies on top of #3066

Cleanup and extend 'Window  Groups' functionality.

##### Description of changes

* Add a 'Bring widget to front' action/shortcut <kbd>ctrl + down_arrow</kbd>
* Preserve window stacking order for saved/activated window groups.
* Move implementation for save/restore window group to WidgetManager.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
ctrl + down_arrow does not seem to do anything on my system... (but if if works for others I am fine with the explanation that my system is strange :))

Also I find the wording of these menu options a bit confusing (Bring widgets to front, Display widgets on top).

ajdapretnar:
On my system, ctrl+down_arrow works. However, when the focus is on a widget, it behaves like regular up/down, despite holding ctrl. For example, when in Edit Domain, ctrl+down_arrow would walk through a list of variables and not bring widgets to front. If this is by design, I'm fine with that. I just thought adding ctrl would have a different functionality than down_arrow used on its own.

One more thing. It is not a consequence of this PR, but since the PR is meant to work on extended screens, I'll just mention it. My full screen Orange looks weird.

screen shot 2018-08-03 at 14 06 21

ales-erjavec:
ctrl+down_arrow works only when the canvas has focus. It's intended as a counterpart to the ctrl + up_arrow in the widget and a more light-weight version of the window groups.

PR #3089, opened 2018-06-25
Reorganize projections

janezd:
I wanted to reimplement the network widget to use the same code base that is used for the scatter plot. I discovered that before doing this, it is about time to reimplement the projections.

I prepared a workable prototype of Scatter plot and MDS, and I welcome any serious comments.

I won't work on this for a while (my summer project is going to be unrelated to Orange), but this prototype should suffice for somebody (maybe me) to continue this in the autumn.

### Issue

TL;DR: `ScatterPlotGraph` is messy and unmaintainable.

**Mess.** The code is impossible to follow. Method `OWScatterPlotGraph.new_data` clears the graph and calls `set_data`, which is inherited by `OWScatterPlotGraph` from `ScaledData`, which scales the data (without any obvious reason, it needs to be unscaled later) and jitters it. And that's it: after calling `OWScatterPlotGraph.new_data`, the widget (scatter plot, at least) calls its `update_graph`, which resets the `zoomStack` (that is no longer used anyway) and then calls `OWScatterPlotGraph.update_data`, which is the obvious name for a method the removes everything from the plot (again) and redraws it from scratch.

`update_data` seems to be actually the only method to change the plot, except for some others that also change it (mess intended) and some of which are called by `update_data`. Since most callbacks are routed through method that sometimes reset too much, they have additional parameters like `resetView` (to reset the view, apparently, though it's not clear whether not resetting it makes any difference) and `keepColors` to reset the colors but keep them.

For instance,  there is `update_density` which should update the density, but instead calls `update_graph` that resets the entire graph.

Due to lack of separation between the model and the view, a lot of notification functions need to be called back and forth.

**Passing data in `Orange.data.Table`**: `OWScatterPlotGraph` was getting the data from `Orange.data.Table`. This is not a problem for color/shape/size, although the latter was problematic in MDS. It, however, is a problem for coordinates, since the Scatter plot is the only widget that actually has coordinates in `Orange.data.Table`. Everybody else needed to pack the data into a table to pass it to the graph - which the extracted the needed columns. Besides being annoying, this resulted in MDS tooltips showing mds-x and mds-y instead of useful attribute values.

**Consequences of unification of projections.** Since all projections were forced into using this messy class, the class contains everything that is needed by any of them. Checkboxes are controlled by `OWScatterPlotGraph`, so since the Scatter plot widget has a checkbox for showing linear regression line, which calls `update_regression_line`, also PCA, MDS and Linear projection had to implement this method. Even better, the scatter plot widget has to know when the color setting changes, so it can reset VizRank. Since the callback for the combo, `update_color`, is in `OWScatterplotGraph`, the graph's `update_color` called `master.update_color`. Although only the scatter plot needs this method, everybody else had `def update_color(self): pass`.

**Other curiosities, like: Handling of sparse data.** This is fun: the widget stores the original data into an attribute `_data`. Then upon each change of color, size, shape or label, it checks whether any of the needed attributes are sparse and if so, it calls its own `self.setData` with densified data (gotten from the stored `_data`) that contained only the needed attributes. I hope these are the only widgets that call their own signal handler to transform its data (or for any other purpose).

#### Conceptually clearer methods

I reimplemented this in a model-view-ish fashion.

The model is actually the widget. There could be a separate model class, but this would multiply the number of needed classes without any obvious benefit. (Essentially, the widget is the controller and the model at once.)

The model needs to provide methods:

- `get_coordinates_data`, `get_size_data`, `get_color_data`, `get_shape_data`, `get_label_data` which return a 1d array (or two arrays, for `get_xy_data`) of `dtype` `float64`, except for `get_label_data`, which returns formatted labels;
- `get_color_labels`, `get_shape_labels`, which are used to construct the legend;
- `is_continuous_color`, which is also used for the legend;

Here are some other methods that perhaps belong to the "view", not the "model" (widget):

- `impute_sizes`, `impute_shapes`, which are called to replace nan's from `get_size_data` and `get_shape_data`;
- `set_palette` - maybe the view should handle the colors;
- `can_draw_density` - you can always draw densities if you have different colors, right?

Note that is any widget has to patch the behaviour of the "view" (graph), it should just derive the new view. E.g. the scatter plot uses a derived view that adds a regression line.

Any of the `get_something` methods can return `None` and all points will have the same color, size and/or shape. If `get_coordinates_data` returns `None, None`, there is no data.

The view has methods `reset_graph`, which should be called only when
- the widget gets entirely new data
- the number of points may change, for instance when selecting x and y in the scatter plot, since the attributes may have some nans.

Otherwise, the "view" provides methods `update_coordinates`, `update_sizes`, `update_colors`, `update_shapes` and `update_labels`, which reset only what they are supposed to reset. These methods are the ones that call `get_something`. E.g. the model's `get_color_data` is called only from the view's `update_colors`.

All `update_something` are very similar, with detail differences due to different nature of properties the control.

Every `update_something` calls the view's `get_something`, which calls the model's `get_something_data`, then it transforms this data into whatever is needed (colors, shapes, scaled sizes) and changes the plot. For the simplest example, here is `update_shapes`:

```
    def update_shapes(self):
        if self.scatterplot_item:
            shape_data = self.get_shapes()
            self.scatterplot_item.setSymbol(shape_data)
        self.make_legend()

    def get_shapes(self):
        shape_data = self.master.get_shape_data()
        shape_data = self.master.impute_shapes(shape_data, len(self.CurveSymbols) - 1)
        return self.CurveSymbols[shape_data]

```

On the model side, `get_something_data` is essentially just:

```
    def get_size_data(self):
        return self.get_column(self.attr_size)
```

where `get_column` retrieves a column while also filtering out the points with missing x and y and so forth. (I'm obviously cheating by showing "shapes" for the view and "sizes" for the model. The colors for the view are more complicated since they deal with different palettes etc,  and the shapes for the view merge infrequent shapes. But conceptually, every function's role is clear.)

#### Separation of settings

The view doesn't know about attributes. The model is solely responsible for getting the data, the view is responsible for showing it. Obviously.

The model (widget) contains settings for combos with attributes. The model has jittering, densities etc.

When derived classes add stuff, everything related to presenting the data, goes into the graph. The Scatter plot widget derives a new view with settings `jitter_continuous_data` and `show_regression_line`.

The widget should not peek into the graph's settings and vice versa. This is a useful implementation guideline.

(Remember the problem with the graph's `update_color` calling the master's `update_color` to inform it that it needs to reset VizRank? This happened because the view controlled the color attribute.)

#### A common base for the widget

I defined an `OWProjectionWidget` with settings `attr_color`, `attr_label`, `attr_shape`, `attr_size`.

To this, `OWScatterPlot` adds 
- `attr_x` and `attr_y`, and `auto_sample`, which are specific
- `auto_send_selection`, as any other widget, 
- `toolbar_selection`, since I fell `OWProjectionWidget` should be limited to what is required from a model
- `tooltip_shows_all` since this is different for the scatterplot, which primarily shows x and y, and all other widgets, hence it doesn't belong to the base widget.

OWScatterPlot then only defines `get_coordinates_data`.

Besides, scatter plots derives a new view that handle (actually: disable if not enabled) jittering of continuous vaues and regression lines.

Similarly, MDS derives a view that shows connected pairs and contains related settings, while the model defines `get_coordinates_data` and overrides `get_sizes` to set the sizes according to stress, if the user chooses so. During optimization, the MDS simply calls `graph.update_coordinates`, which updates the graph without actually resetting it.

#### Changes in GUI

I changed the GUI by grouping the properties with the combos. For instance, I moved "Show density" checkbox below the combo for colors -- before, nobody knew that it's there and what it meant. (I also renamed it to "Show color regions". 

I know we had a similar layout in the past, so I'm regressing. Partially, I realized that it's better to have the properties grouped. In addition, I believe that the new layout is still more compact than the old old one.

This change is unrelated to the entire issue and I regret it, but if you agree with it, please don't force me to split it into two different PRs just because it's the way it should be. (I've lost a lot of time with such requests before, only to discover that exactly those who make them tend to merge unrelated things into the same PR. So I'm not doing it.)

If we, however, decide that we don't like the new GUI, I can go back.

#### TO DO

The Scatter plot and MDS mostly work, though not without bugs. I am not porting PCA and Linear projections before getting comments.

The code needs some cleaning; there may still be unused stuff, but now we can at least see what's called and where, and what not.

There are no tests for the new code; but at least it's now testable.

I made some incompatible changes to owplotgui. This is not unavoidable, though if we care about the backward compatibility I'd prefer to not use it in my new code at all. I doubt anybody uses it.

`OWScatterPlotGraph` is provisionally renamed to `OWScatterPlotBase`; could be renamed to something else or keep the name. The question is whether to keep the old one for backward compatibility.

I removed the module `scaling` in hope the nobody uses it. It was there just for jittering, which can be replaced with a three-line function.

I haven't migrated the settings yet. Waiting for comments before I do.

There are some other parts I haven't touched. `Report` probably does not work.

After we're done, we can actually document these classes so others can use them, too. :)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
Janez, thank you, this was long overdue. 

I like the separation of functionality and how you add specifics to the basic plot. Also, your GUI changed are an usability improvement (placement of majority coloring option). And finally, we get to dump scaling.py!

I was pleasantly surprised how little did you have to change in OWScatterPlot and OWMDS to incorporate new base classes. I do not see any general issues.

janezd:
After discussing it with @BlazZupan (without technical details, though -- thanks for having looked into the PR!) I already asked @VesnaT to implement the missing parts (tests, reporting, migration of settings...) and to port other visualizations that use `ScatterPlotGraph`. She has her hands full right now, but she'll get to this eventually.

ajdapretnar:
@janezd This requires your review I believe.

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -598,480 +491,308 @@ def update_tooltip(self, modifiers):
...
+    def _jitter_data(self, x, seed=0):
+        random = np.random.RandomState(seed=seed)
+        off = self.jitter_size * (np.max(x) - np.min(x)) / 25
+        return x + random.uniform(-off, off, len(x))
+
+    # Sizes
+    def get_sizes(self):
+        size_column = self.master.get_size_data()
+        if size_column is None:
+            if self.master.attr_size == OWPlotGUI.SizeByOverlap:
```
janezd: Please don't do this. One of the basic rules of the refactoring (citing from the PR description) is **The view doesn't know about attributes.** The size make come from wherever. The widget does not necessarily have an attribute `attr_size`. This condition may only appear in the class derived from this class, if at all.

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -598,480 +491,308 @@ def update_tooltip(self, modifiers):
...
+                self.update_colors()  # color points by most frequent color
```
janezd: I don't like `update_coordinates`, `update_sizes`, `update_shapes`, `update_colors`, `update_labels` calling each other. I believe (hope?) that this is not violated anywhere, except here.

The logic of the separation between the widget and the plot is currently simple: if the widget calls `update_something`, the plot will respond by calling `get_something_data`. If it can also result in calling `get_somethingelse_data`, the whole logic of updating may again become difficult to follow.

If handling overlaps would be the responsibility of the plot - which I believe is not, it must find the way out of this without re-calling `update_colors`.

However, I believe that this is responsibility of the widget. What happens in this code here is that the sizes are provided by the plot instead of by the widget. If we want sizes by overlaps, it is the widget that needs to compute the overlaps and return the appropriate sizes.

Also, the widget, not the plot, must provide the colors.

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -598,480 +491,308 @@ def update_tooltip(self, modifiers):
...
-                self.set_labels(axis, None)
+        self.update_label_coords(x, y)
+        self.update_density()  # Todo: doesn't work: try MDS with density on
+        self._reset_view(x, y)
 
-        # compute overlaps of points for use in compute_colors and compute_sizes
-        self.overlaps = []
         self.coord_to_id = defaultdict(list)
-        for i, xy in enumerate(zip(self.x_data, self.y_data)):
+        for i, xy in enumerate(zip(x, y)):
```
janezd: This code looks expensive, but it's needed only when sizing and coloring by overlaps.

Plus, I believe it belong to the widget, not the plot (see other comments).

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -598,480 +491,308 @@ def update_tooltip(self, modifiers):
...
+        off = self.jitter_size * (np.max(x) - np.min(x)) / 25
+        return x + random.uniform(-off, off, len(x))
+
+    # Sizes
+    def get_sizes(self):
+        size_column = self.master.get_size_data()
+        if size_column is None:
+            if self.master.attr_size == OWPlotGUI.SizeByOverlap:
+                self.update_colors()  # color points by most frequent color
+                return np.multiply(np.ones(self.n_points) * self.point_width,
```
janezd: You can use `np.full(self.n_points, self.point_width)` instead of `np.ones(self.n_points) * self.point_width`.

However, `return np.full(self.n_points, self.point_width) * self.overlap` doesn't work?

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -598,480 +491,308 @@ def update_tooltip(self, modifiers):
...
+                                   self.overlap_factor)
+            return np.ones(self.n_points) * self.point_width
```
janezd: `np.full(self.n_points, self.point_width)`?

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -1288,40 +969,186 @@ def box_zoom_select(self, parent):
...
+            freq_vals = [i for i, f in enumerate(infrequent) if not f]
+            for i, f in enumerate(infrequent):
+                result[all_data == i] = MAX - 1 if f else freq_vals.index(i)
+            return result
+
+    # Sizes
+    def get_size_data(self):
+        return self.get_column(self.attr_size)
+
+    def impute_sizes(self, size_data):
```
janezd: Move impute methods from widget to plot.

ajdapretnar:
Iris.tab, set Size to Overlap, then set color to None.

ajdapretnar:
Feature question:
- Is scaled difference when using the Symbol size slider really the best approach? I am not a big fan, but if there is a solid argument for its use, then let's leave it.
- When color is a numeric attribute and size discrete, we see two legends. If I deselect 'Show legend', only the size legend will disappear. Feature or a bug? (I'd go with a bug.)

janezd:
> Is scaled difference when using the Symbol size slider really the best approach? I am not a big fan, but if there is a solid argument for its use, then let's leave it.

What are the alternatives? Just dividing by the maximal value, without subtracting the minimal? This wouldn't work if the data comes from an interval far from zero, like 1900 - 2018. Of course we can let the user decide, but I'd hate adding too many options.

janezd:
> When color is a numeric attribute and size discrete, we see two legends. If I deselect 'Show legend', only the size legend will disappear. Feature or a bug? (I'd go with a bug.)

Bug, of course.

ajdapretnar:
> What are the alternatives? Just dividing by the maximal value, without subtracting the minimal? This wouldn't work if the data comes from an interval far from zero, like 1900 - 2018. Of course we can let the user decide, but I'd hate adding too many options.

No, no, this is not the problem. The problem is the slider is used as a multiplier of difference between points. It is coupled with Size attribute, while before it wasn't. I think these two should be independent, this is why I also suggested separate boxes. In my opinion, Symbol size refers to how big or small I want the points to be and when changing it, all the points will increase or decrease linearly. In other words, when I set symbol size to the lowest value and setting Size by some attribute, there is no difference between points. Which I think is confusing.

ajdapretnar:
I was wrong. This is the same as before. Still, not particularly fond of it, but since it doesn't change anything...

ajdapretnar:
Only a trained bug-hunting eye will notice a small gray box at the top right of the graph. This is the blank legend. Should be off when empty.

screen shot 2018-09-17 at 16 02 56

ajdapretnar:
When scrolling in and out, sometimes you get a weird result, like this. Which is fine.
screen shot 2018-09-17 at 16 05 09

But then you realize your mistake and would like to see the graph as before. You press the Reset zoom and it won't take you straight away to the nice plot. One has to press Reset zoom a couple of times to see the 'normal' graph.

This is the result of the first click:
screen shot 2018-09-17 at 16 05 12

ajdapretnar:
Filter in Find informative projections should not be case sensitive.

Also, I don't see a good use case for editable feature pairs. I think this should be off and clicking on the feature pair should always propagate the x, y to the plot. Now, when selecting the feature pair, changing one attribute manually in the widget and re-selecting the feature pair, the combo goes into edit mode instead of propagating the pair to the widget.

janezd:
> editable feature pairs.

Wow, you're indeed on a hunt here. :)

I could not replicate this exactly as you described, but found another way: you can edit a cell if you double-click it.

This is however unrelated to this PR and the module that needs fixing (the common Vizrank dialog is implemented in `Orange.widgets.visualize.utils`) has not been touched by the PR, so I opened this as two separate issues (case sensitivity, editing).

ajdapretnar:
> There is, just smaller.

The problem is, the difference is not noticeable. At least to me, they all look the same. Unless you have very sharp eyes. 😁 

screen shot 2018-09-18 at 10 01 09

> Wow, you're indeed on a hunt here.

You told me to do a job, I do the job. 🐛 🔫

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -598,493 +569,530 @@ def update_tooltip(self, modifiers):
...
+        """
+        size_column = self.master.get_size_data()
+        if size_column is None:
+            return np.ones(self.n_points) * self.point_width
+        size_column = size_column.copy()
+        size_column -= np.min(size_column)
+        mx = np.max(size_column)
+        if mx > 0:
+            size_column /= mx
+        return self.MinShapeSize + self.point_width * size_column
```
janezd: Maybe `return self.MinShapeSize + (5 + self.point_width) * size_column`. @ajdapretnar, can you try and see if you like it?

janezd:
The slider goes from 1 to 20. When its all the way to the left, the size is either 6 or 7. If you cant' see the difference you need to have your eyes checked.

Orange/widgets/visualize/owscatterplotgraph.py
ajdapretnar: Yes, that's better. At least I can see there some difference between points even at the smallest scale.

janezd:
Remaining minor issues, including lint complaints, will be resolved in the following days in separate PRs.

PR #3090, opened 2018-06-26
[FIX] owkmeans: fix initialization choice

markotoplak:
##### Issue
Initialization (random or k-means++) was wrongly passed from widget to the library.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


lanzagar:
Nice catch. While you are fixing this, could you improve this one step further and remove the constants in the middle of the code?
I think something like changing `INIT_METHODS` to a dict or tuple of tuples with both our strings and scikit parameters would be much clearer and more future-proof.

markotoplak:
@lanzagar, thank you, I did as you suggested.

Issue #3103, opened 2018-06-29
Replicable workflows


markotoplak:
Same data set + same algorithm = same results, unless explicitly randomized

So we have to set random seed in widgets to some fixed value by default. I propose these to saved as widget setting `random_state` (an integer).

Widgets that need changing:
- [ ] Data Sampler need a default Replicable sampling as chosen by default
- [x] k-means needs a set random seed #3075
- [X] Random forests already have a `random_state` setting
- [ ] Is this an issue in Test&Score? #2035
- [ ] Logistic regression with liblinear solver (used in the widgets) can be stochastic

There is still a question whether we want to expose these seeds. Would perhaps be the Data Sampler (which could shuffle the data if we allow the proportion to go to 100%) be enough for showing effects of randomness?

kernc:
Could we make our "Learners" have a default param `random_state=0` (instead of `=None`). It's simple but covers a lot.

janezd:
#3955 solves the first point and sets random seed for the last point - logistic regression. Is there another source of randomness within the solver?

Test & Score is already non-random, I think.

Does this close the issue? If we want to make all workflows replicable, somebody must comb through all the code. The task may be neigh impossible. For instance, even using `max(d, key=d.get)` to find the dict key belonging to the maximal value may yield random results.

markotoplak:
I think for LR this should be it. By default, sklearns  `random_state` is None and we usually leave it as default.

Should we fix such random seeds only in widgets or in the library and we could then leave widgets as they are?

Also PCA (and probably many other projection) would need to have random_seed fixed by default, because we are using a randomized variant of it for speed. Just last week I was wondering for a while why I was getting different results.

markotoplak:
I checked this a bit. So, in the library, we are currently using `random_state=None` everywhere but in pca.randomized_pca(), CrossValidation(), ShuffleSplit(). There, the default is 0.

PR #3104, opened 2018-06-29
owkmeans: always use the same random seed

### LINKED ISSUE #3075: k-means restores different clusters when loading a workflow
##### Expected behavior
Saving a workflow and opening it should return exactly the same clusters. 

Perhaps k-means should also have a replicable option similar to the data sampler widget. But I think that saving and opening a workflow should return the same clusters even if replicable is not selected. 

##### Actual behavior
K-means results change if you save and open schemas. Because also the order of classes changes, everything downstream is problematic.

markotoplak:
##### Issue
#3103

Fixes #3075

##### Description of changes
K-means now always uses the same random seed and outputs the same data

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owkmeans.py
+++ b/Orange/widgets/unsupervised/owkmeans.py
@@ -19,6 +19,9 @@ from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.widget import Input, Output
 
 
+RANDOM_STATE = 0
+
+
 class ClusterTableModel(QAbstractTableModel):
     def __init__(self, parent=None):
         super().__init__(parent)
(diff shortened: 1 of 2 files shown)


Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -128,6 +128,7 @@ class Warning(widget.OWWidget.Warning):
     max_iterations = Setting(300)
     n_init = Setting(10)
     smart_init = Setting(0)  # KMeans++
+    random_state = Setting(0)
```
kernc: You make this a setting, but it's not tunable?

lanzagar: Now that you point it out, it looks like at least a comment here would be appropriate.
But to explain, we discussed how to allow users to change the random seed and not yet come to the final conclusion. Maybe a common option in Orange's settings or per widget buttons to change the seed... But everyone mostly agrees that widgets should by default produce reproducible results, so we want to start fixing the seeds and then add the option to change that if/where we decide it is best (proposals should probably go into #3103).
For now I don't mind if random_state is a constant or a setting (with a comment to avoid confusion).

markotoplak: I made it a constant.

PR #3109, opened 2018-07-02
[FIX] postgres: Fix wrong discrete values

### LINKED ISSUE #3108: owsql: Wrong auto-discovered DiscreteVariable values
##### Orange version
3.14.0 dev

##### Expected behavior
SQL Table correctly discovers categorical variables.

##### Actual behavior
`DiscreteVariable` values for char fields have additional with space on the end.

SQL for Postgres:

``` sql
CREATE TABLE COMPANY(                                                                                                
   ID INT PRIMARY KEY     NOT NULL,
   ADDRESS        CHAR(50)
);
INSERT INTO COMPANY (ID,ADDRESS) VALUES (1, 'California');
```
In SQL Table select tabel company

This issue is because on data query `::text` is used that removes the additional white space.

robertcv:
##### Issue
Fixes #3108 
##### Description of changes
Remove traling spaces.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/sql/backend/postgres.py
+++ b/Orange/data/sql/backend/postgres.py
@@ -165,6 +165,8 @@ class Psycopg2Backend(Backend):
         if type_code in CHAR_TYPES:
             if inspect_table:
                 values = self.get_distinct_values(field_name, inspect_table)
+                # remove trailing spaces
+                values = [v.rstrip() for v in values]
                 if values:
                     return DiscreteVariable.make(field_name, values)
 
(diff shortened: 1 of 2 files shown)


Orange/data/sql/backend/postgres.py
```diff
@@ -165,6 +165,8 @@ def _guess_variable(self, field_name, field_metadata, inspect_table):
         if type_code in CHAR_TYPES:
             if inspect_table:
                 values = self.get_distinct_values(field_name, inspect_table)
+                # remove trailing spaces
+                values = [v.rstrip() for v in values]
```
astaric: Why not modify query to cast to ::text like it will be done in further selects?

robertcv: `get_distinct_values` encapsulates `filed_name` into quotes and the query becomes invalid. On request, I move the encapsulation from `get_distinct_values` and add outside function calls.

astaric: no, it is ok the way it is.

PR #3111, opened 2018-07-03
[ENH] Move Louvain clustering from prototypes to core

### LINKED ISSUE #3110: Louvain Clustering
Orange is missing some more advanced means of data clustering. One of them is network-based Louvain clustering that is becoming popular due to speed and almost-automatic assignment of number of clusters.

- develop Louvain clustering and store it in Orange/Clustering/louvain.py
- include a corresponding widget and place it by priority just after k-Means widget

pavlin-policar:
##### Issue
Fixes #3110

##### Description of changes
Move Louvain clustering widget from prototypes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/unsupervised/owlouvainclustering.py
```diff
@@ -0,0 +1,361 @@
...
+    Graph = None
+
+
+_MAX_PCA_COMPONENTS = 50
+_DEFAULT_PCA_COMPONENTS = 25
+_MAX_K_NEIGBOURS = 200
+_DEFAULT_K_NEIGHBORS = 30
+
+
+METRICS = [('Euclidean', 'l2'), ('Manhattan', 'l1')]
```
lanzagar: Are these the only two metrics that are supported?

Orange/widgets/unsupervised/owlouvainclustering.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def _update_pca_components(self):
+        self._invalidate_pca_projection()
+        self.commit()
+
+    def _update_metric(self):
+        self._invalidate_graph()
+        self.commit()
+
+    def _update_k_neighbors(self):
```
lanzagar: As long as you do not expect these methods will be expanded in the future with code specific to changing one of the settings I don't see a reason to have 3 methods with the same body but different names.
I think having just one _update_graph and using it as the callback for all 3 controls would be better.

Orange/clustering/louvain.py
```diff
@@ -0,0 +1,123 @@
...
+
+"""
+
+import networkx as nx
+import numpy as np
+from sklearn.neighbors import NearestNeighbors
+
+import Orange
+from Orange.data import Table
+from community import best_partition
```
lanzagar: I guess this import belongs to the above group of external libraries now

lanzagar:
The widget works well as far as I was able to test it.

What I am missing is documentation - at least a stub with something completely basic that can be expanded later.
Also, I was wondering what numbers to set for Resolution and was thinking that a tooltip explaining what I am doing there wouldn't be bad either. But maybe when the widget has some documentation, that will be enough.

Another thought was, why not a slider for Resolution like for PCA components. If anything I would sooner expect a spinbox for the number of components (everybody understands the number, I know what I am doing when setting e.g. 5 or 20). For Resolution, I have no idea what 0.5 means and how much bigger 1.5 is or 2 (or why 5 is the max). So a resolution slider from smaller scale communities to larger scale could make more sense.

lanzagar:
Oh, and the only issue I have with the functionality is that the pca slider now commits on any change so dragging it starts a lot of computations and isn't smooth. Commit should happen when I stop dragging.

Orange/widgets/unsupervised/owlouvainclustering.py
pavlin-policar: For the time being, only these two metrics are supported because of the way sklearn computes nearest neighbours. The only other metrics I had considered adding was the cosine, pearsonr and spearmanr distances, but unfortunately the fast tree methods for finding nearest neighbours in sklearn don't support these and fall back to the O(n^2) brute force approach, which is prohibitively expensive for larger data sets, so we decided it is better to leave them out.

If you feel any of the other supported metrics should be added, they are listed in the source: [KDTree](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/neighbors/kd_tree.pyx#L14-L15) and [BallTree](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/neighbors/ball_tree.pyx#L13-L22). Most of these don't appear anywhere else in Orange, so including them doesn't really make sense.

pavlin-policar:
Using sliders makes more sense, I agree. I've changed it now. As far as the documentation goes, I can write some documentation, but I can't for the life of me figure out how to incorporate that into Orange.

ajdapretnar:
Ajda to the rescue! 🏇
Here is where you put the .rst file of your widget: https://github.com/biolab/orange3/tree/master/doc/visual-programming/source/widgets
Best thing is to copy an existing file and just change the content, so you keep the form. Then just include it in your PR. You will have to run make html to see how the output looks like.
Also important:
- stamper for numbers on images can be found here: https://github.com/biolab/orange-tools
- in that same repo you will find trim.py that removes a background of the image (say for complex workflows)
- you will have to index images for tests to pass (I recommend Gimp)

If you have any questions, just drop by. I would be happy if you just provide the key content, I can later take care of the rest.

pavlin-policar:
Great, thanks! I had no idea this existed.

I've added a fairly minimal stub explaining what the controls do and how the widget works. I've tried to make the stamping consistent with the other widgets. Let me know if I should add anything or anything is poorly worded.

doc/visual-programming/source/widgets/unsupervised/louvainclustering.rst
```diff
@@ -0,0 +1,38 @@
+Louvain Clustering
+=======
```
ajdapretnar: You need to extend the signs to encompass at minimum the entire title.

doc/visual-programming/source/widgets/unsupervised/louvainclustering.rst
```diff
@@ -0,0 +1,38 @@
...
+    Graph (with the Network addon)
+		the weighted k-nearest neighbor graph
+
+
+The widget first converts the input data into a k-nearest neighbor graph. To preserve the notions of distance, the Jaccard index for the number of shared neighbors is used to weight the edges. Finally, a `modularity optimization <https://en.wikipedia.org/wiki/Louvain_Modularity>`_ community detection algorithm is applied to the graph to retrieve clusters of highly interconnected nodes. The widget outputs a new dataset in which the cluster index is used as a meta attribute.
+
+
+.. figure:: images/Louvain-stamped.png
+
+1. PCA processing is typically be applied to the original data to remove noise.
```
ajdapretnar: typically ~be~ applied

doc/visual-programming/source/widgets/unsupervised/louvainclustering.rst
```diff
@@ -0,0 +1,38 @@
...
+2. The distance metric is used for finding specified number of nearest
+   neighbors.
+3. The number of nearest neighbors to use to form the KNN graph.
+4. Resolution is a parameter for the Louvain community detection algorithm that
+   affects the size of the recovered clusters. Smaller resolutions recover
+   smaller, and therefore a larger number of clusters, and conversely, larger
+   values recover clusters containing more data points.
+5. When *Apply Automatically* is ticked, the widget will automatically
+   communicate all changes. Alternatively, click *Apply*.
+
```
ajdapretnar: At the bottom it is great to provide a minimum workflow for the widget. Or a cool way how to use it, preferably on a simple data set that exists in Orange.

ajdapretnar:
🤔 I think tests won't pass if images aren't indexed. [This](https://en.wikipedia.org/wiki/Indexed_color) is what I had in mind. It is as simple as opening Gimp and using Image - Mode - Indexed (Convert). Then File - Export As (and click all the Export buttons that appear). This is what you add then to the repo.

pavlin-policar:
@lanzagar I've fixed the images, all the tests are passing. The only thing remaining are some pointless pylint errors. Should I try to fix them and add ignore statements where necessary or is this fine?

Orange/widgets/unsupervised/owlouvainclustering.py
```diff
@@ -0,0 +1,377 @@
...
+
+            if self.__cancelled:
+                self.on_cancel.emit()
+                return
+
+            def __task_progress(percentage):
+                current_progress = idx / num_tasks
+                # How much progress can each task contribute to the total
+                # work to be done
+                task_percentage = len(self.__tasks) ** -1
```
astaric: What is wrong with 1 / len(self.__tasks)? :)

pavlin-policar: I think I wrote this code at a point where I was doing a considerable amount of calculus and I guess I found this clearer at that point.

pavlin-policar:
@lanzagar The only remaining issue is the coverage. I've written all the tests that I think are reasonable for this widget, and I think most of the code is hit at least once, so I'm not sure where this poor coverage comes from. Should I look into this or can this be merged?

astaric:
https://codecov.io/gh/biolab/orange3/compare/56fdfc936d0f8294a335e1ef9ff1f8889ffc0bbc...1e705f9a629ee488143b46ed6ffef7ce824c75a3/diff

PR #3113, opened 2018-07-03
Add support for PostgreSQL servers without extensions

### LINKED ISSUE #3097: OWSql: Support PostgreSQL servers without extensions
Orange requires two extensions to be installed on PostgreSQL server it connects to; quantile (used for discretization) and tsm_system_time (used for efficient sampling). Instructions on how to install these extensions on the server are available on wiki: https://github.com/biolab/orange3/wiki/Installation-of-PostgreSQL-extensions

Many users, however, connect to servers that do not have these extensions installed and sometimes do not have the necessary permissions to install them. For a list of reports, see gh-2083

As extensions are only required for online work with the data, owsql widget could be modified to force offline mode (enable Download data to local memory checkbox and make it non-editable) and show a warning instead of an error. This way, users can still access the data, they just need to load the whole table into memory to perform the analysis.

robertcv:
##### Issue
Fixes #3097 
##### Description of changes
Replace error on PostgresSQL missing extensions with warning and forced data download.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


robertcv:
There are no tests for this widget. Should I add some based on [sql tests](https://github.com/biolab/orange3/tree/master/Orange/tests/sql)?

Orange/data/sql/backend/postgres.py
```diff
@@ -45,8 +46,9 @@ def _create_extensions(self):
                 query = "CREATE EXTENSION IF NOT EXISTS {}".format(ext)
                 with self.execute_sql_query(query):
                     pass
-            except OperationalError:
+            except BackendError:
                 warnings.warn("Database is missing extension {}".format(ext))
+                self.missing_extension = ext
```
lanzagar: Are you sure you want to overwrite  `missing_extension` with ext in the for loop?
This will only save the last one missing... maybe missing_extensions could be a list and ext is appended here.

Orange/widgets/data/owsql.py
```diff
@@ -235,6 +240,12 @@ def connect(self):
                 password=self.password
             ))
             self.Error.connection.clear()
+            if hasattr(self.backend, 'missing_extension') and \
+                    self.backend.missing_extension:
```
lanzagar:
```py
if getattr(self.backend, 'missing_extension', False):
```

Orange/data/sql/backend/postgres.py
```diff
@@ -22,6 +22,7 @@ class Psycopg2Backend(Backend):
     display_name = "PostgreSQL"
     connection_pool = None
     auto_create_extensions = True
+    missing_extension = ""
```
lanzagar: This should probably be an instance attribute declared in `__init__` as it depends on the connection_params

Orange/widgets/data/owsql.py
```diff
@@ -74,7 +74,9 @@ class Information(OWWidget.Information):
     class Error(OWWidget.Error):
         connection = Msg("{}")
         no_backends = Msg("Please install a backend to use this widget")
-        missing_extension = Msg("Database is missing extension{}: {}")
+
+    class Warning(OWWidget.Warning):
+        missing_extension = Msg("Database is missing extension: {}")
```
lanzagar: The previous Error printed all missing extensions. I think that is better than just showing 1 (if 2 are missing), as it can be a pain to install that one and come back to Orange to find out later that another one is missing too.

PR #3115, opened 2018-07-04
[ENH] io: Speedup write_data

ales-erjavec:
##### Issue

Speed `write_data`.

##### Description of changes

Assemble column formatters beforehand, avoid inline logic deciding the
correct format for every cell value.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -13,7 +13,7 @@ from functools import lru_cache
 from importlib import import_module
 from itertools import chain, repeat
 from math import isnan
-from numbers import Number
+
 from os import path, remove
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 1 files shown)


ales-erjavec:
```ipython
In [1]: import Orange
In [2]: adult = Orange.data.Table("adult")
In [3]: %timeit adult.save("test-save-temp.tab")
```
before:
```1.07 s ± 7.55 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)```
after:
```386 ms ± 5.88 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)```

BlazZupan:
This is a great speed-up. I have tested it on data with over 20,000 features and 100 data instances, where speed-up was 6-fold.

PR #3117, opened 2018-07-04
[ENH] Match Keywords in Widget Search

irgolic:
##### Issue
At the moment, searching for widgets only queries widget titles. This leads to one having to type the widget's exact name to find it (e.g. "k-means"). With searching by keyword, widgets can be assigned other tags (e.g. "means", "kmeans").

##### Description of changes
Overriding filterAcceptsRow() in quick menu, matching both keywords and widget title.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/document/quickmenu.py
+++ b/Orange/canvas/document/quickmenu.py
@@ -325,18 +325,16 @@ class SortFilterProxyModel(QSortFilterProxyModel):
     def filterAcceptsRow(self, row, parent=QModelIndex()):
         flat_model = self.sourceModel()
         index = flat_model.index(row, self.filterKeyColumn(), parent)
-        keywords = flat_model.data(index, role=QtWidgetRegistry.WIDGET_DESC_ROLE).keywords
+        description = flat_model.data(index, role=QtWidgetRegistry.WIDGET_DESC_ROLE)
+        name = description.name
+        keywords = description.keywords
 
-        # match keywords
+        # match name and keywords
         accepted = False
-        for keyword in keywords:
+        for keyword in [name] + keywords:
             if self.filterRegExp().indexIn(keyword) > -1:
                 accepted = True
                 break
-
-        # if does not match keywords, match title
-        if not accepted:
-            accepted = QSortFilterProxyModel.filterAcceptsRow(self, row, parent)
 
         # if matches query, apply filter function (compatibility with paired widget)
         if accepted and self.__filterFunc is not None:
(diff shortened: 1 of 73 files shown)


irgolic:
Note: Most widgets currently do not hold keyword data.

markotoplak:
A very important improvement. Thank you for working on it!

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -90,6 +90,7 @@ class OWKMeans(widget.OWWidget):
                   "quality estimation."
     icon = "icons/KMeans.svg"
     priority = 2100
+    keywords = ["means", "kmeans"]
```
lanzagar: "means" is unnecessary as that worked before keywords already... The names are split into words so k-Means matched k and Means (and prefixes).
I would add "clustering" however.

lanzagar:
Another comment: a commit like 6d5f8c3 (reimplemented title matching) is usually not useful to have in the git history. When fixing/improving/reworking your own changes you should squash the commits into one containing the final version.

markotoplak:
I see that to some widgets empty keywords were added. Widgets should also work properly if the keywords were not defined (as it is the case of most current add-on).

Commits with undefined keywords (keywords = []) should be removed from this PR.

Orange/canvas/document/quickmenu.py
```diff
@@ -323,7 +323,20 @@ def filterFunc(self):
...
-        accepted = QSortFilterProxyModel.filterAcceptsRow(self, row, parent)
+        flat_model = self.sourceModel()
+        index = flat_model.index(row, self.filterKeyColumn(), parent)
+        description = flat_model.data(index, role=QtWidgetRegistry.WIDGET_DESC_ROLE)
+        name = description.name
+        keywords = description.keywords
+
+        # match name and keywords
+        accepted = False
+        for keyword in [name] + keywords:
```
kernc: How about also iterating over `filter(None, re.split(r'\W+', description.description.lower()))`? This way, many widgets won't even need separate keywords, e.g. K-means already has "clustering", "silhouette", ...

irgolic: I'd rather manually set keywords to be able to finely control what shows up in search, avoiding the overcrowding of search results. Each widget shouldn't need more than a few keywords I think.

irgolic:
To clarify, widgets work properly without keywords defined (even without an empty list of keywords). I just added an empty list in case someone notices it in the future and thinks of something good to fill it with.

PR #3121, opened 2018-07-05
ubuntu 16.04 lts

d-demirci:
remediation for error:
    from PyQt5.QtCore import *
ImportError: /home/forensic/Documents/Workspace/orange-ml/lib/python3.5/site-packages/PyQt5/QtCore.so: undefined symbol: PySlice_AdjustIndices

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/requirements-doc.txt
+++ b/requirements-doc.txt
@@ -3,4 +3,4 @@ docutils
 numpydoc
 recommonmark>=0.1.1
 Sphinx>=1.3
-PyQt5
+PyQt5==5.10.1


lanzagar:
Is this related to https://github.com/conda-forge/pyqt-feedstock/pull/25?
Anyway, I don't think we will want to pin the qt version to 5.10.1 in general. E.g. I am using 5.11.1 and it is working fine.
If specific system configurations / environments have issues, I think downgrading/upgrading a package there is more appropriate.
However, if some version of a package causes problems for everyone, we can add a not-equal version restriction.

PR #3123, opened 2018-07-06
Fixed workflow basedir when saving

PrimozGodec:
##### Issue

`basedir` points to the parent directory of the basedir.

##### Description of changes
`basedir` changed to actually point to the base directory. 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
@ales-erjavec, you changed this line in ff96b96 and now @PrimozGodec wants to change it back. Do you perhaps remember what did you want to achieve with this change?

ales-erjavec:
>Do you perhaps remember what did you want to achieve with this change?

No. But it was an mistake.

Issue #2937, opened 2018-03-03
[IDEA] Include errors/warnings in widget help


astaric:
While Errors and Warnings should be as descriptive as possible, they usually describe the problem, not the way to solve it.

One way to address this would be to list the error/warning messages in documentation along with the suggested solutions.

ajdapretnar:
Where would you put this file? I can prepare a template with some examples, just let me know what is the proper location for this.

astaric:
I would include it in widget's help file. Not sure where though. Maybe a separate section?

janezd:
@ajdapretnar, let us consider this a guideline and remember it when writing or improving documentation. But guidelines don't belong to issues, so let's close it.

PR #3128, opened 2018-07-12
[RFC][ENH] Extend documentation with error and warning descriptions

ajdapretnar:
##### Issue

Implements #2937.

##### Description of changes
Extend existing documentation with error, warning and information description.

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation

### DIFF
--- a/doc/visual-programming/source/widgets/evaluation/testandscore.rst
+++ b/doc/visual-programming/source/widgets/evaluation/testandscore.rst
@@ -11,7 +11,6 @@ Inputs
     Learner
         learning algorithm(s)
 
-
 Outputs
     Evaluation Results
         results of testing classification algorithms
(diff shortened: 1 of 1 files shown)


doc/visual-programming/source/widgets/evaluation/testandscore.rst
```diff
@@ -101,3 +100,34 @@ want to know only the sex and status of the survived and omit the age.
...
+*Target variable has no values* - Target variable in the data has only missing values meaning the model can't be constructed. Use :doc:`Select Columns <../data/selectcolumns>` to select a different target variable.
+*Target variable has only one value* - Target variable in the data has a constant value meaning the model can't be constructed. Use :doc:`Select Columns <../data/selectcolumns>` to select a different target variable.
+
+Warnings
+++++++++
+
+"Instances with unknown target values were removed from data." - If a data instance has missing values, it was not considered in the computation. To change this, use :doc:`Impute <../data/impute>` to impute the the missing values.
+"Missing separate test data input." - To Test on test data, use Test Data input. Double-click on the connection between the data and Test & Score and rewire the connection to Test Data.
+"Some scores could not be computed." - The number of folds exceeds the number of instances in the smallest class. Decrease the number of folds to be the same or lower than the number of smallest class or use Leave one out option.
```
PrimozGodec: I would actually rename that warning to something like `AUC could not be computed`, as I know this issue appears only because of AUC.

ajdapretnar: Yes, I agree.

PrimozGodec: I just opened the PR with this warning renamed. https://github.com/biolab/orange3/pull/3130

lanzagar: I would not rename _this_ warning (we can have a specific warning for AUC with too many folds, or solve that in another way).
That means also that the documentation of this warning should be changed - it does not imply that "the number of folds exceeds..." but that something (anything!) went wrong with some scorer. Currently, however, the AUC one is a common example that is the most frequent cause of it. That can be mentioned in the documentation, or even better, handled explicitly (different warning / code checking).

ajdapretnar: Yes, I agree with having a separate warning. This is also necessary, because the documentation is supposed to provide a guide to solving the issues. We can't just say 'something went wrong'. The idea is to guide the user wherever possible.

markotoplak:
I think we should not accept this. We are extending documentation with something that, for most of the warnings, does not really add any information. Perhaps we could just rephrase the questionable ones.

I would say a warning description in the documentation is useless unless there is a directy link to it when the warning appears.

ajdapretnar:
I don't think you can always neatly explain what the user needs to do. For example, a target variable is missing doesn't tell the users they need to use Select Columns and put something to Target variable box.
Unfortunately, also, most people don't read what the error tells them. Nor they do read the documentation, so both attempts are fairly futile.

ajdapretnar:
As we haven't agreed on extending docs with warning descriptions, I think I should close this. If at any point we decide we should go ahead with this, I'll reopen.

astaric:
@markotoplak, it does not have to be documentation, but IMHO, all errors that include "Target Variable" could benefit from some clarification as they are quite technical.

1. *Train data input requires a target variable*
    and
2. *Too many target variables*
    Would benefit a lot by having "Use :doc:`Select Columns <../data/selectcolumns>` to select a single target variable." written somewhere, but I do not see it as a part of the error message, as it would make it too long.

3. *Test and train datasets have different target variables*
   This one is also confusing and is usually a result of not using make in some place in the code, using python script or feature constructor or create a class widget and then using the result in combination with some data read from a file.

Even if we do not list all errors and warnings in the documentation, having these three listed in a "Common errors and warnings" section, would IMHO be an improvement.

PR #3130, opened 2018-07-12
[RFC] [FIX] Warning renamed in Test and Score widget

PrimozGodec:
##### Issue

The warning tells just that one score is missing. 
As I know this warning will be triggered just by AUC.

##### Description of changes
The warning message was changed to tell that the AUC score is missing. 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
The issue of uninformative messages is partially addressed in #3128. However, having informative messages is much better than rerouting to docs. If the score issue runs deeper than AUC, I agree we should have two errors, one for the specific case of low number of instances, and one for other unknown errors.

PR #3134, opened 2018-07-13
[ENH] Matplotlib output for Scatter plot

markotoplak:
##### Issue
Our outputs are ugly and non-configurable. Outputting python code could solve output flexibility problem for people with basic programming knowledge. 

##### Description of changes
The changes here just enable the user to save graphs as Python code. Currently a pyplot plot is drawn directly.

The code gets the data for drawing from the pyqtgraph's graph.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/tests/test_evaluation_scoring.py
+++ b/Orange/tests/test_evaluation_scoring.py
@@ -1,10 +1,7 @@
 # Test methods with long descriptive names can omit docstrings
 # pylint: disable=missing-docstring
 
-import sys
-import contextlib
 import unittest
-from unittest.mock import Mock
 import numpy as np
 
 from Orange.data import DiscreteVariable, ContinuousVariable, Domain
@@ -337,23 +334,9 @@ class TestComputeCD(unittest.TestCase):
         cd = scoring.compute_CD(avranks, 30, test="bonferroni-dunn")
         np.testing.assert_almost_equal(cd, 0.798)
 
-        @contextlib.contextmanager
-        def mock_module(name):
-            if not name in sys.modules:
-                try:
-                    sys.modules[name] = Mock()
-                    yield
-                finally:
-                    del sys.modules[name]
-            else:
-                yield
-
         # Do what you will, just don't crash
-        with mock_module("matplotlib"), \
-                mock_module("matplotlib.pyplot"), \
-                mock_module("matplotlib.backends.backend_agg"):
-            scoring.graph_ranks(avranks, "abcd", cd)
-            scoring.graph_ranks(avranks, "abcd", cd, cdmethod=0)
+        scoring.graph_ranks(avranks, "abcd", cd)
+        scoring.graph_ranks(avranks, "abcd", cd, cdmethod=0)
 
 
 class TestLogLoss(unittest.TestCase):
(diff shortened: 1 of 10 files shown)


BlazZupan:
This is a good first step, but it needs some polishing. In the case when dots where colored I am missing a legend. Also, the code includes two calls to `plt.scatter`, where the first one unnecessary plots a black edge around the point, while removing it I get exactly Orange's plot, as intended:

In any case, the output is much better than current PDF output. It would be great to work on this prototype and complete it to PR, and then ask Jure to help in polishing it. In summary:

- add the legend (this may require to split `plt.scatter` to several calls for discrete legend)
- remove black edges
- minimize the output by limiting precision for `sizes` and `edgecolors` (use, say, np.float16)

BlazZupan:
Also, could you please add "Portable Document Format (from Matplotlib) (*.pdf)" to the list of supported outputs. The order of the outputs would then be:

1. Portable Document Format (from Matplotlib) (*.pdf)
2. png
3. Scalable Vector Format
4. Python code (with Matplotlib)
5. Portable Document Format (*.pdf)

BlazZupan:
Works well! When exporting matplotlib pdf's on Mac, OS X asks to modify the extension and the whole process gets confusing. Could you please remove .matplotlib from the extension and simply use pdf? Also, if possible, when saving to .py, reduce the precision edgecolors and facecolors to reduce the size of the file. When these are saved for categorical values, perhaps the length of this file could be reduced even more if edgecolors and facecolors would be created programmatically, avoiding creating arrays with huge chunks of arrays with same numbers.

markotoplak:
I added "compression" of colors to shorten the generated code. Generating code with different numpy versions is a pain - therefore I avoided numpy's repr. This exporter is now only enabled for the scatterplot as the only ".pdf" exporter.

Missing functionality:
- a legend

A test for the CD diagram fail now. It only fails if I run all the tests, not just that one. Perhaps some file imports matplotlib before and that confuses it. @lanzagar ? If we are making matplotlib a dependency, that is easy to solve: just stop using mock in that test.

Warning: I added matplotlib to requirements. Before merging we would need to take care that package build files are updated.

Looking back, I think this PR should be reimplemented. Reading from pyqtgraph structures has many hidden details, and the code generator is then not aware of what the content really means. implementing equivalent functionality in the corresponding widgets (which also know the meaning of what they draw) should be better. I did not go that way because:

1. It was easier to integrate my current approach with how graphs are saved (FileFormats taking a scene).
2. Because scatterplot code is so dirty that was hard to find what to draw in there. 

A mistake, probably!

BlazZupan:
@markotoplak, on "I think this PR should be reimplemented. Reading from pyqtgraph structures has many hidden details, and the code generator is then not aware of what the content really means." -- I agree, it would probably be easier if the implementation would be widget-specific, but at this stage, what do you suggest. I guess we have three options:

1. merge these changes reimplement
2. close with no merge, reimplement

I wish this implementation would set an example how to implement this functionality in other widgets, where reading the pyqtgraph structures would be even more problematic or not possible since other means of plotting where used. Example include hierarchical clustering, box plot, and silhouette plot.

markotoplak:
If you need it now, merge. Before that, someone need to take care of
matplolib dependecy and a the CD test (I can only do it after 12.8.), which
now, that matplotlib is actually imported, behaves strengely. Then we will
at least have some idea how people use it.

markotoplak:
@ales-erjavec, could you please check if I added the matplotlib dependency correctly?

PR #3135, opened 2018-07-17
OWPreprocess: Make widget inheritable

VesnaT:
Fixes:
https://github.com/biolab/orange3-single-cell/issues/243

Make the widget inheritable for single-cell add-on purposes.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpreprocess.py
+++ b/Orange/widgets/data/owpreprocess.py
@@ -212,14 +212,14 @@ class ContinuizeEditor(BaseEditor):
             rb = QRadioButton(
                 text=text,
                 checked=self.__treatment == treatment)
-            group.addButton(rb, _enum_to_index(ContinuizeEditor._Type,
-                                               treatment))
+            group.addButton(rb, enum_to_index(ContinuizeEditor._Type,
+                                              treatment))
             self.layout().addWidget(rb)
 
         self.setSizePolicy(QSizePolicy.Minimum, QSizePolicy.Fixed)
 
     def setTreatment(self, treatment):
-        buttonid = _enum_to_index(ContinuizeEditor._Type, treatment)
+        buttonid = enum_to_index(ContinuizeEditor._Type, treatment)
         b = self.__group.button(buttonid)
         if b is not None:
             b.setChecked(True)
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/utils/preprocess.py
```diff
@@ -412,7 +413,7 @@ def createWidgetFor(self, index):
         Create a QWidget instance for the index (:class:`QModelIndex`)
         """
         definition = index.data(DescriptionRole)
-        widget = definition.viewclass()
+        widget = definition.viewclass(self._parent)
```
markotoplak: I checked how does the change in Controller influence my preprocessors in orange-spectroscopy. It is OK for most of them, but some assume that parent is a keyword argument and crash with your changes.

I though about changing mine, but then I saw that even QWidget uses parent as a keyword argument (see http://pyqt.sourceforge.net/Docs/PyQt4/qwidget.html). Therefore I propose changing this line to `widget = definition.viewclass(parent=self._parent)`

http://pyqt.sourceforge.net/Docs/PyQt4/qwidget.html

VesnaT: Done.

Orange/widgets/data/utils/preprocess.py
```diff
@@ -412,7 +413,7 @@ def createWidgetFor(self, index):
         Create a QWidget instance for the index (:class:`QModelIndex`)
         """
         definition = index.data(DescriptionRole)
-        widget = definition.viewclass()
+        widget = definition.viewclass(parent=self._parent)
```
ales-erjavec: Why? The `self._parent` is not even the correct type (it can be any QObject, but QWidgets must be parented to other QWidgets), and has no particular specified relation to the parent-child hierarchy that the widgets are/will be in.

VesnaT: I need a reference to OWWidget. I should probably have used some other keyword argument.

markotoplak: If you use any other keyword argument, you are, I think, going to break the existing preprocessors.

I do not know what is a good solution here. You could perhaps use `widget.parent_widget = self._parent` after init. At least that is what I did when I subclassed the Controller.

Issue #3079, opened 2018-06-19
Test & Score crashes on housing dataset using NN


pruales:
##### Orange version

3.13.0

##### Expected behavior

With the flow shown below, the test & score widget is expected to score the model on the training dataset. It works as expected on the Iris dataset.

##### Actual behavior
On the housing dataset, if you test and score on the training data the GUI crashes and you get a Bus error shown below.

##### Steps to reproduce the behavior
Use default parameters on the neural network and feed the raw housing data. I haven't tested on other datasets.

##### Additional info (worksheets, data, screenshots, ...)

``` /miniconda3/bin/pythonw: line 3: 90460 Bus error: 10         /miniconda3/python.app/Contents/MacOS/python "$@"```

<img width="1028" alt="screen shot 2018-06-19 at 6 23 57 pm" src="https://user-images.githubusercontent.com/21148863/41629244-1d87a8d4-73ee-11e8-855b-d2fd92f2f492.png">

pruales:
I installed with Anaconda and ran the GUI from the command line. It seems to be an issue with this type of usage. I downloaded the MacOS bundle and it worked as expected

Issue #3050, opened 2018-06-04
K-means: crahses on large data sets


ajdapretnar:
##### Orange version

3.14.dev

##### Expected behavior
K-means works for large data.

##### Actual behavior
On ~54000 rows, k-Means crashes with segfault.

##### Steps to reproduce the behavior
File (https://www.kaggle.com/zynicide/wine-reviews --> winemag-data-130k-v2.csv) - Select Rows (country is US) - k-Means.

thocevar:
This data set has several discrete variables with a large number of values. K-means performs continuization, which inflates the number of variables to ~3000 and you end up with a table of size 1.2 GB.

ajdapretnar:
Shoot. Is there any sensible way around this? 🤔

janezd:
Not really. K-means can't cope with discrete variables, AFAIK.

PR #3138, opened 2018-07-18
[FIX] utils/concurrent: Switch default thread pool

ales-erjavec:
##### Issue

Fixes gh-3079, gh-3050

##### Description of changes

The default stack size of QThreadPool created threads is insufficient
to call OpenBLAS in common builds (see numpy issue 11551).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


thocevar:
ThreadExecutor doesn't seem to fully utilize the CPU. Execution time is the same with 1 or 4 max_workers on a 4-core CPU (Windows 10). Are the threads assigned to the same core?

ales-erjavec:
Where? What task?

PR #3140, opened 2018-07-19
[FIX][RFC] KMeans upgrade sparse support

### LINKED ISSUE #2913: K-means: work on sparse
https://github.com/biolab/orange3/blob/1eb9b69d37e35b159bd6892a606606803c537da1/Orange/clustering/kmeans.py#L29

```len(X)``` should be ```X.shape[0]``` beacause it raise error ```sparse matrix length is ambiguous; use getnnz() or shape[0]``` when use tf-idf sparse matrix with Kmeans

Error flow
error

robertcv:
##### Issue
Fixes #2913 
Additionally, I noticed that you can not predict sparse data.
##### Description of changes
Replaced `len(X)` with `X.shape[0]`.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/clustering/kmeans.py
+++ b/Orange/clustering/kmeans.py
@@ -27,7 +27,7 @@ class KMeans(SklProjector):
         proj.silhouette = np.nan
         try:
             if self._compute_silhouette and 2 <= proj.n_clusters < X.shape[0]:
-                if len(X) <= SILHOUETTE_MAX_SAMPLES:
+                if X.shape[0] <= SILHOUETTE_MAX_SAMPLES:
                     proj.silhouette_samples = \
                         silhouette_samples(X, proj.labels_)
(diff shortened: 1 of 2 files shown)


ajdapretnar:
Would you mind adding a nice informative error when the user connects pure Corpus into k-Means (without bag of words)? Now the error is useless. It should say something like: "Data has no attributes." or "Data is missing features."

robertcv:
@ajdapretnar I can add it but this pull request is probably not the right place for it.

ajdapretnar:
Ok, let's have it in a different PR then. As far as I am concerned, this PR works.

ajdapretnar:
Would you mind making pylint happy, so the tests will pass? :) 🙈 

> third party import "import numpy as np" should be placed before "import Orange"

PR #3145, opened 2018-07-20
Introduce sizeHints for Scatter Plot

ajdapretnar:
##### Issue

Certain visualization widgets open with wrong initial size.

##### Description of changes
Introduce sizeHint that follows 16:9 ratio.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -2,7 +2,7 @@ from itertools import chain
 
 import numpy as np
 
-from AnyQt.QtCore import Qt, QTimer
+from AnyQt.QtCore import Qt, QTimer, QSize
 from AnyQt.QtGui import QPen, QPalette
 from AnyQt.QtWidgets import QApplication
 
@@ -217,6 +217,10 @@ class OWScatterPlot(OWWidget):
         for w in [MatplotlibFormat, MatplotlibPDFFormat]:
             for ext in w.EXTENSIONS:
                 self.graph_writers[ext] = w
+
+
+    def sizeHint(self):
+        return QSize(1132, 708)
 
     def keyPressEvent(self, event):
         super().keyPressEvent(event)


ajdapretnar:
Honestly, I don't like this, but here's how a 16:9 ratio would look like.

astaric:
why 16:9?

markotoplak:
> Why 16:9. 

Some designer said so  - I asked him why and he said it is a nice ratio. And supposedly it would be beneficial if our widgets follow a single ratio. (I think both are... Whoops, let's just say I am not that convinced)

astaric:
I'd go with 1:1.6 as it is the same ratio A4 paper uses :) and that would probably be for the visualization itself, not the whole widget.

markotoplak:
The designer was open about the ratio... He said we could try different options.

ajdapretnar:
I don't think this will make it into 3.16. We didn't decide on the ratio. I can modify to 1:1.6, if you want.

lanzagar:
Why wouldn't it make it into 3.16?
The code changes are minimal, and the current state is **_terrible_**.

I agree we have not decided on the appropriate size/ratio for a all of Orange or even just visualizations, but this should be renamed to "Fix Scatter Plot size" and merged before 3.16. I don't really care what the size is, anything will be better than the current default :)

So please rebase, and you can change the aspect ration while at it to whatever you find best.
The general topic will remain open after this, we just need a quickfix until we find the ultimate answer.

ajdapretnar:
After discussing with @irgolic, we decided to got with @astaric's suggestion. Now Scatter Plot uses 1:1.6 ratio, but horizontally for SP, since the other way around simply doesn't look good. Can be merged before 3.16.

kernc:
Btw, like 16:9, it's called [*16:10*](https://en.wikipedia.org/wiki/16:10) aspect ratio.

Issue #3144, opened 2018-07-20
Hiding and showing of control area should keep the aspect ratio of the main (plotting) area


BlazZupan:
##### Expected behavior

Hiding the control area should resize the window, keeping the right edge of the window where it is now and shrinking the window from the left, such that the main part of the widget's window (say, the visualization) keep the size and the position.

##### Actual behavior

Hiding and showing of control area is currently triggered by clicking the splitter in the widget's window or by pressing Cmd-Shift-D. The action keeps the size of the window, and hence changes the aspect ration of any visualization in the main widgets area:

![screen shot 2018-07-20 at 11 15 22](https://user-images.githubusercontent.com/726604/42994515-9086bbc6-8c0e-11e8-8423-48d4ec44d352.png)

PR #3141, opened 2018-07-19
Let the main area shrink further when control area is hidden.

thocevar:
##### Issue
Fixes #3132.

##### Description of changes
The computed minimum size of the widget prevents the main area from shrinking further after hiding the control area. This is overridden by explicitly setting a small minimum width of the widget when control area is hidden.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #3146, opened 2018-07-20
[ENH] OWWidget: Collapse/expand the widget on control area toggle

ales-erjavec:
##### Issue

Implements gh-3144

##### Description of changes

Collapse/expand the widget on control area toggle by the control area width

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/test_widget.py
+++ b/Orange/widgets/tests/test_widget.py
@@ -194,8 +194,7 @@ class WidgetMsgTestCase(WidgetTest):
             want_control_area = True
 
         w = Widget()
-        splitter = w._OWWidget__splitter  # type: OWWidget._Splitter
-        splitter.setControlAreaVisible(False)
+        w._OWWidget__setControlAreaVisible(False)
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I made a mess. I didn't realize merging #3141 also touches widget.py. Should I revert and keep this one instead?

ales-erjavec:
No. I can rebase.

thocevar:
There is a visible blink/movement of items in the main area when the widget moves and then shrinks. Is it perhaps possible to do this in a single update?

ales-erjavec:
> Is it perhaps possible to do this in a single update?
Probably not. This seems platform dependent.

ajdapretnar:
I like this. I don't experience any blinking on OSX (but this is after the last commit). There is an occasional update of the plot (or at least it looks like the graph is replotted), but it doesn't bother. Not sure what would happen on bigger data set, but perhaps it is not replotting but just a small glitch. :)

Issue #3136, opened 2018-07-17
Save Data enhancement


lanzagar:
It is now possible to use compression when saving to any file format (e.g. pickle + gzip), but it is not evident how to do that. Currently one has to click Save As, then choose Files of type: Pickled Python object file (...) and save with a filename test.pkl for no compression or test.pkl.gz or test.pkl.xz to use appropriate compression.

Instead, we should make the choices clearer in the widget itself.
Below is an example of how the widget could look like. Comments welcome.

![save data](https://user-images.githubusercontent.com/4579717/42817464-7b7bcbdc-89ce-11e8-9189-25d974bbda49.png)

Save As would open the same dialog as now, but would be limited to files of the selected type/compression.

_File type_ options:
- Tab-delimited (.tab)
- Comma-separated values (.csv)
- Pickle (.pkl)
- [future] Excel (.xlsx) and/or Open Document Spreadsheet (.ods)?

_Compression_ options:
- gzip (.gz)
- bzip2 (.bz2)
- lzma (.xz)

[**RFC**] Default options could either be:
- tab or csv with no compression: better for smaller data
- pickle with compression: better for big(ger) data
- choosing one of the above based on data size?

lanzagar:
Done in #3147

PR #3147, opened 2018-07-23
[ENH] Selection of format and compression in save data widget

golbog:
##### Issue
Resolves issue #3136 

##### Description of changes
Updated widget according to above issue. 
Does not include changes for biolab/orange3-network#80.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owsave.py
+++ b/Orange/widgets/data/owsave.py
@@ -46,11 +46,12 @@ class OWSave(widget.OWWidget):
     def get_writer_selected(self):
         writer = FileFormat.get_reader(self.type_ext)
         try:
-            writer.EXTENSIONS = [writer.EXTENSIONS[writer.EXTENSIONS.index(self.type_ext + self.compression_ext)]]
+            writer.EXTENSIONS = [
+                writer.EXTENSIONS[writer.EXTENSIONS.index(self.type_ext + self.compress_ext)]]
             return writer
         except ValueError:
             self.Error.not_supported_extension()
-            return
+            return None
 
     @classmethod
     def remove_extensions(cls, filename):
@@ -69,7 +70,7 @@ class OWSave(widget.OWWidget):
         self.filename = ""
         self.basename = ""
         self.type_ext = ""
-        self.compression_ext = ""
+        self.compress_ext = ""
         self.writer = None
 
         form = QFormLayout(
@@ -118,9 +119,9 @@ class OWSave(widget.OWWidget):
 
     def adjust_label(self):
         if self.filename:
-            # filename = os.path.split(self.filename)[1]
             text = "Auto save as '{}'" if self.auto_save else "Save as '{}'"
-            self.save.button.setText(text.format(self.basename + self.type_ext + self.compression_ext))
+            self.save.button.setText(
+                text.format(self.basename + self.type_ext + self.compress_ext))
 
     @Inputs.data
     def dataset(self, data):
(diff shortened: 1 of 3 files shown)


Orange/widgets/data/owsave.py
```diff
@@ -22,58 +39,108 @@ class Inputs:
...
     def get_writers(cls, sparse):
         return [f for f in FileFormat.formats
                 if getattr(f, 'write_file', None) and getattr(f, "EXTENSIONS", None)
                 and (not sparse or getattr(f, 'SUPPORT_SPARSE_DATA', False))]
 
+    def get_writer_selected(self):
+        type = FILE_TYPES[self.filetype]
+        compression = COMPRESSIONS[self.compression] if self.compress else ''
+        writer = FileFormat.get_reader(type)
+        writer.EXTENSIONS = [writer.EXTENSIONS[writer.EXTENSIONS.index(type + compression)]]
```
lanzagar: You are looking up an index of something in the list and then indexing that list with the found index... you will probably be getting that "something" back most of the time ;)
Writing this directly `writer.EXTENSIONS = [type + compression]` looks a bit dangerous since you are overwriting the list, but it should always be one of the available options anyway. Probably it would not be bad to have an assert before this to make sure that the chosen extension was really supported (more for the benefit of someone reading the code).

Orange/widgets/data/owsave.py
```diff
@@ -22,58 +39,108 @@ class Inputs:
...
+            box, self, "filetype",
+            callback=None,
+            items=FILE_TYPES_NAMES,
+            sendSelectedValue=True,
+        )
+        form.addRow("File type", self.controls.filetype, )
+
+        gui.comboBox(
+            box, self, "compression",
+            callback=None,
```
lanzagar: Instead of `callback=None` for both of these combo boxes, you could call a function that fixes self.filename.
E.g. I save a tab file with gzip compression to `test.tab.gz`, then when changing to csv, the stored filename could be updated to `test.csv.gz` and similarly for changes to compression.
Currently, after changing the type and clicking Save again, the file is stored with the previous type since that is what the stored filename reflects...

Orange/widgets/data/owsave.py
```diff
@@ -22,58 +38,126 @@ class Inputs:
...
     auto_save = Setting(False)
+    filetype = Setting(FILE_TYPES[0][0])
+    compression = Setting(COMPRESSIONS[0][0])
+    compress = Setting(False)
+
+    def get_writer_selected(self):
+        writer = FileFormat.get_reader(self.type_ext)
+        try:
+            writer.EXTENSIONS = [
+                writer.EXTENSIONS[writer.EXTENSIONS.index(self.type_ext + self.compress_ext)]]
```
lanzagar: You still have the problem mentioned in https://github.com/biolab/orange3/pull/3147#discussion_r206903890

Instead of the try/except, you should just check
```py
ext = self.type_ext + self.compress_ext
if ext not in writer.EXTENSIONS:
    self.Error.not_supported_extension()
    return None
writer.EXTENSIONS = [ext]
return writer
```

BTW, just a minor suggestion, but a better name for `not_supported_extension` would be `unsupported_extension`.

Orange/widgets/data/owsave.py
```diff
@@ -1,12 +1,28 @@
...
 from Orange.widgets import gui, widget
 from Orange.widgets.settings import Setting
-from Orange.data.io import FileFormat
 from Orange.widgets.utils import filedialogs
 from Orange.widgets.widget import Input
 
+FILE_TYPES = [
+    ("Tab-delimited (.tab)", ".tab", False),
+    ("Comma-seperated values (.csv)", ".csv", False),
+    ("Pickle (.pkl)", ".pkl", True),
```
lanzagar: Again, these are hard coded constants, which could become incorrect after changes elsewhere.
How did you know where to write False and where True? If you checked the attribute SUPPORT_SPARSE_DATA in reader classes it is better to import and reference that here directly...
Maybe even construct the whole triple directly from the class:

```py
FILE_TYPES = [
    ("{} ({})".format(w.DESCRIPTION, w.EXTENSIONS[0]),
     w.EXTENSIONS[0],
     w.SUPPORT_SPARSE_DATA)
    for w in (TabReader, CSVReader, PickleReader)
]
```

Please test the above code and fix if necessary. It was written here as a suggestion without trying it out.

Also, I think the description for PickledReader could be changed to "Pickled Orange data".

PR #3148, opened 2018-07-23
[FIX] WebView fixes

thocevar:
##### Issue
1. _evalJS of WebEngine ignores the returned value since 3b11a9090b3111d0052faa8617e015de3ccdc1f7.
2. Report_plot doesn't work for webview (although it should).

##### Description of changes
1. Store ids together with results in IdStore for scheduled JS executions.
2. Unsuccessful search for `<svg>` tag results in ValueError (and not IndexError).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/webview.py
+++ b/Orange/widgets/utils/webview.py
@@ -475,23 +475,23 @@ elif HAVE_WEBENGINE:
         def __init__(self):
             self.id = 0
             self.lock = threading.Lock()
-            self.ids = set()
+            self.ids = dict()
 
         def create(self):
             with self.lock:
                 self.id += 1
                 return self.id
 
-        def store(self, id):
+        def store(self, id, value):
             with self.lock:
-                self.ids.add(id)
+                self.ids[id] = value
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/utils/webview.py
```diff
@@ -475,23 +475,23 @@ class IdStore:
...
-        def store(self, id):
+        def store(self, id, value):
             with self.lock:
-                self.ids.add(id)
+                self.ids[id] = value
 
         def __contains__(self, id):
             return id in self.ids
 
         def remove(self, id):
```
astaric: Maybe this could be renamed to pop?

Orange/widgets/utils/webview.py
```diff
@@ -556,9 +556,9 @@ def _evalJS(self, code):
             if sip.isdeleted(self):
                 return
```
astaric: Lint suggests that you return None here.

(Either all return statements in a function should return an expression, or none of them should.)

PR #3151, opened 2018-07-24
Usage Statistics Tracking

irgolic:
##### Issue
Anonymous usage statistics are used to streamline the UX design process. Statistics are to be collected anonymously, only with explicit user permission.

##### Description of changes
- [x] Permission request
- [x] Usage tracking
    - [X] Tracks the addition of widgets
    - [x] Stores statistics persistently
- [x] Communication with server

##### Includes
- [X] Code changes
- [ ] Tests
- [X] Documentation


requirements.txt
```diff
@@ -1,3 +1,4 @@
+requests
```
lanzagar: This should go into requirements-core

lanzagar:
We could not find out with @acopar where the results of usage statistics can be found or if even they actually get stored somewhere. Either there is a problem somewhere or we just didn't know where to look.
In any case, it would be good if this functionality came with some documentation (can be a wiki page here on github), so that others know what is happening.

irgolic:
A wiki page has been added clarifying the nature of statistics tracking: https://github.com/biolab/orange3/wiki/Anonymous-Usage-Statistics-Tracking

PR #3155, opened 2018-07-25
[FIX][RFC] OWHeatMap: remove empty clusters from visualization

### LINKED ISSUE #2907: Heat Map doesn't output a specific selection
##### Orange version

3.10.dev

##### Expected behavior
Heat Map outputs data subset when selected in the plot.

##### Actual behavior
Heat Map fails to output only some instances.

File - Heat Map - Data Table
Cluster by k-means. Select bottom part of yellow instances.

Data set available for testing. Ask in comment.

screen shot 2018-02-14 at 11 54 20

robertcv:
##### Issue
Fixes #2907 
##### Description of changes
Kmeans in the OWHeatMap can produce clusters with no elements. Before those clusters were shown in the visualization and could be selected. Because there was no data in those clusters if they were selected no output was given and it seemed like the widget's selection had a bug. This fix removes those empty clusters from the visualization and they can no longer be selected. Additionally, a warning is shown.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owheatmap.py
+++ b/Orange/widgets/visualize/owheatmap.py
@@ -444,6 +444,9 @@ class OWHeatMap(widget.OWWidget):
             "Not enough instances for k-means merging")
         not_enough_memory = Msg("Not enough memory to show this data")
 
+    class Warning(widget.OWWidget.Warning):
+        empty_clusters = Msg("Empty clusters were removed")
+
     def __init__(self):
         super().__init__()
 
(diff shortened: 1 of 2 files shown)


lanzagar:
Seems ok to me. Just waiting on the appveyor tests...
Are you not sure about something else or do you consider this ready to be merged? In the latter case, you don't need the [RFC] tag...

robertcv:
Ok, I consider it ready.

Issue #3057, opened 2018-06-07
ROC Analysis widget: show thresholds


markotoplak:
The widget does not provide any way to get the threshold that was used for a point on curve. I would suggests a tooltip.

Threshold is important to understand results of cost optimization.

astaric:
One way to implement tooltips with ScatterPlotItems: https://groups.google.com/forum/#!topic/pyqtgraph/3osqs25pEIc

ales-erjavec:
This was done in gh-3172

Issue #1316, opened 2016-06-07
When selecting multiple rows in Data Table, Image Viewer crashes


rok-povsic:
##### Orange version
3.3.6.dev0+ab40d8

##### OS
Ubuntu 16.04 64bit, kernel 4.4.8-040408

##### Expected behavior
Displaying multiple images in Image Viewer widget.

##### Actual behavior
Error window is displayed with an error. No images are displayed.

##### Steps to reproduce the behavior
Schema:
File -> Data Table -> Image Viewer

Data Table contains one column with absolute paths to images on local disk.

##### Additional info (worksheets, data, screenshots, ...)

https://snag.gy/8UeklE.jpg
https://snag.gy/UWSRPA.jpg

Error log:
```text
--------------------------------------------------------------------------------
RuntimeError                                  Traceback (most recent call last):
  File "/home/rok/faks/ozip/orange3/Orange/widgets/data/owimageviewer.py", line 488, in set_pixmap
    thumb.setPixmap(pixmap)
  File "/home/rok/faks/ozip/orange3/Orange/widgets/data/owimageviewer.py", line 150, in setPixmap
    self.pixmapWidget.setPixmap(pixmap)
  File "/home/rok/faks/ozip/orange3/Orange/widgets/data/owimageviewer.py", line 53, in setPixmap
    self.updateGeometry()
RuntimeError: wrapped C/C++ object of type GraphicsPixmapWidget has been deleted
--------------------------------------------------------------------------------
RuntimeError                                  Traceback (most recent call last):
  File "/home/rok/faks/ozip/orange3/Orange/widgets/data/owimageviewer.py", line 488, in set_pixmap
    thumb.setPixmap(pixmap)
  File "/home/rok/faks/ozip/orange3/Orange/widgets/data/owimageviewer.py", line 150, in setPixmap
    self.pixmapWidget.setPixmap(pixmap)
  File "/home/rok/faks/ozip/orange3/Orange/widgets/data/owimageviewer.py", line 53, in setPixmap
    self.updateGeometry()
RuntimeError: wrapped C/C++ object of type GraphicsPixmapWidget has been deleted
--------------------------------------------------------------------------------
RuntimeError                                  Traceback (most recent call last):
  File "/home/rok/faks/ozip/orange3/Orange/widgets/data/owimageviewer.py", line 488, in set_pixmap
    thumb.setPixmap(pixmap)
[...]

PR #3172, opened 2018-07-30
[ENH] ROC analysis: show thresholds

matejklemen:
##### Issue
Implements #3057.

##### Description of changes
When mouse is put over a point on ROC curve, the threshold that was used for calculation now appears. If threshold value is _nan_, the value does not get shown. Also, the displaying of thresholds is currently not implemented when showing individual curves.

My questions:
- currently thresholds are displayed when mouse goes 'perfectly' over a point - should this be changed so that they get displayed if mouse is in some very small radius of the point?
- what kind of a test do I write for this functionality (i. e. mouse events) - if any?

General comments are also welcome.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owrocanalysis.py
+++ b/Orange/widgets/evaluate/owrocanalysis.py
@@ -10,8 +10,8 @@ from collections import namedtuple, deque, OrderedDict
 import numpy
 import sklearn.metrics as skl_metrics
 
-from AnyQt.QtWidgets import QListView, QLabel, QGridLayout, QFrame, QAction
-from AnyQt.QtGui import QColor, QPen, QBrush, QPainter, QPalette, QFont
+from AnyQt.QtWidgets import QListView, QLabel, QGridLayout, QFrame, QAction, QToolTip
+from AnyQt.QtGui import QColor, QPen, QBrush, QPainter, QPalette, QFont, QCursor
 from AnyQt.QtCore import Qt
 import pyqtgraph as pg
 
@@ -396,6 +396,7 @@ class OWROCAnalysis(widget.OWWidget):
 
         self.plotview = pg.GraphicsView(background="w")
         self.plotview.setFrameStyle(QFrame.StyledPanel)
+        self.plotview.scene().sigMouseMoved.connect(self._on_mouse_moved)
 
         self.plot = pg.PlotItem(enableMenu=False)
         self.plot.setMouseEnabled(False, False)
@@ -528,7 +529,6 @@ class OWROCAnalysis(widget.OWWidget):
                 graphics = curve.merge()
                 curve = graphics.curve
                 self.plot.addItem(graphics.curve_item)
-                graphics.curve_item.scene().sigMouseMoved.connect(self._on_mouse_moved)
 
                 if self.display_convex_curve:
                     self.plot.addItem(graphics.hull_item)
@@ -556,7 +556,6 @@ class OWROCAnalysis(widget.OWWidget):
 
                 self.plot.addItem(graphics.curve_item)
                 self.plot.addItem(graphics.confint_item)
-                graphics.curve_item.scene().sigMouseMoved.connect(self._on_mouse_moved)
 
             hull_curves = [curve.avg_vertical.hull for curve in selected]
 
@@ -565,7 +564,6 @@ class OWROCAnalysis(widget.OWWidget):
                 graphics = curve.avg_threshold()
                 self.plot.addItem(graphics.curve_item)
                 self.plot.addItem(graphics.confint_item)
-                graphics.curve_item.scene().sigMouseMoved.connect(self._on_mouse_moved)
 
             hull_curves = [curve.avg_threshold.hull for curve in selected]
 
@@ -607,7 +605,7 @@ class OWROCAnalysis(widget.OWWidget):
         self.warning(warning)
 
     def _on_mouse_moved(self, pos):
-        curves = self.plot.curves
+        curves = [crv for crv in self.plot.curves if isinstance(crv, pg.PlotCurveItem)]
         for i in self.selected_classifiers:
             sp = curves[i].childItems()[0]
             act_pos = sp.mapFromScene(pos)
@@ -619,9 +617,9 @@ class OWROCAnalysis(widget.OWWidget):
                 curve_data = curve_data.merged
             elif self.roc_averaging == OWROCAnalysis.Vertical:
                 curve_data = curve_data.avg_vertical
-            elif OWROCAnalysis.Threshold:
+            elif self.roc_averaging == OWROCAnalysis.Threshold:
                 curve_data = curve_data.avg_threshold
-            elif OWROCAnalysis.NoAveraging:
+            elif self.roc_averaging == OWROCAnalysis.NoAveraging:
                 # currently not implemented
                 return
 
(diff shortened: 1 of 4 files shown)


Orange/widgets/evaluate/owrocanalysis.py
```diff
@@ -561,6 +565,7 @@ def _setup_plot(self):
                 graphics = curve.avg_threshold()
                 self.plot.addItem(graphics.curve_item)
                 self.plot.addItem(graphics.confint_item)
+                graphics.curve_item.scene().sigMouseMoved.connect(self._on_mouse_moved)
```
ales-erjavec: Connect the signal once in the `__init__` method: 
 `self.plotview.scene().sigMouseMoved.connect(self._on_mouse_moved)`

Orange/widgets/evaluate/owrocanalysis.py
```diff
@@ -601,6 +606,50 @@ def _setup_plot(self):
...
+
+                curve_pts = curve_data.points
+
+                # Find closest point on curve and display it
+                idx_closest = numpy.argmin([numpy.linalg.norm(mouse_pt - [curve_pts.fpr[idx], curve_pts.tpr[idx]])
+                                            for idx in range(len(curve_pts.thresholds))])
+
+                thresh = curve_pts.thresholds[idx_closest]
+                if not numpy.isnan(thresh):
+                    item = pg.TextItem(text="{:.3f}".format(thresh), color=(0, 0, 0))
```
ales-erjavec: Why not `QToolTip.showText`?

matejklemen: Mainly because I am new to QT and was just following a few resources online.  
I've fixed most of the things that you've mentioned and some other bugs.  
Thank you for the helpful comments!

Orange/widgets/evaluate/owrocanalysis.py
```diff
@@ -601,6 +604,52 @@ def _setup_plot(self):
                 warning = "All ROC curves are undefined"
         self.warning(warning)
 
+    def _on_mouse_moved(self, pos):
+        curves = [crv for crv in self.plot.curves if isinstance(crv, pg.PlotCurveItem)]
+        for i in range(len(self.selected_classifiers)):
+            sp = curves[i].childItems()[0]
```
ales-erjavec: This assumes implicitly that the order of the curves in the `self.plot.curves` corresponds to the `self.selected_classifiers`. This seems fragile.

You can use:
```python
curves = [(clsf_idx, self.plot_curves(self.target_index, clsf_idx), self.curve_data(self.target_index, clsf_idx))
          for clsf_idx in self.selected_classifiers]  # type: List[Tuple[int, plot_curves, ROCData]]
```
to retrieve triples of (classifier_index, plot_curves, curve_data). Use `plot_curves` similarly as `curve_data` in ROC averaging dispatch below, but use

```python
plot_curves.merge().curve_item
plot_curves.avg_vertical().curve_item
plot_curves.avg_threshold().curve_item
```
to get the `pg.PlotCurveItem` instances

matejklemen: My bad - should be fixed now. I've also fixed some other bugs:
- if mouse was hovered over a point on curve for some classifier and then the same classifier's curve would be hidden and mouse hovered over the same point as before, the cache would not reset and therefore an invalid threshold would be shown,
- if tooltip was not being shown and the cache point was "valid" (i. e. same point as previously shown), a tooltip with message "" would be created, which resulted in hiding a tooltip.

I have also encountered the following exception (just once), which I can not seem to be able to reproduce anymore:
```
--------------------------- RuntimeError Exception ----------------------------
  File "/home/matej/.local/lib/python3.5/site-packages/pyqtgraph/GraphicsScene/GraphicsScene.py", line 162, in mouseMoveEvent
    self.sendHoverEvents(ev)
  File "/home/matej/.local/lib/python3.5/site-packages/pyqtgraph/GraphicsScene/GraphicsScene.py", line 228, in sendHoverEvents
    items = self.itemsNearEvent(event, hoverable=True)
  File "/home/matej/.local/lib/python3.5/site-packages/pyqtgraph/GraphicsScene/GraphicsScene.py", line 424, in itemsNearEvent
    if hoverable and not hasattr(item, 'hoverEvent'):
RuntimeError: wrapped C/C++ object of type ScatterPlotItem has been deleted
```

When browsing through old issues, I've found out that there's been some instances of similar issues (#1316), though I can't say I completely understand your fix for this.

Orange/widgets/evaluate/owrocanalysis.py
```diff
@@ -601,6 +604,57 @@ def _setup_plot(self):
...
+                    else:
+                        QToolTip.showText(QCursor.pos(), "")
+                        self._tooltip_cache = None
+
+                curve_pts = curve.curve.points
+
+                # Find closest point on curve and display it
+                idx_closest = numpy.argmin(
+                    [numpy.linalg.norm(mouse_pt - [curve_pts.fpr[idx], curve_pts.tpr[idx]])
+                     for idx in range(len(curve_pts.thresholds))])
```
ales-erjavec: This loop could be vectorized
```python
roc_points = numpy.column_stack((curve_pts.fpr, curve_pts.tpr))
query_pt = numpy.array([[mouse_pt.x(), mouse_pt.y()]])
diff = roc_points - query_pt
idx_closest = numpy.argmin(numpy.linalg.norm(diff, axis=1))
```
(... although a slightly moot point as the `sp.pointsAt` call above already runs through all the points in a python loop)

matejklemen: True - fixed.  
I've also made another minor change: now QCursor.pos() is directly used everywhere instead of using it directly in some cases and in the other cases saving it and then using the variable (just a consistency fix).

ales-erjavec:
> currently thresholds are displayed when mouse goes 'perfectly' over a point - should this be changed so that they get displayed if mouse is in some very small radius of the point?

I think the hit radius is sufficient as it is. However it might be better to collect points from all the curves under the mouse and display a list of all thresholds with classifier names. Right now if two or more curves are close to each other it is not clear to which does the shown threshold belong to.

> what kind of a test do I write for this functionality (i. e. mouse events) - if any?

I would use [QTest.mouseMove](http://doc.qt.io/qt-5/qtest.html#mouseMove) targeting the QGraphicsView's viewport (`plotview.viewport()` in the OWROCAnalysis). Try to map one of the curve's points to scene coordinates  (`QGraphicsItem.mapToScene`) and those to the viewport (`QGraphicsWidget.mapFromScene`).

ales-erjavec:
> The QTest.mouseMove(...) seems to work ok when manually testing - it throws the cursor to (0, 0) (in plot coordinates), but does not trigger the mouse callback in the test.

Maybe [QTBUG-5232](https://bugreports.qt.io/browse/QTBUG-5232). I swear I saw this when writing https://github.com/biolab/orange3/pull/3014/files#diff-da80849ceff100f955c9ec05920ba089R138.

If I use that function and modify your test like this:

```python
        ...
        # curve data for classifier 0
        curve = self.widget.plot_curves(self.widget.target_index, 0)
        curve_merge = curve.merge()

        view = self.widget.plotview
        item = curve_merge.curve_item  # type: pg.PlotCurveItem
        xs, ys = item.getData()
        x, y = xs[len(xs) // 2], ys[len(ys) // 2]
        pos = item.mapToScene(x, y)
        pos = view.mapFromScene(pos)
        mouseMove(view.viewport(), pos)
        # example assertion
        self.assertTrue(QToolTip.isVisible())
```
then the test passes. Can you rebase on current master and use the *mouseMove* function from `test_combobox.py` (move it to e.g. *Orange/widgets/tests/utils.py* and import it from there)

matejklemen:
Thank you, that does fix the issue.  

I've stumbled upon another issue though - the points which I'm trying to move the mouse to are not being mapped correctly. For example, if I try to simulate a mouse move to (x=0.0, y=0.0) and (x=1.0, y=1.0), it produces the same threshold/tooltip.  

I've noticed that these two points get mapped to very similar positions inside the test (`QPoint(56, 1)` and `QPoint(57, 2)` in my case), while during "manual" testing they get mapped far apart and correctly.  
_My guess_ is that this is happening because the widget, created inside the test, is tiny, and therefore everything is getting mapped to pretty much the same coordinates - but I'm not sure how to fix it.

ales-erjavec:
Try adding
```
vb = self.widget.plot.getViewBox()
vb.childTransform()  # Force pyqtgraph to update transforms
```
somewhere before mapping the points.

matejklemen:
Thanks once again. :smiley: 
I've moved the `mouseMove(...)` function and added tests for tooltips - hopefully I didn't mess anything up.

Orange/widgets/evaluate/owrocanalysis.py
```diff
@@ -336,6 +336,7 @@ def __init__(self):
         self._plot_curves = {}
         self._rocch = None
         self._perf_line = None
+        self._tooltip_cache = None
```
ales-erjavec: I do not understand why the tooltip cache is necessary. Can you add a comment explaining the reasoning for its use?

ales-erjavec: But it already checks all points in ` pts = sp.pointsAt(act_pos)`

matejklemen: Right, but it also does it again later (although it is vectorized): `idx_closest = numpy.argmin(numpy.linalg.norm(diff, axis=1))`.

Considering your point, it does seem that the performance gain from tooltip cache might not be as big as I thought. Should I remove it and make the code clearer that way?

ales-erjavec:
> Should I remove it and make the code clearer that way?

It does not matter really, you can leave it.

PR #3174, opened 2018-07-31
[ENH] Add cosine distance

BlazZupan:
##### Issue
Silhouette widget supported only Euclidean and Manhattan distance.

##### Description of changes
Added a cosine distance. Could add other distances, but that would require more elaborate fix, as some of them crash with this widget. Cosine distance is needed in some of the demonstrations with current Orange, so this is a minimal fix at this stage.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
To elaborate - cosine distance is not defined for all-zero vectors. The distance returns NaNs and then the silhouette crashes.
There are several options of what to do in those cases. Of the ones I considered, I think it is best to just omit the all-zero vectors in the plot similarly to how instances with missing cluster values are omitted: there is already a mask that is applied to the data, it could be extended to include the problematic instances (any distance in its column/row is NaN). A separate warning message similar to the missing cluster one would be nice.

BlazZupan:
This PR has too many issues for one-linear, I am closing it. Need to start fresh.

PR #3176, opened 2018-07-31
Silhouette Plot: Add cosine distance

lanzagar:
Added cosine distance to Silhouette Plot.
Handle nan values in the computed dist matrix (e.g. in case of all-zero vectors for cosine) by omitting instances and showing a warning.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/distance/distance.py
+++ b/Orange/distance/distance.py
@@ -348,7 +348,8 @@ class Cosine(FittedDistance):
             dist = safe_sparse_dot(data1, data2.T)
             np.clip(dist, 0, 1, out=dist)
             if x2 is None:
-                dist.flat[::dist.shape[0] + 1] = 1.0
+                diag = np.diag_indices_from(dist)
+                dist[diag] = np.where(np.isnan(dist[diag]), np.nan, 1.0)
             return 1 - dist
 
 
--- a/Orange/widgets/visualize/owsilhouetteplot.py
+++ b/Orange/widgets/visualize/owsilhouetteplot.py
@@ -90,6 +90,7 @@ class OWSilhouettePlot(widget.OWWidget):
     class Warning(widget.OWWidget.Warning):
         missing_cluster_assignment = Msg(
             "{} instance{s} omitted (missing cluster assignment)")
+        nan_distances = Msg("{} instance{s} omitted (undefined distances)")
 
     def __init__(self):
         super().__init__()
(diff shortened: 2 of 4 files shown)


ales-erjavec:
In case of selected Cosine distance can you check the input domain to ensure it has no discrete columns?

Either show an error and stop, or show a warning and drop them from the domain before computing the distance.

The way that `Cosine` treats discrete columns means that it implicitly depends on the *variable reuse&trade;* meaning it can produce different results depending on the history and order of loaded data

For instance using 
[discrete-confound-a.tab.txt](https://github.com/biolab/orange3/files/2249029/discrete-confound-a.tab.txt)
[discrete-confound-b.tab.txt](https://github.com/biolab/orange3/files/2249030/discrete-confound-b.tab.txt)
```sh
$ cat discrete-confound-a.tab
A	B	C
d	d	d
		class
a1	b1	+
a1	b2	+
a3	b3	-
a1	b2	-
a2	b3	+
a3	b4	-
$ cat discrete-confound-b.tab
A	B	C
d	d	d
		class
a0	a1	+
a1	a2	+
a3	a3	-
a1	b2	-
a2	b3	+
a3	b4	-
```
compare
```python
import Orange
A = Orange.data.Table("discrete-confound-a.tab")
print(Orange.distance.Cosine(A).round(3))
```
which prints:
```
[[0.      nan   nan   nan   nan   nan]
 [  nan 0.    0.293 0.    0.293 0.293]
 [  nan 0.293 0.    0.293 0.    0.   ]
 [  nan 0.    0.293 0.    0.293 0.293]
 [  nan 0.293 0.    0.293 0.    0.   ]
 [  nan 0.293 0.    0.293 0.    0.   ]]
```
and
```python
import Orange
B = Orange.data.Table("discrete-confound-b.tab")
A = Orange.data.Table("discrete-confound-a.tab")
print(Orange.distance.Cosine(A).round(3))
```
that produces 
```
[[0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0.]
 [0. 0. 0. 0. 0. 0.]]
```

lanzagar:
> The way that Cosine treats discrete columns means that it implicitly depends on the variable reuse™ meaning it can produce different results depending on the history and order of loaded data

Not just that, it seems to me that the way Cosine currently treats discrete columns is just plain wrong. It basically differentiates between the first and all other values (which it treats as equal)?
I am not sure if this is wanted in some circumstances or was just a first approach to make it work and never redone? Maybe @janezd remembers anything about this?
Why I find it strange is that it explicitly implements `discrete_to_indicators` that does this instead of e.g. simply calling Continuize on the data. Maybe due to performance concerns? Anyway, we should probably either change `supports_discrete` to False or change to one hot encoding (or something even better?).

lanzagar:
I have changed Cosine to not advocate support of categorical features until this is resolved. This affects the Distances widget too - everything above is the same there as well. Now it shows a warning and ignores categorical features for Cosine distance.
I added the same warning in Silhouette for metrics that do not support categorical features.

Issue #3028, opened 2018-05-08
Better representation of repeated points in scatter plot


smihael:
Currently, when a data point is repeated multiple times it will just overlap the previous ones on a scatter plot. By adding some jittering, one can get an impression that there are multiple points with the sme value, but it is not always the best decision.

It would be nice to have a possibility to make points size dependent on the count of overlapping points, something like bubble chart in certain software.

There is a solution how you would achieve the same in matplotlib:
https://stackoverflow.com/questions/37759475/scatter-plot-same-point-repeated-several-times-python

ajdapretnar:
I absolutely agree this is needed. It would also be great to implement this is Geo add-on as well.

markotoplak:
Well, the stackoverflow solution would not be quite enough: we also need to show multiple classes (different colors) which can also overlap, and, if the widget is showing categorical data, they do overlap exactly.

ajdapretnar:
We could turn off coloring by class in this case? This would be an easy solution. Or color by mode...

ajdapretnar:
OR! Implement pie charts for coloring dots. 🤣

janezd:
When refactoring visualizations in #3089, @VesnaT and I have ported this feature, but we haven't made any real effort to make the code nice because I don't like neither this feature nor its implementation (not from the code perspective but from its usability).

1. Coloring by majority is even more misleading than jittering. (Pie charts would be better, but ... no.) Hence, this can only be used when all points are of the same color and jittering is disabled. Implementing it like this complicates the user interface: if the user sets Sizes to Overlap, then Color should be automatically set to Same color. If the user sets Colors to anything but Same, Orange has to check whether Sizes are Overlap and set them to Same size.

2. This shouldn't be used together with jittering. If overlap is computed from jittered coordinates, it is wrong, since points can overlap by random. If it's computed from original coordinates (when jittering is otherwise enabled), it doesn't make sense because the points that are grouped are not actually plotted at the same spot. At least, setting Sizes to Overlap should disable jittering.

3. I would argue that if the data contains so many overlapping points that jittering gives a wrong impression, then Scatterplot is the wrong visualization in the first place. If this functionality is needed, it could be implemented differently -- with a widget that would data instances with the same value of a subset of attributes into a single instance, add a count attribute and perhaps remove all attributes except the counted. Or something along this line. Scatterplot can then visualize these counts by using the count as size.

[...]

PR #3177, opened 2018-08-01
[ENH] Scatterplot: indicate overlap of points.

thocevar:
##### Issue
Implements #3028 

##### Description of changes
Scatterplot computes the set of indices at each coordinate for later use in compute_sizes and compute_colors. The sizes are multiplied by log_2 of the number of points. The color represents the most frequent color (or one of them if there are several).

scatterplot-overlap

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -56,6 +56,12 @@ class OWMDSGraph(OWScatterPlotGraph):
         self.plot_widget.setAspectLocked(True, 1)
 
     def compute_sizes(self):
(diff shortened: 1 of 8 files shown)


VesnaT:
For me the updated scatter plot looks confusing, since it states 'Same size', but there appear to be points of different sizes.
I suggest adding an option to Size drop down, which enables a user to set point sizes according to the number of overlapping points, and keeping the default look as it was.

ajdapretnar:
I agree. Adding an option to Size would make a lot of sense!

PR #3178, opened 2018-08-01
[FIX] OWSql: enforce data download for non PostgreSQL databases

### LINKED ISSUE #3043: SQL DB access error using widgets in Orange 3.13
##### Orange version

3.13

##### Expected behavior

Load data from SQL server database

##### Actual behavior

Throws exception whenever a widget is used to access the data:

 b"Incorrect syntax near the keyword 'SELECT'.DB-Lib error message 20018, severity 15:

use the SQL Table widget to connect to SQL database attach and use a widget (select columns, select rows, box plot, data table etc.O

orange sql db error

robertcv:
##### Issue
Fixes #3043 
##### Description of changes
SqlTable has poor support for non-PostgreSQL databases. This enforces the download of the selected table so that it works as a normal Table.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owsql.py
+++ b/Orange/widgets/data/owsql.py
@@ -250,6 +250,10 @@ class OWSql(OWWidget):
                 self.download = True
                 self.downloadcb.setEnabled(False)
 
+            if not is_postgres(self.backend):
+                self.download = True
+                self.downloadcb.setEnabled(False)
+
             self._save_credentials()
             self.database_desc = OrderedDict((
                 ("Host", self.host), ("Port", self.port),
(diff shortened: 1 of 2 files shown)


Ryujjin:
Thank you for the patch but with the auto download always enabled the data is to large to fit in memory.  No data is loaded and thus no widgets can be used to manipulate the information. What is the workaround for working with large databases?

Thanks!

ajdapretnar:
I thought we went with sampling here... But it could be we discarded this option for some reason.

@Ryujjin Did you try using Data Sampler immediately after SQL and sample it yourself?

Ryujjin:
Thanks for the reply, I need to be able to pick date ranges or time intervals, last I looked the sampler appeared to be a random range but I could be wrong will look at it again. 

Thanks!

lanzagar:
If you know what kind of subset of a large data set you want to explore it will probably be best to use an SQL select query to get just that data. This will also be much faster than transferring large amounts of data and sampling on the client.

PR #3180, opened 2018-08-02
[ENH] Add pandas_compat.table_to_frame(tab)

apetrov:
### Feature

Convert Orange.data.Table instance to pandas dataframe

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation

### DIFF
--- a/Orange/data/pandas_compat.py
+++ b/Orange/data/pandas_compat.py
@@ -11,7 +11,7 @@ from Orange.data import (
     ContinuousVariable,
 )
 
-__all__ = ['table_from_frame']
+__all__ = ['table_from_frame', 'table_to_frame']
 
 
 def table_from_frame(df, *, force_nominal=False):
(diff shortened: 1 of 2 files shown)


Orange/data/pandas_compat.py
```diff
@@ -76,3 +76,42 @@ def _is_datetime(s):
...
+    Parameters
+    ----------
+    tab : Table
+
+    Returns
+    -------
+    pandas.DataFrame
+    """
+    def _column_to_series(col, vals):
+        print(col.name)
```
astaric: I guess this was used for debugging. Please remove.

Orange/data/pandas_compat.py
```diff
@@ -76,3 +76,42 @@ def _is_datetime(s):
...
+        if col.is_discrete:
+            labels = [col.values[i] for i in np.vectorize(int)(vals)]
```
kernc: `vals.astype(int)` avoids a call into python.

But what if some values are NaN?

Alternative:
```py
return col.name, pd.Categorical.from_codes(codes=pd.Series(vals).fillna(-1).astype(int),
                                           categories=col.values, ordered=col.ordered)

File: Orange/data/pandas_compat.py [hunk 3]
```diff
@@ -76,3 +76,42 @@ def _is_datetime(s):
...
+            return (col.name, pd.Series(labels).astype('category'))
+        elif col.is_time:
+            return (col.name, pd.to_datetime(vals,unit='s').to_series())
```
kernc: spaces after commas.

Orange/data/pandas_compat.py
```diff
@@ -76,3 +76,42 @@ def _is_datetime(s):
...
+            return (col.name, pd.Series(vals))
+    def _columns_to_series(cols, vals):
+        return [_column_to_series(col, vals[:,i]) for i,col in enumerate(cols)]
+
+    x,y,metas = [],[],[]
+    domain = tab.domain
+    if domain.attributes:
+        x = _columns_to_series(domain.attributes,tab.X)
+    if domain.class_vars:
+        y = _columns_to_series(domain.class_vars,tab.Y.reshape(tab.Y.shape[0],1))
```
kernc: Reshaping Y here doesn't work with multiple target columns. Perhaps:
```py
table.Y.reshape(table.Y.shape[0], len(domain.class_vars))

Orange/data/pandas_compat.py
apetrov: thanks

Orange/data/pandas_compat.py
apetrov: Great suggestion, thanks, fixed.

Orange/data/pandas_compat.py
apetrov: Thanks

apetrov:
@astaric sure, added basic tests and looks good on Iris data set.
checking for datasets with datetime columns so I could test datetime columns properly.

apetrov:
update: fixed pylinting.

apetrov:
ok, looks like a couple of datasets doesn't work with it:
```python
from Orange.data import Table
from Orange.data.pandas_compat import table_to_frame
datasets = [ f.split('.')[0] for f in os.listdir("Orange/datasets") if '.tab' in f]
for ds in datasets:
    print(ds)
    table = Table(ds)
    table_to_frame(table)
~

astaric:
I'd add it as a test, but mark it as @skip, so it can be uncommented when needed.

You wrote in an earlier comment that the function did not work on a couple of datasets. Could you test on them as well? (not sure what the problem was)

apetrov:
@astaric ok, that's a good point.

Yes, it helped me to troubleshoot a problem with datetime column. works on any Orange build-in dataset

PR #3187, opened 2018-08-05
[ENH] Data Info display data set name

BlazZupan:
##### Issue
No data set name displayed in the info box. This could be important when checking if the names of the data sets are set correctly.

##### Description of changes
Includes the box with the data set name.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/data/owdatainfo.py
```diff
@@ -99,6 +101,11 @@ def update_size():
 
         threading.Thread(target=update_size).start()
 
+        if hasattr(data, "name"):
+            self.data_set_name = data.name
+        else:
+            self.data_set_name = "N/A"
```
lanzagar:
```py
self.data_set_name = getattr(data, "name", "N/A")
```

PR #3200, opened 2018-08-13
Put red dot of Canvas LinkItem on top

borondics:
##### Issue
The NodeItem was cutting the processing indicator dot in two, which was visible only when the colors didn't match, i.e. during processing, while the dot is red.

##### Description of changes
Looks better if the red processing indicator dot is on top.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/canvas/items/linkitem.py
+++ b/Orange/canvas/canvas/items/linkitem.py
@@ -286,7 +286,7 @@ class LinkItem(QGraphicsWidget):
     """
 
     #: Z value of the item
-    Z_VALUE = 0
+    Z_VALUE = 200  # the red dot should appear on top of the node
 
     #: Runtime link state value
     #: These are pulled from SchemeLink.State for ease of binding to it's


lanzagar:
While it is slightly nicer if the red dot is full instead of cut in two, I don't think this is really a big issue.
However, this change also puts links above nodes, which I do not think is a good idea.

See example:
linksabove

borondics:
@lanzagar you are right, this doesn't look too good at all with the widget below. Maybe separating the line and the dot would fix this...

lanzagar:
I don't really mind the current state of the red dot that much so I wouldn't complicate things if not necessary.
Maybe @irgolic has some aesthetic/UI comments on this and is willing to work on a better indication of the current execution point in the workflow if the current one is not sufficient.

irgolic:
This is a very welcome change, however as @lanzagar has pointed out, it should be implemented in such a way that LinkAnchorIndicator is drawn over NodeItem, drawn over LinkCurveItem.

On the topic of alternative ways of displaying the current execution point, I believe a simple animation would be most effective. If a dashed line were not already used to show a no-data-flow state, I would have suggested a scrolling dashed line similar to this:

scrolling-dashed-line

Perhaps a dotted line?

PR #3217, opened 2018-08-31
[FIX] Variable: Prevent hashing of Values of DiscreteVariable.

### LINKED ISSUE #2912: Value.__hash__ is wrong
Hash function should return the same value for all objects that are considered equal.

steps to reproduce:
```
from Orange.data.variable import StringVariable, Value
val = Value(StringVariable("s"), "test")
val == "test"
# True
hash(val) == hash("test")
# False
```

thocevar:
##### Issue
Values of StringVariable are hashed incorrectly (fixes #2912).

Values of DiscreteVariable are impossible to hash as dictated by the `__eq__` function. For example:
```
from Orange.data.variable import *
val = Value(DiscreteVariable("var", ["red","green","blue"]), 1)
print(val == 1, hash(val) == hash(1))  # True True
print(val == "green", hash(val) == hash("green"))  # True False
```

##### Description of changes
As we can not make `hash(1) == hash("green")`, we prevent hashing of Values of DiscreteVariables with an exception. Users should hash directly ids or domain values instead.

This might break things that use dictionaries or sets of Values. However, the tests didn't discover any such cases.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -207,6 +207,11 @@ class Value(float):
         raise TypeError("invalid operation on Value()")
 
     def __hash__(self):
+        if self.variable.is_discrete:
+            # It is not possible to hash the id and the domain value to the same number as required by __eq__.
+            # hash(1) == hash(Value(DiscreteVariable("var", ["red", "green", "blue"]), 1)) == hash("green")
+            # User should hash directly ids or domain values instead.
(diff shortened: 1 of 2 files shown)


brunap:
Lan, this change disables plotting in Biopharm add-on: ProcessPlot widget.

Issue #3223, opened 2018-09-01
Canvas won't save NN model to file using Save Model widget


enjoysmath:
Orange version: 3.15
Expected behavior:  Model saved to file
Actual behavior: Exception is thrown related to pickling a local variable, 0kb resulting in file
   1.  Train a NN widget
   2.  Connect it to Save Model widget
   3.  Try saving model with the Save Model widget
Additional Info:
[Screenshot](https://screenshot.net/8qq1gsv)

gbaptista:
Same probleme here:
```
  File "/orange3/Orange/widgets/model/owsavemodel.py", line 122, in browse
    self.save(filename)
  File "/orange3/Orange/widgets/model/owsavemodel.py", line 90, in save
    pickle.dump(self.model, f)
AttributeError: Can't pickle local object 'OWNNLearner.__update.<locals>.callback'
```

For some reason `pickle.dump(self.model, f)` is not working...

I tried to change the code and use [dill](https://pypi.org/project/dill/) instead of [pickle](https://docs.python.org/3/library/pickle.html), but without success...

enjoysmath:
I'm pretty good at pickling things.  Usually callback functions aren't pickleable or you have to do something extra.

enjoysmath:
Looks like they're trying to pickle a local variable.  I will look into it.  I've installed the dev version of orange3 \o/

enjoysmath:
```

Exception:
AttributeError: Can't pickle local object 'OWNNLearner.__update.<locals>.callback'
Module:
Orange.widgets.model.owsavemodel:90
Widget Name:
Save Model
Widget Module:
Orange.widgets.model.owsavemodel:106
Widget Scheme:
C:\Users\FRUITF~1\AppData\Local\Temp\ows-ew10kef7.ows.xml
Version:
3.16.0.dev0+7857c4b
Environment:
Python 3.6.5 on Windows 10 10.0.17134 AMD64
Installed Packages:
[...]

PR #3230, opened 2018-09-05
Fix OWNeuralNetwork model pickling

markotoplak:
##### Issue
#3223 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


enjoysmath:
How do / when can I use this fix?

markotoplak:
@enjoysmath, you could checkout the branch with the fixes, and then install and run it. Perhaps an easier option is to find `owneuralnetwork.py` on your computer and add that 1 line I added there in this PR (the rest are tests).

enjoysmath:
When will it be integrated into the main branch so I can just download zip
and install?  That's what I did last time...

Sry, new to github!

I can help on this project.  I have like 4 years of PyQt5 experience.

And am really good at debugging, at least once I have my env set up.

I use WingWare pro.

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -222,6 +222,7 @@ def _task_finished(self, f):
         else:
             self.model.name = self.learner_name
             self.model.instances = self.data
+            self.model.skl_model.orange_callback = None  # remove unpicklable callback
```
lanzagar: This fixes the widget output which can now be pickled. But a general NN model (unfixed, not produced by the widget) is still unpicklable right? Are we fine with that?

PR #3237, opened 2018-09-07
[FIX] io: Handle mismatched number of header/data values

### LINKED ISSUE #1471: Strip empty columns in .xlsx files
Hi,
I am new to orange and I get the error while importing my data. Hope you could help me.
##### Orange version

3.3.6 (on windows 7)
##### Expected bahavior

Dataset has 61 columns and 338 rows. It is read as a XLSX file but not as a CSV file
##### Actual behavior

Gives the error: "Cannot parse dataset libro1.csv: too many indices for array".

[Libro1.xlsx](https://github.com/biolab/orange3/files/371787/Libro1.xlsx)

robertcv:
##### Issue
Fixes #1471 
Orange can't load fils where there are more data then header columns
##### Description of changes
Strip data columns
##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -618,9 +618,13 @@ class FileFormat(metaclass=FileFormatMeta):
         # Determine maximum row length
         rowlen = max(map(len, (names, types, flags)))
 
+        strip = False
+
         def _equal_length(lst):
+            nonlocal strip
             if len(lst) > rowlen > 0:
                 lst = lst[:rowlen]
+                strip = True
             elif len(lst) < rowlen:
                 lst.extend(['']*(rowlen - len(lst)))
             return lst
@@ -629,6 +633,9 @@ class FileFormat(metaclass=FileFormatMeta):
         data = [_equal_length([s.strip() for s in row])
                 for row in data if any(row)]
         data = np.array(data, dtype=object, order='F')
+
+        if strip:
+            warnings.warn("Columns with no headers were removed.")
 
         # Data may actually be longer than headers were
         try:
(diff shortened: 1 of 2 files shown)


ajdapretnar:
Hmm... I think it was assumed here that the values will be empty, meaning no column names AND no values. At least that's what happened sometimes with Excel.
If there are more items in a row than there are headers, then I think we should go with the second approach, i.e. Feature 1. You don't want to discard values when there are no columns names.
A warning in either case would be nice.

astaric:
+1 for warning.

my vote is for 1. since it is already implemented (and less complicated :))

Orange/tests/test_io.py
```diff
@@ -104,6 +105,20 @@ def test_open_bad_pickle(self):
...
+        a, b
+        1, 0,
+        1, 2,
+        """
+        c = io.StringIO(samplefile)
+        with self.assertWarns(UserWarning) as cm:
+            table = CSVReader(c).read()
+        self.assertEqual(len(table.domain.attributes), 2)
+        self.assertEqual(cm.warning.args[0],
+                         "Columns with no headers were striped.")
```
ajdapretnar: Striped --> Removed or stripped.

Orange/data/io.py
```diff
@@ -618,15 +618,25 @@ def data_table(cls, data, headers=None):
...
+                lst.extend(['']*(rowlen - len(lst)))
             return lst
 
         # Ensure all data is of equal width in a column-contiguous array
         data = [_equal_length([s.strip() for s in row])
                 for row in data if any(row)]
         data = np.array(data, dtype=object, order='F')
 
+        if strip:
+            warnings.warn("Columns with no headers were striped.")
```
ajdapretnar: I would prefer "Columns with no headers were removed.", since stripped is a bit Pythonish. Or at least 'stripped'.

Issue #1563, opened 2016-09-16
Hierarchical Clustering doesn't retain settings


ajdapretnar:
Hierarchical Clustering doesn't retain selection or label settings.

1.) Selection is gone when reloading data. To reproduce: load iris, select a branch in dendrogram, press 'Reload' in File widget. Selection disappears.

2.) Label settings default to enumeration, are empty to begin with and are lost upon reloading the data. Label (annotation) settings should be retained if domain matches. Possibly even just the particular column selected as label (to be discussed). Also, enumeration as a label should probably be the default only if there's no class label or meta data. Class label should probably be the default and followed by a meta attribute.

astaric:
Should be fix in #1748. Can you check?

ajdapretnar:
2.) is fixed, 1.) is not.
Selections don't get saved in schemas, which is problematic.

robertcv:
![](https://github.trello.services/images/mini-trello-icon.png) [Hierarchical Clustering doesn't retain settings](https://trello.com/c/p4BE3ZUj/1-hierarchical-clustering-doesnt-retain-settings)

PR #3249, opened 2018-09-14
[FIX] OWHierarchicalClustering: save manually selected branches

robertcv:
##### Issue
Fixes 1.) in #1563
##### Description of changes
Save and restore selected branches.
##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


ajdapretnar:
Tests fail. Please check.

Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -1174,6 +1189,7 @@ def commit(self):
             return
 
         selection = self.dendrogram.selected_nodes()
+        self.selected_trees = list(map(hash, selection))
```
ales-erjavec: Hashes should not be considered as stable ids suitable for persistence (neither are they guaranteed unique).

robertcv: I am aware that they are not completely unique however I think they are still a suitable as ids because the probability of having a collision is very small. Do you recommend any other parameter I could use as ids?

PR #3260, opened 2018-09-19
OWProjectionWidget: Introduce new parent for projection widget

VesnaT:
##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owradviz.py
+++ b/Orange/widgets/visualize/owradviz.py
@@ -8,40 +8,30 @@ from sklearn.model_selection import cross_val_score
 from sklearn.neighbors import KNeighborsClassifier, KNeighborsRegressor
 
 from AnyQt.QtGui import QStandardItem, QColor
-from AnyQt.QtCore import (
-    Qt, QEvent, QRectF, QPoint, pyqtSignal as Signal
-)
-from AnyQt.QtWidgets import (
-    qApp, QApplication, QToolTip, QGraphicsEllipseItem
-)
+from AnyQt.QtCore import Qt, QRectF, QPoint, pyqtSignal as Signal
+from AnyQt.QtWidgets import qApp, QApplication
 
 import pyqtgraph as pg
 from pyqtgraph.graphicsItems.ScatterPlotItem import ScatterPlotItem
 
-from Orange.data import Table, Domain, ContinuousVariable, StringVariable
-from Orange.data.sql.table import SqlTable
+from Orange.data import Table, Domain, StringVariable
 from Orange.preprocess.score import ReliefF, RReliefF
 from Orange.projection import radviz
-from Orange.widgets import widget, gui, settings, report
+from Orange.widgets import widget, gui
 from Orange.widgets.gui import OWComponent
-from Orange.widgets.settings import Setting
-from Orange.widgets.utils.annotated_data import (
-    create_annotated_table, ANNOTATED_DATA_SIGNAL_NAME, create_groups_table
-)
+from Orange.widgets.settings import Setting, ContextSetting, SettingProvider
+from Orange.widgets.utils import vartype
 from Orange.widgets.utils.itemmodels import VariableListModel
 from Orange.widgets.utils.plot import VariablesSelection
-from Orange.widgets.visualize.owscatterplotgraph import OWProjectionWidget
 from Orange.widgets.visualize.utils import VizRankDialog
-from Orange.widgets.visualize.utils.component import OWVizGraph
-from Orange.widgets.visualize.utils.plotutils import (
-    TextItem, VizInteractiveViewBox
-)
-from Orange.widgets.widget import Input, Output
+from Orange.widgets.visualize.utils.component import OWGraphWithAnchors
+from Orange.widgets.visualize.utils.plotutils import TextItem
+from Orange.widgets.visualize.utils.widget import OWAnchorProjectionWidget
 
 
 class RadvizVizRank(VizRankDialog, OWComponent):
     captionTitle = "Score Plots"
-    n_attrs = settings.Setting(3)
+    n_attrs = Setting(3)
     minK = 10
 
     attrsSelected = Signal([])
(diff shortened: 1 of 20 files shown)


Orange/widgets/visualize/utils/widget.py
```diff
@@ -0,0 +1,623 @@
...
+    def set_data(self, data):
+        self.closeContext()
+        self.clear()
+        self.data = data
+        self.check_data()
+        self.init_attr_values()
+        self.openContext(self.data)
+        self.cb_class_density.setEnabled(self.can_draw_density())
+
+    def check_data(self):
```
janezd: I'm not sure any of these checks will fit any widget (Freeviz, maybe). I think the method should just call `clear_messages`.

- The parent already handles sparse data, except for coordinates. Scatter plot can (or should) handle sparse data for coordinates, too. Other widgets maybe as well. So the first check is (generally) not OK.
- Data is invalid if coordinates can't be computed. This depends upon the selection of attribute subset (e.g. x and y in scatter plot, variables in radviz...). This test is thus generally wrong. More importantly, computing `self.valid_data` here is misleading.

Orange/widgets/visualize/utils/widget.py
```diff
@@ -0,0 +1,623 @@
...
+    def get_subset_mask(self):
+        if self.subset_indices:
+            return np.array([ex.id in self.subset_indices
+                             for ex in self.data[self.valid_data]])
+        return None
+
+    # Plot
+    def init_embedding_coords(self):
+        """Setting self._embedding_coords.
+        Should be overridden when subclassed."""
```
janezd: Is this code useful for any derived class? If not, remove.

Orange/widgets/visualize/utils/widget.py
```diff
@@ -0,0 +1,623 @@
...
+        x_data = self.data.X
+        x_data[x_data == np.inf] = np.nan
+        x_data = np.nanmean(x_data[self.valid_data], 1)
+        y_data = np.ones(len(x_data))
+        self._embedding_coords = np.vstack((x_data, y_data)).T
+
+    def get_coordinates_data(self):
+        return (self._embedding_coords[:, 0], self._embedding_coords[:, 1]) \
+            if self._embedding_coords is not None else (None, None)
```
janezd: Maybe `return self._embedding_coords.T if self._embedding_coords is not None else (None, None)`? It makes it more obvious that it returns the complete `_embedding_coords` and not just the first two columns.

Orange/widgets/visualize/utils/widget.py
```diff
@@ -0,0 +1,623 @@
...
+from Orange.widgets.utils.sql import check_sql_input
+from Orange.widgets.visualize.owscatterplotgraph import OWScatterPlotBase
+from Orange.widgets.widget import OWWidget, Input, Output, Msg
+
+
+MAX_CATEGORIES = 11  # maximum number of colors or shapes (including Other)
+MAX_POINTS_IN_TOOLTIP = 5
+
+
+class OWProjectionWidgetBase(OWWidget):
```
janezd: Names of classes (already in the previous PR, so it's my fault) are not very informative.

Perhaps this could be `OWProjectionWidget` and the derived class, which is currently `OWProjectionWidget` should become `OWDataProjectionWidget`. This reflects the difference - this one can show any projection, including networks, while `OWDataProjectionWidget` has a data signal.

Orange/widgets/visualize/utils/widget.py
```diff
@@ -0,0 +1,623 @@
...
+
+    class Error(OWProjectionWidgetBase.Error):
+        sparse_data = Msg("Sparse data is not supported")
+
+    settingsHandler = DomainContextHandler()
+    selection = Setting(None, schema_only=True)
+    auto_commit = Setting(True)
+
+    GRAPH_CLASS = OWScatterPlotBase
+    graph = SettingProvider(OWScatterPlotBase)
```
janezd: I'm not sure how the widget's meta classes work, and nobody's sure how they will work in the future. Adding a setting provider like this can have (or "will can have") side-effects due to processing done to the meta class, which maybe cannot be cancelled in derived classes.

I think it would be better if the base widget wouldn't declare a setting provider and leave this to derived classes.

astaric: Declaring something as a setting provider should be harmless.

The only thing it does is tell the SettingHandler that there is a "component" named graph that also holds some settings:
https://github.com/biolab/orange3/blob/master/Orange/widgets/settings.py#L125
(called from SettingHandler.create)

janezd: I would still prefer removing this because the provider would always be replaced. Probably no projection widget will just reuse the base graph --- scatter plot adds a regression line, radviz adds a circle, freeviz adds axes... --- so this provider is always a wrong one.

janezd:
I have so far reviewed only the base widget. I like it, it's well organized.

My general remark is that I don't like methods `_add_top_controls`, `_add_middle_controls`, `_add_bottom_controls`. First, I don't know which is which. I supposed that top is just, for instance, x and y in the scatter plot. But the top is actually "points properties" and middle is "effects" and "plot properties".

I think I would probably do it like this (instead of having three methods for different parts of the widget):

```
    def setup_gui(self):
        g =  self.graph.gui
        control = self.controlArea

        self.point_box = g.point_properties_box(control)
        self.effects_box = g.effects_box(control)
        self.plot_box = g.plot_properties_box(control)
        self.control_area_stretch = gui.widgetBox(
            sizePolicy=QSizePolicy.Maximum, QSizePolicy.Minimum)
        self.graph.box_zoom_select(control)
        gui.auto_commit(control, self, "auto_commit",
                        "Send Selection", "Send Automatically")
        self._add_graph()
```

All boxes are accessible and can be appended. In general, widgets would override `setup_gui` like

```
    def setup_gui(self):
        # Add anything that needs to be at the very top of the control area
        super().setup_gui()
        # Add additional controls to existing boxes or add new boxes
```

Stretch is added with a QWidget so it can be removed instead of inserting stretch, so it can be removed if needed.

I will check tomorrow how this would work in actual widgets and whether there are other nasty exceptions like stretch.

If we keep `_add_top_controls` and similar, please rename them to `_add_point_properties` and so forth; a separate method for each box.

janezd:
Github doesn't allow me to comment this in the code since you haven't touched it. But I guess we can fix this in this PR anyway.

    def _vizrank_color_change(self):
        is_enabled = self.data is not None and not self.data.is_sparse() and \
            len(self.model_other) + len(self.model_selected) > 3 and \
            len(self.data[self.valid_data]) > 1 and \
            np.all(np.nan_to_num(np.nanstd(self.data.X, 0)) != 0)
        self.btn_vizrank.setEnabled(
            is_enabled and self.attr_color is not None
            and not np.isnan(self.data.get_column_view(
                self.attr_color)[0].astype(float)).all())
        self.vizrank.initialize()

Some conditions are in `is_enabled`, then some more are added in the call. I suppose they should be moved to `is_enabled`, and `self.vizrank.initialize()` should be guarded by `if is_enabled:`.

More importantly, some conditions are incorrect, I guess.

`len(self.model_other) + len(self.model_selected) > 3` should probably be just `len(self.model_selected) > 3`. This also means that this method should also be called from `_model_selected_changed`. And thus renamed from `_vizrank_color_change` to `_prepare_vizrank` or something.

`np.all(np.nan_to_num(np.nanstd(self.data.X, 0)) != 0)`: this one should check only attributes in `selected_model`. As far as I understand, it now prohibits vizrank if *any* attribute contains nans, as well as if any attribute is constant. Furthermore, please check whether `nanstd` is needed. I don't see why radviz wouldn't handle constant attributes.

janezd:
Wrt. to the above: keep the first condition as it is, but the rest stays, I guess. About `nanstd`: changing `radviz.normalize` to 

def normalize(x):
    """
    MinMax normalization to fit a matrix in the space [0,1] by column.
    """
    a = x.min(axis=0)
    b = x.max(axis=0)
    diff = b - a
    diff[diff == 0] = 1
    return (x - a[np.newaxis, :]) / diff[np.newaxis, :]

should already do the trick. Let's do it in this PR anyway. :)

Orange/widgets/visualize/owradviz.py
```diff
@@ -360,19 +346,18 @@ def set_data(self, data):
...
-                self.data = None
+                error(self.Error.no_instances)
             elif len([v for v in domain.variables + domain.metas
                       if v.is_primitive()]) < 3:
-                self.Error.no_features()
-                self.data = None
-            elif np.all(np.isnan(self.data.X).any(axis=1)):
-                self.Warning.missing_coords()
-                self.data = None
+                error(self.Error.no_features)
```
janezd: Doesn't this condition already ensure that `len(self.model_other) + len(self.model_selected) > 3`? (see `init_attr_values`)

Orange/widgets/visualize/owradviz.py
```diff
@@ -360,19 +346,18 @@ def set_data(self, data):
...
 
     def init_attr_values(self):
         super().init_attr_values()
```
janezd: Add

```
def primitive_variables(self):
    return [v for v in chain(domain.attributes, domain.metas)
                         if v.is_primitive()]
```

and call it from `set_data`, `check_data`, `init_attr_values` instead of doing this same in all of them. If `len(self.model_other) + len(self.model_selected) > 3` in `_vizrank_color_changed` is still needed, it can be replaced with checking `len(self.primitive_variables)`, too.

janezd:
I think that `get_embedding` should not call `self.graph.clear()`. When the selected vars are changed, the widget calls `self.setup_plot()` (I don't particularly like this -- it should directly call `self.graph.reset_graph()`, so we know what is happening without a detour.) `reset_graph` will already call `clear_graph`, so another call is redundant.

I would prefer treating `clear_graph` as `_clear_graph`-ish.

Orange/widgets/visualize/owradviz.py
VesnaT: No. The error is shown when len(self.model_other) + len(self.model_selected) < 3. So when the number of variables is 3, widget still shows the graph, only vizrank is disabled.

ajdapretnar:
Not sure if this relates to this PR (probably not), but Find Informative Projections doesn't clear filter box when the domain changes... Which is... weird.

janezd:
... though not unexpected. :)

ajdapretnar:
When zooming out and adjusting jittering, the projection is rescaled to the original zoom. Now this is unexpected. :) Even though, this is not the issue of this PR, but an issue that existed in old projections as well.

ajdapretnar:
Also funny. When I move the legend out of its normal position, then change one of the attributes that affects the legend (say add Shape), the legend moves.

ajdapretnar:
Weird #3. Zoom out of the projection, then click Class density and Reset zoom. Zoom won't be reset. However, before this PR, Class density rescales on its own! 😮

ajdapretnar:
Zooming in MDS is much faster in this implementation! 🙌 However, granularity of class density is a bit off.

screen shot 2018-10-05 at 16 33 05

janezd:
@VesnaT, I added one more commit. Sorry. :)

1. Derived graphs or widgets had no good way to create new selection groups or append to the last selection group since behaviour of `select_by_indices` was controlled by modifier keys. In https://github.com/janezd/orange3-network/tree/network-explorer-2, I need this for https://github.com/janezd/orange3-network/blob/network-explorer-2/orangecontrib/network/widgets/OWNxExplorer.py#L351.

    The way they are implemented now should also be easier to test, though I have not (yet?) added any specific tests because I'm working on networks.

2. If user adds multiple selection groups but then eliminate some, some group indices may be missing. This behaviour was particularly annoying in the network analysis widget. Group indices are now "compressed".

Except for these hiccups, new scatterplotgraphbase and *your* base widget work like charm.

markotoplak:
Thanks @ajdapretnar, I'll look into that.

Issue #3198, opened 2018-08-10
I have a data set with just 11% of positive target. Therefore, I would like to create my binary classification model after doing an undersampling. Is there a widget for this? I saw the data sampler, but looks like it is only for training and validation


saprissa2018:

ajdapretnar:
You are right, Data Sampler doesn't include over/undersampling. It would be a feature request.

astaric:
You can always split your data on class (using Select Rows) over/under sample one class with DataSampler and concatenate the result.

![screen shot 2018-09-12 at 11 13 45](https://user-images.githubusercontent.com/588601/45415021-f9538380-b67c-11e8-9b56-e12ee05a1769.png)

markotoplak:
@astaric, I like your approach. Should we put it into a Data Sampler documentation?

ajdapretnar:
I will do it.

astaric:
If #3244 helps, you can use it, otherwise close it once you make your PR.

Issue #2436, opened 2017-06-27
[ENH] Over/undersampling widget


Zefeng-Wu:
How to deal with unbanlanced data, for example, too much negative instances and few positive instances?
 Is there an option of class weight  to solve the unbanlance when building a model like SVM or other classifiers?

tastyminerals:
This question is not related to **Orange** specifically.
There is plenty of information about unbalanced datasets and they way to handle them, just google.
Here is one paper that might help you: http://sci2s.ugr.es/keel/pdf/algorithm/congreso/kubat97addressing.pdf

MrMauricioLeite:
but do we have SMOTE to oversample our datasets on Orange GUI?
Would we be able to undersample a dataset based on a specific classification forcing, let's say, result A to pre 50% of the output sample?

This would be a great start for unbalanced data mining as the options usually are Ignore it, Undersample it or Oversample it.

ajdapretnar:
This calls for the implementation of a new Over/undersampling widget.
Suggestions for implementation and contributions welcome.

PR #3269, opened 2018-09-27
[WIP][RFC] SMOTE oversampling

matejklemen:
##### Issue
Implements SMOTE (or rather the SMOTE-NC variation) - relevant issues: #3198, #2436.

##### Description of changes
This pull request adds the [SMOTE oversampling method](https://jair.org/index.php/jair/article/view/10302) and an interface to use it via Data sampler.
It requires more testing and likely some other changes, but I wanted to get some feedback and ask a question.

The main thing that I'm not sure about is the way output should be provided. With other methods, there's a pretty clear distinction between what belongs in the _sample_ and what in the _remaining data_ (the 2 ouputs of Data sampler).  
With oversampling, I see two ways to divide up the data:
(1.) Putting the existing data, merged with newly generated data all into the 'sample' output slot and None into the 'remaining' output slot (**current way**) or
(2.) putting the newly generated data into into the 'sample' output slot and the existing data into the 'remaining' output slot.  
I would like to get a second opinion / new suggestion on how this should be done. 

General comments are also much appreciated.

##### Things I am aware I need to fix:
- more testing (particularly the case where input is a SQLTable, which I've not given any attention yet),
- move the SMOTE logic into a separate function (i.e. outside of `__call__()`),
- include meta attributes in distance calculation (same way as categorical ones),
- figure out how to move generation of new data into a new thread (GUI currently freezes on bigger data sets). 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owdatasampler.py
+++ b/Orange/widgets/data/owdatasampler.py
@@ -37,8 +37,9 @@ class OWDataSampler(OWWidget):
     resizing_enabled = False
 
     RandomSeed = 42
-    FixedProportion, FixedSize, CrossValidation, Bootstrap = range(4)
+    FixedProportion, FixedSize, CrossValidation, Bootstrap, Oversample = range(5)
     SqlTime, SqlProportion = range(2)
+    MethodSMOTE = 0
 
     use_seed = Setting(False)
     replacement = Setting(False)
@@ -51,6 +52,10 @@ class OWDataSampler(OWWidget):
     sampleSizeSqlPercentage = Setting(0.1)
     number_of_folds = Setting(10)
     selectedFold = Setting(1)
+    oversampling_factor = Setting(1)
+    oversampling_method = Setting(MethodSMOTE)
+    oversampling_k = Setting(5)
+    min_class = Setting(0)
 
     class Warning(OWWidget.Warning):
         could_not_stratify = Msg("Stratification failed\n{}")
@@ -61,6 +66,9 @@ class OWDataSampler(OWWidget):
         sample_larger_than_data = Msg("Sample must be smaller than data")
         not_enough_to_stratify = Msg("Data is too small to stratify")
         no_data = Msg("Dataset is empty")
+        too_many_neighbors = Msg("Number of neighbors considered must be lower than " 
+                                 "number of all examples")
+        invalid_target = Msg("Target variable is either undefined or non-categorical")
 
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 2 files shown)


janezd:
I don't think this is a good idea.

First, I don't know how this widget would be used. I imagine that the only use of such data would be for learning. However, oversampled data must not be used in cross validation or similar schema because it would result in having duplicates of training data in test data. Oversampling would have to happen within cross validation, on training data only. But for this, the method would have to be implemented as a preprocessor, and not within this widget.

Second, I checked the paper and I don't understand how they managed to publish it. If I understood it correctly, they published ROC curves in which individual points correspond to different oversampling settings. In other words, their "ROC curves" (which are not really ROC curves in the usual meaning) show how oversampling affects cross validation -- whatever this is supposed to show. Unless I'm mistaken about this, I really wonder how could this get past reviewers.

janezd:
We discussed this PR at today's meeting.
 
1. Oversampling does not belong to this widget. *If anywhere*, it should be in preprocessing.
2. *If anywhere* refers to the observation that oversampling does not work (except perhaps in papers about oversampling).
3. If we still decide we want to have it in Orange, we should select an appropriate algorithm. Based on my reading the paper, SMOTE's approach doesn't look OK, and it was not properly tested, either.

janezd:
It would be if it worked. My observation is that it works just in paper about oversampling.

Balancing imbalanced data results in models that are incorrectly biased towards the minority class. With 95 % majority, it makes perfect sense to always predict the majority class ... unless your goal is detecting rare cases (which it actually probably is). But then you should not use algorithms that were designed for classification and that optimize classification accuracy (balancing will decrease it), predicted probabilities (balancing will obviously change them) or a similar score. Balancing data won't translate a problem of detecting anomalies into a classification problem.

Thus oversampling (although in such a way that will only oversample the training data) would lead to bad practice, while I don't see any scenarios in which this would actually be useful.

One could object that Orange is just a tool and it's the user's responsibility to not misuse it. I can think of two other issues in which we had similar discussions. But think about programming languages: modern languages are all about null-safety, though one could say it's user's responsibility to not dereference them. Kotlin's switch requires listing all cases (or explicitly give a default); wouldn't it be the programmers responsibility to cover all that can happen? Language designers would be hesitant to add features that are unreadable, confusing, ambiguous. I believe that some reasonable guidance is a good thing.

AuSpotter:
> It would be if it worked. My observation is that it works just in paper about oversampling.
> 
> Balancing imbalanced data results in models that are incorrectly biased towards the minority class. With 95 % majority, it makes perfect sense to always predict the majority class ... unless your goal is detecting rare cases (which it actually probably is). But then you should not use algorithms that were designed for classification and that optimize classification accuracy (balancing will decrease it), predicted probabilities (balancing will obviously change them) or a similar score. Balancing data won't translate a problem of detecting anomalies into a classification problem.
> 
> Thus oversampling (although in such a way that will only oversample the training data) would lead to bad practice, while I don't see any scenarios in which this would actually be useful.
> 
> One could object that Orange is just a tool and it's the user's responsibility to not misuse it. I can think of two other issues in which we had similar discussions. But think about programming languages: modern languages are all about null-safety, though one could say it's user's responsibility to not dereference them. Kotlin's switch requires listing all cases (or explicitly give a default); wouldn't it be the programmers responsibility to cover all that can happen? Language designers would be hesitant to add features that are unreadable, confusing, ambiguous. I believe that some reasonable guidance is a good thing.

Hi @janezd - actually the problem of imbalanced classes is a very relevant and important problem in the field of geoscience data analysis. Addressing imbalance is fundamentally important, e.g., problems that relate to the prediction of rock types from geophysical data (as one example [https://www.sciencedirect.com/science/article/pii/S2590197419300011](url). Note the use of SMOTE in this paper

** I will freely admit ** that SMOTE isn't a perfect solution, but those of us in the geoscience/spatial data fields need (at least) a way to randomly undersample majority classes; optimally a robust way to balance the classes in a sensible way

bellini1:
Hi. Emergency physician, pandemic advisor and Ai practitioner: We really need SMOTE for health care research and modeling of the pandemic. As above, not perfect but the vast majority of our outcome classes are undersampled and we need to figure out all sorts of synthetic data methods to improve our models and save lives, SMOTE, digital twins etc.
Please think about adding, Orange folks!
Thanks

AuSpotter:
Wow - that's a much more important use-case than my rock related one!

AuSpotter:
Is it possible to re-open this request, @janezd ?

Issue #3255, opened 2018-09-17
Selecting visualizations in VizRank dialogs


janezd:
Double clicking a cell allows the user to edit it, which does not make sense.

If the user selects a pair in the scatter plot vizrank, then changes the attribute in one of the combos in the widget, and then again clicks the same (already selected) cell in the vizrank dialog, the selection should be passed to the widget. It however doesn't because the selection is not changed.

ajdapretnar:
Is there no Qt flag such as `table.setEditTriggers(QAbstractItemView::NoEditTriggers)`?

janezd:
Sorry, I don't understand this comment.

ajdapretnar:
I thought we are just talking about editable fields. Then I realized the issue is also that the selection isn't communicated onwards. I prefer the second solution. If I click on the field x times, the widget should be updated x times... I guess.

markotoplak:
As long is changing selection with a keyboard still works. :)

janezd:
I've implemented an improved version of the first alternative (https://github.com/biolab/orange3/commit/c67d22ec46c87f4c2c5b826faaf3215b6a704b85). When the user changes the selection in the widget, the corresponding row in vizrank should be highlighted, otherwise the widget and vizrank are not synchronized.

Communication between the widget and vizrank is trivial if we use signals.

janezd:
Fixed in #3276.

PR #3276, opened 2018-10-01
Selection of attributes in the widget is reflected in VizRank

janezd:
##### Issue

#3255, second part. The problem can be re-viewed from a different perspective: the problem is that after the user selects something in VizRank, he can change the selection in the widget, while VizRank still incorrectly highlights the previously selected combination of attributes.

With this in mind, the correct solution is as follows.

##### Description of changes

As the user changes the selection in the widget, the widget emits signal that the attributes have been changed. Its vizrank highlights the newly selected combination if it exists, otherwise it clears the selection.

Caveat: if the user changes the selection while vizrank is running, nothing bad will happen. However, if the combination (s)he selected has not been evaluated by VizRank *yet*, it will not be highlighted neither at that moment (obviously) nor later. I don't think this is worth losing sleep over, though.

Radviz is a special case since the user can move anchors there, so reselecting the same combination makes sense - it resets the anchors. (Plus, signals in this widget are more complicated so I was unable to do it in the same way as in others. :))

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -56,6 +56,8 @@ class MosaicVizRank(VizRankDialog, OWComponent):
         self.marginal = {}
         self.last_run_max_attr = None
 
+        self.master.attrs_changed_manually.connect(self.on_manual_change)
+
     def sizeHint(self):
         return QSize(400, 512)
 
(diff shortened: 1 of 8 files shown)


janezd:
@lanzagar, you showed me some problem with this PR. I'm trying to replicate it from memory, but can't. Can you tell me how?

PR #3279, opened 2018-10-01
Fix incremental pca

janezd:
Tests in PR's fail in incremental PCA. This happens with scikit-learn >= 0.20. I suppose the problem is in https://github.com/scikit-learn/scikit-learn/issues/12234. This PR fixes it by extending the data (reusing some instances) so that it's size is a multiple of the batch size that will be used by IncrementalPCA.

Tests then fail because the PCA the first principal components for entire and reduced data sets do no match to 4 but just to two decimals. I thus reduced `4 -> 2` (as already a few lines above). @lanzagar, OK?

### DIFF
--- a/Orange/projection/pca.py
+++ b/Orange/projection/pca.py
@@ -114,6 +114,19 @@ class IncrementalPCA(SklProjector):
 
     def fit(self, X, Y=None):
         proj = self.__wraps__(**self.params)
+
+        # Fix for https://github.com/scikit-learn/scikit-learn/issues/12234
+        n_components = self.params.get("n_components")
+        if n_components is not None:
+            n_samples, n_features = X.shape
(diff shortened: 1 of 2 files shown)


janezd:
You mean forcing a batch size in tests? This would do the trick. In general, we can't search for batch sizes that satisfy `sample_size % batch_size == 0 or sample_size % batch_size > n_components`.

Another solution is to clip the data instead of extending it. This is equally (or more) wrong. I don't know what the upstream fix will be -- I haven't got an answer yet.

I agree with your solution to fix the tests, either by removing them or by forcing a better batch size. Which one do you prefer? Can you take over this issue?

lanzagar:
Replaced by #3280 for now, which skips the problematic test.
We can revisit the actual fix when more info is available upstream.

PR #3280, opened 2018-10-02
Skip failing Iterative PCA test

lanzagar:
Skip Iterative PCA test which started failing in scikit-learn 0.20.
See https://github.com/scikit-learn/scikit-learn/issues/12234

Depending on the upstream solution, we should do something about the problem itself too. However, Iterative PCA is not actually available in Orange Canvas (except as a prototype in combination with remote orange), so skipping the test for now should be ok and give us some time to decide what to do.


Orange/tests/test_pca.py
```diff
@@ -62,6 +64,8 @@ def __rnd_pca_test_helper(self, data, n_com, min_xpl_var):
         proj = np.dot(data.X - pca_model.mean_, pca_model.components_.T)
         np.testing.assert_almost_equal(pca_model(data).X, proj)
 
+    @unittest.skipIf(sklearn_version.startswith('0.20'),
```
kernc: Instead, skip if upstream `IncrementalPCA` raises for particular data?

janezd: I don't understand. Can you explain?

kernc: Skip the test for as long as https://github.com/scikit-learn/scikit-learn/issues/12234 is open. Or assert it is indeed open. Revisit the test once the upstream issue is fixed, not once 0.21 hits (which may not contain the fix).

lanzagar: Don't understand the change or the comment?
I did not want to just unconditionally skip the test, because then it is forgotten and skipped for eternity. The problems appear only in version 0.20, this temporarily makes them go away until 0.21, when it needs to be reviewed (hopefully just removing the skip because it is fixed :)). Ideally, we should fix this properly before that, as Incremental PCA still has issues, which is not good even though it is not used much.
@kernc that would indeed be better in terms of not skipping the test anymore immediately once the bug is fixed (e.g. 0.20.2), but if it is not fixed we will have this _temporary_ solution forever and a broken Incremental PCA. I am more worried about the latter. Also, your proposal includes more work :)

janezd: I didn't understand what @kernc meant. I prefer @lanzagar's way, also because it will force us to recheck this with every new version of scikit-learn.

PR #3284, opened 2018-10-04
Modify Mosaic widget to use domain models

### LINKED ISSUE #3283: Mosaic widget does not use DomainModel
... except for coloring.

This surprised me in https://github.com/biolab/orange3/pull/3276/commits/e83c9c80bfc8d9a5ac5a402988a7c3dc5b5b598c#diff-fc3272b6fde65b41a30d305ce903e3afR251. At first I made a mistake here because I reasonably supposed I'd get Variables, not their names.

janezd:
Fixes #3283 

Best if reviewed one commit at a time.

Change to DomainModels was not trivial because widget uses discretized data. Models are filled with original domain so that variables have the right icons, but the widget then has to convert variables from original to discretized and back.

PR also includes some refactoring. `interior_coloring` was declared as setting, yet it served just as an attribute and was not even needed since it was synonymous with `variable_color is None`.

I also reduced the number of lines to 1000. :)

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -1222,7 +1222,7 @@ def rename_setting(settings, old_name, new_name):
         settings[new_name] = settings.pop(old_name)
 
 
-def migrate_str_to_variable(settings, names=None):
+def migrate_str_to_variable(settings, names=None, none_placeholder=None):
     """
     Change variables stored as `(str, int)` to `(Variable, int)`.
 
@@ -1235,6 +1235,8 @@ def migrate_str_to_variable(settings, names=None):
         var, vartype = settings.values[name]
         if 0 <= vartype <= 100:
             settings.values[name] = (var, 100 + vartype)
+        elif var == none_placeholder and vartype == -2:
+            settings.values[name] = None
 
     if names is None:
         for name, setting in settings.values.items():
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -654,8 +654,7 @@ class OWMosaicDisplay(OWWidget):
                           y0 + currpos + height * 0.5 * perc,
                           y0 - self.ATTR_VAL_OFFSET,
                           y0 + currpos + height * 0.5 * perc]
-                    if side == 0:
-                        CanvasText(self.canvas, val, xs[side], ys[side], align)
+                    CanvasText(self.canvas, val, xs[side], ys[side], align)
                     space = height if side % 2 else width
                     currpos += perc * space + spacing * (total_attrs - side)
 
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -137,7 +137,7 @@ class MosaicVizRank(VizRankDialog, OWComponent):
                 self.attr_ordering = [a for _, a in attrs]
 
     def _compute_class_dists(self):
-        return self.master.interior_coloring == self.master.CLASS_DISTRIBUTION
+        return self.master.variable_color is not None
 
     def state_count(self):
         """
@@ -266,12 +266,9 @@ class OWMosaicDisplay(OWWidget):
         selected_data = Output("Selected Data", Table, default=True)
         annotated_data = Output(ANNOTATED_DATA_SIGNAL_NAME, Table)
 
-    PEARSON, CLASS_DISTRIBUTION = 0, 1
-
     settingsHandler = DomainContextHandler()
     settings_version = 1
     use_boxes = Setting(True)
-    interior_coloring = Setting(CLASS_DISTRIBUTION)  # TODO: This is not setting
     variable1 = ContextSetting(None)
     variable2 = ContextSetting(None)
     variable3 = ContextSetting(None)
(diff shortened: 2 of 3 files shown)


Orange/widgets/visualize/owmosaic.py
```diff
@@ -359,21 +356,12 @@ def __init__(self):
...
 
-    def _compare_with_total(self):
-        if self.data is not None and \
-                self.data.domain.class_var is not None and \
-                self.interior_coloring != self.CLASS_DISTRIBUTION:
-            self.interior_coloring = self.CLASS_DISTRIBUTION
-            self.coloring_changed()  # This also calls self.update_graph
-        else:
-            self.update_graph()
-
```
janezd: If I haven't overlooked something, this method was called by a combo that was disabled when `self.interior_coloring` was not `self.CLASS_DISTRIBUTION`, so the condition was never `True`.

mstrazar:
I think you have a wrong issue/PR reference there.

janezd:
True. Fixed. Thanks.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -676,80 +637,48 @@ def draw_text(side, attr, x0_x1, y0_y1,
...
+                    perc = counts[i] / float(total)
+                    xs = [x0 + currpos + width * 0.5 * perc,
+                          x0 - self.ATTR_VAL_OFFSET,
+                          x0 + currpos + width * perc * 0.5,
+                          x1 + self.ATTR_VAL_OFFSET]
+                    ys = [y1 + self.ATTR_VAL_OFFSET,
+                          y0 + currpos + height * 0.5 * perc,
+                          y0 - self.ATTR_VAL_OFFSET,
+                          y0 + currpos + height * 0.5 * perc]
                     if side == 0:
```
VesnaT: Feature values are now shown only for the first feature (bottom side). Is this intentional? If it is, why setting xs and ys for other sides as well?

Orange/widgets/visualize/owmosaic.py
```diff
@@ -676,80 +637,48 @@ def draw_text(side, attr, x0_x1, y0_y1,
...
-                    x0 + (x1 - x0) / 2,
-                    y0 - self.ATTR_VAL_OFFSET - self.ATTR_NAME_OFFSET,
-                    align, bold=1)
-            else:
-                CanvasText(
-                    self.canvas, attr,
-                    x1 + max_ylabel_w2 + self.ATTR_VAL_OFFSET,
-                    y0 + (y1 - y0) / 2,
-                    align, bold=1, vertical=True)
+                        CanvasText(self.canvas, val, xs[side], ys[side], align)
```
VesnaT: Feature values are now shown only for the first feature (bottom side). Is this intentional? If it is, why setting xs and ys for other sides as well?

VesnaT: Feature values are now shown only for the first feature (bottom side). Is this intentional? If it is, why setting xs and ys for other sides as well?

VesnaT:
Since features are no longer saved as string, the migration of these settings should be handled within the PR too.

janezd:
Migrations of which settings? Not those? https://github.com/biolab/orange3/pull/3284/files#diff-fc3272b6fde65b41a30d305ce903e3afR921

VesnaT:
image
The widget crashes with the following workflow (select one instance in Data Table). 
I guess it is also an old issue.

VesnaT:
About migration: I'm not sure. Can look deeper into that, but an old workflow doesn't work.

janezd:
Could you attach the workflow?

Orange/widgets/settings.py
```diff
@@ -1235,6 +1235,8 @@ def _fix(name):
         var, vartype = settings.values[name]
         if 0 <= vartype <= 100:
             settings.values[name] = (var, 100 + vartype)
+        elif var == none_placeholder and vartype == -2:
+            settings.values[name] = None
```
janezd: @astaric, could you check this? `DomainModel` and `VariableModel` do not contain a string placeholder `"(None)"` but an actual `None`. Hence migration from string settings to variables, must replace `"(None)"` with `None`.

I however wonder why haven't we noticed this problem before, for instance in scatter plot's shape combo. Is Mosaic somehow specific? Or is the bug somewhere else and this fix somehow masks it?

Orange/widgets/visualize/owmosaic.py
```diff
@@ -476,31 +456,21 @@ def reset_graph(self):
         self.update_graph()
 
     def set_color_data(self):
-        if self.data is None or len(self.data) < 2 or len(self.data.domain.attributes) < 1:
+        if self.data is None or len(self.data) < 2:
```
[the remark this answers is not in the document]
janezd: This doesn't look like a necessary condition for setting the color, but I suspect it was added as a fix for some crash that occurs sometime later if colors are set -- instead of fixing it at its proper place. I would leave this as it this.

Orange/widgets/visualize/owmosaic.py
janezd: Fixed.

janezd:
I opened a new issue (#3416) about crashing since this is unrelated to this PR.

I hopefully fixed migrations (please check). Changes are in new commits. It's ready for re-review on Monday.

Orange/widgets/visualize/owmosaic.py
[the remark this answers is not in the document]
VesnaT: No, this indeed is a bug and should be fixed (gonna open an issue). Not the removed condition, but the existing one. `len(self.data) < 2` is used only here which causes wrong value of `self.discrete_data` when resetting the graph. Domains (current `self.data.domain` and the domain `self.discrete_data` was built from) then differ and the widget crashes.

VesnaT:
The migration and labels now work as expected. I would merge the PR, unless @astaric has anything to add.

PR #3290, opened 2018-10-05
owselectbydataindex: also output unmatched data

markotoplak:
##### Issue
After transformation of the data such as PCA and then selecting some of the transformed data, users wish to remove the selected instances from the original data.

##### Description of changes
Added two outputs to Select by Data Index widget. Don't be shy If you remember better names for these outputs.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation


janezd:
I'm merging this, but I'm unhappy with the name of the widget. I had no idea what it does before I saw its signal names and recalled that we have talked about this. Maybe "Select by subset" could be a better name? Index does mean anything, even to me.

markotoplak:
Its original name was "Select Subset". I agree with the name you propose because we are already using subsets frequently.

Issue #2035, opened 2017-02-21
Replicable sampling in Test & Score


astaric:
Add an option to choose replicable sampling in Test & Score widget. It should do the same as the Replicable checkbox in Data Sampler

MaxGreil:
I can have a look at this issue.

astaric:
@MaxGreil, how is it going? Did you encounter any problems adding the checkbox?

MaxGreil:
Hi @astaric , I am sorry but I couldn't start to implement yet. I had some troubles installing the software on my Ubuntu `(ImportError: ../lib/python3.5/site-packages/PyQt5/QtCore.so: undefined symbol: PySlice_AdjustIndices)`.
The solution to this was to add `pyqt5==5.10.1` in the `requirement-dev.txt`.

ajdapretnar:
Similar to #3104, no? @markotoplak Would you be interested in adding this?

markotoplak:
Hmmm, @astaric, the Test & Score already seems to pass a fixed random seed forward. Do we need a non-replicable option?

astaric:
Is there another way to get different AUC results on a dataset?

markotoplak:
Shuffle your data? But perhaps in a different widget? 

Currently, data sampler can do a len-1 sample, which is almost shuffling.

astaric:
This could work. In order to consider this issue resolved, I would like the following though:
  1. Modified DataSampler that allows selecting all instances (not just n-1)
  2. DataSampler doc mentioning that the widget shuffles the dataset
  3. A test that checks that DataSampler really shuffles the dataset
  4. An example in Test & Score docs showcasing how this can be done (like #3245)

@markotoplak, what do you think?

markotoplak:
I like the concept to have a randomization on one place only. But perhaps it would be more natural for that functionality to come in the Randomize widget. I would certainly look for it there first.

Issue #3287, opened 2018-10-05
Orange library should use fixed random seeds by default


markotoplak:
Our workflows are not replicable (#3103).

As a first step, let's do what @kernc suggested in #3103: change default random seeds in `Orange.` to a fixed value.

markotoplak:
We decided that, if we use an external library, we keep its defaults. Then, random seeds should be set in widgets, in needed.

PR #3293, opened 2018-10-05
Replicable sampling in Test & Score

MaxGreil:
##### Issue
Issue #2035 
The Test & Score widget should also have a Replicable checkbox similar to the Data Sampler widget.

##### Description of changes
Added an option to choose replicable sampling in Test & Score widget.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -500,6 +500,9 @@ class OWTestLearners(OWWidget):
         if self.__needupdate:
             self.__update()
 
+    def settings_changed(self):
+        self._param_changed()
+
     def kfold_changed(self):
         self.resampling = OWTestLearners.KFold
         self._param_changed()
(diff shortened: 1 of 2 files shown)


ajdapretnar:
Partially addresses #3103 and #3287. I think we are solving this issue by passing a parameter `random_seed` to the constructor with a fixed value of 0.

@markotoplak Could you please review and see what can be improved to be in line with #3287?

markotoplak:
Thank you for your PR. 

Currently, Test&Score is replicable by default. What is your use case where you would like to have it disabled? Test & Score is already complex and we are trying to keep widgets simple.

To see how do the different random splits affect performance, you could use select rows to sample your data (sample len(data) - 1). That will reorder your data and therefore the splits will be applied differently.

MaxGreil:
@markotoplak thank you for your comment.
Did I understand issue #2035 then incorrectly because sampling in the Test&Score widget is already replicable?

markotoplak:
@MaxGreil, as I understand it (please correct me if I failed to notice something), your code tries to add a no-replicable option to the widget. Otherwise the random seed was already set before.

I do not know what that bug was about. Perhaps there are some things in the backend that make results non-replicable. For example, if any learner has random elements is should be made replicable too.

So for me fixing the bug should start by finding an example which does not currently replicate, write it as a test, and then fix it. Or did you already try it?

MaxGreil:
@markotoplak since the requirements from the original issue changed, can this PR be closed?

Issue #3157, opened 2018-07-25
Provide percent missing values in Info box 


stuart-cls:
Hi,

The Data Table widget provides a nice overview of the dataset in the "Info" box, specifically including the percent missing values. It would be nice if the File widget "Info" box did the same to help diagnosing problematic datasets.

Thanks!

##### Orange version
3.14.0

##### Expected behavior
File "Info" box provides the same text as Data Table "Info" box.

##### Actual behavior
File "Info" has less info.

rokgomiscek:
Closed via #3305

PR #3305, opened 2018-10-11
[FIX] File: Provide percent missing values in Info box

MaxGreil:
##### Issue
Fixes issue #3157.

##### Description of changes
Info Box in widget File now displays the same information as the info box in widget Data Table

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1065,9 +1065,19 @@ class Table(MutableSequence, Storage):
         missing_x = not sp.issparse(self.X) and bn.anynan(self.X)   # do not check for sparse X
         return missing_x or bn.anynan(self._Y)
 
+    def has_missing_attribute(self):
+        """Return `True` if there are any missing attribute values."""
+        return not sp.issparse(self.X) and bn.anynan(self.X)    # do not check for sparse X
+
     def has_missing_class(self):
         """Return `True` if there are any missing class values."""
         return bn.anynan(self._Y)
+
+    def get_nan_percentile_attribute(self):
+        return np.isnan(self.X).sum() / self.X.size * 100
+
+    def get_nan_percentile_class(self):
+        return np.isnan(self._Y).sum() / self._Y.size * 100
 
     def checksum(self, include_metas=True):
         # TODO: zlib.adler32 does not work for numpy arrays with dtype object
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1073,11 +1073,11 @@ class Table(MutableSequence, Storage):
         """Return `True` if there are any missing class values."""
         return bn.anynan(self._Y)
 
-    def get_nan_percentile_attribute(self):
-        return np.isnan(self.X).sum() / self.X.size * 100
+    def get_nan_frequency_attribute(self):
+        return np.isnan(self.X).sum() / self.X.size
 
-    def get_nan_percentile_class(self):
-        return np.isnan(self._Y).sum() / self._Y.size * 100
+    def get_nan_frequency_class(self):
+        return np.isnan(self._Y).sum() / self._Y.size
 
     def checksum(self, include_metas=True):
         # TODO: zlib.adler32 does not work for numpy arrays with dtype object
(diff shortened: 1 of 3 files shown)


MaxGreil:
@markotoplak can you please have a look at this?

Orange/widgets/data/owfile.py
```diff
@@ -385,19 +385,36 @@ def _describe(self, table):
...
+                text += " (no missing values)"
         elif domain.has_discrete_class:
-            text += "<br/>Classification; categorical class with {} values.".\
+            text += "<br/>Classification; categorical class with {} values".\
                 format(len(domain.class_var.values))
+            if table.has_missing_class():
+                miss = "%.1f" % table.get_nan_percentile_class()
+                text += " ({}% missing values)".format(miss)
+            else:
+                text += " (no missing values)"
```
janezd: These five lines repeat twice. Could you construct this part of the string before`if` (but only if data has class of any kind)?

Orange/widgets/data/owfile.py
```diff
@@ -385,19 +385,36 @@ def _describe(self, table):
...
         elif domain.has_discrete_class:
-            text += "<br/>Classification; categorical class with {} values.".\
+            text += "<br/>Classification; categorical class with {} values".\
                 format(len(domain.class_var.values))
+            if table.has_missing_class():
+                miss = "%.1f" % table.get_nan_percentile_class()
+                text += " ({}% missing values)".format(miss)
```
janezd: Why not something like

    text += " ({:.1f}% missing values)".format(table.get_nan_percentile_class())

Or

    miss = table.get_nan_percentile_class()
    text += " ({:.1f}% missing values)".format(miss)

, so that we don't format the string in two lines in two different ways.

Orange/widgets/data/owfile.py
```diff
@@ -385,19 +385,36 @@ def _describe(self, table):
...
+            else:
+                text += " (no missing values)"
         elif table.domain.class_vars:
             text += "<br/>Multi-target; {} target variables.".format(
                 len(table.domain.class_vars))
```
janezd: If you extract the above five lines, you could also report the percentage of unknowns for this case (there's no reason why this case should be different).

Orange/data/table.py
```diff
@@ -1065,10 +1065,20 @@ def has_missing(self):
...
+
     def has_missing_class(self):
         """Return `True` if there are any missing class values."""
         return bn.anynan(self._Y)
 
+    def get_nan_percentile_attribute(self):
+        return np.isnan(self.X).sum() / self.X.size * 100
+
+    def get_nan_percentile_class(self):
+        return np.isnan(self._Y).sum() / self._Y.size * 100
```
janezd: I think it would be better if these two functions reported relative frequencies, not percentages. That is, I would remove `* 100`. Percentages are more "human-friendly", while these two functions look general and could also be used in other, non-GUI-related code.

Orange/data/table.py
```diff
@@ -1065,10 +1065,20 @@ def has_missing(self):
...
+    def has_missing_attribute(self):
+        """Return `True` if there are any missing attribute values."""
+        return not sp.issparse(self.X) and bn.anynan(self.X)    # do not check for sparse X
+
     def has_missing_class(self):
         """Return `True` if there are any missing class values."""
         return bn.anynan(self._Y)
 
+    def get_nan_frequency_attribute(self):
+        return np.isnan(self.X).sum() / self.X.size
```
janezd: Sorry, I just noticed one more: could you return `0` if `self.X.size` is 0, to avoid division by zero? Same in the next function?

Orange/widgets/data/owfile.py
MaxGreil: I don't understand what you have in mind here. Which `if` do you mean and how should I then construct the string before it?

janezd:
Thank you both. I also added some tests.

PR #3310, opened 2018-10-12
[RFC] color palette: Change default colors

VesnaT:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
I'm sorry, but I think these colors are quite bad. :( I don't think they can be separated well visually, since they are so dark/muted.
I don't know if these are really improvement over our current palette.

markotoplak:
@VesnaT, could your reference the color palette? I think something like this belongs in a comment. :)

VesnaT:
I've used one of suggested palettes in http://colorbrewer2.org/#type=qualitative&scheme=Accent&n=3

ajdapretnar:
Ugh, these palettes are.... um.... interesting.
Could we not go with http://colorbrewer2.org/#type=qualitative&scheme=Set1&n=3 since it is the most similar to what we have now?

ajdapretnar:
Is the first one really color blind safe? It looks like it might not be... 3-class Dark2 seems to be...

markotoplak:
What is the number of colors we need in our color palettes so that our widgets do not work strangely on data with lots of labels? The old palette was larger.

No palette on that site is marked as colorblind safe for more than 4 colors, but yes, we should use one that is safe for the first few colors at least.

kernc:
:bulb: We could use two sets of colors and a Options toggle "Use colorblind-safe color palette" which would, upon loading the app, switch `colorpalette.DefaultRGBColors` palette for the weird one.

---
[This Bokeh palette](https://bokeh.pydata.org/en/latest/docs/reference/palettes.html#usability-palettes) looks safe for 7 colors.

lanzagar:
We currently have a limit of 10 different items/colors in legends + 1 grey for all others.
I think the number of unique colors in our default palette should be very high on the priorities list - ideally 10 so we don't need to decrease the limit, and if not, then as close to that as possible.
Keep in mind that we are destroying data/information, when the limit is exceeded (e.g. analyzing >10 countries, the plots would merge some of them on visualizations).

If the first few (3-4) colors can be colorblind safe in the default palette that would be good enough IMHO. If not, we can have a setting in Options to toggle to a special colorblind-safe one, like @kernc suggested.
In the link @VesnaT provided, there are only 2 schemes (_Paired_, _Set3_) that support 10 colors. The first, _Paired_, is colorblind safe for 4 classes, but I personally don't like it since the first two colors are light and dark blue (and the next two are both green), so commentating a visualization becomes very annoying.

Unless someone finds some better color palettes (that are not on this site), I suggest we have a colorblind unsafe default (_Set3_) and an option to switch to a colorblind safe one (_Paired_).
We could maybe even bump the red color in _Set3_ from the fourth to the second position, making the first two kind of similar to current ones? And I will not try to explain why _Set3_ is better than what we currently have, but someone can try :)
(In case this discussing does not come to a conclusion soon, we can still just add an alternative colorblind safe option of _Paired_)

markotoplak:
For reference, here are the choices we have and our original schema on _zoo_. I did not include _Paired_ because we agree it is not a good default.

mo
mset1
mset2
mset3
so
sset1
sset2
sset3

markotoplak:
_Set1_ supports 7 colors, _Set2_ supports 8, and _Set3_ eleven (+ each has another gray tone).

_Set3_ has lots of colors that are very light, which works for maps, but doesn't do well for thin elements (like lines or small dots) on white.

For me, it is harder to decide between _Set1_ and _Set2_. _Set2_ is colorblind safe for 3 colors and Agnieszka said it looks better.

ajdapretnar:
I strongly vote for having standard _and_ colorblind option available. None of the new palettes look good to me. Stylistically, Set2 is best, I agree, but it is hard to see instantly what the visualization communicates (content), mostly because of the paleness of colors. Set2 is also hard to describe, because I don't think many people know what teal shade is. The one we have now is easy to refer to and has strong, striking colors than non-colorblind people can easily tell apart.

janezd:
I strongly agree with the strong opinion of @ajdapretnar. While everything should be accessible by wheelchairs, staircases are still optimal for all who can use them. I find the original palette more appealing and would keep it as the default, but make it clear it's possible to switch to a color-blind palette.

Visualization widgets *could* have a palette icon in the status bar, beside help and report. The problem, though, is that palettes are defined by `Variable` (see, e.g. https://github.com/biolab/orange3/blob/master/Orange/widgets/visualize/utils/widget.py#L217 in projections), so it's difficult to switch after the data is already loaded. Unless we'd like to implement it as an override (uglifying the existing code in all widgets), this can only be a global preference setting.

janezd:
@ajdapretnar raise another concern: you should be able to name the colors. I would have to lecture with set 2 and call the colors "yellowish" and "blueish" to start a discussion about whether they are actually "orangish" and "purplish". Set 1 is somewhat better in this respect, but it looks aggressive in comparison with the nice default we currently have.

markotoplak:
I think it is easier to refer to _Set1_ than to the current one: it only has one shade of each color. For the _zoo_, we currently see "blue" and "light blue".

_Set1_ and _Set2_ have another advantage to our current color set: their first 5 colors are equally dark so they seem equally important.

janezd:
After some consideration and discussion we decided to seek expert help and to probably implement it as an option instead of enforcing this palette.

PR #3311, opened 2018-10-12
[FIX] Save data widget crash on no data

### LINKED ISSUE #3256: Save Data crashes on no data
##### Orange version
3.16, caused by #3147 

##### Actual behavior
Error that is_spare does not exist for None object.

Connect and disconnect a file to the File widget. You can also connect an empty selection.

golbog:
##### Issue
Fixes #3256.

##### Description of changes
Does not crash on empty data anymore.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


stuart-cls:
Thanks. Should there be a test added as well so this can't happen to other widgets?

Orange/widgets/data/owsave.py
```diff
@@ -126,18 +126,26 @@ def adjust_label(self):
 
     @Inputs.data
     def dataset(self, data):
-        self.data = data
         self.save.setDisabled(data is None)
         self.save_as.setDisabled(data is None)
-        if data is not None:
-            self.save_file()
+        if data is None:
+            return
```
janezd: `self.data = data`? Or, equivalently, `self.data = None`?

Orange/widgets/data/owsave.py
```diff
@@ -126,18 +126,26 @@ def adjust_label(self):
...
 
-        self.controls.filetype.clear()
-        if self.data.is_sparse():
-            self.controls.filetype.insertItems(0, [item for item, _, supports_sparse in FILE_TYPES
-                                                   if supports_sparse])
-        else:
+        if (self.data is None and data.is_sparse()) or (
```
janezd: These conditions do not cover the case `self.data is None and not data.is_sparse()`.

Connect some sparse data to the widget. This will remove the formats that do not support sparse data. Then remove the signal to set `self.data` to `None`. Then connect  some non-sparse data. After this, the widget will (incorrectly) show only formats for sparse data.

janezd:
Could you squash the commits?

Orange/widgets/data/owsave.py
golbog: I can't seem to replicate this bug. 
I've tested like you described here, but it shows correct formats.

janezd: Have you added `self.data = data` before line 132? This line is needed to ensure that `self.data` stores the current data and just not the last data seen by the widget. At the moment it may not matter, but it may introduce bugs in the future by defying expectations of what is in `self.data`.

I still don't understand the purpose of the change. However, even if there is/was no bug, the code is overly complicated and repetitive. I suppose the intention was to update the combo only when the sparsity of the new data is different from the previous. This translates to `self.data is None or self.data.is_sparse() != data.is_sparse()` (condition `data is None` is take care of earlier). This makes the code more readable and half shorter:

        if self.data is None or self.data.is_sparse() != data.is_sparse():
            self.controls.filetype.clear()
            allowed = [item for item, _, supports_sparse in FILE_TYPES
                       if supports_sparse or not data.is_sparse()]
            self.controls.filetype.insertItems(0, allowed)
            self.filetype = allowed[0]
            self.update_extension()

Orange/widgets/data/owsave.py
```diff
@@ -126,18 +126,24 @@ def adjust_label(self):
...
-        if self.data.is_sparse():
-            self.controls.filetype.insertItems(0, [item for item, _, supports_sparse in FILE_TYPES
-                                                   if supports_sparse])
+        if data.is_sparse():
+            items = [item for item, _, supports_sparse in FILE_TYPES
+                     if supports_sparse]
         else:
-            self.controls.filetype.insertItems(0, [item for item, _, _ in FILE_TYPES])
+            items = [item for item, _, _ in FILE_TYPES]
+        if not items == [self.controls.filetype.itemText(i)
```
janezd: `if items != ...`

Orange/widgets/data/owsave.py
```diff
@@ -126,18 +126,24 @@ def adjust_label(self):
...
-        self.controls.filetype.clear()
-        if self.data.is_sparse():
-            self.controls.filetype.insertItems(0, [item for item, _, supports_sparse in FILE_TYPES
-                                                   if supports_sparse])
+        if data.is_sparse():
+            items = [item for item, _, supports_sparse in FILE_TYPES
+                     if supports_sparse]
         else:
-            self.controls.filetype.insertItems(0, [item for item, _, _ in FILE_TYPES])
+            items = [item for item, _, _ in FILE_TYPES]
```
janezd: OK, that's one way to do it. It may be a matter of opinion, but isn't

```
items = [item for item, _, supports_sparse in FILE_TYPES
         if supports_sparse or not data.is_sparse()]
```

more concise and shorter than if's with separate comprehensions? :)

PR #3313, opened 2018-10-12
utils.updates_disabled: Make it a decorator as well

janezd:
##### Issue

Disabling updates may be useful for widgets that make a lot of subsequent changes, for instance putting a lot of labels on a pyqtgraph plot. Existing `updates_disabled` is a context manager, yet in practice we can often disable updates for entire methods. It is more practical to do so with decorators -- if for no other reason, it reduces the indentation.

##### Description of changes

The function is changed into a class with appropriate methods for context managing and calling (as decorator).

**Question:** Does this belong to Orange.**canvas**.gui.utils or is there a better place outside canvas? Widgets should (strive to) not import things from the canvas.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/canvas/gui/utils.py
+++ b/Orange/canvas/gui/utils.py
@@ -7,6 +7,7 @@ import sys
 import traceback
 
 from contextlib import contextmanager
+from functools import wraps
 
 from AnyQt.QtWidgets import (
     QWidget, QMessageBox, QStyleOption, QStyle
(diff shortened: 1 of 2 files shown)


janezd:
When this PR is merged, remove it from graphview.py (note: this module may be renamed to something else by then) in the network add-on.

Orange/canvas/gui/utils.py
```diff
@@ -21,16 +22,46 @@
 QWIDGETSIZE_MAX = ((1 << 24) - 1)
 
 
-@contextmanager
-def updates_disabled(widget):
-    """Disable QWidget updates (using QWidget.setUpdatesEnabled)
+# Serves as a decorator and context manager, hence a function-like name
+# pylint: disable=invalid-name
```
kernc: Put on the next line, otherwise it disables for the whole indentation level (i.e. module).

ales-erjavec:
> Disabling updates may be useful for widgets that make a lot of subsequent changes, for instance putting a lot of labels on a pyqtgraph plot.

From what I can tell `setUpdatesEnabled` should have no effect on the pyqtgraph plot (i.e. Qt Graphics View framework).

ales-erjavec:
> Won't `setUpdatesEnabled(False)` stop any updates of any widget, including Qt graphics, so pyqtgraph won't have to (re)paint the content on every update of the scene? (https://github.com/qt/qtbase/blob/5.11/src/widgets/kernel/qwidget.cpp#L11106)

Widgets in general do not paint except on event loop scheduled coalesced `paintEvents`.
Only if explicit `repaint` method is called is the painting forced immediately.

janezd:
Some visualizations, such as MDS and Networks update the plot in a separate thread. Doesn't the event loop of the widget then run in parallel and can handle an update event at an inappropriate moment?

The problem I think I encountered was that the Network widget moved nodes and edges, and the update happened in between. Worse still, I suspect that update sometimes happened when the `ScatterPlotItem` was in some inconsistent state within `setData` (the `'targetRect'` column of `ScatterPlotItem`'s `data` was set to `None` because it was being recomputed), so paint crashed. I had to do this to fix the problem:

```
class ScatterPlotItem(pg.ScatterPlotItem):
    def __init__(self, *args, **kwargs):
        self.lock = threading.Lock()
        super().__init__(*args, **kwargs)

    def paint(self, painter, option, widget=None):
        with self.lock:
            super().paint(painter, option, widget)

    def setData(self, *args, **kwargs):
        with self.lock:
            super().setData(*args ,**kwargs)
```

ales-erjavec:
You should not do that. The Graphics View framework is not thread safe. Only update the from the main GUI thread.

ales-erjavec:
MDS is single threaded. It does all work on the main GUI thread.

janezd:
I see. Thanks for this explanation. I've fixed the network widget to update from the main thread, and I'll remove locks from ScatterPlotItem..

Can we ever need this context/decorator in Orange then? Should I remove it altogether to avoid future confusions? It's used only once, as far as I see.

PR #3317, opened 2018-10-13
[WIP] Projections: Clean-up

janezd:
After reimplementing projections widgets in #3089 and #3260, a lot of old classes (like multiple implementations of legends etc) remained lying around. Naming of old and new classes is also rather bad. This PR will clean this up.

##### Description of changes

- The graph creates an instance of `OWPlotGUI.OWPlotGUI`, which is then used exclusively by the widget. The only exception is `box_zoom_select`, which is a method of the graph although other methods of this kind are in `OWPlotGUI`. https://github.com/biolab/orange3/commit/b66894423a324af48128bc689b14b5d287cec932 moves `box_zoom_select` to `OWPlotGUI` and the instance of `OWPlotGUI` to widget.

- ScatterPlotItem had locks to prevent setData during paint and vice versa. As @ales-erjavec explained in #3313, Qt Graphics framework is not thread safe and should be updated from main thread only, hence the lock is redundant. This PR removes it.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/plot/owplotgui.py
+++ b/Orange/widgets/utils/plot/owplotgui.py
@@ -28,10 +28,8 @@ This module contains functions and classes for creating GUI elements commonly us
 import os
 
 import unicodedata
-from functools import reduce
-from operator import itemgetter
 
-from Orange.data import ContinuousVariable, DiscreteVariable, Variable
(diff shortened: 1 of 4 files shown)


janezd:
Split into separate pull requests.

PR #3323, opened 2018-10-19
Test and Score: Warn about transformation, raise error if all is nan

janezd:
##### Issue

- Domain transformations in Test and Score (and Predictions?) are implicit. Users should be informed about them.

##### Description of changes

- Test and Score shows this exception as a user-readable error (without mentioning "domains")
- If train and test data have different domains but transformation does not result in all-nans, an information about the transformation appears.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -9,6 +9,7 @@ import scipy
 
 from Orange.data import Table, Storage, Instance, Value
 from Orange.data.filter import HasClass
+from Orange.data.table import DomainTransformationError
 from Orange.data.util import one_hot
 from Orange.misc.wrapper_meta import WrapperMeta
 from Orange.preprocess import Continuize, RemoveNaNColumns, SklImpute, Normalize
@@ -246,6 +247,13 @@ class Model(Reprable):
             if isinstance(data, Instance):
                 data = Table(data.domain, [data])
             if data.domain != self.domain:
+                if self.original_domain.attributes != data.domain.attributes \
+                        and data.X.size \
+                        and not np.isnan(data.X).all():
+                    data = data.transform(self.original_domain)
+                    if np.isnan(data.X).all():
+                        raise DomainTransformationError(
+                            "domain transformation produced no defined values")
                 data = data.transform(self.domain)
             prediction = self.predict_storage(data)
         elif isinstance(data, (list, tuple)):
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -32,6 +32,7 @@ from AnyQt.QtCore import pyqtSlot as Slot
 from Orange.base import Learner
 import Orange.classification
 from Orange.data import Table, DiscreteVariable, ContinuousVariable
+from Orange.data.table import DomainTransformationError
 from Orange.data.filter import HasClass
 from Orange.data.sql.table import SqlTable, AUTO_DL_LIMIT
 import Orange.evaluation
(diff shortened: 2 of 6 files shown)


janezd:
@markotoplak, something like this?

Orange/base.py
```diff
@@ -244,9 +245,17 @@ def __call__(self, data, ret=Value):
             prediction = self.predict(data)
         elif isinstance(data, (Table, Instance)):
             if isinstance(data, Instance):
-                data = Table(data.domain, [data])
-            if data.domain != self.domain:
-                data = data.transform(self.domain)
+                trans_data = Table(data.domain, [data])
+            elif data.domain != self.domain:
```
markotoplak: The first elif should stay as if. The domain of a single instance was not yet converted.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -209,6 +211,8 @@ class Error(OWWidget.Error):
         memory_error = Msg("Not enough memory.")
         no_class_values = Msg("Target variable has no values.")
         only_one_class_var_value = Msg("Target variable has only one value.")
+        test_data_incompatible = Msg(
+            "Test data may be incompatible with train data.")
```
markotoplak: may? :)

janezd: May. I don't want to be too categorical, it could already have had nan's before. But I can change "may be" to "is" if you think it's better. Your call.

markotoplak: I did not consider this... Keep it "may".

markotoplak:
@janezd, hmm, I really do not know. I'd go either for an exception (without the check) or a warning. Exception + check seems like a strange compromise...

janezd:
@markotoplak, I had to add some checks to prevent raising exceptions in some semi-legitimate cases.

markotoplak:
I tried writing tests and I noticed that for scikit-learn based learners this PR does not really work. For scikit-learn methods Orange adds some very aggressive preprocessors, which work hard to remove all nans. For example, with logistic regression, I can still build a learner on iris and apply its output to titanic.

I think preprocessing is so agressive to handle potentially missing columns in the test data. Ouch.

janezd:
We could still merge it because it works for non-skl methods, like Naive Bayes. And it will gradually work better as we gradually get rid of skl. :)

I mean, it doesn't hurt to have this check, although it doesn't always work.

To make it better, we could add a flag to preprocessors that would tell them to raise an exception if all data is `nan`. When using them for preprocessing training data, we'd set this flag. Since a model cannot be fit to a table of nans, raising such an exception would make sense, right?

markotoplak:
How about first transforming the data to the original domain, then checking for NaNs, and afterwards continuing with however the classifier wants to preprocess? I just tried this:

                # transform to the original domain and check if the data is compatible
                data = data.transform(self.original_domain)
                 if data.X.size \
                         and np.isnan(data.X).all() \
                         and not np.isnan(orig.X).all() \
                         and data.domain.attributes != orig.domain.attributes:
                     raise DomainTransformationError(
                         "domain transformation produced no defined values")
                # apply transformations added by the learner
                data = data.transform(self.domain)

Surprisingly, all tests pass. Perhaps we could then even remove some conditions, but I did not look well into it.

Do you see any potential problems with this approach?

janezd:
Thanks. Fixed as suggested, except that I don't transform the data to the original domain if this cannot raise an exception.

markotoplak:
I added some tests. I think it is ready to merge. Please check the tests, Janez.

janezd:
Thanks for the tests. I think they're OK. If you agree with the code, you can merge.

PR #3335, opened 2018-10-26
[WIP] OWRules: fix default rule

rokgomiscek:
##### Issue
Sometimes the default rule would not classify to majority class.
##### Description of changes
When no more rules are found, the default rule classifies to majority class.
##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


rokgomiscek:
Fixed. Tests pass now.

PR #3336, opened 2018-10-26
[WIP] [FIX] OWMergeData removes table meta attributes

### LINKED ISSUE #3289: Merge Data removes table meta attributes
##### Orange version
3.16

##### Expected behavior
Merge data should retain table meta attributes (like taxonomy id) from the primary Data Table.

##### Actual behavior
Merge data throws away table meta attributes. 

Run this workflow: [test.ows.zip](https://github.com/biolab/orange3/files/2449733/test.ows.zip) and compare the two Data Infos.
(Requires scOrange).

AndrejaKovacic:
##### Issue

Fixes #3289 

##### Description of changes
I added an optional parameter in table.py, method from_numpy can now accept data_attributes.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Could you also add the argument `data_attributes` to `Table.from_list`, for the sake of consistency? Alternatively, you can remove the argument from `from_numpy` and just attach attributes in the widget (instead of passing them as an argument), that is, without modifying Orange.data.table. I have no preference.

Please remove the unwanted unrelated change in application.py. :)

Also, please squash the commits into a single one.

lanzagar:
Well, in addition to `from_list` there are `from_domain`, `from_table`, `from_table_rows` above and none have this parameter. So my preference is to leave the Table API unchanged and fix the widget instead.
And looking at how the tables are merged, I would say copying the attributes is not the only thing missing. At least the fact that `id`s are not retained is also a problem. Using `from_numpy` has been problematic in other cases (many of them now fixed) and is not a good choice here as well. Appending columns from extra_data to some primary data set should probably retain as much as possible from the main dataset (including `id`s and `attributes`).

AndrejaKovacic:
Replaced by  #3474.

PR #3339, opened 2018-10-27
[FIX] Louvain Clustering: Do not invalidate output on PCA slider change with apply disabled

### LINKED ISSUE #3334: Louvain clustering reruns on every PCA components change
Louvain clustering re-runs every time the PCA components slider is changed, even if Apply PCA preprocessing is disabled.

##### Orange version
master

##### Expected behavior
Only re-run if Appy PCA processing is enabled.

1. Uncheck Apply PCA preprocessing
2. Check Apply automatically
3. Change PCA components slider

pavlin-policar:
##### Issue
Fixes #3334

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owlouvainclustering.py
+++ b/Orange/widgets/unsupervised/owlouvainclustering.py
@@ -136,6 +136,9 @@ class OWLouvainClustering(widget.OWWidget):
 
     def _invalidate_pca_projection(self):
         self.pca_projection = None
+        if not self.apply_pca:
+            return
+
         self._invalidate_graph()
         self._set_modified(True)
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/unsupervised/owlouvainclustering.py
```diff
@@ -135,6 +135,9 @@ def __init__(self):
         )  # type: QWidget
 
     def _invalidate_pca_projection(self):
+        if not self.apply_pca:
+            return
+
         self.pca_projection = None
```
ales-erjavec: This line should be moved before the check and early return. Right? Toggling the apply_pca will only invalidate the partial results starting with the graph.

pavlin-policar: You're right of course.

PR #3343, opened 2018-10-29
travis: Update test matrix

### LINKED ISSUE #3341: Remove tests (and support) for Python 3.4?
We discussed that we could stop supporting Python 3.4,  rather than fixing the reasons for the failed test.

Even disregarding the failed test, I propose we stop testing on 3.4 and start testing on 3.7. @lanzagar?

lanzagar:
Updates the travis test matrix as follows:

Remove travis testing for Python 3.4.
Py3.5 is used only in conjunction with PyQt4 in a single "legacy" test.
The main testing is done using Py3.6.
There is also a Py3.7 test, which requires some additional travis parameters (see https://github.com/travis-ci/travis-ci/issues/9815).

Fixes #3341

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,6 @@ matrix:
           env: PYQT4=true
         - python: '3.6'
           env: UPLOAD_COVERAGE=true
-        - python: '3.7'
-          sudo: required
-          dist: xenial
     fast_finish: true
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,18 @@ dist: trusty
 matrix:
     include:
         - env: RUN_PYLINT=true
-          python: '3.4'
+          python: '3.6'
           cache: {}
         - env: BUILD_DOCS=true
-          python: '3.5'
+          python: '3.6'
           script: source $TRAVIS_BUILD_DIR/.travis/build_doc.sh
-
-        - python: '3.4'
-          env: PYQT4=true
         - python: '3.5'
-          env: UPLOAD_COVERAGE=true
+          env: PYQT4=true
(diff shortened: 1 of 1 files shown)


janezd:
There's apparently a new failing test on all versions on travis. I haven't checked what it actually does, but does it belong to this PR, too?

```
FAIL: test_credential_manager (test_credentials.TestCredentialManager)
  File "/home/travis/build/biolab/orange3/build/travis-test/Orange/widgets/tests/test_credentials.py", line 15, in test_credential_manager
    self.assertEqual(cm.key, 'Foo')
AssertionError: None != 'Foo'
```

It would be great if this PR is merged ASAP, those red crosses are frustrating.

lanzagar:
PyQt4: I don't mind dropping that too, I just wanted this to be a quicker fix for removing py3.4 and not wait for a (possibly) longer discussion about pyqt
If there are no problems dropping support for it, I suggest to just delete the py3.5&pyqt4 test.

The keyring problem looks to be independent of this and is probably due to the new version released 2 days ago. I updated it on my computer too, but could not reproduce the failed test - everything still works on my computer, so it looks like it is just travis that has problems with it?

There are however some additional errors in the py3.7 tests. I didn't find out what is causing them yet (again, I am running py3.7 for some time, but cannot reproduce the problems locally).
If we want to merge this PR quickly, I suggest someone just drops the py3.7 test from this until we can get it to work.
This can than be merged (with just the py3.6 tests). However, to finally get everything green again, another PR needs to address the keyring thing.
I am away from computer until next week, so feel free to update this PR and merge it in the meantime.

janezd:
🥇

Issue #2019, opened 2017-02-15
Consider including documentation in built wheels


astaric:
Like it is done in example add-on:

https://github.com/biolab/orange3-example-addon/pull/4

PR #3345, opened 2018-10-29
[ENH] Install widget help files

ales-erjavec:
##### Issue

The widget html help files should be available offline on installation.

Fixes gh-2019

##### Description of changes

Install widget help files by default. This can be overridden by a --without-htmlhelp cmd line flag.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
After merging this the BUILD_DOCS test on travis started failing.
I see it was my fault for being to eager to merge - not really sure why I didn't rerun the tests, but it probably looked like just the old tests were failing which were later removed. It turns out there is an actual problem with building the docs that needs to be investigated.

Issue #3342, opened 2018-10-29
Data transformation using PCA and FreeViz


BlazZupan:
##### Motivation

Given the training data, we would like to develop a data projection or embedding model to be used on separate test data. The present set of widgets only supports transformations of the training data.

##### Requirements

Develop a widget called Transforms with input channels "Data" (Orange.data.Table) and "Preprocessor" (Orange.preprocess.preprocess.Preprocess) and the output channel "Transformed Data" (Orange.data.Table). The widget applies preprocessor on the input data. It has only a simple GUI:

![screen shot 2018-10-29 at 09 57 05](https://user-images.githubusercontent.com/726604/47639235-0fe86800-db61-11e8-8404-3d14424e6df0.png)

In the case of no data, write "No data on input.". In the case of no preprocessor, write "No preprocessor on input.".

The primary use of this widget is with transformation and embedding models. Change both PCA and FreeViz widgets so that both include an adequate Preprocessor output channel. 

##### Vision

To be used with MDS, UMAP, and t-SNE as well. We need to issue additional pull requests to cover these as well.

PR #3346, opened 2018-10-30
OWTransform: Add new widget

VesnaT:
##### Issue

Implements #3342

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/data/owtransform.py
```diff
@@ -0,0 +1,105 @@
...
+    def set_preprocessor(self, preprocessor):
+        self.preprocessor = preprocessor
+
+    def handleNewSignals(self):
+        self.apply()
+
+    def apply(self):
+        self.transformed_data = None
+        if self.data is not None and self.preprocessor is not None:
+            self.transformed_data = self.preprocessor(self.data)
```
janezd: Preprocessor could raise an exception here. I think the widget has to catch and report it. Maybe something like "An error occurred while transforming data", and the second line (visible only on tooltip) would contain the exception message.

BlazZupan:
Works as intended. Therefore merge. The documentation and subsequently a blog, once the widget enters the distribution, are missing. Also, for testing, could we release both widgets (PCA, Transform) into Prototypes (change name to, say, Test PCA, and Test Transform)?

PR #3352, opened 2018-11-03
Edit Links window redesign

### LINKED ISSUE #3258: Responsive links and boxes in Edit Links dialog
##### Orange version
3.15.0

##### Expected behavior
Edit Links dialog opens on double click on a link between two widgets. A typical dialog shows the two widgets, the channels they implement, and the established link:

screen shot 2018-09-17 at 15 52 41

In this dialog, the user can click on the link to remove it. For the starters, it is hard to know that this action is available. Mouse over link does not change anything, and it would be nice if on mouse over, the link would "react", say, with shadowing the lines in a similar way that the links react in the workflows upon mouseover.

In this dialog, the user can make new links by drawing lines from an output box of left widget to the output box of the widget on the right. Mouse over box makes the box react (shadow), but perhaps the effect should be stronger. Also, when drawing the line to the other side (the compatible box of another widget), hitting that box anywhere should anchor the link to the right position, and the box should react in some way. Compare this current behavior to correctly implemented interface in canvas where connecting the two widgets.

irgolic:
Resolves #3258 

##### Description of changes
- Enlarges mouse hit area of link
- Highlights link on hover (shadow, color change)
- Disables channel box if there are no viable links
- Disabled non-viable channel boxes on start line drag (try Polynomial Classification (Educational Add-on) -> Test & Score)
- Nicer box hover effect

preview

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


irgolic:
What should be displayed in the tooltips?
On hover over a channel, it shows the channel type in this format:
```
Orange.data.table.Table
```
I want disabled channels to have on-hover text saying 'No possible link to other channels'. I'm just adding it next to the current tooltip for now.

ajdapretnar:
I haven't noticed the tooltips before. 😮 Are these really useful for the GUI version of Orange? What are they communicating to the user? I don't think they say much. I would prefer just keeping the 'No possible link to other channels' tooltip. Actually, if the tooltip appears over a certain option, perhaps 'Link not possible' might be enough?

BlazZupan:
Works nice. Please consider changing the hover messages to:

- No compatible output channel.
- No compatible input channel.
- Click and drag to connect widgets.

ales-erjavec:
If I click on an anchor without starting a drag, the incompatible channels remain disabled on the mouse release.

Orange/canvas/document/editlinksdialog.py
```diff
@@ -782,3 +864,74 @@ def setDocument(self, doc):
...
+        self.setGraphicsEffect(self.__shadow)
+        self.prepareGeometryChange()
+        self.__shadow.setEnabled(False)
+
+    @overload
+    def setLine(self, line: QLineF) -> None:
+        pass
+
+    @overload
+    def setLine(self, x1: float, y1: float, x2: float, y2: float) -> None:
```
ales-erjavec: Such use of `overload` is only valid since python 3.5.2 (note the micro release). It raises on error on Python<=3.5.1. I would avoid its use or put it in a `if False:` block.

Orange/canvas/document/editlinksdialog.py
```diff
@@ -695,23 +768,32 @@ def setChannel(self, channel):
...
+            self.setBrush(self.enabledBrush)
+        else:
+            self.setBrush(self.disabledBrush)
+
+    def paint(self, painter, option, widget=None):
+        QGraphicsRectItem.paint(self, painter, option, widget)
+        # if disabled, draw X over box
+        if not self.isEnabled():
+            painter.drawLine(0, 0, int(self.rect().width()), int(self.rect().height()))
+            painter.drawLine(int(self.rect().width()), 0, 0, int(self.rect().height()))
```
ales-erjavec:
```python
painter.drawLine(self.rect().topLeft(), self.rect().bottomRight())
painter.drawLine(self.rect().topRight(), self.rect().bottomLeft())
```
ChannelAnchor is a QGraphicsRectItem, it does not necessarily have (0, 0) as its top right corner.

Orange/canvas/document/editlinksdialog.py
ales-erjavec: Never mind. It seems we are on longer even syntax compatible with Python 3.5 (since gh-3403).

Orange/canvas/document/editlinksdialog.py
```diff
@@ -310,7 +310,21 @@ def mouseMoveEvent(self, event):
...
+                    if clamped_pos.x() < source_widget_geometry.left():
+                        clamped_pos.setX(source_widget_geometry.left())
+                    elif clamped_pos.x() > sink_widget_geometry.right():
+                        clamped_pos.setX(sink_widget_geometry.right())
+                    if clamped_pos.y() < source_widget_geometry.top():
+                        clamped_pos.setY(source_widget_geometry.top())
+                    elif clamped_pos.y() > source_widget_geometry.bottom():
+                        clamped_pos.setY(source_widget_geometry.bottom())
+
+                    line.setP2(clamped_pos)
```
ales-erjavec: Maybe set a fixed displayed scene rect on the view instead?

```diff
diff --git a/Orange/canvas/document/editlinksdialog.py b/Orange/canvas/document/editlinksdialog.py
index c7f175775..60735a96a 100644
--- a/Orange/canvas/document/editlinksdialog.py
+++ b/Orange/canvas/document/editlinksdialog.py
@@ -118,6 +118,7 @@ class EditLinksDialog(QDialog):
         left, top, right, bottom = self.getContentsMargins()
         self.view.setFixedSize(size.toSize() + \
                                QSize(left + right + 4, top + bottom + 4))
+        self.view.setSceneRect(self.scene.editWidget.geometry())
```

Orange/canvas/document/editlinksdialog.py
janezd: Do these overloads have any effect? PyCharm and Pylint don't recognize them, and the redefinitions trigger `function-redefined` in pylint.

If they are necessary, please disable pylint warning. If they have no effect, please remove them.

irgolic: They're not necessary, but PyQT already overloads the method. I thought it would be most elegant to expose the same interface if I am overriding the method.

Orange/canvas/document/editlinksdialog.py
```diff
@@ -278,11 +277,13 @@ def mouseMoveEvent(self, event):
...
                 # Start a line drag
-                line = QGraphicsLineItem(self)
+                line = LinkLineItem(self)
                 start = self.__dragStartItem.boundingRect().center()
                 start = self.mapFromItem(self.__dragStartItem, start)
-                line.setLine(start.x(), start.y(),
-                             event.pos().x(), event.pos().y())
+
+                eventPos = event.pos()
+                # False positive; pylint: disable=too-many-function-args
```
ales-erjavec: @janezd 

This obstructs the readability/flow/interpretation of the code. When I read this I expect the comment to elucidate the following code. It is only at the middle that I realize I am on the wrong path and need to backtrack.

Put `# pylint: ...` at the beginning so the reader can bail out earlier. Although I would prefer to remove it altogether.

PR #3358, opened 2018-11-06
[FIX]OWFreeViz: Fix optimization for data with missing values

VesnaT:
##### Issue

The widget crashed when received dataset with missing values on the input.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owfreeviz.py
+++ b/Orange/widgets/visualize/owfreeviz.py
@@ -353,12 +353,12 @@ class OWFreeViz(OWAnchorProjectionWidget):
             self._X = self._Y = self.valid_data = None
             return
 
-        self._X = self.data.X.copy()
-        self._X -= np.nanmean(self._X, axis=0)
-        span = np.ptp(self._X[self.valid_data], axis=0)
+        self._X = self.data.X[self.valid_data]
+        self._X -= np.mean(self._X, axis=0)
+        span = np.ptp(self._X, axis=0)
         self._X[:, span > 0] /= span[span > 0].reshape(1, -1)
 
-        self._Y = self.data.Y
+        self._Y = self.data.Y[self.valid_data]
         if self.data.domain.class_var.is_discrete:
             self._Y = self._Y.astype(int)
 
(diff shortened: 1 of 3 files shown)


Orange/widgets/visualize/owfreeviz.py
```diff
@@ -370,9 +370,10 @@ def init_embedding_coords(self):
     def get_embedding(self):
         if self.data is None:
             return None
-        embedding = np.dot(self._X, self.projection)
-        embedding /= \
-            np.max(np.linalg.norm(embedding[self.valid_data], axis=1)) or 1
+        EX = np.dot(self._X, self.projection)
+        EX /= np.max(np.linalg.norm(EX, axis=1)) or 1
+        embedding = np.zeros((len(self.data), 2), dtype=np.float)
```
lanzagar: If the original data had missing values, and the embedding is not available for all points, I think it is better to assign missing values to those points as the embedding too (instead of zeros).

lanzagar: Also, while I doubt these projections will ever be generalized to more than 2D, I would still prefer to not make implicit assumptions like that in the code and use `EX.shape[1]` instead of `2`

lanzagar:
I see that @janezd was quicker than my comments... :)

janezd:
I apologize for being over-eager. :)

`np.zeros` bothered me, too, and I started commenting, but then I saw it's filtered out immediately afterwards, so I ignored it.

You're right about 2D. On the other hand, if it would say `EX.shape[1]`, it would make may wonder "Hey, isn't this always 2?!". `2` is more explicit and informative. Besides, Freeviz won't be generalized beyond 2D because it can't be plotted then (you'd need another projection over it). And if it was generalized, a lot of other code would have to be rewritten anyway.

lanzagar:
I know it is not going to be generalized. I just really dislike hardcoded constants in code (numbers, strings) when they can be avoided :)
But I guess both options have their arguments. Using shape might not make it immediately obvious what the final size will be. But you know that EX will fit into the container without dimension errors, by just looking at these 2 lines (no need to be familiar with the widget, freeviz algorithm, what self.projections is, ...)
But I wouldn't spend time over it, either way is acceptable.

However, I would change the zeros to nans, as I think that is conceptually more correct.

janezd:
I agree about nans.

I guess @VesnaT should make a change in a new PR. I promise I won't merge it. :)

PR #3360, opened 2018-11-06
[FIX] File widget gives reader error 

### LINKED ISSUE #3349: File widget gives "reader error" on empty URL
##### Orange version

Any recent.

##### Expected behavior

In the file widget, selecting URL when the line edit is empty should output None as data, but without reporting any error.

##### Actual behavior

The widget informs the user about the error when reading the data.

AndrejaKovacic:
##### Issue

Fixes #3349

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/application.py
+++ b/Orange/canvas/application/application.py
@@ -14,7 +14,8 @@ class CanvasApplication(QApplication):
     def __init__(self, argv):
         if hasattr(Qt, "AA_EnableHighDpiScaling"):
             # Turn on HighDPI support when available
-            QApplication.setAttribute(Qt.AA_EnableHighDpiScaling)
+            #QApplication.setAttribute(Qt.AA_EnableHighDpiScaling)
+            pass
         QApplication.__init__(self, argv)
         self.setAttribute(Qt.AA_DontShowIconsInMenus, True)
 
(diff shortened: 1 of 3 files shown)


janezd:
There is another situation where a similar bug could occur. If `self.source == self.LOCAL_FILE`, `_get_reader` assumes that `self.recent_paths` is not empty. Initially it indeed contains files from documentation, but these can be later replaced by user files. If those are removed from disk, Orange will, I assume, remove them from `self.recent_paths` so it will be empty. If this happens, `path` will be `None` and `reader_class(path)` will return `None`.

I suppose you need to add

```
if path is None:
    return self.NoFileSelected
```

after `path = self.last_path()`.

(This is also the reason why I called the singleton `NoFileSelected` and not `NoUrl`.)

ajdapretnar:
I would also kindly point out some redundant commits, which I suggest get sorted before merging. :)

janezd:
@ajdapretnar, you should work in diplomacy.

janezd:
Here's a test you can add.

```
    def test_nothing_selected(self):
        widget = self.widget = \
            self.create_widget(OWFile, stored_settings={"recent_paths": []})
        
        widget.Outputs.data.send = Mock()
        widget._try_load()
        widget.Outputs.data.send.assert_called_with(None)

        widget.Outputs.data.send.reset_mock()
        widget.source = widget.URL
        widget._try_load()
        widget.Outputs.data.send.assert_called_with(None)
```

AndrejaKovacic:
Replaced by  #3402.

PR #3362, opened 2018-11-07
[ENH] Correlations: Move from prototypes to core

VesnaT:
##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation


lanzagar:
Add Correlations widget (without input) -> press Start -> crash

PrimozGodec:
I think the button should be disabled when no input. It is disabled when I change the correlation but not when the widget is placed at the canvas.

PR #3365, opened 2018-11-07
Projections: Assign missing values to embedding

VesnaT:
##### Issue

If the original data had missing values, and the embedding is therefore not available for all points, missing values should be assigned to those points instead of zeros.

##### Description of changes
Assign np.nan-s to embedding instead of zeros.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owfreeviz.py
+++ b/Orange/widgets/visualize/owfreeviz.py
@@ -372,7 +372,7 @@ class OWFreeViz(OWAnchorProjectionWidget):
             return None
         EX = np.dot(self._X, self.projection)
         EX /= np.max(np.linalg.norm(EX, axis=1)) or 1
-        embedding = np.zeros((len(self.data), 2), dtype=np.float)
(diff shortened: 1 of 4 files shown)


Orange/widgets/visualize/owfreeviz.py
```diff
@@ -372,7 +372,8 @@ def get_embedding(self):
             return None
         EX = np.dot(self._X, self.projection)
         EX /= np.max(np.linalg.norm(EX, axis=1)) or 1
-        embedding = np.zeros((len(self.data), 2), dtype=np.float)
+        embedding = np.empty((len(self.data), 2), dtype=np.float)
+        embedding.fill(np.nan)
```
janezd: I guess `np.full((len(self.data), 2), np.nan)` does the same job as these two lines. Or is there a difference?

No need to change this PR, I'm just saying it for next time.

VesnaT: Have changed it anyway :)

PR #3393, opened 2018-11-13
[ENH] Pythagorean Tree: children order

### LINKED ISSUE #3330: Overlap of leaves in Pythagorean Tree
The leaves of the Pythagorean Tree overlap under certain conditions. I wonder if this is something that can be corrected for better visual representation or in these cases the options that produce the confusing output should be disabled?

##### Orange version

3.17.0.dev0+8f507ed

##### Expected behavior
Since this is a visual representation it would be important that the Pythagorean Tree leaves don't overlap.

##### Actual behavior
image

Dataset: heart_disease.tab

Workflow:
image

With other settings the output looks fine.
image

thocevar:
##### Issue
Fixes #3330.

##### Description of changes
Implemented two new features that should enable visualizations of Pythagorean trees with less overlap:
- Add a button to completely redraw the tree with a random left/right ordering of children (this should give a user a good starting point).
- Double click on a node reverses the order of its children (for manual fine-tuning).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owpythagorastree.py
+++ b/Orange/widgets/visualize/owpythagorastree.py
@@ -120,8 +120,7 @@ class OWPythagorasTree(OWWidget):
             box_plot, self, 'show_legend', label='Show legend',
             callback=self.update_show_legend)
 
-        button_redraw = gui.button(
-            self.controlArea, self, label="Redraw", callback=self.redraw)
+        gui.button(self.controlArea, self, label="Redraw", callback=self.redraw)
 
         # Stretch to fit the rest of the unsused area
         gui.rubber(self.controlArea)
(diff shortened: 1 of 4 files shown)


thocevar:
Further work:
- Make randomized/hand-crafted visualizations consistent between sessions.
- Generate a good visualization automatically.

thocevar:
Updating the scene rectangle repositions the tree in the current view. If we have a horizontal layout of the tree and flip something that results in a vertical layout, we would have to add some horizontal margin to keep the tree in place besides computing the right view parameters (such that the scene/view transformation keeps the items in place).

I resorted to updating the entire view with `_update_main_area`. It's not ideal as it moves the item that the user just clicked but I'll leave that as another issue.

PR #3402, opened 2018-11-16
[Fix] Missing url error

### LINKED ISSUE #3349: File widget gives "reader error" on empty URL

AndrejaKovacic:
##### Issue

Fixes  #3349

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -122,6 +122,7 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
 
     class NoFileSelected:
         pass
+
     def __init__(self):
         super().__init__()
         RecentPathsWComboMixin.__init__(self)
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owfile.py
```diff
@@ -120,6 +120,8 @@ class Error(widget.OWWidget.Error):
         sheet_error = widget.Msg("Error listing available sheets.")
         unknown = widget.Msg("Read error:\n{}")
 
+    class NoFileSelected:
+        pass
```
janezd: Add an empty line after this.

Orange/widgets/data/tests/test_owfile.py
```diff
@@ -176,6 +176,19 @@ def test_file_not_found(self):
         self.open_dataset("iris")
         self.assertFalse(self.widget.Error.file_not_found.is_shown())
 
+   def test_nothing_selected(self):
```
janezd: Tests don't pass because the indent is too small by one space. (Trick: first run tests on your machine. Not that I'm without a sin in this respect, though. :))

AndrejaKovacic:
Thank you for your comments. It seems that everytime i do something quickly to save time, it later takes more time to do it right.
The test should work now.

PR #3404, opened 2018-11-16
OWFile: Lint

janezd:
The only remaining warning now is `too-many-ancestors`, but this happens for most widgets.

##### Includes
- [X] Code changes


Orange/widgets/data/owfile.py
```diff
@@ -356,7 +363,7 @@ def _get_reader(self):
             if self.recent_paths and self.recent_paths[0].sheet:
                 reader.select_sheet(self.recent_paths[0].sheet)
             return reader
-        elif self.source == self.URL:
+        else:  # self.source == self.URL:
```
VesnaT: Same as above as far as comment is concerned.

Orange/widgets/data/tests/test_owfile.py
```diff
@@ -176,10 +176,12 @@ def test_file_not_found(self):
         self.open_dataset("iris")
         self.assertFalse(self.widget.Error.file_not_found.is_shown())
 
+
```
VesnaT: Lint doesn't complain, but PEP8 does. Could you please remove this line, since we are tidying up the files with this PR.

Orange/widgets/data/owfile.py
```diff
@@ -92,6 +93,11 @@ class Outputs:
         match_values=PerfectDomainContextHandler.MATCH_VALUES_ALL
     )
 
+    # pylint seems to want declarations separated from definitions?!
```
VesnaT: I think this comment is not needed and should be removed.

janezd: I'd keep it. It's not a complaint about pylint, it's here to explain why these lines are separated. Otherwise somebody could merge `recent_urls: List[str]` and `recent_urls = Setting([])` into `recent_urls: List[str] = Setting([])`, which is a legal way to do it, but pylint doesn't recognize it.

I removed `?!` though.

Orange/widgets/data/tests/test_owfile.py
janezd: Done. And pushed. Please be back when tests are finished. :)

PR #3407, opened 2018-11-21
[ENH] tSNE: Output preprocessor

VesnaT:
##### Issue

Implements third part of #3342

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/manifold.py
+++ b/Orange/projection/manifold.py
@@ -10,8 +10,10 @@ import fastTSNE
 
 import Orange
 from Orange.data import Table, Domain, ContinuousVariable
+from Orange.data.util import ComputeValueProjector
 from Orange.distance import Distance, DistanceModel, Euclidean
 from Orange.projection import SklProjector, Projector, Projection
+from Orange.projection.base import TransformDomain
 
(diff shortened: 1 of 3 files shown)


Orange/projection/manifold.py
```diff
@@ -360,15 +375,15 @@ def __call__(self, data: Table) -> TSNEModel:
 
         # The results should be accessible in an Orange table, which doesn't
         # need the full embedding attributes and is cast into a regular array
-        tsne_cols = [ContinuousVariable('t-SNE-%d' % (i + 1))
-                     for i in range(self.tsne.n_components)]
+        n = self.tsne.n_components
+        postfixes = ['x', 'y'] if n == 2 else list(range(n))
```
pavlin-policar: I don't like `list(range(n))`. While it would very rarely come up, it would imply that the variables are named `t-SNE-0`, `t-SNE-1`, `t-SNE-2`, ... I guess it is consistent with the current behaviour of Manifold learning, which outputs `C0`, `C1`. Honestly, I think `t-SNE-x` and `t-SNE-y` should be removed in favour of `t-SNE-1`, `t-SNE-2`, and if the axes are to be labelled with x and y, this should be done on the t-SNE widget level.

PR #3411, opened 2018-11-22
settings: Continue workflow after error in setting migration

### LINKED ISSUE #3399: Errors in settings migration block the workflow
##### Orange version
master

##### Expected behavior
If an error in setting or context migration happens (and it inevitably will, someday), a user should see our error window (and hopefully report the error), but afterwards he should still be able to use the workflow, perhaps with widgets that failed to migrate set to default settings.

##### Actual behavior
Currently the workflow is blocked. Nothing works, even widgets on parallel branches, until the problematic widget is deleted.

I can see blocking nodes in the log:

    Blocking nodes: {SchemeNode(description_id=Orange.widgets.visualize.owscatterplot, title='Scatter Plot', ...)}

Raise an Exception in `migrate_settings` or `migrate_context`.

This is how the workflow with a crash looks like. 

ss

VesnaT:
##### Issue

Fixes #3399

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -29,6 +29,7 @@ so they can be used alter. It should be called before widget starts modifying
 (initializing) the value of the setting attributes.
 """
 
+import sys
 import copy
 import itertools
 import os
@@ -462,7 +463,12 @@ class SettingsHandler:
     def _migrate_settings(self, settings):
         """Ask widget to migrate settings to the latest version."""
         if settings:
-            self.widget_class.migrate_settings(settings, settings.pop(VERSION_KEY, 0))
+            try:
+                self.widget_class.migrate_settings(
+                    settings, settings.pop(VERSION_KEY, 0))
+            except Exception:
+                sys.excepthook(*sys.exc_info())
+                settings.clear()
 
     def _select_provider(self, instance):
         provider = self.provider.get_provider(instance.__class__)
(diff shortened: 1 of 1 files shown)


Orange/widgets/settings.py
```diff
@@ -462,7 +463,11 @@ def initialize(self, instance, data=None):
     def _migrate_settings(self, settings):
         """Ask widget to migrate settings to the latest version."""
         if settings:
-            self.widget_class.migrate_settings(settings, settings.pop(VERSION_KEY, 0))
+            try:
+                self.widget_class.migrate_settings(
+                    settings, settings.pop(VERSION_KEY, 0))
+            except Exception:
+                sys.excepthook(*sys.exc_info())
```
astaric: @janezd, when context migration fails, the offending context is removed. Should settings also be cleared if they cannot be migrated? (settings.clear())

janezd: You are the only one who really knows the mechanics of settings. If they can't be migrated, they mustn't be used. If calling `settings.clear()` is the way to achieve this, then, yes, we should call `settings.clear()` here.

Orange/widgets/settings.py
```diff
@@ -241,9 +241,9 @@ def unpack(self, instance, data):
         data : dict
             packed data
         """
-        for setting, data, instance in self.traverse_settings(data, instance):
-            if setting.name in data and instance is not None:
-                setattr(instance, setting.name, data[setting.name])
+        for setting, _data, inst in self.traverse_settings(data, instance):
```
janezd: I guess it should be `data_`. But not this time. :)

PR #3417, opened 2018-11-24
Lint of imports throughout Orange

janezd:
Rebase of #2856 (force-push into closed PRs is treated as a new PR).

Most of the PR is just moving and removing imports.

Pylint > 2.0 no longer omits `unused-import` for import used in type comments, thus this PR also removes some `disable=unused-import`.

The important change is the introduction of `WidgetTestRunMixin` which contains the usual `main` function for widgets, so the "main" part of the file is reduced to, for instance,

```
if __name__ == "__main__":  # pragma: no cover
    OWSieveDiagram.test_run(Table("zoo"))
```

, which uses deafult signals, or 

```
if __name__ == "__main__":  # pragma: no cover
    data = Table("iris")
    OWScatterPlot.test_run(set_data=data, set_subset_data=data[:30])
```

, which sets the signal explicitly. Mixin is designed to be configurable, so further modifications are possible by overriding its [methods](https://github.com/biolab/orange3/blob/20928608176dd401020deb47a5673ad5269dc350/Orange/widgets/utils/test_run.py).

Besides removing the redundant repetitive code from widegt modules, the benefit of this change is to follow the proper procedure (including saving settings, tearing down the widgets...) in all widgets.

The use of the mixin is explained in documentation for widget developers (see [rst](https://github.com/biolab/orange3/blob/20928608176dd401020deb47a5673ad5269dc350/doc/development/source/testing.rst) and [docstrings in the code](https://github.com/biolab/orange3/blob/20928608176dd401020deb47a5673ad5269dc350/Orange/widgets/utils/test_run.py)).

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation


Orange/preprocess/__init__.py
```diff
@@ -6,10 +9,3 @@
 from .remove import *
 from .preprocess import *
 
-from Orange.misc.lazy_module import _LazyModule
-transformation = _LazyModule("preprocess.transformation")
-score = _LazyModule("preprocess.score")
-discretize = _LazyModule("preprocess.discretize")
-continuize = _LazyModule("preprocess.continuize")
-normalize = _LazyModule("preprocess.normalize")
-remove = _LazyModule("preprocess.remove")
```
janezd: `preprocess` is already a lazy module itself, so there's no point in making these lazy.

markotoplak:
I wanted to click merge but I'll ask: any reservations? If we do it, let's do it quckly.

lanzagar:
This looks great. Looking forward to not seeing so many different versions of testing widgets in `"__main__"` functions :)
Maybe @ales-erjavec can just double-check [test_run](https://github.com/biolab/orange3/pull/3417/commits/46dc21bfa77e5b3c8860ea28f7d46f61783d5191#diff-c62ee995715e72e50116d90c2d5fdf1f) to confirm that widgets are tore down and garbage collected as they should be etc.?

Also, before merging, can the lint test be saved (=made green) so we are sure that after so many changes the lint testing is still working properly? At the moment it seems to be timeouting somewhere in the middle.

janezd:
[test_run](https://github.com/biolab/orange3/pull/3417/commits/46dc21bfa77e5b3c8860ea28f7d46f61783d5191#diff-c62ee995715e72e50116d90c2d5fdf1f) contains everything I've seen in different widgets, but I'd appreciate, too, if @ales-erjavec can take a look.

I suspect that lint takes too much time because it has too many files to check. It works locally. Do you think it will help if we proceed (just for this PR?) the line `.travis/check_pylint_diff` in `.travis/stage_script.sh` with `travis_wait`? (See [this](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received).)

lanzagar:
I also assumed it was the large diff that caused lint to have to check so many files and make it time out. But after rerunning it, I saw that it got stuck in the same place and if it is not producing any output for more than 10min (which as far as I can tell should be the default) that seems suspicious. I can't imagine a single file to take that long to lint...
We can definitely try changing the waiting time for this PR. If it passes, we can change back, merge and hope it will work faster in the future. And if it still does not pass, then we at least know there is a problem.

janezd:
I tried with 20 minutes. It's not a timeout.

I updated pylint from 2.1 to 2.2.1, which we use on travis. Now it also crashes locally, (at least) on `pylint doc/development/source/orange-demo/orangedemo/*.py`. This however also happens on master, but usually doesn't cause problems because no PRs touch this file. Another example is Orange/projections/manifold.py (which wasn't changed in this PR).

It crashes in a checker for reporting external dependencies (RP0401), which we don't need anyway. I disabled it and pylint now works locally, but still timeouts on travis.

Orange/tests/test_sparse_reader.py
```diff
@@ -7,7 +7,7 @@
 
 import numpy as np
 
-import Orange.data._io as _io
+import Orange.data._io as _io  # pylint: disable=useless-import-alias
```
astaric: Why does it have to be imported like this? (and not from Orange.data import _io)

Orange/widgets/utils/tests/test_test_run.py
```diff
@@ -0,0 +1,90 @@
...
+
+# prevent test_run from calling sys.exit
+sys.exit = MagicMock()
+
+# prevent multiple initializations of QApplication
+app = QApplication([])
+test_run.QApplication = lambda *_: app
+
+# prevent entering the application's main loop
+app.exec_ = MagicMock()
```
ales-erjavec: Move all of the above to `setUpClass `and then undo the modification s in `tearDownClass`.

janezd:
> I do not think the `test_run` etc. should be members of OWWidget.

You're right. I put it there because of practicality, but now I moved it out and it's much better and more useful.

Orange/widgets/utils/tests/test_test_run.py
janezd: Moved. `app = QApplication([])` remains outside, I suppose?

PR #3432, opened 2018-11-30
Projections: Move OWPlotGUI instance from graph to widget

janezd:
After reimplementing projections, the reference to an instance of `OWPlotGui` remained on the plot although it is actually used by the widget. This PR moves it where it belongs.

Justification for the move is obvious from the diff. :)

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/utils/plot/owplotgui.py
+++ b/Orange/widgets/utils/plot/owplotgui.py
@@ -438,9 +438,9 @@ class OWPlotGUI:
 
     JITTER_SIZES = [0, 0.1, 0.5, 1, 2, 3, 4, 5, 7, 10]
 
-
-    def __init__(self, plot):
-        self._plot = plot
+    def __init__(self, master):
+        self._master = master
+        self._plot = master.graph
         self.color_model = DomainModel(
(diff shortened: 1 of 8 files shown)


Orange/widgets/utils/plot/owplotgui.py
```diff
@@ -439,7 +439,8 @@ class OWPlotGUI:
     JITTER_SIZES = [0, 0.1, 0.5, 1, 2, 3, 4, 5, 7, 10]
 
 
-    def __init__(self, plot):
+    def __init__(self, master, plot):
+        self._master = master
```
VesnaT: OWLinePlot crashes. We could (temporarily) use
```
    def __init__(self, master, plot=None):
        plot = plot or master
```
I don't like it, but I cannot think of any better solution since OWLinePlot is in another add-on. The hack works because the widget only uses OWPlotGUI for Zoom/Select toolbar

Orange/widgets/utils/plot/owplotgui.py
```diff
@@ -439,7 +439,8 @@ class OWPlotGUI:
...
         self._plot = plot
```
VesnaT: Why not `self._plot = master.graph`?

PR #3434, opened 2018-11-30
[ENH] Remove discrete attributes from scatter plot's axes

### LINKED ISSUE #3318: Why does Orange allow plotting discrete variables in scatter plots?!
I'm just preparing lectures and I'm thinking about how to persuade students against using scatter plots for showing discrete variables. Student often (ab)use the scatter plot instead of using more appropriate visualizations. But then I thought --- why does Orange allow it in the first place?!?!

A semi-useful case is to put the gender on y-axis and age on x-axis and see whether the age of women in our data is lower or higher or more spread out than the men's. However, what we actually observe in this case are distributions, and there's another, better widget for that (Distribution, Box plot). (And which does not require explaining away the jittering of genders.)

Plotting two discrete variables in a scatter plot is probably always a bad idea. If we plot, say, gender vs. four different types of chest pain, we project all data into just 2 x 4 = 8 points and then use jittering to spread them around these eight points. What kind of information can we get from this visualization? We can count the points around each of the few points (8, in above case). Or use the famous "overlap by size" feature to represent the counts by sizes --- which are known to be a bad way to visualize and compare magnitudes. This information is presented much better in Box plots, Distributions and Mosaic. This also goes for observing colors of jittered points; mosaic does it better.

Does anybody have a good argument for allowing discrete variables for scatter plots axes? Something that is not done much better with Distributions, Mosaic and Sieve? Aren't we promoting a bad practice here?

janezd:
##### Issue

Fixes #3318.

##### Includes
- [X] Code changes
- [X] Modified tests

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -158,7 +158,7 @@ class OWScatterPlot(OWDataProjectionWidget):
     class Outputs(OWDataProjectionWidget.Outputs):
         features = Output("Features", AttributeList, dynamic=False)
 
-    settings_version = 3
+    settings_version = 4
     auto_sample = Setting(True)
     attr_x = ContextSetting(None)
     attr_y = ContextSetting(None)
@@ -427,12 +427,15 @@ class OWScatterPlot(OWDataProjectionWidget):
 
     @classmethod
     def migrate_context(cls, context, version):
+        values = context.values
         if version < 3:
-            values = context.values
             values["attr_color"] = values["graph"]["attr_color"]
             values["attr_size"] = values["graph"]["attr_size"]
             values["attr_shape"] = values["graph"]["attr_shape"]
             values["attr_label"] = values["graph"]["attr_label"]
+        if version < 4:
+            if values["attr_x"][1] % 100 == 1 or values["attr_y"][1] % 100 == 1:
+                raise IncompatibleContext()
 
 
 if __name__ == "__main__":  # pragma: no cover
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -11,16 +11,15 @@ from AnyQt.QtGui import QColor
 
 import pyqtgraph as pg
 
-from Orange.data import Table, Domain, DiscreteVariable, Variable
+from Orange.data import Table, Domain, DiscreteVariable, Variable, \
+    ContinuousVariable
 from Orange.data.sql.table import SqlTable, AUTO_DL_LIMIT
 from Orange.preprocess.score import ReliefF, RReliefF
 
-from Orange.widgets import gui, report
+from Orange.widgets import gui
 from Orange.widgets.io import MatplotlibFormat, MatplotlibPDFFormat
 from Orange.widgets.settings import (
-    Setting, ContextSetting, SettingProvider
-)
-from Orange.widgets.utils import get_variable_values_sorted
+    Setting, ContextSetting, SettingProvider, IncompatibleContext)
 from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.visualize.owscatterplotgraph import OWScatterPlotBase
@@ -97,7 +96,6 @@ class ScatterPlotVizRank(VizRankDialogAttrPair):
 
 class OWScatterPlotGraph(OWScatterPlotBase):
     show_reg_line = Setting(False)
-    jitter_continuous = Setting(False)
 
     def __init__(self, scatter_widget, parent):
         super().__init__(scatter_widget, parent)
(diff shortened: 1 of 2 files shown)


VesnaT:
Until this PR the widget expected self.effective_variables are always set when self.data is not None. 
With this change, if there are not continuous variables in input data, self.effective_variables happens to be [None, None], which causes several errors.
Maybe you could add a check in check_data() (clearing the plot and showing an error message would be nice as well).
```
        if self.data is not None:
            if not self.data.domain.has_continuous_attributes(True, True):
            	self.Error.no_continuous_vars()
                self.data = None
```
I don't mind any other solution, just make sure (a potential test - that should probably be added) passes:
```
    def test_change_data(self):
        self.send_signal(self.widget.Inputs.data, self.data)
        self.send_signal(self.widget.Inputs.data, Table("titanic"))
```

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -122,37 +120,11 @@ def update_coordinates(self):
         self.update_regression_line()
         self.update_tooltip()
```
VesnaT: I think this line is no longer needed.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -231,7 +203,6 @@ def _add_controls(self):
         self._add_controls_axis()
         self._add_controls_sampling()
         super()._add_controls()
-        self.gui.add_widget(self.gui.JitterNumericValues, self._effects_box)
```
VesnaT: Is this property (JitterNumericValues) still needed?

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -465,10 +423,8 @@ def _get_send_report_caption(self):
             ("Label", self._get_caption_var_name(self.attr_label)),
             ("Shape", self._get_caption_var_name(self.attr_shape)),
             ("Size", self._get_caption_var_name(self.attr_size)),
-            ("Jittering", (self.attr_x.is_discrete or
-                           self.attr_y.is_discrete or
-                           self.graph.jitter_continuous) and
-             self.graph.jitter_size)))
+            ("Jittering", (self.graph.jitter_size > 0
+             and "{} %".format(self.graph.jitter_size)))))
```
VesnaT: The same method exists on the parent, so this one should be removed.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -428,19 +392,13 @@ def set_attr(self, attr_x, attr_y):
...
-        self.cb_reg_line.setEnabled(self.can_draw_regresssion_line())
         self.setup_plot()
         self.commit()
 
     def setup_plot(self):
         super().setup_plot()
-        for axis, var in (("bottom", self.attr_x), ("left", self.attr_y)):
-            self.graph.set_axis_title(axis, var)
-            if var and var.is_discrete:
-                self.graph.set_axis_labels(axis,
```
VesnaT: set_axis_labels() should probably be removed from OWScatterPlotGraph

Orange/widgets/visualize/owscatterplot.py
janezd: Good find!

Orange/widgets/visualize/owscatterplot.py
janezd: I'd dissolve OWPlotGui or at least clean it. I wrote an issue (#3446).

Orange/widgets/visualize/owscatterplot.py
janezd: Removed.

janezd:
Thanks for the fix and the test (I also added a few checks).

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -208,6 +172,7 @@ class Warning(OWDataProjectionWidget.Warning):
         missing_coords = Msg(
             "Plot cannot be displayed because '{}' or '{}' "
             "is missing for all data points")
+        no_continuous_vars = Msg("Data has not continuous variables")
```
VesnaT: Data has no continuous variables.

Orange/widgets/visualize/tests/test_owscatterplot.py
```diff
@@ -591,6 +577,15 @@ def assert_equal(data, max):
         data.Y[42] = np.nan
         assert_equal(data, MAX_CATEGORIES + 1)
 
+    def test_change_data(self):
+        self.send_signal(self.widget.Inputs.data, self.data)
+        self.send_signal(self.widget.Inputs.data, Table("titanic"))
+        self.assertTrue(self.widget.Warning.no_continuous_vars.is_shown())
+        self.assertIsNone(self.widget.data)
```
VesnaT: Instead of checking self.widget.data I'd suggest checking output.
`self.assertIsNone(self.get_output(self.widget.Outputs.annotated_data))`.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -428,19 +390,13 @@ def set_attr(self, attr_x, attr_y):
...
         super().setup_plot()
-        for axis, var in (("bottom", self.attr_x), ("left", self.attr_y)):
-            self.graph.set_axis_title(axis, var)
-            if var and var.is_discrete:
-                self.graph.set_axis_labels(axis,
-                                           get_variable_values_sorted(var))
-            else:
-                self.graph.set_axis_labels(axis, None)
+        self.graph.set_axis_title("bottom", self.attr_x)
+        self.graph.set_axis_title("left", self.attr_y)
```
VesnaT: Shouldn't this be implemented on the graph?
It is indeed an old issue, but the axis titles are not cleared when data is removed.

Orange/widgets/visualize/tests/test_owscatterplot.py
janezd: Right. But it doesn't hurt to do both.

Orange/widgets/visualize/owscatterplot.py
VesnaT: Yes, let's do it in another PR

PR #3306, opened 2018-10-11
Saving data to excel and ods

eadaradhiraj:
##### Issue

Fixes #3096

##### Description of changes
This pull request is to solve issue # 3096 which requests support for excel and ods file formats while saving data

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

PR #3437, opened 2018-12-03
[FIX] Save Data: Reset writer upon changing the extension

ajdapretnar:
##### Issue

Save Data doesn't change the writer when the user changes the extension.

##### Description of changes
Upon changing the extension, writer is set accordingly.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owsave.py
+++ b/Orange/widgets/data/owsave.py
@@ -55,7 +55,6 @@ class OWSave(widget.OWWidget):
         self.basename = ""
         self.type_ext = ""
         self.compress_ext = ""
-        self.writer = None
 
         form = QFormLayout(
             labelAlignment=Qt.AlignLeft,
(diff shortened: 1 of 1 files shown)


janezd:
What about eliminating the attribute `self.writer`, and calling `get_writer_selected()` at the two places when the widget needs the writer? There is no need to store data that can be trivially computed from the data you already have.

It's preferable to have a *single source of truth* whenever possible and practical. The bug that you're fixing here occurred because there were two and they were not properly synchronized.

ajdapretnar:
@janezd Fixed as suggested. Could you perhaps give it another look and merge, if everything is ok. #3306 depends on this.

janezd:
I think that tests shouldn't call `update_ext` because the widget doesn't call it either.

I thus removed this and the related redundant change of method name. This simplifies the PR so that it fixes only the problem with the writer.

@lanzagar, can you quickly check that it's still OK? Changes are now minimal.

PR #3450, opened 2018-12-07
[FIX] Scatter Plot: Always setup plot

VesnaT:
##### Issue

When inputing features and data into Scatterplot at the same time handleNewSignals() is invoked only once (handleNewSignals() then calls setup_plot()). Therefore attr_x and attr_y, set from settings, could be the same as input features and self.setup_plot() would never be called.

##### Description of changes
Always call self.attr_changed() when receiving Features on the input.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -407,8 +407,8 @@ class OWScatterPlot(OWDataProjectionWidget):
                 self.data.domain is not None and \
                 all(attr in self.data.domain for attr
                         in self.attribute_selection_list):
-            self.set_attr(self.attribute_selection_list[0],
-                          self.attribute_selection_list[1])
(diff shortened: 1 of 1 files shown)


janezd:
Why do we need to call `setup_plot` when feature input does not actually change the current `attr_x` and `attr_y`? Does it have any useful side effects or does it just trigger an unnecessary update?

janezd:
Answer to self after getting an answer from @VesnaT: if this input features signal is present, `super().handleNewSignals()` is not called because it's in the `else` part. The fix is rather dangerous though, because `super().handleNewSignals()` is still not called; it works just because `attr_changed()` calls all the same methods that are called by `handleNewSignals()`, but this can change in the future.

We can merge this PR out of urgency, but this has to be fixed by setting `__invalidate` to `True` instead.

PR #3460, opened 2018-12-09
[MNT] Fix pylint issues in visualization widgets and tests

janezd:
This PR fixes 90 issues that required some superficial changes.

Remaining 25 issues are about too many attributes, ancestors, branches or lines, or about accessing protected members. Fixing them requires deeper refactoring.

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -483,7 +483,8 @@ class OWBoxPlot(widget.OWWidget):
             return
 
         if not self.is_continuous:
-            return self.display_changed_disc()
+            self.display_changed_disc()
+            return
 
         self.mean_labels = [self.mean_label(stat, attr, lab)
                             for stat, lab in zip(self.stats, self.label_txts)]
@@ -502,7 +503,8 @@ class OWBoxPlot(widget.OWWidget):
             return
 
         if not self.is_continuous:
-            return self.display_changed_disc()
+            self.display_changed_disc()
+            return
 
         self.order = list(range(len(self.stats)))
         criterion = self._sorting_criteria_attrs[self.compare]
@@ -578,6 +580,7 @@ class OWBoxPlot(widget.OWWidget):
             self.conts = self.conts[np.sum(np.array(self.conts), axis=1) > 0]
 
             if self.sort_freqs:
+                # pylint: disable=invalid-unary-operand-type
                 self.order = sorted(self.order, key=(-np.sum(self.conts, axis=1)).__getitem__)
         else:
             self.boxes = [self.strudel(self.dist)]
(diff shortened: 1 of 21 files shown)


Orange/widgets/visualize/owboxplot.py
```diff
@@ -1015,10 +1018,8 @@ def line(y0, y1):
         if self.compare == OWBoxPlot.CompareNone or len(self.stats) < 2:
             return
 
-        if self.compare == OWBoxPlot.CompareMedians:
-            crit_line = "median"
-        else:
-            crit_line = "mean"
+        crit_line = "median" if self.compare == OWBoxPlot.CompareMedians \
+            else "mean"
```
janezd: This reduces the number of branches (radon wouldn't fall for this trick, but pylint does).

Orange/widgets/visualize/owdistributions.py
```diff
@@ -201,26 +201,26 @@ def __init__(self):
                     callback=self._on_set_smoothing, createLabel=False)
         self.l_smoothing_r = gui.widgetLabel(box2, "Precise")
 
-        self.cb_disc_cont = gui.checkBox(
+        gui.checkBox(
```
janezd: I've done this only in this widget. We can reduce the number of attributes in other widgets in the same way.

It's not just about lint. The benefit of doing this is that we actually reduce the number of names that programmer must cope with: the name of the control (e.g. checkbox) is derivable from the name of the attribute it controls, so no more `cb_`-somethings.

Orange/widgets/visualize/owtreeviewer.py
```diff
@@ -358,28 +358,26 @@ def update_node_info_cls(self, node):
         else:
             modus = np.argmax(distr)
             tabs = distr[modus]
-            text = self.domain.class_vars[0].values[int(modus)] + "<br/>"
+            text = f"{self.domain.class_vars[0].values[int(modus)]}<br/>"
         if tabs > 0.999:
-            text += "100%, {}/{}".format(total, total)
+            text += f"100%, {total}/{total}"
```
janezd: PyLint complained about this line. The nicest solution was to use f-strings. Other changes in this function are for consistency -- as well as for readability, too.

Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -1035,9 +1040,6 @@ def setText(self, text):
             self._text = text
             self.text.setPlainText(text)
 
-    def text(self):
```
janezd: This function was shadowed by an attribute set in `__init__` and was inaccessible.

Orange/widgets/visualize/owdistributions.py
astaric: +1

Orange/widgets/visualize/owboxplot.py
janezd: You're right. I reverted this change. Even though it silenced the warning, the function was still too complicated and should be split into several functions, so the warning should stay.

Such, more complicated, issues should be fixed in separate PRs for each widget, though. This PR only deals with simple changes, otherwise it would be too big to review and merge.

PR #3462, opened 2018-12-11
Scatterplot: Refactor and enhancement

### LINKED ISSUE #3456: OWScatterPlot sometimes doesn't call handleNewSignals
#3450 is a temporary patch and has to be replaced with one that sets `__invalidated` to `True` instead. See the comment in #3450.

VesnaT:
##### Issue

Fixes #3456

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -223,8 +223,32 @@ class WidgetTest(GuiTest):
         wait : int
             The amount of time to wait for the widget to complete.
         """
+        return self.send_signals([(input, value)], *args,
+                                 widget=widget, wait=wait)
+
+    def send_signals(self, signals, *args, widget=None, wait=-1):
+        """ Send signals to widget by calling appropriate triggers.
+        After all the signals are send, widget's handleNewSignals() in invoked.
+
+        Parameters
+        ----------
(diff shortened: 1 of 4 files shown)


Orange/widgets/tests/base.py
```diff
@@ -225,6 +225,36 @@ def send_signal(self, input, value, *args, widget=None, wait=-1):
         """
         if widget is None:
             widget = self.widget
+        self._send_signal(widget, input, value, *args)
+        widget.handleNewSignals()
+        if wait >= 0 and widget.isBlocking():
+            spy = QSignalSpy(widget.blockingStateChanged)
+            self.assertTrue(spy.wait(timeout=wait))
+
+    def send_signals(self, inputs, values, *args, widget=None, wait=-1):
```
janezd: One more thing: I think a better signature would be `def send_signals(self, signals, *args, widget=None, wait=-1):` where `signals` is a list of tuples. Otherwise the caller has to correctly pair elements of two lists, and the method is then only simpler (it just avoids a `zip`.

janezd:
Kudos for this PR. In particular tests. And `send_signals`. And `update_axes`.

PR #3474, opened 2018-12-14
[FIX] OWMergeData removes table meta attributes

### LINKED ISSUE #3289: Merge Data removes table meta attributes

AndrejaKovacic:
##### Issue

Fixes #3289

##### Description of changes
Merged data is added to original table instead of creating a new one.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owmergedata.py
+++ b/Orange/widgets/data/owmergedata.py
@@ -360,16 +360,16 @@ class OWMergeData(widget.OWWidget):
         of rows given in indices"""
         if not len(indices):
             return None
-        domain = Orange.data.Domain(
+        self.data.domain = Orange.data.Domain(
             *(getattr(self.data.domain, x) + getattr(reduced_extra.domain, x)
               for x in ("attributes", "class_vars", "metas")))
-        X = self._join_array_by_indices(self.data.X, reduced_extra.X, indices)
-        Y = self._join_array_by_indices(
+        self.data.X = self._join_array_by_indices(self.data.X, reduced_extra.X, indices)
+        self.data.Y = self._join_array_by_indices(
             np.c_[self.data.Y], np.c_[reduced_extra.Y], indices)
-        string_cols = [i for i, var in enumerate(domain.metas) if var.is_string]
-        metas = self._join_array_by_indices(
+        string_cols = [i for i, var in enumerate(self.data.domain.metas) if var.is_string]
+        self.data.metas = self._join_array_by_indices(
             self.data.metas, reduced_extra.metas, indices, string_cols)
-        return Orange.data.Table.from_numpy(domain, X, Y, metas)
+        return self.data
 
     @staticmethod
     def _join_array_by_indices(left, right, indices, string_cols=None):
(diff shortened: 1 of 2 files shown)


markotoplak:
Is this changing the input data? The input data should never be modified (if any other widgets were also using the same data they would break). To support workflows, we should never modify data in place.

lanzagar:
What @markotoplak said. And although that could be fixed by making a copy of the data first and modifying that, I don't think you should just overwrite the domain of a Table, the arrays X, Y, metas, etc. It would be too easy to end up with an inconsistent Table (the shape of X might not agree with the size of the domain, ids, ...).

AndrejaKovacic:
I think the most appropriate solution would be to use the function from_numpy and add an option to add metadata as a parameter. I know @lanzagar disagrees with it, because there are some issues with that function, but maybe we should fix from_numpy, instead of trying to create new table here.

janezd:
About travis errors: lint issues are obvious, while to avoid segfault in documentation, you need to rebase to today's master.

Orange/widgets/data/owmergedata.py
```diff
@@ -369,8 +369,12 @@ def _join_table_by_indices(self, reduced_extra, indices):
         string_cols = [i for i, var in enumerate(domain.metas) if var.is_string]
         metas = self._join_array_by_indices(
             self.data.metas, reduced_extra.metas, indices, string_cols)
-        return Orange.data.Table.from_numpy(domain, X, Y, metas)
-
+        table = Orange.data.Table.from_numpy(domain, X, Y, metas)
+        table.name = getattr(self.data, 'name')
+        table.attributes = getattr(self.data, 'attributes')
+        table.ids = getattr(self.data, 'ids')
+        return table
```
janezd: I don't know this code well, but judging by what I see in the `Table` class, I think that you should have `getattr(self.data, 'name', '')` and `getattr(self.data, 'attributes', {})`.

`ids` seems to be always present, so it's just `table.ids = self.data.ids`.

(Perhaps `name` and `attributes` are always present, too, but a lot of code does `getattr` for them.)

Orange/widgets/data/owmergedata.py
```diff
@@ -369,7 +369,11 @@ def _join_table_by_indices(self, reduced_extra, indices):
         string_cols = [i for i, var in enumerate(domain.metas) if var.is_string]
         metas = self._join_array_by_indices(
             self.data.metas, reduced_extra.metas, indices, string_cols)
-        return Orange.data.Table.from_numpy(domain, X, Y, metas)
+        table = Orange.data.Table.from_numpy(domain, X, Y, metas)
+        table.name = getattr(self.data, 'name', '')
+        table.attributes = getattr(self.data, 'attributes', '')
```
janezd: Default for `attributes` should be `{}`.

(When fixed, don't make a new commit but amend this one, then `push -f`)

janezd:
Burning some midnight oil?

PR #3480, opened 2018-12-15
[FIX] Feature statistics fixes

pavlin-policar:
##### Issue
1. Feature Statistics would crash when it got a discrete variable with no `values`
2. Feature Statistics would crash when switching from a data set with a target variable to a data set with no target variable

##### Description of changes
1. Fix. This crashed because `DiscreteVariable.str_val` crashes if it has no `values`. This means that the problem may exist in other widgets as well. I thought about fixing the `str_val` implementation, but I don't really know what it should do in this case ([implementation here](https://github.com/biolab/orange3/blob/master/Orange/data/variable.py#L704-L715)). In both cases a string is returned, so it's not clear what to do here. Perhaps this was even left out by design, so I fixed the issue on the widget.
2. Fix.
3. Update tests to use new signal style i.e. `self.send_data("Data", data)` to `self.send_data(self.widget.Inputs.data, data)`.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfeaturestatistics.py
+++ b/Orange/widgets/data/owfeaturestatistics.py
@@ -787,6 +787,7 @@ class OWFeatureStatistics(widget.OWWidget):
 
         if data is not None:
             self.color_var_model.set_domain(data.domain)
+            self.color_var = None
             if self.data.domain.class_vars:
                 self.color_var = self.data.domain.class_vars[0]
         else:
--- a/Orange/widgets/data/owfeaturestatistics.py
+++ b/Orange/widgets/data/owfeaturestatistics.py
@@ -245,7 +245,8 @@ class FeatureStatisticsTableModel(AbstractSortTableModel):
         def __mode(x, *args, **kwargs):
             if sp.issparse(x):
                 x = x.todense(order="C")
-            return ss.mode(x, *args, **kwargs)[0]
+            # return ss.mode(x, *args, **kwargs)[0]
+            return ut.nanmode(x, *args, **kwargs)[0]  # Temporary replacement for scipy < 1.2.0
 
         self._center = self.__compute_stat(
             matrices,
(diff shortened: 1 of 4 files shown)


Orange/widgets/data/owfeaturestatistics.py
```diff
@@ -500,7 +500,7 @@ def data(self, index, role):
             if role == Qt.DisplayRole:
                 if isinstance(attribute, DiscreteVariable):
                     output = self._center[row]
-                    if not np.isnan(output):
```
janezd: What about `else`? Should `output` be set to `""` in this case? `np.nan` is properly handled below with `if output==output: output = ""`, but this won't work if an attribute has no values and `self._center[row]` is ... whatever it is in this case (probably `0`).

pavlin-policar:
I've removed the commit that handled discrete variables with no values since this should never happen. In the event that it *does* happen, the underlying issue should be fixed and patched up in this widget.

lanzagar:
I am not sure why this passes on travis, but fails on appveyor (maybe different versions of libraries?).
But the error there looks relevant. Can you take a look and try to fix whatever is causing the problem?

thocevar:
I can reproduce appveyor's error on windows. test_owfeaturestatistics -> force_render_table -> FeatureStatisticsTableModel.data tries to get a string value of a DiscreteVariable that has an empty list of values (owfeaturestatistics.py, line 515).

pavlin-policar:
Yes, that's actually the problem. Building a `DiscreteVariable` with no values shouldn't be possible, and apparently, this does not happen on non-Windows systems (for some reason). It's fairly easy to fix the problem here and check for such an event, but that would mask the underlying issue i.e. a discrete variable with no possible values!

thocevar:
We traced the problem to scipy's mode function. For an array of nans it should return nan. It does so in the latest version 1.2, but not in version 1.0 that's used by appveyor.

PR #3490, opened 2018-12-19
Spearman distances: Return 0 when comparing a single row to itself

janezd:
##### Issue

Spearman distance fails on Scipy 1.2 because of stats.spearmanr failing when given a single matrix with a single row.

##### Description of changes

This distance should always return 0, so we can just hard code it.

##### Includes
- [X] Code changes


janezd:
Doc building is still broken (core dump) but tests now pass.

PR #3491, opened 2018-12-20
[FIX][MNT] Remove segfault in tests for building documentation

janezd:
##### Issue

create_widget_catalog.py segfaults when Python is freeing memory.

##### Description of changes

It seems that removing numpy helped - it almost always crashed before, and never crashed after. Numpy was used just to create a table with 50*50*4 zeros. `b"\x00" * 50 * 50 *4"` works just as well.

While searching for the problem I reordered the import order to match that in `canvas.__main__`. I kept it reordered.

I also added `QApplication.setAttribute(Qt.AA_ShareOpenGLContexts)`, as recommended in the output. This flag has to be set before `QApplication` to have an effect. I don't know about other flags, but I moved them, too.

##### Includes
- [X] Code changes

### DIFF
--- a/scripts/create_widget_catalog.py
+++ b/scripts/create_widget_catalog.py
@@ -2,12 +2,9 @@ from copy import copy
 import json
 from os import path, makedirs
 
-import numpy as np
-
-from AnyQt import QtGui
-from AnyQt.QtCore import QRectF, Qt, QTimer
-from AnyQt.QtGui import QImage
+from AnyQt.QtGui import QImage, QPainter
 from AnyQt.QtWidgets import QGraphicsScene, QApplication, QWidget, QGraphicsView, QHBoxLayout
+from AnyQt.QtCore import QRectF, Qt, QTimer
 
 from Orange.canvas import config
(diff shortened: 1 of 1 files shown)


janezd:
@ajdapretnar, this PR hopefully fixes a crash that caused all new PRs to fail. It is related to building a catalog of widgets. I don't know how the catalog on the web is updated, but I suppose it uses the script I changed here.

I'm assigning this PR to you: could you check if the catalog - in particular the icons - is still OK with this script? Or find somebody who knows how to do it? (I don't.)

ajdapretnar:
I am not sure what you mean by widget catalog. Documentation does not have a catalog. If you are referring to the website, this is done elsewhere (website repo) and I don't think it should be affected by numpy at all.
I could only test the documentation locally and it was fine (apart from some small fixes I can do later).

janezd:
@ajdapretnar, I think this is the same script that generates icons for the catalog on the web. If they look OK locally (they do on my machine), let's hope they'll look the same on the web.

PR #3492, opened 2018-12-20
[ENH] OWLouvain: Ensure deterministic clustering

### LINKED ISSUE #3278: Set a fixed random seed in Louvain Clustering
##### Expected behavior
Clusters in Louvain clustering are reproducible between different runs.

##### Actual behavior
Clusters are always different because random seed is not set. Set random seed to 0.

See orange3-single-cell/OWTsne.

pavlin-policar:
##### Issue
Fixes #3278.

##### Description of changes
Version 0.12 of `python-louvain` was just released and with it the `random_state` parameter, which was missing before. I've added the parameter to our wrapper and set the random_state to a fixed value in the widget, so the results are deterministic in the widget.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/clustering/louvain.py
+++ b/Orange/clustering/louvain.py
@@ -8,6 +8,8 @@ https://sites.google.com/site/findcommunities/
 
 import numpy as np
 import networkx as nx
+# NOTE: The ``community`` package might be renamed in the near future, see
+# GH issue https://github.com/taynaud/python-louvain/issues/23
 from community import best_partition
 from sklearn.neighbors import NearestNeighbors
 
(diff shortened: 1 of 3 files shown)


pavlin-policar:
I've also fixed the version to 0.12 because the API might change in the next version https://github.com/taynaud/python-louvain/issues/23.

janezd:
Fixing a version looks like a good idea (as anybody who just spent a day trying to find why tests suddenly fail would confirm), yet it may be better to fix problems as they come, instead of jumping a few versions at some later time and having to fix the whole heap then.

I would let it fail when the new version comes and just change the import name when it happens. But it would be a good idea to add a comment at the point where we'll have to rename the import.

pavlin-policar:
I've removed the version from the requirements and added a note to the import.

PR #3497, opened 2018-12-21
image export: remove custom error dialog

markotoplak:
##### Issue
Image export opened a custom dialog.

##### Description of changes
Now the exception runs as-is and users can therefore report it to sentry.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/saveplot.py
+++ b/Orange/widgets/utils/saveplot.py
@@ -1,7 +1,5 @@
 import os.path
-import traceback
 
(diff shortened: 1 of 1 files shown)


janezd:
You're right. Still, we should give nice messages only for I/O-related errors, not errors in our code. What about maybe catching just `OSError` (with subclasses)?

ales-erjavec:
> What about maybe catching just OSError (with subclasses)?

Yes. That would be best.

PR #3500, opened 2018-12-21
appveyor, travis: Speedup build

VesnaT:
##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -558,6 +558,7 @@ class Table(MutableSequence, Storage):
                 self.metas[i, j] = var.to_val(val)
         if weights is not None:
             self.W = np.array(weights)
+        self.attributes = {}
         return self
 
     @classmethod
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -13,6 +13,8 @@ matrix:
           script: source $TRAVIS_BUILD_DIR/.travis/build_doc.sh
         - python: '3.6'
           env: UPLOAD_COVERAGE=true
+        - python: '3.7'
+          env: UPLOAD_COVERAGE=true
```
lanzagar: We don't need UPLOAD_COVERAGE twice (it is already checked in 3.6).
But python 3.7 needs sudo, so change the env line for this:
`sudo: true`

For more on this see: https://github.com/travis-ci/travis-ci/issues/9815

janezd:
> Table has methods append, insert, extend and concatenate, which, to my knowledge, nobody (in his right mind) uses.

... except for `concatenate`, which is reasonably used in `OWConcatenate` and `OWVennDiagram`. I will create another PR with this commit. I will keep `concatenate`, but change it so that it won't resize arrays but stack them.

janezd:
I added a commit that:

- removes the failing tests for methods that mutate the table so that it's size changes (`append`, `insert`, `extend`, `__delitem__`). The methods themselves will be removed in another pull request because this will require some further checking, while changing the way we run tests is rather urgent;
- reimplements `Table.concatenate` so that it uses `vstack` instead of `extend`. It also removes some unused functionality;
- fixes the code in few other places.

The commit obviously decreases the coverage.

PR #3502, opened 2018-12-24
[MNT] Fix warnings in widget tests

janezd:
##### Issue

Logs are polluted with warnings. Some are relevant and are fixed, and some are expected by tests and should be silenced. Logs are also printed out instead of being tested.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/projection/manifold.py
+++ b/Orange/projection/manifold.py
@@ -109,7 +109,7 @@ class MDS(SklProjector):
     __wraps__ = skl_manifold.MDS
     name = 'MDS'
 
-    def __init__(self, n_components=2, metric=True, n_init=4, max_iter=300,
+    def __init__(self, n_components=2, metric=True, n_init=None, max_iter=300,
                  eps=0.001, n_jobs=1, random_state=None,
                  dissimilarity='euclidean', init_type="random", init_data=None,
                  preprocessors=None):
@@ -147,6 +147,8 @@ class MDS(SklProjector):
             init_data = None
 
         params["dissimilarity"] = dissimilarity
+        if params["n_init"] is None:
+            params["n_init"] = 4 if init_data is None else 1
         mds = self.__wraps__(**params)
         mds.fit(X, y=Y, init=init_data)
         mds.domain = domain
--- a/Orange/projection/manifold.py
+++ b/Orange/projection/manifold.py
@@ -109,7 +109,7 @@ class MDS(SklProjector):
     __wraps__ = skl_manifold.MDS
     name = 'MDS'
 
-    def __init__(self, n_components=2, metric=True, n_init=None, max_iter=300,
+    def __init__(self, n_components=2, metric=True, n_init=4, max_iter=300,
                  eps=0.001, n_jobs=1, random_state=None,
                  dissimilarity='euclidean', init_type="random", init_data=None,
                  preprocessors=None):
@@ -147,8 +147,6 @@ class MDS(SklProjector):
             init_data = None
 
         params["dissimilarity"] = dissimilarity
-        if params["n_init"] is None:
-            params["n_init"] = 4 if init_data is None else 1
         mds = self.__wraps__(**params)
         mds.fit(X, y=Y, init=init_data)
         mds.domain = domain
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -353,30 +353,30 @@ def stats(X, weights=None, compute_variance=False):
 
 
 def _nan_min_max(x, func, axis=0):
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", ".*All-NaN slice encountered.*",
-                                RuntimeWarning)
-        if not sp.issparse(x):
+    if not sp.issparse(x):
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", ".*All-NaN slice encountered.*",
+                                    RuntimeWarning)
             return func(x, axis=axis)
-        if axis is None:
-            extreme = func(x.data, axis=axis) if x.nnz else float('nan')
-            if sparse_has_implicit_zeros(x):
-                extreme = func([0, extreme])
-            return extreme
-        if axis == 0:
-            x = x.T
-        else:
-            assert axis == 1
+    if axis is None:
+        extreme = func(x.data, axis=axis) if x.nnz else float('nan')
+        if sparse_has_implicit_zeros(x):
+            extreme = func([0, extreme])
+        return extreme
+    if axis == 0:
+        x = x.T
+    else:
+        assert axis == 1
 
-        # TODO check & transform to correct format
-        r = []
-        for row in x:
-            values = row.data
-            extreme = func(values) if values.size else float('nan')
-            if sparse_has_implicit_zeros(row):
-                extreme = func([0, extreme])
-            r.append(extreme)
-        return np.array(r)
+    # TODO check & transform to correct format
+    r = []
+    for row in x:
+        values = row.data
+        extreme = func(values) if values.size else float('nan')
+        if sparse_has_implicit_zeros(row):
+            extreme = func([0, extreme])
+        r.append(extreme)
+    return np.array(r)
 
 
 def nanmin(x, axis=None):
@@ -427,19 +427,12 @@ def nanmean(x, axis=None):
     """ Equivalent of np.nanmean that supports sparse or dense matrices. """
     def nanmean_sparse(x):
         n_values = np.prod(x.shape) - np.sum(np.isnan(x.data))
+        if not n_values:
+            warnings.warn(RuntimeWarning, "Mean of empty slice")
+            return np.nan
         return np.nansum(x.data) / n_values
 
-    if not x.size:
-        res = np.nan if np.min(x.shape) == 0 else 0
-        if axis is None:
-            return res
-        else:
-            return np.full(x.shape[:axis] + x.shape[axis + 1:], res)
-
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", ".*Mean of empty slice.*",
-                                RuntimeWarning)
-        return _apply_func(x, np.nanmean, nanmean_sparse, axis=axis)
+    return _apply_func(x, np.nanmean, nanmean_sparse, axis=axis)
 
 
 def nanvar(x, axis=None):
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -354,10 +354,7 @@ def stats(X, weights=None, compute_variance=False):
 
 def _nan_min_max(x, func, axis=0):
     if not sp.issparse(x):
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore", ".*All-NaN slice encountered.*",
-                                    RuntimeWarning)
-            return func(x, axis=axis)
+        return func(x, axis=axis)
     if axis is None:
         extreme = func(x.data, axis=axis) if x.nnz else float('nan')
         if sparse_has_implicit_zeros(x):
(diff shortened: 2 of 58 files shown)


Orange/widgets/data/owrank.py
```diff
@@ -368,31 +371,33 @@ def set_learner(self, scorer, id):
...
+                try:
+                    scores = np.array([estimator(data, attr)
+                                       for attr in data.domain.attributes])
+                except ValueError:
+                    log.error("%s doesn't work on this data", method.name)
+                    scores = np.full(len(data.domain.attributes), np.nan)
+                else:
+                    log.warning("%s had to be computed separately for each "
+                                "variable", method.name)
+            return scores
```
janezd: Reorganized to give an error or a warning (or neither), not both.

Orange/widgets/data/owrank.py
```diff
@@ -142,14 +142,17 @@ def headerData(self, section, orientation, role=Qt.DisplayRole):
...
-        else:
-            vmax = np.nanmax(values)
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", ".*All-NaN slice encountered.*",
+                                    RuntimeWarning)
+            try:
+                vmin = np.nanmin(values)
+                if np.isnan(vmin):
+                    raise TypeError
+            except TypeError:
```
lanzagar: I prefer explicit code that checks before doing something "stupid" instead of doing it first and catching exceptions.
If we had that here, we would also not need to suppress warnings...

I am guessing it would be enough to check `if np.isfinite(values).any()` and use the computed values, otherwise set to inf?
Shorter, clearer, explicit.
(and not tested, so maybe not enough :))

Sidenote: If sometime in the future, we pass a list of values (instead of np.array), the warning would change from
All-NaN slice encountered
to
All-NaN axis encountered
Just to keep in mind when setting this filter in functions that are otherwise fine with any iterable.

Orange/widgets/data/owrank.py
```diff
@@ -368,31 +371,33 @@ def set_learner(self, scorer, id):
...
-        try:
-            scores = np.asarray(estimator(data))
-        except ValueError:
-            log.warning("Scorer %s wasn't able to compute all scores at once",
-                        method.name)
+        # These errors often happen, but they result in nans, which
+        # are handled correctly by the widget
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', '.*true_divide.*')
+            warnings.filterwarnings('ignore', '.*double_scalars.*')
```
lanzagar: I think it would be better to solve the problem of warnings at the source, i.e. in the scoring method that produces them, then in potentially several places where they are used.
Also, it is not hard to imagine that a general warning filter for all methods here can cover up an actual problem/bug. If not now, then maybe in the future.

Orange/widgets/data/tests/test_owdatasampler.py
```diff
@@ -64,7 +66,9 @@ def select_sampling_type(self, sampling_type):
     def test_no_intersection_in_outputs(self):
         """ Check whether outputs intersect and whether length of outputs sums
         to length of original data """
-        self.send_signal("Data", self.zoo)
+        iris = Table("iris")
+        iris.X = np.arange(iris.X.size).reshape(iris.X.shape)
```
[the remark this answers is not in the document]
janezd: No, it's to avoid warning about less than 4 instances in some classes (see the comment below).

Orange/widgets/data/owrank.py
janezd: I agree but we don't understand the code well enough to dare to remove catching the `TypeError`. Hence I only changed the code to filter only warnings only in `vmin = np.nanmin(values)`.

Orange/projection/manifold.py
```diff
@@ -109,7 +109,7 @@ class MDS(SklProjector):
     __wraps__ = skl_manifold.MDS
     name = 'MDS'
 
-    def __init__(self, n_components=2, metric=True, n_init=4, max_iter=300,
+    def __init__(self, n_components=2, metric=True, n_init=None, max_iter=300,
```
lanzagar: I would prefer if our wrappers retain the same signature/defaults as the wrapped function as much as possible.
It is then much easier to update them, if/when these values change upstream (and they often do).

This leaves us with a couple of choices of how to combat the current awkward defaults:
- Fix the calls of MDS, and explicitly set `n_init` to 1 when `init` is used. Not sure it that is how the authors imagined it... (does not sound appealing)
- Let sklearn change it and filter out the warning when calling fit, if that bothers us
- We make the same change before them in our call. Basically what you did in L150, but do it when `init` is given instead of when `n_init` is not.

janezd: The first option was the simplest because this happened only in one place.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -169,12 +171,16 @@ def error(err):
                 error(self.Error.not_enough_rows)
             elif not self.data.domain.attributes:
                 error(self.Error.no_attributes)
-            elif not self.data.is_sparse() and \
-                    not np.sum(np.nanstd(self.data.X, axis=0)):
-                error(self.Error.constant_data)
             elif not self.data.is_sparse() and \
                     np.all(~np.isfinite(self.data.X)):
                 error(self.Error.no_valid_data)
+            elif not self.data.is_sparse():
```
lanzagar: If we now have an `elif not sparse` then the above elif can be moved inside this one...

Orange/widgets/visualize/owboxplot.py
```diff
@@ -680,16 +680,19 @@ def compute_tests(self):
...
+
+        # pylint: disable=comparison-with-itself
         def stat_ttest():
             d1, d2 = self.stats
-            if d1.n == 0 or d2.n == 0:
+            if d1.n < 2 or d2.n < 2:
                 return np.nan, np.nan
             pooled_var = d1.var / d1.n + d2.var / d2.n
+            # pylint: disable=comparison-with-itself
+            if pooled_var == 0 or pooled_var != pooled_var:
```
lanzagar: Isn't `np.isnan` clearer?

Orange/widgets/visualize/owboxplot.py
```diff
@@ -724,15 +729,21 @@ def stat_ANOVA():
                 # t = "Mann-Whitney's z: %.1f (p=%.3f)" % (z, self.p)
             else:
                 t, self.p = stat_ttest()
-                t = "Student's t: %.3f (p=%.3f)" % (t, self.p)
+                if t == t:
```
lanzagar: Isn't `np.isnan` clearer?

Orange/widgets/visualize/owboxplot.py
```diff
@@ -724,15 +729,21 @@ def stat_ANOVA():
...
+                    t = 0
         else:
             if self.compare == OWBoxPlot.CompareMedians:
                 t = ""
                 # U, self.p = -1, -1
                 # t = "Kruskal Wallis's U: %.1f (p=%.3f)" % (U, self.p)
             else:
                 F, self.p = stat_ANOVA()
-                t = "ANOVA: %.3f (p=%.3f)" % (F, self.p)
+                if F == F:
```
lanzagar: Isn't `np.isnan` clearer?

Orange/widgets/visualize/owboxplot.py
```diff
@@ -724,15 +729,21 @@ def stat_ANOVA():
                 # t = "Mann-Whitney's z: %.1f (p=%.3f)" % (z, self.p)
             else:
                 t, self.p = stat_ttest()
-                t = "Student's t: %.3f (p=%.3f)" % (t, self.p)
+                if t == t:
+                    t = "Student's t: %.3f (p=%.3f)" % (t, self.p)
+                else:
+                    t = 0
```
lanzagar: `t = ""`
As in ANOVA below and CompareMedians...

Orange/widgets/visualize/owsieve.py
```diff
@@ -43,8 +43,9 @@ def __init__(self, data, attr1, attr2):
         self.probs_x = self.observed.sum(axis=0) / self.n
         self.probs_y = self.observed.sum(axis=1) / self.n
         self.expected = np.outer(self.probs_y, self.probs_x) * self.n
-        self.residuals = \
-            (self.observed - self.expected) / np.sqrt(self.expected)
+        with np.errstate(divide="ignore", invalid="ignore"):
+            self.residuals = \
+                (self.observed - self.expected) / np.sqrt(self.expected)
```
lanzagar: Can we not divide by zero instead of ignoring warnings?

Orange/classification/naive_bayes.py
```diff
@@ -48,22 +49,62 @@ def __init__(self, log_cont_prob, class_prob, domain):
 
     def predict_storage(self, data):
         if isinstance(data, Instance):
-            data = [data]
-        if len(data.domain.attributes) == 0:
-            probs = np.tile(self.class_prob, (len(data), 1))
+            data = np.atleast_2d(data.X)
+        elif isinstance(data, Storage):
+            data = data.X
```
lanzagar: The `Storage` class does not have `X` - `Table` has it, and even there it is not declared properly :)
Joking aside, our Learners and Models can implement fit/predict which work with numpy arrays or fit_storage/predict_storage which do not assume/require numpy data.
Instead of `data = data.X` you could replace the implementation of predict_storage with the `predict` method. That can have implications for non-numpy based storages (SqlTable)...
In short - this change needs some consideration and should probably be in its own PR.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -240,17 +241,22 @@ def effective_variables(self):
...
-        self.vizrank_button.setToolTip(text)
+        with warnings.catch_warnings():
+            # nanstd may trigger this warning, but it's harmless
+            warnings.filterwarnings("ignore", "Degrees of freedom .*",
+                                    RuntimeWarning)
+            self.vizrank.initialize()
+            is_enabled = self.data is not None and \
+                not self.data.is_sparse() and \
+                len(self.xy_model) > 2 and len(self.data[self.valid_data]) > 1 \
+                and np.all(np.nan_to_num(np.nanstd(self.data.X, 0)) != 0)
```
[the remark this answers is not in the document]
janezd: You're right, I removed this from the commit.

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -864,7 +864,10 @@ def _get_continuous_colors(self, c_data, subset):
         except the former is darker. If the data has a subset, the brush
         is transparent for points that are not in the subset.
         """
-        self.scale = DiscretizedScale(np.nanmin(c_data), np.nanmax(c_data))
+        with warnings.catch_warnings():
+            # DiscretizedScale handle's nans just fine
+            warnings.filterwarnings("ignore", "All-NaN slice.*", RuntimeWarning)
+            self.scale = DiscretizedScale(np.nanmin(c_data), np.nanmax(c_data))
```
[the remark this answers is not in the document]
janezd: Done.

Orange/widgets/visualize/owsieve.py
janezd: No, it's a vector. I could create a mask, but it would be annoying. It's simpler this way.

Orange/classification/naive_bayes.py
janezd: Moved to #3540.

Orange/statistics/util.py
```diff
@@ -353,7 +354,10 @@ def weighted_mean():
 
 def _nan_min_max(x, func, axis=0):
     if not sp.issparse(x):
-        return func(x, axis=axis)
+        with warnings.catch_warnings():
```
lanzagar: This currently still changes the behaviour of nanmin/nanmax compared to numpy's version.
I thought we agreed that for numpy arrays our functions should be equivalent to np.nanmin/max

PR #3503, opened 2018-12-26
[MNT] Fix test warnings in scripting part

janezd:
**Similar to #3502**, this PR deals with most warnings in the scripting part, except those in postgresql and a few in the code I don't sufficiently understand.

### DIFF
--- a/Orange/evaluation/clustering.py
+++ b/Orange/evaluation/clustering.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy as np
 from sklearn.metrics import silhouette_score, adjusted_mutual_info_score, silhouette_samples
 
@@ -35,17 +37,22 @@ class ClusteringScore(Score):
 
     def from_predicted(self, results, score_function):
         # Clustering scores from labels
-        if self.considers_actual:
-            return np.fromiter(
-                (score_function(results.actual.flatten(), predicted.flatten())
-                 for predicted in results.predicted),
-                dtype=np.float64, count=len(results.predicted))
-        # Clustering scores from data only
-        else:
-            return np.fromiter(
-                (score_function(results.data.X, predicted.flatten())
-                 for predicted in results.predicted),
-                dtype=np.float64, count=len(results.predicted))
+        # This warning filter can be removed in scikit 0.22
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore", "The behavior of AMI will change in version 0\.22.*")
+            if self.considers_actual:
+                return np.fromiter(
+                    (score_function(results.actual.flatten(),
+                                    predicted.flatten())
+                     for predicted in results.predicted),
+                    dtype=np.float64, count=len(results.predicted))
+            # Clustering scores from data only
+            else:
+                return np.fromiter(
+                    (score_function(results.data.X, predicted.flatten())
+                     for predicted in results.predicted),
+                    dtype=np.float64, count=len(results.predicted))
 
 
 class Silhouette(ClusteringScore):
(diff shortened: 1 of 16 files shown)


Orange/classification/naive_bayes.py
```diff
@@ -57,9 +57,10 @@ def predict_storage(self, data):
                 np.log(self.class_prob) +
                 np.array([np.zeros_like(self.class_prob)
                           if isnan(ins.x).all() else
-                          np.sum(attr_prob[:, int(attr_val)]
-                                 for attr_val, attr_prob in zip(ins, self.log_cont_prob)
-                                 if not isnan(attr_val))
+                          np.sum([attr_prob[:, int(attr_val)]
+                                  for attr_val, attr_prob in
+                                  zip(ins, self.log_cont_prob)
+                                  if not isnan(attr_val)], axis=0)
```
lanzagar: Just changing `np.sum` to `sum` would be shorter and I think clearer, than creating a python list, and then using axis (I have to think about axis=0/1 every time).
But this works just as well.

janezd: I removed this commit: naive Bayesian classifier is modified in #3540 and this patch is moved there to prevent conflicts.

I used `sum`, as suggested.

janezd:
Build crashed (output too large) after removing the fix for `np.sum(generator)`. Now I got to the bottom of it.

`"once"` filter in `warnings` has a nice undocumented behaviour: if the list of filters is changed, the registry of printed warnings is cleared. Hence any call to `filterwarnings`, even within `catch_warnings` context, resets the registry.

The problem in this PR is [silencing the warnings in `RowInstance.__init__`](https://github.com/biolab/orange3/pull/3503/files#diff-1a46745833e0557e8e3cbff98c867258R73) which resets the filters at every iteration of the loop over all instances in naive Bayesian classifier.

If you agree, I'll:

- I'll leave silencing of warnings within tests. It's harmless since the unittest framework resets the filters anyway.
- Almost all filters in actual code silence the scikit-sparse's use of matrix subclass. I would disable this warning for entire Orange, that is, put this filter to `Orange.__init__`. Benefits:
    - It avoids randomly resetting the filters.
    - We never want to see this warning anyway because it's never our fault and we can do nothing about it.
    - It will be more trivial to remove when it's fixed in scikit-sparse
- I will try to remove other filters, either by
    - fixing the code so they're not issued or
    - using `np.errstate` instead or
    - keeping filters, if they are unlikely to cause harm,

    or I may keep them, but hesitantly unless we expect them to go away some day.

When/If I get a green light from you, I'll first fix this PR and then do the same in #3502 (there seem to be just six places, also thanks to your cautious approach reviewing it).

lanzagar:
At least now we know why filtering out some warnings can actually lead to more of them in the output!
I agree with all three points, and am a big fan of "fixing the code so they're not issued" approach anyway :)

Orange/evaluation/clustering.py
```diff
@@ -35,17 +37,22 @@ class ClusteringScore(Score):
...
-        # Clustering scores from data only
-        else:
-            return np.fromiter(
-                (score_function(results.data.X, predicted.flatten())
-                 for predicted in results.predicted),
-                dtype=np.float64, count=len(results.predicted))
+        # This warning filter can be removed in scikit 0.22
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore", "The behavior of AMI will change in version 0\.22.*")
```
janezd: I kept this `catch_warnings` although it's not in tests. It's supposed to be temporary.

PR #3506, opened 2018-12-27
[ENH] Label subset

### LINKED ISSUE #3376: Projections: Show labels for subset
In visualizations, "Label only selected points" could also apply to points in the subset. Or there could be a separate check box.

janezd:
##### Issue

Fixes #3376: subsets are similar to selections, so it would make sense to label the subset data instances, too.

**Based on #3500 for faster CI testing, but otherwise unrelated.** If reviews are positive, I'll rebase to master so it can be merged.

##### Description of changes

To keep the GUI simple, this PR does not introduce another checkbox, but just adds subset instances to those that are labelled. Checkbox is renamed from *Label only selected points* to *Label only selection and subset*.

out

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -1024,25 +1024,22 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         for label in self.labels:
             self.plot_widget.removeItem(label)
         self.labels = []
-        if self.scatterplot_item is None \
-                or self.label_only_selected and self.selection is None:
-            self._signal_too_many_labels(False)
+
+        mask = None
+        if self.scatterplot_item is not None:
+            x, y = self.scatterplot_item.getData()
+            mask = self._label_mask(x, y)
+
+        if mask is not None:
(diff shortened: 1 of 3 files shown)


Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -1052,13 +1049,31 @@ def update_labels(self):
...
             self._too_many_labels = too_many
             self.too_many_labels.emit(too_many)
 
+    def _label_mask(self, x, y):
+        (x0, x1), (y0, y1) = self.view_box.viewRange()
+        mask = np.logical_and(
+            np.logical_and(x >= x0, x <= x1),
+            np.logical_and(y >= y0, y <= y1))
+        if self.label_only_selected:
+            sub_mask = self._filter_visible(self.master.get_subset_mask())
```
markotoplak: Aaaa, having masks that can be None make the code so ugly... But I did not find a better solution. 

Perhaps in the future we could have some helper functions for logical_* where arguments could be None? Or just have masks that are never None?

PR #3509, opened 2019-01-03
image export: show custom error dialog only for OSError

markotoplak:
##### Issue
Any errors in image export open a custom dialog and they can not be reported to sentry. 

##### Description of changes
Catch OSError as suggested in #3497. This still reports user-dependent errors (wrong permissions, out of disk space) to the user but our errors will open Orange's standard error dialogue.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
This is the latest master. I did not play with #3500, but I did restart you last PR (which is based on #3500) and it crashes the same.

PR #3507, opened 2018-12-28
[ENH] Warn when data subset is unrelated to data

janezd:
##### Issue

Fixes #3292.

Also switches to `np.fromiter` when constructing a subset mask to avoid creating a Python list.

##### Includes
- [X] Code changes
- [X] Tests

PR #3514, opened 2019-01-04
OWProjectionWidgetBase: Move checking of subsets to handleNewSignals

janezd:
##### Issue

#3507 introduced checking of subsets in `get_subset_data`.

- `handleNewSignals` is a better place because this is where one would expect such test.
- checking the length of the mask was not the best idea because data instances appear multiple times; it's better to subtract the two sets
- checks need to be done on entire data not just `valid_data` to prevent false warnings.

##### Description of changes

Code is changed, but existing tests remain valid.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/visualize/utils/widget.py
+++ b/Orange/widgets/visualize/utils/widget.py
@@ -367,7 +367,7 @@ class OWDataProjectionWidget(OWProjectionWidgetBase):
             "Subset data contains some instances that do not appear in "
             "input data")
         subset_independent = Msg(
-            "None of subset data instance appear in input data")
+            "No subset data instances appear in input data")
 
     settingsHandler = DomainContextHandler()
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/utils/widget.py
```diff
@@ -455,11 +455,21 @@ def use_context(self):
...
         self.controls.graph.alpha_value.setEnabled(subset is None)
 
     def handleNewSignals(self):
+        self.Warning.subset_independent.clear()
+        self.Warning.subset_not_subset.clear()
+        if self.data is None or self.subset_data is None:
+            self.subset_indices = set()
+        else:
+            self.subset_indices = {e.id for e in self.subset_data}
+            ids = {ex.id for ex in self.data}
```
lanzagar: `ids = set(self.data.ids)`

PR #3518, opened 2019-01-07
[ENH] Scatterplot: Draw separate regression lines for colors; add orthonormal regression

### LINKED ISSUE #3519: Scatter plot shows regression lines; orthogonal regression may be better
This is based on #3518, but the concern is equally valid without.

## Expected:

iris-expected

## Actual

iris-actual

## Not Really

"Expected" is expected only when x and y are independent; the picture at the top uses orthogonal regression instead of linear. "Actual" (which we have now) models y as a function of x, which is sometimes what we want. Consider heart disease data where we currently (well, in #3518) have

age-vs-hr,

which shows the maximal HR as a function of age, while the following shows nothing.

heart-disease-orthogonal

Comments welcome. As much as I prefer to keep GUIs simple, I find this worth a checkbox. And as much as I like to avoid showing the user some options that (s)he wouldn't understand, I could justify adding such checkbox because it's important that users are aware of the difference.

janezd:
##### Issue

It would be interesting to see separate regression lines for different colors - when points in the scatter plot are colored with discrete colors.

Also fixes #3519.

##### Description of changes

age-vs-hr

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -100,15 +100,19 @@ class OWScatterPlotGraph(OWScatterPlotBase):
 
     def __init__(self, scatter_widget, parent):
         super().__init__(scatter_widget, parent)
-        self.reg_line_item = None
+        self.reg_line_items = []
 
     def clear(self):
         super().clear()
-        self.reg_line_item = None
+        self.reg_line_items.clear()
 
     def update_coordinates(self):
         super().update_coordinates()
         self.update_axes()
+        self.update_regression_line()
+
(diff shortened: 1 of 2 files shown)


janezd:
To be done: remove points with missing values for color.

markotoplak:
I like this a lot!

markotoplak:
Also, documentation should be updated, but probably not in this PR. The old docs still show some discrete features.

janezd:
I changed widths to 2. I could now even remove the argument, but decided to keep it if we change our minds in the future.

PR #3529, opened 2019-01-11
[FIX] Python script widget: prevent data loss 

### LINKED ISSUE #3531: Python scripts not saving in ows file
##### Orange version
3.19 on Windows 10

##### Expected behavior
Python script object changes should save with project.

##### Actual behavior
After saving project with python script and closing Orange, the next time I open the project the python script is back to the default "Hello World" example instead of my saved script.

1. Create new project with python script object
2. Modify the python script to something other than Hello World
3. Save the project
4. Close orange
5. Re-open Orange 
6. Open the project
7. The python script reverts to Hello World

markotoplak:
##### Issue
Users were losing their scripts:
1. If they added a new enrty to the library (button "+") they lost the current script.
2. If they did not save the script into the library it got lots.

Fixes #3531

##### Description of changes
1. Adding an entry to the library initializes that script with the current script text.
2. The current script text is saved as a schema_only settings and is restored on load. I also added an option to restore the library version.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpythonscript.py
+++ b/Orange/widgets/data/owpythonscript.py
@@ -463,14 +463,17 @@ class OWPythonScript(widget.OWWidget):
 
         new_from_file = QAction("Import Script from File", self)
         save_to_file = QAction("Save Selected Script to File", self)
+        restore_saved = QAction("Undo Changes to Selected Script", self)
         save_to_file.setShortcut(QKeySequence(QKeySequence.SaveAs))
 
         new_from_file.triggered.connect(self.onAddScriptFromFile)
         save_to_file.triggered.connect(self.saveScript)
+        restore_saved.triggered.connect(self.restoreSaved)
 
         menu = QMenu(w)
         menu.addAction(new_from_file)
         menu.addAction(save_to_file)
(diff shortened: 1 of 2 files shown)


markotoplak:
Currently, this PR is changing a setting on every textChanged. I tried with long texts and it did not seem problematic.

Still, I think saving scriptText belongs to storeSpecificSettings, but it is not called without a context. Should I then rather add a context handler?

astaric:
ContextHandler calls `self.settings_from_widget(widget)` in `pack_data`, which in turn calls `widget.storeSpecificSettings()`.

You could consider "promoting" this behaviour to SettingsHandler thus making it available to all widgets, not just those that use contexts?

All widgets have this method (https://github.com/biolab/orange3/blob/master/Orange/widgets/widget.py#L882) and it does nothing by default so it should not break anything.

astaric:
Then I would at least use the same widget method (storeSpecificSettings) for the job.

This way, if the functionality is ever moved to SettingsHandler, a search for storeSpecificSettings will find this widget and PrepareSavingSettingsHandler can be removed at that point.

PR #3537, opened 2019-01-16
[FIX] appveyor: Bump scipy to 1.2.0

thocevar:
##### Issue
Scipy.stats.mode handles nans incorrectly in version 1.0 (see https://github.com/biolab/orange3/pull/3480#issuecomment-454775335)

##### Description of changes
Bump scipy version to 1.2.0 on appveyor.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ environment:
     BUILD_GLOBAL_OPTIONS: build -j1
     BUILD_ENV: wheel==0.29.0 pip==9.0.1 numpy==1.9.3 sphinx==1.8.2
     # SIP 4.19.4+ with PyQt5==5.9.1+ segfault our tests (GH-2756)
-    TEST_ENV: sip==4.19.6 PyQt5==5.9.2 numpy~=1.14.0 scipy~=1.0.0 scikit-learn pandas==0.21.1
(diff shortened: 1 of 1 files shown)


janezd:
I'm reassigning this to @ales-erjavec. As I understand, it bumps the version in Appveyor test environment, but not in the build. Ales, can you verify that we use SciPy >= 1.2.0 in Windows builds? Also, it seems to me that we use numpy 1.9.3 on Windows, while we test on 1.14? Can you check?

Background: this PR is needed to fix a problem in #3480, which is due to a bug in `scipy.stats.mode` that is fixed in 1.2.0.

ales-erjavec:
conda does not yet have scipy 1.2.0 builds for windows. The current miniconda based installer still has scipy 1.0.1 (although 1.1.0 packages are available). 

> Also, it seems to me that we use numpy 1.9.3 on Windows, while we test on 1.14? Can you check?

We build against numpy 1.9 for wider binary compatibility. The tests then run with 1.14. This is also the version in the installers.

pavlin-policar:
Perhaps it would then be best if we wrap the `mode` function in `statistics.utils` and make it do what we want, and eventually remove it when it isn't needed anymore.

PR #3540, opened 2019-01-17
[ENH] Naive Bayes: Implement predict, fix predict_storage

janezd:
Implement `predict` to:

- avoid numpy warning about passing generator to sum
- loop over columns instead of over rows (usually better + more friendly to pandas sometime in the future)
- natively support sparse matrices

`predict_storage` now calls `predict` when possible.

`predict_storage` also didn't work for `Instance`. Fixed.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/classification/naive_bayes.py
+++ b/Orange/classification/naive_bayes.py
@@ -48,10 +48,10 @@ class NaiveBayesModel(Model):
         self.class_prob = class_prob
 
     def predict_storage(self, data):
+        if isinstance(data, Instance):
+            data = Table(np.atleast_2d(data.x))
         if type(data) is Table:  # pylint: disable=unidiomatic-typecheck
             return self.predict(data.X)
-        if isinstance(data, Instance):
-            return Table(np.atleast_2d(data.x))
 
         if not len(data) or not len(data[0]):
             probs = np.tile(self.class_prob, (len(data), 1))
--- a/Orange/classification/naive_bayes.py
+++ b/Orange/classification/naive_bayes.py
@@ -1,7 +1,8 @@
 import numpy as np
+import scipy.sparse as sp
 
 from Orange.classification import Learner, Model
-from Orange.data import Instance, Storage
+from Orange.data import Instance, Storage, Table
 from Orange.statistics import contingency
 from Orange.preprocess import Discretize, RemoveNaNColumns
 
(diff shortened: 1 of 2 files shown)


Orange/classification/naive_bayes.py
```diff
@@ -47,23 +48,81 @@ def __init__(self, log_cont_prob, class_prob, domain):
...
 
     def predict_storage(self, data):
+        if type(data) is Table:  # pylint: disable=unidiomatic-typecheck
+            return self.predict(data.X)
         if isinstance(data, Instance):
-            data = [data]
-        if len(data.domain.attributes) == 0:
+            return Table(np.atleast_2d(data.x))
+
+        if not len(data) or not len(data[0]):
```
janezd: `data` can be a `Storage` or a numpy array or, in principle, any sequence of iterables over floats. Thence `len(data[0])`. The original test would (almost) always crash; apparently that line was never executed.

Orange/classification/naive_bayes.py
```diff
@@ -47,23 +48,81 @@ def __init__(self, log_cont_prob, class_prob, domain):
...
                 np.log(self.class_prob) +
-                np.array([np.zeros_like(self.class_prob)
-                          if isnan(ins.x).all() else
-                          np.sum(attr_prob[:, int(attr_val)]
-                                 for attr_val, attr_prob in zip(ins, self.log_cont_prob)
-                                 if not isnan(attr_val))
-                          for ins in data]))
+                np.array([
+                    zeros if isnan(ins.x).all() else
+                    sum(attr_prob[:, int(attr_val)]
```
janezd: This was changed because of pending deprecation warning about `np.sum` over a generator.

Orange/tests/test_naive_bayes.py
```diff
@@ -2,39 +2,42 @@
...
+    def setUp(self):
+        self.model = self.learner(self.data)
+
     def test_NaiveBayes(self):
         results = CrossValidation(self.table, [self.learner], k=10)
         ca = CA(results)
         self.assertGreater(ca, 0.7)
         self.assertLess(ca, 0.9)
 
-    def test_predict_single_instance(self):
```
janezd: I removed these tests because they test only whether the model does not crash and whether it returns two things when called with `self.model.ValueProbs`. New tests test this and more.

Orange/tests/test_naive_bayes.py
```diff
@@ -53,3 +56,160 @@ def test_allnan_cv(self):
...
+        values, probs = self.model.predict_storage(self.data[50:])
+        values2, probs2 = self.model.predict_storage(data2[50:])
+        np.testing.assert_equal(values, values2)
+        np.testing.assert_equal(probs, probs2)
+
+    def test_predictions(self):
+        self._test_predictions(sparse=None)
+
+    def test_predictions_csr_matrix(self):
+        with warnings.catch_warnings():
```
janezd: I preferred silencing this warnings (which will be removed once `scipy.sparse` stops using the matrix subclass or silences the warnings there) here rather than in `_test_predictions` to save horizontal space there.

janezd:
I removed almost all existing tests but I kept the testing of CA as an (admittedly insensitive) sanity check.

markotoplak:
Good to have good tests (I did not redo your calculations though), thanks.

It must have been a little late while you were coding this. :) I added two commits.  I added the code from the last commit just for my testing, but then I added it, because it makes tests more explicit.

Orange/classification/naive_bayes.py
```diff
@@ -48,10 +48,10 @@ def __init__(self, log_cont_prob, class_prob, domain):
         self.class_prob = class_prob
 
     def predict_storage(self, data):
+        if isinstance(data, Instance):
+            data = Table(np.atleast_2d(data.x))
```
janezd: Thanks. I somehow thought I've done this correctly. (But I have no good explanation for `assert_not_called` in tests. :))

PR #3541, opened 2019-01-18
[WIP] (store|retrieve)SpecificSettings called by non-context settings

markotoplak:
##### Issue
storeSpecificSettings was only called with a ContextHandler. Before, if was called after packing normal settings and before packing context settings.

In #3529 I implemented a special SettingsHandler because I needed to prepare settings for saving just before saving them.

##### Description of changes
storeSpecificSettings is not called in SettingsHandler

##### TODO
- [ ] test if they are really called before packing
- [ ] retrieveSpecificSettings

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcolor.py
+++ b/Orange/widgets/data/owcolor.py
@@ -367,6 +367,8 @@ class OWColor(widget.OWWidget):
 
     def storeSpecificSettings(self):
         # Store the colors that were changed -- but not others
+        if self.current_context is None:
+            return
         self.current_context.disc_data = \
             [(var.name, var.values, "colors" in var.attributes and var.colors)
              for var in self.disc_colors]
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -434,6 +434,8 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
         return text
 
     def storeSpecificSettings(self):
+        if self.current_context is None:
+            return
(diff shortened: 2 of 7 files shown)


ales-erjavec:
Put Color widget on canvas, delete it without any input connections:
```
  File "/Users/aleserjavec/workspace/orange3/Orange/canvas/scheme/widgetsscheme.py", line 795, in eventFilter
    widget.saveSettings()
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/widget.py", line 899, in saveSettings
    self.settingsHandler.update_defaults(self)
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/settings.py", line 684, in update_defaults
    widget.storeSpecificSettings()
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/data/owcolor.py", line 372, in storeSpecificSettings
    for var in self.disc_colors]
AttributeError: 'NoneType' object has no attribute 'disc_data'
```

markotoplak:
@ales-erjavec, thank you. So the problem is storeSpecificSettings wes only called before if the context was valid.

Therefore I needed to add checks for context validity in OWFile and OWColor. So if we go forward with this we risk things breaking. I quickly searched and no add-on on my disk is using this methods.

Would it be nicer if we deprecated storeSpecificSettings and used a signal "aboutToSaveData" instead?

markotoplak:
I give up. A different attempt, with a signal, is in #3567.

PR #3545, opened 2019-01-21
[ENH] Improve Save widget's gui

### LINKED ISSUE #3539: Improve the layout of OWSave
I nominate this widget for the ugliest widget in Orange core. Suggestions for improvements:

- Compression doesn't need a checkbox; "None" (or "No compression") can be an option in the combo. (Don't forget to migrate settings).
- Form layout shouldn't override the default system alignment of labels.
- Save buttons are too wide.
- The checkbox beside the save button is rather unimpressive, and does not have the same meaning as elsewhere. It tells whether the widget reacts after receiving new data and not (as is assumed elsewhere) to changes in the widget. This can be resolved by replacing it with a checkbox with a label, something like "Autosave upon receiving new data".

##### Expected behavior

Looks nice.

##### Actual behavior

Looks ugly. 

Open it.

Rather not.

janezd:
##### Issue

Fixes #3539.

OWSave widget didn't age well.

screenshot 2019-01-21 at 10 26 23

- **Choosing the format in a separate combo box instead of in the save dialog?** The GUI of its previous incarnation was clear to his creator (me), but I would have to explain it at workshops - not a good sign. But it made sense for as long as one set the format with "Save as". The current incarnation required setting the format in a combo and then use Save as (or Save) to set the name -- the save dialog was passed the filter chosen in the combo. Unusual and cumbersome.

- **Compression options.** I suppose this complication was needed because of the "Compression" checkbox, which the user would have to decide before clicking Save (and it's availability depended on the format). The widget offered three different compressions, one of which was even new to me. Gzip should suffice.

- **Auto save checkbox didn't follow the usual semantics.** Auto-apply buttons apply the changes in the widgets, while input data is always handled, disregarding the checkbox. In this widget, it was the opposite.

- **Ugly Save and Save As buttons.** The previous layout had a single line, with Save button, that included the file name, being arbitrarily long. Additional controls required a more standard layout, resulting in ugly wide buttons.

- **Selection of filters depended on data sparsity..** Sparse data was handled by reducing the number of available formats in the combo. The code for this was complicated and could, I suppose, switch the file format behind the user's back. Perhaps a better way is to just show an user error message.
  
- **Mess under the hood.** Contrived GUI resulted in contrived code.  (This was not a result of a single coder; every change made the widget more complex.) One notable self-inflicted injury was that the widget had to manage the base name (chosen in the save dialog - extension was thrown away!) and then two extensions (file format and compression) controlled by the checkboxes. To maintain this `update_extension` was called in various places. The code became dangerous to touch.

##### Description of changes

screenshot 2019-01-21 at 10 24 56

screenshot 2019-01-21 at 10 24 42

I decided to discard Save As, which confused the user and have "Set File Name" instead. This also sets the file format. Visibility of checkboxes for compression and type annotations depend upon the format (I generally oppose hiding controls because they change the widget size and confuse the user, but in this case it does no such harm.)

The function of the auto save checkbox is stated explicitly.

There is only a single option for compression.

I also added report and info in the status bar.

The code is now rather simple -- the widget contains a constructor, input data handler, two handlers for push buttons, a method that updates GUI and a method checking whether to enable saving.

Setting migrations are minimal (one renamed setting).

**There are two commits in the PR.** The first contains an interim design that mostly just shuffled the controls in the interface. The picture below does not contain the save button - it appears after the file is set (with Save As). While this looks nice, it still contains all the problems listed above and was impossible to improve on. I will squash the commits before the PR is merged.

screenshot 2019-01-21 at 10 30 28

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -987,7 +987,7 @@ class ExcelReader(FileFormat):
     """Reader for excel files"""
     EXTENSIONS = ('.xlsx',)
     DESCRIPTION = 'Microsoft Excel spreadsheet'
-    SUPPORT_COMPRESSED = True
+    SUPPORT_COMPRESSED = False
     SUPPORT_SPARSE_DATA = False
 
     def __init__(self, filename):
--- a/Orange/widgets/data/owsave.py
+++ b/Orange/widgets/data/owsave.py
@@ -196,6 +196,23 @@ class OWSave(widget.OWWidget):
     def _extension_from_filter(selected_filter):
         return re.search(r".*\(\*?(\..*)\)$", selected_filter).group(1)
 
+    def _valid_filters(self):
+        if self.data is None or not self.data.is_sparse():
+            return self.filters
+        else:
+            return {filt: writer for filt, writer in self.filters.items()
+                    if writer.SUPPORT_SPARSE_DATA}
+
+    def _default_valid_filter(self):
+        if self.data is None or not self.data.is_sparse() \
+                or self.filters[self.filter].SUPPORT_SPARSE_DATA:
+            return self.filter
+        for filt, writer in self.filters.items():
+            if writer.SUPPORT_SPARSE_DATA:
+                return filt
+        # This shouldn't happen and it will trigger an error in tests
+        return None   # pragma: no cover
+
     # As of Qt 5.9, QFileDialog.setDefaultSuffix does not support double
     # suffixes, not even in non-native dialogs. We handle each OS separately.
     if sys.platform == "darwin":
@@ -210,13 +227,13 @@ class OWSave(widget.OWWidget):
             def no_suffix(filt):
                 return filt.split("(")[0] + "(*.*)"
 
-            mac_filters = {no_suffix(f): f for f in self.filters}
+            mac_filters = {no_suffix(f): f for f in self._valid_filters()}
             filename = self._initial_start_dir()
             while True:
                 dlg = QFileDialog(
                     None, "Save File", filename, ";;".join(mac_filters))
                 dlg.setAcceptMode(dlg.AcceptSave)
-                dlg.selectNameFilter(no_suffix(self.filter))
+                dlg.selectNameFilter(no_suffix(self._default_valid_filter()))
                 dlg.setOption(QFileDialog.HideNameFilterDetails)
                 dlg.setOption(QFileDialog.DontConfirmOverwrite)
                 if dlg.exec() == QFileDialog.Rejected:
@@ -232,13 +249,13 @@ class OWSave(widget.OWWidget):
                     return filename, selected_filter
 
     elif sys.platform == "win32":
-        # TODO: This is not tested!!!
+        # TODO: Behaviour of getSaveFileName on Windows is not tested!!!
         # Windows native dialog may work correctly; if not, we may do the same
         # as for macOS?
         def get_save_filename(self):  # pragma: no cover
             return QFileDialog.getSaveFileName(
                 self, "Save File", self._initial_start_dir(),
-                ";;".join(self.filters), self.filter)
+                ";;".join(self._valid_filters()), self._default_valid_filter())
 
     else:  # Linux and any unknown platforms
         # Qt does not use a native dialog on Linux, so we can connect to
@@ -271,8 +288,8 @@ class OWSave(widget.OWWidget):
         def get_save_filename(self):
             dlg = self.SaveFileDialog(
                 None, "Save File", self._initial_start_dir(),
-                ";;".join(self.filters))
-            dlg.selectNameFilter(self.filter)
+                ";;".join(self._valid_filters()))
+            dlg.selectNameFilter(self._default_valid_filter())
             if dlg.exec() == QFileDialog.Rejected:
                 return "", ""
             else:
--- a/Orange/widgets/data/owsave.py
+++ b/Orange/widgets/data/owsave.py
@@ -1,30 +1,20 @@
 import os.path
-import pathlib
 
-from AnyQt.QtWidgets import QFormLayout
 from AnyQt.QtCore import Qt
+from AnyQt.QtGui import QBrush
+from AnyQt.QtWidgets import QGridLayout
 
 from Orange.data.table import Table
-from Orange.data.io import Compression, FileFormat, TabReader, CSVReader, PickleReader, \
-    ExcelReader
+from Orange.data.io import TabReader, CSVReader, PickleReader, ExcelReader
 from Orange.widgets import gui, widget
 from Orange.widgets.settings import Setting
 from Orange.widgets.utils import filedialogs
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.widget import Input
 
-FILE_TYPES = [
-    ("{} ({})".format(w.DESCRIPTION, w.EXTENSIONS[0]),
-     w.EXTENSIONS[0],
-     w.SUPPORT_SPARSE_DATA)
-    for w in (TabReader, CSVReader, PickleReader, ExcelReader)
-]
 
-COMPRESSIONS = [
-    ("gzip ({})".format(Compression.GZIP), Compression.GZIP),
-    ("bzip2 ({})".format(Compression.BZIP2), Compression.BZIP2),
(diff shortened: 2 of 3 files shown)


ales-erjavec:
> Or it could have a signal that accepts either this or that (a union-like signal). We don't have this in Orange.

This can be achieved by defining an ad-hoc abstract meta class for the input and registering all the accepted types.

```python
InputType(abc.ABC):
    pass
InputType.register(Type1)
InputType.register(Type2)
```
This works for the inputs, but not outputs.

janezd:
Thanks for this. But I'd still avoid having such widget unless we also solve other problems, including design.

ajdapretnar:
I don't like the fact that the widget prompts me to set the name before saving in the form of a warning.

Could there be a line edit where one sets the name? A way to avoid opening a new window if not necessary? There could be a default name suggested, for example, which would speed up the saving process.

This is what Excel has:
screen shot 2019-01-25 at 10 57 13

It is not collapsed by default, but ours could be. Expanding it allows the user to navigate the file system and decide where to save the data.

janezd:
We can't implement our own dialog for selecting a file, because it would have to be different for every operating system (or even every window manager?). Even Qt doesn't have it - it has its own, non-native dialog that doesn't look good. Excel can have it, because Microsoft can afford and maintain it.

If it's just a line edit -- we could, but I'd prefer the "normal way" of selecting a file in a GUI. No ordinary user can type /usr/janez/projects/foo/bar/x.tab or ~/janez/blah/blah.

ajdapretnar:
> No ordinary user can type /usr/janez/projects/foo/bar/x.tab or ~/janez/blah/blah.

No, of course not. I meant simply writing _iris_, which would be saved with os.path + name + extension.

Speaking of which, when I edit the file name from iris.tab to, say, rozice, the file would be saved without extension, even though .tab is selected below. I hate this very much. We surely can check if the extension exists and then add it ourselves if it doesn't.

janezd:
I no longer like the idea that the widget should show only file formats that are allowed for a particular setting. If the user selects compression or type annotations and clicks Save As..., (s)he won't see the Excel format and probably won't know it exists until (s)he, by pure chance, disable Excel.

The same arguments holds even if the filtering is based only on sparsity. I prefer offering all formats but show an error that the user should select the pickle format (which is currently the only format that supports sparse data anyway).

janezd:
@ajdapretnar said that she usually just puts the data into this widget and saves it. Hence Save top left. And hence a small little tiny widget. (Compare with the screenshot at the top.)

screenshot 2019-01-31 at 23 25 35

File name is again in the button; it's clear and there's no better place where to put it. (I tried label. Ugly.) Save As... belongs beside this button, and the auto save checkbox should be immediately below.

The other two checkboxes are below. They trigger warnings if incompatible with the chosen format but don't prevent saving.

Sparse data triggers error "Use .pkl format for sparse data." if format does not support sparse data (only .pkl does).

Annotation check box can be removed; the user could chose "Tab-delimited with type annotations (.tab)" as a format in the save dialog. Gzip can't be handled similarly because of QTBUG-44227 with double extensions (.tab.gz) on mac Os.

ajdapretnar:
This is probably the best solution to our problem. Once the tests pass, it is ready to merge.

lanzagar:
A couple of issues:
- When adding the widget, it starts with an error (File name is not set) - I don't like being greeted by errors before I have a chance to do anything
- When clicking Save As it crashed - I think it was a problem with old settings, it doesn't happen now after I cleared them (but users should not get a crash in any case)
- The compression option/checkbox does not affect extensions available in the system file chooser opened by Save As. This has several downsides: can't find and click a file I want to save to, the warning/question about overwriting files is wrong, etc
- I am not really sure why we switched the position of Save / Save As buttons from bottom to top? Isn't the previous more standard and what one would expect - first you choose some parameters like compression/annotation, then you save - (western) people usually read from top to bottom and software usually has buttons that do something at the bottom.

EDIT: I managed to reproduce the settings crash:
```
  File "/home/lan/dev/orange3/Orange/widgets/data/owsave.py", line 107, in save_file_as
    data_name += self.filt_ext[self.filter]
KeyError: 'Tab-separated values (.tab)'
```

janezd:
> * The compression option/checkbox does not affect extensions available in the system file chooser opened by Save As. This has several downsides: can't find and click a file I want to save to, the warning/question about overwriting files is wrong, etc

Existing widget already had this problem: if you selected a file, then checked compression and then saved again, it could overwrite an existing file without a warning.

Apparently the only way to do this properly is to list compressed formats in the dialog (instead of having a compression checkbox) and to implement the file selection dialog for specific OSes.

- On macOS, native dialog treats "foo.csv.gz" as having an extension ".gz" and changes it to "foo.csv.gz.csv.gz" (and adds more and more of them).
- On Linux, Qt does not use native dialog but its own *which also has a bug in setting double extensions(!)*.
- On Windows, I've no idea.

On Linux, the bug was fixable by connecting to some signals. On Mac, the save dialog offers format but doesn't show (and know about) extensions. After the user selects a file, (s)he is asked whether to overwrite an existing file. If (s)he doesn't, the file dialog reappears. Behaviour is thus similar as usual. The only downside is that the extension is assigned after selecting a file (e.g., the user selects iris.foo, which changes to iris.tab if Tab delimited format is chosen). This is similar to default Windows behaviour (adding a .txt in notepad). :)

Widget is now pretty tiny.

screenshot 2019-02-24 at 22 52 58

File selection on macOS (note the absence of extensions in the filter):

screenshot 2019-02-24 at 22 53 11

File selection on Linux (note that flies with the wrong extension are not show - that's up to Qt):

screenshot from 2019-02-24 22-48-23

> When adding the widget, it starts with an error (File name is not set) - I don't like being greeted by errors before I have a chance to do anything

This error now only shows if auto save is on.

> When clicking Save As it crashed - I think it was a problem with old settings, it doesn't happen now after I cleared them (but users should not get a crash in any case)

Migrations are yet to be tested.

> I am not really sure why we switched the position of Save / Save As buttons from bottom to top? Isn't the previous more standard and what one would expect - first you choose some parameters like compression/annotation, then you save - (western) people usually read from top to bottom and software usually has buttons that do something at the bottom.

It's because I wanted the auto save checkbox under the buttons. So with @lanzagar we decided to put a box around the two checkboxes above (annotations and compression). With compression gone, I gave up and put auto save above the buttons.

**To do:**
- [ ] Test on Windows
- [ ] Test migrations

Diff coverage won't be improved (much) because of the code that deals with platform-specific file dialogs.

Orange/data/io.py
```diff
@@ -987,7 +987,7 @@ class ExcelReader(FileFormat):
     """Reader for excel files"""
     EXTENSIONS = ('.xlsx',)
     DESCRIPTION = 'Microsoft Excel spreadsheet'
-    SUPPORT_COMPRESSED = True
+    SUPPORT_COMPRESSED = False
```
janezd: This is not a change but a fix.

Orange/widgets/data/owsave.py
```diff
@@ -34,170 +19,270 @@ class OWSave(widget.OWWidget):
...
+                # If not found, uncompressed may have been erroneously set
+                # for a writer that didn't support if (such as .xlsx), so
+                # fall through to uncompressed
+            for afilter in cls.filters:
+                if ext in afilter:
+                    settings["filter"] = afilter
+                    return
+
+        if version < 2:
+            migrate_to_version_2()
```
janezd: Migrations are in a local function because the code uses `return`-s to abort migration if it fails or when it's done. Not using `return`-s would make the code overly `if`-i -- this is way simpler. However, if these were returns from `migrate_setting`, they would complicate future migrations. Hence a local function.

ajdapretnar:
I really like this. It works perfectly from what I have tested.
Only a small comment that does not have to be addressed here, but just as a note. The widget is quite small. When it reports an error, it cannot fit both the error and status on the same line, so it cuts the status. Perhaps we should ensure this doesn't happen globally.
screen shot 2019-02-28 at 11 50 04

thocevar:
Works ok on Windows with the exception that selecting a double extension clears the file name. I guess we're giving up on that and hope that Windows will start handling double extensions better.

Issue #3438, opened 2018-12-03
conda upgrade to 3.18.0 doesn't account for new dependency on scikit-learn>=0.20


stuart-cls:
##### Orange version
3.18.0

##### Expected behavior
`conda update orange3` installs working 3.18.0 version

##### Actual behavior
scikit-learn is left at 0.19.0, which doesn't satisfy the current requirements and results in many import errors for `from sklearn.impute import SimpleImputer`

##### Steps to reproduce the behavior
run `conda update orange3` on an existing anaconda install of orange3 (from conda-forge)

##### Additional info (worksheets, data, screenshots, ...)
Happened to me on both win64 and linux64

pavlin-policar:
I would note here that this is not an issue due to the `fastTSNE` dependency, but rather that the [conda recipe](https://github.com/biolab/orange3/blob/master/conda-recipe/meta.yaml#L39) was not updated to >=0.20 while the [pip dependency](https://github.com/biolab/orange3/blob/master/requirements-core.txt#L4) was.

I'm guessing that the conda recipe just needs to be updated and pushed to the Orange feedstock.

janezd:
@pavlin-policar, can this be closed?

pavlin-policar:
Yes, this was fixed in the t-SNE library.

However, the conda recipe sklearn version doesn't match the pip dependecy version. It's trivial to update it here, but I guess the changes need to be pushed to our orange3 feedstock repository as well.

PR #3550, opened 2019-01-25
Update conda recipe scikit-learn version to >=0.20

pavlin-policar:
##### Issue

https://github.com/biolab/orange3/issues/3438#issuecomment-443900519

##### Description of changes
Like I mention in a lower comment, the changes probably need to be applied to the orange3 conda forge feedstock as well.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -36,7 +36,7 @@ requirements:
     - setuptools
     - numpy         >=1.9.0
     - scipy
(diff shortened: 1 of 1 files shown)


markotoplak:
We should not forget to put these kind of changed into ales-erjavec/orange3-installers. There, we define which versions to package, but @ales-erjavec already did it two months ago.

PR #3552, opened 2019-01-25
Update openTSNE version to >=0.3

pavlin-policar:
##### Issue
Due to some breaking API changes in my t-SNE implementation, I had fixed the version to 0.2.13. There are likely not going to be any more API changes, so this shouldn't be an issue anymore.

##### Description of changes
fastTSNE has been renamed to openTSNE in the meantime. Also, I've fixed a bug here and there from v0.2.13. I had also changed the way openTSNE.initialization.random was called, so I changed this here.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/manifold.py
+++ b/Orange/projection/manifold.py
@@ -8,9 +8,9 @@ from scipy.linalg import eigh as lapack_eigh
 from scipy.sparse.linalg import eigsh as arpack_eigh
 import sklearn.manifold as skl_manifold
 
-import fastTSNE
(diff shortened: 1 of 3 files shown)


lanzagar:
Just a reminder, that the requirement needs to be changed in the following repositories too:
https://github.com/conda-forge/orange3-feedstock
https://github.com/ales-erjavec/orange3-installers

(not a blocker for merging here)

PR #3553, opened 2019-01-25
Set a fixed random state in Louvain clustering

mstrazar:
##### Issue
Set a fixed random seed by default in Louvain clustering to promote reproducible results in scOrange.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
Wasn't this fixed in #3492 ?
Are you trying to make it deterministic by default from scripts as well or is using the widget still not deterministic?

mstrazar:
Correct,  I had another source of uncertainty somewhere.

PR #3568, opened 2019-02-01
Orange.base.Model: Return 1d predictions for 1d data

### LINKED ISSUE #3561: Should models really return 2d probabilities for Instance?
`Model.__call__(self, inst)`, where `inst` is an `Instance`, returns a 2d array of probabilities with a single row. I find this weird. Shouldn't this be a single vector?

It is not inconceivable that nobody actually calls a model with a single instance, which would explain why nobody noticed this.

janezd:
##### Issue

Fixes #3561.

Interestingly, documentation for scripting indexed the result from the model (`[0]`) without explaining why.

##### Description of changes

`Instance` is not the only kind of 1d data that can be passed to `Model.__call__`: one can also pass a 1d array, list or tuple. The trick with recursion was thus inconvenient, also because the method can return either a single value or a pair, in which case going back to 1d requires taking the first element of each element in the pair.

The method now remembers whether the data was 1d, and turns it back into 1d at the end.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -271,10 +271,7 @@ class Model(Reprable):
             if not isinstance(data[0], (list, tuple)):
                 data = [data]
                 one_d = True
-            data = Table(
-                self.original_domain,
-                data,
-                np.zeros((len(data), len(self.original_domain.class_vars))))
+            data = Table.from_list(self.original_domain, data)
             data = data.transform(self.domain)
             prediction = self.predict_storage(data)
         else:
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -209,12 +209,16 @@ class Model(Reprable):
     Probs = 1
     ValueProbs = 2
 
-    def __init__(self, domain=None):
+    def __init__(self, domain=None, original_domain=None):
         if isinstance(self, Learner):
             domain = None
         elif domain is None:
             raise ValueError("unspecified domain")
         self.domain = domain
+        if original_domain is not None:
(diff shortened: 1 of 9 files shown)


Orange/base.py
```diff
@@ -209,12 +209,16 @@ class Model(Reprable):
     Probs = 1
     ValueProbs = 2
 
-    def __init__(self, domain=None):
+    def __init__(self, domain=None, original_domain=None):
```
markotoplak: When would `original_domain` be different than domain? Isn't the point of `domain` to specify the original domain in which the model was built? (I know that I was among the people editing this stuff, but I can not remember it now)

markotoplak: Ok, now I get it: domain before preprocessing. :)

Orange/base.py
```diff
@@ -260,7 +270,8 @@ def __call__(self, data, ret=Value):
         elif isinstance(data, (list, tuple)):
             if not isinstance(data[0], (list, tuple)):
                 data = [data]
-            data = Table(self.original_domain, data)
+                one_d = True
+            data = Table.from_list(self.original_domain, data)
```
markotoplak: @janezd, I used a different workaround here.

janezd:
@markotoplak, I approve your changes. Merge if you will.

Appveyor failed when trying to upload something. Restart it if you want, or ignore.

PR #3570, opened 2019-02-01
[ENH] t-SNE: Add Normalize data checkbox

### LINKED ISSUE #3448: t-SNE: PCA preprocessing normalization
##### Orange version

master

##### Issue

We almost always want to perform PCA preprocessing before handing off the data to t-SNE. The issue now becomes that sometimes we want to run PCA preprocessing with normalized data. This can be addressed by manually inserting a "Preprocess" widget that standardizes the data, but that requires the user to actually *know* that they want to do that.

One idea would be to split the preprocessing step into a separate settings box (in the left settings toolbar) which would group together the number of PCA components and a checkbox indicating whether to perform normalization. The problem with this is that now there's even more parameters than before.

This problem can be demonstrated with the *wine* data set as shown below

image

Without data normalization
image

With data normalization:
image

pavlin-policar:
##### Issue
Fixes #3448

##### Description of changes
- The widget was still using SVD on sparse data. I have changed this to PCA, since our PCA does support sparse data since #3532
- I basically copied over the functionality from the PCA widget. PCA currently doesn't support normalization on sparse data, so this option is disabled on sparse data, just like in the PCA widget.

- Lastly, I increased the error margin on one test. The test does this: it embeds a data set using t-SNE then embeds the same data onto the existing embedding (transform). Unfortunately, the points are bound to be jittered around each original corresponding point, but also how far away is also determined by the neighborhoods, so there's no clean way to check for this. But if we visualize what this actually produces, we can see that the result is still correct. And given that the space spans from -20 to 20, increasing the error margin from 1 to 3 shouldn't really impact results too much.
image

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owtsne.py
+++ b/Orange/widgets/unsupervised/owtsne.py
@@ -7,7 +7,7 @@ from AnyQt.QtWidgets import QFormLayout
 
 from Orange.data import Table, Domain
 from Orange.preprocess.preprocess import Preprocess, ApplyDomain
-from Orange.projection import PCA, TSNE, TruncatedSVD
+from Orange.projection import PCA, TSNE
 from Orange.projection.manifold import TSNEModel
 from Orange.widgets import gui
 from Orange.widgets.settings import Setting, SettingProvider
@@ -216,11 +216,9 @@ class OWtSNE(OWDataProjectionWidget):
         self.__set_update_loop(self.tsne_iterator)
 
     def pca_preprocessing(self):
-        if self.pca_data is not None and \
-                self.pca_data.X.shape[1] == self.pca_components:
+        if self.pca_data is not None and self.pca_data.X.shape[1] == self.pca_components:
             return
-        cls = TruncatedSVD if self.data.is_sparse() else PCA
-        projector = cls(n_components=self.pca_components, random_state=0)
+        projector = PCA(n_components=self.pca_components, random_state=0)
         model = projector(self.data)
         self.pca_data = model(self.data)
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/unsupervised/owtsne.py
```diff
@@ -215,12 +226,25 @@ def stop(self):
...
+        super().set_data(data)
+
+        if data is not None:
+            # PCA doesn't support normalization on sparse data, as this would
+            # require centering and normalizing the matrix
+            self.normalize_cbx.setDisabled(data.is_sparse())
+
     def pca_preprocessing(self):
-        if self.pca_data is not None and \
-                self.pca_data.X.shape[1] == self.pca_components:
```
pavlin-policar: I removed this check because I changed line 146 to invalidate the PCA projection, therefore it will be set to `None` whenever the number of components changes.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -215,12 +226,32 @@ def stop(self):
...
+        super().set_data(data)
+
+        if data is not None:
+            # PCA doesn't support normalization on sparse data, as this would
+            # require centering and normalizing the matrix
+            self.normalize_cbx.setDisabled(data.is_sparse())
+            if data.is_sparse():
+                self.normalize = False
+                self.normalize_cbx.setToolTip(
+                    "Data normalization is not supported on sparse matrices."
```
lanzagar: Why is normalization not supported on sparse data?

pavlin-policar: Normalization for sparse data is currently not supported anywhere where PCA is used. As far as I can tell, this includes the PCA, t-SNE and Louvain clustering widget.

I would prefer to add normalization support in a separate PR, adding it everywhere where it's missing, and leave this as is.

lanzagar:
> I would prefer to add normalization support in a separate PR, adding it everywhere where it's missing, and leave this as is.

Fine with me.

PR #3571, opened 2019-02-02
[FIX] ScatterplotGraph: Use mapToView instead of mapRectFromParent

### LINKED ISSUE #2789: Scatter Plot: selection is not working with very high feature values
##### Orange version
3.8.0.dev0+301d13a

##### Expected behavior
If a feature value is much higher than the rest of the data, scatter plot shall perform normally.

##### Actual behavior
If a feature value is much higher than the rest of the data, scatter plot does not allow point selection. Mouse rectangle selection is also not working ok (e.g. in my two instances dataset if I select the instance with the extra-high value with rectangle, the second instance is selected :) ).

##### Steps to reproduce the behaviour

Load attached data, go with them to Scatter Plot, try to select data

screen shot 2017-11-21 at 17 19 06

[data.xlsx](https://github.com/biolab/orange3/files/1492119/data.xlsx)

janezd:
##### Issue

Fixes #2789.

##### Description of changes

- To show selection rectangle, (pyqtgraph's) `ViewBox.updateScaleBox` is overloaded to use `self.mapToView` instead of `self.childgroup.mapRectFromParent`. This may not be equivalent if there are multiple viewboxes in the same view or something similarly complicated, but we don't have this.
- To fix selection, (our) `mouseDragEvent` is modified in the same way
- An unrelated fix: `select_by_rectangle` assumed that the top left corner is above and to the left of the bottom right. This was true only if the user dragged in this direction. `select_by_rectangle` now sorts the coordinates.

The PR doesn't pass the diff coverage threshold, but I won't write any tests because they would involve simulating user interaction with mouse, yet at the end there wouldn't be anything to actually test.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -1233,10 +1233,10 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         if self.scatterplot_item is not None:
             self.select(points)
 
-    def select_by_rectangle(self, value_rect):
+    def select_by_rectangle(self, rect):
         if self.scatterplot_item is not None:
-            x0, y0 = value_rect.topLeft().x(), value_rect.topLeft().y()
-            x1, y1 = value_rect.bottomRight().x(), value_rect.bottomRight().y()
+            x0, x1 = sorted((rect.topLeft().x(), rect.bottomRight().x()))
+            y0, y1 = sorted((rect.topLeft().y(), rect.bottomRight().y()))
             x, y = self.master.get_coordinates_data()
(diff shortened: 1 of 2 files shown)


VesnaT:
The selection now works, but zoom does not. The `mouseDragEvent` should probably be modified for zooming as well.

janezd:
@VesnaT, done. Please check (assuming CI tests pass).

Interestingly, pyqtgraph calls `mapToView` when showing the rectangle, which is why it was shown correctly while dragging, but then it calls the broken `mapRectFromParent` at `ev.finish`. This is why it sufficed to reimplement only a part of `mouseDragEvent`.

PR #3574, opened 2019-02-02
[ENH] Use %g (including sci notation) if number of decimals is not set

### LINKED ISSUE #2692: Small values
##### Orange version
screen shot 2017-10-18 at 11 15 27

##### Expected behavior
Small values shall be visible in exp format: e.g. 0.000001 as  1e-6

##### Actual behavior
Value 0.000001 is shown as 0

open attached excel file 

check Data Table 

check Scatter Plot with  show labels

[small-numbers.xlsx](https://github.com/biolab/orange3/files/1394064/small-numbers.xlsx)
scatter-plot
data-table

janezd:
##### Issue

Fixes #2692.

Numeric variables are currently printed with three decimals, unless the number is set to another default. The latter most often happens when the data is read from text file (tab, csv, but not Excel). Consequently, values close to 0 are printed as 0. To fix this, some code sets the number of decimals to 7, which defeats the purpose and even doesn't help for values below 1e-7.

##### Description of changes

- When number of decimals is not set (as in regression coefficients, variables constructed by PCA...), `%g` is now used for formatting, as suggested by @markotoplak. This will print a reasonable number of decimals or switch to scientific notation.
- When the number of decimals is set, nothing changes.

Some code used `Variable.number_of_decimals` to construct own format strings instead of calling `Variable.val_to_str`. `Variable` now exposes `format_str`, which can be either `%g` or `%f` with prescribed number of decimals. I tried to hunt down the code where this change was appropriate.

The change is backward incompatible in the sense that some outputs changed. Some tests had to be fixed accordingly.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/classification/simple_tree.py
+++ b/Orange/classification/simple_tree.py
@@ -274,35 +274,36 @@ class SimpleTreeModel(Model):
                 node = self.node
         n = node.contents
         if self.type == Classification:
-            decimals = 1
+            format_str = format_leaf = format_node = None
         else:
-            decimals = self.domain.class_var.number_of_decimals
+            format_str = f"({self.domain.class_var.format_str}: %s)"
+            format_leaf = " --> " + format_str
+            format_node = "%s " + format_str
         if n.children_size == 0:
             if self.type == Classification:
-                node_cont = [round(n.dist[i], decimals)
+                node_cont = [round(n.dist[i], 1)
                              for i in range(self.cls_vals)]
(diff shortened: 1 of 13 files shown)


markotoplak:
I suspected problems with saving into a file and reloading it, because %g sometimes skips decimals. I found none: our file saving is smart enough to directly use .value, not __str__.  :)

Issue #2943, opened 2018-03-06
Naive Bayes is confounded by variable reuse.


ales-erjavec:
##### Orange version

3.11.dev, 3.10, 3.9, 3.8 ...

##### Expected behavior

Can run learner performance evaluation sequentially in the same session on different datasets and get predictable deterministic results (not withstanding explicit randomness sources).

##### Actual behavior

Cannot run *Test & Score* on anneal.tab and glass.tab (both included in the distribution) sequentially without one affecting the other.

A by product of the Variable reuse.

Also related gh-2500

##### Steps to reproduce the behavior

<img width="259" alt="screen shot 2018-03-06 at 09 53 28" src="https://user-images.githubusercontent.com/4716745/37022813-66c0db8c-2124-11e8-9a56-8347e8857197.png">

In a fresh session load the anneal.tab in the *File* widget. Observe and record the reported AUC (10 fold cross validation, stratified, averaged over all classes); reload as many times as necessary to establish that the AUC is stable: 0.979.

Load glass.tab and then anneal.tab again. The AUC for Naive Bayes is different: 0.844.

Presumably the problem is in variable reuse. The first time anneal.tab is loaded y has values 1, 2, 3, 5, U the second time 1, 2, 3, 5, U, 6, 7 (the values from glass.tab#y are merged).

This is not a problem in general for skl-learners that use LabelEncoder to encode targets (and ignore the values not actually present in the target vector), but NaiveBayes uses the full y domain for Laplace estimation of class probabilities.

```python
import Orange
anneal = Orange.data.Table("anneal")
l1 = Orange.classification.NaiveBayesLearner()(anneal)
print(l1.class_prob)
_ = Orange.data.Table("glass")
l2 = Orange.classification.NaiveBayesLearner()(anneal)
print(l2.class_prob)
```

In Orange 2 at least there was on option to disable the variable reuse. Never the less, the default should be no reuse.

lanzagar:
Variable reuse can be problematic in unexpected places and should be carefully (re)considered.

But until something is changed on that front (might take some time?) we could/should change naive bayes to not use all variable values, but just those found in training data (as all other learners?).
Looks like `class_freq` is obtained with `get_contingency` so it might be as easy as filtering out class values with 0 representation after that is computed? Might need a fix in Model as well to not crash on unseen values?

janezd:
+1 for solution for NB.

[...]

PR #3575, opened 2019-02-02
[FIX] Naive Bayes: Ignore existing classes in Laplacian smoothing

janezd:
##### Issue

As suggested by @lanzagar, issue #2943 can be fixed by ignoring empty classes in Laplacian smoothing.

Plainly, say that `y` has values `male`, `female`, `yes` and `no` because `y` appeared in two different data sets loaded in the same session. Of course, only two values appear in each set. With this PR, only 2 is added to denominators in computation of probabilities, and the probabilities of empty classes are set to 0.

Variable reuse remains a large open issue.

##### Description of changes

The change also required a minor change in computation of probabilities: instead of computing `exp(log(class_probs) + sum(conditional_probs))`, the code now computes `class_probs * exp(sum(conditional_probls))` because `class_probs` can now be 0. This should not affect numerical stability.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/classification/naive_bayes.py
+++ b/Orange/classification/naive_bayes.py
@@ -33,11 +33,23 @@ class NaiveBayesLearner(Learner):
         cont = contingency.get_contingencies(table)
         class_freq = np.array(np.diag(
             contingency.get_contingency(table, table.domain.class_var)))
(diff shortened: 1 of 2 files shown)


lanzagar:
Looks good to me.
It still has the [RFC] tag, but unless others have additional comments, I am fine with changing this to [FIX] and merging.

PR #3576, opened 2019-02-02
[ENH] DomainEditor: Indicate changed variables with bold face font

### LINKED ISSUE #2673: File: Mark modifications in DomainEditor
There is no clear indication whether the header has been modified after the file has been loaded. (in the Columns part of the widget)

The least we could do is to indicate which rows have been modified, either with color or some other indicator (checkbox?). We could also use a Reset button which resets variables to their default values, it should be disabled if no modifications have been made.

Modifications are done manually by the user or restored from settings (when a dataset that has been modified is loaded again).

### LINKED ISSUE #3563: Domain Editor in File widget should have a Reset button

janezd:
##### Issue

Fixes #2673 and fixes #3563.

##### Description of changes

Model remembers the original data and implements `FontRole` to return a bold font if the line was changed.

I sincerely don't know how to reasonably test this. It's all just a game of signals.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -215,6 +215,8 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
             callback=lambda: self.browse_file(True), autoDefault=False)
         gui.rubber(box)
 
+        gui.button(
+            box, self, "Reset", callback=self.reset_domain_edit)
         self.apply_button = gui.button(
             box, self, "Apply", callback=self.apply_domain_edit)
         self.apply_button.setEnabled(False)
@@ -439,6 +441,10 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
     def retrieveSpecificSettings(self):
         if hasattr(self.current_context, "modified_variables"):
             self.variables[:] = self.current_context.modified_variables
+
+    def reset_domain_edit(self):
+        self.domain_editor.reset_domain()
+        self.apply_domain_edit()
 
     def apply_domain_edit(self):
         if self.data is None:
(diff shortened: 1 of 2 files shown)


lanzagar:
Does not work correctly for me.
After some clicking, currently it fails even just opening the File widget and changing sepal width to meta (it does not become bold), then clicking Reload makes it bold and combinations of changes with clicking Reset, Apply, Reload almost always lead to inconsistencies.
(I cleared widget settings and it still fails as early as described above, which is strange, because at first I though it was working fine, until some random clicking).

janezd:
It's amazing that it worked for me. The problem was in a list that had to be deepcopied, not just copied.

I noticed that changing a variable's position (e.g. feature -> meta) is correctly indicated with bold font, but has no effect on the output. This is an unrelated problem that already exists on master.

janezd:
LOL, changing feature -> meta of course works. I was observing the output on Select columns, which just reversed any change -- because that's exactly what it does. :)

ajdapretnar:
This works well for me after a short test. However, changes are indicated with italics not bold. I like it that way, but we could make it clear in the PR as well.

janezd:
```
                font = QFont()
                font.setBold(True)
```

gives italic!? It's bold on my machine.

ajdapretnar:
Sorry, I was being stupid. I was looking at Edit Domain, I obviously didn't read the PR title well enough. It is indeed in bold. And it works, too. I mean the one in File widget.

markotoplak:
@janezd, I particularly like that reset button. Looks cool. :)

PR #3580, opened 2019-02-04
OwTSNE: Don't catch errors during optimization

### LINKED ISSUE #3548: TSNE widget error handling catches too many errors
##### Orange version
master

##### Expected behavior
Our programming errors (for example due to incompatible version of library, as in biolab/orange3-single-cell#323) show the error window so that the error can be reported.

##### Actual behavior
All errors are shown as errors in the widget, which makes them hard to report.

biolab/orange3-single-cell#323

##### Notes
TSNE widget is wrongly catching `Exception`.

pavlin-policar:
##### Issue
Fixes #3548.

Catching errors during optimization and showing them in the error
message bar makes them difficult to report. Furthermore, no exceptions
should be thrown during optimization and if something does go wrong, we
need to fix it.

##### Description of changes
Remove the `optimization_error`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owtsne.py
+++ b/Orange/widgets/unsupervised/owtsne.py
@@ -94,7 +94,6 @@ class OWtSNE(OWDataProjectionWidget):
         constant_data = Msg("Input data is constant")
         no_attributes = Msg("Data has no attributes")
         out_of_memory = Msg("Out of memory")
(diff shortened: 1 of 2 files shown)


pavlin-policar:
My best guess is that something was failing due to the way things were being mocked. This is superseded by #3604.

PR #3582, opened 2019-02-04
[ENH] PCA: Preserve f32s & reduce memory footprint when computing means

pavlin-policar:
##### Issue
While working towards getting the improved PCA merged into scikit-learn, I've found two improvements.

##### Description of changes
1. `np.float32` are now preserved
2. Apparently, scipys sparse `x.mean` method isn't the most memory efficient, and scikit-learn's utility function `mean_variance_axis` is much better (see benchmarks [here](https://github.com/scikit-learn/scikit-learn/pull/12841#issuecomment-460238233))

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/pca.py
+++ b/Orange/projection/pca.py
@@ -1,9 +1,9 @@
 import numbers
-
 import six
 import numpy as np
 import scipy.sparse as sp
 from scipy.linalg import lu, qr, svd
+
 from sklearn import decomposition as skl_decomposition
 from sklearn.utils import check_array, check_random_state
 from sklearn.utils.extmath import svd_flip, safe_sparse_dot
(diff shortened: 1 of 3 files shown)


Orange/projection/pca.py
```diff
@@ -45,10 +46,21 @@ def randomized_pca(A, n_components, n_oversamples=10, n_iter="auto",
...
     n_samples, n_features = A.shape
 
-    c = np.atleast_2d(A.mean(axis=0))
+    # Use `mean_variance_axis` instead of `A.sum` as it is much more memory
+    # efficient than scipy's implementation. See GH
+    # https://github.com/scikit-learn/scikit-learn/pull/12841#issuecomment-460238233
+    if sp.issparse(A):
+        means, _ = mean_variance_axis(A, axis=0)
+    else:
+        means = np.mean(A, axis=0)
```
lanzagar: We already have `Orange.statistics.util.mean` which is supposed to handle dense and sparse matrices, but it does not have the axis parameter (unlike most other functions in that module). Maybe you could improve that function instead and use it here?

lanzagar:
If you add new code with no new tests it can still be covered by existing tests, thus increasing the coverage... ;)

Orange/projection/pca.py
pavlin-policar: That's a much better idea. Unfortunately, `mean_variance_axis` just ignores NaNs and provides no feedback if the data had any NaNs. This means that `mean` wouldn't properly handle NaNs and we'd have no good way of knowing where they occurred.

However, changing `nanmean` to use this is completely fine since this is exactly the behaviour we want there. So I did that.

PR #3585, opened 2019-02-06
[DOC] Add VizRank info to Scatter Plot

ajdapretnar:
##### Issue

Scatter plot's documentation did not sufficiently explain how feature pairs are scored.

##### Description of changes
Explain two scoring methods.

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation

### DIFF
--- a/doc/visual-programming/source/widgets/visualize/scatterplot.md
+++ b/doc/visual-programming/source/widgets/visualize/scatterplot.md
@@ -43,7 +43,11 @@ Here is an example of the **Scatter Plot** widget if the *Show class density* an
 Intelligent Data Visualization
 ------------------------------
 
-If a dataset has many attributes, it is impossible to manually scan through all the pairs to find interesting or useful scatter plots. Orange implements intelligent data visualization with the **Find Informative Projections** option in the widget. The goal of optimization is to find scatter plot projections where instances are well separated.
+If a dataset has many attributes, it is impossible to manually scan through all the pairs to find interesting or useful scatter plots. Orange implements intelligent data visualization with the **Find Informative Projections** option in the widget.
+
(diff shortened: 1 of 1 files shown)


doc/visual-programming/source/widgets/visualize/scatterplot.md
```diff
@@ -45,6 +45,10 @@ Intelligent Data Visualization
 
 If a dataset has many attributes, it is impossible to manually scan through all the pairs to find interesting or useful scatter plots. Orange implements intelligent data visualization with the **Find Informative Projections** option in the widget. The goal of optimization is to find scatter plot projections where instances are well separated.
 
+If categorical variable is selected in the Color section, the option will score variable combinations based on how well class values are separated. For each data instance, the scoring will check 10 closest neighbors and give a higher score to those combinations, where closest neighbor are of the same value.
```
janezd: It is not clear that distances are measured in the projected space.
I don't like "of the same value". Maybe "have the same value"? The text also refers to "class value", although the color is not necessarily class. Also, "separation" is repeated from the last paragraph and thus doesn't add any new information. Maybe:

*If a categorical variable is selected in the Color section, the score is computed as follows. For each data instance, VizRank finds 10 nearest neighbours in the projected space, that is, on the figure as seen by the user. It then checks how many of them have the same color. The total VizRank score of the projection is then the average number of same-colored neighbours.*

*Computation for continuous colors is similar, except that the [coefficient of determination](https://en.wikipedia.org/wiki/Coefficient_of_determination) is used for measuring the local homogeneity of the projection.*

janezd: Or, more compact:

*For each data instance, VizRank finds 10 nearest neighbours in the projected space, that is, on the figure as seen by the user. If a categorical variable is selected for coloring data points, VizRank checks how many neighbours have the same color. The total VizRank score of the projection is then the average number of same-colored neighbours. Computation for continuous colors is similar, except that the [coefficient of determination](https://en.wikipedia.org/wiki/Coefficient_of_determination) is used for measuring the local homogeneity of the projection.*

markotoplak:
Perhaps it would be useful to also add a reference to the VizRank paper.

ajdapretnar:
@janezd Fixed.

PR #3588, opened 2019-02-07
[ENH] Preprocess: implement Select Relevant Feature's percentile

### LINKED ISSUE #3226: Preprocess: Select Relevant Feature's percentile still off
##### Orange version

3.16.dev

##### Expected behavior
Percentile works and has a tooltip.

##### Actual behavior
Percentile is off for 4 years now. We should make it work or remove it.

While at it, it would be nice to rename 'Strategy' to something more intuitive. 'Number of features' for example.

matejklemen:
##### Issue

Fixes #3226.

##### Description of changes
Implements the "Percentile" option for Select Relevant Features preprocessor.  
The option enables the selection of all features that fall >= the entered percentile. The rank of the first feature to be taken is computed using [nearest-rank method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method). There are 2 "special" cases: 0th percentile is defined as all the features and 100th percentile is defined as the best feature.  
  
I've also renamed "Strategy" to "Number of features" in Select Random Features and Select Relevant Features as was suggested in the original issue.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/fss.py
+++ b/Orange/preprocess/fss.py
@@ -16,10 +16,10 @@ __all__ = ["SelectBestFeatures", "SelectRandomFeatures"]
 class SelectBestFeatures(Reprable):
     """
     A feature selector that builds a new dataset consisting of either the top
-    `k` features (`k` can also represent a percentile value between 0.0 and 1.0)
-    or all those that exceed a given `threshold`. Features are scored using the
-    provided feature scoring `method`. By default it is assumed that feature
-    importance diminishes with decreasing scores.
+    `k` features (if `k` is an `int`) or a proportion (if `k` is a `float`
+    between 0.0 and 1.0), or all those that exceed a given `threshold`. Features
+    are scored using the provided feature scoring `method`. By default it is
+    assumed that feature importance decreases with decreasing scores.
 
     If both `k` and `threshold` are set, only features satisfying both
     conditions will be selected.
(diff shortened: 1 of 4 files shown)


Orange/preprocess/fss.py
```diff
@@ -50,6 +51,12 @@ def __init__(self, method=None, k=None, threshold=None, decreasing=True):
         self.decreasing = decreasing
 
     def __call__(self, data):
+        n_attrs = len(data.domain.attributes)
+        if isinstance(self.k, float):
+            idx_attr = np.ceil(self.k * n_attrs).astype(int)
+            # edge case: 0th percentile would result in selection of `(n_attrs + 1)` attrs
+            self.k = min(n_attrs - idx_attr + 1, n_attrs)
```
janezd: `__call__` shouldn't change the state of the object (unless when caching, for instance). In this case, imagine you have

```
fss = SelectBestFeatures(k=0.5)
data1 = <some data set with 10 attributes>
data2 = <some data set with 100 attributes>
fss(data1)  # this sets fss.k to 5!
fss(data2)  # and so this selects just 5 instead of 50 attributes
```

One option is to introduce

```
            effective_k = min(n_attrs - idx_attr + 1, n_attrs)
    else:
            effective_k = self.k
```

and replace further occurrences of `self.k` in this method with `effective_k`.

matejklemen: Good catch, I didn't think about that.  
  
I've fixed it (done the same thing for selection of random features as well, since it had the same problem) and added a test for this.

janezd:
I added another commit to switch from percentiles to proportion, so that `k` (in the class) and spin boxes (in the widget) have parallel meanings, and that higher numbers always mean more attributes.

janezd:
I changed percentiles to proportions, so that `k` (in the class) and combos (in widget) have parallel meaning (more is always more).

PR #3591, opened 2019-02-11
[ENH] Correlations: fixes and enhancements

### LINKED ISSUE #3504: OWCorrelations: fixes and enhancements
Fixes:
- When data changes, the output shows old data instead of the new one. Call commit when new data on input.

Enhancements:
- Split attribute pairs into two columns (in view) and elide as in Data Table.
- Add p-values to Correlations table output.
- Add option to compute all correlations with a chosen feature.
- Check heuristics. Needs discussion: @lanzagar

VesnaT:
##### Issue

Fixes #3504
Heuristic will be updates in a following PR.

##### Description of changes
- Send outputs when feed data
- Split attribute pairs into two columns
- Output p-values
- Compute correlations with chosen feature

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcorrelations.py
+++ b/Orange/widgets/data/owcorrelations.py
@@ -20,6 +20,7 @@ from Orange.statistics.util import FDR
 from Orange.widgets import gui
 from Orange.widgets.settings import Setting, ContextSetting, \
     DomainContextHandler
+from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.signals import Input, Output
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.visualize.utils import VizRankDialogAttrPair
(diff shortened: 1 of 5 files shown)


Orange/widgets/data/owcorrelations.py
```diff
@@ -142,11 +149,18 @@ def check_preconditions(self):
...
             return self.heuristic.get_states(initial_state)
         else:
             return super().iterate_states(initial_state)
 
+    def iterate_states_by_feature(self):
+        for j in range(len(self.attrs)):
+            if j != self.sel_feature_index:
+                yield self.sel_feature_index, j
+
     def state_count(self):
```
janezd: You should also change `state_count`.

```
if self.sel_feature_index is not None:
    return len(self.attrs) - 1
```

janezd:
We decided to replace p-Value with False discovery rate.

PR #3592, opened 2019-02-11
[FIX] t-SNE speed-ups

pavlin-policar:
##### Issue
t-SNE was slow

##### Description of changes
- The multiscale option makes things slow for larger data sets, so turn it off by default.
- Use faster approximation setting for Barnes-Hut
- Update t-SNE plot only at the start and at the end, avoiding potentially many redraws
- Selecting data in t-SNE was slow because `transform` was being called. Avoid this.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owtsne.py
+++ b/Orange/widgets/unsupervised/owtsne.py
@@ -73,7 +73,7 @@ class OWtSNE(OWDataProjectionWidget):
     settings_version = 3
     max_iter = Setting(300)
     perplexity = Setting(30)
-    multiscale = Setting(True)
+    multiscale = Setting(False)
     exaggeration = Setting(1)
     pca_components = Setting(20)
     normalize = Setting(True)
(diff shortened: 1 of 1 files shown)


Orange/widgets/unsupervised/owtsne.py
```diff
@@ -258,6 +265,9 @@ def pca_preprocessing(self):
     def __start(self):
         self.pca_preprocessing()
 
+        # Dirty flag indicating whether the embedding has been drawn at least once
+        self.needs_to_draw = True
```
lanzagar: This can probably be removed now. In any case the comment is no longer true.

pavlin-policar: This cannot be removed because, at the start of the optimization, we have to draw *something*, so we just draw some random points, because the graph widget *needs* some data to be shown. However, we'd like to replace this random data with something slightly more meaningful as soon as possible, so setting this to `True` here will ensure that after the first batch of iterations is done, the widget will display something better than random data.

PR #3593, opened 2019-02-11
Scatterplotgraph: Do not animate resizing

### LINKED ISSUE #3442: Scatterplotgraph: Don't animate resizing when too many points
... because it's too slow. On my computer the limit is around 500 points, I believe.

It may be a good first issue, but don't forget to write tests.

VesnaT:
##### Issue

Fixes #3442

##### Description of changes
Maximum data size to animate resizing is now set to 1000.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -37,6 +37,7 @@ from Orange.widgets.visualize.utils.plotutils import (
 
 
 SELECTION_WIDTH = 5
+MAX_N_VALID_SIZE_ANIMATE = 1000
 
 
 class PaletteItemSample(ItemSample):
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -766,7 +767,8 @@ def __call__(self):
                     widget.scatterplot_item.setSize(size)
                     widget.scatterplot_item_sel.setSize(size + SELECTION_WIDTH)
 
-            if np.sum(current_size_data) / self.n_valid != -1 and np.sum(diff):
+            if np.sum(current_size_data) / self.n_valid != -1 and \
+                    max(abs(diff)) and self.n_valid <= MAX_N_VALID_SIZE_ANIMATE:
```
lanzagar: Can we make this condition clearer?
a) `max(abs(diff))` if better then before, but `np.any(diff != 0)` would be even better / more explicit
b) As far as I can see, sizes are always positive. So the first check could just be `np.all(current_size_data > 0)` i.e. legit sizes, or you can check for default initialization directly `np.any(current_size_data == -1)`
c) I would put the last check `<= MAX_N...` in first place so that for large data other checks don't need to be performed at all.

PR #3604, opened 2019-02-15
[ENH] OWTSNE: Offload computation to separate thread

pavlin-policar:
##### Issue
The t-SNE widget previously performed all computation inside the main thread, while simulating a thread i.e. it would pause every once in a while and handle user action. For larger data sets, this meant the widget would freeze for a while until the widget received control again.

##### Description of changes

The widget now offloads all of the computation into a separate thread, and the widget is only responsible for preparing a `Task` specification, which is then handled by the other thread.

Other changes include:
- The widget now remembers partial results better, i.e. changing the exaggeration parameter doesn't recompute all the nearest neighbours again.
- More descriptive status message and the progress bar now takes into account all the steps of the procedure.
- Since t-SNE is deterministic, once the optimization is complete, pressing "Start" does nothing now, because the results will invariably the same.
- The PCA slider now changes its maximum value based on the number of features in the data.
- Add Info thing to indicate that the settings have been changed and the user should rerun the optimization.
- I've made the t-SNE API a bit more granular so that it's possible to gain access to the affinities (nearest neighbors). In principle, this is a bad idea, because this means the widget is tied to my `openTSNE` implementation, but I think it's justified because 1. in case we ever opt for a different implementation, removing this from the widget requires little effort and 2. finding the nearest neighbors can take a very long time (up to half of all the computation time for larger data sets) therefore the widget should be smart enough to cache results wherever possible.
- The widget now replots fairly frequently again. I did this because it seems that doing the work in a separate thread and drawing in the main one play well together, and the speed difference between updating often and updating rarely is small.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owtsne.py
+++ b/Orange/widgets/unsupervised/owtsne.py
@@ -279,8 +279,6 @@ class OWtSNE(OWDataProjectionWidget, ConcurrentWidgetMixin):
         self.tsne_embedding = None  # type: Optional[manifold.TSNEModel]
         self.iterations_done = 0    # type: int
 
-        self.__invalidated = False
-
     def _add_controls(self):
         self._add_controls_start_box()
         super()._add_controls()
@@ -350,7 +348,6 @@ class OWtSNE(OWDataProjectionWidget, ConcurrentWidgetMixin):
         self._set_modified(True)
 
     def _invalidate_output(self):
-        self.__invalidated = True
         self.cancel()
         self.run_button.setText("Start")
 
@@ -434,10 +431,10 @@ class OWtSNE(OWDataProjectionWidget, ConcurrentWidgetMixin):
             else:
                 self.normalize_cbx.setToolTip("")
 
-        self.run()
+        if self._invalidated:
+            self.run()
 
     def run(self):
-        self.__invalidated = False
         self._set_modified(False)
 
         # When the data is invalid, it is set to `None` and an error is set,
--- a/Orange/widgets/unsupervised/owtsne.py
+++ b/Orange/widgets/unsupervised/owtsne.py
@@ -231,6 +231,7 @@ class OWtSNEGraph(OWScatterPlotBase):
 
 
 class invalidated:
+    # pylint: disable=invalid-name
     pca_projection = affinities = tsne_embedding = False
 
     def __set__(self, instance, value):
@@ -376,6 +377,7 @@ class OWtSNE(OWDataProjectionWidget, ConcurrentWidgetMixin):
 
         # `super().check_data()` clears all messages so we have to remember if
         # it was shown
+        # pylint: disable=assignment-from-no-return
         should_show_modified_message = self.Information.modified.is_shown()
         super().check_data()
 
--- a/Orange/widgets/unsupervised/owtsne.py
+++ b/Orange/widgets/unsupervised/owtsne.py
@@ -1,60 +1,235 @@
 import warnings
+from functools import partial
+from types import SimpleNamespace as namespace
+from typing import Optional  # pylint: disable=unused-import
 
 import numpy as np
 
-from AnyQt.QtCore import Qt, QTimer
+from AnyQt.QtCore import Qt
 from AnyQt.QtWidgets import QFormLayout
 
 from Orange.data import Table, Domain
 from Orange.preprocess import preprocess
-from Orange.projection import PCA, TSNE
-from Orange.projection.manifold import TSNEModel
+from Orange.projection import PCA
+from Orange.projection import manifold
 from Orange.widgets import gui
 from Orange.widgets.settings import Setting, SettingProvider
+from Orange.widgets.utils.concurrent import TaskState, ConcurrentWidgetMixin
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.visualize.owscatterplotgraph import OWScatterPlotBase
 from Orange.widgets.visualize.utils.widget import OWDataProjectionWidget
 from Orange.widgets.widget import Msg
 
+_STEP_SIZE = 25
+_MAX_PCA_COMPONENTS = 50
+_DEFAULT_PCA_COMPONENTS = 20
+
+
+class Task(namespace):
+    """Completely determines the t-SNE task spec and intermediate results."""
+    data = None             # type: Optional[Table]
+    normalize = None        # type: Optional[bool]
+    pca_components = None   # type: Optional[int]
+    pca_projection = None   # type: Optional[Table]
+    perplexity = None       # type: Optional[float]
+    multiscale = None       # type: Optional[bool]
+    exaggeration = None     # type: Optional[float]
+    initialization = None   # type: Optional[np.ndarray]
+    affinities = None       # type: Optional[openTSNE.affinity.Affinities]
+    tsne_embedding = None   # type: Optional[manifold.TSNEModel]
+    iterations_done = 0     # type: int
+
+    # These attributes need not be set by the widget
+    tsne = None             # type: Optional[manifold.TSNE]
+
+
+def pca_preprocessing(data, n_components, normalize):
+    projector = PCA(n_components=n_components, random_state=0)
+    if normalize:
+        projector.preprocessors += (preprocess.Normalize(),)
+
+    model = projector(data)
+    return model(data)
+
+
+def prepare_tsne_obj(data, perplexity, multiscale, exaggeration):
+    # type: (Table, float, bool, float) -> manifold.TSNE
+    """Automatically determine the best parameters for the given data set."""
+    # Compute perplexity settings for multiscale
+    n_samples = data.X.shape[0]
+    if multiscale:
+        perplexity = min((n_samples - 1) / 3, 50), min((n_samples - 1) / 3, 500)
+    else:
+        perplexity = perplexity
+
+    # Determine whether to use settings for large data sets
+    if n_samples > 10_000:
+        neighbor_method, gradient_method = "approx", "fft"
+    else:
+        neighbor_method, gradient_method = "exact", "bh"
+
+    # Larger data sets need a larger number of iterations
+    if n_samples > 100_000:
+        early_exagg_iter, n_iter = 500, 1000
+    else:
+        early_exagg_iter, n_iter = 250, 750
+
+    return manifold.TSNE(
+        n_components=2,
+        perplexity=perplexity,
+        multiscale=multiscale,
+        early_exaggeration_iter=early_exagg_iter,
+        n_iter=n_iter,
+        exaggeration=exaggeration,
+        neighbors=neighbor_method,
+        negative_gradient_method=gradient_method,
+        theta=0.8,
+        random_state=0,
+    )
+
 
 class TSNERunner:
-    def __init__(self, tsne: TSNEModel, step_size=50, exaggeration=1):
-        self.embedding = tsne
-        self.iterations_done = 0
-        self.step_size = step_size
-        self.exaggeration = exaggeration
+    @staticmethod
+    def compute_pca(task, state, **_):
+        # Perform PCA preprocessing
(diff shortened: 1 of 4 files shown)


Orange/widgets/unsupervised/owtsne.py
```diff
@@ -1,60 +1,247 @@
...
+            for job, job_weight in job_queue:
+                job_weight /= total_weight
+
+                def _progress_callback(val):
+                    state.set_progress_value(
+                        100 * progress_done + 100 * val * job_weight
+                    )
+
+                if state.is_interruption_requested():
+                    raise InterruptRequested()
```
VesnaT: Why raising an exception here and excepting it a few lines below? Maybe you could just return the `task`.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -165,79 +342,148 @@ def _add_controls_start_box(self):
...
-        elif self.__state in (OWtSNE.Finished, OWtSNE.Waiting):
-            self.__start()
-
-    def stop(self):
-        self.__state = OWtSNE.Paused
-        self.__set_update_loop(None)
-
-    def resume(self):
-        self.__set_update_loop(self.tsne_iterator)
+            self.run_button.setText("Stop")
```
VesnaT: I think this line is not needed, since the text is already set in `start()`.

VesnaT:
Maybe you could write some tests for the TSNERunner. Each its method is easily testable. And it would also be good to check that the embedding is not modified inplace.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -248,154 +431,159 @@ def set_data(self, data: Table):
...
-        super().setup_plot()
-        self.start()
+    def on_done(self, task):
+        # type: (Task) -> None
+        self.run_button.setText("Start")
+        # NOTE: All of these have already been set by on_partial_result,
+        # we double check that they are aliases
+        if task.pca_projection is not None:
+            self.__ensure_task_same_for_pca(task)
+            assert task.pca_projection is self.pca_projection
```
VesnaT: Since you already have a method with assert statements (self.__ensure_task_same_for_pca()), why not putting this line there? You'd only have to make sure that the method is called after setting `self.pca_projection` in `self.on_partial_result()`.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -248,154 +431,159 @@ def set_data(self, data: Table):
...
+        if task.initialization is not None:
+            self.__ensure_task_same_for_initialization(task)
+            assert task.initialization is self.initialization
```
VesnaT: Same as above.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -248,154 +431,159 @@ def set_data(self, data: Table):
...
+        if task.affinities is not None:
+            assert task.affinities is self.affinities
```
VesnaT: Same as above.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -248,154 +431,159 @@ def set_data(self, data: Table):
...
+        if task.tsne_embedding is not None:
+            self.__ensure_task_same_for_embedding(task)
+            assert task.tsne_embedding is self.tsne_embedding
```
VesnaT: Same as above.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -1,60 +1,235 @@
...
+            task.iterations_done += step_size
+            state.set_partial_result(("tsne_embedding", task))
+            if progress_callback is not None:
+                progress_callback(task.iterations_done / total_iterations_needed)
+
+            if state.is_interruption_requested():
+                return
+
+        # Run regular optimization phase
+        while task.iterations_done < total_iterations_needed:
```
VesnaT: This code is duplicated (with the exception of some parameters). Consider writing a function and calling it twice with different parameters.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -1,60 +1,235 @@
...
+                (partial(cls.compute_pca, task, state), weights["pca"])
+            )
+        if task.initialization is None:
+            job_queue.append(
+                (partial(cls.compute_initialization, task, state), weights["init"])
             )
+        if task.affinities is None:
+            job_queue.append(
+                (partial(cls.compute_affinities, task, state), weights["aff"])
+            )
```
VesnaT: How about something like this instead?
```
tasks = [(task.pca_projection, cls.compute_pca, weights["pca"]),
               (task.initialization, cls.compute_initialization, weights["init"]),
               (task.affinities, cls.compute_affinities, weights["aff"])]
job_queue = [(partial(y, task, state), w) for x, y, w in tasks if x]

```

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -164,79 +326,100 @@ def _add_controls_start_box(self):
...
         super().set_data(data)
 
         if data is not None:
+            n_attrs = len(data.domain.attributes)
+            self.controls.pca_components.setMaximum(
+                min(_MAX_PCA_COMPONENTS, n_attrs)
+            )
+            self.controls.pca_components.setValue(
+                min(_DEFAULT_PCA_COMPONENTS, n_attrs)
+            )
```
VesnaT: This prevents setting the value of `pca_components` to the right value (the one saved in settings).

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -248,154 +431,159 @@ def set_data(self, data: Table):
...
+            self.run_button.setText("Stop")
 
-        # Compute perplexity settings for multiscale
-        n_samples = self.pca_data.X.shape[0]
-        if self.multiscale:
-            perplexity = min((n_samples - 1) / 3, 50), min((n_samples - 1) / 3, 500)
-        else:
-            perplexity = self.perplexity
+        # Cancel current running task
+        self.cancel()
```
VesnaT: Is this needed? The task is canceled when self.start() is invoked.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -164,79 +326,100 @@ def _add_controls_start_box(self):
...
+            self.controls.pca_components.setMaximum(
+                min(_MAX_PCA_COMPONENTS, n_attrs)
+            )
+            self.controls.pca_components.setValue(
+                min(_DEFAULT_PCA_COMPONENTS, n_attrs)
+            )
+
             # PCA doesn't support normalization on sparse data, as this would
             # require centering and normalizing the matrix
             self.normalize_cbx.setDisabled(data.is_sparse())
```
VesnaT: In know this is an old code but still. It should be placed in `self.enable_controls()`. Besides, in this place `data` may differ from `self.data`, because the later could be set to None in `self.check_data()`. The same goes to the code above.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -248,154 +431,159 @@ def set_data(self, data: Table):
...
-        self.needs_to_draw = True
-
-        # We call PCA through fastTSNE because it involves scaling. Instead of
-        # worrying about this ourselves, we'll let the library worry for us.
-        initialization = TSNE.default_initialization(
-            self.pca_data.X, n_components=2, random_state=0)
+        # When the data is invalid, it is set to `None` and an error is set,
+        # therefore it would be erroneous to clear the error here
+        if self.data is not None:
+            self.Error.clear()
```
VesnaT: Is is possible to have any error messages show at this point?

Orange/widgets/unsupervised/owtsne.py
pavlin-policar: Removed the exception altogether.

Orange/widgets/unsupervised/owtsne.py
pavlin-policar: This is definitely shorter and much nicer to look at. However, this now seems a little complicated for what is basically several if statements and it takes a couple of seconds to see what's going on. May I propose a very straightforward alternative:

```python
        job_queue = []
        # Add the tasks that still need to be run to the job queue
        if task.pca_projection is None:
            job_queue.append((cls.compute_pca, weights["pca"]))

        if task.initialization is None:
            job_queue.append((cls.compute_initialization, weights["init"]))

        if task.affinities is None:
            job_queue.append((cls.compute_affinities, weights["aff"]))

        total_iterations = task.tsne.early_exaggeration_iter + task.tsne.n_iter
        if task.tsne_embedding is None or task.iterations_done < total_iterations:
            job_queue.append((cls.compute_tsne, weights["tsne"]))

        job_queue = [(partial(f, task, state), w) for f, w in all_jobs]
```

I do like your version, but I think this one will take a couple of seconds less to understand when one of us inevitably returns to this code.

Orange/widgets/unsupervised/tests/test_owtsne.py
```diff
@@ -24,66 +37,70 @@ def setUpClass(cls):
...
+        self.widget.cancel()
+        try:
+            self.tsne.stop()
+            self.tsne_model.stop()
+        # If `restore_mocked_functions` was called, stopping the patchers will raise
+        except RuntimeError as e:
+            if str(e) != "stop called on unstarted patcher":
+                raise e
 
     def restore_mocked_functions(self):
```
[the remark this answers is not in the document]
pavlin-policar: Kind of. It's called once now. Also, the way I'm using the patchers is pretty unconventional (I haven't seen this type of usage anywhere), and it's turned on by default, so t-SNE isn't actually run while testing the widget. But, if we want to turn this off, and _actually_ run t-SNE, we can do that with this function, without having to scratch our heads on how to do that.

Orange/widgets/unsupervised/owtsne.py
pavlin-policar: Done.

Orange/widgets/unsupervised/owtsne.py
pavlin-policar: Yes, it's needed: below this statement is an early exit condition if `data is None`. So if a job is running, and we remove the input signal (`data = None`), the running task won't be cancelled without this line.

Orange/widgets/unsupervised/owtsne.py
pavlin-policar: Yes, you're right. This remained here from earlier when we were catching general errors in t-SNE optimization. We later agreed that it was better to crash so the error could be reported and not hidden away in the red error box thing.

Orange/widgets/unsupervised/owtsne.py
pavlin-policar: That's neat, I didn't know this function existed.

Orange/widgets/unsupervised/owtsne.py
pavlin-policar: Fixed.

Orange/widgets/unsupervised/owtsne.py
VesnaT: It seems that the widget is still running the task after removing the input signal despite that line.

VesnaT:
The following still happens:
- OWFile (Iris) -> OWtSNE (Check Preserve global structure -> Info message appears) -> OWFile (Reload file) -> OWtSNE (Info message has disappeared)
- OWFile (Iris) -> OWtSNE
				-> OWDataTable (Select 10 rows) -> OWtSNE (the same widget as above (input subset)) -> OWtSNE (Check Preserve global structure) -> OWDataTable (the same widget as above - reselect 10 rows) -> Crash

pavlin-policar:
@VesnaT I fixed the two issues you found above. I've fixed them in a very simple and hacky way. I tried to find an elegant solution, but any of these would require touching the base widget and then fixing the other widgets.

PR #3605, opened 2019-02-15
[FIX] Orange restart dialogs: Improve wording

### LINKED ISSUE #2581: Orange quits after "Reset Widget Settings..."
##### Orange version
3.5.0

##### Expected behavior
Reset Widget Settings ("Clear all widget settings on next restart") warns the user that the "A restart of the application is necessary for the changes to take effect", but does that not imply that Orange will quit after pressing Ok. It implies that upon quitting by the user the settings will be cleared.

##### Actual behavior
Orange quits after clicking Ok in Reset Widget Settings dialog.

JakaKokosar:
##### Issue
Fixes #2581

##### Description of changes

Reset widget settings:
clear_settings

Also for add-on manager:
add-ons

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/application/addons.py
+++ b/Orange/canvas/application/addons.py
@@ -28,7 +28,7 @@ from AnyQt.QtWidgets import (
     QWidget, QDialog, QLabel, QLineEdit, QTreeView, QHeaderView,
     QTextBrowser, QDialogButtonBox, QProgressDialog,
     QVBoxLayout, QStyle, QStyledItemDelegate, QStyleOptionViewItem,
-    QApplication, QHBoxLayout, QPushButton, QFormLayout
+    QApplication, QHBoxLayout, QPushButton, QFormLayout, QMessageBox
 )
 
 from AnyQt.QtGui import (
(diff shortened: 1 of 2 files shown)


lanzagar:
Maybe:
A restart of the application is necessary -> Orange needs to be restarted
?

We are also somewhat misleading here by letting the user select "restart" and then stare at the computer when orange closes and does not start again :)
But I am fine with this if it is hard to actually restart the application (we would need to close the current one to save settings etc, before starting a new instance).

Also, note that pylint is complaining - it seems you removed a call to a function that is now imported, but never used.

JakaKokosar:
@lanzagar, is this clearer?

Reset widget settings: "Close Orange or cancel operation"
screenshot 2019-02-15 at 15 20 28


add-on manager: "Close Orange or do it later (operation was already performed)"
screenshot 2019-02-15 at 15 20 10

lanzagar:
I think the second part of the message was good before ("for the changes to take effect") since it indicates what will happen if the user clicks Cancel (-> then changes will not take effect).
I would also add "now" at the end of "press ok to close orange"

In summary, the whole message should be:

> Orange needs to be restarted for the changes to take effect.
> Press OK to close Orange now.

PR #3611, opened 2019-02-18
[ENH] OWPythonScript: dropping and pasting of python scripts

markotoplak:
##### Issue
Drag and drop of files did not work in the Python script widget

##### Description of changes
Files can be dragged and dropped (or pasted) onto the script editor or widget's control area. Nothing changed for dropping or pasting text.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpythonscript.py
+++ b/Orange/widgets/data/owpythonscript.py
@@ -17,6 +17,7 @@ from AnyQt.QtGui import (
 )
 from AnyQt.QtCore import Qt, QRegExp, QByteArray, QItemSelectionModel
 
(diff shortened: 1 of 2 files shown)


markotoplak:
As @BlazZupan suggested, now, this widgets pop ups a dialog that asks whether to replace the current script or append to it.

Another option (that I prefer) is to never ask and always insert the text at the cursor position.

markotoplak:
We talked about it with @BlazZupan and we decided that if a user drags/pastes in a file, the current script is replaced with that file (there are no pop-up dialogs asking you where to put it as in the first attempt). I took care that undo still works.

PR #3614, opened 2019-02-19
[FIX] Orange: add more informatino on missing compiled libraries

### LINKED ISSUE #3482: Need informative error if Orange does not find compiled libraries needed
Importing Orange should fail with a more informative error if its compiled parts can not be found. Now it just complains about not finding _variable, like in #3481 or #1838.

Even installation should fail with an error in such cases.

robertcv:
##### Issue
Fixes #3482
There is unfortunately no straightforward way to add post install for wheels ([pypi/packaging-problems#64](https://github.com/pypa/packaging-problems/issues/64))
##### Description of changes
Add additional information if import of compiled libraries failes.
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/__init__.py
+++ b/Orange/__init__.py
@@ -1,6 +1,13 @@
 # This module is a mixture of imports and code, so we allow import anywhere
 # pylint: disable=wrong-import-position,wrong-import-order
 
+try:
+    from Orange.data import _variable
+except ImportError:
+    raise ImportError("Compiled libraries cannot be found.\n"
+                      "Try reinstalling the package with:\n"
+                      "pip install --no-binary Orange3") from None
+
 from Orange import data
 
 from .misc.lazy_module import _LazyModule


Orange/__init__.py
```diff
@@ -1,6 +1,12 @@
 # This module is a mixture of imports and code, so we allow import anywhere
 # pylint: disable=wrong-import-position,wrong-import-order
 
+try:
+    from Orange.data import _variable
+except ImportError:
+    raise ImportError("Compiled libraries cannot be found.\n"
+                      "Try reinstalling the package with: pip install --no-binary Orange3") from None
```
janezd: Break this line and then I'm merging. :)

Since you'll modify the commit anyway, do you think it would make sense to import other modules, too? It's probably possible that `_variable` would exist, but some other module would fail to compile.

In the future, we probably won't remember to add new modules here when we introduce them, but it's still better to have at least those that exist now.

This is just a thought. If you disagree, just break the line and we're done. :)

robertcv: I added the brake.
If the compile fails pip shows errors. I think the import problems lies somewhere else.
I would therefore argue that either all library work or none does so there is no point in adding all the imports.

janezd: I meant breaking the line because of pylint, not to add a `\n` to the message, too. But this is OK as well.

PR #3620, opened 2019-02-21
[FIX] MDS: Handle subset data

### LINKED ISSUE #3617: MDS no longer handles data subsets
##### Orange version
3.20.1

##### Expected behavior
MDS should mark data points that come from Data Subset channel.

##### Actual behavior
The input in Data Subset channel has no effect on the visualisation.

screen shot 2019-02-20 at 15 14 34

Fails with any data set, I have used brown-selected.

VesnaT:
##### Issue

Fixes #3617

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -947,6 +947,21 @@ class ProjectionWidgetTestMixin:
         self.widget.setup_plot.assert_called_once()
         self.widget.commit.assert_called_once()
 
+    def test_subset_data_color(self, timeout=DEFAULT_TIMEOUT):
+        self.send_signal(self.widget.Inputs.data, self.data)
+
+        if self.widget.isBlocking():
+            spy = QSignalSpy(self.widget.blockingStateChanged)
+            self.assertTrue(spy.wait(timeout))
+
+        self.send_signal(self.widget.Inputs.data_subset, self.data[:10])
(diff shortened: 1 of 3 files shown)


janezd:
I agree with extracting `_handle_subset_data` to reduce the complexity of `handleNewSignals`.

But I'm not too happy with the fix. The root cause of the bug is that `handleNewSignals` does not call `super().handleNewSignals()`. We'll get similar bugs every time somebody adds something to `OWDataProjectionWidget.handleNewSignals` (or `OWDataProjectionWidget.setup_plot()`:  `OWDataProjectionWidget.handleNewSignals` calls it and `OWMDS.handleNewSignals` doesn't!).

Derived methods that don't call inherited methods are a violation of OOP principles and a potential source of problems (especially when unexpected) that we should avoid. (Please don't say *look, who's preaching!* :))

You can:

- change `OWMDS.handleNewSignals` to call the inherited method, possibly moving some code from `OWMDS.handleNewSignals` to `OWMDS.setup_plot`, or
- remove `OWMDS.handleNewSignals` entirely and move all of its (additional) code to `OWMDS.setup_plot` (I would prefer this solution because this code indeed seems to belong there),
- convince me that above solutions are too impractical.

VesnaT:
I went for the second solution.

PR #3622, opened 2019-02-21
Preprocess: changed normalization features

### LINKED ISSUE #2739: [Feature Request]: Add options for 0,1 and -1,1 normalization in Preprocess
##### Feature
Add options to normalize to range [0,1] or [-1,1].

##### Use case
Great for comparing values in Timeseries add-on, e.g. when plotting two attributes against one another in Line Chart.
With current preprocessing:
screen shot 2017-11-03 at 14 57 13

Ideally, both attributes would be scaled to 0,1.

Ejmric:
Fixes #2739 
Added radio buttons and options to normalize by span to [0,1] and [-1,1].

- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpreprocess.py
+++ b/Orange/widgets/data/owpreprocess.py
@@ -29,7 +29,7 @@ from Orange.widgets.utils.overlay import OverlayWidget
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.widget import Input, Output
-
+from Orange.preprocess import Normalize
 from Orange.widgets.data.utils.preprocess import (
     BaseEditor, blocked, StandardItemModel, DescriptionRole,
     ParametersRole, Controller, SequenceFlow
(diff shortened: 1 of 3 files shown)


Orange/widgets/data/owpreprocess.py
```diff
@@ -637,55 +637,82 @@ def enum_to_index(enum, key):
...
+            return Normalize(norm_type=Normalize.NormalizeBySD)
+        elif method == Scale.NormalizeSpan:
+            return Normalize(norm_type=Normalize.NormalizeBySpan)
+        elif method == Scale.NormalizeSpan2:
+            return Normalize(norm_type=Normalize.NormalizeBySpan2, zero_based=False)
+        else:
+            method, defaults = Scale.Imputers[method]
+            defaults = dict(defaults)
+            defaults.update(params)
+            return preprocess.Normalize(method=method)
```
janezd: When can this `else` occur?

Orange/widgets/data/owpreprocess.py
```diff
@@ -637,55 +637,82 @@ def enum_to_index(enum, key):
...
+        elif method == Scale.CenterByMedian:
+            center = params.get("center", _Scale.CenteringType.Median)
+            scale = params.get("scale", _Scale.ScalingType.NoScaling)
+            return _Scale(center=center, scale=scale)
+        elif method == Scale.NormalizeSD:
+            return Normalize(norm_type=Normalize.NormalizeBySD)
+        elif method == Scale.NormalizeSpan:
+            return Normalize(norm_type=Normalize.NormalizeBySpan)
+        elif method == Scale.NormalizeSpan2:
+            return Normalize(norm_type=Normalize.NormalizeBySpan2, zero_based=False)
```
janezd: `params` can no longer contain a key `"center"`, so `center = params.get("center", _Scale.CenteringType.NoCentering)` is equivalent to `center = _Scale.CenteringType.NoCentering`. Better

```
center = params.get("center", _Scale.CenteringType.Mean)
scale = params.get("scale", _Scale.ScalingType.NoScaling)
return _Scale(center=center, scale=scale)
```

is equivalent to 

```
return _Scale(_Scale.CenteringType.Mean, _Scale.ScalingType.NoScaling)
```

and so forth

(I would omit `center=` and `scale=` because the names of constants are descriptive enough).

Orange/widgets/data/owpreprocess.py
```diff
@@ -29,7 +29,7 @@
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.widget import Input, Output
-
+from Orange.preprocess import Normalize
```
janezd: You can use `Normalize`, but since the code currently uses `Scale`, why wouldn't you modify `Scale` to support `zero_based` flag (which would default to `True`, but the widget would set it to `False` for `NormalizeSpan2`? This could simplify `createInstance` - you'd just use two lists or dictionaries instead of all the ifs.

Orange/widgets/data/owpreprocess.py
```diff
@@ -637,55 +637,82 @@ def enum_to_index(enum, key):
     enum key to its int position"""
     return list(enum).index(key)
 
-
 class Scale(BaseEditor):
-    NoCentering, CenterMean, CenterMedian = 0, 1, 2
-    NoScaling, ScaleBySD, ScaleBySpan = 0, 1, 2
+    NoCentering, CenterByMean, CenterByMedian, NormalizeSD, NormalizeSpan, NormalizeSpan2 = 0, 1, 2, 3, 4, 5
```
janezd: `NormalizeSpan2` is not very descriptive. :) What about `NormalizeBySD`, `NormalizeBySpan_ZeroBased`, `NormalizeSpan_NonZeroBased`. It is not entirely clear what the latter is based on, but it's still closer to being understandable. :)

lanzagar:
If we are moving away from two controls that allow to set centering and scaling independently, how about the following list of options:

- Standardize (mean=0, std=1)
- Normalize to [0, 1]
- Normalize to [-1, 1]
- Center only (mean=0)
- Scale only (std=1)

The first would be the default.

Also, it seems the library already supported all necessary transformations, so it does not need to be changed (files Orange/preprocess/preprocess.py and normalize.py). Just update the widget.

janezd:
Apparently we write very robust tests: although they were using old parameters that no longer exist (`"center"` and `"scale"`), they still passed.

I have added some tests.

PR #3625, opened 2019-02-25
[ENH] FDR: Calculate FDR using numpy

VesnaT:
##### Issue

Current implementation of False discovery rate is using python list and for loops and is therefore inefficient.

##### Description of changes
Implement FDR using numpy.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -5,7 +5,7 @@ and once used from the bottlechest package (fork of bottleneck).
 It also patches bottleneck to contain these functions.
 """
 import warnings
-import math
+from typing import Iterable
 
 import numpy as np
 import bottleneck as bn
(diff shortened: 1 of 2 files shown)


ajdapretnar:
At first I thought this would be a great fix for Text add-on, but then I realized Text implements its own FRD! 😮 I don't think this is optimal. Would you care to check that implementation as well and perhaps replace it with this one? https://github.com/biolab/orange3-text/blob/master/orangecontrib/text/stats.py

Orange/statistics/util.py
```diff
@@ -607,59 +607,35 @@ def std(x, axis=None, ddof=0):
...
-        return all(l[i] <= l[i + 1] for i in range(len(l) - 1))
-
-    if not ordered:
-        ordered = is_sorted(p_values)
+    if p_values is None or len(p_values) == 0 or \
+            (m is not None and m <= 0):
+        return None
 
+    is_list = isinstance(p_values, list)
+    p_values = np.array(p_values) if is_list else p_values
```
janezd: `p_values = np.asarray(p_values)`. It won't copy if it's already an array.

Orange/statistics/util.py
```diff
@@ -607,59 +607,35 @@ def std(x, axis=None, ddof=0):
...
+    m = len(p_values) if m is None else m
```
janezd: We usually use

```python
if m is None:
    m = len(p_values)
```

, I suppose, because it's way easier to read. Admittedly a question of taste, though.

VesnaT:
I think those implementations were the same. So Text add-on should probably use the new one in Orange core, instead of implementing its own.

Orange/statistics/util.py
```diff
@@ -607,59 +607,35 @@ def std(x, axis=None, ddof=0):
...
-    fdrs = []
-    cmin = tmp_fdrs[-1]
-    for f in reversed(tmp_fdrs):
-        cmin = min(f, cmin)
-        fdrs.append( cmin)
-    fdrs.reverse()
+        m *= sum(1 / np.arange(1, m + 1))
 
+    fdrs = (p_values * m / np.arange(1, len(p_values) + 1))[::-1]
+    fdrs = np.array(np.minimum.accumulate(fdrs)[::-1])
```
janezd: Kudos.

Orange/tests/test_statistics.py
```diff
@@ -264,11 +264,31 @@ def test_nanstd_with_ddof(self):
...
+        p_values = np.array([0.0002, 0.0004, 0.00001, 0.0003, 0.0001])
         np.testing.assert_almost_equal(
-            np.array([0.00005, 0.00025, 0.00033, 0.00038, 0.0004]),
+            np.array([0.00033, 0.0004, 0.00005, 0.00038, 0.00025]),
             FDR(p_values), decimal=5)
 
+    def test_FDR_dependent(self):
+        p_values = np.array([0.0002, 0.0004, 0.00001, 0.0003, 0.0001])
+        np.testing.assert_almost_equal(
+            np.array([0.00076, 0.00091, 0.00011, 0.00086, 0.00057]),
```
janezd: Have you gotten these number independently, that is, from another source or calculated manually, not from the code itself? (I'm asking because I used to be a big sinner myself.)

Orange/statistics/util.py
```diff
@@ -607,59 +607,35 @@ def std(x, axis=None, ddof=0):
...
     """ `False Discovery Rate <http://en.wikipedia.org/wiki/False_discovery_rate>`_
         correction on a list of p-values.
 
-    :param p_values: a list of p-values.
+    :param p_values: list or np.ndarray of p-values.
     :param dependent: use correction for dependent hypotheses (default False).
     :param m: number of hypotheses tested (default ``len(p_values)``).
     :param ordered: prevent sorting of p-values if they are already sorted
         (default False).
+    :return: list or a np.ndarray - depends on the input
```
janezd: `:return: list or a np.ndarray, same as the input`

ajdapretnar:
Agree.

Orange/statistics/util.py
```diff
@@ -607,59 +607,35 @@ def std(x, axis=None, ddof=0):
...
     """
-    def is_sorted(l):
-        return all(l[i] <= l[i + 1] for i in range(len(l) - 1))
-
-    if not ordered:
-        ordered = is_sorted(p_values)
+    if p_values is None or len(p_values) == 0 or \
+            (m is not None and m <= 0):
+        return None
```
janezd: Cover this line in tests to get approval by codecov (and to make sure we don't forget these cases at any future refactoring).

Orange/statistics/util.py
```diff
@@ -607,59 +607,35 @@ def std(x, axis=None, ddof=0):
...
+    fdrs = (p_values * m / np.arange(1, len(p_values) + 1))[::-1]
+    fdrs = np.array(np.minimum.accumulate(fdrs)[::-1])
     if not ordered:
-        new = [None] * len(fdrs)
-        for v, i in zip(fdrs, indices):
-            new[i] = v
-        fdrs = new
-
-    return fdrs
+        fdrs[indices] = fdrs.copy()
```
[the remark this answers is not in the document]
VesnaT: It doesn't :) ..and I took me a while to find the bug the absence of copy caused :)

Orange/tests/test_statistics.py
VesnaT: I calculated those using the old function.

ajdapretnar:
FDR in Text add-on replaced with this implementation: https://github.com/biolab/orange3-text/pull/416

PR #3626, opened 2019-02-28
Remove sparse support for Hamming

ajdapretnar:
##### Issue

Hamming distance does not support sparse as stated in [sklearn documentation](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html).

##### Description of changes
Set support_sparse to False. Add test to ensure error is raised when trying to compute distances with metrics that do not support sparse.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/distance/distance.py
+++ b/Orange/distance/distance.py
@@ -662,7 +662,7 @@ class MahalanobisDistance:
         return Mahalanobis(axis=axis).fit(data)
 
 class Hamming(Distance):
-    supports_sparse = True
+    supports_sparse = False
     supports_missing = False
     supports_normalization = False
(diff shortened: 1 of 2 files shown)


ajdapretnar:
@janezd Please check if the test makes sense and is general to test all distances.

janezd:
The test that you added makes sense, but you could also add the opposite one: if some distance does support sparse data, call it - no need to check the result, just see that it doesn't raise an exception. Such test would detect the bug that you're fixing, while the test that you added actually detects the opposite bug.

ajdapretnar:
@janezd Done.

PR #3632, opened 2019-03-01
[FIX] Send usage statistics in a thread at startup

### LINKED ISSUE #3597: Usage statistics block exiting Orange on certain networks
##### Orange version
Any after #3151

##### Expected behavior
Closing a workflow closes the windows immediately.

##### Actual behavior
The workflow window closes immediately on no connection or on a very good connection. In some cases, when my network is connected but unreachable in a certain way closing the window blocks.

Open Orange, add a widget, close Orange on a network with automatic proxy configuration but with proxies not set manually in Orange, and wait. Happens both on OS X and Windows.

Closing works properly if system proxies are set manually, because these kind of proxies are autodetected.

I can also reproduce this bug if I connect to a proxy-less network that requires me to login in a browser before I can continue (and I do not login). 

Currently, closing a workflow makes a HTTP request and then blocks until answered/errored/timeout (supposedly, I never waited long enough for it to timeout).

A short timeout would help, but even waiting a few seconds for a window to close is still bad design.

I propose a different implementation. When a workflow is closed, usage statistics are just saved (as happens now if there is no connection). Then, when Orange is restarted, the data is sent in the background, similarly to our version checks.

rokgomiscek:
##### Issue

Fixes #3597

##### Description of changes
When closing Orange, usage statistics are saved. Then, when starting Orange, usage statistics are sent in the background.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -276,8 +276,12 @@ def send_usage_statistics():
 
     class SendUsageStatistics(QThread):
         def run(self):
-            usage = UsageStatistics()
-            usage.send_statistics()
+            try:
+                usage = UsageStatistics()
+                usage.send_statistics()
+            except Exception:  # pylint: disable=broad-except
+                # exceptions in threads would crash Orange
+                log.warning("Failed to send usage statistics.")
 
     thread = SendUsageStatistics()
     thread.start()
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -34,6 +34,7 @@ from Orange.canvas.application.errorreporting import handle_exception
 from Orange.canvas.gui.splashscreen import SplashScreen
 from Orange.canvas.config import cache_dir
 from Orange.canvas import config
+from Orange.canvas.document.usagestatistics import UsageStatistics
 
 from Orange.canvas.registry import qt
 from Orange.canvas.registry import WidgetRegistry, set_global_registry
@@ -271,6 +272,16 @@ def check_for_updates():
         thread.start()
         return thread
 
+def send_usage_statistics():
+
+    class SendUsageStatistics(QThread):
+        def run(self):
+            usage = UsageStatistics()
+            usage.send_statistics()
+
+    thread = SendUsageStatistics()
+    thread.start()
+    return thread
 
 def main(argv=None):
     if argv is None:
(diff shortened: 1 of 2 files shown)


Orange/canvas/document/usagestatistics.py
```diff
@@ -105,6 +105,31 @@ def log_node_added(self, widget_name, extended_widget=None):
     def set_node_type(self, addition_type):
         self.__node_addition_type = addition_type
 
+    def send_statistics(self):
+        def store_data(d):
```
markotoplak: We probably do not need to store data in this function - as far as I understand we only need to clear when it was sent successfully.

Orange/canvas/__main__.py
```diff
@@ -271,6 +272,16 @@ def compare_versions(latest):
         thread.start()
         return thread
 
+def send_usage_statistics():
+
+    class SendUsageStatistics(QThread):
```
markotoplak: What happens if there is an uncaught exception in a QThread for any reason? I faintly remember some problems with crashing and uncaught exceptions in threads, but that might be just my imagination.

Also, as this addition is a non-vital part of Orange, I recommend being defensive. See send_usage_statistics.

Orange/canvas/document/usagestatistics.py
```diff
@@ -105,6 +105,31 @@ def log_node_added(self, widget_name, extended_widget=None):
...
+            with open(statistics_path, 'w') as f:
+                json.dump(d, f)
+        if release and config.settings()["error-reporting/send-statistics"]:
+            if os.path.isfile(statistics_path):
+                with open(statistics_path) as f:
+                    data = json.load(f)
+            else:
+                data = []
```
markotoplak: Why else here? Why would we want to send empty data?

Orange/canvas/document/usagestatistics.py
rokgomiscek: You are right, I will remove this part.

Orange/canvas/document/usagestatistics.py
rokgomiscek: I moved the sending part inside if os.path.isfile(statistics_path) for the rare case, when there is no file (e.g. directly after installation).

Orange/canvas/__main__.py
rokgomiscek: Did you mean like this? I added an except Exception to send_statistics.

markotoplak: I added a commit that moves checking for exception to run(). Because, for example, an exception could also happen in json loading (if file was saved incorrectly). Could you verify if everything is still OK?

PR #3633, opened 2019-03-01
[FIX] OWTable: Include attributes from meta attributes in header row labels

### LINKED ISSUE #2777: Widget Data table shows colours in header
##### Orange version
screen shot 2017-11-14 at 16 34 26

##### Expected behavior
1. If feature color changed in Color widget, it will not influence data in Data table.

2. Attribute names for metas shall be displayed in data table header.

##### Actual behavior
1. If some feture color is changed in Color widget, Data table widget displays color values as feature attribute.
colors-error
2. If color is changed for meta data, Data table widget does not put attribute name (in this case color) in the header
screen shot 2017-11-14 at 16 47 55

iris dataset:
1. change color for some feature
2. set some feature as meta and change color for it

janezd:
##### Issue

Fixes #2777

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -689,7 +689,8 @@ class OWDataTable(widget.OWWidget):
                 RichTableModel.Labels | RichTableModel.Name)
 
             labelnames = set()
-            for a in model.source.domain.variables:
+            domain = model.source.domain
+            for a in itertools.chain(domain.metas, domain.variables):
                 labelnames.update(a.attributes.keys())
             labelnames = sorted(
                 [label for label in labelnames if not label.startswith("_")])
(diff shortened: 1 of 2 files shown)


markotoplak:
This is certainly a fix of the point (2) in #2777.

But you decided that you won't fix the point (1), right?

janezd:
I didn't decide, I just overlooked it. :)

But if I didn't overlook it, I still wouldn't implement it. I think it doesn't hurt to have this information there. If we however decide to remove it, we can easily do this by renaming "color" to "_color".

What's your opinion? And @BlazZupan, @brunap, others?

markotoplak:
@janezd, I am against hiding "color". I have no opinion about "_color".

janezd:
If you're against hiding, that's it. We would rename it to `_color` if we wanted to hide it -- all attributes that start with `_` are automatically hidden.

PR #3637, opened 2019-03-02
OWContinuize: PyLint

janezd:

... except for `len-as-condition` caused by `if self.data is not None and len(self.data)`. We need to decide to define `Table.is_empty` or `Table.__bool__`. I vote for `Table.is_empty` because it's more explicit. We could also have `.size` like numpy arrays, but it is not clear whether this would include meta attributes or not.

### DIFF
--- a/Orange/widgets/data/owcontinuize.py
+++ b/Orange/widgets/data/owcontinuize.py
@@ -129,7 +129,7 @@ class OWContinuize(widget.OWWidget):
 
     def commit(self):
         continuizer = self.constructContinuizer()
-        if self.data is not None and len(self.data):
+        if self.data:
             domain = continuizer(self.data)
             data = self.data.transform(domain)
             self.Outputs.data.send(data)
@@ -262,7 +262,7 @@ def continuize_var(var,
                    continuous_treatment=Continuize.Leave,
                    zero_based=True):
 
-    if var.is_continuous:
+    def continuize_continuous():
         if continuous_treatment == Normalize.NormalizeBySpan:
             return [normalize_by_span(var, data_or_dist, zero_based)]
         elif continuous_treatment == Normalize.NormalizeBySD:
@@ -270,7 +270,7 @@ def continuize_var(var,
         else:
             return [var]
 
-    elif var.is_discrete:
+    def continuize_discrete():
         if len(var.values) > 2 and \
                 multinomial_treatment == Continuize.ReportError:
             raise ValueError("{0.name} is a multinomial variable".format(var))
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owcontinuize.py
```diff
@@ -34,6 +34,11 @@ class Outputs:
     buttons_area_orientation = Qt.Vertical
     resizing_enabled = False
 
+    multinomial_treatment: int
+    zero_based: int
+    continuous_treatment: int
+    class_treatment: int
```
lanzagar: I am not completely up to speed on what kind of type annotations are preferred, but I don't think I have seen many (any) uses in Orange classes, where we would first annotate class attributes and then assign values later. I personally prefer the one-line annotations in comments, e.g.:
```
multinomial_treatment = Setting(0)  # type: int
```
Mostly for stupid and simple reasons like less lines of code and smaller chance of changing a default value from 0 to False and forgetting to update the annotation 5 lines higher.

I know this is nitpicking, but since it seem that Settings will need to be annotated in other classes too in the future, it would be nice to have them all done in the same way. So this is setting a precedent.
Also - do you think we should use type annotations every time when declaring Settings?

janezd: No, I don't see this as nitpicking at all. I'd like to have a consistent format, too. And I strongly dislike annotations in separate lines - which is why I haven't added them systematically.

I would prefer

```python
multinomial_treatment: int = Setting(0)
```

because it's shortest and consistent with (the simplest) annotations of argument types in function definitions, which we already use. However, in `int x: int = f()` pylint ignores `int` and uses the type of `f`.

It also ignores the annotation you suggested,`multinomial_treatment = Setting(0)  # type: int`. Apparently assignments take precedence over hints?!

I also tried annotating `Setting.__new__` as a generic, and many other tricks. Nope. This works:

```
class Setting(int):
     ...
```

but it's too weird and it also tells that `Settings` is an `int` - not a `str`, which may trigger other warnings. I also tried a union,

```
T = TypeVar("Setting", int, float, str)

class Setting(Generic[T]):
```

but it doesn't work.

Since pylint is incapable of understanding our settings (you're welcome to look for solutions), the only solution without extra annotations or disabling the check altogether is to disable type checking for `Settings`. I discovered that this could be done by puzzling pylint with a second  definition of `Settings`:

```
if 1 == 0:
    class Setting:
        pass
```

Unless I overlooked another, simpler solution we either

- disable this warning (it's a shame, it can be useful),
- use annotations in separate lines (too long, and we'll add them just for attributes that pylint complains about -> inconsistency),
- disable checking `Settings` using the above (unconventional) method.

If you agree, I'd add the fake class definition and then remove type annotations from PRs I made.

janezd:
After merging #3662, this PR contains just a few trivial changes, so I'm merging it myself.

PR #3642, opened 2019-03-02
OWDatasets: PyLint

janezd:

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -25,10 +25,10 @@ from serverfiles import LocalFiles, ServerFiles, sizeformat
 
 import Orange.data
 from Orange.misc.environ import data_dir
-from Orange.widgets import widget, settings, gui
+from Orange.widgets import settings, gui
 from Orange.widgets.utils.signals import Output
 from Orange.widgets.utils.widgetpreview import WidgetPreview
-from Orange.widgets.widget import Msg
+from Orange.widgets.widget import OWWidget, Msg
 
 
 log = logging.getLogger(__name__)
@@ -124,7 +124,7 @@ class Namespace(SimpleNamespace):
             self.title = self.filename
 
 
-class OWDataSets(widget.OWWidget):
+class OWDataSets(OWWidget):
     name = "Datasets"
     description = "Load a dataset from an online repository"
     icon = "icons/DataSets.svg"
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owdatasets.py
```diff
@@ -522,15 +528,17 @@ def onDeleteWidget(self):
...
         return QSize(900, 600)
 
     def closeEvent(self, event):
         self.splitter_state = bytes(self.splitter.saveState())
         self.header_state = bytes(self.view.header().saveState())
         super().closeEvent(event)
 
-    def load_data(self, path):
+    @staticmethod
+    def load_data(path):
```
lanzagar: This method does not do much here, but is overridden (at least) in orange3-single-cell. It is not a staticmethod there, so we might want to keep it non-static here as well for consistency.

Orange/widgets/data/owdatasets.py
```diff
@@ -171,11 +172,17 @@ class Outputs:
...
         self.local_cache_path = os.path.join(data_dir(), self.DATASET_DIR)
 
-        self._header_labels = [header['label'] for _, header in self.HEADER_SCHEMA]
-        self._header_index = namedtuple('_header_index', [info_tag for info_tag, _ in self.HEADER_SCHEMA])
-        self.Header = self._header_index(*[index for index, _ in enumerate(self._header_labels)])
+        self._header_labels = [
+            header['label'] for _, header in self.HEADER_SCHEMA]
+        self._header_index = namedtuple(
+            '_header_index', [info_tag for info_tag, _ in self.HEADER_SCHEMA])
+        self.header = self._header_index(
```
lanzagar: Changing `Header` to `header` breaks the scOrange datasets widget, which extends this one and redefines `Header`.

I see no reason why it was capitalized before, so I agree with the change. But it needs to be coordinated with changes in widgets that extend this one.

Orange/widgets/data/owdatasets.py
janezd: Agreed.

janezd:
I didn't know that this widget is extended. But that's the reason why I only self-merged the PRs with trivial changes.

I would prefer if we decide to never subclass widget classes. Better still, we should never import a module with a widget, unless this is an abstract widget class. Let's discuss this on Friday, and if we agree, I'll make widget classes final unless explicitly declared as open (with a reasonable period of deprecation warnings). NB, all Kotlin classes are final by default.

Argument: if widgets can be subclassed, our hands are rather tied when it comes to any redesigning. Touch any widget and derived widgets will break.

PR #3647, opened 2019-03-03
OWImpute: PyLint

janezd:

This PR is non-trivial (compared to others in the series) because it splits a method with too many branches into several parts.


mstrazar:
Tested the OWImpute widget and tests, works OK, passes PyLint.

PR #3651, opened 2019-03-03
OWPaintData: PyLint

janezd:
Pylint reported a lot of irrelevant issues for Paint Data, so this PR is mostly about disabling them.

### DIFF
--- a/Orange/widgets/data/owpaintdata.py
+++ b/Orange/widgets/data/owpaintdata.py
@@ -216,6 +216,7 @@ class DataTool(QObject):
             self._cursor = QCursor(cursor)
             self.cursorChanged.emit()
 
+    # pylint: disable=unused-argument,no-self-use,unnecessary-pass
     def mousePressEvent(self, event):
         return False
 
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owpaintdata.py
```diff
@@ -594,6 +596,8 @@ def undo(self):
         self._model.execute(self._undo)
 
     def mergeWith(self, other):
+        # other is another instance of the same class, thus
+        # pylint: disable=protected-access
         if self.id() != other.id():
```
janezd: @ales-erjavec, isn't `id` always `1`? If I'm not mistaken, this check, the above comment, and method `id` can be removed?

Orange/widgets/data/owpaintdata.py
```diff
@@ -1154,20 +1158,20 @@ def _on_editing_finished(self):
...
     def execute(self, command):
-        if isinstance(command, (Append, DeleteIndices, Insert, Move)):
-            if isinstance(command, (DeleteIndices, Insert)):
-                self._selected_indices = None
+        assert isinstance(command, (Append, DeleteIndices, Insert, Move)), \
+            "Non normalized command"
+        if isinstance(command, (DeleteIndices, Insert)):
+            self._selected_indices = None
 
-                if isinstance(self.current_tool, SelectTool):
```
janezd: @ales-erjavec, should this `if` really be nested? Am I overlooking something or is this condition always false?

If resetting the tool is redundant, `_reset()` method can remain private.

Orange/widgets/data/owpaintdata.py
ales-erjavec: The check can be, but the `id` method [should stay.](https://doc.qt.io/qt-5/qundocommand.html#id)

Orange/widgets/data/owpaintdata.py
ales-erjavec: Yes the condition can be true and should be nested.

janezd: Oops, apologies for asking. I was so sloppy that I overlooked that this one checks `self.current_tool`, not `command`, so I didn't understand why is this needed. :(

PR #3652, opened 2019-03-03
OWPreprocess: PyLint

janezd:
Now passes pylint with disabling only `missing-docstring`, `no-name-in-module`, `too-many-ancestors`, `no-else-return`, except for file length.


janezd:
@ales-erjavec, has this try-except remained here from the time before migrations, and we should remove it to avoid masking errors in settings? Or is it necessary and correct?

```
        try:
            model = self.load(self.storedsettings)
        except Exception:
            model = self.load({})
```

ales-erjavec:
> @ales-erjavec, has this try-except remained here from the time before migrations, and we should remove it to avoid masking errors in settings? Or is it necessary and correct?

I guess it can be removed.

PR #3655, opened 2019-03-03
OWRandomize: PyLint

janezd:
Fix pylint issue related to `Setting` not being recognized as `int`.


Orange/widgets/data/owrandomize.py
```diff
@@ -31,7 +31,7 @@ class Outputs:
     shuffle_attrs = Setting(False)
     shuffle_metas = Setting(False)
     scope_prop = Setting(80)
-    random_seed = Setting(0)
+    random_seed = Setting(False)
```
janezd: This is not because of pylint: using `0` was confusing, in particular because the name of the setting itself is wrong.

PR #3656, opened 2019-03-04
Make widget classes final, unless explicitly declared open

janezd:
##### Issue

We always silently assume that widget classes are final. Without this, we wouldn't be able to redesign widgets, not even to change a badly named attribute.

Some widgets, like `OWDatasets` are, however, unexpectedly subclassed, as I was warned in #3642.

I propose to make all widgets final unless explicitly declared open. We will open abstract widgets like `OWWidget`, `OWProjectionWidgetBase`, `OWProjectionWidget`. If we really wish, we can also open non-abstract widgets, such as `OWDatasets`, but opening them will represent a commitment to avoid backward incompatible changes.

##### Description of changes

An open class shall be declared by adding `openclass=True` keyword to its definition, *e.g.* `class OWProjectionWidgetBase(OWWidget, openclass=True):`. Ordinary class attribute would not work because it would be inherited.

To implement this, metaclass puts a `_final_class` attribute into class namespace of final classes, and `__init_subclass__` hook in `OWWidget` checks it.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/test_widget.py
+++ b/Orange/widgets/tests/test_widget.py
@@ -1,7 +1,7 @@
-# Test methods with long descriptive names can omit docstrings
-# pylint: disable=all
+# pylint: disable=protected-access
 
 import gc
+import warnings
 import weakref
 
 from unittest.mock import patch, MagicMock
@@ -19,7 +19,7 @@ from Orange.widgets.utils.messagewidget import MessagesWidget
 
 
 class DummyComponent(OWComponent):
-    b = None
+    dummyattr = None
 
 
 class MyWidget(OWWidget):
@@ -43,8 +43,8 @@ class WidgetTestCase(WidgetTest):
         setattr(widget, 'field', 1)
         self.assertEqual(widget.field, 1)
 
-        setattr(widget, 'component.b', 2)
-        self.assertEqual(widget.component.b, 2)
+        setattr(widget, 'component.dummyattr', 2)
+        self.assertEqual(widget.component.dummyattr, 2)
 
         setattr(widget, 'widget.field', 3)
         self.assertEqual(widget.widget.field, 3)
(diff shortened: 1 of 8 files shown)


Orange/widgets/widget.py
```diff
@@ -65,20 +65,29 @@ class WidgetMetaClass(type(QDialog)):
 
     #noinspection PyMethodParameters
     # pylint: disable=bad-classmethod-argument
-    def __new__(mcs, name, bases, kwargs):
-        cls = super().__new__(mcs, name, bases, kwargs)
+    def __new__(mcs, name, bases, namespace, openclass=False, **kwargs):
+        cls = super().__new__(mcs, name, bases, namespace, **kwargs)
```
janezd: Renaming `kwargs` to `namespace` and adding `**kwargs` is unrelated to this PR. These names were incorrect and puzzling.

markotoplak:
I am also guilty of subclassing OWFIle and shamelessly importing code from owpreprocess.py. I think I should fix it in the Spectroscopy add-on.

I do not have any objections to enforcing final widget classes.

janezd:
I think the proper thing to do would be to extract the code that you need from the File widget into another class (within the `OWWidget` hierarchy - as a common ancestor to `OWFile` and another widget, or outside the hierarchy), and put this code outside the `OWFile` module. We tend to keep such code backwards compatible, while incompatible changes to widgets were always normal practice.

Actually, it would be cool to have a base class for different types of file widgets.

ales-erjavec:
+1 on making them final.

Orange/widgets/tests/test_widget.py
```diff
@@ -192,6 +192,18 @@ def test_status_bar_action(self):
...
         self.assertFalse(action.isChecked())
 
+    def test_widgets_cant_be_subclassed(self):
+        # pylint: disable=unused-variable
+        with self.assertWarns(RuntimeWarning):
+            class MySubWidget(MyWidget):
+                pass
+
+        with warnings.catch_warnings():
+            warnings.filterwarnings("error", ".*", RuntimeWarning)
```
VesnaT: Why should we catch warnings here? There aren't any. Or have I missed something?

Orange/widgets/tests/test_widget.py
```diff
@@ -192,6 +192,18 @@ def test_status_bar_action(self):
...
+            class MyWidget2(OWWidget, openclass=True):
+                pass
```
VesnaT: While at it, please add a blank line here.

Orange/widgets/tests/test_widget.py
```diff
@@ -192,6 +192,18 @@ def test_status_bar_action(self):
...
+            class MySubWidget2(OWWidget):
+                pass
```
VesnaT: Same as above.

PR #3657, opened 2019-03-04
[ENH] Support sparse Jaccard

ajdapretnar:
##### Issue

Jaccard did not support sparse data, making it useless for text mining.

##### Description of changes
Custom support for sparse Jaccard.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/distance/distance.py
+++ b/Orange/distance/distance.py
@@ -3,6 +3,7 @@ from unittest.mock import patch
 
 import numpy as np
 from scipy import stats
+from scipy import sparse as sp
 import sklearn.metrics as skl_metrics
 from sklearn.utils.extmath import row_norms, safe_sparse_dot
 from sklearn.metrics import pairwise_distances
@@ -11,7 +12,7 @@ from Orange.distance import _distance
 from Orange.statistics import util
 
 from .base import (Distance, DistanceModel, FittedDistance, FittedDistanceModel,
-                   SklDistance, _orange_to_numpy, SparseJaccard)
+                   SklDistance, _orange_to_numpy)
 
 class EuclideanRowsModel(FittedDistanceModel):
     """
@@ -395,6 +396,9 @@ class JaccardModel(FittedDistanceModel):
         compute distances between rows without missing values, and a slower
         loop for those with missing values.
         """
+        if sp.issparse(x1):
+            return self.sparse_jaccard(x1, x2)
+
         nonzeros1 = np.not_equal(x1, 0).view(np.int8)
         if self.axis == 1:
             nans1 = _distance.any_nan_row(x1)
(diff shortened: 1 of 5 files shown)


ajdapretnar:
@lanzagar @thocevar @janezd 
This requires some discussion.
1. Before, if Distances are given sparse discrete data, Jaccard didn't work. Currently fixed with a no-so-nice hack.
2. DistMatrix cannot be tested with np.testing.assert_array_equal, because max() returns an int instead of numpy.int64.
3. Distance.Jaccard().fit does not work for any sparse data. How to solve this?

All in all, the code needs to be made nicer and cleaner. Any suggestions welcome.

Orange/widgets/unsupervised/owdistances.py
```diff
@@ -119,7 +119,9 @@ def _check_sparse():
         def _fix_discrete():
             nonlocal data
             if data.domain.has_discrete_attributes() and (
-                    issparse(data.X) and getattr(metric, "fallback", None)
+                    issparse(data.X) and getattr(metric, "fallback",
+                                                 None) and metric is not
+                                distance.Jaccard
```
janezd: Is there a reason for not checking `metric.supports_sparse` instead of `metric is not distance.Jaccard`?

The condition that specifically checks for Jaccard a few lines later is needed because there is no specific flag signalling whether a metric supports distances by columns. Here, we have a flag to check, unless I overlooked something.

ajdapretnar: Jaccard is the only distance that supports discrete attributes via fallback. Other metrics fall back to sklearn's methods, which won't work with discrete.

janezd:
https://github.com/biolab/orange3/pull/3657/commits/551effc45154da84bc60cb76e73f0854e750da5a changes the base class so that it can handle numpy arrays if fallback is not provided. Perhaps we can stop passing numpy arrays to fallbacks and use fallbacks only for sparse data.

https://github.com/biolab/orange3/pull/3657/commits/fd1fd1930e1e6ee13e2c90cb028a1a9c90104a3a moves @ajdapretnar's code for sparse Jaccard to the proper class and also contains minor fixes in tests.

@ajdapretnar, you still have to compute the probabilities in sparse data fitter for handling missing values when the model is used on dense data.

janezd:
`Jaccard.compute_distances` is now ugly. Dense data is handled in `compute_distances` itself, while sparse data is in another method. It should be symmetric: one method for dense and one for sparse, and `compute_distances` calls one or another. That is, extract most of the code from `compute_distances`, except for the first two lines, into a separate method.

`sparse_jaccard` is also not a very good name: `jaccard` is already the class name. Maybe `_compute_sparse` and `_compute_dense`.

janezd:
Cosine tests failed because of wrong clipping: sklearn (correctly) clips to [0, 2], while our function for dense matrices (effectively) clipped to [0, 1]. This commit doesn't belong to this PR, but this PR adds tests for comparison between distances on sparse and dense matrices, which revealed this problem, hence its more practical to fix this here.

Orange/distance/distance.py
```diff
@@ -450,7 +453,7 @@ def fit_rows(self, attributes, x, n_vals):
         frequencies of non-zero values per each column.
         """
         if sp.issparse(x):
-            ps = None  # wrong!
+            ps = x.getnnz(axis=0)
```
ajdapretnar: If I understand the Cython code correctly, this should be ok. @janezd

PR #3662, opened 2019-03-07
[MNT] Confuse pylint to stop checking type for Setting

janezd:
#3637, @lanzagar argued against extra annotation needed to tell pylint that `Setting(0)` is an `int`. I agree. Without a better solution, we can only disable checking the type by confusing pylint with a fake annotation that is never executed.

The code has no side-effects, and should not have detrimental side-effects for pylint: currently it *thinks* it knows the type of setting, and after the PR it apparently *knows it doesn't*.

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -90,6 +90,24 @@ class Setting:
         return (self.default, )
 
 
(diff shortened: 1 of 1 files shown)


markotoplak:
Hahahaha! But please do not forget a commend why is this there. :D

PR #3665, opened 2019-03-08
Scrollable control areas of widgets

### LINKED ISSUE #3058: Widgets should work normally with 700 pixels of vertical space
Widgets should work normally with 700 pixels of vertical space. This is important for presentations on x768 projectors and for display on small laptop screens.

Some Windows laptops are quite small, have 1920x1080 resolution, and due to the save, have often scaling set to 150%. Which means their effective resolution is x720, which can make at least Scatter plot and MDS widgets unusable.

markotoplak:
##### Issue
Eventually attempts to fix #3058.

##### Includes
- [ ] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -364,6 +364,26 @@ class WidgetTest(GuiTest):
         finally:
             QTest.keyRelease(self.widget, Qt.Key_BassBoost, old_modifiers)
 
+    def test_minimum_size(self):
+        widget = getattr(self, "widget", None)
+        if widget is None:
+            return
+
+        def invalidate_cached_size_hint(w):
+            # as in OWWidget.setVisible
+            if w.controlArea is not None:
+                w.controlArea.updateGeometry()
+            if w.buttonsArea is not None:
+                w.buttonsArea.updateGeometry()
+            if w.mainArea is not None:
+                w.mainArea.updateGeometry()
+
+        invalidate_cached_size_hint(widget)
+        min_size = widget.minimumSizeHint()
+        # FIXME high-dpi screens
+        self.assertLess(min_size.width(), 800)
+        self.assertLess(min_size.height(), 700)
+
 
 class TestWidgetTest(WidgetTest):
     """Meta tests for widget test helpers"""
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -90,6 +90,17 @@ class WidgetTest(GuiTest):
 
     widgets = []  # type: List[OWWidget]
 
+    def __init_subclass__(cls, **kwargs):
+
+        def test_minimum_size(self):
+            widget = getattr(self, "widget", None)
+            if widget is None:
+                self.skipTest("minimum size not tested as .widget was not set")
+            self.check_minimum_size(widget)
+
+        if not hasattr(cls, "test_minimum_size"):
+            cls.test_minimum_size = test_minimum_size
+
     @classmethod
     def setUpClass(cls):
         """Prepare environment for test execution
@@ -364,10 +375,7 @@ class WidgetTest(GuiTest):
         finally:
             QTest.keyRelease(self.widget, Qt.Key_BassBoost, old_modifiers)
 
-    def test_minimum_size(self):
-        widget = getattr(self, "widget", None)
-        if widget is None:
-            return
+    def check_minimum_size(self, widget):
 
         def invalidate_cached_size_hint(w):
             # as in OWWidget.setVisible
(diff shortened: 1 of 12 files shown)


Orange/widgets/tests/base.py
```diff
@@ -364,6 +364,26 @@ def modifiers(self, modifiers):
         finally:
             QTest.keyRelease(self.widget, Qt.Key_BassBoost, old_modifiers)
 
+    def test_minimum_size(self):
+        widget = getattr(self, "widget", None)
+        if widget is None:
+            return
```
astaric: Or skip the tests using self.skipTest("minimum size needs to be tested manually")?

Otherwise, it will be marked as passing, even if it does not do anything.

Orange/widgets/tests/base.py
```diff
@@ -364,6 +364,26 @@ def modifiers(self, modifiers):
...
+            if w.controlArea is not None:
+                w.controlArea.updateGeometry()
+            if w.buttonsArea is not None:
+                w.buttonsArea.updateGeometry()
+            if w.mainArea is not None:
+                w.mainArea.updateGeometry()
+
+        invalidate_cached_size_hint(widget)
+        min_size = widget.minimumSizeHint()
+        # FIXME high-dpi screens
```
astaric: On mac, size returns size in virtual, not device pixels, so the values you have are ok.

markotoplak:
@astaric, thanks.

There is a problem with adding tests to the base class: because of the import, they are also run separately with every test file. And ideas how to prevent this? 

Before I did not even notice it because of the return. :)

astaric:
As far as I know, there are no simple tricks you can use.

The proper way would probably be a mixin that contains just the test methods, but maybe something could be done with metaclasses? A metaclass that removes the method when the baseclass is instantiated, but remembers in somewhere so it can put it back for the subclasses? :)

astaric:
Something like this :)
```
class MyMeta(type(GuiTest)):
    def __new__(cls, name, bases, namespace):

        if name == "WidgetTest":
            cls.tests = {}
            for key in list(namespace):
                if key.startswith("test"):
                    cls.tests[key] = namespace.pop(key)
        else:
            for key, method in cls.tests.items():
                namespace[key] = method

        obj = super().__new__(cls, name, bases, namespace)
        return obj

class WidgetTest(GuiTest, metaclass=MyMeta):
```

janezd:
Metaclasses probably work, but `__init_subclass__` may be more appropriate because this is something that `WidgetTest` wants to do to its subclasses.

```
class A:
    def __init_subclass__(cls):
        def f(self, x):
            return self.k * x
        cls.f = f

class B(A):
    def __init__(self, k):
        self.k = k

b = B(2)
print(b.f(3))  # prints 6


a = A()
print(a.f(3))  # error
```

markotoplak:
To make large widgets work with smaller screens, I added scroll areas to widgets' left sides. 

For now, to highlight potential issues, I added it to all widgets. Could anyone play around with this?

I added size hints so that scroll bars should not appear if you add a new widget, but if a user saved a workflow on a Mac (which has quite a compact interface) and opened on Ubuntu (with lots of empty space), I think the scroll bars would appear as window sizes are saved (didn't try this yet, though).

Therefore I recommend adding an attribute to OWWidget that controls if scroll panes are used for a certain widget. Thoughts? @lanzagar, @janezd, @ales-erjavec

lanzagar:
The scrollbar looks and functions well on my system.

I agree with this being optional and only manually turned on for certain widgets to reduce unwanted side effects like size changes for saved workflows on different systems.

One downside currently is that our (separate) very common problem of bad minimum widget size settings becomes even more evident. For example Scatter plot has a long standing issue where the main area can be reduced too much and looks bad when that sometimes happens (unintentionally). Control area was at least enforcing a vertical minimum size and now even that might be gone.
So together with this, I think we should set some minimum sizes, at least for widgets where scrolling will be turned on (but wouldn't hurt in some other cases too).

janezd:
It's unfortunate that macOS shows scroll bars only when scrolling (unless the user changes the default preference). This causes the visualization to resize. But we'll live with that.

I'd prefer moving code out of the huge Orange/widgets/widget.py. But gui.py is also one large mess. If you think this code will be useful elsewhere, please move it to gui.py. If you won't, then say, and we can merge.

ajdapretnar:
For reference: go to System Preferences - General - Show scroll bars.

thocevar:
There's a problem with hiding the control area. If the scrollbar is present and you hide and reopen control area, the sizes compute incorrectly and the scrollbar overlaps some content.

janezd:
Indeed, this also happens on macOS, if you set Show Scroll Bars in system preferences to Always.

markotoplak:
@thocevar, @janezd, thank you for the wrong resize problem if it is collapsed. It can reproduce it easily on a mac. I though I had seen it before and also fixed it while I was experimenting.

markotoplak:
I think that the problem with wrong size after control area show is not fixed.

As @janezd mentions, scroll bars, when hidden as on a Mac if you are not using a mouse :), should not resize the window, but I have no idea how to implement that properly.

I moved some reusable code to gui and the other class to inside OWWidget. I think this is ready to be merged.

PR #3666, opened 2019-03-11
setup.py: fix generating coverage output

markotoplak:
##### Issue
`python setup.py coverage` did not generate anything.

##### Description of changes
I added `coverage combine` into setup.py. I guess that command was probably necessary since we added concurrency to .coveragerc.

I was thinking of adding widget tests to the coverage command but I opted not to, because `python setup.py test` does not run them either.


lanzagar:
I think we should add `.coverage` to `.gitignore` as well (`htmlcov/*` is already there).

PR #3668, opened 2019-03-12
PCA: remove RemotePCA

robertcv:
##### Issue
RemotePCA randomly fails in tests.
##### Description of changes
As it is tightly connected to orangecontrib.remote and nobody uses that add-on, we decided to remove it. 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/unsupervised/owpca.py
```diff
@@ -46,8 +40,6 @@ class Outputs:
 
     ncomponents = settings.Setting(2)
     variance_covered = settings.Setting(100)
-    batch_size = settings.Setting(100)
-    address = settings.Setting('')
     auto_update = settings.Setting(True)
```
lanzagar: `auto_update` needs to be removed too right?

robertcv: resolved

PR #3669, opened 2019-03-12
[FIX]  VizRankDialog: Use extended thread pool to prevent segfaults

VesnaT:
##### Issue

When fed large datasets to owcorrelations, the widget exited with segmentation fault, (probably) due to insufficient stack size for created task.

To reproduce: Generate a dataset with at least 26 features and 3361 rows, and feed it to OWCorrelations.

##### Description of changes
- VizRankDialog now inherits ConcurrentMixin, which uses ThreadExecutor that provides sufficient stack size.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcorrelations.py
+++ b/Orange/widgets/data/owcorrelations.py
@@ -59,7 +59,7 @@ class KMeansCorrelationHeuristic:
     def __init__(self, data):
         self.n_attributes = len(data.domain.attributes)
         self.data = data
-        self.states = None
+        self.clusters = None
         self.n_clusters = int(np.sqrt(self.n_attributes))
 
     def get_clusters_of_attributes(self):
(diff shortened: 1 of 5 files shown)


ales-erjavec:
I get various
```
-------------------------- AssertionError Exception ---------------------------
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/utils/concurrent.py", line 954, in _on_task_done
    self.on_done(future.result())
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/visualize/utils/__init__.py", line 327, in on_done
    assert not self.state_count() or self.rank_model.rowCount() == self.state_count()
AssertionError
```
or
```
-------------------------- AssertionError Exception ---------------------------
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/utils/concurrent.py", line 954, in _on_task_done
    self.on_done(future.result())
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/visualize/utils/__init__.py", line 327, in on_done
    assert self.saved_state is None
AssertionError
```
if I  Pause/Continue multiple time during a run then let it run till the end (testing on [phoneme.tab](https://www.dropbox.com/s/z7d6ogn4uynsb01/phoneme.tab?dl=0))

VesnaT:
I just realized these asserts should be removed because an exception can occur while computing the score and therefore is not put into the table.
But while testing `phoneme.tab` that is not the case, since all the scores can be computed. So apparently there is a bug related to pausing calculation.

ales-erjavec:
Might I suggest just setting a larger initial stack size on the QThread instance, then maybe revisit this latter.

I think the ConcurrentMixin is no a good fit for this.

Issue #3526, opened 2019-01-11
OWTransform: add option to keep original features.


mstrazar:
##### Orange version
df87c8b58751af7365ea94d4fa76693fa0235c7d

##### Expected behavior
The Transform widget needs to retain original data, and place the embedding variables as meta features. See e.g. t-SNE widget.

##### Actual behavior
Original data is overridden by new features. While this is sometimes desirable, we need an option to chose between the two modalities. 

##### Steps to reproduce the behavior
See this workflow:
[tsne-transform.ows.zip](https://github.com/biolab/orange3/files/2749324/tsne-transform.ows.zip)

mstrazar:
Also, be sure not to loose any Table metadata.

mstrazar:
Fixed by #3549

PR #3673, opened 2019-03-13
[FIX] Transform: Replace 'Preprocess' input with 'Template Data' input

### LINKED ISSUE #3623: Transform widget should (probably) apply domain, not run preprocessors
Transform widget currently gets a list of preprocessors and runs a new data through them. This is unexpected. If Preprocess widget discretizes some data and we connect it to Transform, we expect transform to discretize the new data into the same intervals as Preprocess widget. Currently, it computes a new discretization.

We discussed three solutions. One is that Transform accepts both, a list of preprocessors and a list of transformers (fitted preprocessors). This is however confusing for the user (*which signal should I use*?) and also ambiguous if both signals are connected.

Another solution is that Transform accepts only a single signal, fitted preprocessors, actually stored as `Domain`. The drawback is that every widget (think Select Columns) would then need to provide a signal of this type.

The most flexible solution is that Transform accepts "template data". The function of the widget is then: it gets some data and it transforms it into the same "form" as the template data. If the template data has discretized features, it transforms the given data into the same discretized features. With this, we can probably simplify some widgets (e.g. PCA no longer needs to output preprocessors), and arbitrary widgets' outputs (think Select Columns, again) can be used as input for Transform.

I think we settled for the third option, but we need a good name for "template data" that will describe what it does and that won't imply that one can connect Select Rows to Transform.

VesnaT:
##### Issue

Fixes #3623 and #3675.

##### Description of changes
- remove 'Preprocessor' output from projection widgets (PCA, tSNE, Freeviz, Radviz, LinearProjection)
- OWTransform: use domain to transform data (instead of preprocessor)

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtransform.py
+++ b/Orange/widgets/data/owtransform.py
@@ -1,6 +1,10 @@
+from typing import Optional
+
+import numpy as np
+
 from Orange.data import Table, Domain
-from Orange.preprocess.preprocess import Preprocess, Discretize
 from Orange.widgets import gui
+from Orange.widgets.report.report import describe_data
 from Orange.widgets.settings import Setting
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.utils.widgetpreview import WidgetPreview
@@ -18,29 +22,29 @@ class OWTransform(OWWidget):
 
     class Inputs:
         data = Input("Data", Table, default=True)
-        preprocessor = Input("Preprocessor", Preprocess)
+        template_data = Input("Template Data", Table)
 
     class Outputs:
         transformed_data = Output("Transformed Data", Table)
 
     class Error(OWWidget.Error):
-        pp_error = Msg("An error occurred while transforming data.\n{}")
+        error = Msg("An error occurred while transforming data.\n{}")
 
     resizing_enabled = False
     want_main_area = False
 
     def __init__(self):
         super().__init__()
-        self.data = None
-        self.preprocessor = None
-        self.transformed_data = None
+        self.data = None  # type: Optional[Table]
+        self.template_domain = None  # type: Optional[Domain]
+        self.transformed_info = describe_data(None)  # type: OrderedDict
 
         info_box = gui.widgetBox(self.controlArea, "Info")
         self.input_label = gui.widgetLabel(info_box, "")
-        self.preprocessor_label = gui.widgetLabel(info_box, "")
+        self.template_label = gui.widgetLabel(info_box, "")
         self.output_label = gui.widgetLabel(info_box, "")
         self.set_input_label_text()
-        self.set_preprocessor_label_text()
+        self.set_template_label_text()
 
         box = gui.widgetBox(self.controlArea, "Output")
         gui.checkBox(box, self, "retain_all_data", "Retain all data",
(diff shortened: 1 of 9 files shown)


Orange/widgets/unsupervised/tests/test_owpca.py
```diff
@@ -205,9 +205,9 @@ def test_do_not_mask_features(self):
...
         self.send_signal(self.widget.Inputs.data, self.iris)
-        pp = self.get_output(self.widget.Outputs.preprocessor)
-        self.assertIsInstance(pp, Preprocess)
-        transformed_data = pp(self.iris[::10])
+        template = self.get_output(self.widget.Outputs.transformed_data)
+        data = self.iris[::10]
+        transformed_data = data.transform(template.domain)
         self.assertIsInstance(transformed_data, Table)
         self.assertEqual(transformed_data.X.shape, (15, 2))
         output = self.get_output(self.widget.Outputs.transformed_data)
```
janezd: Which function does this test test now? `data.transform`? :)

I suppose you can remove it. And also the one in test_owtsne; I believe (but may be wrong) that they just waste time.

Orange/widgets/data/owtransform.py
```diff
@@ -18,29 +17,29 @@ class OWTransform(OWWidget):
...
+        error = Msg("An error occurred while transforming data.\n{}")
 
     resizing_enabled = False
     want_main_area = False
 
     def __init__(self):
         super().__init__()
         self.data = None
-        self.preprocessor = None
+        self.template_data = None
```
janezd: Could you perhaps store just the domain (in `template_domain` instead of `template_data`)? It would save some space and contribute to clarity -- it would be more obvious (from the code if not from the input signal type) that the widget needs just the domain.

Orange/widgets/data/owtransform.py
```diff
@@ -75,27 +75,28 @@ def set_data(self, data):
...
         self.clear_messages()
         self.transformed_data = None
-        if self.data is not None and self.preprocessor is not None:
+        if self.data is not None and self.template_data is not None:
             try:
-                self.transformed_data = self.preprocessor(self.data)
-            except Exception as ex:   # pylint: disable=broad-except
-                self.Error.pp_error(ex)
+                self.transformed_data = self.data.transform(
+                    self.template_data.domain)
```
janezd: `transformed_data` is stored just for report? Consider pre-preparing (that part of) report and storing it instead.

VesnaT:
Since OWTransform now makes transformation using arbitrary domain, we should probably reconsider #3526.

Orange/widgets/data/owtransform.py
VesnaT: It is actually used many times (for setting labels, reporting, merging data). 
Of course, I could prepare all needed properties, but I'd rather not. It would make the code less clear, and I'd would have to save something anyway.

janezd: Using instance attributes to pass (temporary-ish) data between methods is like using global variables to pass data between functions. Attributes are meant for storing data that needs to be *stored*. (Though practicality beats purity.)

Please feel free to remind me of that when *I* am doing it. :)

janezd:
Is this still WIP?

What about renaming the widget - do we have a better name?

I agree we should reconsider #3526.

VesnaT:
It's WIP because of #3526.
Besides, any ideas on what to do with signal to make saved workflows work?

janezd:
#3526 can be dealt with later.

Name: let's find one tomorrow.

What about not caring about saved workflows in this case? The widget was not around for long so it couldn't had been used extensively, while supporting this, even if perhaps possible (though not entirely clear how), would include too much dead code.

PR #3676, opened 2019-03-14
[FIX] Test & Score: hide warnings for hidden scores.

### LINKED ISSUE #2088: Test & Score: AUC is not computed if some class value in missing in the test set
##### Orange version
master

Take iris data set, Logistic Regression, build the classifier on the whole data set and only use some instances for testing in Test&Score. Even if samples with two different class values are present in the test set, AUC is not computed.

I understand that AUC can not be computed with a test set with only a single class value, but two class values should be enough.

thocevar:
##### Issue
Fixes #2088

A user sees a warning although all displayed scores are successfully computed. The problematic score is not AUC but LogLoss, which the widget tries to compute (and fails) but is not displayed by default.

##### Description of changes

- hide warnings for hidden scores
- fix an error where the user couldn't display additional scores on Windows

We might want to consider fixing LogLoss for the case described in #2088.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -525,6 +525,7 @@ class OWTestLearners(OWWidget):
         for section in range(1, model.columnCount()):
             col_name = model.horizontalHeaderItem(section).data(Qt.DisplayRole)
             header.setSectionHidden(section, col_name not in self.shown_scores)
+        self.view.resizeColumnsToContents()
 
     def _update_stats_model(self):
         # Update the results_model with up to date scores.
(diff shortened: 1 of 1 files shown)


Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -584,13 +585,13 @@ def _update_stats_model(self):
...
             if stats is not None:
-                for stat in stats:
+                for stat, scorer in zip(stats, self.scorers):
                     item = QStandardItem()
                     if stat.success:
                         item.setText("{:.3f}".format(stat.value[0]))
                     else:
                         item.setToolTip(str(stat.exception))
-                        has_missing_scores = True
+                        has_missing_scores = scorer.name in self.shown_scores
```
janezd: Should this be `has_missing_scores = has_missing_scores or scorer.name in self.shown_scores`? It can already be `True`.

thocevar: Good catch.

PR #3681, opened 2019-03-15
In-place settings, and proper encoding and matching of lists of variables

janezd:
##### Issue

If setting is a list, writing it to the widget does not change the list in place but replaces it with a new list. This is a problem if the list is (wrapped in a) model. In particular, models behind list views wrap a setting, and they are not updated when settings are changed.

I encountered the problem in RadViz and Linear projection. The base projection widget, which is common to all projections (scatter plot, mds, tsne, networks...) has a conspicuous method `use_context`, which is there just to be overridden in Radviz and Linear projection. Currently it is more complicated, but after #3508 it will have to do something like

```
def use_context(self):
    self.selection_model.wrap(self.selected_variables)
```

where `selected_variables` is a setting (and a list behind the model at the same time). `use_context` is essentially a fix after application of a context.

##### Description of changes

- Class `Setting` has an additional flag `inplace`. If `True`, the setting is changed in-place. This requires (and checks) that the setting is a `list`, `dict` or `set`.
- `setattr`'s are replaced by calling a `Settings`'s method `apply_to_widget` that sets the setting in-place or not; handler's initialize still uses `setattr` and it already deep-copies mutable values
 
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -39,11 +39,13 @@ import pprint
 import time
 import warnings
 from operator import itemgetter
+from typing import Any
 
 from Orange.data import Domain, Variable
 from Orange.misc.environ import widget_settings_dir
 from Orange.util import OrangeDeprecationWarning
 from Orange.widgets.utils import vartype
+from Orange.widgets.gui import OWComponent
 
 log = logging.getLogger(__name__)
 
@@ -106,6 +108,27 @@ class Setting:
 if 1 == 0:
     class Setting:  # pylint: disable=function-redefined
         pass
+
+
+def _apply_setting(setting: Setting, instance: OWComponent, value: Any):
+    """
+    Set `setting` of widget `instance` to the given `value`, in place if
+    possible.
+
+    If old and new values are of the same type, and the type is either a list
+    or has methods `clear` and `update`, setting is updated in place. Otherwise
+    the function calls `setattr`.
+    """
+    target = getattr(instance, setting.name, None)
+    if type(target) is type(value):
+        if isinstance(value, list):
+            target[:] = value
+            return
+        elif hasattr(value, "clear") and hasattr(value, "update"):
+            target.clear()
+            target.update(value)
+            return
+    setattr(instance, setting.name, value)
 
 
 class SettingProvider:
@@ -261,7 +284,7 @@ class SettingProvider:
         """
         for setting, _data, inst in self.traverse_settings(data, instance):
             if setting.name in _data and inst is not None:
-                setattr(inst, setting.name, _data[setting.name])
+                _apply_setting(setting, inst, _data[setting.name])
 
     def get_provider(self, provider_class):
         """Return provider for provider_class.
@@ -570,7 +593,7 @@ class SettingsHandler:
         """
         for setting, _, inst in self.provider.traverse_settings(instance=instance):
             if setting.packable:
-                setattr(inst, setting.name, setting.default)
+                _apply_setting(setting, inst, setting.default)
 
 
 class ContextSetting(Setting):
@@ -855,7 +878,7 @@ class ContextHandler(SettingsHandler):
                 continue
 
             value = self.decode_setting(setting, data[setting.name])
-            setattr(instance, setting.name, value)
+            _apply_setting(setting, instance, value)
             if hasattr(setting, "selected") and setting.selected in data:
                 setattr(instance, setting.selected, data[setting.selected])
 
@@ -1006,7 +1029,7 @@ class DomainContextHandler(ContextHandler):
                 continue
 
             value = self.decode_setting(setting, data[setting.name], domain)
-            setattr(instance, setting.name, value)
+            _apply_setting(setting, instance, value)
             if hasattr(setting, "selected") and setting.selected in data:
                 setattr(instance, setting.selected, data[setting.selected])
 
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -1033,21 +1033,34 @@ class DomainContextHandler(ContextHandler):
             if hasattr(setting, "selected") and setting.selected in data:
                 setattr(instance, setting.selected, data[setting.selected])
 
+    @staticmethod
+    def encode_variable(var):
+        return var.name, 100 + vartype(var)
+
     def encode_setting(self, context, setting, value):
         if isinstance(value, list):
-            return copy.copy(value)
-        elif isinstance(setting, ContextSetting):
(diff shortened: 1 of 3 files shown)


janezd:
@astaric, what's your opinion about this?

It does its job on #3508 (except for an unrelated problem, setting the number of selected instances, for which I'd need a mutable `int` :( )

astaric:
Is there any downside to this? Could all lists/dicts/sets be applied like this?

I do not mind the change, I would probably put the apply code on the SettingsHandler, since Setting is currently used as a data object. Or, if the code stays on setting, additional functionality (such as encode and decode) could move to setting as well.

For non-mutable types, one could use [connect_control](https://github.com/biolab/orange3/blob/master/Orange/widgets/gui.py#L142) to get notified when the value changes. (gui components use this machinery to update themselves, models could as well)

Alternatively, if model extends OWComponent, it can contain its own settings, that get set directly. (Assuming the reference to the model is set on the widget and the model is marked as a SettingProvider).

janezd:
@astaric, I eliminated the flag: settings are now always in-place if they're a `list` or have `update` and `clear` methods. If anybody holds a reference to a setting, he probably expects the value to change. If he wanted a copy, he'd make a copy.

1. I don't know where to put the method. I like keeping `Setting` as data class. But the method is needed in provider and in handler. For now it's in provider although also called from handler.

2. Unrelated setting-related question: refactored Radviz and Linear projection have a list of `Variable` as setting. [`match_list`](https://github.com/biolab/orange3/blob/0d19866a0f75579cb3d60b2fd503b586c3992ce7/Orange/widgets/settings.py#L1075) calls [`_is_valid_item`](https://github.com/biolab/orange3/blob/0d19866a0f75579cb3d60b2fd503b586c3992ce7/Orange/widgets/settings.py#L1108) for each element of the list. The latter returns `True` if the element is not a tuple -- which it is not. Hence the list of variables will always match (and Radviz reuses the Iris attributes on heart disease). Do I understand correctly the the problem in [`DomainContextHandler.encode_setting`](https://github.com/biolab/orange3/blob/0d19866a0f75579cb3d60b2fd503b586c3992ce7/Orange/widgets/settings.py#L1014)? Should I fix it by checking whether all elements of the list are instances of `Variable` (or `None`) and in this case encode list elements as tuples? And change `decode_settings` accordingly? And perhaps use `-3` to denote encoded lists of variables (just like `-2` is now used for "non-variables")?

astaric:
> 1. I don't know where to put the method. I like keeping `Setting` as data class. But the method is needed in provider and in handler. For now it's in provider although also called from handler.

Since it is a static method, it could also be turned into a function :)

> 2. Unrelated setting-related question: refactored Radviz and Linear projection have a list of `Variable` as setting. [`match_list`](https://github.com/biolab/orange3/blob/0d19866a0f75579cb3d60b2fd503b586c3992ce7/Orange/widgets/settings.py#L1075) calls [`_is_valid_item`](https://github.com/biolab/orange3/blob/0d19866a0f75579cb3d60b2fd503b586c3992ce7/Orange/widgets/settings.py#L1108) for each element of the list. The latter returns `True` if the element is not a tuple -- which it is not. Hence the list of variables will always match (and Radviz reuses the Iris attributes on heart disease). Do I understand correctly the the problem in [`DomainContextHandler.encode_setting`](https://github.com/biolab/orange3/blob/0d19866a0f75579cb3d60b2fd503b586c3992ce7/Orange/widgets/settings.py#L1014)? Should I fix it by checking whether all elements of the list are instances of `Variable` (or `None`) and in this case encode list elements as tuples? And change `decode_settings` accordingly? And perhaps use `-3` to denote encoded lists of variables (just like `-2` is now used for "non-variables")?

You are correct, changing encode_setting to mark lists of variables somehow and using that information when decoding is the way to go.
Heuristically deciding whether a value is a list of variable should work, another option is to subclass ContextSetting into VariableListSetting (or similar) and use that to explicitly mark settings which contain lists of variables.

janezd:
> Since it is a static method, it could also be turned into a function :)

I thought about it, too, but I hoped you had a better idea. Now it's a function.

I modified encoding of lists so that each element is encoded/decoded by the same rules as individual setting. It works for Radviz and Linear projection, where it all started, but core dumps on Correlations (ironically: it happens because it internally encodes list of Variables as tuples, exactly like this PR!).

Two more questions:

- While implementing this and checking for any other places that need to be modified, I saw that a significant amount of the code in many methods is dedicated to settings with an attribute `selected`, which, as I recall, deals with storing the selection in a list view (whereas the content of the list view is another setting). We don't use this, AFAIK, it is not documented and it's probably unsafe. Eliminate?

- Another antique are variables stored as strings instead of a tuple. These are unreliable and may also result in false matches (a string accidentally matches a variable name). AFAIK, we eliminated them in core widgets, but nobody knows about add-ons. Deprecate with some considerable delay before elimination?

astaric:
> * While implementing this and checking for any other places that need to be modified, I saw that a significant amount of the code in many methods is dedicated to settings with an attribute `selected`, which, as I recall, deals with storing the selection in a list view (whereas the content of the list view is another setting). We don't use this, AFAIK, it is not documented and it's probably unsafe. Eliminate?

Sure. At the time Settings were ported from 2.0, all the "features" were included. Reservoirs were already removed, and nobody  complained since nobody knew they ever existed. I guess it is going to be the same for selections.

> * Another antique are variables stored as strings instead of a tuple. These are unreliable and may also result in false matches (a string accidentally matches a variable name). AFAIK, we eliminated them in core widgets, but nobody knows about add-ons. Deprecate with some considerable delay before elimination?

Removing this will break support for old workflows (or require some pretty involved migrate_settings for all the widgets that ever used this). Deprecate, and after some time remove encoding part only?

While deprecating, you could also consider deprecating ContextSetting's attributes:
```
self.not_attribute
self.exclude_attributes
self.exclude_metas
self.required
```
The first one made sense, when setting contained variable name, you could use it to avoid false matches.
The second and third could be used to limit a setting to only metas or attributes + classvars. Since meta variables are now considered equal, they both default to False.
The last one is a nice idea, but not sure if anyone else knows what it does, or uses it.

janezd:
Thanks. I'll 
- remove `selected`,
- deprecate variables encoded as strings and remove encoding (but not decoding) in the future, as you suggested,
- remove `not_attribute`, `exclude_attributes`, `exclude_metas` (unless they're used somewhere), but keep `required`, which may be useful and should be used.

But I'll do this in #3697. I'd prefer this PR merged without further additions because it's required by #3508.

PR #3683, opened 2019-03-19
distance.py: Fix zero division in Jaccard

ajdapretnar:
##### Issue

Jaccard did not handle two all zero instances.

##### Description of changes
Check to avoid division by zero. Add tests for all distances.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


ajdapretnar:
TODO:
All tests pass except for Cosine. Cosine returns nan for all zero instances when dense and 1 when sparse.

PR #3693, opened 2019-03-22
[FIX] OWTable: Don't set selection when there is no data on input

### LINKED ISSUE #3451: OWTable crashes when loading a missing file with selected data
##### Orange version

master

##### Issue
A similar issue to what is described in #3449. Loading a workflow where the source data file can no longer be found causes crashes. One would expect that in this case, `None` would be sent to inputs (just like removeing a widget signal on canvas), however it seems that this is not the case?

I've gone through most of the widgets in the Data and Visualize section and haven't been able to reproduce this bug, but it might be hiding in other widgets as well.

```
  File "/home/pavlin/dev/orange3/Orange/canvas/scheme/widgetsscheme.py", line 1073, in process_signals_for_widget
    handler(*args)
  File "/home/pavlin/dev/orange3/Orange/widgets/data/owtable.py", line 535, in set_dataset
    self.set_selection()
  File "/home/pavlin/dev/orange3/Orange/widgets/data/owtable.py", line 775, in set_selection
    model = view.model()
AttributeError: 'NoneType' object has no attribute 'model'

```

1. Open a workflow and connect File with some data to a Table, select some rows and save the workflow
2. Remove/rename the data file in question so that the File widget can no longer find it
3. Re-open the workflow in Orange → crash

rokgomiscek:
##### Issue
Fixes #3451.

##### Description of changes
Fixes the bug by adding an additional check.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -516,6 +516,9 @@ class OWDataTable(widget.OWWidget):
             current = self.tabs.currentWidget()
             if current is not None:
                 self.set_info(current._input_slot.summary)
+        else:
+            self.__pending_selected_rows = None
+            self.__pending_selected_cols = None
 
         self.tabs.tabBar().setVisible(self.tabs.count() > 1)
         self.openContext(data)


janezd:
This fixes the symptom, but the real problem is that selection is non-empty although there is no data. Such state is invalid and may cause other problems elsewhere in the widget. Could you investigate why this happens and how to prevent it?

rokgomiscek:
I think this now fixes the problem. The old selection was stored in __pending_selected_rows and __pending_selected_cols, which were used to restore selected_rows and selected_cols respectively. My previous fix didn't empty the selections, only didn't pass it forward.

PR #3694, opened 2019-03-22
image export: export PDF directly for Qt >= 5.12

### LINKED ISSUE #2887: Saving images as PDF does not work as it should
##### Orange version
All versions tested since 3.4.2. (3.4.2, 3.4.3, 3.4.4, 3.4.5, 3.5.0, 3.9, 3.10).

##### Expected behavior
WYSIWYG

##### Actual behavior
screenshot_20180125_103312
screenshot_20180125_103306

PDF was introduced in #2210 . See also ##2108 .

markotoplak:
##### Issue
Fixes #2887 for Qt >= 5.12.

##### Description of changes
Use Qt's PDF Writer for output on Qt >= 5.12, because some bugs with line widths and fonts prevent this to work properly on older Qt version.

A potential problem: because I wanted to support multiple formats with the same extension, I changed the format of OWWidget.graph_writers from dict to a list. This could crash unknown add-ons that were touching it. I do not see a good workaround though.

##### Includes
- [X] Code changes
- [X] (Some) tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/io.py
+++ b/Orange/widgets/io.py
@@ -3,7 +3,7 @@ import tempfile
 from warnings import warn
 
 from AnyQt import QtGui, QtCore, QtSvg
-from AnyQt.QtCore import QMimeData
+from AnyQt.QtCore import QMimeData, QMarginsF
 from AnyQt.QtWidgets import (
     QGraphicsScene, QGraphicsView, QWidget, QApplication
 )
@@ -27,8 +27,8 @@ class ImgFormat(FileFormat):
         raise NotImplementedError
 
     @staticmethod
-    def _get_target(scene, painter, buffer):
-        raise NotImplementedError
+    def _get_target(scene, painter, buffer, source):
+        return QtCore.QRectF(0, 0, source.width(), source.height())
 
     @staticmethod
     def _save_buffer(buffer, filename):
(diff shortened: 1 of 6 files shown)


ajdapretnar:
Looks ok on Windows. Similar to @lanzagar's one with thick dots, but at least it is readable. It doesn't bother me, really.

[iris.pdf](https://github.com/biolab/orange3/files/3047018/iris.pdf)

ajdapretnar:
If all in favour, I would merge this. How about codecov/patch?

markotoplak:
@ajdapretnar, please try it with something else than scatterplot. What you are showing is PDF generated through Matplotlib, which does not use the new code at all.

Coverage is a problem, because I am only using some code with PyQt >= 5.12.

ajdapretnar:
@markotoplak Ok, any suggestions what/how I should test it?

markotoplak:
@ajdapretnar, with any visualization, even scatterplot, but be careful to select the writer without "from matplotlib".

ajdapretnar:
Right. I didn't realize Scatter Plot's .pdf option saves from matplotlib as opposed to MDS's regular .pdf. Here's the regular option from MDS. Looks even nicer than the previous one.

[mds.pdf](https://github.com/biolab/orange3/files/3047161/mds.pdf)

markotoplak:
@ajdapretnar, also, your scatterplot should now show 2 pdf options.

markotoplak:
And @ajdapretnar, you are either testing on PyQt < 5.12 or you are not on this branch.

ajdapretnar:
All done now. I was indeed left in master after some mucking about. Still the result is good for this branch as well. The lines are a little thick, but not an issue of this PR I suppose.

[mds2.pdf](https://github.com/biolab/orange3/files/3047253/mds2.pdf)

PR #3695, opened 2019-03-22
[ENH] k-Means: Output centroid labels

### LINKED ISSUE #3687: K-means centroids output has no labels
K-means has an output containing cluster centroids. Plotting them in e.g. MDS (or doing whatever else with them) is problematic, since the user can't label the points and does not know which cluster they correspond to (to see for example which clusters are similar).

I propose we add a meta column "Cluster" with cluster ids "C1", "C2", ... just as in the default Data output from K-means. Up for discussion - for the sake of consistency we could add the Silhouette meta attribute too.

janezd:
##### Issue

Fixes #3687.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owkmeans.py
+++ b/Orange/widgets/unsupervised/owkmeans.py
@@ -11,6 +11,7 @@ from Orange.clustering import KMeans
 from Orange.clustering.kmeans import KMeansModel, SILHOUETTE_MAX_SAMPLES
 from Orange.data import Table, Domain, DiscreteVariable, ContinuousVariable
 from Orange.data.util import get_unique_names
+from Orange.preprocess.impute import ReplaceUnknowns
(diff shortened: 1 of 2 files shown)


lanzagar:
The output table with centroids has no `name` (was like that before this PR already). Can we think of something (anything) and set that attribute (so that e.g. Source from the Concatenate widget output makes more sense)?
I am fine with any naming scheme, maybe something like:
name (centroids), centroids - name, ...
where `name = self.data.name`

janezd:
I went for "name centroids", for instance "iris centroids".

I didn't write a separate test but just added assertions to existing tests. Test do not mock clustering (I won't add mocking within this PR - I tried but it's not trivial), so separate tests would slow down the testing suite.

PR #3697, opened 2019-03-26
Settings cleanup

janezd:
##### Issue

Context settings contain obsolete unused undocumented code that can be removed or deprecated (and removed).

##### Description of changes

- Remove 'selected' flag that could be used (but was not used in Orange 3) for storing selections in list views.
- Remove attribute `not_attribute`.
- Warn that support for variables stored as string settings will be removed in 3.26.
- Require that `settings_to_widget` and `settings_from_widget` support `*args`; exception for those not accepting it was supposed to be removed in 3.10.
- PyLint.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -937,6 +937,13 @@ class DomainContextHandler(ContextHandler):
                 .format(name), OrangeDeprecationWarning
             )
 
+    @staticmethod
+    def _warn_about_str_var_settings(setting):
+        warnings.warn(
+            "Storing variables as strings in settings is deprecated.\n"
+            "Support for this will be dropped in Orange 3.26.\n"
+            f"Change f{setting.name} to store an instance of `Variable`.")
+
     def encode_domain(self, domain):
         """
         domain: Orange.data.domain to encode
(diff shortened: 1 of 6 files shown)


janezd:
Last Friday, @lanzagar argued that some combos should only show attributes, but not metas. If we decide so, we should not remove the `exclude_metas` flag and, to keep it symmetrical, also not `exclude_attributes`. We need to discuss this.

janezd:
Even without making a final decision about exclusion of meta attributes from combos, it appears that we could decide so in the future, or that some future widget might require it. We should thus retain `exclude_metas` and `exclude_attributes`.

astaric:
What was the use case for having a combo that displays only metas or only attributes? (having it written down will help if/when we consider removing it again)

Orange/widgets/settings.py
```diff
@@ -1016,23 +1023,24 @@ def settings_to_widget(self, widget, domain, *args):
     def encode_variable(var):
         return var.name, 100 + vartype(var)
 
-    def encode_setting(self, context, setting, value):
+    @classmethod
```
astaric: What is the reasoning behind changing this to a class method?

janezd: Class methods are easier/cleaner to test because they don't require having an instance.  I actually test this method like this, https://github.com/biolab/orange3/pull/3697/files#diff-0528176a310e674da65e2227537d9a05R285.

I see no disadvantages; derived classes can still have a non-class method and I do not see having `cls` instead of `self` as a different signature. (Neither does pylint consider them as different.)

Besides, I locally use a stricter setting for pylint with just a few disabled warnings,

```
disable=
    no-name-in-module, no-member  # too many false positives from Qt
    too-many-ancestors, # I don't think this is a problem
    no-else-return      # else's may actually improve readability
```

so both pylint and pycharm suggest turning this into a class method.

astaric: Thanks for the explanation. I have nothing against the change, I was just curious.

janezd:
Subtitles regarding `exclude_metas`:
- @lanzagar works on data with lots of metas that overcrowd the VizRank. Currently, there is no way to have a variable in the data, but not include it in VizRank. (An obvious solution for this particular case, though, is a checkbox in VizRank.)
- @BlazZupan argues that the whole point of metas is that they are not included in models. Projections are a kind of model, hence metas should occur only as colors, size..., but not as coordinates.

This open, early discussion indicates that we may still need these flags some time in the future. If they were not in the code, I wouldn't add them. But since they are, I'd prefer to not eagerly remove them.

Orange/widgets/tests/test_domain_context_handler.py
```diff
@@ -276,6 +282,23 @@ def test_backward_compatible_params(self):
             self.assertIn(OrangeDeprecationWarning,
                           [x.category for x in w])
 
+    def test_deprecated_str_as_var(self):
+        if LooseVersion(Orange.__version__) >= LooseVersion("3.26"):
+            # pragma: no cover
+            self.fail("Remove support for variables stored as string settings "
+                      "and this test.")
```
lanzagar: Should we add a comment here or open an issue about what this removal of support for string settings includes?
My wish for that list is that it contains the removal of the constant `100` from the code :)

PR #3699, opened 2019-03-28
[RFC] [NOMERGE] MessagesWidget: Stylesheet for experimenting

### LINKED ISSUE #3377: Widget errors are hard to notice
One of my colleagues commented that looking for red triangles is part of the fun in using Orange. 

If the widget is open, errors (in the status bar) are hard to notice. While that could be fine for information and (maybe) warnings, errors should be more noticeable. I do not know a good way to do it, but #3372  inspired me to write this issue.

Also, they could include hyperlinks (perhaps into documentation).

janezd:
##### Issue

Fixes #3377 - errors are too easy to overlook.

This is not a proper implementation, it is meant just for experimenting. If we confirm the change, I (or somebody else) will do it properly.

Screenshot 2019-03-28 at 19 02 44

@markotoplak, @irgolic, @BlazZupan and everybody else, play and design something better if you wish. Test it by, say, giving housing to logistic regression.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/utils/messagewidget.py
+++ b/Orange/widgets/utils/messagewidget.py
@@ -344,6 +344,14 @@ class MessagesWidget(QWidget):
             openExternalLinks=self.__openExternalLinks,
             sizePolicy=QSizePolicy(QSizePolicy.Preferred, QSizePolicy.Minimum)
         )
+        self.__textlabel.setStyleSheet("""
+            QLabel {
+                padding-left: 8px;
+                padding-right: 8pt;
+                background-color: #b00;
+                color: white;
+                font-weight: 550;
+                }""")
         #: Indicator that extended contents are accessible with a click on the
         #: widget.
         self.__popupicon = QLabel(


ajdapretnar:
It works well on Windows. Two examples:

logreg
distance-warning

However, some design touches should be added.
On OSX, we are used to rounded corners and having sharp edges looks a tad strange. We probably also want to go with unique colors for each type of error message, as @janezd suggested. But then yellow would require black letters I suppose?

janezd:
The question was actually not whether it works on Windows - the current implementation is totally wrong anyway. I'd like to have the design. Round corners would definitely look better. As for colors, only errors will be colored, while warnings and information don't need extra exposure and highlighting them would be distracting, IMHO.

ajdapretnar:
I do not like the way it looks currently. Is it noticeable? Yes. Could it look better? In my opinion - yes. Do I have an idea how it should look like? Not really. Perhaps:
- rounded corners
- no space at the beginning of the message
- experiment with removing bold
- use appropriate colors for different types
- figure out how to include the icon into the message (not necessarily into the line, but make them look a part of the same section)

irgolic:
From what I understand, the problem is in the error's noticeability, not its readability.

Could try adding an animation like this on update (more discreet):

blink

Or this (less discreet, concept):

error-bounce-animation

ajdapretnar:
I don't like the animations very much - I prefer the highlighted option. But that's just my opinion.

markotoplak:
@irgolic, these are wonderful! A user learns about the error message and they are much more elegant than this PR. 

@ajdapretnar, these would not need to repeat indefinitely so they wouldn't do much damage... Flashing twice, or bouncing a few times could work.

PR #3711, opened 2019-03-29
OWSQL: PyLint

janezd:
Pylint.


Orange/widgets/data/owsql.py
```diff
@@ -262,6 +253,7 @@ def connect(self):
         except BackendError as err:
             error = str(err).split('\n')[0]
             self.Error.connection(error)
+            self.highlight_error(error)
```
lanzagar: This sets the error highlights when something goes wrong, but they are never cleared!

PR #3715, opened 2019-04-01
Update owneuralnetwork.py to handle initial random state for ANN learner

WolframRinke:
Added the option to preset the initail random value. Extra input field in the form.
0 = randomizer behavior like before
>0 ranomizer is initiated with the defined value.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owneuralnetwork.py
+++ b/Orange/widgets/model/owneuralnetwork.py
@@ -83,6 +83,7 @@ class OWNNLearner(OWBaseLearner):
     max_iterations = Setting(200)
     alpha_index = Setting(0)
     settings_version = 1
+    random_state = 1
 
     alphas = list(chain([x / 10000 for x in range(1, 10)],
                         [x / 1000 for x in range(1, 10)],
@@ -135,6 +136,14 @@ class OWNNLearner(OWBaseLearner):
                 None, self, "max_iterations", 10, 10000, step=10,
                 label="Max iterations:", orientation=Qt.Horizontal,
                 alignment=Qt.AlignRight, callback=self.settings_changed))
+        
+        form.addRow(
+            "Random state:",
+            gui.spin(
+                None, self, "random_state", 0, 100000, step=20,
+                label="Random state:", orientation=Qt.Horizontal,
+                alignment=Qt.AlignRight, callback=self.settings_changed))
+
 
     def set_alpha(self):
         self.strength_C = self.alphas[self.alpha_index]
@@ -154,20 +163,33 @@ class OWNNLearner(OWBaseLearner):
         gui.button(self.apply_button, self, "Cancel", callback=self.cancel)
 
     def create_learner(self):
-        return self.LEARNER(
-            hidden_layer_sizes=self.get_hidden_layers(),
-            activation=self.activation[self.activation_index],
-            solver=self.solver[self.solver_index],
-            alpha=self.alpha,
-            max_iter=self.max_iterations,
-            preprocessors=self.preprocessors)
+        learner = None
+        if (self.random_state > 0):
+            learner = self.LEARNER(
+                hidden_layer_sizes=self.get_hidden_layers(),
+                activation=self.activation[self.activation_index],
+                solver=self.solver[self.solver_index],
+                alpha=self.alpha,
+                max_iter=self.max_iterations,
+                preprocessors=self.preprocessors,
+                random_state=self.random_state)
+        else:
+            learner = self.LEARNER(
+                hidden_layer_sizes=self.get_hidden_layers(),
+                activation=self.activation[self.activation_index],
+                solver=self.solver[self.solver_index],
+                alpha=self.alpha,
+                max_iter=self.max_iterations,
+                preprocessors=self.preprocessors)
+        return learner
 
     def get_learner_parameters(self):
         return (("Hidden layers", ', '.join(map(str, self.get_hidden_layers()))),
                 ("Activation", self.act_lbl[self.activation_index]),
                 ("Solver", self.solv_lbl[self.solver_index]),
                 ("Alpha", self.alpha),
-                ("Max iterations", self.max_iterations))
+                ("Max iterations", self.max_iterations),
+                ("Random state", self.random_state))
 
     def get_hidden_layers(self):
         layers = tuple(map(int, re.findall(r'\d+', self.hidden_layers_input)))


janezd:
Thanks for this pull request. 

We had a discussion at our weekly meeting. You're right about the issue, allowing the user to choose between replicable or random behavior is a good idea. But we don't like allowing the user to set a (favorable? :)) seed. Somebody mentioned another widget (random forest) that allows it, so we said we'll remove it from there, too.

So we agreed to have a checkbox "Replicable training" (if anybody has a better phrase, welcome), but without letting the user manipulate the seed. This will thus be the same as in Data Sampler widget.

I'm still going to add a few comments, as guidelines for the next time you (hopefully) decide to contribute.

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -83,6 +83,7 @@ class OWNNLearner(OWBaseLearner):
     max_iterations = Setting(200)
     alpha_index = Setting(0)
     settings_version = 1
+    random_state = 1
```
janezd: This should be `random_state = Setting(1)` so the widget will remember this setting. Also, it should be above `settings_version = 1` -- because it's a setting, while `settings_version` has a number that is increased when settings are changed in such a way that they need to be migrated.

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -135,6 +136,14 @@ def add_main_layout(self):
                 None, self, "max_iterations", 10, 10000, step=10,
                 label="Max iterations:", orientation=Qt.Horizontal,
                 alignment=Qt.AlignRight, callback=self.settings_changed))
+        
+        form.addRow(
+            "Random state:",
```
janezd: We, programmers, know what "Random state" this means, but it may not be understood by muggles. We try to explain these in non-technical terms -- by what it means (replicability...) and not how it's implemented (fixed initial random state).

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -135,6 +136,14 @@ def add_main_layout(self):
...
+            gui.spin(
+                None, self, "random_state", 0, 100000, step=20,
```
janezd: What does `step=20` accomplish? Is the difference between 10 and 30 larger than between 10 and 11? :)

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -154,20 +163,33 @@ def setup_layout(self):
...
+                random_state=self.random_state)
+        else:
+            learner = self.LEARNER(
+                hidden_layer_sizes=self.get_hidden_layers(),
+                activation=self.activation[self.activation_index],
+                solver=self.solver[self.solver_index],
+                alpha=self.alpha,
+                max_iter=self.max_iterations,
+                preprocessors=self.preprocessors)
+        return learner
```
janezd: These two calls are the same, except that the second uses the default values for `random_state`. You should have a single call with a variable parameter.

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -154,20 +163,33 @@ def setup_layout(self):
...
     def create_learner(self):
-        return self.LEARNER(
-            hidden_layer_sizes=self.get_hidden_layers(),
-            activation=self.activation[self.activation_index],
-            solver=self.solver[self.solver_index],
-            alpha=self.alpha,
-            max_iter=self.max_iterations,
-            preprocessors=self.preprocessors)
+        learner = None
+        if (self.random_state > 0):
```
janezd: No parentheses here in Python.

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -154,20 +163,33 @@ def setup_layout(self):
...
 
     def create_learner(self):
-        return self.LEARNER(
-            hidden_layer_sizes=self.get_hidden_layers(),
-            activation=self.activation[self.activation_index],
-            solver=self.solver[self.solver_index],
-            alpha=self.alpha,
-            max_iter=self.max_iterations,
-            preprocessors=self.preprocessors)
+        learner = None
```
janezd: This is redundant; `learner` will be set to something in one of the following lines.

Orange/widgets/model/owneuralnetwork.py
WolframRinke: I agree.

Orange/widgets/model/owneuralnetwork.py
WolframRinke: Its is not necessary to enter an initial value as long there is a way to fix it or use the same start value each time. So probalby a checkbox might do the same.

Orange/widgets/model/owneuralnetwork.py
WolframRinke: No issue, it can be what ever value you like. ;-)
So a step size of 1 is fine, but if we drop direct entry it does not make sense anyway.

Orange/widgets/model/owneuralnetwork.py
WolframRinke: Exactly. This is the reason for the conditional statement here. As I don't see a way to access the random state from the underlying library, the easiest way is to have two different calls. One with the fixed random.state and the other without it. Or you fix it in the random.state setup. For me this is more clear, than messing with hidden object states.

Orange/widgets/model/owneuralnetwork.py
WolframRinke: Makes it more readable to me without any drawbacks.

Orange/widgets/model/owneuralnetwork.py
WolframRinke: Than we can drop it.

WolframRinke:
Thank you for considering this feature.

Orange/widgets/model/owneuralnetwork.py
ajdapretnar: I think @janezd was referring to the naming of the checkbox. It should say something like 'Replicable', the way it is done in Data Sampler.

WolframRinke: I agree. 'Replicable' is fine for me.

BTW: Is there any further action expected from my side, as the  status has been closed?
Shall I open another pullrequest with the code changes or will this be implemented by someone else?

Orange/widgets/model/owneuralnetwork.py
ajdapretnar: You pass the value of self.random_state here (probably 0) or None if the checkbox is off.

WolframRinke: I see.

Orange/widgets/model/owneuralnetwork.py
janezd: > Is there any further action expected from my side

I opened #3724. I suppose it will be implemented on Friday.

However, you can do it if you wish. In this case, consider the comments that I've written as guidelines. Including removing redundant parentheses in conditions. :)

WolframRinke: Ok, great.
So I will consider your suggestions and do the changes and test it within my application.

PR #3721, opened 2019-04-04
[ENH] Transform: Values of primitive variables as feature names

### LINKED ISSUE #3677: Transpose: Support ContinousVariable as Feature name
The Transpose widget should support naming features after ContinousVariables in metas.

##### Orange version
3.20.1

##### Expected behavior
Select ContinousVariable from metas, str(number) is set as Feature name in transposed table.

##### Actual behavior
"From meta attribute" remains greyed out, because:

````
DomainModel(
            order=DomainModel.METAS, valid_types=StringVariable,
            alphabetical=True)
````

### PR DESCRIPTION
**Pull request by VesnaT**
##### Issue

Implements #3677

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -20,7 +20,7 @@ from Orange.data import (
 )
 from Orange.data.util import SharedComputeValue, vstack, hstack, \
     assure_array_dense, assure_array_sparse, \
-    assure_column_dense, assure_column_sparse
+    assure_column_dense, assure_column_sparse, get_unique_names_duplicates
 from Orange.statistics.util import bincount, countnans, contingency, \
     stats as fast_stats, sparse_has_implicit_zeros, sparse_count_implicit_zeros, \
     sparse_implicit_zero_weights
@@ -1604,8 +1604,8 @@ class Table(MutableSequence, Storage):
         return contingencies, unknown_rows
 
     @classmethod
-    def transpose(cls, table, feature_names_column="", meta_attr_name="Feature name",
-                  feature_name="Feature"):
+    def transpose(cls, table, feature_names_column="",
+                  meta_attr_name="Feature name", feature_name="Feature"):
         """
         Transpose the table.
 
@@ -1614,6 +1614,7 @@ class Table(MutableSequence, Storage):
             use for feature names
         :param meta_attr_name: str - name of new meta attribute into which
             feature names are mapped
+        :param feature_name: str - default feature name prefix
         :return: Table - transposed table
         """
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: Orange/data/table.py [hunk 1]
```diff
@@ -1625,10 +1626,16 @@ def transpose(cls, table, feature_names_column="", meta_attr_name="Feature name"
...
         # - arbitrary meta column to feature names
         self.X = table.X.T
-        attributes = [ContinuousVariable(str(row[feature_names_column]))
-                      for row in table] if feature_names_column else \
-            [ContinuousVariable(feature_name + " " + str(i + 1).zfill(
-                int(np.ceil(np.log10(n_cols))))) for i in range(n_cols)]
+        if feature_names_column:
+            names, attributes = [], []
+            for row in table:
+                name = get_unique_names(names, str(row[feature_names_column]))
```
janezd: This has quadratic complexity -- `get_unique_names` is linear. My suggestion to use `get_unique_names` was probably bad. It could be better to use a dict (or `Counter`) to count the occurrences of each name.

Orange/widgets/data/owtranspose.py
```diff
@@ -54,9 +57,7 @@ def __init__(self):
         edit.editingFinished.connect(self._apply_editing)
 
         self.meta_button = gui.appendRadioButton(box, "From meta attribute:")
-        self.feature_model = DomainModel(
-            order=DomainModel.METAS, valid_types=StringVariable,
-            alphabetical=True)
+        self.feature_model = DomainModel(DomainModel.METAS, alphabetical=True)
```
janezd: I don't think it needs to be a meta.

I am thinking about prohibiting discrete variables, though. They are always a bad choice, I suppose.

Orange/widgets/data/owtranspose.py
```diff
@@ -34,6 +34,9 @@ class Outputs:
     feature_names_column = ContextSetting(None)
     auto_apply = Setting(True)
 
+    class Warning(OWWidget.Warning):
+        duplicate_names = Msg("Duplicate feature names are indexed.")
```
janezd: I'm not sure that "to be indexed" is a proper phrase here. "Duplicate names have been given indices" could be better, but too long. What about "Values of '{}' are not unique.\nTo avoid multiple features with the same name, names were augmented with indices."? The second line only shows as tooltip.

Orange/data/table.py
```diff
@@ -1625,10 +1626,16 @@ def transpose(cls, table, feature_names_column="", meta_attr_name="Feature name"
...
+                names.append(name)
+                attributes.append(ContinuousVariable(name))
+        else:
+            attributes = [
+                ContinuousVariable(feature_name + " " + str(i + 1).zfill(
+                    int(np.ceil(np.log10(n_cols))))) for i in range(n_cols)]
```
janezd: I know you inherited this code, but it's, ehm, somewhat suboptimal. :) What about something like this?

```python
places = int(np.ceil(np.log10(n_cols)))
attributes = [ContinuousVariable(f"{feature_name} {i:0{places}}")
              for i in range(1, n_cols + 1)]
```

Orange/widgets/data/owtranspose.py
VesnaT: The second option is not much shorter once you add the name(s).

VesnaT: I prefer the first suggestion.

janezd: What about without the variable name? And then mentioning the variable name in the second line?

janezd:
Knuth would be so proud.

PR #3722, opened 2019-04-04
[FIX] stats: Fix statistics for primitive variables

### LINKED ISSUE #3671: Orange.statistics.util.py: improper computation of missing meta values
1. Metas are always read as dtype=object. Try to read them as float64 if possible (numeric, categorical, datetime).
2. If dtype=object (at least one string meta variable), compute better statistics for primitive variables.

Example no1:
Datasets (Forest Fires) - Data Table.
One meta column has TRUE/FALSE values, which are read as bool. I don't this FALSE should be treated as missing.
Also - strange things happen. If I remove a random attribute with Select Columns, not touching the metas at all, Data Table correctly reports there are no missing values in metas.

Example no2:
Datasets (Poker Hand) - Data Table.
The single meta column has values train/test, which are read as categorical. The value that gets assigned a 0 is treated as missing.

VesnaT:
##### Issue

Fixes #3671

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -361,7 +361,9 @@ def stats(X, weights=None, compute_variance=False):
             X.shape[0] - non_zero,
             non_zero))
     else:
-        nans = (~X.astype(bool)).sum(axis=0) if X.size else np.zeros(X.shape[1])
+        X_str = X.astype(str)
+        nans = ((X_str == "nan") | (X_str == "")).sum(axis=0) \
+            if X.size else np.zeros(X.shape[1])
         return np.column_stack((
             np.tile(np.inf, X.shape[1]),
             np.tile(-np.inf, X.shape[1]),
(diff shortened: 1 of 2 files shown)


Orange/statistics/util.py
```diff
@@ -361,7 +361,9 @@ def weighted_mean():
             X.shape[0] - non_zero,
             non_zero))
     else:
-        nans = (~X.astype(bool)).sum(axis=0) if X.size else np.zeros(X.shape[1])
+        X_str = X.astype(str)
+        nans = ((X_str == "nan") | (X_str == "")).sum(axis=0) \
```
ajdapretnar: I have my doubts about a literal 'nan' being treated as missing.

ajdapretnar:
@janezd @lanzagar @markotoplak How do you feel about a literal 'nan' being treated as missing? 'nan' can have meaning in English other than missing values (e.g. grandmother, naan bread).

ajdapretnar:
Ok, we have decided to accept the current solution. We consider 'nan' to be a borderline case and _if_ it ever arises, we will fix it then.

Issue #3371, opened 2018-11-08
Grouping feature values in Edit Domain


markotoplak:
A user had some data with 6 class values and wanted to group them into two (three by three). In the end, he used two consecutive Create Class widgets (the first one to add meaningful labels to existing class values and the second one to actually select according to the the added labels).

Perhaps there should be a way to join values in Edit Domain.

lanzagar:
I had a similar (but not the same) problem/task recently.
When a categorical variable has many values (e.g. 100) or some vary rare values, I would want to merge several values into one. Typically I want to keep some number of the most frequent values and merge the rest into something like "Other".
I currently have a script for this that can be used in Python Script, but I think something of this form would be useful to have available as a standard tool for preprocessing/transforming in Orange.
The question becomes how best to satisfy different use-cases since you probably want the ability to manually merge specific values, I want automatic merging based on frequency etc.
Maybe this is a single new widget, or we put the functionality in Edit Domain or Preprocess or both (manual in former, automatic in latter?).

PR #3725, opened 2019-04-05
[ENH] Edit Domain: Merge categorical values

ales-erjavec:
##### Issue

Implements gh-3371

Also fixes and assertion error when two values are renamed to the same value.
```
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/Orange/widgets/data/oweditdomain.py", line 1165, in commit
    var = apply_transform(v, tr)
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.6/lib/python3.6/functools.py", line 803, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/Orange/widgets/data/oweditdomain.py", line 1430, in apply_transform_discete
    dest_codes = positions(dest_values)
  File "/Applications/Orange3.app/Contents/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/Orange/widgets/data/oweditdomain.py", line 1427, in positions
    assert len(rval) == len(values)
AssertionError
```

##### Description of changes

Implement merging of categorical values.

Merging can be achieved by simply renaming individual values to the same name, or by selecting multiple values and pressing the merge button (shortcut <kbd>ctrl + =</kbd>) and entering/selecting a name in an editable combo box.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/oweditdomain.py
+++ b/Orange/widgets/data/oweditdomain.py
@@ -518,7 +518,7 @@ class CountedListModel(itemmodels.PyListModel):
             key = self.key(index)
             try:
                 counts[key] += 1
-            except TypeError:
+            except TypeError:  # pragma: no cover
                 warnings.warn(f"key value '{key}' is not hashable")
         self.__counts_cache = counts
         return self.__counts_cache
@@ -900,7 +900,7 @@ class DiscreteVariableEditor(VariableEditor):
         view = self.values_edit
         rows = view.selectionModel().selectedRows(0)
         if not rows:
-            return
+            return  # pragma: no cover
         for index in rows:
             model = index.model()
             state = index.data(EditStateRole)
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/oweditdomain.py
```diff
@@ -9,29 +9,35 @@
...
 from AnyQt.QtCore import pyqtSignal as Signal, pyqtSlot as Slot
-from AnyQt.QtCore import Qt, QEvent, QSize, QModelIndex
 
 import numpy as np
 
 import Orange.data
 
 from Orange.preprocess.transformation import Identity, Lookup
 from Orange.widgets import widget, gui, settings
+from Orange.widgets.data.owconcatenate import unique
```
janezd: I'd avoid importing functions from widget modules if possible because it messes up dependencies and complicates future refactorings. I think we'd better treat all functions in widget modules as private (even if not named as such), just as we treat widget classes as final.

Do we have a better place for functions like `unique`?

Perhaps better, isn't `unique(s)` same as `iter(dict.fromkeys(s))` (formally from Python 3.7, de facto already in 3.6). In below code, it's used as `list(unique(s))`, which is, I suppose, the same as `list(dict.fromkeys(s))`.

PR #3726, opened 2019-04-05
[ENH] OWScatterPlotBase: don't use special color for selection

robertcv:
##### Issue
When selecting some data in the scatter plot the starting color is orange. When the user selects an additional group the color of the previously selected points changes to blue and the newly selected points are red. When selecting the fourth group the same orange from the being is reused. 
I find the color change confusing and thought that it was a bug. Perhaps I am the only one who finds it strange.
##### Description of changes
Use the same color pallet from the beginning. The first selection becomes blue.
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -993,17 +993,11 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
             pen = [nopen] * self.n_shown
         else:
             sels = np.max(self.selection)
-            if sels == 1:
-                pen = np.where(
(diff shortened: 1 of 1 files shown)


janezd:
Groups should be marked with the same color succession so that coloring by group in further widgets matches the selection colors.

Yet, the first color (that is, the color if the selection is not "grouped") should preferably not be blue, because this is the color that always appears, and any selected blue points would appear as larger blue points.

After a lively debate we decided that the only change we'd consider making is to change the orange color to some other color that does not appear in the palette.

PR #3245, opened 2018-09-12
Updated documentation for Data Sampler

ajdapretnar:
##### Issue

Implements #3198. 

##### Description of changes
Updated Data Sampler documentation to include an example with oversampling. Some minor changes added.

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation

PR #3727, opened 2019-04-05
Data sampler: Allow selecting 100% of data, shuffle it

### LINKED ISSUE #2035: Replicable sampling in Test & Score

janezd:
##### Issue

Closes #2035. 

##### Description of changes

#2035 required:

1. Modified DataSampler that allows selecting all instances (not just n-1)
2. DataSampler doc mentioning that the widget shuffles the dataset
3. A test that checks that DataSampler really shuffles the dataset
4. An example in Test & Score docs showcasing how this can be done (like #3245)

I've done them all, but I didn't have a good use case in 4. @ajdapretnar, in an already agitated state because of certain issues around #3722, told me in no uncertain terms to remove it. I'm married for 17 years, so I know better than to complain. On the contrary, I *must say* that I didn't like that text either, so I was very happy to remove it.

Documentation still explains that choosing the entire data set shuffles it, which, I must say, should suffice.

(Joke aside, it indeed does.)

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/widgets/data/owdatasampler.py
+++ b/Orange/widgets/data/owdatasampler.py
@@ -60,7 +60,7 @@ class OWDataSampler(OWWidget):
 
     class Error(OWWidget.Error):
         too_many_folds = Msg("Number of folds exceeds data size")
-        sample_larger_than_data = Msg("Sample must be smaller than data")
+        sample_larger_than_data = Msg("Sample can't be larger than data")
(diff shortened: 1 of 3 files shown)


astaric:
Removed requirement #4, as I do not remember anymore what the intended use case was.

PR #3731, opened 2019-04-07
Extend list_tables_query to show views

oyvinrog:
Views should be supported for analysis in Orange Canvas

##### Issue

SQL Table widget in Orange Canvas only displays tables. However, most database servers have  very useful views.

You could argue that why not just use "custom SQL", but the "custom SQL" part of the widget can become too complicated for large SQLs, and it does not support CTEs.

##### Description of changes
Extended select query from information_schema.tables to include views in addition to table

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/sql/backend/mssql.py
+++ b/Orange/data/sql/backend/mssql.py
@@ -29,8 +29,8 @@ class PymssqlBackend(Backend):
             raise BackendError("Incorrect format of connection details")
 
     def list_tables_query(self, schema=None):
-        return """        
-   SELECT [TABLE_SCHEMA], [TABLE_NAME]
+        return """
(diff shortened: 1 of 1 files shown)


lanzagar:
Thank you for the contribution.

Some background for anybody interested:
Views are definitely useful and I believe we used to support them, but they were removed because of problems with efficient (block-level) sampling on them. But that was mostly related to the postgresql backend and (very very) big data. We have since made the change which forces the download option on non-postgres databases and allows us to treat views the same as tables here.
In light of this PR I think we should allow views for the postgresql backend as well and just force the download in that case too, to avoid the sampling issues.

PR #3751, opened 2019-04-12
OWScatterplotGraph: Emit signals when resizing data points

janezd:
##### Issue

In Network Explorer, I wanted to draw lines to the edge of the circles, not to the center. I tried to override `update_sizes` to call `self.update_edges()` after `super().update_sizes()`... but this didn't work because `update_size` does not change sizes immediately but updates them from a separate thread.

Furthermore, it would be nice if edges were animated together with circles.

##### Description of changes

I added signals that are emitted at the start and end of resize, and at each step.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -340,6 +340,9 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
     (in methods `get_<something>`).
     """
     too_many_labels = Signal(bool)
+    begin_resizing = Signal()
+    step_resizing = Signal()
+    end_resizing = Signal()
 
     label_only_selected = Setting(False)
     point_width = Setting(10)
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -766,17 +769,24 @@ def __call__(self):
...
             if self.n_valid <= MAX_N_VALID_SIZE_ANIMATE and \
                     np.all(current_size_data > 0) and np.any(diff != 0):
                 # If encountered any strange behaviour when updating sizes,
                 # implement it with threads
+                self.begin_resizing.emit()
                 self.timer = QTimer(self.scatterplot_item, interval=50)
                 self.timer.timeout.connect(Timeout())
                 self.timer.start()
             else:
+                self.begin_resizing.emit()
```
janezd: I know that this line repeats and could be before `if`, but I prefer having pairs begin-end.

Issue #1848, opened 2016-12-30
Cumulative distribution function widget


brylie:
# User story
> As a data analyst
I would like to view the cumulative distribution function for observations in my data set
so that I can intuitively grasp the probability that an observation will fall above or below a given threshold

# Feature
A *cumulative distribution function* widget would calculate the CDF of a variable among a set of observations. The widget would output an array of x, y coordinates that could be visualized as line charts:

![cdf for several normal distributions](https://upload.wikimedia.org/wikipedia/commons/thumb/c/ca/Normal_Distribution_CDF.svg/500px-Normal_Distribution_CDF.svg.png)

The widget may wrap, or derive from, an existing CDF library, such as [the Cdf class created by Allen B. Downey](http://greenteapress.com/thinkstats/Cdf.html)

## Meta
This widget might appear in an *Analyze* section of the Orange toolbox, among a few other widgets useful for the **analysis stage of a machine learning workflow**.

# Related resources
- Wikipedia: [Cumulative distribution function](https://en.wikipedia.org/wiki/Cumulative_distribution_function)
- Allen B. Downey: Think Stats [Cumulative distribution function sourcecode](http://greenteapress.com/thinkstats/Cdf.html)

kernc:
> This widget might appear in an Analyze section of the Orange toolbox, among a few other widgets useful for the analysis stage of a machine learning workflow.

Why not as a check box in the _Distributions_ widget?

janezd:
Decision: we'll add a check box to Distributions widget.

janezd:
Closed via #3766.

PR #3766, opened 2019-04-19
[ENH] OWDistributions: add cumulative distribution

rokgomiscek:
##### Issue

Implements #1848.

##### Description of changes
Added check box to display cumulative ditribution. 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Nice.

It is a bit weird that curves end at the maximal value, but on the other hand it gives additional information while improving ink/info ratio.

PR #3768, opened 2019-04-23
[FIX] OWLinePlot: legible bottom axis labels

### LINKED ISSUE #3741: OWLinePlot: legible axis labels
Follow up from #2110.

Current status of Line Plot widget. Labels overlap on x and the final attribute name, if long, falls out of view.
Screen Shot 2019-04-12 at 11 06 58

Proposed solution: elide labels and enforce minimum widget width. If there are hundreds of attributes (gene expressions), show every n-th.

VesnaT:
##### Issue

Fixes #3741

##### Description of changes
Subclass `pg.AxisItem` and override `tickStrings()` to display feature's names.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owlineplot.py
+++ b/Orange/widgets/visualize/owlineplot.py
@@ -91,6 +91,25 @@ class LinePlotStyle:
     MEAN_DARK_FACTOR = 110
 
 
+class LinePlotAxisItem(pg.AxisItem):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._ticks = None
+
+    def set_ticks(self, ticks):
+        self._ticks = dict(enumerate(ticks, 1)) if ticks else None
+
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I think this is really nice. ⭐️

lanzagar:
Just writing down a couple of observations I had:

1. Right part of right-most label is cut off (e.g. heart-disease)
2. Actual width of labels not taken into account? Sometimes too much pruning even when everything would fit (e.g. housing)
3. I don't like completely omitting labels unless really necessary, even just a single letter and a dot could be enough to indicate something should have been there. And flexible ellipsing depending on available space would be ideal.
4. When some things really do need to be omitted, it would be nice to have a tooltip or similar to be able to check one specific unlabelled variable. Currently this can be done with zooming, though a bit awkwardly.

I understand that some of these issues aren't simple to solve and don't think of them as blockers that have to be addressed.

PR #3770, opened 2019-04-24
Tests: Remove also saved metadata

ajdapretnar:
##### Issue

Tests fail to remove metadata files.

##### Description of changes
Remove also metadata files, along with saved files.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
As there are often several tests in a single function adding another `try`/`finally` layer for temp dir management would make them even more complicated so the straightforward approach is fine for now.
However, if someone wants to spend some time on improving these tests, that would be very welcome. The fact remains that creating files in a temporary dir would be more appropriate than doing that in the current location from where the tests are being run.

PR #3774, opened 2019-04-26
[DOC] Replace numpydoc with sphinx.ext.napoleon

ales-erjavec:
##### Issue

Reference documentation fails to build with sphinx==2.0.1 and numpydoc==0.9.1. (raises `numpydoc.docscrape.ParseError: ...`)

##### Description of changes

Use sphinx.ext.napoleon instead of numpydoc

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
I checked the differences between docs built with numpydoc 0.8 and napoleon. Most of the time we get something nearly equivalent, but there are things that napoleon actually handled better. I did not find any cases where numpydoc was better, but I was not exhaustive.

PR #3775, opened 2019-04-26
[FIX] OWPythagorasTree: Enable node selection from forests with categorical variables.

### LINKED ISSUE #3414: Pythagorean Tree crashes when selecting data with categorical variables
##### Orange version
master

##### Expected behavior
Selecting a node in tree should always send data to output.

##### Actual behavior
For a specific case, this fails.

image

In File, select a data set containing categorical variables (e.g. titanic, breast-cancer), select a tree in the random forest visualization and then try to select a node in the Pythagorean forest widget.

The widget crashes with an error like this:
```
  File "/home/pavlin/dev/orange3/Orange/widgets/visualize/owpythagorastree.py", line 309, in commit
    data = self.tree_adapter.get_instances_in_nodes(nodes)
  File "/home/pavlin/dev/orange3/Orange/widgets/visualize/utils/tree/skltreeadapter.py", line 260, in get_instances_in_nodes
    leaf_samples = self.get_samples_in_leaves()
  File "/home/pavlin/dev/orange3/Orange/widgets/visualize/utils/tree/skltreeadapter.py", line 243, in get_samples_in_leaves
    leaf_indices = assign(0, items)
  File "/home/pavlin/dev/orange3/Orange/widgets/visualize/utils/tree/skltreeadapter.py", line 229, in assign
    column = self.instances.X[indices, feature_idx]
IndexError: index 10 is out of bounds for axis 1 with size 9
```

thocevar:
##### Issue
Fixes #3414

##### Description of changes
Transform data to the discretized domain when computing the samples in leaves (SklTreeAdapter).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/utils/tree/skltreeadapter.py
+++ b/Orange/widgets/visualize/utils/tree/skltreeadapter.py
@@ -151,8 +151,9 @@ class SklTreeAdapter(BaseTreeAdapter):
 
     def attribute(self, node):
         feature_idx = self.splitting_attribute(node)
-        if feature_idx != self.FEATURE_UNDEFINED:
-            return self.domain.attributes[self.splitting_attribute(node)]
+        if feature_idx == self.FEATURE_UNDEFINED:
+            return None
+        return self.domain.attributes[self.splitting_attribute(node)]
 
     def splitting_attribute(self, node):
         return self._tree.feature[node]
--- a/Orange/widgets/visualize/utils/tree/skltreeadapter.py
+++ b/Orange/widgets/visualize/utils/tree/skltreeadapter.py
@@ -236,7 +236,7 @@ class SklTreeAdapter(BaseTreeAdapter):
                 feature_idx = self._tree.feature[node_id]
                 thresh = self._tree.threshold[node_id]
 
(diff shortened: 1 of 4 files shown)


Orange/widgets/visualize/utils/tree/skltreeadapter.py
```diff
@@ -236,7 +236,7 @@ def assign(node_id, indices):
                 feature_idx = self._tree.feature[node_id]
                 thresh = self._tree.threshold[node_id]
 
-                column = self.instances.X[indices, feature_idx]
+                column = self.instances.transform(self.domain).X[indices, feature_idx]
```
pavlin-policar: Running transform in the `assign` function could be problematic since it's called recursively. If the tree is large, then this will be called many times. It would be better to cache the thing outside `assign`.

Also, could you add a test for this?

pavlin-policar:
I hate to be annoying, but this test really belongs to the tree widget tests, not the forest widget tests. If the behaviour is in the tree widget, the tests should be added to the tree widget tests. Otherwise, the tests for the forest widget are also testing the tree widget, which would quickly become messy.

PR #3777, opened 2019-04-27
[FIX] Minor improvements to pythagorean trees

pavlin-policar:
##### Description of changes
This fixes a couple of minor things I came across while reviewing #3775.

1. Remember selection in the Pythagorean forest widget so that reopening a workflow won't clear selection.
2. Remove some dead code from both widgets
3. Properly use context settings in Pythagorean tree and forest widget (before, it wasn't using the domain context settings handler.
4. The tree widget would crash if there was no data and we clicked "Redraw".

The tree widget does not currently remember selection (both when reopening a workflow or when redrawing), and this will be a bit trickier to implement, since Orange trees have no identifier for each node.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owpythagorastree.py
+++ b/Orange/widgets/visualize/owpythagorastree.py
@@ -75,8 +75,7 @@ class OWPythagorasTree(OWWidget):
         super().__init__()
         # Instance variables
         self.model = None
-        self.instances = None
-        self.clf_dataset = None
+        self.data = None
         # The tree adapter instance which is passed from the outside
         self.tree_adapter = None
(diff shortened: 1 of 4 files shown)


thocevar:
There is a problem with the settings of Pythagorean trees that come from a forest. For example, you can set the depth of a tree to 1, then select a different tree and the visualization does not match the depth setting. Similar situation arises when you change the data. The depth is set to the context-aware value but is not reflected in the visualization.

pavlin-policar:
@thocevar This issue should now be resolved.

PR #3788, opened 2019-05-09
[ENH] Error Animations

### LINKED ISSUE #3377: Widget errors are hard to notice

irgolic:
##### Issue
Resolves #3377, #3699 

##### Description of changes
Using QPropertyAnimation and an opacity property, makes warnings and errors flash on update.

In nodeitem.py, GraphicsIconItem now subclasses QGraphicsWidget instead of QGraphicsItem (to allow QPropertyAnimation). I've not tested this change extensively, and am not aware of its implications.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/canvas/items/nodeitem.py
+++ b/Orange/canvas/canvas/items/nodeitem.py
@@ -17,8 +17,8 @@ from AnyQt.QtGui import (
     QPainterPathStroker
 )
 from AnyQt.QtCore import (
-    Qt, QEvent, QPointF, QRectF, QRect, QSize, QTime, QTimer, QPropertyAnimation
-)
+    Qt, QEvent, QPointF, QRectF, QRect, QSize, QTime, QTimer, QPropertyAnimation,
+    QEasingCurve)
 from AnyQt.QtCore import pyqtSignal as Signal, pyqtProperty as Property
 
 from .graphicspathobject import GraphicsPathObject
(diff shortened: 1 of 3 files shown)


irgolic:
Some widgets call update when clicking any button/slider in the widget window, but some (due to the warning) completely suppress it. 

An example of this is **File (titanic) -> Scatterplot**, where Scatterplot displays a warning but does not animate it upon interacting with the window.

Is there a way to enable such behavior for all widgets, or would each individual widget behaving this way have to be changed?

janezd:
Funny, I just started typing a similar comment about inconsistent behaviour of widgets.

This is not easy to fix. The logic of clearing and showing error messages is rather complex. Some widgets clear errors on receiving new data and then re-show them in set_data or related functions. Others do it when udpating the scene or the output. And some do it in callbacks, thus anywhere. This will never be uniform.

janezd:
We discussed this at the meeting: we like flashing, but perhaps five times instead of just two.

It is not nice that errors are not re-animated or that they will sometimes be reanimated without obvious reason, but we can live with it.

irgolic:
I think flashing 5 times might be a bit excessive. The errors' visibility is increased with flashing, but it helps less with warnings due to their bright color.

Also, where are general informative messages used (not warnings/errors), should those flash too?

janezd:
Info: I guess not.

I'll check the new commit tomorrow.

janezd:
I will merge the PR so everybody will the change and comment. :) We can easily decrease the number of flashes (yes, it is a bit too much :)) later.

PR #3789, opened 2019-05-09
OWNomogram: Add feature list as output

### LINKED ISSUE #3783: Nomograms should output data with selected (shown) features
We had data with a large number of attributes. We'd like to select the first 10 features based on their absolute importance and see if the model is still as good as before. I guess the only way to do it currently is to manually select those features in select columns ... which is annoying.

janezd:
Fixes #3783.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
@ajdapretnar, could you kindly add the documentation for this? The intended use is

Screenshot 2019-05-09 at 22 04 21

PR #3795, opened 2019-05-10
[ENH] Reset settings button

### LINKED ISSUE #3564: Every widget should have a Reset Settings button
... in the status bar.

janezd:
Fixes #3564.

It would be nice if every widget had a reset settings button. This must be implemented specifically for each widget.

This PR prepares the GUI and implements the functionality for a single widget. If widget implements a method named `reset_setting`, a reset button will appear and call this method.

`reset_settings` will usually be simple: it must
- copy the default settings (e.g. by calling the provided `self._reset_settings`),
- resolving context related settings if needed (e.g. default for x and y in the scatter plot),
- calling the necessary callbacks.

The icons is passable but could be improved upon. For one thing, it is a bit too large.
 
##### Includes
- [X] Code changes


markotoplak:
This restores last saved widget settings well. It should probably reset the settings to the defaults.

markotoplak:
@janezd, I added some tests. While I was writing them, pycharm suggested that I use resetSettings(), which seems to do something very similar to _reset_settings. Which one should we keep?

janezd:
- The widget gets a reset button if it implements `reset_settings`.
- `_reset_settings` is a helper method (of `OWComponent`) that can be called by `reset_settings`.
- `resetSettings` is a method that is apparently used only in tests for `OWFeatureStatistics`. As I see, it does not restore to original settings but probably to the last saved settings.

I would remove `resetSettings`.

PR #3803, opened 2019-05-16
[FIX] VerticalItemDelegate: Do not cut long vertical labels

VesnaT:
##### Issue

Long vertical labels (i.e. labels for table's vertical header) are cut off: 
Screen Shot 2019-05-15 at 11 17 06

##### Description of changes
Show the whole label.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -3084,6 +3084,10 @@ class VerticalItemDelegate(QStyledItemDelegate):
     # Extra text top/bottom margin.
     Margin = 6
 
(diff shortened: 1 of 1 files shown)


janezd:
This seems OK for this particular use case because there is nothing in the cell above and below this one. What if they are not empty? Maybe `VerticalItemDelegate` should have a flag whether to extend the text outside the cell or not?

VesnaT:
Done.

PR #3808, opened 2019-05-20
Stop suppressing ".*the matrix subclass.* warnings

markotoplak:
Since Scipy==1.3.0, which was released on 17 May 2019, these warnings stopped appearing.


markotoplak:
@janezd, are you fine with this? You will see warnings until you update to 1.3.0, of course.

janezd:
I am very fine with it.

PR #3809, opened 2019-05-20
[FIX] OWRadviz: legible axis labels

### LINKED ISSUE #3744: OWRadViz: legible axis labels
Follow up on #2110.

Current widget status. Similar to #3743. Limit the number of features displayed to 20. If features names are long, elide them after 20 chars.
Screen Shot 2019-04-12 at 11 26 33

robertcv:
##### Issue
Fixes #3744
##### Description of changes
* Rotate variable labels according to the angel of position
* Shorten labels to 20 characters
* Show tooltip with the full label name 
* Limit number of selected variables to 20

radviz

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/plot/owplotgui.py
+++ b/Orange/widgets/utils/plot/owplotgui.py
@@ -58,12 +58,19 @@ class VariableSelectionModel(VariableListModel):
     SortRole = next(OrangeUserRole)
     selection_changed = pyqtSignal()
 
-    def __init__(self, selected_vars):
+    def __init__(self, selected_vars, max_vars=None):
         super().__init__(enable_dnd=True)
         self.selected_vars = selected_vars
+        self.max_vars = max_vars
 
     def is_selected(self, index):
         return self[index.row()] in self.selected_vars
+
+    def is_full(self):
+        if self.max_vars is None:
+            return False
+        else:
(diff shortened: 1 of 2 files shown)


BlazZupan:
In case of longer, multi-word labels, there is sufficient place to split the labels to, say, two lines, and end the label with ellipsis where there is an end of the word.

For example, we could split "Gross national income (GNI) per capita" into

Gross national
income (GNI) ...

or "Estimated gross national income per capita" into

Estimated
gross national ...

In this way the width of the labels can get short her. Currently, and in the example above, labels take much space making the graph smaller.

Display of full text of the label when mouse-over is great and it should of course stay.

robertcv:
I add the split. Is it ok?
Screenshot from 2019-05-31 13-00-15

BlazZupan:
This looks way better. Another tiny request for change, though, before closing this PR: if the text appears on the left side of the circle than it should be right-aligned. This was irrelevant for the one-liners before. For instance, in the example above, "Gender development" on the bottom left should be right aligned. If this change is not difficult to make, please revise the PR.

robertcv:
Screenshot from 2019-06-07 13-59-19

PR #3812, opened 2019-05-21
Remove two print statements from test_io

markotoplak:
##### Issue
Tests on master have two additional prints. I did not check that code thoroughly before merging it yesterday...


PrimozGodec:
@markotoplak sorry, I totally forgot about those two prints.

Issue #3805, opened 2019-05-17
Different interfaces for clustering methods


PrimozGodec:
##### Orange version

Any

##### Expected behavior

Clustering algorithms should have the same interface, such that other scripts can use them the same way or even replace the clustering method.

##### Actual behavior

There are different interfaces for clustering algorithms. 

1. `KMeans` has `fit` function that accepts `X` and `Y` matrices and returns the instance of `KMeansModel` class as the result. `KMeansModel` has a `__call__` function that accepts either a `Table`, `Instance`, or `np.ndarray`.

2. `DBSCAN` has `fit` function that accepts `X` and `Y` matrices or `Table` and returns the instance of `DBSCANModel` class as the result. `DBSCANModel` has a `__call__` function that accepts either a `Table`, `Instance` (even it is not supported), or `np.ndarray`.

3. `Louvain` has `fit` function that accepts `X` and `Y` matrices or `Table` and does not return anything. It has also a `__call__` function that returns clusters directly.

The `__call__` function of `KMeasnModel` and `DBSCANModel` also does not work when getting the `Table` at the input since there is a bug - checking if `Domain` matches with `self.pre_domain` which do not exist.

`DBSCANModel` call `fit_predict` instead of predict. It is probably made since `sklearn.cluster.DBSCAN` does not have `predict` method implemented, but what is then the purpose of having `Projector` and `Projection` a separate class.

pavlin-policar:
Following the discussion with @PrimozGodec and @markotoplak I propose the following interface for clustering methods:

```python
class Clustering:
    preprocessors = []
    
    def __init__(self):
        self.labels = None
        
    def preprocess(self, data):
        for pp in self.preprocessors:
            data = pp(data)
        return data

    def __call__(self, data: Table) -> np.ndarray:
        data = self.preprocess(data)
        return self.fit_predict(data.X, data.Y)

    def fit(self, X: np.ndarray, y: np.ndarray = None) -> None:
        """Run clustering on the data."""
        raise NotImplementedError()

    def fit_predict(self, X: np.ndarray, y: np.ndarray = None) -> np.ndarray:
        """Run clustering on the data and return the assigned labels."""
        self.fit(X)
        return self.labels
```

It's stateful like scikit-learn models i.e. attributes are set in `fit`. That's not ideal, but if we really want to compute the silhouette score during clustering (like KMeans currently does), that's pretty much the only option.

[...]

PR #3814, opened 2019-05-22
[ENH] Unified clustering API

PrimozGodec:
##### Issue

Fixes https://github.com/biolab/orange3/issues/3805

##### Description of changes
Clustering algorithms are simplified and unified. Each clustering algorithm (kMeans, DBSCAN, Lovain) inherit from a common `Clustering` class. Clustering itself is stateless (it does not remember the data or parameters (such as centroids)). After it is called it returns clusters or when `fit_storage` is called it creates a new instance of type `ClusteringModel` which holds the model parameters (in case of k-means they are clusters) - model can be created just for k-Means. Other algorithms do not have parameters to store or do not enable predicting.

The interface to call the clustering is now different. Here is the example for kMeans:

    d = Table("iris")
    km = KMeans(preprocessors=None, n_clusters=3)  #init the clustering
    clusters = km(d)  # compute clusters

or in the case when we need model:
    
    d = Table("iris")
    km = KMeans(preprocessors=None, n_clusters=3)  #init the clustering
    model = km.fit_storage(d)  # fit the model

I also removed the silhouette computing from k-means class since I think it is not the part of the clustering but its own unit. From now on it will be called directly from the widget.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


PrimozGodec:
@janezd and @markotoplak are you ok with the proposed scheme? I will push the changes for tests and widgets after we agree with the clustering scheme.

pavlin-policar:
Conceptually, some of this doesn't make sense. This is what I would expect to happen when using clustering classes: I call DBScan with some data and I get a model. I would assume the cluster labels are available at this point (via some `cluster_assignments` field or something). Now I get some new data. I notice the dbscan model has a transform method, so why not use this to check which clusters these new samples would belong to. Ok, I call predict, and now am baffled why there may or may not be more clusters in the new data than in the original data. If I overlay the original data with the new data, I notice that the clusters don't overlap at all. What gives!? I look into the code and lo and behold, transform doesn't actually do what I think it does (or what I would expect it to do from having used scikit-learn) but actually computes a new clustering on this data. Why? After another 5min of searching around the internet, I find that dbscan does not support `transform`.

In k-means, this makes complete sense, but dbscan and Louvain don't have a transform method, nor is it straightforward to add one.

I propose that the cluster assignments of the training data be available on the model as a field (so I can get cluster assignments by just calling fit) and if the method supports `transform`, then this can be implemented there. If it doesn't support transform, it should fail loudly, via a NotImplementedError.

PrimozGodec:
@pavlin-policar I agree with your solution. I also do not like that we call clustering twice for the same results as well that we pretend that there is a transform function. @janezd what you think about Pavlins idea.

PrimozGodec:
Tests are failing since the widget for Lovain clustering is not fixed to new clusterings yet. I would like to hear the opinion about @pavlin-policar idea anyway.

lanzagar:
Main points of some additional discussion:
- removing the `compute_silhouette_score` parameter from kMeans should be done by deprecating it first and not crashing on use of previous API
- when wrapping sklearn methods in Orange, we copy the parameters (I guess mostly for introspection, code completion, etc). This practice can be discussed, but I would prefer to keep it like it was in this PR and do that separately if someone wants to propose a general change (not just for clustering methods)
- because `ClusteringModel` is not really something that is often used/useful, the `Clustering.__call__` method could default to the fit_predict behaviour and just return the clusters. For methods that support it (kmeans), if someone wants the model they can get it by specifically calling `fit`/`fit_storage` instead of `__call__`

PrimozGodec:
@lanzagar can you check whether the clustering implementation is what we want. I meanwhile I will fix lint and also add some more tests for clustering.

Orange/clustering/clustering.py
```diff
@@ -0,0 +1,86 @@
+import numpy as np
+
+from Orange.data import Table, Instance
+from Orange.preprocess import Continuize, SklImpute
+
+
+def preprocess_data(preprocessors, data):
```
lanzagar: I do not think this is the right place for this function.

Orange/clustering/clustering.py
```diff
@@ -0,0 +1,86 @@
...
+    def __call__(self, data):
+        return self._fit_predict(data)
+
+    def fit_storage(self, data):
+        raise NotImplementedError
+
+    def fit(self, X: np.ndarray, y: np.ndarray = None):
+        return self.__wraps__(**self.params).fit(X, y)
+
+    def _fit_predict(self, data, return_model=False):
```
lanzagar: Didn't we discuss that there should be a separate function to get a model, instead of a function like this, that can return different types of results based on a parameter.
Also I think that function should be public (not start with an underscore).

Orange/clustering/clustering.py
```diff
@@ -0,0 +1,86 @@
...
+    preprocessors = [Continuize(), SklImpute()]
+    model_class = None
+
+    def __init__(self, preprocessors, parameters):
+        self.preprocessors = tuple(preprocessors or self.preprocessors)
+        self.params = {k: v for k, v in parameters.items()
+                       if k not in ["self", "preprocessors", "__class__"]}
+
+    def __call__(self, data):
+        return self._fit_predict(data)
```
lanzagar: Why is `__call__` a single function call? Couldn't that function body be here instead?

Orange/clustering/clustering.py
```diff
@@ -0,0 +1,86 @@
...
+            raise Exception("The type of your data is not supported.")
+
+    def predict(self, X):
+        raise NotImplementedError(
+            "This clustering algorithm does not support predicting.")
+
+
+class Clustering:
+    """
+    ${skldoc}
```
lanzagar: This remains in the docstrings of subclasses. Check e.g. `help(Orange.clustering.DBSCAN)`

pylintrc
```diff
@@ -61,7 +61,7 @@ confidence=
 # --disable=W"
 disable=
     missing-docstring,  # I guess not
-    no-name-in-module, no-member  # too many false positives from Qt
+    no-name-in-module, no-member,  # too many false positives from Qt
```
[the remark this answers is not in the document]
PrimozGodec: This is not in my commit - this a commit which accidentally became part of this PR. Will rebase again.

PrimozGodec:
@lanzagar I think it is done now. Can you check if there is anything to correct?

Orange/tests/test_clustering_kmeans.py
```diff
@@ -2,64 +2,134 @@
 # pylint: disable=missing-docstring
 
 import unittest
+import warnings
 
 import numpy as np
-from scipy.sparse import csc_matrix
+from scipy.sparse import csc_matrix, csr_matrix
```
lanzagar: I find it a bit strange, that we already tested learning on csc, but you added csr for testing the models. Even worse is that giving a model a csc matrix would result in a crash (although this is avoided in tests)...
I know this is baggage inherited from the copy paste of call from learning models, but I think it needs fixing (probably in a separate PR).

PR #3828, opened 2019-05-29
OWLinePlot: controlArea too wide

ajdapretnar:
##### Issue

controlArea of LinePlot is unnecessarily wide.

##### Description of changes
Make controlArea narrower.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
This doesn't seem like a nice fix. If there's a better way, suggestions welcome.

markotoplak:
@ajdapretnar, I rather set sizeHint instead of widget, because setting the widget width disables size hints. Could you check if the effect is what you wanted?

PR #3838, opened 2019-05-31
Update readme

markotoplak:
"Installing with pip" section now installs released version, as the conda section does. I also added pyqtwebengine to main instructions. Then, I removed some obsolete information about Windows.


lanzagar:
What is the problem with codecov :(
(it is reporting decreased coverage in `owlouvainclustering.py` while changes only in `README.md`)
This is getting annoying...

PR #3841, opened 2019-05-31
OWMergeData: Warn if any variables are non-unique and hidden

### LINKED ISSUE #2810: Merge Data: can't select what to match by
##### Orange version
3.8.dev

##### Expected behavior
If two datasets have columns that can be matched, one can set to match them by these two columns.

##### Actual behavior
When one sets the first variable to match by, often the second option remains only to merge by Position.
For example, I can't match by Datum, even though the column is present in the second data and does have corresponding column to match by in the original data.

File (iris) + File (1) (iris) - Merge Data. Try to merge by sepal length and sepal length. Not possible in current version.

I am working with datetime features, but I think this applies to other uses as well. If the second data set can be matched with only certain rows (not all datetime values are present in the second data set), then Orange should match the corresponding datatime and leave the unmatching rows empty (i.e. fill with nans).

janezd:
##### Issue

Fixes #2810.

##### Includes
- [X] Code changes
- [x] Tests

### DIFF
--- a/Orange/widgets/data/owmergedata.py
+++ b/Orange/widgets/data/owmergedata.py
@@ -52,6 +52,12 @@ class OWMergeData(widget.OWWidget):
 
     class Warning(widget.OWWidget.Warning):
         duplicate_names = widget.Msg("Duplicate variable names in output.")
+        non_unique_variables = widget.Msg(
+            "Columns with non-unique values can't be used for matching.\n"
+            "The second variable must uniquely define the row in the second "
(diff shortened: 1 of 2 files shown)


ajdapretnar:
Do you think the warning should be present even when the user selects two unique variables?

janezd:
Yes, definitely. Perhaps the user has chosen something and would like to change it, but can't see it.

The warning is about the available selection, not about what is selected.

PR #3844, opened 2019-06-02
[FIX] Update output on new input, even if auto commit is disabled

### LINKED ISSUE #3761: Widgets don't Apply, Send or Run when their input signals changed
##### Orange version

3.21-dev

##### Expected behavior

According to #3455 and #3755, widgets should Apply, Send or Run when their input signals changed even if auto-apply/send/run is disabled.

##### Actual behavior

Widgets are not consistent. Some widgets do not Apply, Send or Run when the input signal change.
Those widgets are:

- [x] k-Means
- [x] Merge Data
- [x] Concatenate
- [x] Transpose
- [x] Randomize
- [x] Color
- [x] Neighbours
- [x] Box Plot
- [x] Distance Matrix
- [x] Manifold Learning

The same error also happens for widgets reported in following issue https://github.com/biolab/orange3/issues/3763 in combination with data selection issue.

Connect the widget's input to the data source (e.g. File Widget) and connect the output to visualization widget (e.g. Data Table). Uncheck apply automatically checkbox and change the input data. Observe if output data changes (widget do something with the input).

janezd:
##### Issue

Fixes #3761.

##### Description of changes

Call `unconditional_apply` instead of `apply`; remove the pointless auto commit checkbox in box plot.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owcolor.py
+++ b/Orange/widgets/data/owcolor.py
@@ -189,7 +189,6 @@ class ContColorTableModel(ColorTableModel):
         self.dataChanged.emit(self.index(0, 1), self.index(self.n_rows(), 1))
 
 
-
 class ColorTable(QTableView):
     """The base table view for discrete and continuous attributes."""
 
@@ -369,7 +368,7 @@ class OWColor(widget.OWWidget):
             self.cont_model.set_data(self.cont_colors)
             self.disc_view.resizeColumnsToContents()
             self.cont_view.resizeColumnsToContents()
-        self.commit()
+        self.unconditional_commit()
 
     def storeSpecificSettings(self):
         # Store the colors that were changed -- but not others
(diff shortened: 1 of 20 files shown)


janezd:
I yesterday assigned this to @ajdapretnar; for this one, maybe @VesnaT would have to check it, too.

ajdapretnar:
I think considering this PR is dealing with apply, it would be nice if it fixed the mismatching buttons in Color and Concatenate, which are unlike those of other widgets.

janezd:
This is one of supported generic options. It is more obvious than the usual (checkbox without a label) but we seldom use it because most widgets are not wide enough. I can of course remove it (the label is set by an explicit argument to `gui.auto_commit`, so I just have to remove the extra argument), but I'm undecided about whether the extra explanatory label indeed bothers me. @markotoplak, @irgolic, what are your opinions?

PrimozGodec:
All widgets in this PR works correctly now, so I will merge it. With this PR debate about the inconsistency of auto-commit button started. I will open the separate issue to continue it.

PR #3846, opened 2019-06-02
Discretization: Proper labels for discretized TimeVariable

janezd:
##### Issue

Discretized `TimeVariable` has float-like labels instead of dates/times.

##### Changes

`Discretizer.create_discretized_var` now calls `str_val` to create labels for discretized variable.

This fixes the problem with `TimeVariable`.

Also, for numeric variables, it will properly format small numbers; current code circumvented the fix from 0a205f87cbe3faf68effb9e95eb8f5580f4711ed, which introduced formatting with `%g` when the number of decimals is too large. For normal numbers, the change should have not effect (although rounding is now replaced by stripping 0s and the dot).

##### Includes
- [X] Code changes
- [X] Updated tests

### DIFF
--- a/Orange/preprocess/discretize.py
+++ b/Orange/preprocess/discretize.py
@@ -1,3 +1,5 @@
+import re
+
 import numpy as np
 import scipy.sparse as sp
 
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I like this change a lot. It fixes lots of problems.
There's one issue left... when datetime is not discretized, it still shows as a float. I am referring to Distributions (without binning) and Scatter Plot. This PR is about discretization, so I am merging it, but should we not fix those instances as well?

PR #3849, opened 2019-06-03
Sieve, Mosaic, CanvasText: Elide long labels

### LINKED ISSUE #3742: OWSieveDiagram: legible axis labels
Follow up on #2110.

Current widget status.
Screen Shot 2019-04-12 at 11 13 15

Solve overlap by eliding over the width of the column. If column too small, hide the label.
Also, display proper datetime, not floats.

janezd:
##### Issue

Fixes #3742 (eliding of long labels for Sieve)

Moves code from #3845 (eliding for Mosaic) to `CanvasText`, so it can be used in other widgets (like Sieve), too.

##### Includes
- [X] Code changes


ajdapretnar:
This works well for the two widgets. There are some other places where labels could (should?) be elided, but I will open a separate issue for that.

PR #3853, opened 2019-06-05
OWPredictions: Pylint

janezd:
Trivial pylint fixes.


janezd:
Changes from this commit are now included in #3856.

PR #3854, opened 2019-06-06
OWHierarchicalClustering: Add labels to dendrogram

### LINKED ISSUE #1088: Hierarchical Clustering widget: show group ID on cluster selection
Dendrogram in Hierarchical Clustering widget should display cluster IDs when clusters are selected. Think about where to put the label: onto the cluster, to the left, or to the right of the cluster. TBD.

janezd:
##### Issue

Closes #1088.

##### Description of changes

Screenshot 2019-06-06 at 12 17 26

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -10,8 +10,8 @@ import numpy as np
 from AnyQt.QtWidgets import (
     QGraphicsWidget, QGraphicsObject, QGraphicsLinearLayout, QGraphicsPathItem,
     QGraphicsScene, QGraphicsView, QGridLayout, QFormLayout, QSizePolicy,
-    QGraphicsSimpleTextItem, QGraphicsLayoutItem, QAction, QComboBox
-)
+    QGraphicsSimpleTextItem, QGraphicsLayoutItem, QAction, QComboBox,
+    QGraphicsItemGroup)
 from AnyQt.QtGui import (
     QTransform, QPainterPath, QPainterPathStroker, QColor, QBrush, QPen,
     QFont, QFontMetrics, QPolygonF, QKeySequence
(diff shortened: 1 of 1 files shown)


Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -462,29 +496,30 @@ def select_item(self, item, state):
             self._re_enumerate_selections()
             self.selectionChanged.emit()
 
+    @staticmethod
+    def _create_path(item, path):
```
janezd: This code is just factored out from lines 477 and 592. This change is not necessary; I've needed it in my initial attempt, but then kept it because it's better this way.

Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -1766,7 +1795,8 @@ def __init__(self, parent=None, orientation=Qt.Vertical, value=0.0,
         super().__init__(parent, **kwargs)
 
         self.setAcceptedMouseButtons(Qt.LeftButton)
-        self.setPen(make_pen(brush=QColor(50, 50, 50), width=1, cosmetic=False))
+        self.setPen(make_pen(brush=QColor(50, 50, 50), width=1, cosmetic=False,
+                             style=Qt.DashLine))
```
janezd: Cut-off line often crosses the new labels, which looked ugly. Dashed line looks better.

BlazZupan:
Nice, works well. When a cluster is a singlet, the distance of the text from the clustering line is disproportionally high. Consider C2 and C4 in the following clustering:

image

Another minor change could be to position cluster labels below the line for the lower clustering branches. That is, in the example, you have provided above, C4 should be placed below the line so that the line does not cross the label. Using this rule, C1 in your visualization should be placed below the line as well. Whether or not to implement this is debatable, I like the consistency of labels being all above the clustering line while placing them above and below would reduce the clutter between the text and the graphics.

janezd:
I moved the labels for singlets.

For singlets I also moved the rectangle itself to the left to right-align it with other selection polygons. If you prefer what we had before, I can move them back.

Moving labels below the line, is not doable with this implementation of labels. I only know the shapes of polygons and not where in the structure they lie.

PR #3856, opened 2019-06-06
Add performance scores to Predictions widget

### LINKED ISSUE #2191: Evaluation of Trained Models on New Data
##### Usecase scenario
Say I once trained and saved a classifier. Later, when some more labeled data arrives, I would like to asses the performance of my trained model on the new data (like with Test & Score). Hence I load the model and new data set but can only pass them to `Predictions`. What's missing is the calculation of various metrics, like we have it in Test & Score, from the predictions on new data.

Similarly, this would be useful for training a classifier and evaluating it on the held out data set. Though, this could be done with `Test & Score` to obtain the score, the user has no access to the classifier trained in `Test & Score` and is required to train an other model if he want's to save it.

#### Possible approaches
* `Test & Score` could also accept classifiers and not only learners and would only offer options `Test on test data` and `Test on train data`. Though, this somehow conflicts with the `Test` part of the widget name.
* `Test & Score` could accept predictions (evaluation results) from the `Predictions` widget and it would only calculate and show scores. Same concern as above.
* A new widget for calculating metrics that accepts trained models and new data. Some functionality would probably be duplicated with `Test & Score`.

janezd:
##### Issue

Resolves #2191.

##### Description of changes

- Factor out the table with scores out of Test and Score; it's valuable mainly because column chooser
- Add the table to Predictions

##### Includes
- [X] Code changes
- [X] Some tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -8,9 +8,8 @@ import numpy
 from AnyQt.QtWidgets import (
     QTableView, QListWidget, QSplitter, QStyledItemDelegate,
     QToolTip, QAbstractItemView, QStyleOptionViewItem, QStyle,
-    QApplication,
-)
-from AnyQt.QtGui import QPainter, QHelpEvent
+    QApplication)
+from AnyQt.QtGui import QPainter, QHelpEvent, QStandardItem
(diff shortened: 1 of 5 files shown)


BlazZupan:
Nice, works great!

PR #3862, opened 2019-06-06
Add a stricter version of pylint rcfile

janezd:
This is pylint rcfile that I use internally and against which I test the pylint fixing PRs I'm occasionally making. The difference is that instead of disabling half of tests, I disable only

```
disable=
    no-name-in-module, no-member  # too many false positives from Qt
    too-many-ancestors, # I don't think this is a problem
    no-else-return      # else's may actually improve readability
```

Compared to our usual file, I ignore `too-many-ancestors`, which we treat as the US debt ceiling -- we increase it as needed. :)

I don't yet suggest switching to these settings, but recommend using them internally.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -28,7 +28,7 @@ unsafe-load-any-extension=no
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code
-extension-pkg-whitelist=
+extension-pkg-whitelist=Orange.distance._distance
 
 # Allow optimization of some AST trees. This will activate a peephole AST
 # optimizer, which will apply various small optimizations. For instance, it can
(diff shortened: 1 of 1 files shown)


markotoplak:
Ah, that is why you were fixing attributes defined outside init... Why shouldn't we just switch to this pylintrc?

janezd:
I suppose we could. @lanzagar?

lanzagar:
I don't know how big a difference that will be in practice (how many more PRs will have red crosses because of this). But I am OK with trying the stricter version to see if we can keep up with it.
What I would not like is to start ignoring pylint checks in PRs from time to time (like we are doing with codecov). But we can always go back to the old config (or just make some additional modification) in that case.

janezd:
I'm also concerned that we'd start adding `pylint disable=` everywhere where we're too lazy to actually fix the problem. (I know this temptation so well.)

Alright, I now changed the official pylintrc. Let's see what happens. :)

PR #3869, opened 2019-06-09
Make Results (more like) a dataclass, separate Validation hierarchy

### LINKED ISSUE #3857: Orange.evaluation.testing.Results should be data class
Class `Orange.evaluation.testing.Results` is typical for early days of Orange 3. It's called `Results` and stores results (predictions and probabilities), but derived classes are `CrossValidation` etc.

`Results` should be a data class (possibly with some smart constructors for merging results, and possibly with some methods for splitting results by methods or folds; functions for this are currently implemented in the Test and Score widget). Sampling methods should have a separate hierarchy.

Since we know that this constructor didn't work properly (until #3858), we can be sure that nobody used it in this way (and it's doubtful that any code outside the core Orange used this class anyway). So it may be safe to refactor this.

janezd:
Fixes #3857.

`Results` stores results of validation, but derived classes also perform the validation. A vaguely remember making this (bad) decision at our retreat in Bohinj. :)

These changes break the backward compatibility with `Results` class, but the class itself has been broken for five years, so I doubt anybody used it. To my observation, all code calls `Results()` without arguments.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/evaluation/clustering.py
+++ b/Orange/evaluation/clustering.py
@@ -4,7 +4,7 @@ import numpy as np
 from sklearn.metrics import silhouette_score, adjusted_mutual_info_score, silhouette_samples
 
 from Orange.data import Table
-from Orange.evaluation.testing import Results
+from Orange.evaluation.testing import Results, Validation
 from Orange.evaluation.scoring import Score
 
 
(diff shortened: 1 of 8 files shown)


janezd:
- Remove `__new__` in derived classes. In the base class, explicitly call `__init__` before calling `fit`.
- Replace `fit` with `__call__`. Keep `fit = __call__` for compatibility?
- Add deprecation warning to `__new__` with argument `data` and `learner`.

markotoplak:
@janezd, I removed further unneeded stuff from inits. I think that what worked before this PR also still works now. If you are fine with my commit then we can merge this.

janezd:
I like your changes. You can merge - thanks.

I didn't realize that `__init__` can have different arguments than `__new__`, but in this case it obviously can.

PR #3871, opened 2019-06-10
[ENH] Faster drawing in scatterplot

### LINKED ISSUE #3589: Slow plotting in Scatter Plot
##### Orange version

3.21.0.dev0+920e474

##### Expected behavior
Plots should not slow down drastically when receiving big(ish) datasets.

##### Actual behavior
Plotting 40000 points in scatterplot is already very slow.

Load big dataset
Connect Scatter Plot
Change Axes and observe replotting speed

Using the following code in command line pyqtgraph plots 100000 points in  about 1 sec.
```
import pyqtgraph as pg
import numpy as np
a = np.random.random(100000)
b = np.random.random(100000)
pg.plot(a, b, pen=None, symbol='o')
```

markotoplak:
##### Issue
Fixes #3589

##### Description of changes
1) Reused QPen, QBrush objects and rounded sizes to half a pixel so that the number of different symbols to draw is lower.
2) Deferred updateSpots() that gets called by pyqtgraph in set*() to paint(), where spot atlas is actually needed.

Now adult  with spots of same color, shape=occupation, size=fnlwgt displays in less than 1 second instead of 10.5s as before.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -714,6 +714,7 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
 
         The method is called by `update_sizes`. It gets the sizes
         from the widget and performs the necessary scaling and sizing.
+        The output is rounded to half a pixel for faster drawing.
 
         Returns:
             (np.ndarray): sizes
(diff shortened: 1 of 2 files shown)


BlazZupan:
Marko, this looks great. Will this change propagate to other point-based visualizations, like in t-SNE and MDS widgets?

markotoplak:
@BlazZupan, it will propagate to any scatterplot-based point visualization. With how large data sets did you see a difference?

markotoplak:
@janezd, I replaced lambdas with a function.

PR #3877, opened 2019-06-12
Box plot: Use str_val to print out labels

### LINKED ISSUE #3740: OWBoxPlot: legible axis labels
Follow up from #2110.

Current status of Box Plot widget. Long labels illegible. Elide if labels longer than 10 chars or show just every n-th.
Screen Shot 2019-04-12 at 11 00 09

At the same time, handle reading of datetime variables better. Display datetime, not float.

janezd:
Fixes #3740.

The remaining problem is that axis labels may stick out of view.

Screenshot 2019-06-12 at 18 00 46

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -200,7 +200,8 @@ class OWBoxPlot(widget.OWWidget):
 
         self.label_txts = self.mean_labels = self.boxes = self.labels = \
             self.label_txts_all = self.attr_labels = self.order = []
-        self.scale_x = self.scene_min_x = self.scene_width = 0
+        self.scale_x = 1
+        self.scene_min_x = self.scene_max_x = self.scene_width = 0
         self.label_width = 0
 
         self.attrs = VariableListModel()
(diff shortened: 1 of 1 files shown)


ajdapretnar:
I think this is waaaaaay better than before. 🎉 

- As for labels out of view, I would suggests Robert's great implementation of tooltip on hover. I think that should be ok.
- The +- thing in mean is hilarious. 😁 Ideally it would be sth like 34min 2sec. But I am not sure it is worth putting the effort towards this.
- Mean as such could be rounded? I dunno, actually. But maybe rounding to a second is enough?

janezd:
- I fixed the out-of-view labels. It doesn't optimize the selection (it goes from left to right and omits labels that are out of view or overlap with the last label), but this is only an issue with long labels, such as dates.

- Time differences should be repred by a (not yet existing) method provided by `TimeVariable`. Not in this PR.

- Similarly, `TimeVariable.str_val` should know what to print and what to skip. Also to be discussed, and not in this PR.

With the last commit (and if tests pass), I consider this PR rounded off. :)

PR #3881, opened 2019-06-13
[ENH] Calibration plot (add performance curves) and a new Calibrated Learner widget

### LINKED ISSUE #1267: Feature request: cutoff/threshold value for classification
When using classification usually a p-value is calculated to determine
the class of the target, by default orange uses 0.5 as cutoff or
threshold value if u like. Depending the on this cutoff value the model performs differently.
- It would be nice to obtain these cutoff value(s) (eg as output) from a widget like roc analyis or lift
- Also it would be nice to be able to change the cutoff value
  *\* For example in the learner
  *\* Or in test and score
  *\* or ..

janezd:
This PR revamps an old widget, adds a new one, adds some core functionality, and adds a few minor fixes to base classes.

- `Orange.evaluation.performance_curves.Curves` is a class that computes performance curves (ca, f1, sens, spec...) as functions of probabilities,
- Widget Calibration curves shows such curves,
- A new module `Orange.model.calibration` contains learners and classifiers that set and user thresholds and calibrate probabilities,
- Calibrated Learner is a new widget that produces such models.

Working on this uncovered some hidden bugs and glitches in other parts of the code, hence a large number of commits.

Fixes #1267.


Orange/evaluation/testing.py
```diff
@@ -317,7 +317,7 @@ def split_by_model(self):
                 res.probabilities = self.probabilities[(i,), :, :]
 
             if self.models is not None:
-                res.models = self.models[:, i]
+                res.models = self.models[:, i:i+1]
```
janezd: I believe this was a bug because it changed self.models from 2d to 1d array. Apparently nobody needed this so far.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -756,7 +757,8 @@ def __update(self):
                     stratified=self.shuffle_stratified,
                     random_state=rstate)
             elif self.resampling == OWTestLearners.TestOnTrain:
-                sampler = Orange.evaluation.TestOnTrainingData()
+                sampler = Orange.evaluation.TestOnTrainingData(
+                    store_models=True)
```
janezd: Without these changes, Orange canvas can't produce evaluation results with valid models that Calibration plot can output. This will also be useful in, say ROC Analysis that will now also be able to output a model with some specific threshold.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -315,7 +315,7 @@ def set_learner(self, learner, key):
             # Removed
             self._invalidate([key])
             del self.learners[key]
-        else:
+        elif learner is not None:
```
janezd: We haven't encountered this bug before because (I suppose) no learner widget can fail to produce a learner. Now it can - Calibrated Learner outputs `None` if it doesn't have a learner on its input.

Orange/widgets/evaluate/contexthandlers.py
```diff
@@ -1,47 +1,30 @@
+from Orange.data import Variable
 from Orange.widgets import settings
-from Orange.widgets.utils import getdeepattr
 
 
 class EvaluationResultsContextHandler(settings.ContextHandler):
-    def __init__(self, targetAttr, selectedAttr):
-        super().__init__()
-        self.targetAttr, self.selectedAttr = targetAttr, selectedAttr
+    """Context handler for evaluation results"""
```
janezd: This is obviously a new context handler with the same name as an old one that nobody used because it probably didn't work and it was weird. ROC Analysis and Lift Curves currently don't have context; they can now have this one.

janezd:
The potentially problematic commit (but I think it's OK) is https://github.com/biolab/orange3/pull/3881/commits/f742ff919d4c2985a38cdee5790cb7230962c398.

Currently, widgets for learners set the default name and the user can change it. The default name was always fixed. In Calibrated Learner however, the name should depend on the widget's input and settings to produce a meaningful name. The widget could reset the name (as a part of its context?), but with this it would override user's changes.

This commit sets the default name as editline's placeholder, so the user sees the default, (s)he can enter a different name, the widget can change the default (without having to pay attention to any user's changes) and it the user then removes hir input, (s)he gets the widget's default.

ajdapretnar:
There's a glitch when changing the threshold. Try it with Logistic Regression, which has a long name. The difference between the size of the numbers makes the widget change size. I suggest a little padding to account for long model names.

calibrationplot

Orange/widgets/evaluate/owcalibrationplot.py
```diff
@@ -117,106 +211,274 @@ def clear(self):
...
+        wrapped = None
+        results = self.results
+        if results is not None:
+            problems = [
+                msg for condition, msg in (
+                    (len(results.folds) > 1,
+                     "each training data sample produces a different model"),
+                    (results.models is None,
+                     "test results do not contain stored models - try testing on"
+                     "separate data or on training data"),
```
ajdapretnar: This produces a badly formatted warning, specifically "try testing onseparate data". Please introduce a space here or a newline.

janezd:
Huh. I limited the model name to 20 characters (elide if longer; I can't add tooltips because Qt's rich text doesn't support them). Not the most elegant and foolproof solution, but I don't have a better one.

ajdapretnar:
`Exception: | AttributeError: 'OWCalibratedLearner' object has no attribute 'calibrate'`

Press Report in Calibrated Learner widget.

janezd:
> `Exception: | AttributeError: 'OWCalibratedLearner' object has no attribute 'calibrate'`
>
> Press Report in Calibrated Learner widget.

I obviously haven't tried it. :) Now it works.

ajdapretnar:
To merge this PR I would suggest:
- adding an icon for Calibrated Learner (at least a temporary one)
- check the formatting of report in Calibration Plot
- make sure Calibration Plot works with sparse (or at least doesn't crash)

janezd:
Merged due to lack of activity.

PR #3883, opened 2019-06-14
[ENH] t-SNE: Load openTSNE lazily

### LINKED ISSUE #3523: Orange.projection loads slowly after introduction of fastTSNE
After f5eeaa6cf24b7ab62a8a24847b43ec8489647048 (tSNE: Implement Orange wrapper around Pavlin's implementation of tSNE) loading of Orange.projection became even slower.

On my computer, it used to take 0.9 second, now it takes 2.3 second.

Should we import fastTSNE only when actually needed?

pavlin-policar:
##### Description of changes
Fixes #3523.

Load openTSNE (and, in turn, numba) as a lazy module.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/manifold.py
+++ b/Orange/projection/manifold.py
@@ -8,10 +8,6 @@ from scipy.linalg import eigh as lapack_eigh
 from scipy.sparse.linalg import eigsh as arpack_eigh
 import sklearn.manifold as skl_manifold
 
-import openTSNE
-import openTSNE.affinity
-import openTSNE.initialization
-
 import Orange
 from Orange.data import Table, Domain, ContinuousVariable
 from Orange.distance import Distance, DistanceModel, Euclidean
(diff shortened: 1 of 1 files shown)


pavlin-policar:
@markotoplak please check.

markotoplak:
@pavlin-policar, works well. I am back to 1 second. Thanks!

Orange/projection/manifold.py
```diff
@@ -190,6 +182,23 @@ def __init__(self, n_components=2, affinity='nearest_neighbors', gamma=None,
...
 
 
+class lazy_openTSNE:
+    """openTSNE uses numba, which is slow to load, so load it lazily."""
+    def __getattr__(self, item):
+        import sys
+        import openTSNE
+        import openTSNE.affinity
+        import openTSNE.initialization
+        if "openTSNE" in sys.modules:
```
janezd: @pavlin-policar, I'm trying to revive this, but since the change caused a crash on travis in bioinformatics, it is difficult to debug.

Lookin for a cause, I found this condition very strange. When you import a module, it always appears in `sys.modules`, doesn't it? Why did you need this line, what did it try to prevent, what was the case when the condition was false?

Have you by any chance tried adding `global openTSNE` to the function?

Is it possible that bioinformatics tried to do something in parallel and this function was executed in parallel (though I don't see how this could have happened), and numba initialized twice? Qt, for instance, does (or used to) crash if you initialized QApplication twice.

PR #3885, opened 2019-06-14
[FIX] OWLinearProjection: limit axis

### LINKED ISSUE #3743: OWLinearProjection: legible axis labels
Follow up on #2110.

Current widget status. Linear Projection is hard to 'destroy'. But if one has a really, really large number of attributes, it is possible. I suggest limiting the number of attributes that can be displayed to, say, 20. If feature names are reaaaaaaaaallllly long, elide them at... dunno, 20 chars.
Screen Shot 2019-04-12 at 11 21 29

robertcv:
##### Issue
Fixes #3743
##### Description of changes
Limit the number of displayed axis and labels length to 20.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owlinearprojection.py
+++ b/Orange/widgets/visualize/owlinearprojection.py
@@ -35,6 +35,9 @@ from Orange.widgets.visualize.utils.widget import OWAnchorProjectionWidget
 from Orange.widgets.widget import Msg
 
 
+MAX_LABEL_LEN = 20
+
+
 class LinearProjectionVizRank(VizRankDialog, OWComponent):
     captionTitle = "Score Plots"
     n_attrs = Setting(3)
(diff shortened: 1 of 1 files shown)


janezd:
@robertcv, I'm so sorry but ... could you keep just eliding and remove the limit on the number of attributes?

This widget is different from Radviz where variables are always placed around a circle; here you can place them arbitrarily (or use PCA) and the labels can overlap even with just 5 (well, just 2) attributes. So limiting their number doesn't help a lot.

Also, in Linear projection you can hide attributes whose axis are shorter than some limit. So you can have 300 atributes, run PCA and hide those with short axes. Or even hide all attributes because you'd just like to see the projection. With the limitation to 20 attributes this is no longer possible.

robertcv:
@janezd I removed the limit on the number of attributes.

PR #3887, opened 2019-06-14
owheatmap: greater dynamic range for colors

### LINKED ISSUE #3884: Heat Map: change how colors are computed
Colors lose precision when you decrease the color range with High and Low. For a better dynamic range implement coloring as in the HyperSpectra widget from the spectroscopy add-on.

markotoplak:
##### Issue
Closes #3884 

The change is not that big. For the shortest possible interval between high and low settings (5%) the old code still used 14 different colors. Now, up to 255 can be used, but only if the data has that many different values in the range that is set.

##### Description of changes
Before this change low and high thresholds cut the color palette. Thus, less colors were used for drawing. Now thresholds only change how the full palette is applied. Therefore, dynamic range improved.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owheatmap.py
+++ b/Orange/widgets/visualize/owheatmap.py
@@ -182,14 +182,11 @@ def color_palette_model(palettes, iconsize=QSize(64, 16)):
     return model
 
 
-def color_palette_table(colors, samples=255,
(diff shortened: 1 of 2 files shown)


janezd:
It's indeed way better!

PR #3897, opened 2019-06-21
[FIX] Documentation links

rokgomiscek:
##### Issue
Some widgets reported missing documentation even if this wasn't the case.

##### Description of changes
Fixed names so documentation links point to the right file.

##### Includes
- [x] Code changes
- [ ] Tests
- [x] Documentation


Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -125,7 +125,7 @@ class State(enum.Enum):
 
 
 class OWTestLearners(OWWidget):
-    name = "Test & Score"
+    name = "Test and Score"
```
ajdapretnar: This changes the name of the widget in Orange. Which means all schemas that had "Test & Score" before could fail. Please check.

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -94,7 +94,7 @@ class NotEnoughData(ValueError):
 
 
 class OWKMeans(widget.OWWidget):
-    name = "k-Means"
+    name = "k-Means clustering"
```
ajdapretnar: This changes the name of the widget and makes it longer. I don't think this is ok. If docs don't match, we can change the docs.

doc/visual-programming/source/widgets/data/applydomain.md
```diff
@@ -1,4 +1,4 @@
-Transform
+Apply Domain
 =========
```
ajdapretnar: Underscores have to fit the length of the title.

doc/visual-programming/source/widgets/evaluation/confusionmatrix.md
```diff
@@ -1,4 +1,4 @@
-Predictions
+Confusion Matrix
 ===========
```
ajdapretnar: Same as above.

doc/visual-programming/source/widgets/data/applydomain.md
ajdapretnar: Also, the change in not reflected in the index.rst. Could you please fix this too?

ajdapretnar: Sorry for these continuous comments. The documentation for Apply Domain should be changed to reflect the changes in the widget name. Namely the references to the widget, image names (so we know what they refer to) and likely the example workflows.

Orange/widgets/evaluate/owtestlearners.py
rokgomiscek: It works (I checked on an example workflow).

Orange/widgets/unsupervised/owkmeans.py
rokgomiscek: Changing the name would make this similar to Hierachical and Louvain Clustering (though "clustering" should be capitalized).

doc/visual-programming/source/widgets/data/applydomain.md
rokgomiscek: Done.
Apply Domain (ex Transform) not present in example workflows.

Orange/widgets/unsupervised/owkmeans.py
lanzagar: Hierarchical and Louvain can not be used by themselves - "clustering" is part of the name there.
k-Means is clear enough and can be used as a name by itself.
I think we should strive for short widget names, so I would not expand it in this case.
BTW, clustering is already a keyword for k-means so using it in a search lists it alongside other methods.

rokgomiscek: Ok. Changed back widget name and changed documentation file name.

doc/visual-programming/source/widgets/data/applydomain.md
ajdapretnar: Sorry, I didn't mean example workflows in Orange tutorial section, but workflows that are in Example section of the documentation.

PR #3898, opened 2019-06-21
[FIX] Change normalization to Scaling in SVM

### LINKED ISSUE #3202: SVM fails when mixing original and BoW attributes
##### Orange version

3.16.dev

##### Expected behavior
SVM works with election-tweets-2016.tab from Text add-on.

##### Actual behavior
SVM shows 'Fitting failed. Non-zero offset in normalization of sparse data.'
This only happens when additional, non-bag-of-words attributes are present.

Corpus - Data Table (select a couple of instances for performance) - Bag of Words (Count) - SVM.

Try to remove original attributes (Favourite count, Retweet count, Latitude, Longitude, Is retweet and Language) and SVM starts working. I think the issue might be in different handling of attributes (from original data and from bag of words).

See https://github.com/biolab/orange3-text/issues/312

AndrejaKovacic:
##### Issue

Fixes  #3202

##### Description of changes
SVM was failing because Corpus widget outputs sparse matrix which must be then turned to dense, if we want to mean shift it. We would like to avoid that, so the current proposal is to just scale the data. The user can opt out trough a check box. Currently the check box is in the optimization parameters box, perhaps something else is needed.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/preprocess.py
+++ b/Orange/preprocess/preprocess.py
@@ -18,7 +18,7 @@ from . import impute, discretize, transformation
 __all__ = ["Continuize", "Discretize", "Impute", "RemoveNaNRows",
            "SklImpute", "Normalize", "Randomize", "Preprocess",
            "RemoveConstant", "RemoveNaNClasses", "RemoveNaNColumns",
-           "ProjectPCA", "ProjectCUR", "Scale"]
+           "ProjectPCA", "ProjectCUR", "Scale", "SmartNormalize"]
 
 
 class Preprocess(_RefuseDataInConstructor, Reprable):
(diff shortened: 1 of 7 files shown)


Orange/preprocess/preprocess.py
```diff
@@ -568,3 +568,52 @@ def __call__(self, data):
         for pp in self.preprocessors:
             data = pp(data)
         return data
+
+
+class SmartNormalize(Preprocess):
```
thocevar: We need a better name for this kind of normalization that handles both dense and sparse cases.

Orange/preprocess/preprocess.py
```diff
@@ -568,3 +568,52 @@ def __call__(self, data):
...
+        passed to Normalize
+
+    norm_type : NormTypes (default: Normalize.NormalizeBySD)
+        passed to Normalize
+
+    transform_class : bool (default=False)
+        passed to Normalize
+
+    center :
+        passed to Normalize as boolean and to Scale as recieved in constructor
```
thocevar: This is ignored in scale, right?

Orange/preprocess/preprocess.py
```diff
@@ -568,3 +568,52 @@ def __call__(self, data):
...
+
+    normalize_datetime : bool (default=False)
+        passed to Normalize
+
+    scale : ScaleTypes (default: Scale.NoCentering)
```
thocevar: Default seems to be Scale.Std but I'm not sure it makes sense to scale by std if you don't center the data at the mean (std is computed as deviation from the mean). Instead, default to Scale.Span?

Orange/preprocess/preprocess.py
```diff
@@ -568,3 +568,52 @@ def __call__(self, data):
...
+                 zero_based=True,
+                 norm_type=Normalize.NormalizeBySD,
+                 transform_class=False,
+                 normalize_datetime=False,
+                 center=Scale.Mean,
+                 scale=Scale.Std):
+        self.normalize_pps = Normalize(zero_based,
+                                       norm_type,
+                                       transform_class,
+                                       ~(center is not False),
```
thocevar: Normalize centers the mean. Passing center=Scale.Median would give unexpected results. Try to make it consistent with Normalize.

Orange/widgets/model/owsvm.py
```diff
@@ -29,6 +31,10 @@ class OWSVM(OWBaseLearner):
     class Outputs(OWBaseLearner.Outputs):
         support_vectors = Output("Support vectors", Table, explicit=True)
 
+    class Warning(OWWidget.Warning):
+        sparse_data = Msg('Input data is sparse, default preprocessing is to scale it.')
+        outdated_learner = Msg("Press Apply to submit changes.")
```
thocevar: Is `outdated_learner` used anywhere?

Orange/widgets/model/tests/test_owsvm.py
```diff
@@ -87,3 +90,15 @@ def test_kernel_spins(self):
...
                               for j in range(3)], hidden)
+
+    def test_sparse_warning(self):
+        """Check if the user is warned about sparse input"""
+        data = Table("iris")
+        data.X = csr_matrix(data.X)
+        self.send_signal("Data", data)
+        self.assertTrue(self.widget.Warning.sparse_data.is_shown())
+
+    def test_dense_input(self):
```
thocevar: I think you can merge these two functions since both test a "sparse warning".

Orange/widgets/model/owsvm.py
AndrejaKovacic: It is used in the parent class. I redefine it here, since i needed to add sparse data warning.

Orange/preprocess/preprocess.py
AndrejaKovacic: Thanks, it slipped my mind that scaling has a set sentering paramter. I changed it to boolean.

Orange/widgets/model/owsvm.py
thocevar: Inherit from OWBaseLearner.Warning in the same way as is done for Outputs.

Issue #1847, opened 2016-12-30
Random sample generator widget


brylie:
# User story
> As a data analyst
I would like to generate some random samples, fitted to a given distribution
so that I can visualize how well my observed data fits a general distribution

# Feature
A random sample generator widget could generate *n* samples from a selected distribution with necessary input parameters. The output data could be fed into a visualization widget, possibly allowing comparison of a general distribution with the distribution of values within an observation data set.

# Related resources
The [`np.random` module](https://docs.scipy.org/doc/numpy/reference/routines.random.html) has many types of random data generators, such as:

- [binomial](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.binomial.html#numpy.random.binomial)
- [normal](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.normal.html#numpy.random.normal)
- [poisson](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.poisson.html#numpy.random.poisson)

janezd:
Widgets like this are easy to implement. This is dangerous since we could easily end up with hundreds and hundreds of widgets, which would confuse the users on one side, and make it a maintenance nightmare on the other.

The functionality you propose could be added to the Paint widget. This would require renaming the widget to something like *Create data*. @BlazZupan?

brylie:
> we could easily end up with hundreds and hundreds of widgets

This is a proposal for one multi-purpose widget that would generate random samples based on the `np.random` library.

janezd:
Sorry, I wasn't clear. With "Widgets like this" I didn't mean widgets for generating distribution samples, but simple (single multipurpose) widgets like this one. We come up with ideas for new (single) widgets every day. Since implementing a widget for a simple data manipulation or a widget that wraps a numpy or scikit function takes so little time, we have to resist the temptation to create seven new widgets a week. Many of our internal discussions are about whether a new widget is necessary or can the functionality be added to another widget, without making that widget too complicated.

This is why I think we should consider merging your suggestion into a general data creation widget. In `Paint`, we would just add a button for adding a random sample; the user would choose the distribution and its parameters in a dialog, and the widget would add a sample to the existing data in the plot.

brylie:
Cool, so the 'Paint' widget might take on a new name, such as 'Paint/generate', to match its extended purpose?

shivin9:
Hi @brylie,
[...]

PR #3901, opened 2019-06-25
[NOMERGE] Random Data: New widget

janezd:
Implements #1847. Open for comments.

This will be moved to Educational add-on. (I can't do the work there because `python setup.py develop` for educational add-on still breaks my conda environment!? After I install the add-on in this way, orange crashes with Abort trap 6. Even if I uninstall the add-on, the environment is still broken?!)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owrandomdata.py
+++ b/Orange/widgets/data/owrandomdata.py
@@ -1,0 +1,474 @@
+from collections import namedtuple
+from itertools import chain
+from math import ceil, log10
+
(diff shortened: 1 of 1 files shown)


janezd:
I managed to (non-replicably) install the educational add-on without breaking my environment, so this continues in https://github.com/biolab/orange3-educational/pull/70.

PR #3911, opened 2019-06-28
[FIX] OWNeighbours fix manual apply for some options

PrimozGodec:
##### Issue

Manual apply didn't work for `number of neighbors` and `Exclude rows`. Those two fields were applied automatically even when `Apply automatically` unchecked.

##### Description of changes

Manual apply is enabled for `number of neighbors` and `Exclude rows`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owneighbors.py
+++ b/Orange/widgets/data/owneighbors.py
@@ -77,11 +77,11 @@ class OWNeighbors(OWWidget):
             callback=self.recompute)
         gui.spin(
             box, self, "n_neighbors", label="Number of neighbors:",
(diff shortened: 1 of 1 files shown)


janezd:
Good catch.

I replaced `commit` with a lambda function.

Some widgets have `commit` and some have `apply`, which is annoying enough because you never know what to call when working on different widgets. Now after this PR we would have `apply` *and* `commit`, which would have the same effect, but only after `gui.auto_commit` is called. This is even more confusing. :)

I think it's better to have either lambda or methods like `on_n_neighbours_changed` (which then call `commit`. Or `apply`. Whichever :)).

PR #3922, opened 2019-07-04
Fix 404ing widget documentation URL

irgolic:
After the website change, the widget catalog URL changed. This PR updates it to the correct URL.


markotoplak:
Good catch! We should also fix the web site. Let's not forget that none of our users is on master currently and that many are running older versions.

irgolic:
I've messaged @acopar about it. :)

Issue #3705, opened 2019-03-29
Remove Variable.make


lanzagar:
See also the following (and other) issues:
#2500, #2346, #2943, #3521 

Can we remove `Variable.make` and solve the problems it is supposed to be solving in another way?
We were discussing with @janezd that we could try to automatically match variables when that is needed instead of reusing _and changing_ existing objects.
For example building a model on training data and then trying to use it on new test data would have problems without `make`. But the model needs to transform new data into its domain (that of the training data) anyway, and could instead do that in a better way. Just map variables with the same name, type and compute_value (so that variable age with values 1-100 can't map to a normalized version with values 0-1). For discrete variables, `values` need special care since some can be missing in train/test data in which case we need to remap the values.

Before starting work on this, we need a better estimate of how many things would need to be changed. The main one is `Table.from_table(domain2, data1)` which also covers the use of `transform`. Then there is `Table.get_column_view()` and the underlying `Domain` operations (`index`, `__get_item__`, etc).
We also need to take care that transformations (through their use of `compute_value`) still work.

janezd:
1. Search the code for all places that require changing (assigned to nitpicker @lanzagar).
2. Change it.

Issue #3521, opened 2019-01-08
Predictions differ when training/predicting in the same instance versus pickling the model and loading it into a new instance for predictions


pomalley44:
##### Orange version
3.17.0

##### Expected behavior
Two methods below should yield the same predictions:

**Method A:**
Start new python instance
Load training data to Orange table (train_data)
Create learner
```
learner = RandomForestRegressionLearner(random_state=0, n_estimators=50, min_samples_split=2)
```
Create model from learner trained on train_data
```
model = learner(train_data)
```
Pickle model
```
with open('model.pickle', "wb") as f:
    pickle.dump(model, f, 0)
```
Load target data to Orange table (predict_data)
Unpickle model
```
with open('model.pickle', "rb") as f:
    model = pickle.load(f)
```
Run predictions
```
pred = model(predict_data)
```

**Method B:**
Start new python instance
Load target data to Orange table (predict_data)
Unpickle model
```
with open('model.pickle', "rb") as f:
    model = pickle.load(f)
```
Run predictions
```
pred = model(predict_data)
```

##### Actual behavior
Unpickling the model in a new instance and running predictions give different results than when predictions were run from the same instance where the model was trained.

##### Steps to reproduce the behavior
I have tried reproducing this error with built-in datasets, but I have not been successful doing so.

##### Additional info (worksheets, data, screenshots, ...)

I'm using random forest regression with ```random_state=0, n_estimators=50, min_samples_split=2``` . My training data set is ~15,000 rows with ~350 columns--a mixture of continuous features, binary features, and categorical features (that I'm relying on Orange to preprocess with one-hot encoding).

The results from **Method A** and **Method B** are significantly different. The predicted populations have different means and standard deviations.  The predictions in **Method B** appear much "flatter", with the mean and standard deviation both much lower. However, there is still some correlation--items with high prediction in **Method A** tend to have a high prediction in **Method B** (relative to the overall range).

I've included pickling/unpickling in **Method A** above for symmetry's sake. Leaving the model in memory yields the same results. I've also tried:

- loading (unpickling) the same pickle file into a different object in memory
- making a copy of the pickle file in the OS and loading (unpickling) that file into memory

Both of the above experiments still yield the same prediction results in **Method A**.

My suspicion is that some sort of preprocessing is being saved in memory and not saved in the pickled model, but I don't know what that could be or how to figure it out.

I'm happy to try some different things if someone can point me in the right direction.

ales-erjavec:
[...]

Issue #3049, opened 2018-06-01
Neural network: Different result in GUI and python


horvathj:
##### Orange version

3.13

##### Expected behavior
Get same result in GUI and python when using neural network modell prediction (for binary classification).

##### Actual behavior
Different result in GUI. Maybe it is wrong?

##### Steps to reproduce the behavior
(Small files, modell with train in 2-3 seconds)
![_flow](https://user-images.githubusercontent.com/10976599/40847788-9cac507c-65bd-11e8-850e-bbc3c0b91e77.jpg)

1. Rename to (.csv) and Load the attached train.txt file, and connect into a Neural Network modell.
[train.txt](https://github.com/biolab/orange3/files/2063099/train.txt)

2. Parameters for Neural Network:
- Hidden neurons: 1
- Activation: ReLu
- Optimizer: L-BFGS-B
- alpha: 0,0001
- Max iterations: 300

3) Connect the modell in to prediction modul.

4) Rename to .csv and Load the test.csv  and connect into prediciton modul.
[test.txt](https://github.com/biolab/orange3/files/2063100/test.txt)

5) Check the predictions, and retrain the neural network modell while the first 2 row is bus, and the second 2 is police. AND the third row is 0.43 : 0.57

![_results](https://user-images.githubusercontent.com/10976599/40847836-c36c23a4-65bd-11e8-974b-29a2b8cd9c78.jpg)

6) Save the neural network modell into a pickle file "neural.pkcls" with Save model.
--
7) Activate a python environment with installed Orange and pickle packages.

8) Run this code in python:
```
import Orange
import pickle
nn = pickle.load( open( "neural.pkcls", "rb" ) )
nn.predict([[0.327,0.926],[0.077,0.498],[0.518,0.512],[0.499,0.335]])
```
9) The first array shows the categorisation (on my computer it is wrongly: [0., 0., 0., 1.] and the predict_proba numbers are not the same as in prediciton result table. 

![_cmd](https://user-images.githubusercontent.com/10976599/40848142-9c921eb8-65be-11e8-8fa8-7dea324e974a.jpg)

##### Additional info (worksheets, data, screenshots, ...)

If I list out the coefs_ and intercepts_ and do the MLP math in excel, I get the python results. But the GUI shows something else, what I can not understand.

horvathj:
If I train the model in python I get the same result as in the prediction window.

```
train = Orange.data.Table("train.csv")
test = Orange.data.Table("test.csv")

[...]

PR #3925, opened 2019-07-05
Remove Variable.make

janezd:
Implements #3705.   Probably also #3521, #3949, #3049.

##### Includes

- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -117,7 +117,7 @@ class Learner(ReprableWithPreprocessors):
                             self.__class__.__name__)
 
         model = self._fit_model(data)
-        model.used_vals = [np.unique(y) for y in data.Y[:, None].T]
+        model.used_vals = [np.unique(y).astype(int) for y in data.Y[:, None].T]
         model.domain = data.domain
         model.supports_multiclass = self.supports_multiclass
         model.name = self.name
@@ -226,53 +226,31 @@ class Model(Reprable):
         def fix_dim(x):
             return x[0] if one_d else x
 
-        def backmap_value(value, probs):
-            if backmapper is None:
-                return value
-            value = backmapper(value)
-            nans = np.isnan(value)  # What happens if there's just a single value?
+        def data_to_model_domain():
+            if data.domain == self.domain:
+                return data
 
-            if not np.any(nans):
-                return value
-            if probs is not None:
-                value[nans] = backmapper(np.argmax(probs[nans], axis=1))
-            else:
-                value[nans] = np.RandomState(0).choice(
-                    backmapper(np.arange(0, len(dataclass.values) - 1))
-                    (np.sum(nans), ))
-            return value
+            if self.original_domain.attributes != data.domain.attributes \
+                    and data.X.size \
+                    and not np.isnan(data.X).all():
+                new_data = data.transform(self.original_domain)
+                if np.isnan(new_data.X).all():
+                    raise DomainTransformationError(
+                        "domain transformation produced no defined values")
+                return new_data.transform(self.domain)
 
-        def backmap_probs(probs):
-            if backmapper is None:
-                return probs
-            new_probs = np.zeros((len(probs), len(dataclass.values)),
-                                 dtype=probs.dtype)
-            for col in range(probs.shape[1]):
-                target = backmapper(col)
-                if not np.isnan(target):
-                    new_probs[:, int(target)] = probs[:, col]
-            tots = np.sum(new_probs, axis=1)
-            tots[tots == 0] = 1
-            new_probs = new_probs / tots[:, None]
-            return new_probs
+            return data.transform(self.domain)
 
-        if not 0 <= ret <= 2:
-            raise ValueError("invalid value of argument 'ret'")
-        if ret > 0 and any(v.is_continuous for v in self.domain.class_vars):
-            raise ValueError("cannot predict continuous distributions")
+        def get_backmappers():
+            backmappers = []
+            n_values = []
 
-        # Call the predictor
-        one_d = False
-        backmapper = None
-        if isinstance(data, np.ndarray):
-            one_d = data.ndim == 1
-            prediction = self.predict(np.atleast_2d(data))
-        elif isinstance(data, scipy.sparse.csr.csr_matrix):
-            prediction = self.predict(data)
-        elif isinstance(data, (Table, Instance)):
-            if data.domain.class_var is not None:
-                dataclass = data.domain.class_var
-                modelclass = self.domain.class_var
+            dataclasses = data.domain.class_vars
+            modelclasses = self.domain.class_vars
+            if len(dataclasses) != len(modelclasses):
+                raise DomainTransformationError(
+                    "Mismatching number of model's classes and data classes")
+            for dataclass, modelclass in zip(dataclasses, modelclasses):
                 if dataclass != modelclass:
                     if dataclass.name != modelclass.name:
                         raise DomainTransformationError(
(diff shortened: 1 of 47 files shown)


janezd:
This passes rather lots of tests without significant modifications in code (except in tests that do ugly things, like renaming variables).

`Variable.name` is not read-only property because it's used in `Variable.__hash__`.

I think we should deprecate `Domain.index` and `Domain.__contains__` because they are not well-defined.

janezd:
@lanzagar, can you take a look? There are still things to smooth out (like removing all calls of Variable.make, which are currently forwarded to normal constructors), but in principle this seems to work. Comments from others very welcome, too, of course.

lanzagar:
I can look at it on Thursday/Friday.

As for `Domain.__contains__` - I think it is ok the way it works / is defined now: checks if a variable is (anywhere) in the domain... What was/is a big inconsistency is that it does not match `__iter__`, but that is already deprecated and I am very much in favour of changing it - one can already iterate over attributes, class vars, variables (atts + clsvars), or metas in an explicit and well-defined way.
`__iter__` could be removed altogether, but it might be useful to have it iterate over all vars in the domain in some agreed upon and defined order? In any case `index` should almost certainly match `__iter__` to make sense, and not abuse negative indices, which confuses everyone that knows python.

janezd:
> As for Domain.__contains__ - I think it is ok the way it works / is defined now: checks if a variable is (anywhere) in the domain...

The problem is that two variables are now equal if they are "compatible". As a consequence, `var in domain` will no longer be equivalent to any(var == var0 for var0 in domain)`. I don't think `Domain.__contains__` is very useful anyway.

markotoplak:
I like this approach that now all discrete variable with the same name (without compute_value) are equal and then DomainConversion creates a mapping as a compute_value-type function. I have no counterexamples yet: I do not see a reason why it couldn't work, but I am also not sure that it will. :)

lanzagar:
Some bugs will probably need to be ironed out, but I think this is good conceptually.
The first one to investigate, which produces a crash in the current version is:
File (iris) -> Logistic Regression -> Predictions  <- File (iris12),
where iris12 is a file with only 2 class values from iris, e.g. first 100 instances

```
  File "/home/lan/dev/orange3/Orange/widgets/evaluate/owpredictions.py", line 612, in predict
    return self.predict_discrete(predictor, data)
  File "/home/lan/dev/orange3/Orange/widgets/evaluate/owpredictions.py", line 618, in predict_discrete
    return predictor(data, Model.ValueProbs)
  File "/home/lan/dev/orange3/Orange/classification/base_classification.py", line 47, in __call__
    probs_ext[:, c, cv] = probs[:, c, i]
IndexError: index 2 is out of bounds for axis 2 with size 2
```

janezd:
@lanzagar: the problem was caused by inconsistency in `SklModelClassification`, which overrode `__call__` instead of `predict` and so my new code was called from the wrong place. These are good news: the concept is still sound, it's broken by existing bad code. (I'm not pointing fingers, its author was here for three weeks when writing it. :)

More good news: the schema I used (logistic regression trained on two-class Iris from Select Rows is tested on three-class directly from File) does not work on master branch. Here it now works, after the last fixes.

janezd:
@lanzagar, @markotoplak, now would be a good time to check and merge -- if OK. This will surely introduce some bugs, we must use it for some time before releasing it.

"47 changed files" looks scary, but most are just fixes in tests -- removal of clear_all_caches and giving names to test variables with empty names.

markotoplak:
I tried this with tests from Spectroscopy. Your changes do not break anything significant. Great work!

What worries me that now tests of spectroscopy took 302 sec. On master they finish in 61 seconds. These tests are pretty heavy on data processing and classification where the test data has to be transformed. I'll try to narrow this down a bit.

Orange/data/domain.py
```diff
@@ -254,6 +258,13 @@ def empty(self):
         """True if the domain has no variables of any kind"""
         return not self.variables and not self.metas
 
+    def _get_equivalent(self, var):
+        if isinstance(var, Variable):
```
markotoplak: I tried this:

    import time
    import Orange
    
    data = Orange.data.Table([list(range(5000))])
    
    t = time.time()
    pdata = Orange.preprocess.SklImpute()(data)
    print("first transform", time.time()-t, "s")
    
    t = time.time()
    data.transform(pdata.domain)
    print("Table.tranform", time.time()-t, "s")

On master I get 

    first transform 0.6564505100250244 s
    Table.tranform 0.5023694038391113 s

but on this branch

    first transform 11.536584615707397 s
    Table.tranform 0.5088396072387695 s

My profiler led me here.

Orange/widgets/evaluate/owpredictions.py
```diff
@@ -230,20 +230,22 @@ def set_predictor(self, predictor=None, id=None):
                 PredictorSlot(predictor, predictor.name, None)
 
     def _set_class_var(self):
+        self.class_var = self.data and self.data.domain.class_var
```
lanzagar: `self.class_var` now uses the variable from test data instead of the one from models.
This is wrong, test data's target is mostly irrelevant (we also allow it to be None) and should only be used to double check that the user is not making a mistake by using incompatible models and data.

Arguably, the mismatch error could have been a warning instead, since if we allow no target, why should we stop everything when it is present, but different. This was probably your thinking as well, when you removed `self.class_var = None` after the mismatch error? I think the policy in Orange is to stop everything on Errors and use Warnings when something is not as it should be, but the work can continue.

lanzagar: An example of this causing problems is the same as the one I mentioned above.
> File (iris) -> Logistic Regression -> Predictions <- File (iris12),
> where iris12 is a file with only 2 class values from iris, e.g. first 100 instances
The control area changes to show only 2 class values after test data is connected. There are also only 2 probabilities shown, although the model should report 3. Compare this with removing the target from test data.

Orange/base.py
```diff
@@ -211,39 +211,151 @@ def predict_storage(self, data):
...
+        def backmap_value(value, mapped_probs, n_values, backmappers):
+            if backmappers is None:
+                return value
+
+            if value.ndim == 2:  # For multitarget, recursive call by columns
+                new_value = np.zeros(value.shape)
+                for i, n_value, backmapper in zip(
+                        itertools.count(), n_values, backmappers):
+                    new_value[:, i] = backmap_value(
+                        value[:, i], probs[:, i, :], [n_value], [backmapper])
```
janezd: `probs` -> `mapped_probs`.

janezd:
Thanks to discussions with @lanzagar, the Predictions widget will now look very different.

- Assume that you created models for predicting different things for some data; some may be numeric, some categorical. The widget now accepts all those models and shows their predictions although they may have different targets.
- If the data has target class, the widget still shows a table with performance scores for models that predict that particular target (and it shows a warning about models that don't).

This uses Bayesian classifiers to predict gender and chest pain, and linear regression for age; none matches the target in the data, so no scores are shown at the bottom:

Screenshot 2019-09-15 at 00 20 17

This makes widget more useful while also simplifying the code. But the GUI still consists of mostly redundant options, IMHO.

- Box "Output" allows the user to choose what appears in the output: original data, predictions and predicted probabilities. The output without the original data is probably useless. Furthermore, having a few more (meta) columns in the output never hurts. Hence I would eliminate the box and always output all predictions and probabilities.

- Box "Data View" allows the user to choose whether the widget also shows the original data or just classes and meta attributes. This made sense when attributes were shown to the left of the class and metas. Now that their columns are on the right and mostly invisible, I would eliminate this, too, and always show them.

- Box "Show" includes a list view that allows the user to select the probabilities (s)he wants to see. This list view is now more complicated because it contains all class values that appear in the target of any model. But even without this, I doubt that users ever select multiple classes. I suppose we could limit the options to show

   - probabiliities for all classes
   - probability for the class that is predicted
   - probability for a single class selected by the user

   I would keep the first checkbox (Predicted class, but rename it to "Prediction") and replace the rest with a combo.

- It is also questionable whether we really need the first check box. What about always showing prediction?

- Distributions bars are not very useful, but they are a nice decoration and don't hurt anybody, so let's just have them always on ... and remove the checkbox.

This would replace 7 checkboxes with 1 or 0 checkbox and 1 combo (or three radio buttons and a combo). I guess I would then even dare putting these on top of main area (and move the info to status bar), thus giving the widget more horizontal space at a minimal cost of the vertical.

Comments?

janezd:
This (or less -- prehaps replace listview with a reasonable combo and move it to main area) should suffice, IMHO.

Screenshot 2019-09-15 at 13 12 06

markotoplak:
Khm... I am a bit embarrassed to complicate this further, but now I do not understand why do models need backmappers. Should the model care about the test (or prediction) class_vars at all?

As far as I see, backmappers are only useful for scoring. There we have the model (predicted) class_vars and test class_vars and we could just map before scoring. Am I missing something?

janezd:
Code-wise, it's trivial; to remove backmapping from models, we'd just delete 5 lines of code (and add them at various other places). But I think we shouldn't.

First and foremost, backwards compatibility. With backmapping, models work the same as they have before. Without, predictions returned by models (in core Orange) wouldn't match the class from test data, so everybody calling the model would have to backmap them (or suffer the consequences (s)he won't understand.)

Conceptually, consider models as if they (pretend to) operate in the "test-data space". They convert the data from test domain to training domain, and convert the results back. Just as the caller doesn't care about the attributes of the original domain of the model, (s)he shouldn't need to care about the class of the original domain. If we eliminate backmapping, should we also remove "forward" mapping?

Predictions widget is an exception, not a rule. I believe that an ordinary user would expect backmapping to happen, so anybody who doesn't want it should specifically disable it (by removing the class from test data).

Furthermore, Predictions widget is an exception because we decided so (in the last commit). Not counting it, the effects of this PR are very limited. If we don't backmap, I believe the side-effects would spread.

markotoplak:
@janezd, I agree that removing backmapping does not belong to this PR.

Forward mapping is necessary for models to work, but back mapping is not. We only need when we compute accuracy - it therefore seems to belong there. But yes, a potential compatibility breaking change and perhaps best handled separately.

PR #3929, opened 2019-07-08
[ENH] Feature constructor: deduce discrete values from function

### LINKED ISSUE #3056: Feature Constructor: easier categorical features
Categorical features are hard to create, because currently:
- the user has to list all possible values
- and then the expression has to return indices for the list of values

I propose that the expression could directly return strings and that possible values were automatically created.

janezd:
Fixes #3056.

@markotoplak, like this?

If the user defines values, the function must return integer indices. If function returns strings (or instances of `Value`), the user mustn't define values. It is contrived, but you can't have both. We will have to mention this in documentation. :) If you have a better solution, feel free to change the code.

The widget now also casts string values into strings, so `name[0]` is now a valid expression for zoo. Before, you'd have to write `str(name)[0]` -- and, of course, list all possible values.

I'll add tests when we confirm the approach.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -25,8 +25,7 @@ import numpy as np
 from AnyQt.QtWidgets import (
     QSizePolicy, QAbstractItemView, QComboBox, QFormLayout, QLineEdit,
     QHBoxLayout, QVBoxLayout, QStackedWidget, QStyledItemDelegate,
-    QPushButton, QMenu, QListView, QFrame
-)
+    QPushButton, QMenu, QListView, QFrame, QLabel)
 from AnyQt.QtGui import QKeySequence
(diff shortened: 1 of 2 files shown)


markotoplak:
I like it! Fixes the issue and, as far as I see, it is also completely backwards compatible.

ajdapretnar:
I love this! @janezd Any chance of increasing coverage or should we just ignore it?

janezd:
> Any chance of increasing coverage or should we just ignore it?

Sure, I planned to add tests: "I'll add tests when we confirm the approach." I'll do it now.

ajdapretnar:
I guess my reading skills are impaired today. 😊

janezd:
Closed in favor of #3936, which is based on this PR but adds more general casting in further commits.

Issue #3931, opened 2019-07-10
Reset widget settings not working


ajdapretnar:
##### Orange version

2.23.dev

##### Expected behavior
~CSV File Import is empty when clearing widget settings.~
Actually, clear settings should work but it doesn't.

##### Actual behavior
CSV File Import loads previous data.

##### Steps to reproduce the behavior
Open .csv file with CSV File Import. Press Reset widget settings (or run --clear-widget-settings) and reopen Orange. Place CSV File Import on the canvas. It loads previous data.

PR #3772, opened 2019-04-25
[ENH] Separate canvas and base widget

ales-erjavec:
##### Issue

Ref gh-3527 (Standalone canvas)

Separate canvas and base widget out of Orange3
Using:

https://github.com/biolab/orange-canvas-core/
https://github.com/biolab/orange-widget-base/

This pull request builds on top and supersedes gh-2792

##### Description of changes

* Of the canvas, only the main application entry point remains.

* The workflow model with the base widget is moved to orange-widget-base. This includes some but not all of gui.py, utilits/..., basically what is independent of Table, Variable, ... but the package structure is reconstructed here.

* Orange.data.io.FileFormat is no longer used to registers image writers `Orange.data.io.FileFormat.img_writers` returns empty sequence.

* The Orange.widgets.widget.OWWidget extends OWBaseWidget with Table,Domain,... related report methods.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #3932, opened 2019-07-10
[FIX] main: Fix widget settings directory path to clear

ales-erjavec:
### Issue

Fixes gh-3931

Since gh-3772 the `environ.widget_settings_dir` is no longer the authoritative directory path.

##### Description of changes

* Deprecate `Orange.misc.environ.widget_settings_dir`, (but make it return the correct path).
* Change the application name to 'Orange' to preserve historical directory layout (the application name is used as the path component in the settings path).
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/misc/environ.py
+++ b/Orange/misc/environ.py
@@ -57,15 +57,14 @@ def widget_settings_dir(versioned=True):
     """
     Return the platform dependent directory where widgets save their settings.
 
-    This a subdirectory of ``data_dir(versioned)`` named "widgets"
+    .. deprecated:: 3.23
     """
     warnings.warn(
-        f"'{__name__}.widget_settings_dir' is deprecated. "
-        "Use 'Orange.widgets.settings.widget_settings_dir'",
+        f"'{__name__}.widget_settings_dir' is deprecated.",
         DeprecationWarning, stacklevel=2
     )
-    from Orange.canvas import config
-    return config.widget_settings_dir(versioned)
+    import orangewidget.settings
+    return orangewidget.settings.widget_settings_dir(versioned)
 
 
 def cache_dir(*args):
--- a/Orange/misc/environ.py
+++ b/Orange/misc/environ.py
@@ -9,16 +9,10 @@ $DATA_HOME/Orange/$VERSION/
 
 where DATA_HOME is a platform dependent application directory
 (:ref:`data_dir_base`) and VERSION is Orange.__version__ string.
-
-``canvas`` subdirectory is reserved for settings/preferences stored
-by Orange Canvas
-``widget`` subdirectory is reserved for settings/preferences stored
-by OWWidget
-
 """
-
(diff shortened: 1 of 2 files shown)


lanzagar:
I am still a bit confused. Which is the correct one to use (and why aren't all others deprecated)?

```py
>>> Orange.version.version
'3.23.0.dev0+15d4fe6'
>>> Orange.canvas.config.widget_settings_dir()
'/home/lan/.local/share/Orange/3.23.0.dev/widgets'
>>> Orange.widgets.settings.widget_settings_dir()  # Wrong version?
'/home/lan/.local/share/Orange/0.0.0/widgets'
>>> Orange.misc.environ.widget_settings_dir()
'/home/lan/.local/share/Orange/3.23.0.dev/widgets'
>>> Orange.canvas.utils.environ.widget_settings_dir  # str not function! Is deprecated
'/home/lan/.local/share/Orange/3.23.0.dev/widgets'
```

Orange/misc/environ.py
```diff
@@ -65,7 +59,13 @@ def widget_settings_dir(versioned=True):
 
     This a subdirectory of ``data_dir(versioned)`` named "widgets"
     """
-    return os.path.join(data_dir(versioned=versioned), "widgets")
+    warnings.warn(
+        f"'{__name__}.widget_settings_dir' is deprecated. "
+        "Use 'Orange.widgets.settings.widget_settings_dir'",
```
lanzagar: Should this be `'Orange.canvas.config.widget_settings_dir'`?
And `'Orange.widgets.settings.widget_settings_dir'` removed since it is wrong?

ales-erjavec: `Orange.widgets.settings.widget_settings_dir` is the effective directory (technically it is `orangewidget.settings.widget_settings_dir`.  It will *become* the correct path once the application starts and configures.

```
>>> Orange.widgets.settings.widget_settings_dir()
'/Users/aleserjavec/Library/Application Support/Orange/0.0.0/widgets'
>>> Orange.canvas.config.Config().init()
'/Users/aleserjavec/Library/Application Support/Orange/3.23.0.dev/widgets'
```
This actually has to happen very soon before the widgets are imported.

lanzagar: OK, so after configuration they both return the same correct path. But still, when should one use which
`Orange.canvas.config.widget_settings_dir`
`Orange.widgets.settings.widget_settings_dir`
and could we have just one?

Currently they also have different calls to orangewidget:
`orangewidget.workflow.config.widget_settings_dir`
`orangewidget.settings.widget_settings_dir`
I think the public API of orangewidget should have one path from which to import widget_settings_dir.
If the same thing is available from multiple locations (either in orangewidget or in orange) it just creates confusion and provides opportunities for problems in the future IMO.

PR #3935, opened 2019-07-11
Projection widgets: Reduce number of shapes to 6 (5 + other)

### LINKED ISSUE #3748: Legend and Size setting: don't show more than 5 values
##### Issue
Visualizations enable displaying shapes of points, which are derived from categorical variables. If a categorical variable has many values, strange things begin to happen --> see how shapes represent different values in each plot?

##### Scatter Plot
Screen Shot 2019-04-12 at 12 47 25

##### Linear Projection
Screen Shot 2019-04-12 at 12 47 17

##### Geo Map
Screen Shot 2019-04-12 at 12 35 37

In three different widgets we show three different things. This is confusing and, quite frankly, ugly.

##### Proposed solution
Do not allow selecting a categorical attribute with more than 5 values. Edit Domain now has the option to merge several discrete values into one. ⭐️ 
Another argument for limiting the upper bound to 5 is that one cannot keep 9 shapes in mind while inspecting a plot. Even 5 is a lot.

janezd:
##### Issue

Fixes #3748

##### Description of changes

Projections now show six different shapes; if there are more,  they show five most frequent + "Other".

The first five shapes were, imho, also most visually distinct, I just swapped the sixth (square) with the last (star), which also makes sense because it will represent "Other".

Changes are not backward compatible, but I suppose (ehm, hope) nobody should notice. I prefer having clear arguments than methods with contrieved arguments for backward compatibility when this can be avoided. @VesnaT?

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -371,7 +371,7 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
 
     resolution = 256
 
-    CurveSymbols = np.array("o x t + d s t2 t3 p h star ?".split())
+    CurveSymbols = np.array("o x t + d star ?".split())
     MinShapeSize = 6
     DarkerValue = 120
     UnknownColor = (168, 50, 168)
--- a/Orange/widgets/visualize/tests/test_owprojectionwidget.py
+++ b/Orange/widgets/visualize/tests/test_owprojectionwidget.py
@@ -1,7 +1,7 @@
 # Test methods with long descriptive names can omit docstrings
 # pylint: disable=missing-docstring
 import unittest
-from unittest.mock import patch, Mock
+from unittest.mock import Mock
 
 import numpy as np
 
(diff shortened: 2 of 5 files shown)


Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -371,7 +371,7 @@ def get_size_data(self):
 
     resolution = 256
 
-    CurveSymbols = np.array("o x t + d s t2 t3 p h star ?".split())
+    CurveSymbols = np.array("o x t + d star t2 t3 p h s ?".split())
```
VesnaT: Do we ever need more than 6 symbols. If not, how about removing the rest (apart from '?')?
`CurveSymbols = np.array("o x t + d star".split())`

janezd: I guess this should be safe, yes.

After this, I also changed `MAX_SHAPES = 6` to `MAX_SHAPES = len(OWScatterPlotBase.CurveSymbols) - 1`.

PR #3937, opened 2019-07-13
Rank widget: Allow up to 999 features

### LINKED ISSUE #3934: Rank Widget - Select best ranked attributes
##### Orange version

3.22

##### Change request
When one species the number of best ranked attributes, the counter stops at 100. It would be nice if the counter could go higher (say, 999).

Also, a numerical threshold for the importance would also be convenient.

janezd:
Fixes #3934.


leeherman99:
Thanks for fixing this. I have a very wide data table that I wanted to do some light feature elimination. I suggested 999 just because you were allowing 3 digits - I only needed 120 in my case. My suggestion about an absolute threshold was based on an inspection of the distribution of importance scores - I was basically trimming off a long tail - like using a scree plot.

PR #3943, opened 2019-07-26
Add test and train times to owtestlearners

### LINKED ISSUE #3784: Running times in Test & Score
It would be nice to see running times of algorithms -- both training and learning.

A particular problem that we had was that we used 10 different learning algorithms and Test & Score was really slow. We wanted to know which to remove.

However, this information is also interesting by itself.

AndrejaKovacic:
##### Issue

Fixes #3784

##### Description of changes
Adds train and test times to owtestlearners

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -24,7 +24,7 @@ def _identity(x):
 
 def _mp_worker(fold_i, train_data, test_data, learner_i, learner,
                store_models):
-    predicted, probs, model, failed = None, None, None, False
+    predicted, probs, model, failed, train_time, test_time = None, None, None, False, None, None
     try:
         if not train_data or not test_data:
             raise RuntimeError('Test fold is empty')
--- a/Orange/evaluation/testing.py
+++ b/Orange/evaluation/testing.py
@@ -2,6 +2,7 @@
 # pylint: disable=arguments-differ
 from warnings import warn
 from collections import namedtuple
+from time import time
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


Orange/evaluation/testing.py
```diff
@@ -22,20 +24,27 @@ def _identity(x):
 
 def _mp_worker(fold_i, train_data, test_data, learner_i, learner,
                store_models):
-    predicted, probs, model, failed = None, None, None, False
+    predicted, probs, model, failed, train_time, test_time = None, None, None, False, None, None
     try:
         if not train_data or not test_data:
             raise RuntimeError('Test fold is empty')
+        # training
+        t0 = time()
```
janezd: This is a bit crude. It can work for now, but it would be much better to measure the actual execution time of the thread, if possible.

AndrejaKovacic: In python 3.7 there is [thread_time](https://docs.python.org/3/library/time.html#time.thread_time), in python 3.6 we have [process_time](https://docs.python.org/3/library/time.html#time.process_time). Do you think the latter will do?

janezd:
If I recall correctly (please check), we currently don't use multiple threads here, so process_time should be OK until we require Python 3.7.

@ales-erjavec, does Qt provide a way to measure time spent in a thread?

ales-erjavec:
> does Qt provide a way to measure time spent in a thread?

None that I can find.

AndrejaKovacic:
I am still using time(), since process_time() returns much too large values and i am not sure why.  Process_time is the sum of system and CPU time, current implementation of the widget offloads the entire computation into a separate thread. Any ideas welcome.

thocevar:
I've tried swapping in process_time instead of time and I get pretty much the same values.

Eventually, we will want to use thread_time. If you have two "test and score" widgets and launch evaluations simultaneously, you will get results that are twice as large. We could try to use thread_time and fall back on time in case the former is not available.

AndrejaKovacic:
Try neural network learner with heart disease dataset, cross validation with 10 folds supposedly lasts more than a minute, according to process_time.

thocevar:
`Process_time` can return values that are larger than the wall-clock returned by `time` if a process consists of several threads running in parallel. Maybe the underlying neural network model offloads some computation to separate threads.

I think that from a user's perspective, `time` is ok as it provides an equivalent of using a stopwatch to measure when does the widget finish. We will also avoid explaining the mysteries of time measurements to users filing bug reports.

PR #3946, opened 2019-07-27
[FIX] OWProjectionWidgetBase: Update when domain is changed

### LINKED ISSUE #3944: Scatter Plot axes labels not updated automatically
##### Orange version

3.22.0

##### Expected behavior
Scatter Plot should update axes labels whenever it updates its data.

##### Actual behavior
The axes labels only change when being connected to a new widget. Changing the column names for the input table is not reflected in Scatter Plot.

janezd:
Fixes #3944.

@VesnaT, please confirm this is the right way to do it and that nothing else is needed - also considering possible other similar bugs.

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -440,6 +440,9 @@ class OWScatterPlot(OWDataProjectionWidget):
             self.attr_x, self.attr_y = self.attribute_selection_list[:2]
             self.attribute_selection_list = None
         super().handleNewSignals()
+        if self._domain_invalidated:
+            self.graph.update_axes()
+            self._domain_invalidated = False
         self._vizrank_color_change()
 
     @Inputs.features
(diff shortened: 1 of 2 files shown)


VesnaT:
No, this is not the right way to solve the issue. Setting _invalidated to True, causes recomputation of the embedding (in this - Scatterplot's - case there is no calculation and the replotting is not as wasteful as in other widgets that inherit from OWDataProjectionWidget, but should still be avoided).

1. One of the solutions to the issue, could be to call `OWScatterPlotGraph.update_axes` in `OWScatterPlot.handleNewSignals` (not the best one because the axes are always updated):
```
super().handleNewSignals()
self.graph.update_axes()
self._vizrank_color_change()
```

2. Another one (probably the best one) is to introduce a new flag `self._domain_invalidated ` in `OWDataProjectionWidget` and call the `OWScatterPlotGraph.update_axes` only if it is set to True.
```
self._domain_invalidated = not (
                data_existed and self.data is not None and
                effective_data.domain.checksum()
                == self.effective_data.domain.checksum())
```

3. And if you want a general solution (even though I'd rather not plot data again if not changed):

 ```
def handleNewSignals(self):
        self._handle_subset_data()
        if self._invalidated:
            self._invalidated = False
            self.setup_plot()
        else:
            self.graph.update_coordinates()
            self.graph.update_point_props()
        self.commit()
```

janezd:
My reasoning was that renaming variables is so uncommon that we can afford to update the entire graph. Still, your solutions are better. I would prefer the first one; updating axes costs nothing and simplicity counts. But I implemented the second because other widgets may also need to know that domain has changed.

PR #3951, opened 2019-07-30
[FIX] Test and Score: Sort numerically, not alphabetically

### LINKED ISSUE #3947: Bug: Test and Score Evaluation results sort alphabetically
##### Orange version

3.22

##### Expected behavior
In the Test&Score module window, there are five columns by default: Model, MSE, RMSE, MAE & R2.  All but the first column contains numeric values.  It is possible to click on any of the column headers to sort the table by that column.  Expected behavior:  sorting by the numeric columns will use numeric sort behavior.

##### Actual behavior
Clicking on a numeric column header sorts with alphabetic/lexicographic behavior, so a possible "sorted" list might look like:
16992
178
355
4536

1. Associate multiple model modules with a Test and Score module
2. Once the Test and Score module results window
3. Click on a column header to sort by a numeric column.

In the attached screenshot, the table is sorted by column 3, RMSE, alphabetically ascending:
12630
1389
20510
2194779950371
37742
screenshot_200

For sorting to be useful, these values should be sorted:

1389
12630
20510
37742
2194779950371

janezd:
Fixes #3947 by using a QSortFilterProxy that sorts numerically, if column values can be converted to numbers, or case-insensitive, if they can't.

##### Includes
- [X] Code changes
- [ ] Tests

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -533,7 +533,7 @@ class OWTestLearners(OWWidget):
                 for stat, scorer in zip(stats, self.scorers):
                     item = QStandardItem()
                     if stat.success:
-                        item.setText("{:.3f}".format(stat.value[0]))
+                        item.setData(float(stat.value[0]), Qt.DisplayRole)
                     else:
                         item.setToolTip(str(stat.exception))
                         if scorer.name in self.score_table.shown_scores:
--- a/Orange/widgets/evaluate/tests/test_owtestlearners.py
+++ b/Orange/widgets/evaluate/tests/test_owtestlearners.py
@@ -277,10 +277,10 @@ class TestOWTestLearners(WidgetTest):
         # Ensure that the click on header caused an ascending sort
         # Ascending sort means that wrong model should be listed first
         self.assertEqual(header.sortIndicatorOrder(), Qt.AscendingOrder)
-        self.assertEqual(view.model().item(0, 0).text(), "VersicolorLearner")
+        self.assertEqual(view.model().index(0, 0).data(), "VersicolorLearner")
 
         self.send_signal(self.widget.Inputs.test_data, versicolor, wait=5000)
-        self.assertEqual(view.model().item(0, 0).text(), "SetosaLearner")
+        self.assertEqual(view.model().index(0, 0).data(), "SetosaLearner")
 
         self.widget.hide()
 
@@ -365,10 +365,11 @@ class TestOWTestLearners(WidgetTest):
                 [1, 1, 1.23, 23.8, 5.], [1., 2., 3., 4., 3.], "yynnn"))
         )
 
-        self.assertTupleEqual(self._test_scores(
-            table_train, table_test, LogisticRegressionLearner(),
-            OWTestLearners.TestOnTest, None),
-                              (0.667, 0.8, 0.8, 0.867, 0.8))
+        np.testing.assert_almost_equal(
+            self._test_scores(table_train, table_test,
+                              LogisticRegressionLearner(),
+                              OWTestLearners.TestOnTest, None),
+            (2 / 3, 0.8, 0.8, 13 / 15, 0.8))
 
     def test_scores_cross_validation(self):
         """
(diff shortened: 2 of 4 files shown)


ales-erjavec:
Or maybe just undo b4196790548a57ff4de9e52c2a7f87a4b9c0ad4f and go on from there.

```diff
diff --git a/Orange/widgets/evaluate/owtestlearners.py b/Orange/widgets/evaluate/owtestlearners.py
index 139f2cb9c..9a2d77de9 100644
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -533,7 +533,7 @@ class OWTestLearners(OWWidget):
                 for stat, scorer in zip(stats, self.scorers):
                     item = QStandardItem()
                     if stat.success:
-                        item.setText("{:.3f}".format(stat.value[0]))
+                        item.setData(float(stat.value[0]), Qt.DisplayRole)
                     else:
                         item.setToolTip(str(stat.exception))
                         if scorer.name in self.score_table.shown_scores:
diff --git a/Orange/widgets/evaluate/utils.py b/Orange/widgets/evaluate/utils.py
index ebe060327..6caa6811b 100644
--- a/Orange/widgets/evaluate/utils.py
+++ b/Orange/widgets/evaluate/utils.py
@@ -109,6 +109,12 @@ class ScoreTable(OWComponent, QObject):
             size = super().sizeHint(*args)
             return QSize(size.width(), size.height() + 6)
 
+        def displayText(self, value, locale):
+            if isinstance(value, float):
+                return "{:.3f}".format(value)
+            else:
+                return super().displayText(value, locale)
+
     def __init__(self, master):
         QObject.__init__(self)
         OWComponent.__init__(self, master)
```

janezd:
Thanks @ales-erjavec. I reverted the changes in https://github.com/biolab/orange3/commit/b4196790548a57ff4de9e52c2a7f87a4b9c0ad4f, as you proposed.

I still kept a (now simpler) proxy because I like putting the failed tests at the bottom.

thocevar:
Sorting has a problem with `np.nan` because it is a float.

A stable sort would be nice for sorting by multiple criteria (e.g. if the first criteria reaches the max value). Sorting by name and then by some constant column does not retain the alphabetical order. I don't know why this doesn't work out of the box.

This would also solve the problem with a changing order of failed tests. If you have several, they change positions when going from ascending to descending (but stay in place from descending to ascending).

janezd:
> Sorting has a problem with `np.nan` because it is a float.

There should be no nans in this table. If there are (but I haven't seen them yet), this is a different bug; they have to be replaced by errors by the widget.

> A stable sort would be nice for sorting by multiple criteria (e.g. if the first criteria reaches the max value).

Sorting is done by [`QSortFilterProxyModel`](https://doc.qt.io/qt-5/qsortfilterproxymodel.html). My class provides just the comparison, I can't control the algorithm.

> This would also solve the problem with a changing order of failed tests.

It would be nice to keep the order, but it's not such a problem. :)

thocevar:
> There should be no nans in this table. If there are (but I haven't seen them yet), this is a different bug; they have to be replaced by errors by the widget.

True, but as it is a part of the utils, it would be nice if it handled this additional case (besides "" and None) that might come up somewhere else. Actually, anything that is not a float could fall into the same bottom category.

> Sorting is done by [`QSortFilterProxyModel`](https://doc.qt.io/qt-5/qsortfilterproxymodel.html). My class provides just the comparison, I can't control the algorithm.

I've just checked that `QSortFilterProxyModel` behaves as a stable sort in the Predictions widget. Now I can't find the case, where it doesn't, so maybe it does. :)

janezd:
> > There should be no nans in this table.
> 
> True, but as it is a part of the utils, it would be nice if it handled this additional case (besides "" and None) that might come up somewhere else. Actually, anything that is not a float could fall into the same bottom category.

`ScoreModel` is supposed to contain scores, and these cannot (should not) be nan. But you're right: it's better to support them to avoid problems in the future and particularly to avoid bugs if somebody copies this code or moves it to some more general utils module. Done. The changed code is also cleaner.

PR #3954, opened 2019-07-31
MySQL database support

mviana:
First Commit
inluded support for mysql in sql data

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/sql/mysql.py
+++ b/Orange/data/sql/mysql.py
@@ -1,0 +1,288 @@
+import logging
+from contextlib import contextmanager
+
+from Orange.util import Registry
+import mysql.connector as mysql
+
+from Orange.data import StringVariable, TimeVariable, ContinuousVariable, DiscreteVariable
+from Orange.data.sql.backend import Backend
+from Orange.data.sql.backend.base import ToSql, BackendError, TableDesc
+
+log = logging.getLogger(__name__)
+
+class PymysqlBackend(Backend):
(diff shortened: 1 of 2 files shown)


ajdapretnar:
Thank you very much for this contribution!

Would you care adding a few tests to ensure everything is a-ok? You can get a hint of what tests should look like in Orange/tests/sql/base.py (DBTestConnection class and its subclasses) and in the other scripts in the sql folder.

Orange/data/sql/mysql.py
```diff
@@ -0,0 +1,288 @@
+import logging
+from contextlib import contextmanager
+
+from Orange.util import Registry
+import mysql.connector as mysql
```
ajdapretnar: You also have to add the requirement to https://github.com/biolab/orange3/blob/master/requirements-sql.txt

mviana: Done. Still pending tests!

ajdapretnar:
@mviana Thank you for submitting the pull request. We are unfortunately closing it, since the widget doesn't work (see @lanzagar's comments) and there are no appropriate tests in place. If you feel like still working on this, feel free to reopen the PR and submit the changes.

lanzagar:
To add to this: we know many people would like to connect to other databases (like mysql, but also others) and are more and more convinced that it would be worth it to add odbc support and cover all of them instead of adding support one by one. Looks like it could happen soon (fingers crossed) :)

PR #3957, opened 2019-08-01
[FIX] Fix storing and retrieving selection, and unconditional auto commit

### LINKED ISSUE #3762: Widgets do not remember a selection for a data set
##### Orange version

3.21-dev

##### Expected behavior

Widgets should remember what data were selected for the dataset.

##### Actual behavior

Widgets are not consistent. Some widgets do not remember what data were selected before.
Those widgets are:

- [x] Mosaic Display
- [x] Silhouette Plot 
- [x] Distance Map

See the comment below for:

- [ ] Confusion Matrix
- [ ] tSNE
- [ ] Hierarchical Clustering

The same issue also appears for widgets reported in the following issue https://github.com/biolab/orange3/issues/3763 in a combination with auto apply issue.

Connect the widget's input to the data source (e.g. File Widget) set dataset to `Iris`. Select part of data in widget. Change data to some other dataset and change data back to `Iris`. The selection for `Iris` should be still present but it is not.

### LINKED ISSUE #3763: Widgets don't Apply, Send or Run when their input signals changed and widgets do not remember selected subset
##### Orange version

3.21-dev

##### Expected behavior

According to #3455 and #3755, widgets should Apply, Send or Run when their input signals changed even if auto-apply/send/run is disabled.
Widgets should also remember what data were selected for the dataset.

##### Actual behavior

Widgets are not consistent. Some widgets do not Apply, Send or Run when the input signal change and at the same time widgets do not remember what data were selected before.
Those widgets are:

- [ ] Line Plot
- [ ] Heat Map
- [ ] Data Table
- [ ] Venn Diagram

... and the following projections widgets:

- [ ] Scatter Plot
- [ ] FreeViz
- [ ] Linear Projection
- [ ] RadViz
- [ ] MDS
- [ ] tSNE (apply works)

For selection: 
Connect the widget's input to the data source (e.g. File Widget) set dataset to `Iris`. Select part of data in widget. Change data to some other dataset and change data back to `Iris`. The selection for `Iris` should be still present but it is not. 

For apply: 
Uncheck apply automatically checkbox and change the input data. Observe if output data changes (widget do something with the input).

##### Additional info

Venn Diagram does not remove selection when different dataset at the input.

janezd:
Fixes #3762. Fixes #3763.

Following @markotoplak suggestion, instance selections are schema-only because data is not a part of the context.

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -317,6 +317,7 @@ class OWMosaicDisplay(OWWidget):
         self.discrete_data = None
         self.subset_data = None
         self.subset_indices = None
+        self.__pending_selection = None
 
         self.color_data = None
 
@@ -442,6 +443,7 @@ class OWMosaicDisplay(OWWidget):
 
         self.init_combos(self.data)
         self.openContext(self.data)
+        self.__pending_selection = self.selection
 
     @Inputs.data_subset
     def set_subset_data(self, data):
@@ -475,6 +477,11 @@ class OWMosaicDisplay(OWWidget):
     def reset_graph(self):
         self.clear_selection()
         self.update_graph()
+        if self.__pending_selection is not None:
+            self.selection = self.__pending_selection
+            self.__pending_selection = None
+            self.update_selection_rects()
+            self.send_selection()
 
     def set_color_data(self):
         if self.data is None:
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -290,7 +290,7 @@ class OWMosaicDisplay(OWWidget):
     variable3 = ContextSetting(None)
     variable4 = ContextSetting(None)
     variable_color = ContextSetting(None)
-    selection = ContextSetting(set())
+    selection = Setting(set(), schema_only=True)
 
     BAR_WIDTH = 5
     SPACING = 4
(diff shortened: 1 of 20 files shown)


janezd:
On Travis on Python 3.7 this consistently core dumps in Relief, which should be unrelated to these changes. Help!

Orange/widgets/visualize/owmosaic.py
```diff
@@ -442,6 +443,7 @@ def set_data(self, data):
 
         self.init_combos(self.data)
         self.openContext(self.data)
+        self.__pending_selection = self.selection
```
VesnaT: Should'n this be a 'schema only' setting

Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -76,6 +76,8 @@ class Outputs:
     add_scores = settings.Setting(False)
     auto_commit = settings.Setting(True)
 
+    pending_selection = settings.ContextSetting(None, schema_only=True)
```
VesnaT: I think we have not handled 'schema only' context settings, so setting the flag to True has no effect.

Orange/widgets/visualize/owmosaic.py
```diff
@@ -462,8 +462,14 @@ def handleNewSignals(self):
...
                 self.subset_indices = [ex.id in indices for ex in self.data]
+        if self.data is not None and self.__pending_selection is not None:
+            self.selection = self.__pending_selection
+            self.__pending_selection = None
+        else:
+            self.selection = set()
         self.set_color_data()
-        self.reset_graph()
+        self.update_graph()
+        self.send_selection()
```
janezd: @VesnaT, copying from projection widget indeed helped (and you were right about not using `pack_settings`, I should have listened more carefully and be less self confident :)).

However, every widget has also its own tricks, so you can't *just copy*. In this case, it was important to call `send_selection` after `update_graph`.

PR #3960, opened 2019-08-02
Create FUNDING.yml

ajdapretnar:
Add donation support for Orange.


borondics:
@markotoplak we should follow how this gets propagated into other "flavored" Oranges, like Quasar. It would be unfair to count funding to those from the main Orange project.

ajdapretnar:
@borondics This is just a Sponsor button for Github. You can add one for Quasar if you like.

PR #3979, opened 2019-08-21
Fix numpy version on 1.16.*

PrimozGodec:
##### Issue

AppVeyor fails on openTSNE import - reason numba issue.

##### Description of changes

With this PR we fixed the numpy version on 1.1.6.* It can be undone when numba resolves an issue.

##### Includes
- [ ] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ environment:
     BUILD_GLOBAL_OPTIONS: build -j1
     BUILD_ENV: wheel==0.29.0 pip==9.0.1 numpy==1.13.* -r requirements-doc.txt
     # SIP 4.19.4+ with PyQt5==5.9.1+ segfault our tests (GH-2756)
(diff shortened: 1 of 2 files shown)


janezd:
Seem OK. May I (change the title and) merge?

PrimozGodec:
This PR changes just the numpy version for tests I think it is wort test if this is only the issue of AppVeyor or issue of any Windows installation. 

I mean if AppVeyor fails there is a chance that some users have a Windows machine where this will happen to.

PrimozGodec:
I suggest fixing the numpy version for Orange until the issue is fixed.

janezd:
Should we have ==1.6.* (as proposed in this PR) or <1.17.*? Currently it's ==1.14, and 1.14 is still OK.

PrimozGodec:
There is also a similar issue with numpy-1.14.6 (this is what AppVeyor is currently using) 
One of the logs: https://ci.appveyor.com/project/biolab/orange3-75xll/builds/26879946?fullLog=true

lanzagar:
OK, let's fix it to 1.16 then and hope it is soon fixed upstream and we can remove the restriction.

PR #3980, opened 2019-08-22
[FIX] Heat map: Cannot cluster a single instance

### LINKED ISSUE #3975: Heat Map: Row clustering fails on class value with a single instance
**Describe the bug**
Heat Map crashes with
`ValueError: The number of observations cannot be determined on an empty distance matrix.`

**To Reproduce**
File (load test data from below) - Heat Map (Cluster Rows).

**Orange version:**
3.23.dev

**Expected behavior**
Heat Map doesn't crash.

**Additional context**
Test data.

c1 | c2
-- | --
1 | 6
2 | 5

VesnaT:
##### Issue

Fixes #3975

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owheatmap.py
+++ b/Orange/widgets/visualize/owheatmap.py
@@ -318,11 +318,11 @@ class RowPart(RowPart):  # pylint: disable=function-redefined
                                cluster, cluster_ordered)
 
     @property
-    def is_empty(self):
+    def can_cluster(self):
         if isinstance(self.indices, slice):
-            return (self.indices.stop - self.indices.start) == 0
+            return (self.indices.stop - self.indices.start) > 1
         else:
-            return len(self.indices) == 0
+            return len(self.indices) > 1
 
     @property
     def cluster_ord(self):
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owheatmap.py
```diff
@@ -320,9 +320,9 @@ def __new__(cls, title, indices, sortindices, cluster=None,
     @property
     def is_empty(self):
         if isinstance(self.indices, slice):
-            return (self.indices.stop - self.indices.start) == 0
+            return (self.indices.stop - self.indices.start) < 2
         else:
-            return len(self.indices) == 0
+            return len(self.indices) < 2
```
lanzagar: The name of the property becomes a bit awkward (=wrong) after this fix, since it does not check if the group is empty anymore, but if it contains 0 or 1 items.
However, it is only used in one place and apart from the exotic definition of "empty" it works as it should.
I am fine merging this as is.
Possible fixes (if you feel it is worth it) would be for example
- inverting the property to `has_multiple_items` (can't think of a better name) and dropping the `not` where it is called
- remove the property and implement `__len__` instead and check that it is `< 2` in the `cluster_rows` method (only use of the property)

PR #3982, opened 2019-08-23
Boxplot: Show chi2 statistics

### LINKED ISSUE #3971: Add chi**2 to boxplot when multiple categorical features are selected
Nope not really, just an enhancement.

The box plot widget outputs student's t-test or ANOVA statistics, including p-values, when selecting a categorical and numerical feature. I really like that; eyeballing is great and all but having the test stats so easily available really enhances the experience. In the same way that that's useful when comparing group means, I think that a Chi**2 statistic would be useful when comparing two categorical features in the same widget. For example using the [census](https://archive.ics.uci.edu/ml/datasets/adult) dataset, we can compare Income (the target) to the "workclass" feature and get this output in the boxplot widget:

image

We can calculate the chi**2 statistic and p-value for these 2 features with scipy.stats.chi2_contingency:

```Python
import scipy.stats as ss
import pandas as pd

# instantiate the census data into a dataframe called df

ss.chi2_contingency(pd.crosstab(df.workclass, df.income).values)
```
Output is chi2 statistic: 1045.7085997281692 and p-value: 2.026505431120716e-220

I think it'd be nice to have those printed on the bottom of the box plot widget when comparing categorical features.

Calculating it isn't especially arduous so that alternative exists. This is definitely a _nice to have_.

**Additional context**
None necessary I think.

janezd:
##### Issue

Fixes #3971.

##### Description of changes

@ajdapretnar, please do some clicking. It seems to work and it should pass the test, but this widget code is rather contrived, so I wasn't too comfortable changing it.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -11,7 +11,7 @@ from AnyQt.QtGui import QPen, QColor, QBrush, QPainterPath, QPainter, QFont
 from AnyQt.QtCore import Qt, QEvent, QRectF, QSize
 
 import scipy.special
-from scipy.stats import f_oneway, chisquare
+from scipy.stats import f_oneway, chi2_contingency
 
 import Orange.data
 from Orange.data.filter import FilterDiscrete, FilterContinuous, Values
@@ -129,8 +129,8 @@ class OWBoxPlot(widget.OWWidget):
     when the user changes display settings or colors.
 
     For discrete attributes, the flow is a bit simpler: the elements are not
-    constructed in advance (by layout_changed). Instead, layout_changed and
-    display_changed call display_changed_disc that draws everything.
+    constructed in advance (by layout_changed). Instead, _display_changed_disc
+    draws everything.
     """
     name = "Box Plot"
     description = "Visualize the distribution of feature values in a box plot."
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owboxplot.py
```diff
@@ -738,6 +742,16 @@ def stat_ANOVA():
...
 
+    def _compute_tests_disc(self):
+        if self.group_var is None or self.attribute is None:
+            self.stat_test = ""
+        else:
+            chi, p, dof = self._chi_square(self.group_var, self.attribute)
+            if np.isnan(p):
+                self.stat_test = ""
+            else:
+                self.stat_test = f"χ²: {chi:.2f} (p={p:.3f}, dof={dof})"
```
ajdapretnar: I like that this is shown below, but after clicking 'Stretch bars', it disappears and never shows up again (unless attribute is changed).

Orange/widgets/visualize/owboxplot.py
```diff
@@ -378,6 +364,26 @@ def compute_score(attr):
...
+        if min(observed.shape) < 2:
+            return 0, 2, 0
+        expected = \
+            np.outer(observed.sum(axis=1), observed.sum(axis=0)) / \
+            np.sum(observed)
+        dof = (len(group_var.values) - 1) * (len(attr.values) - 1)
+        return chisquare(
+            observed.ravel(), f_exp=expected.ravel(),
+            ddof=sum(observed.shape) - 2
+        ) + (dof, )
```
thocevar: We could replace the computation of expected values and the return by `return scipy.stats.chi2_contingency(observed)[:3]`, which does the same thing (https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.chi2_contingency.html).

PR #3986, opened 2019-08-27
Concatenation of BoW with Corpus

AndrejaKovacic:
##### Issue

Fixes  https://github.com/biolab/orange3-text/issues/243

##### Description of changes
We concluded that bag of words can be concatenated only with other Bow, so error is issued if the user tries to concatenate anything else.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owconcatenate.py
+++ b/Orange/widgets/data/owconcatenate.py
@@ -153,11 +153,11 @@ class OWConcatenate(widget.OWWidget):
             del self.more_data[sig_id]
 
     def handleNewSignals(self):
-        self.Error.clear()
         self.mergebox.setDisabled(self.primary_data is not None)
         if self.incompatible_types():
             self.Error.bow_concatenation()
         else:
+            self.Error.bow_concatenation.clear()
             self.unconditional_apply()
 
     def incompatible_types(self):
--- a/Orange/widgets/data/owconcatenate.py
+++ b/Orange/widgets/data/owconcatenate.py
@@ -20,7 +20,7 @@ from Orange.widgets.settings import Setting
 from Orange.widgets.utils.annotated_data import add_columns
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.utils.widgetpreview import WidgetPreview
-from Orange.widgets.widget import Input, Output
+from Orange.widgets.widget import Input, Output, Msg
 
 
 class OWConcatenate(widget.OWWidget):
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owconcatenate.py
```diff
@@ -152,8 +155,34 @@ def set_more_data(self, data=None, sig_id=None):
             del self.more_data[sig_id]
 
     def handleNewSignals(self):
+        self.Error.clear()
         self.mergebox.setDisabled(self.primary_data is not None)
-        self.unconditional_apply()
+        incompatible = self.incompatible_types()
+        if incompatible:
```
ajdapretnar: Probably just `if self.incompatible_types():` ?

AndrejaKovacic: Yes, thank you, i refactored code and then missed this.

ajdapretnar:
This does not fix the issue, though. Try this (with the same data set):
Screen Shot 2019-08-28 at 10 41 12

ajdapretnar:
While the above issue should be fixed, for incompatbility check I managed to find this:
```
data = [self.more_data[key].is_sparse() for key in self.more_data
                if self.more_data[key].X.size != 0]
        if self.primary_data is not None and self.primary_data.X.size != 0:
            data.append(self.primary_data.is_sparse())
        return not data.count(data[0]) == len(data) if len(data) else False
```

Orange/widgets/data/tests/test_owconcatenate.py
```diff
@@ -17,6 +18,10 @@
 
 class TestOWConcatenate(WidgetTest):
 
+    class DummyTable(Table):
```
ajdapretnar: Is this supposed to be here?
If yes, I am very curious as to what it does.

AndrejaKovacic: It is used to test mismatching inputs, in line 133.  Both Corpus and DummyTable extend Table class, so we are testing the same thing, without the assumption of text addon being present.

ajdapretnar: I am blind apparently. I did not see L133. Cool stuff! 😎

ajdapretnar:
This still doesn't work as it should, sorry. 😢 

Try connecting Corpus + Bag of Word to Concatenate and File, as above. The error appears. Then press 'Intersection' instead of 'Union'. Error still there, all good. Then press 'Union' again. Now the widget crashes as before.

While at it, I would extend the error a bit. I like them short, but here, for clarity, I think we should let the user know what is going on. Sth along the lines of "Incompatible types. Can't concatenate Table and Corpus." What do you think?

PR #3993, opened 2019-08-30
New Icons for the Website

ajdapretnar:
##### Issue

Icons manually cut and edited for Orange website.

##### Description of changes
Use new automatically rendered icons.

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation


lanzagar:
Before merging this, let's double check what our plans are for these icons.
As I understand, they are not used in Orange, but by the website for the widget catalog docs. If that is the case, maybe they should be generated and commited in the website repository instead of here?

markotoplak:
Why don't we generate this when we are building the website? Hugo websites can be build with travis and uploaded anywhere.

ajdapretnar:
> Why don't we generate this when we are building the website?
Because nobody is yet doing it. I'd be ecstatic to have a website that auto-generates icons when new docs are available. There are scripts for auto-generating icons and for batch indexing them available. Help appreciated.

PR #4003, opened 2019-09-04
Update requirements

PrimozGodec:
##### Issue

In #3979 we fixed numpy requirement to 1.16.* because of conflicts with openTSNE (numba). Keeping requirements so hardly defined is not nice practice and package with those requirements cannot be released.

##### Description of changes

- We are changing those requirements back to old values.
- We limit the openTSNE version to >=3.11.0, older versions were built with newer numpy(1.16) and those not compatible with older numpy versions.
- We fix a test which does not work in numpy 1.17

##### Includes
- [ ] Code changes
- [ ] Tests
- [ ] Documentation


pavlin-policar:
This will probably also need to update the minimum required version of opentsne to >=0.3.11, where the wheels are compiled against an older version of numpy

PrimozGodec:
@pavlin-policar done.

PrimozGodec:
It is ready to merge now @markotoplak

Issue #3991, opened 2019-08-30
Venn Diagram should work on features


BlazZupan:
Current Venn Diagram widget operates on data instances and shows overlap in data instances for the input data sets. Here we propose to extend this functionality to features. In Orange, several widgets support feature selection and preprocessing, and would be exciting to analyze the similarities and differences of such procedure.

Redesign Venn Diagram according to the proposed sketch. The default selections are marked.

![ven-diagram-redesigned](https://user-images.githubusercontent.com/726604/64009782-c8cc5a00-cb18-11e9-9efa-aa73ac79d1b4.png)

* When "Rows", the domain of the data sets should match entirely, in features, classes, and meta-features. If it does not, then display error "Input data domains do not match."
* When "Columns", the set of instances should be equal (sets of instance IDs are the same). If not, error "Data sets do not contain the same instances."
* "Use identifier" shows a list of string-typed meta-features with the same name present in all input data sets. If there is no such feature, "Use identifier" is disabled, "Use instance equality" is checked and the list box is empty.
* If "Rows", the output of the widget are rows that correspond to instances selected in the Venn diagram.
* When comparing columns, use column identifier (not names).

ajdapretnar:
I would flip the order of Elements box. Rows seems to be (and should be) the default, hence I'd have it listed first. Then Columns.

janezd:
Implemented via #4006.

PR #4006, opened 2019-09-05
[ENH] Venn Diagram: Add relations over columns, simplify over rows

AndrejaKovacic:
##### Issue

Implements  #3991. 

##### Description of changes
Venn diagram over rows is now simpler. Venn diagram over columns is now possible. I would appreciate comments about output, especially when using columns.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -151,69 +170,81 @@ def __init__(self):
...
+        self.Error.domain_mismatch.clear()
+        self.Error.instances_mismatch.clear()
+        if not self.usecols:
+            domains = [input.table.domain for input in self.data.values()]
+            samedomain = all((d1 == d2) for d1, d2 in pairwise(domains))
+            if not samedomain:
+                self.vennwidget.clear()
+                self.Error.domain_mismatch()
+                self.itemsets = OrderedDict()
+                return False
```
janezd: These four lines repeat again below. You can extract this into a function (including `return False`) and replace these lines with `return name_of_that_function()`.

Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -105,27 +125,26 @@ def __init__(self):
...
-            model = itemmodels.VariableListModel(parent=self)
-            cb = gui.OrangeComboBox(
-                minimumContentsLength=12,
-                sizeAdjustPolicy=QComboBox.AdjustToMinimumContentsLengthWithIcon
+        box = gui.vBox(self.inputsBox, 'feature', addSpace=False)
+        box.setFlat(True)
+        model = itemmodels.VariableListModel(parent=self)
+        self.cb = cb = gui.OrangeComboBox(
+            minimumContentsLength=12,
+            sizeAdjustPolicy=QComboBox.AdjustToMinimumContentsLengthWithIcon
```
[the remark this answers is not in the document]
AndrejaKovacic: I agree, this is clearer. The referenced issue in this pull request has a sketch by @BlazZupan with "Use identifiers", but i think that's because of historical reasons.

janezd:
TL;DR: Checkout my commit and see what I've done. :|

I'm sorry... There was a huge gap between the last radio button and the combo. I tried removing it and noticed there's an unnecessary box. And that the widget uses attribute indices althought the combo box uses a model. After fixing this, I realized that radio buttons are not necessary if we just use "Instance identity" as a placeholder for None. So, changing this, I noticed that the combo is disabled unless the diagram shows rows, hence, the combo actually belongs to one of the options in the other radio button group. This made the control area almost empty, in particular when I removed the empty and useless info box. So we arrived to what I've just committed.

janezd:
Screenshot 2019-11-14 at 23 09 57

janezd:
@AndrejaKovacic, in your last commit you removed context settings. You decided you didn't like them? :)

(I've modified the commit to put them back.)

Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -479,41 +344,21 @@ def _on_selectionChanged(self):
...
+        self.selection = [i for i, area in enumerate(areas) if area.isSelected()]
         self.invalidateOutput()
 
-    def _on_useidentifiersChanged(self):
-        self.inputsBox.setEnabled(self.useidentifiers == 1)
-        # Invalidate all itemsets
-        self._invalidate()
-        self._updateItemsets()
+    def _on_matching_changed(self):
+        self.output_duplicates_box.setEnabled(bool(self.rowwise))
```
lanzagar: Errrr, I don't get why the whole Output box would be disabled for Columns - including autocommit!
Probably only the `Output duplicates` option should be?

Edit: maybe this bug also hints at non-optimal naming - `output_duplicates_box` could be just `output_box` (duplicates setting is only one part of it).

Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -137,221 +100,163 @@ def __init__(self):
...
+        return False
+
+    def settings_compatible(self):
+        self.Error.domain_mismatch.clear()
+        self.Error.instances_mismatch.clear()
+        if self.rowwise:
+            domains = [input.table.domain for input in self.data.values()]
+            self.samedomain = all((d1 == d2) for d1, d2 in pairwise(domains))
+            if not self.samedomain:
+                return self.settings_incompatible()
```
lanzagar: The domains do not need to match. That is why there is an option to change `Rows, matched by:` from `Instance identity` (where it makes sense to match domains first) to a single variable with unique identifiers (matching is done only on that attribute, the rest of the domains can differ).

To test use the zoo dataset, connect to Venn. Also connect to PCA->Scatter plot, select some points and connect to Venn. Venn should be able to match on attribute `Name`.

Orange/widgets/visualize/owvenndiagram.py
```diff
@@ -687,7 +596,7 @@ def __hash__(self):
     def __eq__(self, other):
         # XXX: comparing NaN with different payload
         return (isinstance(other, ComparableInstance)
-                and domain_eq(self.domain, other.domain)
+                and (self.domain == other.domain)
```
lanzagar: This now compares metas as well (before we didn't). Intentional?

lanzagar:
The mentioned problems seem to be fixed. I found a new one, and the documentation update is still needed (can be done last).
- [ ] Instance identity used to ignore metas, now it uses them as well (can cause problems when a Selected column was added etc.). I don't think this was intentional?
- [ ] Update documentation

Orange/widgets/visualize/owvenndiagram.py
AndrejaKovacic: It was requested in the original issue, that domains match completely, when doing comparisons over rows. This is a remnant of that, but as we later decided against the idea, i should have  reverted it. Good catch.

PR #4011, opened 2019-09-06
[FIX] Correlations: Add progress bar, retain responsiveness

### LINKED ISSUE #4008: Correlations makes Orange unresponsive
**Describe the bug**
When given data with many features, Correlations becomes unresponsive after a minute or two of runtime.

**To Reproduce**
Take GDS100 data set from GEO Data Sets (bioinformatics add on), samples in rows.

image

image

Correlations widget is responsive at the start, but then freezes Orage. Clicking on Pause has no effect.

**Expected behavior**
The widget should be responsive and also show the progress bar, even when there are many features on the input.

VesnaT:
##### Issue

Fixes #4008

Since adding the progress bar was somehow trivial, making the widget responsive was not and I'm still not sure about the solution.

The widget was getting unresponsive for datasets with small number of instances and many features. In those cases correlation between a pair of features was calculated 'too fast' which caused a signal, for setting partial results, being emitted too frequently. Consequently the widget was getting unresponsive.

##### Description of changes
To prevent the congestion, the signal is only emitted every 0.01 second.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcorrelations.py
+++ b/Orange/widgets/data/owcorrelations.py
@@ -199,8 +199,7 @@ class CorrelationRank(VizRankDialogAttrPair):
     def start(self, task, *args, **kwargs):
         self.__set_state_ready()
         super().start(task, *args, **kwargs)
-        self.master.progressBarInit()
-        self.master.setBlocking(True)
+        self.__set_state_busy()
 
     def cancel(self):
         super().cancel()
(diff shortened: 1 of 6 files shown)


VesnaT:
The number of states is so small (all the scores are computed within a time period), the tests could actually work.

PR #4039, opened 2019-09-20
Elide long text files

### LINKED ISSUE #4019: Data Table: showing text files is slow
**Describe the bug**
When Data Table is fed longer text files, showing the data table gets really slow.

**To Reproduce**
1. Load any corpus, say *bookexcerpts*.
2. Connect a Data Table and open it.

**Orange version:**
3.23.dev

**Expected behavior**
Data Table shows data immediately / on-the-fly.

**Operating system:**
OSX

ajdapretnar:
##### Issue

Fixes #4019

##### Description of changes
Elide string variables to 100 chars.

Timing on a data set with long text:

Before fix 0.9463250637054443
After fix 0.9104220867156982

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -28,6 +28,7 @@ from AnyQt.QtCore import pyqtSlot as Slot
 import Orange.data
 from Orange.data.storage import Storage
 from Orange.data.table import Table
+from Orange.data import StringVariable
 from Orange.data.sql.table import SqlTable
 from Orange.statistics import basic_stats
 
@@ -55,6 +56,7 @@ class RichTableModel(TableModel):
 
         self._header_flags = RichTableModel.Name
         self._continuous = [var.is_continuous for var in self.vars]
+        self._string = [isinstance(var, StringVariable) for var in self.vars]
         labels = []
         for var in self.vars:
             if isinstance(var, Orange.data.Variable):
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owtable.py
```diff
@@ -76,6 +79,9 @@ def data(self, index, role=Qt.DisplayRole,
                 return (val - dist.min) / (dist.max - dist.min)
             else:
                 return None
+        if role == _Qt_DisplayRole and self._string[index.column()]:
+            val = super().data(index, TableModel.ValueRole)
```
janezd: I think it would be clearner if you called super with `_Qt_DisplayRole`. (And then don't cast to string.)

If I overlooked something and you need `TableModel.ValueRole`, then add `_ValueRole` to arguments, (see `_BarRole`).

janezd:
Are these benchmarks wrong, or do you notice 30 ms speed up, or are you withdrawing the pull request? :)

ajdapretnar:
I called the benchmarks at the beginning and at the end of set_dataset().

janezd:
> I called time() at the beginning and at the end of set_dataset().

I don't understand. Does it shave off only 30 ms or does it do more?

ajdapretnar:
Shaves off.

janezd:
Closing, but we'll reopen if @ajdapretnar replicates the case in which this speed-up is needed and demonstrates that these changes indeed work.

Issue #3822, opened 2019-05-28
Color node forgets color values on saving


iNLyze:
##### Orange version

3.19.0

##### Expected behavior
Color node saves all RGB values for colors once set with the entire workflow. 

##### Actual behavior
After closing orange and reloading an existing workflow containing a color node all RGB values for all colors are replaced by default value. 

##### Steps to reproduce the behavior
Load a table
Select columns
Change default colors in color node
Save workflow
Close Orange
Open Orange
Load workflow
Open color node and find default colors

TomFreudenberg:
I am working with 3.20.1 and it has still the same anoying behavior.

Everytime after pushing "Reload" on my larger (320.000 records) dataset
all color settings are reset to their defaults.

Does not make fun to work with colors currently

Cheers
Tom

TomFreudenberg:
Just found a solution to go with for me.

Open the Color Widget
Disable "Automatic Reply"
Leave Color Widget dialog opened
Reload Dataset
Switch back to Color Widget dialog
The "Apply" button is now enabled but the colors still remain
After pressing "Apply" Button coloring is as before

iNLyze:
I also noticed, that the color widget keeps generating new instances of the color profile on reload, which seems to be the root cause of the problem.

janezd:
I can't replicate this. I'm using 3.21. Also looking at the code, there's nothing suspicious there, either.

Could you please check whether the problem persist in 3.21? It contains some changes in handling of stored setting, which may be related to the bug you encountered.

TomFreudenberg:
Hi, I have updated on 3.21 an loaded my working file.

Unhappy but still loose the previous settings.

Just to know what happens on my side:

1. I open my project.ows file which is about 145 MB
2. After loading the ows file the data "File" element has a warning sign "The file is too large to load automatically. Please reload"
3. Opening the "Color" dialog all is empty
3. After opening "File" object and selecting the underlaying data.tsv file, I press "Reload"
4. When checking the "Color" dialog all settings are resetted :-(

Hope that helps to figure out whats happen

janezd:
The .ows shouldn't be so large. But that's another matter.

Is your data secret or could you upload the first ten lines of data.tsv? 

Does your data have any duplicate column names? I don't think Orange checks for it -- not yet, but it should and will, because they can cause problems.

TomFreudenberg:
Hi Janez @janezd 

the data are secret but not the table layout. So this what we use as data:

``` sql
[...]

PR #4041, opened 2019-09-20
[FIX] Color widget failed after removing Variable.make

janezd:
##### Issue

After removing `Variable.make`, `OWColor` fails when loading settings because it tries to set `Variable.name` when loading settings.

The widget was misbehaving badly. `set_data` constructed the output table(!) and modified its attributes (not only colors but also names!). This most probably also broke `Domain`'s dictionary `_indices`.

##### Description of changes

Fixing this required reimplementing the way in which the widget stores the data. Instead of storing variables, it stores the data (names, colors) and constructs variables before sending output.

#3822 could have also been the result of the questionable behaviour of the widget and may be fixed with this PR.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/tests/test_variable.py
+++ b/Orange/data/tests/test_variable.py
@@ -195,8 +195,6 @@ class TestDiscreteVariable(VariableTest):
         var = DiscreteVariable.make("a", values=["F", "M"])
         self.assertIsNone(var._colors)
         self.assertEqual(var.colors.shape, (2, 3))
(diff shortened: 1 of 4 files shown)


markotoplak:
@janezd: you intentionally deleted ordered_values, right?

janezd:
> @janezd: you intentionally deleted ordered_values, right?

Yes. We haven't use them, as far as I know and check.

PR #4046, opened 2019-09-23
[ENH] Distances: Offload work to a separate thread

### LINKED ISSUE #3810: Distance widget should compute in a new thread
We just use Distance widget with Vesna. We forgot that it is set to measure distances with rows by default. Unfortunately, the data set was not small enough.

Distances should be implemented in a separate thread.

VesnaT:
##### Issue

Partially fixes #3810. 

##### Description of changes
Distance calculation now runs in a separate thread. 
It is only possible to interrupt the Euclidean distance by now. Other distances will be adopted once the changes are approved.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/unsupervised/owdistances.py
```diff
@@ -151,22 +169,34 @@ def _fix_missing():
...
-                              normalize=True)
-            else:
-                return metric(data, axis=1 - self.axis, impute=True)
-        except ValueError as e:
+        if data is not None:
+            for check in (_check_sparse, _fix_discrete,
+                          _fix_missing, _fix_nonbinary):
+                if not check():
+                    data = None
+                    break
```
janezd: Nothing to change here, just a comment: `all` (and `any`) shortcircuits. You could use `if not all(check() for check in ...): data = None`.

Orange/distance/distance.py
```diff
@@ -108,24 +171,43 @@ def compute_distances(self, x1, x2=None):
...
+        def callback(i):
+            return self.callback is not None and self.callback(
+                sum(cb_weights[:cb_counter]) + i *
+                cb_weights[cb_counter] / sum(cb_weights))
+
+        def callback_count():
+            nonlocal cb_counter
+            cb_counter += 1
+            return callback
+
```
janezd: I don't like two functions sharing a nonlocal variable hosted in the parent function. :) Also, I don't like computing the sum at each step. You can put the sum in a closures:

```python
        def callback_count():
            nonlocal cb_counter
            cb_counter += 1
            base = sum(cb_weights[:cb_counter])
            step = cb_weights[cb_counter] / sum(cb_weights)
            return lambda i: self.callback and self.callback(base + i * step)
```

It doesn't have to be a lambda, it just has to be inside `callback_count`.

Perhaps even bettter:

```python
class StepwiseCallbacks:
    def __init__(self, callback, weights):
        self.step_counter = -1
        self.callback = callback
        self.weights = weights

    def next(self):
        self.step_counter += 1
        base = sum(self.weights[:self.step_counter])
        step = self.weights[self.step_counter] / sum(self.weights)
        return lambda i: self.callback and self.callback(base + i * step)
```

and the in function you have `callbacks = StepwiseCallbacks(self.callback, [20, 10, 50, 5, 15])`, and then create callbacks with `callbacks.next()`.

The solution with the class is better because you can use it in multiple functions.

Orange/distance/_distance.pyx
```diff
@@ -15,9 +15,14 @@ cdef extern from "math.h":
...
-    cdef int row1, row2
-    for row1 in range(distances.shape[0]):
+cpdef void lower_to_symmetric(double [:, :] distances, callback):
+    cdef int row1, row2, step
+    n_rows1 = distances.shape[0]
+    step = n_rows1 // 100 or 100
+    for row1 in range(n_rows1):
+        if row1 % step == 0 and callback is not None \
+                and callback(row1 * 100 / n_rows1):
+            break
```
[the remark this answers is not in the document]
janezd: Actually, it's not just division, it's also branching, which can be really expensive. I think you should use double loops here.

janezd:
Please rebase to current master, which fixes the failed tests caused by changes in sklearn 0.22.

PR #4049, opened 2019-09-24
Docs: rename Evaluation to Evaluate

markotoplak:
##### Issue
If we do not rename we need to complicate the website code. I vote for renaming (it is consistent with the category name in canvas).


ajdapretnar:
Apparently the images are not indexed? Even though they probably should be since the tests passed before? 🤔 Why this error here?

ajdapretnar:
Could be. I am sure you weren't de-indexing them. 😆

ajdapretnar:
We decided to ignore Travis warnings about un-indexed images. Sorry, Travis.

PR #4055, opened 2019-09-25
[ENH] Box plot: Add 'order by importance' checkbox to groups

janezd:
Box plot shows distribution of some attribute and, if grouping is enabled, how the distribution of this attribute varies accross groups. It thus conveys information about conditional probability of the target variable given the value of the grouping variable.

If "Order by importance" is checked, the widget computes the chi square or ANOVA between all target variables and the currently selected group. This may help the user answer the question "If I divide the data into such and such groups, which is the attribute by which these groups differ most".

However, since the widget shows the conditional probability of the target given the grouping variable, it might make more sense to sort grouping variables. This will allow the user to set the target (typically the outcome, the class) and see which (grouping) attribute is the most informative about this class. This is currently possible by setting the outcome as the grouping variable and sorting the variables whose distributions we're observing, but in this case the widget shows the wrong conditional probabilities.

Both ways make some sense, but I believe that sorting group variables makes more sense. A circumstantial evidence for the latter is also that if we sort by variables, we compute a mixture of chi-square and ANOVA p-values and sort them. This is not wrong, they should be commensurable. If we sort by groups, we compute either chi-square (if variable is discrete) or ANOVA (if it's numeric) for all groups (because all groups are always discrete).

I changed the widget so that it can be tried out, but haven't thoroughly checked the code yet. I would appreciate some comments before jumping into a change that we might decide to revert the day after tomorrow.

@BlazZupan?

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -21,7 +21,7 @@ from Orange.statistics.contingency import Discrete
 from Orange.widgets import widget, gui
 from Orange.widgets.settings import (Setting, DomainContextHandler,
                                      ContextSetting)
-from Orange.widgets.utils.itemmodels import DomainModel, VariableListModel
+from Orange.widgets.utils.itemmodels import VariableListModel
 from Orange.widgets.utils.annotated_data import (create_annotated_table,
                                                  ANNOTATED_DATA_SIGNAL_NAME)
 from Orange.widgets.utils.widgetpreview import WidgetPreview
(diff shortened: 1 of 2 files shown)


janezd:
Gosh, I forgot what we decided in the end, so please forget me if I'm wrong. 

- We would have a checkbox under each list. This won't create inconsistencies, but may be confusing. In this case, we can make them exclusive.
- When sorting changes, the selected item scrolls into view.
- Open question: where is the class? If defaults are stretched bars, then the appropriate initial selection is to plot class and have no groups. Also in line with @lanzagar's description of what this widget essentially is (grouping is just a fancy addition). Alternative is showing the first variable and grouping by class.

janezd:
@BlazZupan and @lanzagar, please confirm functionality. I'll write/fix tests afterwards.

janezd:
Comment by @blazzupan: when stretching bars makes no sense (when there are no groups or when the grouping variable is the same as the variable shown), bars should not be strectched.

This should disable the checkbox and, preferably, uncheck it.

Does not necessarily belong to this PR.

PR #4056, opened 2019-09-25
[RFC] Short names for widgets whose name elides in toolbox

irgolic:
RFC regarding the chosen widget short names (e.g. Randomize -> Random).

##### Relevant PRs
https://github.com/biolab/orange-canvas-core/pull/31
https://github.com/biolab/orange-widget-base/pull/16

##### Description of changes
In an effort to avoid eliding text in the toolbox, a `short_name` property was added to widgets and `WidgetDescription`, which overrides the text in the toolbox.

Screenshot 2019-09-25 at 19 49 10

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/unsupervised/owsom.py
+++ b/Orange/widgets/unsupervised/owsom.py
@@ -172,6 +172,7 @@ N_ITERATIONS = 200
 
 class OWSOM(OWWidget):
     name = "Self-Organizing Map"
+    short_name = "SOM"
     description = "Computation of self-organizing map."
     icon = "icons/SOM.svg"
     keywords = ["SOM"]
--- a/Orange/widgets/model/owlinearregression.py
+++ b/Orange/widgets/model/owlinearregression.py
@@ -16,6 +16,7 @@ from Orange.widgets.widget import Output
 
 class OWLinearRegression(OWBaseLearner):
     name = "Linear Regression"
+    short_name = "Lin Reg"
     description = "A linear regression algorithm with optional L1 (LASSO), " \
                   "L2 (ridge) or L1L2 (elastic net) regularization."
     icon = "icons/LinearRegression.svg"
--- a/Orange/widgets/model/owlogisticregression.py
+++ b/Orange/widgets/model/owlogisticregression.py
@@ -12,6 +12,7 @@ from Orange.widgets.utils.widgetpreview import WidgetPreview
 
 class OWLogisticRegression(OWBaseLearner):
     name = "Logistic Regression"
+    short_name = "Log Reg"
     description = "The logistic regression classification algorithm with " \
                   "LASSO (L1) or ridge (L2) regularization."
     icon = "icons/LogisticRegression.svg"
--- a/Orange/widgets/model/owsgd.py
+++ b/Orange/widgets/model/owsgd.py
@@ -17,6 +17,7 @@ MAXINT = 2 ** 31 - 1
 
 class OWSGD(OWBaseLearner):
     name = 'Stochastic Gradient Descent'
+    short_name = 'SGD'
     description = 'Minimize an objective function using a stochastic ' \
                   'approximation of gradient descent.'
     icon = "icons/SGD.svg"
--- a/doc/development/source/code/widgetTemplate.py
+++ b/doc/development/source/code/widgetTemplate.py
@@ -4,6 +4,7 @@ from Orange.widgets.settings import Setting
 
 class OWWidgetName(widget.OWWidget):
     name = "Widget Name"
+    short_name = "Name"
     id = "orange.widgets.widget_category.widget_name"
     description = ""
     icon = "icons/Unknown.svg"


Orange/widgets/unsupervised/owsom.py
```diff
@@ -172,6 +172,7 @@ def paint(self, painter, _option, _index):
 
 class OWSOM(OWWidget):
     name = "Self-Organizing Map"
+    short_name = "Self-Org\nMap"
```
janezd: SOM

Orange/widgets/data/owfeatureconstructor.py
```diff
@@ -360,6 +360,7 @@ def is_valid_item(self, setting, item, attrs, metas):
 
 class OWFeatureConstructor(OWWidget):
     name = "Feature Constructor"
+    short_name = "Feature\nConstruct"
```
janezd: I think "Construct" would be OK, too.

Orange/widgets/data/owpreprocess.py
```diff
@@ -958,6 +958,7 @@ def icon_path(basename):
 
 class OWPreprocess(widget.OWWidget):
     name = "Preprocess"
+    short_name = "Preproc"
```
janezd: If you must. I don't like it. :)

Orange/widgets/data/owrandomize.py
```diff
@@ -13,6 +13,7 @@
 
 class OWRandomize(OWWidget):
     name = "Randomize"
+    short_name = "Random"
```
janezd: I think "Shuffle" would be better. Also as the long name. @BlazZupan?

Orange/widgets/unsupervised/owcorrespondence.py
```diff
@@ -44,6 +44,7 @@ def select_rows(view, row_indices, command=QItemSelectionModel.ClearAndSelect):
 
 class OWCorrespondenceAnalysis(widget.OWWidget):
     name = "Correspondence Analysis"
+    short_name = "Corresp\nAnalysis"
```
janezd: Would "Correspondence" fit? It not, I'm OK with this.

Orange/widgets/visualize/owdistributions.py
```diff
@@ -214,6 +214,7 @@ def pdf(x, a, sigma=1, weights=None):
 
 class OWDistributions(OWWidget):
     name = "Distributions"
+    short_name = "Distribute"
```
janezd: I don't like this change to a verb. Predictions -> Predict is OK, because the widget predicts and then shows predictions. But Distributions widget does not distribute anything. "Histogram" would be a better name anyway. Rename the whole widget? @BlazZupan?

Orange/widgets/visualize/ownomogram.py
```diff
@@ -563,6 +563,7 @@ def add_items(self, items):
 
 class OWNomogram(OWWidget):
     name = "Nomogram"
+    short_name = "Nomogram"
```
janezd: Isn't `short_name` name optional?

Orange/widgets/visualize/owpythagorastree.py
```diff
@@ -40,6 +40,7 @@
 
 class OWPythagorasTree(OWWidget):
     name = 'Pythagorean Tree'
+    short_name = 'Pythagora\nTree'
```
janezd: I think this short form doesn't function in English. It would be Pythagoras, and if you randomly cut the last two letters you get Pythagore. I would say Pythagor or Pythagoras.

Orange/widgets/visualize/owpythagoreanforest.py
```diff
@@ -157,6 +157,7 @@ def mousePressEvent(self, event):
 
 class OWPythagoreanForest(OWWidget):
     name = 'Pythagorean Forest'
+    short_name = 'Pythagora\nForest'
```
janezd: Same.

Orange/widgets/visualize/ownomogram.py
irgolic: `short_name` overrides the text elide and line breaking. The elide is (in my opinion) very aggressive. 
I have not yet checked whether the word fits nicely on Windows and Linux.

Orange/widgets/unsupervised/owcorrespondence.py
irgolic: Correspondence is the word that elides in the original name.

Orange/widgets/data/owfeatureconstructor.py
irgolic: "Construct Features"?

lanzagar:
I do not like this at all!
For some widgets (minority) the short name is also good and we could even consider renaming the widget (e.g. Predictions -> Predict, or acronyms instead of full names SGD, SOM).
For others, the "short name" is not a name or word at all, but looks like a typo or the product of a programmer that can't be bothered to use proper English (e.g. Preproc, Corresp Analysis, Hierarch Clustering).
And for some it is a bit worse than the real name, but just extremely confusing. Nobody will notice that 'Pythagoras Tree' is shorter than 'Pythagorean Tree' or "Construct Features" instead of "Feature Constructor" and they will not know why one name is used in the Toolbox and one on the Canvas.

I know that the current state is not good either, but I think that when shortening names because of space issues the eliding makes all the difference. I am fine with "Corresp... Analysis", but not "Corresp Analysis".
Also, I think consistency is key, and unless something was obviously shortened, the same names need to be used - i.e. no changing of word order and no noun-verb transformations.

ajdapretnar:
Thanks, I didn't want to be the one raining on everyone's parade again, but I don't like it either. I think it looks even weirder than elided names.

irgolic:
My goal is to eliminate text elide in the toolbox, whichever way possible.
I believe there's no reason text should elide on fixed size buttons. Either the widget should be renamed or a shortened name should be employed.

Note that when the widget is hovered in the toolbox, the Help menu shows the widget's full name. 

The following short names are abbreviations/acronyms:

Linear Regression –> Lin Reg
Logistic Regression –> Log Reg
Stochastic Gradient Descent –> SGD
Self-Organizing Map –> SOM

The following are shortened words, which still convey the same meaning IMO:

Concatenate –> Concat
Preprocess –> Preproc
Randomize –> Random (Shuffle?)
Distance Transformation –> Distance Transform

The following change the noun to a verb:

Distributions –> Distribute (Histogram?)
Predictions –> Predict
Correlations –> Correlate
Feature Constructor –> Construct Features

The following are shortenings are clumsier, not sure a good short name exists:

Hierarchical Clustering –> Hierarch Clustering
Pythagorean Tree –> Pythagoras Tree (Pythagora Tree?)
Pythagorean Forest –> Pythagoras Forest (Pyhagora Forest?)
Correspondence Analysis –> Correspond Analysis

Nomogram looks as if it should fit on Mac, which is why I set it to simply ignore text elision. Yet to be tested on Windows and Linux.

Another option is a redesign of the toolbox, such that it would fit two widgets per row, but each row would take up less height.

janezd:
Like Ajda, I didn't like this, but I held back.

**Abbreviations**: OK

**Shortened:**
- Concat and Preproc hurt my eyes.
- Randomize could become Shuffle; Random is not OK, because it doesn't produce random data
- Distance Transform is not much shorter than Distance Transformation.
  
**Verb:**
- Distributions can become Histogram (I may have even suggested renaming when reimplementing the widget some months ago); Distribute is unacceptable because it doesn't distribute anything
- Correlate also doesn't work well. The widget shows correlations.
- Construct Features ... I'm OK with this, if we have to, but we should change the long name then.

**Clumsy shortennings**: as @lanzagar wrote.

> I believe there's no reason text should elide on fixed size buttons.

I don't like eliding either, but isn't "Preproc" just eliding without dots? :) And, as Lan wrote, I prefer consistency. Using noun in one place and verb in another looks as if we were just sloppy.

irgolic:
Okay, I'll leave the abbrevations/acronyms and remove the rest.

I'll look for another solution that won't elide words up to the length of 'Correspondence' (the longest word in the array of standard widgets).

irgolic:
Widgets to be considered for renaming:

Randomize -> Shuffle
Distributions -> Histogram
Predictions -> Predict
Feature Constructor -> Construct Features

janezd:
We're closing this today. :)

We said having shorter names could be a good idea for some add-ons and for some other projects, so we merge pull requests in base widget and canvas. But we wouldn't change those widgets.

Issue #4057, opened 2019-09-25
Correlations do not include class attribute


janezd:
Is there any reasonable reason why continuous `class_vars` aren't included in computation of correlations in the Correlation widget?

I tried to demonstrate that among 1000 independent random variables, some will be corelated to the random outcome by pure chance. Currently you have to use a data set with 1000 (well, 1001) attributes and treat one of them as "a target".

**Describe the solution you'd like**

I would not only include class vars, but also change default for `feature_model`. When the data has a single, numeric class variable, I would by default compute correlations with that variable instead of correlations between all pairs.

This would also save the poor lecturers from problems arising from the fact that while the widget can be stopped in principle, it is (on my computer) too busy to react to the Pause button while computing correlations between 1000 * 999 / 2 pairs. :)

PR #4067, opened 2019-10-01
[ENH] Correlations: Include continuous class and meta variables

VesnaT:
##### Issue

Implements #4057

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcorrelations.py
+++ b/Orange/widgets/data/owcorrelations.py
@@ -365,7 +365,11 @@ class OWCorrelations(OWWidget):
 
     def set_feature_model(self):
         self.feature_model.set_domain(self.cont_data and self.cont_data.domain)
-        self.feature = None
+        data = self.data
+        if self.cont_data and data.domain.has_continuous_class:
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owcorrelations.py
```diff
@@ -365,7 +365,11 @@ def set_data(self, data):
 
     def set_feature_model(self):
         self.feature_model.set_domain(self.cont_data and self.cont_data.domain)
-        self.feature = None
+        data = self.data
+        if data and data.domain.has_continuous_class and self.cont_data:
```
janezd: Is it necessary to test for `data` or would `if self.cont_data and data.domain.has_continuous_class` suffice?

If you don't change it by the end of the week (or explain why the test is necessary), I will consider my comment nitpicking and merge the code as it is. :)

VesnaT: I removed the unnecessary condition.

janezd: I'd do it myself but wasn't sure about it. :)

PR #4071, opened 2019-10-01
[ENH] Datasets: Remove control area

### LINKED ISSUE #4069: [RFC] Data Sets widget does not need control area
It contains no useful information (if the number of available and cached data sets is important, it can go to the status line), and Send button could be below the table. If we care about vertical space, the label Filter and the corresponding edit line can be arranged horizontally.

If we eliminate control area, the widget would fit my screen horizontally.

janezd:
Sapientem sat?

Screenshot 2019-10-03 at 14 54 35

(Fixes #4069.)

To show indicators with different colors, it needs https://github.com/biolab/orange-widget-base/pull/17. Before https://github.com/biolab/orange-widget-base/pull/17 is merged, indicators will remain black, as they are now, but the widget will still work.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -118,6 +118,16 @@ class Namespace(SimpleNamespace):
             self.title = self.filename
 
 
+class TreeViewWithReturn(QTreeView):
+    returnPressed = Signal()
+
+    def keyPressEvent(self, e):
+        if e.key() == Qt.Key_Return:
+            self.returnPressed.emit()
+        else:
+            super().keyPressEvent(e)
+
+
 class OWDataSets(OWWidget):
     name = "Datasets"
     description = "Load a dataset from an online repository"
@@ -160,8 +170,6 @@ class OWDataSets(OWWidget):
     #: Selected dataset id
     selected_id = settings.Setting(None)   # type: Optional[str]
 
-    auto_commit = settings.Setting(False)  # type: bool
-
     #: main area splitter state
     splitter_state = settings.Setting(b'')  # type: bytes
     header_state = settings.Setting(b'')    # type: bytes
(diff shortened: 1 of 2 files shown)


PrimozGodec:
I tested it and it works well. There is just a need to change the documentation screenshots and maybe add a sentence about the status color. Should we do it here or in the separate PR?

Orange/widgets/data/owdatasets.py
```diff
@@ -539,8 +518,15 @@ def closeEvent(self, event):
         self.header_state = bytes(self.view.header().saveState())
         super().closeEvent(event)
 
-    def load_data(self, path):  # pylint: disable=no-self-use
-        return Orange.data.Table(path)
+    def load_and_output(self, path):  # pylint: disable=no-self-use
+        if path is None:
+            self.Outputs.data.send(None)
+        else:
+            data = Orange.data.Table(path)
```
lanzagar: The previous `load_data` method is overriden in Single Cell Datasets.
You can still call it here and make this a new method instead of changing it from `load_data`, which should keep everything working downstream.

janezd:
> Users could previously choose a dataset and click Send (or have auto send checked). After we remove the button, double click is probably the only way to do this? Is that obvious enough

I suppose. If nothing happens on single click, the normal reflex is to double-click.

> Upon downloading a dataset warnings are printed: 'processEvents' argument is deprecated.
It has nothing to do with this PR, but it should be as simple as searching for and removing all occurrences of processEvents=None.

I didn't want to do this in this PR, but it's trivial, so now I did it.

> This breaks Single Cell Datasets. There should probably be some refactoring and maybe a common base class, but I would avoid dragging all that into this PR. I think the following should be enough:
> We add openclass=True, just to make it clearer that this widget is reused (it should have been there already)
> You changed load_data to load_and_output. We could leave load_data as it was and just call it inside a new method, which should then work fine with the downstream changes.

I restored `load_data`, but I wouldn't add `openclass=True`. It is preferable to have a warning that reminds us to extract common routines to a base class. `openclass=True` is also a promise that we won't break compatibility; we can only do this after refactoring.

PrimozGodec:
> Users could previously choose a dataset and click Send (or have auto send checked). After we remove the button, double click is probably the only way to do this? Is that obvious enough (honest question, I don't know)?

What about having a send button/apply at the bottom of the widget (similarly than file widget has)?

janezd:
Check whether pressing return also submits the data.

Add Tooltip: "Double click to send".

PR #4073, opened 2019-10-02
Sieve: Enable attribute controls

### LINKED ISSUE #4061: VizRank allows for manual changing of features in Sieve despite Features input
**Describe the bug**

Certainly clearly described in the title. :)

**To Reproduce**

Connect Sieve to File and run Vizrank. Do not close the VizRank dialog afterwards.

Connect File -> Correlations -> Sieve, so that Correlations are used to set the features in Sieve. This connection disables the combo boxes for setting the features in Sieve. A naughty user can howere still change the features in Sieve by clicking on VizRank.

**Expected behavior**

Fix if it's trivial. Close this issue if not. It's an unimportant glitch, not a bug.

VesnaT:
##### Issue

Fixes #4061

##### Description of changes
Due to consistency (with scatter plot) feature combo boxes and vizrank button are enabled even with features on the input.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owsieve.py
+++ b/Orange/widgets/visualize/owsieve.py
@@ -107,18 +107,18 @@ class OWSieveDiagram(OWWidget):
         self.areas = []
         self.selection = set()
 
-        self.attr_box = gui.hBox(self.mainArea)
+        attr_box = gui.hBox(self.mainArea)
         self.domain_model = DomainModel(valid_types=DomainModel.PRIMITIVE)
         combo_args = dict(
-            widget=self.attr_box, master=self, contentsLength=12,
+            widget=attr_box, master=self, contentsLength=12,
             callback=self.attr_changed, sendSelectedValue=True, valueType=str,
             model=self.domain_model)
         fixed_size = (QSizePolicy.Fixed, QSizePolicy.Fixed)
         gui.comboBox(value="attr_x", **combo_args)
-        gui.widgetLabel(self.attr_box, "\u2715", sizePolicy=fixed_size)
+        gui.widgetLabel(attr_box, "\u2715", sizePolicy=fixed_size)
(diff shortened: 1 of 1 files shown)


janezd:
It doesn't fix #4061, it eliminates it. :)

I don't like the change. @ajdapretnar said it makes sense from user perspective, which I have to agree.

But consider a schema where File is connected to Sieve and Correlations, and Correlations is connected to Sieve, through features. The user selects a pair of features in Correlations. Then he overrides this by changing a different pair in Sieve. The schema is not inconsistent. What will happen if we save it? Upon loading, will the stored setting in Sieve override the signal from Correlations, or will Orange first load the settings and then process signals? Or is it undefined?

We have to discuss this. Perhaps we'll decide that you change is OK, but perhaps we'll fix scatter plot instead.

janezd:
I also found an interesting bug which was here before and it still is. Connect a schema like described above, and have VizRank and Correlations open. Choose a pair in VizRank, then a pair in Correlations, then a pair in Vizrank, in Correlations... Instead of having a single chosen item, you will have two chosen items in each of them.

VesnaT:
That is an interesting bug. It actually comes with PyQt5==5.12. With PyQt5==5.9, selecting items (when multiple windows are open) works fine.

VesnaT:
After some more investigation: the bug has nothing to do with multiple windows being open. It happens every time one tries to re-select a row in the Vizrank window, after it has lost focus.
To reproduce: File(Iris) -> Scatterplot -> Find Informative Projections -> Start -> Click on Scatterplot (to make Vizrank loose focus) -> Select a (n unselected) row in Vizrank.

janezd:
We decided to change the scatter plot to disable combos when the features input is present.

PR #4076, opened 2019-10-03
[FIX] Merge data: Rename variables with duplicated names

### LINKED ISSUE #4070: Merge data may output domain with duplicated names
The God sayeth: "*Thou shall not create domains with two variables of the same name.*"

And the Merge widget came forth and said: "*Watch me*", and dared to even announce:

```
    class Warning(widget.OWWidget.Warning):
        duplicate_names = Msg("Duplicate variable names in output.")
```

**Expected behavior**

Maybe put variables through `get_unique_names`. Though this one is a bit more complicated. `get_unique_names` is appropriate when a variable name may already exist and we need to add a number to the name. Here, variables come together from different domains. I would prefer that all variables get numbers if there is a conflict. If both domains have variable `foo`, I'd like to see `foo (1)` and `foo (2)`, not `foo` and `foo (1)`. 

Conceptually (this code doesn't actually fit anywhere literally, but probably works):

```
conflict = set(left_vars) & set(right_vars)
left_vars = [get_unique_names(var, conflict) for var in left_vars]
right_vars = [get_unique_names(var, conflict | left_vars) for var in right_vars]
```

Variables in `left_vars` are renamed if they are in conflict with `right_vars`. `right_vars` are then renamed to consecutive numbers.

VesnaT:
##### Issue

Fixes #4070

##### Description of changes
- implement Variable.rename(name)
- check for duplicate variable names in domain and rename if any 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -1,4 +1,3 @@
-import collections
 import re
 import warnings
 from collections import Iterable
@@ -215,7 +214,8 @@ class Value(float):
 
     def __hash__(self):
         if self.variable.is_discrete:
-            # It is not possible to hash the id and the domain value to the same number as required by __eq__.
+            # It is not possible to hash the id and the domain value to the
+            # same number as required by __eq__.
             # hash(1) == hash(Value(DiscreteVariable("var", ["red", "green", "blue"]), 1)) == hash("green")
             # User should hash directly ids or domain values instead.
             raise TypeError("unhashable type - cannot hash values of discrete variables!")
@@ -451,8 +451,10 @@ class Variable(Reprable, metaclass=VariableMeta):
         # Use make to unpickle variables.
         return make_variable, (self.__class__, self._compute_value, self.name), self.__dict__
 
-    def copy(self, compute_value):
-        var = type(self)(self.name, compute_value=compute_value, sparse=self.sparse)
+    def copy(self, compute_value=None, name=None, **kwargs):
+        var = type(self)(name=name or self.name,
+                         compute_value=compute_value or self.compute_value,
+                         sparse=self.sparse, **kwargs)
         var.attributes = dict(self.attributes)
         return var
 
@@ -507,6 +509,10 @@ class ContinuousVariable(Variable):
     def format_str(self):
         return self._format_str
 
+    @format_str.setter
+    def format_str(self, value):
+        self._format_str = value
+
     @property
     def colors(self):
         if self._colors is None:
@@ -559,9 +565,12 @@ class ContinuousVariable(Variable):
 
     str_val = repr_val
 
-    def copy(self, compute_value=None):
-        var = type(self)(self.name, self.number_of_decimals, compute_value, sparse=self.sparse)
-        var.attributes = dict(self.attributes)
+    def copy(self, compute_value=None, name=None, **kwargs):
+        var = super().copy(compute_value=compute_value, name=name,
+                           number_of_decimals=self.number_of_decimals,
+                           **kwargs)
+        var.adjust_decimals = self.adjust_decimals
+        var.format_str = self._format_str
         return var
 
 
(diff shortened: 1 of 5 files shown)


Orange/data/variable.py
```diff
@@ -451,8 +452,14 @@ def __reduce__(self):
         # Use make to unpickle variables.
         return make_variable, (self.__class__, self._compute_value, self.name), self.__dict__
 
-    def copy(self, compute_value):
-        var = type(self)(self.name, compute_value=compute_value, sparse=self.sparse)
+    def rename(self, name):
```
janezd: This name is misleading because it it implies that it renames a variable. I would eliminate this function and just use `copy` when I need a new variable with a different name.

If you really want to keep rename (which I really don't like, but it's a matter of opinion), we should call it `renamed`. (Like `list.sort`, which works in place, and `list.renamed`, which returns a new list.)

Orange/data/variable.py
```diff
@@ -451,8 +452,14 @@ def __reduce__(self):
...
+        return self._copy(name, self.compute_value)
+
+    def copy(self, compute_value=None):
```
janezd: Perhaps we should just have `def copy(self, compute_value=None, *, name=None)`, and then, of course, `if name is None: name = self.name`. So we avoid superfluous `_copy` function.

Orange/data/variable.py
```diff
@@ -559,8 +566,9 @@ def repr_val(self, val):
 
     str_val = repr_val
 
-    def copy(self, compute_value=None):
-        var = type(self)(self.name, self.number_of_decimals, compute_value, sparse=self.sparse)
+    def _copy(self, name, compute_value):
```
janezd: Same.

Orange/widgets/data/owmergedata.py
```diff
@@ -566,6 +557,29 @@ def _join_table_by_indices(self, reduced_extra, lefti, righti, rightu):
...
+
+        variables = domain.variables + domain.metas
+        proposed_names = [m.name for m in variables]
+        unique_names = get_unique_names_duplicates(proposed_names)
+        duplicates = set()
+        for p_name, u_name, var, c in zip(proposed_names, unique_names,
+                                          variables, lists):
+            if p_name != u_name:
+                duplicates.add(p_name)
+                var = var.rename(u_name)
```
janezd: ... and then this would be `var = var.copy(name=u_name)`.

Orange/widgets/data/owmergedata.py
```diff
@@ -217,13 +217,13 @@ class Outputs:
     resizing_enabled = False
 
     class Warning(widget.OWWidget.Warning):
-        duplicate_names = Msg("Duplicate variable names in output.")
+        renamed_vars = Msg("Some variables ({}) have been renamed.")
```
janezd: Maybe: `"Some variables have been renamed to avoid duplicates.\n{}"`?

Orange/widgets/data/owmergedata.py
VesnaT: I'd prefer `var = var.copy(compute_value=var.compute_value, name=u_name)`.

Speaking of properties we want to keep. Why not copying `color` as well?

janezd: | `var = var.copy(compute_value=var.compute_value, name=u_name)`

OK, if you wish. I view `copy` as a function that copies the variable, but because its attributes are constant, you can also override some by giving them as keyword arguments. Much like `_replace` in `namedtuple` (which is also misleading: for some time I wondered how can it modify a tuple!). This is why I wouldn't pass a value that I am not changing (`compute_value`). If you pass `compute_value`, why not also `values` and `number_of_decimals`?

I think that `_colors` doesn't need to be copied because it is also saved in `attributes`. I hope this is true.

VesnaT: On the other hand. `compute_value` could be considered same as `name`: `if compute_value is None: compute_value = self.compute_value`.

janezd: Yes. On the second point. Anything you do not explicitly change is (should be) copied.

PR #4078, opened 2019-10-03
OWScatterPlotGraph: Don't combine color and shape legend for different data

### LINKED ISSUE #4065: Scatterplotgraph is too eager to combine legends
When checking results of prediction models (typically coming from Predictions widget), we may get this:

Screenshot 2019-09-30 at 14 43 23

Points belonging to C1 are blue and they are circles, and those is C2 are red and crosses. But the legend implies there are only blue circles and red crosses, which is evidently not the case.

Scatterplotgraph decides whether to combine the color legend and the shape legen into a single legend based on whether they have the same labels:

```
        shape_labels = self.master.get_shape_labels()
        color_labels = self.master.get_color_labels()
        if shape_labels == color_labels and shape_labels is not None:
            self._update_combined_legend(shape_labels)
```

A proper way would be something like

```
        if shape_attr is color_attr and shape_attr is not None:
            self._update_combined_legend(...)
```

except that scatterplotgraph does not have access to `shape_attr` and `color_attr`. Something like `color_var` may even not exist. Scatterplotgraph is the base class for all kinds of projections, and the colors of points need not represent values of any variable but can be compute from, for instance, the stress (in MDS).

Hence only the `master`, that is, the widget, can reliably tell the scatterplotgraph whether to combine the color and shape legends. I would hate to have an API with too many methods, but it looks like we will have to add this one.

janezd:
Fixes #4065.

This approach could be costly if getting the colors and sizes from master is slow -- which, I think, we said shouldn't be the case: the graph requests this data multiple times and the widget should have it cached if necessary.

The advantage of this approach is that it doesn't require a new function provided by widget.

@VesnaT's opinion would be appreciated.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -1208,7 +1208,15 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         cont_color = self.master.is_continuous_color()
         shape_labels = self.master.get_shape_labels()
         color_labels = self.master.get_color_labels()
-        if shape_labels == color_labels and shape_labels is not None:
+        if not cont_color and shape_labels is not None \
+                and shape_labels == color_labels:
+            colors = self.master.get_color_data()
+            shapes = self.master.get_shape_data()
+            mask = np.isfinite(colors) * np.isfinite(shapes)
+            combined = (colors == shapes)[mask].all()
+        else:
+            combined = False
+        if combined:
             self._update_combined_legend(shape_labels)
         else:
             self._update_shape_legend(shape_labels)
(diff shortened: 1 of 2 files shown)


VesnaT:
I'm fine with the solution. The performance of the method should not be an issue, since `shape_labels == color_labels` does not happen that often, imo. 

On the other hand, I don't understand why adding something like
```
def shape_attr_is_color_attr(self):
    return self.attr_color is not None and self.attr_color is self.attr_shape
```
to the `OWProjectionWidgetBase` is a problem.

janezd:
> On the other hand, I don't understand why adding something like
> 
> ```
> def shape_attr_is_color_attr(self):
>     return self.attr_color is not None and self.attr_color is self.attr_shape
> ```

I just wouldn't like to increase the number of methods that the widget must provide if it can be avoided and the price is low enough.

PR #4093, opened 2019-10-10
[ENH] Add remove sparse features preprocessor

AndrejaKovacic:
##### Description of changes
Preprocessor widget now enables removing sparse features. Threshold is up to the user. I thought we could use the same icon as feature selection and cur decomposition.

##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation

### DIFF
--- a/Orange/preprocess/preprocess.py
+++ b/Orange/preprocess/preprocess.py
@@ -578,12 +578,12 @@ class SelectDense(Preprocess):
 
     Parameters
     ----------
-    treshold : float
+    threshold : float
         Minimal proportion of non-zero entries of a feature
     """
 
-    def __init__(self, treshold=0.05):
-        self.treshold = treshold
+    def __init__(self, threshold=0.05):
+        self.threshold = threshold
 
     def __call__(self, data):
         if sp.issparse(data.X):
@@ -592,7 +592,7 @@ class SelectDense(Preprocess):
             sparsness = [data_csc[:, i].count_nonzero()/h for i in range(w)]
         else:
             sparsness = np.count_nonzero(data.X, axis=0)/ data.X.shape[0]
-        att = [a for a, s in zip(data.domain.attributes, sparsness) if s >= self.treshold]
+        att = [a for a, s in zip(data.domain.attributes, sparsness) if s >= self.threshold]
         domain = Orange.data.Domain(att, data.domain.class_vars,
                                     data.domain.metas)
         return data.transform(domain)
--- a/Orange/preprocess/preprocess.py
+++ b/Orange/preprocess/preprocess.py
@@ -18,7 +18,8 @@ from . import impute, discretize, transformation
 __all__ = ["Continuize", "Discretize", "Impute", "RemoveNaNRows",
            "SklImpute", "Normalize", "Randomize", "Preprocess",
            "RemoveConstant", "RemoveNaNClasses", "RemoveNaNColumns",
-           "ProjectPCA", "ProjectCUR", "Scale", "AdaptiveNormalize"]
+           "ProjectPCA", "ProjectCUR", "Scale", "SelectDense",
+           "AdaptiveNormalize"]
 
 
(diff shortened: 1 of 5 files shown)


janezd:
I like the preprocessor, but do we have a better name for it? :)

I didn't realize what it does before seeing the code.

Orange/preprocess/preprocess.py
```diff
@@ -569,6 +570,32 @@ def __call__(self, data):
...
+    user-defined treshold.
+
+    Parameters
+    ----------
+    treshold : float
+        Minimal proportion of non-zero entries of a feature
+    """
+
+    def __init__(self, treshold=0.05):
+        self.treshold = treshold
```
ajdapretnar: I don't want to be this person, but perhaps for clarity and for maintaining a good API, I would correct the name to _threshold_.

AndrejaKovacic: Thanks. This, along with length is in my top misspelled words.

BlazZupan:
I would support "sparse feature" in the name of the widget. One reason is that I cannot come up with a better one, and second because this term has been used in this context. For instance, googling '"sparse features" machine learning' returns [83.000 pages](https://www.google.com/search?client=safari&rls=en&q=%22sparse+feature%22+machine+learning&ie=UTF-8&oe=UTF-8).

ajdapretnar:
This works well also on bag of words sparse type. I was wondering would it make sense to have also an int option. As in Max instances with nonzero values... 🤔 In the case of text mining this would then be similar to *document frequency* filter.

AndrejaKovacic:
Sounds good, we can add that in the second iteration of this preprocessor.

markotoplak:
@AndrejaKovacic, I added an icon. I reused something we already had in "Data". What do you think?

AndrejaKovacic:
I think it's fitting, better than select column.

janezd:
@AndrejaKovacic, if you pull master and rebase, it should no longer crash on SOM. I apologize for inconvenience. :)

This is assigned to @BlazZupan. If UX is confirmed, I can check the code and merge.

Orange/preprocess/preprocess.py
```diff
@@ -569,6 +571,32 @@ def __call__(self, data):
...
+    def __init__(self, threshold=0.05):
+        self.threshold = threshold
+
+    def __call__(self, data):
+        if sp.issparse(data.X):
+            data_csc = sp.csc_matrix(data.X)
+            h, w = data_csc.shape
+            sparsness = [data_csc[:, i].count_nonzero()/h for i in range(w)]
+        else:
+            sparsness = np.count_nonzero(data.X, axis=0)/ data.X.shape[0]
```
janezd: Because you'll be rebasing and pushing anyway, you can also add a space before `/`, so it's balanced. Bonus points for putting spaces before and after `/` two lines above.

Orange/widgets/data/owpreprocess.py
```diff
@@ -250,6 +250,37 @@ def createinstance(params):
...
+        self.layout().addLayout(form)
+
+    def setThresh(self, thresh):
+        if self.sparse_thresh != thresh:
+            self.sparse_thresh = thresh
+            self.cspin.setValue(thresh)
+            self.changed.emit()
+
+    def parameters(self):
+        return {'sparse_thresh':self.sparse_thresh}
```
janezd: A space after `:` would be nice, too.

Orange/widgets/data/owpreprocess.py
```diff
@@ -250,6 +250,37 @@ def createinstance(params):
...
+
+    def setParameters(self, params):
+        self.setThresh(params.get('sparse_thresh', 5))
+
+    @staticmethod
+    def createinstance(params):
+        params = dict(params)
+        threshold = params.pop('sparse_thresh', 5)
+        return RemoveSparse(threshold=threshold/100)
```
janezd: And just a pair of spaces here and I'm merging - when it's rebased so it passes SOM.

PR #4098, opened 2019-10-11
[FIX] Datetime conversion to string

### LINKED ISSUE #3963: File: conversion to string destroys datetime
**Describe the bug**
File widget does not convert datetime to string appropriately.

**To Reproduce**
File (open any dataset with datetime)
Convert datetime into text.
Observe the result in Data Table.

**Orange version:**
3.23.dev

**Expected behavior**
String version of 2010-01-01 should be '2010-01-01' not its float value.

rokgomiscek:
##### Issue
Fixes #3963

##### Description of changes
Added additional branch when converting to string. 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/domaineditor.py
+++ b/Orange/widgets/utils/domaineditor.py
@@ -313,7 +313,7 @@ class DomainEditor(QTableView):
                 col_data = self._to_column(col_data, is_sparse)
             elif tpe == StringVariable:
                 var = tpe.make(name)
-                if type(orig_var) == DiscreteVariable:
+                if type(orig_var) in [DiscreteVariable, TimeVariable]:
                     col_data = [orig_var.repr_val(x) if not np.isnan(x) else ""
                                 for x in self._iter_vals(col_data)]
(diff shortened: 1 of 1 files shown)


Orange/widgets/utils/domaineditor.py
```diff
@@ -322,6 +322,9 @@ def numbers_are_round(var, col_data):
                                 str(int(x)) if round_numbers else
                                 orig_var.repr_val(x)
                                 for x in self._iter_vals(col_data)]
+                elif type(orig_var) == TimeVariable:
+                    col_data = [orig_var.repr_val(x) if not np.isnan(x) else ""
+                                for x in self._iter_vals(col_data)]
```
lanzagar: This looks to be the same as in the above `if type(orig_var) == DiscreteVariable`,
we could change the first check to match both types, e.g. `if type(orig_var) in (DiscreteVariable, TimeVariable)`,
so we can avoid copy/pasting code.

Issue #4103, opened 2019-10-14
Preprocess: Select Random Features, increase number of features


BlazZupan:
**The problem**
Select Random Features preprocessor as implemented in the Preprocess widget allows the user to enter only a very limited number of attributes. I believe 99 is the current limit. The number of features on the input is usually larger. Say, in image embedding, we have thousands of features. The limit of 99 is therefore too stringent.

**Solution**
Remove any limit on the number of requested features. Provide enough space to enter, say, a million features.

![image](https://user-images.githubusercontent.com/726604/66750804-b5d5d500-ee8d-11e9-9206-2cf88d5ef727.png)

PR #4105, opened 2019-10-14
#4103 - Updating maximum number of features for Preprocess: Select Random Fea…

aturanjanin:
…tures

##### Issue 
#4103 - Select Random Features preprocessor as implemented in the Preprocess widget allows the user to enter only a very limited number of attributes.

##### Description of changes
Added a maximum attribute to QSpinBox function in RandomFeatureSelectEditor to allow up to 10000000 features to be selected.


janezd:
If the number is to be arbitrarily large, it may be better to have a line edit with QIntValidator instead of spin box. @BlazZupan?

PR #4110, opened 2019-10-17
Pymssql backend: Nicer error msg

VesnaT:
##### Issue

PymssqlBackend raises exceptions with unreadable messages.

To reproduce:
- install pymssql
- open SQL Table widget
- write "test" in all four text boxes
- press Enter

##### Description of changes
- parse pymssql exception if possible and raise BackendError with readable message

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/data/sql/backend/mssql.py
```diff
@@ -9,6 +9,13 @@
 from Orange.data.sql.backend.base import ToSql, BackendError
 
 
+def parse_ex(ex):
+    try:
+        return ex.args[0][1].decode().split("\n")[-2]
```
janezd: You could use `splitlines()` instead of `split("\n")`. Supposedly safer. Change if you'll be changing anything anyway; leave as it is if not.

lanzagar:
Although I can't reproduce the problem (I had a normal error message already),
this does not seem to break working cases, but helps in some (like @VesnaT's).

PR #4111, opened 2019-10-17
[ENH] Bhatthacharayya distance

AndrejaKovacic:
##### Description of changes
A new metric is introduced in distances widget. Bhatthacharayya is used to measure the distance between distributions. Currently, the user has to already have the data that they see as a distribution. In the future, pivot could be extended to have aggregation of all features, filtered by column values, so distributions could be generated in Orange itself. 

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/distance/distance.py
+++ b/Orange/distance/distance.py
@@ -3,6 +3,7 @@ from unittest.mock import patch
 
 import numpy as np
 from scipy import stats
+from scipy.spatial.distance import cdist
 from scipy import sparse as sp
 import sklearn.metrics as skl_metrics
 from sklearn.utils.extmath import row_norms, safe_sparse_dot
@@ -643,6 +644,21 @@ class PearsonR(CorrelationDistance):
 class PearsonRAbsolute(CorrelationDistance):
     def fit(self, _):
         return PearsonModel(True, self.axis, self.impute)
+
+def _bhattacharyya(a, b):
+    # not a real metric, does not obey triangle inequality
+    return -np.log(np.sum(np.sqrt(a*b)))
+
+class Bhattacharyya(Distance):
+    def fit(self, data):
+        return BhattacharyyaModel(self.axis, self.impute)
+
+class BhattacharyyaModel(DistanceModel):
+
+    def compute_distances(self, x1, x2):
+        if x2 is None:
+            x2 = x1
+        return cdist(x1, x2, _bhattacharyya)
 
 
 class Mahalanobis(Distance):
--- a/Orange/distance/distance.py
+++ b/Orange/distance/distance.py
@@ -5,6 +5,7 @@ import numpy as np
 from scipy import stats
 from scipy import sparse as sp
 import sklearn.metrics as skl_metrics
+from sklearn.utils import check_array
 from sklearn.utils.extmath import row_norms, safe_sparse_dot
 from sklearn.metrics import pairwise_distances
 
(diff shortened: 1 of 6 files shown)


Orange/distance/distance.py
```diff
@@ -644,6 +644,36 @@ class PearsonRAbsolute(CorrelationDistance):
     def fit(self, _):
         return PearsonModel(True, self.axis, self.impute)
 
+def _prob_dist(a):
+    # Makes the vector sum to one, as to mimick probability distribution.
```
janezd: mimic

Orange/distance/distance.py
```diff
@@ -644,6 +644,36 @@ class PearsonRAbsolute(CorrelationDistance):
...
+    return a / np.sum(a)
+
+def _bhattacharyya(a, b):
+    # not a real metric, does not obey triangle inequality
+    a = _prob_dist(a)
+    b = _prob_dist(b)
+    if sp.issparse(a):
+        return - np.log(np.sum(np.sqrt(a.multiply(b))))
+    return - np.log(np.sum(np.sqrt(a * b)))
```
janezd: Sorry for nitpicking. But since you are probably going to change the flie anyway, you can remove space after unary minus (PEP8 is not specific about this, but see the operators in  https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex)

Orange/widgets/unsupervised/owdistances.py
```diff
@@ -157,6 +160,9 @@ def _fix_missing():
                       _fix_discrete, _fix_missing, _fix_nonbinary):
             if not check():
                 return None
+        if (METRICS[self.metric_idx][0] == 'Bhattacharyya') and _min(data.X) < 0:
```
janezd: It is in general a bad idea to use string literals like this. Somebody will rename it ... and there we go. I'd be happier with `METRICS[self.metric_idx][1] is distance.Bhattacharyya` (and, by the way, you can remove the parentheses).

On the other hand, why wouldn't the distance itself (i.e. function `distance. _bhattacharyya`) test the values and raise `ValueError("Bhattcharyya distance requires non-negative values")`? The widget already catches and shows `ValueError` exceptions.

As I proposed on Friday, this is better because it gives a reasonable error also to anybody that would call this distance from a script. With current code, if somebody calls Bhattacharyya with negative values, (s)he will get just a `RuntimeWarning: invalid value encountered in sqrt`, the result will be `nan` ... and this `nan` will propagate until an unrelated function crashes down the road. Testing and raising an exception there is better because it helps debugging.

This is open for discussion -- on Friday.

AndrejaKovacic: I don't think it's really open for discussion, since you are obviously right. Thanks for being courteous about it.

Orange/distance/distance.py
```diff
@@ -645,33 +645,31 @@ class PearsonRAbsolute(CorrelationDistance):
...
-        if a.min() < 0:
-            raise ValueError("Bhattcharyya distance requires non-negative values")
-        return
-    if min(a) < 0:
+
+def check_non_negative(a):
+    # Raise an exception for infinities, nans and negative values
+    check_array(a,
+                accept_sparse=True, accept_large_sparse=True, ensure_2d=False)
+    if a.min() < 0:
```
janezd: @AndrejaKovacic, would this be OK, too?

Dense arrays also have a method `min`, so there's no need for `if`. Also, I think it is better to not change the exception message raised by `check_array` so the caller is informed that, for instance, there are `nan` values in the data.

AndrejaKovacic: Yes, it's more informative this way.

PR #4113, opened 2019-10-17
[FIX] File: Limit the width of the URL combo

### LINKED ISSUE #4107: OWFile: elide or hide url if too long
**Describe the bug**
File remembers settings. So when an url is really long, the widgets is resized to fit the url. This results in huge windows.

**To Reproduce**
1. Open File.
2. Enter the following url in the url field: https://raw.githubusercontent.com/SteffiPeTaffy/machineLearningAZ/master/Machine%20Learning%20A-Z%20Template%20Folder/Part%204%20-%20Clustering/Section%2025%20-%20Hierarchical%20Clustering/Mall_Customers.csv
3. Save the workflow or just close Orange.
4. Open Orange and open the File widget again.

**Orange version:**
3.23.1

**Expected behavior**
File widget is of normal size.

**Screenshots**
Screen Shot 2019-10-14 at 10 11 23

janezd:
##### Issue

Fixes #4107, but introduces a different problem, which may need to be fixed in https://github.com/biolab/orange-widget-base/blob/master/orangewidget/utils/combobox.py.

Enter any URL (a random string is OK), press Enter, then open the combo box's popup. The popup's height will be too small. Add more URL's and the popup will keep being too small by half a line.

This occurs because of `view = self.view()` in https://github.com/biolab/orange-widget-base/blob/master/orangewidget/utils/combobox.py#L32. (Putting a `return` before this line helps, a return below it doesn't.)

Calling `self.view()` triggers construction of a `QComboBoxListView ` via a call to `QComboBoxPrivateContainer* QComboBoxPrivate::viewContainer()` (https://github.com/qt/qtbase/blob/dev/src/widgets/widgets/qcombobox.cpp#L1046).

@ales-erjavec, ideas?

### DIFF
--- a/Orange/data/sql/backend/mssql.py
+++ b/Orange/data/sql/backend/mssql.py
@@ -9,6 +9,13 @@ from Orange.data.sql.backend import Backend
 from Orange.data.sql.backend.base import ToSql, BackendError
 
 
+def parse_ex(ex: Exception) -> str:
+    try:
+        return ex.args[0][1].decode().splitlines()[-1]
+    except:  # pylint: disable=bare-except
+        return str(ex)
+
+
 class PymssqlBackend(Backend):
     display_name = "SQL Server"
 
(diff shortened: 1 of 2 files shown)


janezd:
This is so much simpler. Thanks!

PR #4116, opened 2019-10-18
SQL: Inherit from OWBaseSql

VesnaT:
##### Description of changes
- refactor SQL Table widget so it extends OWBaseSql
- fix crash when postgres is not installed 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/sql/backend/postgres.py
+++ b/Orange/data/sql/backend/postgres.py
@@ -4,7 +4,7 @@ import warnings
 from contextlib import contextmanager
 from time import time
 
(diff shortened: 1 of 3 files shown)


lanzagar:
Connecting to the database leaves the selection in the Tables dropdown the same as it was before, but the output gets cleared. This results in an inconsistent state (table selected, but not output).

lanzagar:
Backend selection in dropdown is not a setting (while the server below is).
This is annoying and even leads to saved workflows not opening correctly.
But that can be addressed in another PR... I'll open an Issue.

Issue #4094, opened 2019-10-10
Box plot freezes when feature has many values


PrimozGodec:
**Describe the bug**
Box plot freezes when a feature has many values and is selected in the subgroup menu.

**To Reproduce**
1. Save telecom-customer-churn dataset locally
2. Change CustomerID to be a discrete variable
3. Open the dataset with File widget
4. Connect box plot
5. If it does not freeze directly change subgroup value to csutomerID

**Orange version:**
3.24dev

**Expected behavior**
Orange would not freeze

**Operating system:**
Any

**Additional context**
The idea for a fix: The box plot does not need to plot a line for each value in those cases. Let's limit the number of plotted subgroups to let say first 100.

janezd:
Many widgets would crash like this.

Alternative, general fix: prohibit discrete variables with > 100 values. These should be text variables. If there are any widgets that could do something useful with such a variable and they don't accept text variables, let them start accepting text variables.

astaric:
Opsi car registration data (https://orange.biolab.si/blog/2017/10/13/diving-into-car-registration-data/) can be nicely visualized in BoxPlot, using Municipality/Car model as a subgroup and Sort by subgroup checked. (I was (ab)using boxplot since distributions widget does not scale too this many values, while boxplot with its slider does).
There are 213 municipalities in the dataset, 107 car models, and boxplot is fast enough. For this usecase, limiting the plot to first X values would be a preferable option to not being able to subgroup by Municipality at all.

janezd:
What about this: if the file widget outputs a discrete variable with > 100 values, it would warn `"Categorical variables with >100 values may decrease performance."`?

PR #4120, opened 2019-10-18
[FIX] Warning for discrete variable with >100 values in OWFile

PrimozGodec:
##### Issue

Fixes https://github.com/biolab/orange3/issues/4094

##### Description of changes
As @janezd suggested the variable with more than 100 different values should not be DiscreteVariable. This PR:

- changes guessing mechanism such that the variable with more than 100 different values is not a StringVariable
- add tests for `guess_data_type` function
- `OWFile` warns the user who set the variable with more than 100 different values as a DiscreteVariable

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/tests/test_io.py
+++ b/Orange/data/tests/test_io.py
@@ -1,0 +1,70 @@
+import unittest
+import numpy as np
+
+from Orange.data import ContinuousVariable, DiscreteVariable, StringVariable
+from Orange.data.io import guess_data_type
+
+
+class TestTableFilters(unittest.TestCase):
+
+    def test_guess_data_type_continuous(self):
+        # should be ContinuousVariable
+        valuemap, values, coltype = guess_data_type(list(range(1, 100)))
+        self.assertEqual(ContinuousVariable, coltype)
+        self.assertIsNone(valuemap)
+        np.testing.assert_array_equal(np.array(list(range(1, 100))), values)
+
+        valuemap, values, coltype = guess_data_type([1, 2, 3, 1, 2, 3])
+        self.assertEqual(ContinuousVariable, coltype)
+        self.assertIsNone(valuemap)
+        np.testing.assert_array_equal([1, 2, 3, 1, 2, 3], values)
+
+        valuemap, values, coltype = guess_data_type(
+            ["1", "2", "3", "1", "2", "3"])
+        self.assertEqual(ContinuousVariable, coltype)
+        self.assertIsNone(valuemap)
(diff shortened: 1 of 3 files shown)


janezd:
As you noted, there are still many ways to force variables to have >100 values. Having a warning in `DiscreteVariable.__init__` doesn't do any good (nobody sees it), and raising an exception would be worse (nobody would catch it).

I wrote a comment in #4094 - maybe the file widget would just give a warning?

Automated detection should of course recognize such variables as strings. Wasn't it already?

ajdapretnar:
> Automated detection should of course recognize such variables as strings. Wasn't it already?

It did, I am sure of it.

I also remember some users asking to increase the number of allowed discrete variables, mostly biologists, geologists and physicists.

PrimozGodec:
> Automated detection should, of course, recognize such variables as strings. Wasn't it already?

Automated detection recognizes variables that have `int(round(len(values)**.7)) < num_unique_variables` as string where `len(values)` is number of items in the table. This threshold is dependent on the number of items size (which looks ok). Last time we had a case when we had about 30000 items and for this data table, this threshold is 1361. Orange recognized a variable that had a few hundred unique values as a discrete and the box-plot freeze. I know this case is very unusual but it can happen. 

Should we put a threshold of 100 hundred unique values besides the current threshold for variable recognition or we leave it as it is? 

I agree with the warning in the File widget.

PrimozGodec:
I removed the warning from the Discrete variable and added the Warning on the File widget. Since now there is a warning on the widget for discrete variables with more than 100 values I suggest that we keep the same rule for guessing (see the previous comment).

janezd:
Domain editor is a component and as such it shouldn't expect the class into which it is inserted to have some specific warning that the component will trigger.

- I think the best solution would be to raise the warning from within OWFile widget. Currently it warns only if the user manually changes the type to categorical. The warning should also be shown in .tab files if the column is specified as discrete and contains too many values, not just if the user overrides the guessing heuristic.

- If we keep the current solution, I believe it would be better if domain editor itself declared the warning (we therefore treat it as a kind of "mixin"). You can do this by removing 
`performance_warning = widget.Msg("Categorical variables with >100 values may decrease performance.")` from widget's warnings and instead add 
```
        widget.Warning.add_message(
            "performance_warning",
            "Categorical variables with >100 values may decrease performance.")
``` to domain editor's `__init__`.

I believe that the former is better, though.

PrimozGodec:
I moved the warning triggering to the file widget and it shows the warning if the column in the file is marked as a discrete.

PR #4123, opened 2019-10-20
[ENH] Nice binning of time variables (Distributions, SOM)

### LINKED ISSUE #3964: Distributions: binning datetime
**Describe the bug**
Distributions doesn't bin datetime variables correctly.

**To Reproduce**
Datasets (Philadelphia Crime or any timeseries data set) - Distributions.

**Orange version:**
3.23.dev

**Expected behavior**
Bin width should not be hidden behind the slider and it should be datetime (i.e. 2019-08-01).
X axis in the plot should be datetime.

**Screenshots**
Example of Kennedy's speeches. Bins are not correctly computed. The final bin contains only 1 item, even though the speeches come in fairly equal intervals between 1960 and 1963. Bin should probably be <1962 and >=1962.

Screen Shot 2019-08-02 at 13 39 08

janezd:
Fixes #3964.

##### Description of changes

- provides a general `time_binning` function, similar to `decimal_binning`, which can be used in any visualization widget that discretizes numeric features
- reverses the order of returned binnings in `decimal_binning`
- improves the `BinDefinition`
- uses the above in Distributions in SOM (the two widgets that already use `decimal_binning`)

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


ajdapretnar:
I ❤️ it!

ajdapretnar:
So far I was not able to find any faults with it. It is responsive, labels are correct, output is as expected. This makes Distributions really useful for visualizing datetime data!

The only thing that doesn't seem to be implemented yet is the legend in SOM - it shows floats for datetime.

Orange/widgets/unsupervised/owsom.py
```diff
@@ -828,13 +828,19 @@ def update_output(self):
...
-            self.thresholds = self.colors = None
+            self.thresholds = self.bin_labels = self.colors = None
         elif self.attr_color.is_discrete:
-            self.thresholds = None
+            self.thresholds = self.bin_labels = None
             self.colors = [QColor(*color) for color in self.attr_color.colors]
         else:
             col = self.data.get_column_view(self.attr_color)[0].astype(float)
-            self.thresholds = decimal_binnings(col, min_bins=4)[0][1:-1]
+            binning = decimal_binnings(col, min_bins=4)[-1]
```
ajdapretnar: I think you forget to check whether self.attr_color is a Time Variable. This doesn't use time_binnings.

Orange/preprocess/discretize.py
```diff
@@ -278,40 +297,174 @@ def unique_bins():
...
-    if add_unique is not None and len(unique) <= add_unique:
-        if bins and bins[-1].nbins == len(unique):
-            del bins[-1]
-        bins.append(unique_bins())
-        if len(unique) < min_unique:
-            del bins[:-1]
 
-    if not return_defs:
-        bins = [get_bins(bin) for bin in bins]
+def time_binnings(data, *, min_bins=2, max_bins=50, min_unique=5, add_unique=0):
```
ajdapretnar: The output of `time_binnings` and `decimal_binnings` are not the same. The latter uses `BinDefinition`, which has thresholds and labels properties, while the former doesn't.

janezd: This can't be. OWDistributions calls either `time_binnings` or `decimal_binnings`, stores the result and then doesn't care where the bins came from.

Orange/widgets/unsupervised/owsom.py
janezd: Oops. I just ported it to new signature of `decimal_binning`. It works now.

I'm not sure I like the legend. It uses short labels.

Screenshot 2019-10-21 at 16 32 15

Perhaps the lower bound for each interval should be in the long form. That is, the last two should be `17 Apr - Jul` and `>= 17 Jul`. The second would be in short form, except of course when the longer is needed, like in `16 Oct - 17 Jan`.

Orange/preprocess/discretize.py
ajdapretnar: But OWSOM doesn't.

Orange/widgets/unsupervised/owsom.py
janezd: Like this.

Screenshot 2019-10-21 at 19 27 15

Orange/preprocess/discretize.py
janezd: This was OWSOM's problem. Fixed.

PR #4127, opened 2019-10-21
SOM: check if data has single attribute

### LINKED ISSUE #4126: OWSOM: crash on single column
**Describe the bug**
SOM crashes when given data with a single column.

**To Reproduce**
1. File (iris)
2. Select Columns (remove all but one attribute=
3. Connect to SOM.

**Expected behavior**
A warning appears.

ajdapretnar:
##### Issue

Fixes #4126. 

##### Description of changes
Check if data has single attribute.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owsom.py
+++ b/Orange/widgets/unsupervised/owsom.py
@@ -211,6 +211,7 @@ class OWSOM(OWWidget):
             Msg("Some data instances have undefined value of '{}'.")
         missing_values = \
             Msg("{} data instance{} with undefined value(s) {} not shown.")
+        single_attribute = Msg("Data contains a single numeric column.")
 
     class Error(OWWidget.Error):
         no_numeric_variables = Msg("Data contains no numeric columns.")
@@ -299,6 +300,8 @@ class OWSOM(OWWidget):
         def prepare_data():
             if len(cont_attrs) < len(attrs):
                 self.Warning.ignoring_disc_variables()
+            if len(cont_attrs) == 1:
+                self.Warning.single_attribute()
             x = Table.from_table(Domain(cont_attrs), data).X
             if sp.issparse(x):
                 self.data = data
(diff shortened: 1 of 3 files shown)


janezd:
There is no reason why SOM wouldn't work with a single variable. (It doesn't make sense to use SOM then, but this is not a reason. :)

#4126 happens because initial weights cannot be initialized with PCA when the data has just a single dimension. If you switching to random initialization, SOM works even with a single attribute.

The proper fix, I think, is to switch to random initialization in this case. We already do the same for sparse data.

```
--- a/Orange/projection/som.py
+++ b/Orange/projection/som.py
@@ -50,7 +50,7 @@ class SOM:
                 f(self.weights, x,
                   sigma / decay, learning_rate / decay)

-        if self.pca_init and not sp.issparse(x):
+        if self.pca_init and not sp.issparse(x) and x.shape[1] > 1:
             self.init_weights_pca(x)
         else:
             self.init_weights_random(x)
```

Orange/widgets/unsupervised/owsom.py
```diff
@@ -338,6 +339,8 @@ def set_warnings():
             cont_attrs = [var for var in attrs if var.is_continuous]
             if not cont_attrs:
                 self.Error.no_numeric_variables()
+            elif len(attrs) == 1:
+                self.Error.single_attribute()
```
janezd: But if we follow this route, check `len(cont_attrs)` not `len(attrs)`.

ajdapretnar: No, I prefer your approach. We are not to block people from doing silly things. :)

ajdapretnar:
I decided for a warning and checking the attributes at projection level.

Orange/widgets/unsupervised/owsom.py
janezd: OK, but you must check `cont_attrs`, not `attrs` (and change the message from `(...) single column` to `(...) single numeric variable`). It doesn't help if the data has 1000 categorical features but a single numeric.

ajdapretnar: It was before my coffee. I was sure now I am checking for cont_attrs.

PR #4150, opened 2019-10-29
OWScatterPlotBase: fix unaligned items on legend

robertcv:
##### Issue
For categorical: symbols on legend are not aligned with text.
For numerical: pallet and text are not in the center of legend, also legend window too wide.
Examples are from Dataset(Illegal waste dumpsites in Slovenia) -> Scatter Plot, then select "Location type" for categorical and "Waste volume" for numerical

Comparison:
catagorical_old -> catagorical_new
numerical_old -> numerical_new

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -15,7 +15,8 @@ from AnyQt.QtWidgets import (
 )
 
 import pyqtgraph as pg
-import pyqtgraph.graphicsItems.ScatterPlotItem
+import pyqtgraph.functions as fn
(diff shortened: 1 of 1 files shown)


robertcv:
This is something that is probably system and screen dependent so I would ask for as many testers as possible.

janezd:
On my machine it is not aligned. `translate(8, 12)` looks better than proposed `translate(8, 14)`.

This is not exact science. If this works OK for @thocevar and @lanzagar, we can keep it as it is.

lanzagar:
12 is better than 14 on my system too!

thocevar:
14 is way off for me. From left to right: current, proposed 12, proposed 14

image

markotoplak:
I'd vote for current.

PR #4151, opened 2019-10-31
Use OWSaveBase for Save Distances and Save Model

### LINKED ISSUE #4027: Derive Save distances and Save Model from the common base for save widgets
@irgolic noticed in #4024 that Save Distances (ab?)uses auto_commit for saving.

After #3994, it would be nice if all save widgets used the same base class.

janezd:
Closes #4027.

Review by commits.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation


Orange/widgets/data/owsave.py
```diff
@@ -55,53 +93,120 @@ def __init__(self, start_row=0):
...
-            return
-        self.filename = filename
-        self.filter = selected_filter
-        self.last_dir = os.path.split(self.filename)[0]
-        self.bt_save.setText(f"Save as {os.path.split(filename)[1]}")
-        self.update_messages()
-        self.save_file()
+    def do_save(self):
+        """
+        Do the savingg.
```
lanzagar: savingg

Orange/widgets/tests/base.py
```diff
@@ -37,7 +37,23 @@
...
+        np.testing.assert_array_equal(table1.Y, table2.Y)
+        np.testing.assert_array_equal(table1.metas, table2.metas)
+
+    def assert_domain_equal(self, domain1, domain2):
+        for var1, var2 in zip(domain1.variables + domain1.metas,
+                              domain2.variables + domain2.metas):
+            self.assertEqual(type(var1), type(var2))
+            self.assertEqual(var1.name, var2.name)
+            if var1.is_discrete:
+                self.assertEqual(var1.values, var2.values)
```
lanzagar: Currently, this is different than `domain1 == domain2`. Maybe we should fix `DiscreteVariable` to check `values` in `__eq__`.
But despite that, do we really want to "reimplement" domain equality here? Is it for debugging purposes (a specific assert fails in case of nonequality)?

janezd: I added a docstring.

PR #4158, opened 2019-11-02
gui: Remove comboBox, use the one from orange-widget-base

janezd:
##### Issue

https://github.com/biolab/orange-widget-base/pull/21 eliminates `valueType` argument from `comboBox`. This PR removes the argument from calls.

https://github.com/biolab/orange-widget-base/pull/21 also implements full comboBox function in orange-widget-base, so the function no longer needs to be reimplemented in this repo.

This slightly modifies behaviour of combo boxes: previously, a box could have assigned a model, but use ints as settings. This was invalid and no longer works. Only one core widget did this, and the error should be easily detected by tests, probably already at widget's initialization.

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -39,7 +39,6 @@ from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.utils.concurrent import ThreadExecutor, TaskState
 from Orange.widgets.widget import OWWidget, Msg, Input, Output
-from orangewidget.utils.itemmodels import PyListModel
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


lanzagar:
Looks good, I just don't understand the second commit (fix imports) completely.
`OrangeComboBox` was added for backcompat, since it was removed above in the first commit, but what is with removing `LAMBDA_NAME`, `disable_opposite` and moving `_addSpace` into its own block (which looks the same)?
Ready to merge otherwise, just curious :)

janezd:
All three are unused, but `_addspace` was (and still is) mentioned in docs. It's moved to a separate block so that pylint-disable applies only to it and not to other imports.

We should probably just remove it from documentation and remove the import. Right?

lanzagar:
But it was moved out of a block that has the same pylint disable...
Anyway, it starts with an underscore and is never used - I think we can remove it from the docs, yes :)

janezd:
> But it was moved out of a block that has the same pylint disable...

Huh? I don't recall how/why.

I now removed it completely.

PR #4161, opened 2019-11-04
Updated icons

larazupan:
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
I don't think Pivot is entirely clear. It looks like this should be the icon for Transpose, not Pivot. 🤷‍♀

ajdapretnar:
I see this is the icon that Excel uses, too. 😱 Then I guess we should leave it.

ajdapretnar:
Classic me. I was trying to solve this, but only managed to mess it up. 🤦‍♀

PR #4169, opened 2019-11-06
OWRandomize: data info displayed in status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owrandomize.py
+++ b/Orange/widgets/data/owrandomize.py
@@ -67,6 +67,9 @@ class OWRandomize(OWWidget):
             box, self, "random_seed", "Replicable shuffling",
             callback=self._shuffle_check_changed)
 
+        self.info.set_input_summary(self.info.NoInput)
+        self.info.set_output_summary(self.info.NoInput)
+
         self.apply_button = gui.auto_apply(self.controlArea, self, box=False, commit=self.apply)
 
     @property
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owrandomize.py
```diff
@@ -87,6 +90,8 @@ def _set_scope_label(self):
     def set_data(self, data):
         self.data = data
         self.unconditional_apply()
+        self.info.set_input_summary(str(len(data))) if data else \
```
VesnaT: That is actually an assignment expression, but result is not assigned to anything. You can e.g. create a new variable and use it to set the summary.
`text = str(len(data)) if self.data else self.info.NoInput`

Orange/widgets/data/tests/test_owrandomize.py
```diff
@@ -74,6 +74,25 @@ def test_replicable_shuffling(self):
...
+        output_sum.assert_called_with(str(len(output)))
+        input_sum.reset_mock()
+        output_sum.reset_mock()
+
+        self.send_signal(self.widget.Inputs.data, None)
+        input_sum.assert_called_once()
+        self.assertEqual(input_sum.call_args[0][0].brief, "")
+        output_sum.assert_called_once()
+        self.assertEqual(output_sum.call_args[0][0].brief, "")
+
```
VesnaT: One blank line too much.

Orange/widgets/data/owrandomize.py
```diff
@@ -100,6 +105,8 @@ def apply(self):
             data = self.data.copy()
             for i, instance in zip(indices, randomized):
                 data[i] = instance
+        self.info.set_output_summary(str(len(data))) if data else\
```
VesnaT: Same as above.

PR #4180, opened 2019-11-11
[WIP][ENH] Output to OWCorrespondence

### LINKED ISSUE #4088: Output for correspondence analysis
Hi,
It would be cool to have an output with the coordinates of the points in the correspondence analysis widget so we may export them and possibly plot them also outside Orange.
Thank you for all the great stuff in Orange!

PrimozGodec:
##### Issue

Fixes #4088

##### Description of changes
Added output with correspondence analysis coordinates. I added coordinates off all components not just selected two.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owcorrespondence.py
+++ b/Orange/widgets/unsupervised/owcorrespondence.py
@@ -7,7 +7,7 @@ from AnyQt.QtGui import QBrush, QColor, QPainter
 from AnyQt.QtCore import QEvent, QItemSelectionModel, QItemSelection
 
 import pyqtgraph as pg
-import Orange.data
+from Orange.data import Table, Domain, ContinuousVariable, StringVariable
 from Orange.statistics import contingency
 
 from Orange.widgets import widget, gui, settings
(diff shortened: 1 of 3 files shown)


ajdapretnar:
Thanks Primož! I would just like to confirm with @ales-erjavec that his work on Correspondence Analysis has been put on hold for the foreseeable future. Aleš was doing a complete rehaul of CA, so I would like to see where we are with that.

ales-erjavec:
> ... that his work on Correspondence Analysis has been put on hold for the foreseeable future.

Yes

Orange/widgets/unsupervised/owcorrespondence.py
```diff
@@ -8,14 +8,16 @@
 
 import pyqtgraph as pg
 import Orange.data
+from Orange.data import Table, Domain, ContinuousVariable, StringVariable
```
ajdapretnar: I don't think the above `import Orange.data` is necessary, since you are importing everything here. Also in this case replace all instances of `Orange.data.Table` with `Table`.

Orange/widgets/unsupervised/owcorrespondence.py
```diff
@@ -8,14 +8,16 @@
...
 from Orange.statistics import contingency
 
 from Orange.widgets import widget, gui, settings
 from Orange.widgets.utils import itemmodels, colorpalette
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 
 from Orange.widgets.visualize.owscatterplotgraph import ScatterPlotItem
-from Orange.widgets.widget import Input
+from Orange.widgets.widget import Input, Output
+from orangewidget.settings import Setting
```
ajdapretnar: Should this not be `from Orange.widgets.settings import Setting`?

PR #4190, opened 2019-11-14
Adapt usage statistics for orange-canvas-core 0.1.9

irgolic:
To be merged after https://github.com/biolab/orange-canvas-core/pull/49.

##### Description of changes
Call the `UsageStatistics.load()` static function instead of instantiating the class and calling the method.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -160,7 +160,6 @@ def open_link(url: QUrl):
 
             if not settings.contains('reporting/machine-id'):
                 settings.setValue('reporting/machine-id', str(uuid.uuid4()))
-        pass
     else:
         QDesktopServices.openUrl(url)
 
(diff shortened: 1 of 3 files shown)


markotoplak:
What does that mean for Orange and canvas core compatibility? If Orange now requires newer (next release) of canvas-core, please, also raise requirements in this PR. And if so, why don't the tests fail?

But pre-this-PR Orange still works with new canvas-core, right? What is missing in current Orange if new canvas-core is released and this PR is not yet accepted?

markotoplak:
We discussed this with @irgolic. A combination of old orange3 and new orange-canvas-core would not send any statistics but also would not crash. We decided that this is fine.

This PR can be accepted after a new release of orange-canvas-core. Everything needed is already in merged to its master branch.

PR #4212, opened 2019-11-22
Feature Statistics: Fix coloring of the last bin

### LINKED ISSUE #4195: Feature statistics colour scale
Orange 2.23, linux
When colouring by a feature in Feature statistics the colour scale of last bin is off (eg. same as at the left side of histogram). This does not happen in scatter plot. 
 Example: colouring by petal length in Orange iris dataset - look at histogram of same feature.

Screenshot from 2019-11-15 15-42-29

janezd:
##### Issue

Fixes #4195.

##### Description of changes

The code that computed bins for coloring produced just 9 bins because data instances with values equal to the last edge were put in the previous bin because of `right=True`. Removing the argument solves the issue.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/data/utils/histogram.py
+++ b/Orange/widgets/data/utils/histogram.py
@@ -359,9 +359,7 @@ class Histogram(QGraphicsWidget):
 
             bins = np.arange(self.n_bins)[:, np.newaxis]
             edges = self.edges if self.attribute.is_discrete else self.edges[1:-1]
-            # Need to digitize on `right` here so the samples will be assigned
-            # to the correct bin for coloring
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/utils/histogram.py
```diff
@@ -361,7 +361,7 @@ def _get_colors(self):
             edges = self.edges if self.attribute.is_discrete else self.edges[1:-1]
             # Need to digitize on `right` here so the samples will be assigned
```
markotoplak: Is perhaps this comment now misleading?

PR #4219, opened 2019-11-25
[ENH] ConcurrentWidgetMixin: Cancel task on input change

VesnaT:
##### Issue

ConcurrentWidgetMixin subclasses don't stop computation when an input changes.

##### Description of changes
Set `invalidated` widget state (instead of `blocking` state) when stopping/starting the task. 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owtsne.py
+++ b/Orange/widgets/unsupervised/owtsne.py
@@ -513,7 +513,6 @@ class OWtSNE(OWDataProjectionWidget, ConcurrentWidgetMixin):
 
     def __ensure_task_same_for_pca(self, task: Task):
         assert self.data is not None
-        assert task.data is self.data
         assert task.normalize == self.normalize
         assert task.pca_components == self.pca_components
         assert isinstance(task.pca_projection, Table) and \
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -563,9 +563,10 @@ class ProjectionWidgetTestMixin:
         properly set/updated"""
         self.send_signal(self.widget.Inputs.data, self.data)
 
-        if self.widget.isBlocking():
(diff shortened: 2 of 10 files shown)


ales-erjavec:
If i connect *File* -> *t-SNE*, and load iris.tab then housing.tab, then move focus (tab) to the 'Recent' combo box and spam the down arrow key, eventually I get a:

```
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/unsupervised/owtsne.py", line 551, in on_partial_result
    self.__ensure_task_same_for_pca(task)
  File "/Users/aleserjavec/workspace/orange3/Orange/widgets/unsupervised/owtsne.py", line 516, in __ensure_task_same_for_pca
    assert task.data is self.data
AssertionError
```

janezd:
Please rebase to current master.

Orange/widgets/unsupervised/owtsne.py
```diff
@@ -513,7 +513,7 @@ def run(self):
 
     def __ensure_task_same_for_pca(self, task: Task):
         assert self.data is not None
-        assert task.data is self.data
+        np.testing.assert_array_equal(task.data.X, self.data.X)
```
ales-erjavec: `numpy.testing` should not be used outside of unittests.

I am assuming the cause of this assert violation is a [check](https://github.com/biolab/orange3/blob/06f7630d8753143f4c7f6b029c7dc79264d76318/Orange/widgets/visualize/utils/widget.py#L453) that the data is the same upon receiving the input but only after having already updated `self.data`. 

If that is the case I would just remove the assert.

VesnaT: It's removed.

PR #4222, opened 2019-11-26
[FIX] Feature Constructor: Compatibility with Python 3.8

lanzagar:
##### Issue

Feature Constructor did not work on Python 3.8 due to AST changes.

##### Description of changes
- Add the newly required argument `posonlyargs`
As far as I can tell, this does not cause any problems on older versions of python and is silently ignored (always empty anyway).
- Add support for new expression type `Constant` (which is replacing `Num`, `Str`, ...)


ajdapretnar:
This seems to work fine on Py3.7.

lanzagar:
@ales-erjavec can you check the code in case I missed something before we merge?

PR #4230, opened 2019-11-28
Provide a separate README for PyPi

ajdapretnar:
##### Issue

README for Github and README for PyPi should not be the same. The description should be shorter, without Github tags.

##### Description of changes
Provide a separate PyPi README.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -17,6 +17,7 @@ recursive-include distribute *.svg *.desktop
 include requirements*.txt
 
 include README.md
+include README.pypi
 include CONTRIBUTING.md
 include CHANGELOG.md
 include COPYING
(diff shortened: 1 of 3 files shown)


ajdapretnar:
Seems like markdown should render just fine on PyPi, but it doesn't for Orange. See: https://pypi.org/project/Orange3/

I have found the issue on pypa: https://github.com/pypa/warehouse/issues/3664
[PyPi](https://packaging.python.org/guides/making-a-pypi-friendly-readme/#validating-restructuredtext-markup) says to use
```
setuptools >= 38.6.0
wheel >= 0.31.0
twine >= 1.11.0
```
which I believe who ever is in charge of uploading to PyPi already does. Could it be something else?

PR #4231, opened 2019-11-28
[FIX] Various fixes of box plot

PrimozGodec:
##### Issue

While working on some data I found few bugs in the Box Plot widget. Some issues were produced by https://github.com/biolab/orange3/pull/4135 some are old. 

- [x] Widget failed with index error since one of the arrays unexpectedly changed the shape when subgroup had just missing values
- [x] Contingency is not correctly computed when row_variable is sparse
- [x] Axis does not stretch correctly when many missing values
- [x] Missing values not plotted when a subgroup is not selected
- [x] Continuous variable raises an error in some cases when the group variable has nans.

##### Description of changes
All of the above errors fixed and added tests

Important before you test run `pip install -e .` in Orange directory since Cython code was changed and C code needs to be compiled.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


Orange/statistics/distribution.py
```diff
@@ -130,6 +130,19 @@ def from_data(cls, data, variable):
...
+        """
+        This property returns a distribution array with unknowns arrays added
+        at the end
+
+        Returns
+        -------
+        np.array
+            Array with concatenated unknowns as a row and column
+        """
+        return np.append(np.array(self), self.unknowns)
```
lanzagar: The documentation for append says it already works on the copy of the first arg - do you really need to make another copy? (or are you doing this to change it from a distribution to an array?)

PrimozGodec: It just to transform to array from distribution. Otherwise, it stays distribution which is not good since we get another distribution with unknown attributes. In this case, the unknowns attribute is joined together with other data so it needs to be an array.

lanzagar:
When stretch bars is disabled, and there are missing values in the subgroup with the largest bar, the size of that subgroup (written to the right of the bar) falls out of view.
(does not happen when there are no missing values)

PrimozGodec:
@lanzagar it should be fixed now.

PR #4234, opened 2019-11-28
[ENH] owheatmap: Add Split By combo box

ales-erjavec:
##### Issue

Heat map splits vertically by discrete class variable. This is hardcoded in the widget but not always desired.

##### Description of changes

Allow (rows) split by other class or meta categorical var (not just the class var), or disable spliting at all.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


ajdapretnar:
This is so great, I cannot even... 👌

PR #4235, opened 2019-11-28
[FIX] File widget: fix name change

PrimozGodec:
##### Issue

File widget fails on an attribute name change. It happens just in cases when only the name is changed and the type of variable stays the same.

##### Description of changes
Fixed name change using the copy function. It seems that the variable name is protected (no setter function).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


PrimozGodec:
@janezd it is ready now.

PR #4237, opened 2019-11-28
Table: Clean-up and deprecations in constructor

### LINKED ISSUE #4224: Empty `Table`
**Describe the bug**

As noted by @vpodpecan in #4221, *if an empty table is created, an exception will happen when pretty-printing it (the nonexisting X attribute is referenced):*

```
import Orange
a = Orange.data.Table()
a  # or print(a)
```

We should improve `str` or, perhaps better, prohibit calling constructor without arguments when `cls is Table`. Derived classes are allowed to have constructors without arguments. (Or are they? @lanzagar? This class is a mess...)

janezd:
##### Issue

Fixes #4224. Also tries to start cleaning up the mess in Table's constructors.

##### Description of changes

- Attributes that every `Table` must have are added is immutable class attributes.
- Implements class `frozendict` to be used for attribute `attributes`
- Warns that calls of table constructor are deprecated, except 

    - `Table(filename_or_url)` because `Table("iris.tab")` is short and unambiguous, 
    - `Table(table)` because it looks like an expected legit way to copy a table,
    - `Table()`, because constructors need it,
    - `Table(X, Y, metas, W, attributes, ids)` as a default-ish constructor.

The latter is seldom used and the penultimate is used only internally.

Documentation recommends using named constructors and doesn't mention calling `Table` directly. I'd keep it like this.

##### Includes
- [X] Code changes
- [X] Tests


lanzagar:
> Table(X, Y, metas, W, attributes, ids) as a default-ish constructor.

I would deprecate that one as well (and force explicit `from_numpy` calls) and instead have the "full" one as a default-ish constructor including `domain` as the first arg:
`Table(domain, X, Y, metas, W, attributes, ids)`

PR #4244, opened 2019-12-04
Continuize: Elaborate docs

### LINKED ISSUE #4243: Continuize "one class per value" does not work if n_classes=4
**Describe the bug**
When trying to use the continuize widget for _one-hot encoding_ of a categorical variable that has four possible values, columns are created for only three of the values 

**To Reproduce**
The enclosed example convincingly shows what's going wrong
[Workflow with input file.zip](https://github.com/biolab/orange3/files/3916946/Workflow.with.input.file.zip)
Screenshot 2019-12-03 at 14 50 01

**Orange version:**
3.23.1

**Expected behavior**
In the example above, there should have been an additional column for category-A

**Screenshots**
See above

**Operating system:**
Mac OSX 10.14.6

ajdapretnar:
##### Issue

Resolves #4243.

##### Description of changes
Be clearer about continuization.

##### Includes
- [X] Code changes
- [ ] Tests
- [X] Documentation

### DIFF
--- a/doc/visual-programming/source/widgets/data/continuize.md
+++ b/doc/visual-programming/source/widgets/data/continuize.md
@@ -17,7 +17,7 @@ The **Continuize** widget receives a data set in the input and outputs the same 
 
 1. [Continuization methods](https://en.wikipedia.org/wiki/Continuity_correction), which define the treatment of multivalued discrete attributes. Say that we have a discrete attribute status with the values low, middle and high, listed in that order. Options for their transformation are:
 
-   - **First value as base**: the attribute will be transformed into two continuous attributes, status=middle with values 0 or 1 signifying whether the original attribute had value middle on a particular example, and similarly, status=high. Hence, a three-valued attribute is transformed into two continuous attributes, corresponding to all except the first value of the attribute.
+   - **First value as base**: the attribute will be transformed into one less continuous variable as there are categorical values. The first value alphabetically will be taken as a base. The order of categorical variables can be changed in [Edit Domain](../data/editdomain).
    - **Most frequent value as base**: similar to the above, except that the data is analyzed and the most frequent value is used as a base. So, if most examples have the value middle, the two newly constructed continuous attributes will be status=low and status=high.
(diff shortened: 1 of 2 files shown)


doc/visual-programming/source/widgets/data/continuize.md
```diff
@@ -17,7 +17,7 @@ The **Continuize** widget receives a data set in the input and outputs the same
 
 1. [Continuization methods](https://en.wikipedia.org/wiki/Continuity_correction), which define the treatment of multivalued discrete attributes. Say that we have a discrete attribute status with the values low, middle and high, listed in that order. Options for their transformation are:
 
-   - **First value as base**: the attribute will be transformed into two continuous attributes, status=middle with values 0 or 1 signifying whether the original attribute had value middle on a particular example, and similarly, status=high. Hence, a three-valued attribute is transformed into two continuous attributes, corresponding to all except the first value of the attribute.
+   - **First value as base**: the attribute will be transformed into one less continuous variable as there are categorical values. The first value alphabetically will be taken as a base. The order of categorical variables can be changed in [Edit Domain](../data/editdomain).
```
janezd: I think the previous approach to explaining the method using an example (but not necessarily the text itself, e.g. "signifying" should be "indicating") was better.

Instructions for changing the order of variables could be moved to one paragraph earlier.

ajdapretnar: I think docs should first explain in general, what the feature is doing, and second how it applies in a particular use case. Saying 'the attribute will be transformed into two continuous attributes' is false for just about all the attributes where values != 3.

If anything, I think we could keep the example after the general description of the method.

The order of categorical values applies only to this particular method, does it not?

janezd: > The order of categorical values applies only to this particular method, does it not?

No, also to both methods that treat it as ordinal.

janezd:
Although `git blame` puts the blame on you, I am pretty sure that this text is mine, at least in its most horrible parts. I believe it must had been copied from some old documentation.

I started anew in #4251. I've recycled your text but also rewrote most of the rest.

PR #4248, opened 2019-12-05
[ENH] Merge Data: Implement context settings

### LINKED ISSUE #4210: Merge data remember Row matching
Merge data should remember Row matching feature names even when input data changes. For example, when inserting/removing preprocessing widgets or changing preprocessing applied by Single cell preprocessing before merging the Row matching features are forgotten. Same happens when data comes from selection on plots and an empty selection is made by mistake between selections. This can be very frustrating if there are hundreds of features which need to be scrolled through to reach the desired one - maybe a search box could be added as well.

Furthermore, sometimes extra data may come from different sources (e.g. clustering or tSNE) and has same feature names which could be used for merging. Thus an instance of Merge widget could remember what was last set for Row matching and apply it automatically if new data has features with such names.

janezd:
Fixes #4210.

##### Includes
- [X] Code changes
- [x] Tests

### DIFF
--- a/Orange/widgets/data/owmergedata.py
+++ b/Orange/widgets/data/owmergedata.py
@@ -11,7 +11,8 @@ import Orange
 from Orange.data import StringVariable, ContinuousVariable, Variable
 from Orange.data.util import hstack, get_unique_names_duplicates
 from Orange.widgets import widget, gui
-from Orange.widgets.settings import Setting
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I have tried several combinations of merging and they work as expected. This seems really nicely done.

PR #4249, opened 2019-12-05
Updated tests to sklearn=0.22.0

ajdapretnar:
##### Issue

Fixes failing tests.

##### Description of changes
Update clustering test to sklearn=0.22.0 and update the requirements.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/evaluation/clustering.py
+++ b/Orange/evaluation/clustering.py
@@ -35,22 +35,18 @@ class ClusteringScore(Score):
     # pylint: disable=arguments-differ
     def from_predicted(self, results, score_function):
         # Clustering scores from labels
-        # This warning filter can be removed in scikit 0.22
-        with warnings.catch_warnings():
-            warnings.filterwarnings(
-                "ignore", "The behavior of AMI will change in version 0\\.22.*")
-            if self.considers_actual:
-                return np.fromiter(
-                    (score_function(results.actual.flatten(),
-                                    predicted.flatten())
-                     for predicted in results.predicted),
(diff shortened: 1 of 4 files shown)


janezd:
Could you also fix the other test, if @lanzagar agrees that the change is innocent?

Suggestion: to avoid such problems in the future, we could change `[c.instances for c in clusters]` into `{frozenset(c.instances) for c in clusters}`, so the order will no longer matter.

ajdapretnar:
Yes, I was looking into this, but I am not sure where the issue appears. The order of clusters where exactly?

janezd:
As I understand, the function return a different *order of clusters*, but clusters are the same, so tests should pass.

Smallprint: however, saved workflows may no longer be compatible. If somebody saved a schema in which k-means' results were fed to Select Rows and there she selected the second cluster, it will no longer be the same cluster. We cannot include this in migrations for Select Rows, hence we can't fix this.

lanzagar:
For the remaining failing test, I meant that if due to some changes in the code in scikit-learn 0.22 the order of clusters is now different we could update our expected results to the new ones.
But I still think we want the results to be deterministic and always the same - after all we explicitly set the random seed to achieve this. So I would avoid comparing (unordered) sets in the test.
Bonus points for the person that finds the reason / change causing the new results.

janezd:
I agree. We have no control over the order, but it's still nice to be informed when it's changed.

I added the commit so it's tested and ready for rebases in the morning.

PR #4254, opened 2019-12-06
[FIX] Select Rows: Fix crash on changed variable type

### LINKED ISSUE #3952: Select rows crashes on changing input var type
##### Orange version

3.23.0 dev

##### Expected behavior
When changing input feature type for Select rows, it should remove/update appropriately conditions on it. 
 

##### Actual behavior
Crash. Widget tries to treat the feature as if it were still of the previous type.

Connect File widget to Select Rows. Add a condition for a numerical feature, it can be an empty one. Change the feature from numeric to category in File widget.

janezd:
##### Issue

Fixes #3952. Widget's settings stored variable names without type, and retrieved settings that did not really match.

##### Description of changes

Add variable types to settings.

~~Previous settings are not migrated because we cannot guess variable types in migration. A possible workaround would be to skip migration and instead add specific code for handling old settings in `decode_setting` and `match`.~~

Settings are not migrated; what crashed will still crash, but the only way to prevent this would be to clear settings entirely, which would harm more users than letting the widget crash for the few users who have existing settings and will change variables types.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -44,25 +44,38 @@ class SelectRowsContextHandler(DomainContextHandler):
         return varname in attrs or varname in metas
 
     def encode_setting(self, context, setting, value):
-        if setting.name == 'conditions':
-            CONTINUOUS = vartype(ContinuousVariable("x"))
-            for i, (attr, op, values) in enumerate(value):
-                if context.attributes.get(attr) == CONTINUOUS:
-                    if values and isinstance(values[0], str):
-                        values = [QLocale().toDouble(v)[0] for v in values]
-                        value[i] = (attr, op, values)
-        return super().encode_setting(context, setting, value)
+        if setting.name != 'conditions':
+            return super().encode_settings(context, setting, value)
+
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owselectrows.py
```diff
@@ -105,6 +124,8 @@ class Outputs:
     purge_classes = Setting(False, schema_only=True)
     auto_commit = Setting(True)
 
+    settings_version = 1
```
VesnaT: Should'n this be set to 2?

Orange/widgets/data/owselectrows.py
```diff
@@ -692,6 +713,14 @@ def send_report(self):
                  ("Non-matching data",
                   nonmatch_inst > 0 and "{} instances".format(nonmatch_inst))))
 
+    # Uncomment this on 2021/1/1
+    #
+    # @classmethod
+    # def migrate_context(cls, context, version):
+    #     if not version:
```
VesnaT: `if version is None or version < 2:`

PR #4255, opened 2019-12-06
[ENH] PCA: Add a signal with original data + pca

### LINKED ISSUE #3801: PCA: retain all data (Unification of embedding methods )
**Issue:** PCA discards the original attributes and replaces them with PCs. Often (e.g. in scOrange), it is desirable to use the original features: annotate PCA plots,  compute differentially-occurring features in different PCA regions, etc. 

**Proposed solution**: Place PCs in meta attributes and rename output to *Data*. 

This would be consistent to *t-SNE* and *MDS*. The user could use *Select columns* to restore current behaviour. This is substantially easier (number of PCs is typically small) than merging original attributes with principal components, as it's possible in the current version. 
 
Such an improvement would also play nice with the new widget *Apply domain*, which enables projecting data to existing PCs and/or attributes.

janezd:
##### Issue

Closes #3801.

##### Description of changes

To maintain backward compatibility, this adds a new signal instead of adding the original data as metas (ugly, not done in other widgets) or moving components to metas (incompatible).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owpca.py
+++ b/Orange/widgets/unsupervised/owpca.py
@@ -31,6 +31,7 @@ class OWPCA(widget.OWWidget):
 
     class Outputs:
         transformed_data = Output("Transformed Data", Table, replaces=["Transformed data"])
+        data = Output("Data", Table, default=True)
         components = Output("Components", Table)
         pca = Output("PCA", PCA, dynamic=False)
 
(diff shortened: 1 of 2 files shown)


BlazZupan:
Nice, separation of the signals and including principal components as meta features is a nice idea.

PR #4258, opened 2019-12-07
[FIX] Louvain Clustering: Update graph output for compatibility with the new network add-on

### LINKED ISSUE #3902: Louvain Clustering: compatibility with the current Network add-on
In the upcoming release, Network add-on will no longer depend on networkx. Hence
`from orangecontrib.network.network import Graph`
won't work.

Louvain Clustering should be updated to match the current status of the Network add-on. I think the current correct import is
`from orangecontrib.network import Network`

janezd:
Fixes #3902.

##### Includes
- [X] Code changes
- [ ] Tests


janezd:
The new code is covered by tests, but not on CI where tests are skipped because the network add-on is not installed.

PR #4259, opened 2019-12-07
[FIX] table_to_frame: metas lost on conversion

itslarin:
##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/pandas_compat.py
+++ b/Orange/data/pandas_compat.py
@@ -78,13 +78,16 @@ def table_from_frame(df, *, force_nominal=False):
                             np.column_stack(M) if M else None)
 
 
-def table_to_frame(tab):
+def table_to_frame(tab, include_metas=False):
     """
     Convert Orange.data.Table to pandas.DataFrame
 
     Parameters
     ----------
     tab : Table
+
+    include_metas : bool, (default=False)
+        Include table metas into dataframe.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


VesnaT:
Thank you for the PR.

Some comments:
- Have you thought of adding an argument to the `table_to_frame` i. e. `include_metas=False ` to ensure backward compatibility?
- Some unittests would be nice.

itslarin:
Hi, Vesna.

>>Have you thought of adding an argument to the table_to_frame i. e. include_metas=False to ensure backward compatibility?

In my personal opinion this parameter can cause confusion in further usage, i believe by default table supposed to be transformed into frame without any columns lost. The user always can drop some columns explicitly if he needs. There might be some incompatibility though in case somebody rely on output frame columns number (maybe in tests). Otherwise i guess additional (in comparison with prev. results) datarfame columns shouldnt provide any harm to existing code.

VesnaT:
All that is true and I agree with you, but still, the existing code should not be broken. 
Nobody knows how many add-ons are using the function, all expecting dataframes without metas. 
If you don't like the additional parameter, you can add it temporarily and raise a deprecation warning, to allow users to fix their code. The parameter can be removed in the next version.

janezd:
> ...  in the next version

Or a year. Better.

janezd:
Decision at the meeting: @VesnaT implements what she proposed here and merges the PR.

Issue #3891, opened 2019-06-18
Widget for statistical comparison of models


ghost:
##### Expected behavior
 
A widget should be added to compare model performances, using t-test, wilcoxon signed rank test etc.
##### Actual behavior

There is no currently available widget for statistical comparison of the model performances

drburakkocak:
At least, it should be possible to see all performance metrics in validation process. For instance, in 10-fold cross-validation, the program should show us the the metric for each fold.

janezd:
I'd rather not promote (frequentist) null-hypothesis testing in Orange (or elsewhere). It is not only conceptually wrong per se, but also doesn't fit well into data mining workflow because constantly reformulating your hypotheses invalidates the tests' premises.

It would be nice though to have a modern Bayesian alternative. Essentially this: https://baycomp.readthedocs.io/en/latest/functions.html#single-data-set, for pairs of methods on single data set. We don't have to depend on [baycomp](https://github.com/janezd/baycomp) for this, this function is just a Bayesian reinterpretation of the t-test.

janezd:
Implemented via #4261.

PR #4261, opened 2019-12-09
[ENH] Test & Score: Add comparison of models

janezd:
Solves #3891.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owtestlearners.py
+++ b/Orange/widgets/evaluate/owtestlearners.py
@@ -1,5 +1,6 @@
 # pylint doesn't understand the Settings magic
 # pylint: disable=invalid-sequence-index
+# pylint: disable=too-many-lines,too-many-instance-attributes
 import abc
 import enum
 import logging
@@ -9,14 +10,17 @@ from functools import partial, reduce
 
 from concurrent.futures import Future
 from collections import OrderedDict, namedtuple
+from itertools import count
 from typing import Any, Optional, List, Dict, Callable
 
 import numpy as np
+import baycomp
 
 from AnyQt import QtGui
-from AnyQt.QtGui import QStandardItem
 from AnyQt.QtCore import Qt, QSize, QThread
 from AnyQt.QtCore import pyqtSlot as Slot
+from AnyQt.QtGui import QStandardItem, QDoubleValidator
+from AnyQt.QtWidgets import QHeaderView, QTableWidget, QLabel
 
 from Orange.base import Learner
 import Orange.classification
@@ -35,7 +39,7 @@ from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.utils.concurrent import ThreadExecutor, TaskState
 from Orange.widgets.widget import OWWidget, Msg, Input, Output
-
+from orangewidget.utils.itemmodels import PyListModel
 
 log = logging.getLogger(__name__)
 
@@ -175,6 +179,10 @@ class OWTestLearners(OWWidget):
     fold_feature = settings.ContextSetting(None)
     fold_feature_selected = settings.ContextSetting(False)
 
+    use_rope = settings.Setting(False)
+    rope = settings.Setting(0.1)
+    comparison_criterion = settings.Setting(0)
+
     TARGET_AVERAGE = "(Average over classes)"
     class_selection = settings.ContextSetting(TARGET_AVERAGE)
 
@@ -275,12 +283,52 @@ class OWTestLearners(OWWidget):
             callback=self._on_target_class_changed,
             contentsLength=8)
 
+        self.modcompbox = box = gui.vBox(self.controlArea, "Model Comparison")
+        gui.comboBox(
+            box, self, "comparison_criterion", model=PyListModel(),
+            callback=self.update_comparison_table)
+
+        hbox = gui.hBox(box)
+        gui.checkBox(hbox, self, "use_rope",
+                     "Negligible difference: ",
+                     callback=self.update_comparison_table)
+        gui.lineEdit(hbox, self, "rope", validator=QDoubleValidator(),
+                     controlWidth=70, callback=self.update_comparison_table,
+                     alignment=Qt.AlignRight)
+
         gui.rubber(self.controlArea)
         self.score_table = ScoreTable(self)
         self.score_table.shownScoresChanged.connect(self.update_stats_model)
+        view = self.score_table.view
+        view.setSizeAdjustPolicy(view.AdjustToContents)
 
         box = gui.vBox(self.mainArea, "Evaluation Results")
         box.layout().addWidget(self.score_table.view)
+
+        self.compbox = box = gui.vBox(self.mainArea, box="Model comparison")
+        table = self.comparison_table = QTableWidget(
+            wordWrap=False, editTriggers=QTableWidget.NoEditTriggers,
+            selectionMode=QTableWidget.NoSelection)
+        table.setSizeAdjustPolicy(table.AdjustToContents)
+        header = table.verticalHeader()
+        header.setSectionResizeMode(QHeaderView.Fixed)
+        header.setSectionsClickable(False)
+
+        header = table.horizontalHeader()
+        header.setSectionResizeMode(QHeaderView.ResizeToContents)
+        avg_width = self.fontMetrics().averageCharWidth()
+        header.setMinimumSectionSize(8 * avg_width)
+        header.setMaximumSectionSize(15 * avg_width)
+        header.setTextElideMode(Qt.ElideRight)
+        header.setDefaultAlignment(Qt.AlignCenter)
+        header.setSectionsClickable(False)
+        header.setStretchLastSection(False)
+        box.layout().addWidget(table)
+        box.layout().addWidget(QLabel(
+            "<small>Table shows probabilities that the score for the model in "
+            "the row is higher than that of the model in the column. "
+            "Small numbers show the probability that the difference is "
+            "negligible.</small>", wordWrap=True))
 
     @staticmethod
     def sizeHint():
@@ -440,6 +488,8 @@ class OWTestLearners(OWWidget):
             self.scorers = []
             return
         self.scorers = usable_scorers(self.data.domain.class_var)
+        self.controls.comparison_criterion.model()[:] = \
+            [scorer.long_name or scorer.name for scorer in self.scorers]
 
     @Inputs.preprocessor
     def set_preprocessor(self, preproc):
@@ -453,6 +503,7 @@ class OWTestLearners(OWWidget):
         """Reimplemented from OWWidget.handleNewSignals."""
         self._update_class_selection()
         self.score_table.update_header(self.scorers)
+        self._update_comparison_enabled()
         self.update_stats_model()
         if self.__needupdate:
             self.__update()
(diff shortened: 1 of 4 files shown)


Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -275,13 +282,108 @@ def __init__(self):
...
+
+        header = table.horizontalHeader()
+        header.setSectionResizeMode(QHeaderView.ResizeToContents)
+        header.setDefaultAlignment(Qt.AlignLeft)
+        header.setStretchLastSection(False)
+        box.layout().addWidget(table)
+        box.layout().addWidget(QLabel(
+            "<small>Table shows probabilities that the score for the model in "
+            "the row is higher than that of the model in the column. "
+            "Small numbers show the probability that the difference is"
```
ajdapretnar: space after final 'is'

ajdapretnar: That said, that should probably be in the message above, not fixed below the table.

ajdapretnar:
This change makes Test&Score even more complicated. Is there a way to support this functionality outside of this widget? Perhaps as something that comes after Test&Score?

Orange/widgets/evaluate/owtestlearners.py
janezd: I added the space, but would prefer having this text visible.

ajdapretnar:
I applaud the documentation commit. 👏 Very diligent.

janezd:
> I applaud the documentation commit.

I pushed two commits. One adds a long-sought Bayesian test for comparison of classifiers in some 100 lines of reasonably decent code (not the best I ever wrote, but it's OK) accompanied with 150 lines of tests that systematically check all branches. I worked on it for, say, 2 days. The second commit adds a few sentences I wrote in ten minutes while waiting for some check-up in a hospital. And you applaud the latter commit?! What an encouragement!

Anyway, I wanted to ask you for an updated screenshot with two new numbers -- assuming the layout is confirmed (but save the workflow, just for a case).

ajdapretnar:
I clicked through the widget and it works well. A couple of comments, though:

1. The width of the columns is dependent on the learner name, so short learns get squeezed into substandard cells, while long names get really comfy. 😁 I propose minimum column width + eliding longer learner names.
Screen Shot 2020-01-10 at 10 53 10

2. When a single learner is on the input, the bottom part should probably be disabled.
Screen Shot 2020-01-10 at 10 54 23

3. Names of learners, both in rows and in columns, appear to be clickable, even though they're not. This can probably be disabled somewhere, so the GUI is not misleading.

4. [IDEA] How about coloring the cells according to the score? Say with light blue and the stronger the color the higher the score?

janezd:
I implemented 1-3, but I won't do the coloring. I agree this would be nice, but it requires too much machinery (going to models and defining delegates to either paint HTMLs or do the painting the hard way...) or it would be quick and dirty.

@VesnaT, you can review this as it is.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -275,13 +283,53 @@ def __init__(self):
...
+        self.modcompbox = box = gui.vBox(self.controlArea, "Model Comparison")
+        gui.comboBox(
+            box, self, "comparison_criterion", model=PyListModel(),
+            callback=self.update_comparison_table)
+
+        hbox = gui.hBox(box)
+        gui.checkBox(hbox, self, "use_rope",
+                     "Negligible difference: ",
+                     callback=self.update_comparison_table)
+        gui.lineEdit(hbox, self, "rope", validator=QDoubleValidator(),
```
VesnaT: Why is this not a spinbox? 
It should probably disabled when `use_rope` is not checked.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -562,6 +620,91 @@ def update_stats_model(self):
...
+
+        for row, row_name, row_scores in zip(count(), names, scores):
+            for col, col_name, col_scores in zip(range(row), names, scores):
+                if row_scores is None or col_scores is None:
+                    continue
+                if self.use_rope and self.rope:
+                    p0, rope, p1 = baycomp.two_on_single(
+                        row_scores, col_scores, self.rope)
+                    self._set_cell(table, row, col,
+                                   f"{p0:.3f}<br/><small>{rope:.3f})</small>",
```
VesnaT: `)` in 68. column should be removed.

Orange/widgets/evaluate/tests/test_owtestlearners.py
```diff
@@ -391,6 +399,204 @@ def test_no_pregressbar_warning(self):
...
+
+        rbs[OWTestLearners.KFold].click()
+        self.get_output(self.widget.Outputs.evaluations_results, wait=5000)
+        self.assertIsNotNone(w.comparison_table.cellWidget(0, 1))
+        self.assertTrue(w.modcompbox.isEnabled())
+        self.assertTrue(w.comparison_table.isEnabled())
+        baycomp.two_on_single.assert_called()
+        baycomp.two_on_single.reset_mock()
+
+    @patch("baycomp.two_on_single", Mock(wraps=baycomp.two_on_single))
```
VesnaT: No need to patch the `baycomp.two_on_single` for this test.

Orange/widgets/evaluate/tests/test_owtestlearners.py
```diff
@@ -391,6 +399,204 @@ def test_no_pregressbar_warning(self):
...
+        learner.name = "majd"
+        self.send_signal(w.Inputs.learner, learner, 1)
+        self.get_output(self.widget.Outputs.evaluations_results, wait=5000)
+        self.assertFalse(w.comparison_table.isEnabled())
+
+        rbs[OWTestLearners.KFold].click()
+        self.get_output(self.widget.Outputs.evaluations_results, wait=5000)
+        self.assertTrue(w.comparison_table.isEnabled())
+
+    @patch("baycomp.two_on_single", Mock(wraps=baycomp.two_on_single))
```
VesnaT: Same as above.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -562,6 +620,91 @@ def update_stats_model(self):
...
+        proxy = self.score_table.sorted_model
+
+        keys = (model.data(proxy.mapToSource(proxy.index(row, 0)), Qt.UserRole)
+                for row in range(proxy.rowCount()))
+        slots = [slot for slot in (self.learners[key] for key in keys)
+                 if slot.results is not None and slot.results.success]
+        return slots
+
+    def _scores_by_folds(self, slots):
+        scorer = self.scorers[self.comparison_criterion]()
```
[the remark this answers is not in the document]
janezd: Can it?

Orange/widgets/evaluate/owtestlearners.py
janezd: It's not a spin box because it has no defined range. It can refer to AUC, that is, between 0 and 1, or it can be RMSE, which is between 0 and infinity -- it can easily be 100000.

Disabling it would make sense, I'll do that.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -470,9 +521,16 @@ def shuffle_split_changed(self):
         self._param_changed()
 
     def _param_changed(self):
+        self.modcompbox.setEnabled(self.resampling == OWTestLearners.KFold)
+        self._update_comparison_enabled()
         self._invalidate()
         self.__update()
 
+    def _update_comparison_enabled(self):
+        self.comparison_table.setEnabled(
```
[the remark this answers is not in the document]
janezd: I like disabling it because it shows the user that it's intentionally blank. Otherwise it looks like a bug when the upper table is filled and this one isn't (e.g. when using Leave one out). Hiding would also be an option, though I like disabling better -- like "something could be here, but currently isn't because I can't compute it in this situation".

I can disable it when there is no data. But in this case we should do the same with the above table, I suppose. We need to discuss this.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -275,13 +283,53 @@ def __init__(self):
...
+        table.setSizeAdjustPolicy(table.AdjustToContents)
+        header = table.verticalHeader()
+        header.setSectionResizeMode(QHeaderView.Fixed)
+        header.setSectionsClickable(False)
+
+        header = table.horizontalHeader()
+        header.setSectionResizeMode(QHeaderView.ResizeToContents)
+        avg_width = self.fontMetrics().averageCharWidth()
+        header.setMinimumSectionSize(8 * avg_width)
+        header.setMaximumSectionSize(15 * avg_width)
```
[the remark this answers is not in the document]
janezd: That is, we can have fixed size. Alternatively, we can stretch all sections. What would you prefer? I prefer stretching. Can you please try `header.setSectionResizeMode(QHeaderView.Stretch)` (in line 318) and tell me whether you like it. It is a bit of a stretch for just two methods, while with three it's already OK.

Orange/widgets/evaluate/owtestlearners.py
```diff
@@ -611,6 +755,8 @@ def _invalidate(self, which=None):
                         item.setData(None, Qt.DisplayRole)
                         item.setData(None, Qt.ToolTipRole)
 
+        self.comparison_table.clearContents()
```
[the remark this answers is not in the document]
janezd: Fixed.

Orange/widgets/evaluate/owtestlearners.py
janezd: I added disabling but didn't like it. Let's let the user change the line edit first and then enable it, if (s)he wishes.

I added a method `_on_use_rope_changed`. You can add a line `self.controls.rope.setEnabled(self.use_rope)` and see for yourself that you won't like it. :)

Orange/widgets/evaluate/owtestlearners.py
janezd: Nice. Fixed.

Orange/widgets/evaluate/owtestlearners.py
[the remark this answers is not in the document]
VesnaT: I just noticed `self.comparison_criterion` setting does not work.
But if it worked and if someone removes scores from scoring module, it probably can.

Orange/widgets/evaluate/owtestlearners.py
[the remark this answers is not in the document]
VesnaT: I like it for three method. I don't like it for two and a single method, and it makes it inconsistent with the evaluation table.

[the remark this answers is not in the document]
janezd: Now it switches between Stretch and Fixed depending upon number of methods. Fixed has a default size of (about the same) as Logistic regression above (15 average chars). Stretch has a minimum size so that it starts scrolling when there are too many methods.

Orange/widgets/evaluate/owtestlearners.py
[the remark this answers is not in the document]
janezd: It is now a schema-only setting, with pending. I added an out-of-range check. If scorers are added or removed, the setting can still be wrong (because it's an index) but it won't crash. I don't find this issue too important.

Orange/widgets/evaluate/owtestlearners.py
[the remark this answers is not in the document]
janezd: Both views now disable under same conditions.

Orange/widgets/evaluate/owtestlearners.py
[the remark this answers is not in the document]
janezd: Not it's really fixed. Probably.

Orange/widgets/evaluate/owtestlearners.py
janezd: This is a won't fix. :)

VesnaT:
Why is the Evaluation Results table now disabled when one learner is on the input?

Right before Model comparison in performed, all the scores are shown in the left upper cell of the Model comparison table for a brief moment.

janezd:
> Why is the Evaluation Results table now disabled when one learner is on the input?

Because I'm stupid, but I've fixed that. (Now I'm smart.)

> Right before Model comparison in performed, all the scores are shown in the left upper cell of the Model comparison table for a brief moment.

Interesting. Fixed, too.

PR #4266, opened 2019-12-13
[ENH] Preprocess: Add filtering by missing values

AndrejaKovacic:
##### Description of changes
I extended Filter sparse features preprocessor with filtering columns by Nan's. We spoke about having 3 options, filter by 0, Nan's or both. I chose not to implement the third option, since the order of operation matters here and the user just use this preprocessor two times and have complete control that way.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/preprocess.py
+++ b/Orange/preprocess/preprocess.py
@@ -572,26 +572,43 @@ class PreprocessorList(Preprocess):
 
 class RemoveSparse(Preprocess):
     """
-    Remove sparse  features. Sparseness is determined according to
-    user-defined treshold.
+    Filter out the features with too many nan's or 0. Threshold is user defined.
 
     Parameters
     ----------
-    threshold : float
-        Minimal proportion of non-zero entries of a feature
+    filter_0: bool
+        filter out by zeros or nan's
+    fixed_threshold: bool
+        threshold is either a fixed number of elements or percentage
+    threshold: int
+        kept as is if fixed threshold or the percent is used to calculate
+        the appropriate number of elements
     """
 
-    def __init__(self, threshold=0.05):
+    def __init__(self, filter_0=True, fixed_threshold=False, threshold=5):
+        self.filter_0 = filter_0
+        self.fixed_threshold = fixed_threshold
         self.threshold = threshold
 
     def __call__(self, data):
-        if sp.issparse(data.X):
-            data_csc = sp.csc_matrix(data.X)
-            h, w = data_csc.shape
-            sparsness = [data_csc[:, i].count_nonzero() / h for i in range(w)]
+        if self.fixed_threshold:
+            tailored_threshold = self.threshold
         else:
-            sparsness = np.count_nonzero(data.X, axis=0) / data.X.shape[0]
-        att = [a for a, s in zip(data.domain.attributes, sparsness) if s >= self.threshold]
+            tailored_threshold = np.ceil(self.threshold/100 * data.X.shape[0])
+
+        if self.filter_0:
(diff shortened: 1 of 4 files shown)


Orange/widgets/data/owpreprocess.py
```diff
@@ -252,35 +252,103 @@ def __repr__(self):
 
 class RemoveSparseEditor(BaseEditor):
 
+    options = ["Nan's", "0's"]
```
ajdapretnar: NaNs and 0s, no apostrophe. I prefer missing instead of NaN, but that's ok, too. 0 should probably be written with a word, 'zeros'.

ajdapretnar:
Other preprocessors have a vertical layout, so I suggest placing the two filtering options vertically instead of horizontally.
Also, 'Select random features' and 'Select relevant features' have the threshold options as 'Fixed' and 'Percentage'. I think it is nice we have the same layout across all preprocessors. That said, there's an option in Text add-on where words are filtered by their absolute frequency if the input is an integer and by their relative value if the input is a float (e.g. 0.1 == 10%). Perhaps something we could have in preprocessors, too. Not sure about the user perspective here, though.

Orange/widgets/data/owpreprocess.py
janezd: I think we should use "missing". Muggles won't understand NaN.

Orange/preprocess/preprocess.py
```diff
@@ -572,26 +572,42 @@ def __call__(self, data):
...
-    user-defined treshold.
+    Filter out the features with too many nan's or 0. Threshold is user defined.
 
     Parameters
     ----------
-    threshold : float
-        Minimal proportion of non-zero entries of a feature
+    threshold: int or float
+        if >= 1, the argument represents the allowed number of 0s or NaNs;
+        if below 0, it represents the allowed proportion of 0s or NaNs
```
lanzagar: `below 0` -> `below 1`

Orange/preprocess/preprocess.py
```diff
@@ -572,26 +572,42 @@ def __call__(self, data):
...
+    filter0: bool
+        if True (default), preprocessor counts 0s, otherwise NoNs
```
lanzagar: `NoNs` -> `NaNs`

Orange/preprocess/preprocess.py
```diff
@@ -572,26 +572,42 @@ def __call__(self, data):
 
 class RemoveSparse(Preprocess):
     """
-    Remove sparse  features. Sparseness is determined according to
-    user-defined treshold.
+    Filter out the features with too many nan's or 0. Threshold is user defined.
```
lanzagar: Filter out features with too many (>threshold) zeros or missing values.

Orange/preprocess/preprocess.py
```diff
@@ -572,26 +572,42 @@ def __call__(self, data):
...
-        Minimal proportion of non-zero entries of a feature
+    threshold: int or float
+        if >= 1, the argument represents the allowed number of 0s or NaNs;
+        if below 0, it represents the allowed proportion of 0s or NaNs
+    filter0: bool
+        if True (default), preprocessor counts 0s, otherwise NoNs
     """
-
-    def __init__(self, threshold=0.05):
+    def __init__(self, threshold=5, filter0=True):
```
lanzagar: I suggest to leave the default value at `threshold=0.05` so it remains backwards compatible.
(this does not impact the widget as it always calls it with a defined threshold anyway)

PR #4267, opened 2019-12-13
[RFC] OWDistances: Include discrete attributes in cosine distance

### LINKED ISSUE #4200: Distances: Cosine similarity should handle categorical features
Take zoo data from the Dataset widget. All features are categorical. If this data set is passed to Distances, Euclidean distance works fine but Cosine fails. In principle, both types of distances require continuous features, so it is conceptually surprising that one distance type works on categorical data and the other one complains about it.

image

I propose to apply continuization preprocessor for cosine distance. If really needed, the widget can still throw a warning that categorical features were continuized, but in this way the widget and the cosine distance did what was expected: computed the distance from all the features.

janezd:
##### Issue

Fixes #4200, maybe.

@BlazZupan, the problem is that columns in cosine distances are not normalized, so variables must be from the same scale. (Cosine normalizes by rows, and mustn't normalize by columns -- think about bags of words.)

Indicator variables aren't on the same scale as other numeric variables. Take heart_disease and compute distances for all attributes or just for continuous: they are the same because discrete attributes don't contribute anything.

We can merge or we can warn that cosine on a mixture of attributes makes no sense. Or we can close this PR (now that we know why we haven't already had this).

However, one also shouldn't compute cosine on heart disease without normalization.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/distance/distance.py
+++ b/Orange/distance/distance.py
@@ -392,28 +392,33 @@ class Manhattan(FittedDistance):
 
 class Cosine(FittedDistance):
     supports_sparse = True  # via fallback
-    supports_discrete = False
+    supports_discrete = True
     fallback = SklDistance('cosine')
 
     @staticmethod
-    def discrete_to_indicators(x, discrete):
-        """Change non-zero values of discrete attributes to 1."""
-        if discrete.any():
-            x = x.copy()
-            for col, disc in enumerate(discrete):
-                if disc:
-                    x[:, col].clip(0, 1, out=x[:, col])
-        return x
+    def discrete_to_indicators(x, n_vals):
+        """One-hot encoding of discrete attributes"""
+        # This function is assumed to copy the data. Don't shortcut over it.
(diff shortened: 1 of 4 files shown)


janezd:
In discussion with @ajdapretnar (followin one with @BlazZupan and @lanzagar) we decided to better close this. While I usually oppose @lanzagar and advocate that Orange should be friendly, I would here side with @ajdapretnar: users must be aware that mixing discrete and continuous features in distances may be fishy.

If we ever revisit this in the future: this pull request does all except normalization. It would be easy to enable normalization for Cosine, but the problem is that the checkbox would normally be off for Cosine and on for Euclidean, so it would be too easy to make a mistake and unknowingly normalize cosine distances. 

An answer to the question why we allow categorical features in Euclidean and Manhattan but not in Cosine could be that mixing discrete and numeric features requires normalization. While normalization (almost) never hurts in Euclidean and Manhattan, it is inconsistent with the common use of cosine distances (in text mining).

PR #4269, opened 2019-12-15
[ENH] OWTreeViewer: Bold predicted values in tree nodes

irgolic:
##### Description of changes

I think this better visually differentiates a node's prediction from its next branching feature.

Screenshot 2019-12-15 at 23 15 37
Screenshot 2019-12-15 at 23 21 31

On a tangent, is the background color of a regression tree's nodes always this color by default? Either the background's or the separator line's color should be changed.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owtreeviewer.py
+++ b/Orange/widgets/visualize/owtreeviewer.py
@@ -358,7 +358,7 @@ class OWTreeGraph(OWTreeViewer2D):
         else:
             modus = np.argmax(distr)
             tabs = distr[modus]
-            text = f"{self.domain.class_vars[0].values[int(modus)]}<br/>"
+            text = f"<b>{self.domain.class_vars[0].values[int(modus)]}</b><br/>"
         if tabs > 0.999:
             text += f"100%, {total}/{total}"
         else:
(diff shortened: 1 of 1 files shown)


ajdapretnar:
You mean is the background always purple? Yes and no. You can set the node color in the Tree Viewer. By default it is colored by, unremarkably, default color, but you can change it to mean value and such. Perhaps the default is not the best... well, default.

irgolic:
@ajdapretnar I'll look into changing the default color.

On another note, in the regression tree, should we bold the errors if they're significant?

ajdapretnar:
> On another note, in the regression tree, should we bold the errors if they're significant?

I like the idea, but would they stand out enough considering the mean value is already bold?

irgolic:
Screenshot 2019-12-18 at 13 34 54

janezd:
> Also, can anyone explain why this PR is failing codecov?

The second of the two lines you've changed is not covered by tests, apparently. The coverage is not worse, but the diff coverage is < 95%.

Colors are indeed ugly.

As for marking significant differences, I don't like it because (a) it doesn't stand out and thus looks more like an inconsistency, (b) the user won't know why are they bold anyway, (c) significances could only be computed on separate data set, otherwise this is pure overfitting. Especially (c) is a problem.

PR #4280, opened 2019-12-19
OWDistance: Limit Mahalanobis to 1000 rows/columns

### LINKED ISSUE #4277: Should we remove the Mahalanobis distance from the Distances widget?
Before on a meeting with @lanzagar and @BlazZupan we discussed that Mahalanobis distance should be removed from the Distance widget. The problem appears when someone accidentally clicks at this distance in the dropdown. On slightly bigger data Orange takes a long time to compute the Mahalanobis distances and there is no other option than to restart the Orange in order to continue with work.

janezd:
##### Issue

Fixes #4277.

The problem occurs if we have too many rows when computing column-wise distances and vice-versa, because Mahalanobis must compute the inverse of the covariance matrix.

##### Description of changes

- When the dimensionality is too large, the widget shows and error and does not compute.
- Mahalanobis is moved towards the end of the list

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/unsupervised/owdistances.py
+++ b/Orange/widgets/unsupervised/owdistances.py
@@ -16,7 +16,6 @@ from Orange.widgets.widget import OWWidget, Msg, Input, Output
 METRICS = [
     ("Euclidean", distance.Euclidean),
     ("Manhattan", distance.Manhattan),
-    ("Mahalanobis", distance.Mahalanobis),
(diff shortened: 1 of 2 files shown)


janezd:
Move Mahalanobis to the bottom of the list.

PrimozGodec:
It looks good. Thank you.

PR #4282, opened 2019-12-19
Fix tests for 5.14

janezd:
##### Issue

Tests started failing when Qt 5.14 was released a few hours ago. I can't replicate it locally, but this may fix it. (And it probably won't. :)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
The test still fails (apparently sorting was correct, but the indicator was not). The big problem, though, is the core dump that follows.

For the time being we can pin Qt to 5.13 until this issue is resolved, so we can continue to use tests.

PR #4283, opened 2019-12-19
[ENH] OWTree: Add 'Classification Tree' keyword

irgolic:
##### Description of changes
Screenshot 2019-12-19 at 21 54 45

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
It is not sufficient to just add a keyword to the widget. You need to generate a new widget.json, so that the documentation reflects the changes. Hence tests fail on docs.

PR #4292, opened 2019-12-22
[ENH] Box Plot: Add box for missing group values

PrimozGodec:
##### Issue

@lanzagar and I discussed that there should be also a box for missing values in subgropus for countinuous variables as it is done for discrete. 

##### Description of changes
When the selected variable is continuous there is now box with missing values if they are present in the subgroup. It shows the distribution of values that are missing in the subgroups variable.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -494,20 +494,21 @@ class OWBoxPlot(widget.OWWidget):
             self.dist = []
             self.conts = contingency.get_contingency(
                 dataset, attr, self.group_var)
+            group_var_labels = self.group_var.values + [
+                f"missing '{self.group_var.name}'"]
             if self.is_continuous:
                 stats, label_texts = [], []
-                for i, cont in enumerate(self.conts):
+                for i, cont in enumerate(self.conts.array_with_unknowns):
                     if np.sum(cont[1]):
                         stats.append(BoxData(cont, attr, i, self.group_var))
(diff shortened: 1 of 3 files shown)


Orange/widgets/visualize/owboxplot.py
```diff
@@ -494,12 +494,13 @@ def compute_box_data(self):
             self.dist = []
             self.conts = contingency.get_contingency(
                 dataset, attr, self.group_var)
+            group_var_labels = self.group_var.values + ["Missing value"]
```
ajdapretnar: I was thinking whether we could simplify the user experience by instead printing, say "missing thal" or "missing sepal length". Same for splitting by missing values. This would tell the user more clearly what those missing values are.

PrimozGodec: I think it is a great idea. I added the names of the variables to the missing label.

janezd:
I have one trivial suggestion, which you may feel free to reject or ignore if not other changes are made.

Screenshot 2020-01-06 at 17 05 21

I think writing `missing thal` in lower case would like nicer here, and even when other values are written in capitals, `missing thal` in lower case would not stick out.

I would also put the attribute name into (single) quotes.

PrimozGodec:
@janezd I agree with you it looks nicer and also using a single quote more clearly shows that it is not the name of the value.

PR #4295, opened 2019-12-23
[ENH] OWTranspose: data info displayed in status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/data/owtranspose.py
```diff
@@ -92,6 +95,10 @@ def set_data(self, data):
         if self.feature_model:
             self.openContext(data)
         self.unconditional_apply()
+        if data:
```
VesnaT: I'd rather this happens before `self.apply()`. The apply could take a while with a big dataset. In that case the info is wrongly set through the whole procedure.

The best place is probably right after `self.data = data` line.

PR #4296, opened 2019-12-23
[WIP] OWSelectRows: data info displayed in status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -7,7 +7,7 @@ import numpy as np
 from AnyQt.QtWidgets import (
     QWidget, QTableWidget, QHeaderView, QComboBox, QLineEdit, QToolButton,
     QMessageBox, QMenu, QListView, QGridLayout, QPushButton, QSizePolicy,
-    QLabel)
+    QLabel, QHBoxLayout)
 from AnyQt.QtGui import (
     QDoubleValidator, QRegExpValidator, QStandardItemModel, QStandardItem,
     QFontMetrics, QPalette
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owselectrows.py
```diff
@@ -628,20 +626,33 @@ def commit(self):
...
         else:
-            lab1.setText(label + "~%s row%s, %s variable%s" %
-                         (sp(data.approx_len()) +
-                          sp(len(data.domain.variables) +
-                             len(data.domain.metas)))
-                        )
+            self.info.set_input_summary(self.info.NoInput)
+
+    def update_output_summary(self, data):
+        def sp(s, capitalize=True):
```
VesnaT: This function is copied from `update_input_summary`. Consider creating a staticmethod to avoid the duplicated code.

EDIT: Actually, the majority of the code in this method is duplicated. You could replace it with something like:

```     
    def _create_summary(self, data):
        def sp(s, capitalize=True):
            return s and s or ("No" if capitalize else "no"), "s" * (s != 1)

        summary, details = self.info.NoInput, ""
        if data:
            n = data.approx_len()
            details = "~%s row%s, %s variable%s" % \
                      (sp(n) + sp(len(data.domain.variables) +
                                  len(data.domain.metas)))
            summary = self.info.format_number(n)
        return summary, details
``` 

and then call it from `update_input_summary` and `update_output_summary`:
`self.info.set_input_summary(*self._create_summary(data))`

Orange/widgets/data/owselectrows.py
```diff
@@ -628,20 +626,33 @@ def commit(self):
...
+    def update_input_summary(self, data):
         def sp(s, capitalize=True):
             return s and s or ("No" if capitalize else "no"), "s" * (s != 1)
 
-        if data is None:
-            lab1.setText("")
+        if data:
+            n = data.approx_len()
+            details = "~%s row%s, %s variable%s" % \
+                      (sp(n) + sp(len(data.domain.variables) + len(data.domain.metas)))
```
VesnaT: Try to keep the line length up to 79 characters.

Orange/widgets/data/owselectrows.py
```diff
@@ -492,7 +490,7 @@ def set_data(self, data):
         else:
             self.add_row()
 
-        self.update_info(data, self.data_in_variables, "In: ")
+        self.update_input_summary(data)
```
VesnaT: If you place this line at the beginning of the method, you can remove upper `self.update_input_summary(None)`.

Orange/widgets/data/owselectrows.py
janezd: Also, consider replacing formatting with `%` with formatting with f-strings.

Orange/widgets/data/owselectrows.py
```diff
@@ -628,20 +626,28 @@ def commit(self):
...
 
-        if data is None:
-            lab1.setText("")
-        else:
-            lab1.setText(label + "~%s row%s, %s variable%s" %
-                         (sp(data.approx_len()) +
-                          sp(len(data.domain.variables) +
-                             len(data.domain.metas)))
-                        )
+        summary, details = self.info.NoInput, ""
```
VesnaT: `summay` needs to be passed as a parameter. 
For output summary `self.info.NoOutput` should be used.

PR #4301, opened 2019-12-27
[ENH] Distributions: Add option to hide bars

### LINKED ISSUE #4119: Distributions: impossible to hide bars and get just smooth kernel density lines
With the updated distributions widget, it seems impossible to get the view that is shown in the documentation:
image

The closest I can get is this but I can't get rid of the bars:
image

Any suggestion?

I'm using v3.23.1

janezd:
##### Issue

Resolves #4119.

##### Description of changes

- Add a checkbox for hiding bars; add an attribute 'hidden' to bars and don't paint when it's True
- Move the check box for cummulative distributions to the box below; for aesthetics and because it belongs there more than above (though both would be OK)
- Fill curves when bars are hidden
- Move legend to the plot above, so that it appears above curves

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owdistributions.py
+++ b/Orange/widgets/visualize/owdistributions.py
@@ -65,7 +65,7 @@ class LegendItem(SPGLegendItem):
 
 class DistributionBarItem(pg.GraphicsObject):
     def __init__(self, x, width, padding, freqs, colors, stacked, expanded,
-                 tooltip):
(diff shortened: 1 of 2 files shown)


ajdapretnar:
Really nice.

prdubois:
Thanks so much for this!

PR #4305, opened 2019-12-30
[ENH] Reorganize continuous palettes

### LINKED ISSUE #1924: Gradient Color Picker
##### Orange version

3.3.5

##### Feature Request:

It would be convenient to have a [custom gradient picker](https://www.google.com/search?q=custom+gradient+picker&source=lnms&tbm=isch&sa=X&ved=0ahUKEwjKoNnWrMDRAhUQ24MKHcYDCAUQ_AUICSgC&biw=1122&bih=776) for better customization of output visualizations.

This applies to any widgets that use color gradients in their visualizations. (distance map is used in the example image)

screen shot 2017-01-13 at 5 59 39 pm

Hope this is possible! :) I imagine there might be some open source color pickers out there, as long as the underlying libraries support manual color input. I would go ahead and look for some, but I'm not the most code-literate, hopefully someone else knows an optimal way to implement this.

janezd:
##### Issue

Fixes #1924

##### Description of changes

- Takes a subset of palettes from colorcet
- Implements classes for DiscretePalette, ContinuousPalette and BinnedContinuousPalette
- Implements a model and a combo box for choosing palettes
- Ports core widgets to new palettes and marks colorbrewer and colorpalette as obsolete
- Replaces gradient picker in the Color widget with predefined palettes
- Eliminates gamma from heatmap because; let the user set the range instead

**To fix**:

- [x] SOM: Color by continuous and select 'Show pie charts'. Error is thrown.
- [x] Tree Viewer (with housing): Color by mean value. Error is thrown.
- [x] Pythagorean Tree with node color Mean: legend is wrong.
- [x] Eliminate centering checkbox in the heatmap; center when the palette is diverging

**To consider for the future (rather than here)**: 
- **Legends.** Provide methods for drawing legends. Some can be ported from existing widgets. Consider using module owlegend. We should decide whether all widgets should eventually use that module for drawing legends.
- **Utility functions** Move `Scale` and some other methods from `owscatterplotgraph` and possibly `owsom` to some utils module so they can be shared with other widgets.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


ajdapretnar:
This looks much better than before. +1 for the PR!

ajdapretnar:
2 bugs.

1. File - Tree - Tree Viewer (select Target class: iris setosa)
```
  File "/Users/ajda/orange/orange3/Orange/widgets/visualize/owtreeviewer.py", line 235, in color_changed
    self.toggle_node_color_cls()
  File "/Users/ajda/orange/orange3/Orange/widgets/visualize/owtreeviewer.py", line 387, in toggle_node_color_cls
    color = colors[self.target_class_index - 1].lighter(
AttributeError: 'numpy.ndarray' object has no attribute 'lighter'
```

2. File - SOM.
```
---------------------------- ValueError Exception -----------------------------
  File "/Users/ajda/orange/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 936, in __process_next
  File "/Users/ajda/orange/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 974, in __process_next_helper
  File "/Users/ajda/orange/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 605, in process_node
  File "/Users/ajda/orange/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 792, in send_to_node
  File "/Users/ajda/orange/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 826, in process_signals_for_widget
    handler(*args)
  File "/Users/ajda/orange/orange3/Orange/widgets/unsupervised/owsom.py", line 354, in set_data
    self.create_legend()
  File "/Users/ajda/orange/orange3/Orange/widgets/unsupervised/owsom.py", line 871, in create_legend
    Qt.gray, color))
  File "/Users/ajda/orange/orange3/Orange/widgets/visualize/utils/__init__.py", line 679, in __init__
    if brush_color:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

ajdapretnar:
Issues:
- Color does not propagate changes at all times (use Copy to all option).
- Distance Map does not show selection in an obvious way. Selections are now way too similar to the palette, so they are practically invisible (especially when you use a reddish color palette).
- SOM: Color by continuous and select 'Show pie charts'. Error is thrown.
- Tree Viewer (with housing): Color by mean value. Error is thrown.
- Pythagorean Tree with node color Mean: legend is wrong.

Don't honestly know what to check in Paint Data, Calibration Plot, Lift Curve and ROC Analysis. Or Venn Diagram.

janezd:
> Color does not propagate changes at all times (use Copy to all option).

I don't see how this could happen. Did you discover this in scatter plot (or some other projection)? I was able to replicate this on master. The problem is in scatter plot -- if you disconnect and reconnect it, the colors are OK. I reported this in #4391.

> Distance Map does not show selection in an obvious way. Selections are now way too similar to the palette, so they are practically invisible (especially when you use a reddish color palette).

Suggestions? Should Distance map (and similar widgets) be aware of palettes and pick different colors for showing selections? Palette classes could include information about which color is "distinguishable" from those in the palette. Should we do this?

> Don't honestly know what to check in Paint Data, Calibration Plot, Lift Curve and ROC Analysis. Or Venn Diagram.

Just that they work. They use new palettes now, but just for getting a list of colors. Not much to go  wrong there.

janezd:
I fixed problems found by @ajdapretnar, except for choosing a better color for marking selections, which we have to discuss.

@ajdapretnar, I updated documentation, except images. I'm attaching a colorful screenshot of a widget. I haven't prepared screenshot with workflow and widgets (because shadows and background...). @ajdapretnar, could you? For numeric variables, consider "rainbow". If you like it -- if you don't, don't.

Screenshot 2020-02-01 at 18 15 26

ajdapretnar:
As for selections, we can address this in another PR I suppose.
I will prepare the documentation. With rainbow, of course, which is one of my favorites.

ajdapretnar:
I have pushed two commits to your branch. They add the documentation and a sorry attempt at tricking the build script to ignore explicitly unindexed images. I do not know anything about bash, so... 🤷‍♀ It should probably work, but who knows.

Orange/widgets/utils/colorbrewer.py
```diff
@@ -1,3 +1,9 @@
+import warnings
+
+warnings.warn("Module 'colorbrewer' is obsolete and will be removed.\n"
+              "Use palettes from 'Orange.misccolorpalettes'.",
```
lanzagar: `Orange.widget.utils.colorpalettes` ?

Orange/widgets/visualize/utils/owlegend.py
```diff
@@ -643,28 +661,14 @@ def set_domain(self, domain):
...
-        vals, colors = list(zip(*values))
-
-        # If the orientation is vertical, it makes more sense for the smaller
-        # value to be shown on the bottom
-        if self.orientation == Qt.Vertical and vals[0] < vals[len(vals) - 1]:
-            colors, vals = list(reversed(colors)), list(reversed(vals))
-
+        vals, palette = values
+        if self.orientation == Qt.Vertical:
+            vals = reversed(vals)
```
lanzagar: I am always nervous with these one-time-use generator iterables.
While it does seem to be ok now, vals are passed on to `ContinuousLegendItem` and there to `_format_values` and no one guarantees that they will use vals only once (even though they do now).
How about keeping the `list( )` around `reversed`?

Orange/widgets/utils/colorpalette.py
```diff
@@ -391,6 +392,12 @@ def colorSchemaChange(self):
 class ColorPalleteListing(QDialog):
     def __init__(self, parent=None, windowTitle="Color Palette List",
                  **kwargs):
+        # Import locally to prevent warning about using an obsolete module
+        # (This module itself is obsolete, but imported by its replacement,
+        # for backward compatibility, so importing colorbrewer globally would
+        # cause a warning.)
+        from Orange.widgets.utils import colorbrewer
```
lanzagar: `colorbrewer` is still imported globally on line 23 and that seems fine as the replacement module (`colorpalettes`?) does not import this one.
I guess all of the above can be removed?

Orange/widgets/utils/colorpalette.py
```diff
@@ -1009,5 +1016,8 @@ def main():  # pragma: no cover
     a.exec()
 
 
+warnings.warn(
+    "Module colorpalette is obsolete; use colorpalettes", DeprecationWarning)
```
lanzagar: Is there a reason this is not at the top of the module like in `colorbrewer`?

Orange/widgets/utils/colorbrewer.py
janezd: Yes, this module travelled around a bit.

PR #4317, opened 2020-01-09
io: Refactor FileFormat

VesnaT:
##### Issue

- `FileFormat.data_table()` consist of almost 200 lines
- the `io` module has to many lines according to pylint

##### Description of changes
- split io.py into three files (io.py, io_base.py, io_utils.py)
- io_base contains FileFormat base classes
- io_util contains utility functions

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
Separating some functionality only relevant to .tab or excel reader into a mixin is a step forward, but may break some external code that relied on the FileFormat. Perhaps we could even ignore this.

The refactored `_FileWriter` contains functions that are probably only relevant for .tab files. These should go directly into the corresponding file format or into a mixin if there are more readers that rely on them.

PR #4319, opened 2020-01-10
Select Rows: Increase settings version

### LINKED ISSUE #4316: Select rows sends and shanges selection when it should not
Similar to https://github.com/biolab/orange3/issues/4278 that was already closed.
When I change group in widget marked by A the selection changes in B. This should not happen. What is even stranger, even if I disable automatic sending in B the data after B changes according to the change in B upon me changing A.
Screenshot from 2020-01-09 10-22-56

Orange 3.25.0 (last master, if I am not mistaken).

ajdapretnar:
##### Issue

Fixes #4316.
At least for me. 😅 

##### Description of changes
Increase settings version.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -126,7 +126,7 @@ class OWSelectRows(widget.OWWidget):
     purge_classes = Setting(False, schema_only=True)
     auto_commit = Setting(True)
 
-    settings_version = 2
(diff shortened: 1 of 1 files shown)


janezd:
Good hunch that helped identify the actual cause. Unfortunately, there is nothing to fix. See #4316.

PR #4321, opened 2020-01-10
Set stateWhenDisabled where appropriate

### LINKED ISSUE #4318: Show proper state for disabled check boxes
Many widgets disable check boxes when they are not applicable. For instance, if "Split by" in Distribution is set to None, check boxes Stack columns and Show probabilities are disabled. However, if they are checked they remain checked (so that they are checked when they are re-enabled). The widget itself ignored the state, but the information that the user sees is wrong.

https://github.com/biolab/orange-widget-base/pull/30 in orange-widget-base (merged, but not yet released) resolves this by introducing an argument `stateWhenDisabled`. In case of Distributions, settings `stateWhenDisabled=False` would ensure that check boxes would be shown as unchecked when they are disabled, although the underlying setting is still `True`. When reenabled, the box will be shown as checked.

Somebody would need to go through all widgets, or perhaps just search the code for `setDisabled` and `setEnabled`, see whether it disables check boxes and if so, provide the `stateWhenDisable` argument with the proper value.

This PR won't be mergeable until orange-widget-base is released.

ajdapretnar:
##### Issue

Resolves #4318.

##### Description of changes
Add stateWhenDisabled where appropriate.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/plot/owplotgui.py
+++ b/Orange/widgets/utils/plot/owplotgui.py
@@ -575,14 +575,15 @@ class OWPlotGUI:
         else:
             return name
 
-    def _check_box(self, widget, value, label, cb_name):
+    def _check_box(self, widget, value, label, cb_name, stateWhenDisabled=None):
         '''
             Adds a :obj:`.QCheckBox` to ``widget``.
             When the checkbox is toggled, the attribute ``value`` of the plot object is set to
             the checkbox' check state, and the callback ``cb_name`` is called.
         '''
         args = dict(master=self._plot, value=value, label=label,
-                    callback=self._get_callback(cb_name, self._plot))
+                    callback=self._get_callback(cb_name, self._plot),
+                    stateWhenDisabled=stateWhenDisabled)
         if isinstance(widget.layout(), QGridLayout):
             widget = widget.layout()
         if isinstance(widget, QGridLayout):
@@ -627,7 +628,8 @@ class OWPlotGUI:
         self._master.cb_class_density = \
             self._check_box(widget=widget, value="class_density",
                             label="Show color regions",
-                            cb_name=self._plot.update_density)
+                            cb_name=self._plot.update_density,
+                            stateWhenDisabled=False)
 
     def regression_line_check_box(self, widget):
         self._master.cb_reg_line = \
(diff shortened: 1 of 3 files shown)


Orange/widgets/data/owselectrows.py
```diff
@@ -126,7 +126,7 @@ class Outputs:
     purge_classes = Setting(False, schema_only=True)
     auto_commit = Setting(True)
 
-    settings_version = 2
+    settings_version = 3
```
janezd: Rebase to master. :)

ajdapretnar:
@janezd I am not entirely satisfied with the solution for owplotgui. Could you check and propose what could be done better?

Orange/widgets/data/owselectrows.py
ajdapretnar: Who made such a stupid PR??? 😅

Orange/widgets/utils/plot/owplotgui.py
```diff
@@ -575,14 +575,15 @@ def _get_callback(self, name, master=None):
         else:
             return name
 
-    def _check_box(self, widget, value, label, cb_name):
+    def _check_box(self, widget, value, label, cb_name, state):
```
janezd: > I am not entirely satisfied with the solution for owplotgui. Could you check and propose what could be done better?

owplotgui should be abolished anyway, so I don't care much. (Idea was to unify GUIs of visualizations. Projections are now unified under a common base widget, and on the first glance, owplotgui is used exclusively in that widget (and derived ones). And even they use just small pieces, while most of it is unused. Check box for animating visualizations!?)

Still, please:

- rename the argument to `stateWhenDisabled`, otherwise this looks like initial state. Plus, it will be clearer that it is passed to `gui.checkBox`.
- give it a default value (`None`).

ajdapretnar: Fixed. I hope appropriately.

ajdapretnar:
Tests don't fail locally. They also don't seem to be related to this PR, unless I am missing something...

PrimozGodec:
Tests started to fail since with adding `stateWhenDisabled=False` a callback function is called when the check box is disabled/enabled. It causes problems just when new data comes and disabling/enabling the checkbox changes the value of `stretched` since disabling was made before computing new data. I fixed it by reordering the lines in two functions.

I also rebased this PR to master (conflicts with a new heatmap). I think this setting is not required in the new heatmap.

janezd:
Amazingly few widgets disable check boxes. Interesting.

PR #4334, opened 2020-01-15
[RFC] Nomogram: Prevent connecting incompatible models

janezd:
##### Issue

Users can connect arbitrary classifiers to nomogram; if the classifier is not naive Bayes or logistic regression, the widget informs the user that this type of classifier is invalid. It would be nicer if classifiers of wrong types couldn't be connected at all.

**Caveat:** are there (or could there be) widgets that output models and this model can be naive Bayes or logistic regression -- but not necessarily? E.g. what if ROC would output the optimal model for some operating condition? In this case, the current code supports connecting and checking later, while this PR prohibits it.

##### Description of changes

Implement a mixin that serves as a flag and mark the acceptable models.

### DIFF
--- a/Orange/classification/base_classification.py
+++ b/Orange/classification/base_classification.py
@@ -17,6 +17,15 @@ class ModelClassification(Model):
     pass
 
 
+class LinearModel:
+    """
+    A mix-in that serves as a flag for linear models, like naive Bayesian
+    classifier and logistic regression (but not SVM which can also be
+    non-linear -- in terms of original variables). The nomogram widget
+    accepts only models with this mix-in.
+    """
+
+
 class SklModelClassification(SklModel, ModelClassification):
     def predict(self, X):
         prediction = super().predict(X)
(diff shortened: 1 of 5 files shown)


ales-erjavec:
You can achieve this with:
```diff
diff --git a/Orange/widgets/visualize/ownomogram.py b/Orange/widgets/visualize/ownomogram.py
index 0ad19e68e..5b1aa1acf 100644
--- a/Orange/widgets/visualize/ownomogram.py
+++ b/Orange/widgets/visualize/ownomogram.py
@@ -591,7 +591,7 @@ class OWNomogram(OWWidget):
     keywords = []
 
     class Inputs:
-        classifier = Input("Classifier", Model)
+        classifier = Input("Classifier", (NaiveBayesModel, LogisticRegressionClassifier))
         data = Input("Data", Table)
 
     class Outputs:
```

> ... are there (or could there be) widgets that output models and this model can be naive Bayes or logistic regression -- but not necessarily?

Load Model is one such widget.

janezd:
> Load Model is one such widget.

Thanks, @ales-erjavec. This was a bad idea.

PR #4335, opened 2020-01-15
Sorting fixes in predictions widget

### LINKED ISSUE #4312: Predictions require clicking twice, can't sort by probabilities
#4307, which I just merged, fixes the problem, but -- as I noticed a moment later -- not entirely. Instead of reverting, I'm writing a PR and described another, related problem that already existed before.

### Twice-click

Say I sort by predictions. I click on some data column and the order is incorrect. But then I click again and the order is correct. I keep clicking on data columns and the order is always correct in a first try (which must be why I haven't noticed the problem!).

But then I click on predictions column. I don't get the correct order of predictions. Click again and the order is correct. If you now try sorting by a data column, you have to click twice, again...

### Sorting by probabilities

`PredictionSortProxyModel.setProbInd` is never called.

If I display probabilities for a single class, I would expect to see column sorted by that probability. If I select multiple, I'd like sorting by the first probability, and ties resolved by the next.

PrimozGodec:
##### Issue

- Fixes #4312
- Order of probabilities does not match the list on the left, colors do not match too.
- SVM works wrong when the class attribute has a value that does not appear in Y. In this case, we expect that probabilities shape match to class values, but they don't, since SVM has its own predict function that works differently than other models. 

##### Description of changes
- Fixing sorting problems from #4312.
- Fixing SVM which now uses the same predict function than other models. 
- Fixing problems with colors (when different orders of probabilities)
- Fix probabilities order when different values order in learners

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -413,7 +413,11 @@ class SklModel(Model, metaclass=WrapperMeta):
 
     def predict(self, X):
         value = self.skl_model.predict(X)
-        if hasattr(self.skl_model, "predict_proba"):
+        # SVM has probability attribute which defines if method compute probs
+        has_prob_attr = hasattr(self.skl_model, "probability")
+        if (has_prob_attr and self.skl_model.probability
+                or not has_prob_attr
+                and hasattr(self.skl_model, "predict_proba")):
             probs = self.skl_model.predict_proba(X)
             return value, probs
         return value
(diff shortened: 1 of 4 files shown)


PrimozGodec:
This should be finished now. Can someone test it very well this PR makes changes of some very essential parts of widget.

PR #4340, opened 2020-01-16
[FIX] Round bhattacharayya

AndrejaKovacic:
##### Issue

Due to floating point precision,  bhattacharayya distance would sometimes return negative distances.

##### Description of changes
Rounding numbers to 13th digit. I think this is precise enough for our purposes, otherwise, we could use decimal class instead of floats.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/distance/distance.py
+++ b/Orange/distance/distance.py
@@ -749,9 +749,9 @@ def _bhattacharyya(a, b):
     check_non_negative(b)
     a = _prob_dist(a)
     b = _prob_dist(b)
-    if sp.issparse(a):
-        return -np.log(np.sum(np.sqrt(a.multiply(b))))
-    return -np.log(np.sum(np.sqrt(a * b)))
+    prod = a.multiply(b) if sp.issparse(a) else a * b
+    return np.clip(-np.log(np.sum(np.sqrt(prod))), 0, None)
+
 
 
 class Bhattacharyya(Distance):


Orange/distance/distance.py
```diff
@@ -750,8 +750,8 @@ def _bhattacharyya(a, b):
     a = _prob_dist(a)
     b = _prob_dist(b)
     if sp.issparse(a):
-        return -np.log(np.sum(np.sqrt(a.multiply(b))))
-    return -np.log(np.sum(np.sqrt(a * b)))
+        return np.clip(-np.log(np.sum(np.sqrt(a.multiply(b)))), 0, None)
+    return np.clip(-np.log(np.sum(np.sqrt(a * b))), 0, None)
```
janezd: I know I'm being an a..., I mean, annoying, but I would find it much nicer if you changed the code to something like

```
if sp.issparse(a):
    prod = a.multiply(b)
else:
    prod = a * b
return np.clip(-np.log(np.sum(np.sqrt(prod))
```

Or even `prod = a.multiply(a) if sp.issparse(a) else a * b`.

AndrejaKovacic: it does look nicer, i changed it.

markotoplak: Hmmm, is the condition even needed? Wouldn't `a.multiply(b)` work to with sparse and dense matrices?

janezd: You seem to think I haven't tried. :)

```
>>> import numpy as np
>>> a = np.arange(20).reshape(4, 5)
>>> a.multiply(a)
AttributeError: 'numpy.ndarray' object has no attribute 'multiply'
```

There is `np.multiply`, but it doesn't work for sparse matrices.

AndrejaKovacic: No, numpy arrays don't have multiply function, only the numpy module has it.

Issue #2663, opened 2017-10-10
OWHierarchialClustering: HC computes sth even once the data is removed


ajdapretnar:
##### Orange version

3.7.dev

##### Expected behavior
Hierarchical Clustering empties instantly once the data input gets disconnected/empty.

##### Actual behavior
Hierarchical Clustering stalls and appears to be computing something even when the input data gets empty (removed).

##### Steps to reproduce the behavior
File (really large data set) - Distances - Hierarchical Clustering. Remove connection before Distances and observe HC.

jerneju:
Widget spends a lot of time clearing all the items.

ales-erjavec:
This is still an issue and can be fixed/improved.

PR #4361, opened 2020-01-23
[FIX] DendrogramWidget: Remove event filters before removing items

ales-erjavec:
##### Issue

Fix what appears to be a quadratic time bottleneck in `DendrogramWidget.clear`

Alleviates gh-2663

##### Description of changes

* Remove event filters before removing items
Each `removeItem` seems to be linear in number of installed filters in the whole scene -> so this would be quadratic for us (not accounting any other filters not part of `DendrogramWidget`).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -311,6 +311,12 @@ class DendrogramWidget(QGraphicsWidget):
             self.setParentItem(parent)
 
     def clear(self):
+        # Remove event filter from all items before removal.
+        # Each removeItem is linear in number of installed filters in the whole
+        # scene -> so this would be quadratic for us (not accounting any other
+        # filters not part of DendrogramWidget).
+        for item in self._items.values():
+            item.removeSceneEventFilter(self)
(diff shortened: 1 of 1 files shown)


ales-erjavec:
Benchmarked using:
```python
import timeit

import numpy as np
from scipy.cluster.hierarchy import linkage

from PyQt5.QtWidgets import QApplication, QGraphicsScene, QGraphicsView
from Orange.widgets.unsupervised.owhierarchicalclustering import DendrogramWidget
from Orange.clustering import hierarchical as hier

app = QApplication([])
scene = QGraphicsScene()
view = QGraphicsView()
view.setScene(scene)

N = 15000

dw = DendrogramWidget()
scene.addItem(dw)

mat = np.random.exponential(0, size=N * (N - 1) // 2,)
Z = linkage(mat)
tree = hier.tree_from_linkage(Z)


def setup():
    dw.set_root(tree)


def run():
    dw.clear()


res = timeit.timeit(
    "run()", setup="setup()", globals=globals(), number=100
)
print(res)
```
Before: 17.68
After: 0.71

PR #4363, opened 2020-01-24
[ENH] Allow concurrent transformation of tables into new domains

markotoplak:
##### Issue
Table.from_table was locked to one thread only. This started being problematic now that more widgets are running in threads. If a long `Table.transform` is running in one thread,  another threads will block if they try to use `Table.transform`.

This happened, for example, while testing quasars/orange-spectroscopy#395 (see comment by @stuart-cls).

##### Description of changes
Table.from_table had a reentrant lock on its conversion cache. That lock was needed so that we knew when to purge the cache. Making the cache thread-local allows concurrent calls.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1,12 +1,13 @@
 import operator
 import os
+import threading
 import warnings
 import zlib
 from collections import Iterable, Sequence, Sized
 from functools import reduce
 from itertools import chain
 from numbers import Real, Integral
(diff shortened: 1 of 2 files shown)


markotoplak:
Codecov complains about some apparently stale code which did not change in this PR. Such changes are dangerous so I would not like to change it here. Please ignore its complaints.

Issue #4349, opened 2020-01-21
Heat Map: slow on delete


ajdapretnar:
**Describe the bug**
With 32,000 rows, Heat Map is slow to delete.

**To Reproduce**
Load any large file and plot it in a Heat Map.
Then delete the widget.

**Orange version:**
3.25.dev

**Expected behavior**
Heat Map is deleted relatively quickly.

**Operating system:**
OSX

PR #4365, opened 2020-01-24
[FIX] Fix slow clear/delete in 'Heat Map' 'Hier. Clustering', 'Distance Map'

ales-erjavec:
##### Issue

owheatmap, owdistancemap and owhierarchicalclustering use QGraphicsLinearLayout to position text labels in a list. Clearing the QGraphicsLinearLayout however is slow (seems to be another quadratic time complexity).

Extract a more limited (but sufficient) TextListWidget from owsilhouetteplot and replace the use of QGraphicsLinearLayout.

Fixes gh-2663
Fixes gh-4349

##### Description of changes

* extract and extend the TextListWidget from owsilhouetteplot
* use the TextListWidget in owheatmap, owhierarchiclustering, owdistancemap

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owdistancemap.py
+++ b/Orange/widgets/unsupervised/owdistancemap.py
@@ -5,11 +5,12 @@ from operator import iadd
 import numpy
 
 from AnyQt.QtWidgets import (
-    QFormLayout, QGraphicsRectItem, QGraphicsGridLayout, QApplication
+    QFormLayout, QGraphicsRectItem, QGraphicsGridLayout, QApplication,
+    QSizePolicy
 )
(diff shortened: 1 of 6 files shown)


thocevar:
Heat map visualizations with a lot of rows (e.g., Adult dataset) are now besides being faster to delete also faster to construct. 

Distance map contains a quadratic number of elements anyway, therefore the difference is not that obvious.

Hierarchical clustering must be doing something strange besides this. It should containt a linear number of elements but is still slow to delete. I don't think it solves #2663.

PR #4379, opened 2020-01-30
[FIX] Continuize: Disable normalizing sparse data

### LINKED ISSUE #4378: Continuize: Normalizing sparse data fails
**Describe the bug**
Continuize widget crashes when normalizing sparse data.

**To Reproduce**
Corpus (book-excerpts.tab) -> Bag Of Words -> Continuize -> In `Numeric Features` box select `Normalize by span`

**Orange version:**
master

**Expected behavior**
Disabled normalization for sparse datasets.

VesnaT:
##### Issue

Fixes #4378

##### Description of changes
Disable `Normalize by span` and `Normalize by standard deviation` radio buttons for sparse datasets.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcontinuize.py
+++ b/Orange/widgets/data/owcontinuize.py
@@ -34,9 +34,12 @@ class OWContinuize(widget.OWWidget):
     buttons_area_orientation = Qt.Vertical
     resizing_enabled = False
 
+    # continuous treats
+    Leave, NormalizeBySpan, NormalizeBySD = range(3)
+
     multinomial_treatment = Setting(0)
     zero_based = Setting(1)
-    continuous_treatment = Setting(0)
+    continuous_treatment = Setting(Leave)
     class_treatment = Setting(0)
 
     transform_class = Setting(False)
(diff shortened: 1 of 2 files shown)


markotoplak:
What kind of normalization can we still have with sparse data then? We could still do normalization that does some division or multiplication, we just have to avoid shifts (plus, minus).

Another thing: I never associated sparse data and discrete values, but yes, why not... Where do we get discrete sparse data in Orange? Is ti directly read from a file or generated by some text-mining widget?

ajdapretnar:
🤔 I think that if you have some discrete variables in the corpus before bag of words, they would inevitably get transformed into sparse alongside words.

VesnaT:
The example is described in the issue, and yes, it does not make sense but it still should not crash.

markotoplak:
Yes, but using appropriate operations for sparse data would be better than disabling options. Normalization by span is something that could still be done for sparse data, it should just not be centered.

VesnaT:
Some types of normalization can be done using a Preprocess widget. Should these be added to the Continuize widget as well?

PR #4389, opened 2020-01-31
OWTranspose: Warning when transposing categorical features

### LINKED ISSUE #4359: Transpose remembers old unexisting data
I have a table file  that I overwrite with different values (rows and columns stay the same). When using a certain version of the file with certain workflow the workflow remembers old data, that was previously overwritten (and does not exist on my disk anymore).  I could not replicate this in a new workflow made from scratch or with Orange data sets. However, it occurs even when I reopen Orange or when I copy the workflow to a new file. 
Below are the workflow and the data and an image of what is happening. 

Screenshot from 2020-01-23 11-41-55

[transpose_remembering.zip](https://github.com/biolab/orange3/files/4102632/transpose_remembering.zip)

Orange: Last master.

aturanjanin:
Fixes #4359.  A warning added if the input data contains categorical features.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtranspose.py
+++ b/Orange/widgets/data/owtranspose.py
@@ -38,6 +38,7 @@ class OWTranspose(OWWidget):
         duplicate_names = Msg("Values are not unique.\nTo avoid multiple "
                               "features with the same name, values \nof "
                               "'{}' have been augmented with indices.")
+        discrete_attrs = Msg("Categorical features have been encoded as numbers.")
 
     class Error(OWWidget.Error):
         value_error = Msg("{}")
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owtranspose.py
```diff
@@ -38,6 +38,9 @@ class Warning(OWWidget.Warning):
         duplicate_names = Msg("Values are not unique.\nTo avoid multiple "
                               "features with the same name, values \nof "
                               "'{}' have been augmented with indices.")
+        discrete_attrs = Msg("Values of categorical features will change.\n"
+                             "They will be encoded with suitable numeric "
+                             "representation.")
```
janezd: I would you past tense here. When the user reads the message, the deed has already been done. :)

aturanjanin: That is true, does the present passive voice sounds okay? 

Values of categorical features _are changed_. They _are encoded_ with suitable numeric representation.

janezd: It is not entirely clear what "suitable" means. Perhaps "They will be encoded as numbers."

Alternative: The widget is so narrow that the message won't fit anyway, so it doesn't hurt if it's too long: we can write just "Categorical features have been encoded as numbers." Without any `\n` and detils.

Issue #4397, opened 2020-02-04
Ability to export data table to csv/other formats with headers


Tarna8:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Tarna8:
Can copy/paste all manually but no headers

ajdapretnar:
Save Data can do all of this.

PR #4398, opened 2020-02-05
owsave: add "export" keyword

markotoplak:
##### Issue
#4397


ajdapretnar:
You need to generate also widget.json in docs (or less ideally add the keyword manually) so that the tests pass.

PR #4405, opened 2020-02-10
[FIX] DataProjectionWidget: Update combos on new data

### LINKED ISSUE #4391: Scatter plot does not detect coloring changes
**Describe the bug**

Scatter plot ignores data input when the affected attributes are unchanged. Other projections probably have a similar problem and it probably also affects size and shape.

**To Reproduce**

1. Connect File -> Color -> Scatter plot
2. In Scatter plot, set color to Cholesterol. X and Y should be, say, age and rest SBP.
3. In Color, change the color for max HR.
4. In Scatter plot, set color to max HR.

After this, max HR will be shown in the original color. I suspect that this occurs because at step 3 "max HR" was not in `effective_data`, so scatter plot ignored the new data input.

**Expected behavior**

I suppose projections should be more careful with `effective_data`. Perhaps this mechanism is flawed.

VesnaT:
##### Issue

Fixes #4391

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/tests/test_owprojectionwidget.py
+++ b/Orange/widgets/visualize/tests/test_owprojectionwidget.py
@@ -11,6 +11,7 @@ from Orange.data import (
 from Orange.widgets.tests.base import (
     WidgetTest, WidgetOutputsTestMixin, ProjectionWidgetTestMixin
 )
+from Orange.widgets.utils.colorpalettes import ContinuousPalettes
 from Orange.widgets.visualize.utils.widget import (
     OWDataProjectionWidget, OWProjectionWidgetBase
 )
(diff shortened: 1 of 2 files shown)


janezd:
At first I didn't understand why this fixes the problem. It is amazing that the widget worked with wrong variables in the model -- I guess it's only because we eliminated `Variable.make`, so the wrong variable matched the same-named variable from the new domain. :)

I added a test to convince myself that it's really about wrong variables in the model. And because it's prudent to have this test to prevent somebody from "re-optimizing" the code in the future.

VesnaT:
Thanks for the test. I tried to write one, but it kept passing with the old code.

PR #4412, opened 2020-02-12
[ENH] Outliers: Offload work onto separate thread

VesnaT:
##### Issue

Responsive Outliers widget: the widget could make calculations in a separate thread to make Orange responsive.

##### Description of changes
- Create a `progress_callback` function that serves as a wrapper function for callbacks. It allocates the corresponding proportion of the progress to the callback.
- Add callback to `Learner` to enable preprocessing interruption.
- Add callback to `_OutlierModel`.
- Do work in a separate thread.
- Still impossible to terminate the fit/predict operations.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -10,11 +10,12 @@ import scipy
 from Orange.data import Table, Storage, Instance, Value
 from Orange.data.filter import HasClass
 from Orange.data.table import DomainTransformationError
-from Orange.data.util import one_hot, progress_callback, dummy_callback
+from Orange.data.util import one_hot
 from Orange.misc.wrapper_meta import WrapperMeta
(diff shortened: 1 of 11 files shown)


markotoplak:
Thank you @VesnaT, I like it a lot!

Orange/base.py
```diff
@@ -101,7 +102,7 @@ def fit_storage(self, data):
         X, Y, W = data.X, data.Y, data.W if data.has_weights() else None
         return self.fit(X, Y, W)
 
-    def __call__(self, data):
+    def __call__(self, data, callback=None):
```
janezd: I glanced over this PR. The only thing I don't like is the name of the argument. `callback` can be anything. Perhaps `progress_callback`? Longer, but more informative.

Issue #4406, opened 2020-02-10
Concatenate categorical


GottM:
Hello @ajdapretnar ,
The Concatenate widget seems dysfunctional whenever categorical or text values are involved. Fields which should be recognized as same in both data tables are either handled as different or concatenated wrongly.
I enclose the simplest case I could find for you to reproduce the bug (see attached screenshots). I load two tables containing respectively A, B, C, D and E, F, G in one single column. The expected output of concatenation of the two would be a column containing A, B, C, D, E, F, G. Instead I get: A, B, C, D, A, B, C.
I hope that you guys will easily find the fix...
Regards,
G.
![Concatenate bug](https://user-images.githubusercontent.com/47322724/74139368-8b094c80-4bf3-11ea-8972-922c9d50831b.png)

janezd:
@ajdapretnar, this is wider than just concatenation. There are two bugs, and the first one is pretty scary.

1. The first is related to #3925, removal of `Variable.make`.

    Additional data is not properly converted into the new domain because `Domain.get_conversion` uses cached conversions, which it gets from a dictionary whose keys are domains. `Domain.__eq__` and `Domain.__hash__` use `Variable.__eq__` and `Variable.__hash__`, which ignore `Discrete.values` -- as they should. But it bites us here.

    I think that `Domain.__eq__` behaves properly, but `Domain.get_conversion` should require the cached domain is the same, not just equal. Or we can remove caching (@markotoplak?)

2. The second bug is in `OWConcatenate`. For categorical attributes with the same name, which are merged into one attribute, it should construct a new attribute with values from both (or all) original attributes.

    If I remove domain caching, it will construct a column with the attribute from the primary data, so all values from the second table that have no corresponding value in the primary table will be missing.

#### Test

Here is a test to be added to the PR that will fix this issue. (I hope it's correct. :)

```python

    def test_same_var_name(self):
        widget = self.widget

        var1 = DiscreteVariable(name="x", values=list("abcd"))
        data1 = Table.from_numpy(Domain([var1]),
                                 np.arange(4).reshape(4, 1), np.zeros((4, 0)))
        var2 = DiscreteVariable(name="x", values=list("def"))
        data2 = Table.from_numpy(Domain([var2]),
                                 np.arange(3).reshape(3, 1), np.zeros((3, 0)))

        self.send_signal(widget.Inputs.primary_data, data1)
        self.send_signal(widget.Inputs.additional_data, data2)
[...]

PR #4422, opened 2020-02-14
Remove domain transformation cache

janezd:
##### Issue

The first part of fix for #4406.

Two domains are considered equal if variables are equal. However, equality of discrete variables does not check for `values`.

##### Description of changes

Eliminate caching of domain conversions.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/data/domain.py
+++ b/Orange/data/domain.py
@@ -397,27 +397,6 @@ class Domain:
     def has_time_class(self):
         return bool(self.class_var and self.class_var.is_time)
 
-    def get_conversion(self, source):
-        """
-        Return an instance of :class:`DomainConversion` for conversion from the
-        given source domain to this domain. Domain conversions are cached to
-        speed-up the conversion in the common case in which the domain
-        is based on another domain, for instance, when the domain contains
-        discretized variables from another domain.
-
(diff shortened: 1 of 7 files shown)


markotoplak:
Before removal of domain conversion cache `python -m unittest -q benchmark.bench_domain` benchmarked as:

```
[transform_copy] with 5 loops, best of 3:
        min 16.1 msec per loop
        avg 16.2 msec per loop
[transform_discretize] with 5 loops, best of 3:
        min 153 msec per loop
        avg 155 msec per loop
[transform_normalize] with 5 loops, best of 3:
        min 137 msec per loop
        avg 138 msec per loop
```

After cache removal:

```
[transform_copy] with 5 loops, best of 3:
        min 32.2 msec per loop
        avg 32.3 msec per loop
[transform_discretize] with 5 loops, best of 3:
        min 163 msec per loop
        avg 164 msec per loop
[transform_normalize] with 5 loops, best of 3:
        min 138 msec per loop
        avg 139 msec per loop
```

Data had 2000 features and 100 instances. `transform_copy` is the ideal use case for this cache, because it only copies features. Transformations with non-trivial operations do not show significant slowdowns.

PR #4424, opened 2020-02-15
[ENH] Import openTSNE lazily for faster loading of Orange

### LINKED ISSUE #3523: Orange.projection loads slowly after introduction of fastTSNE

janezd:
Fixes #3523. 

#3883 crashed travis for unknown reasons. This fix is implemented a bit differently and replaces the lazy object with actual module once it is loaded.

It survived three runs of tests. If we're worried about random crashes, we can restart a few more times.

### DIFF
--- a/Orange/projection/manifold.py
+++ b/Orange/projection/manifold.py
@@ -8,10 +8,6 @@ from scipy.linalg import eigh as lapack_eigh
 from scipy.sparse.linalg import eigsh as arpack_eigh
 import sklearn.manifold as skl_manifold
 
-import openTSNE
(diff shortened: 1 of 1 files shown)


janezd:
I'll dare merging it. It won't be difficult to revert if needed.

Issue #4382, opened 2020-01-30
Domain: Prevent constructing domain with non-unique variable names


VesnaT:
**Is your feature request related to a problem? Please describe.**
We have an agreement to construct domains with unique variable names, but there are no checks for that in the code.

**Describe the solution you'd like**
Raise an error when constructing a domain with duplicated variable names.

**Describe alternatives you've considered**
Create unique variable names under the hood when constructing the domain.

janezd:
I support raising an exception because we should prefer failing early. The cost of this will be unhandled exceptions at various places, but this is better than unreported errors in various places (and miscomputations later).

I don't like the under-the-hood idea.

VesnaT:
Before we accept this enhancement, it is necessary to fix all widgets that may construct a new domain (which seems impossible).

AndrejaKovacic:
### Widgets, that are changing domains
#### Data widgets
  * File allows changing of names indiscriminately
  * Paint data
  * Pivot
    * change var name by which you are aggregating to Aggregate
  * Concatenate
    * name id column iris or whatever you please
  * Transpose
    * make class var for iris : Feature 001 Iris-setosa
  * Preprocess
    * for PCA: class var for iris : PC1, resulting column names are perhaps incosistent with the rest of widgets, no duplicates though
  * Python script
  * Feature constructor
   * any name goes

VesnaT:
Another proposal: Prevent constructing domains from tuples and lists. Create custom collection e.g. SortedSet. The collection in returned by `get_unique_names`. The purpose is to force the user to always use the `get_unique_names` when constructing the domain.

janezd:
Clever idea. But it would be backwards incompatible. Plus ... think about tests with hardcoded names - which are obviously unique. I'd hate having to call `get_unique_names` there. Plus if everybody just called `get_unique_names`, this is not much different from `Domain` calling it itself, I guess.

janezd:
I usually remember to edit "Partially fixes ..." to "Fixes (partially) ..." to prevent closing of an issue. In #4578 I forgot. :)

PR #4425, opened 2020-02-15
[FIX] Concatenate: Fix wrong merging of categorical features

### LINKED ISSUE #4406: Concatenate categorical

janezd:
##### Issue

Fixes #4406. Also fixes part of #4382.

The problem was that when merging domains, attributes from the first domain are used and other attributes values' that don't appear in the first table are turned into nan's.

This fix applies only when "primary table" is not present. When it is present, #4422 suffices.

##### Description of changes

When filtering for unique attributes, two (or more) attributes with different values are replaced by a new attribute.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -556,11 +556,16 @@ class ContinuousVariable(Variable):
     str_val = repr_val
 
     def copy(self, compute_value=None, *, name=None, **kwargs):
-        var = super().copy(compute_value=compute_value, name=name,
-                           number_of_decimals=self.number_of_decimals,
-                           **kwargs)
-        var.adjust_decimals = self.adjust_decimals
-        var.format_str = self._format_str
+        # pylint understand not that `var` is `DiscreteVariable`:
+        # pylint: disable=protected-access
+        number_of_decimals = kwargs.pop("number_of_decimals", None)
+        var = super().copy(compute_value=compute_value, name=name, **kwargs)
+        if number_of_decimals is not None:
+            var.number_of_decimals = number_of_decimals
+        else:
+            var._number_of_decimals = self._number_of_decimals
+            var.adjust_decimals = self.adjust_decimals
+            var.format_str = self._format_str
         return var
 
 
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -460,6 +460,10 @@ class Variable(Reprable, metaclass=VariableMeta):
         var.attributes = dict(self.attributes)
         return var
 
+    def renamed(self, new_name):
+        # prevent cyclic import, pylint: disable=import-outside-toplevel
+        from Orange.preprocess.transformation import Identity
+        return self.copy(name=new_name, compute_value=Identity(variable=self))
 
 del _predicatedescriptor
 
--- a/Orange/widgets/data/owconcatenate.py
+++ b/Orange/widgets/data/owconcatenate.py
@@ -6,14 +6,17 @@ Concatenate (append) two or more datasets.
 
 """
 
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 from functools import reduce
+from itertools import chain, count
+from typing import List
 
 import numpy as np
 from AnyQt.QtWidgets import QFormLayout
 from AnyQt.QtCore import Qt
 
 import Orange.data
+from Orange.data.util import get_unique_names_duplicates
 from Orange.util import flatten
 from Orange.widgets import widget, gui, settings
 from Orange.widgets.settings import Setting
@@ -42,6 +45,10 @@ class OWConcatenate(widget.OWWidget):
 
     class Error(widget.OWWidget.Error):
         bow_concatenation = Msg("Inputs must be of the same type.")
+
+    class Warning(widget.OWWidget.Warning):
+        renamed_variables = Msg(
+            "Variables with duplicated names have been renamed.")
 
     merge_type: int
     append_source_column: bool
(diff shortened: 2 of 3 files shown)


Orange/data/variable.py
```diff
@@ -585,16 +591,19 @@ class DiscreteVariable(Variable):
...
                  *, sparse=False):
         """ Construct a discrete variable descriptor with the given values. """
-        self.values = list(values)
-        if not all(isinstance(value, str) for value in self.values):
+        values = list(values)  # some people (including me) pass a generator
+        if not all(isinstance(value, str) for value in values):
             raise TypeError("values of DiscreteVariables must be strings")
+
         super().__init__(name, compute_value, sparse=sparse)
+        self.values = values
```
VesnaT: I'd rather make `self.values` a property and check `self._value_index - self._value` consistency in its setter. It's too easy to obtain inconsistent state with this api.

Orange/data/variable.py
```diff
@@ -700,13 +709,16 @@ def to_val(self, s):
...
+        if s not in self._value_index:
+            raise ValueError(f"Value {s} does not exist")
+        return self._value_index[s]
 
     def add_value(self, s):
         """ Add a value `s` to the list of values.
         """
         if not isinstance(s, str):
             raise TypeError("values of DiscreteVariables must be strings")
+        self._value_index[s] = len(self.values)
```
VesnaT: Since this is a public method, I'd add a check `if s not in self._value_index`, before inserting a new value.

Orange/widgets/data/owconcatenate.py
```diff
@@ -172,18 +179,17 @@ def incompatible_types(self):
...
             tables = self.more_data.values()
-            if self.merge_type == OWConcatenate.MergeUnion:
-                domain = reduce(domain_union,
-                                (table.domain for table in tables))
-            else:
-                domain = reduce(domain_intersection,
-                                (table.domain for table in tables))
+            domains = [table.domain for table in tables]
+            oper = set.union if self.merge_type == OWConcatenate.MergeUnion \
+                else set.intersection
```
VesnaT: Why passing `oper` parameter? It can be set inside `self.merge_domains`.

Orange/widgets/data/owconcatenate.py
```diff
@@ -236,36 +242,74 @@ def send_report(self):
...
+            if desc.original:
+                new_attr = attr
+            elif desc.template.is_discrete:
+                new_attr = attr.copy()
+                for val in desc.values[len(attr.values):]:
+                    new_attr.add_value(val)
+            else:
+                assert desc.template.is_continuous
+                new_attr = attr.copy(number_of_decimals=desc.number_of_decimals)
+            new_attrs.append(new_attr)
```
VesnaT: If `number_of_decimals` is passed as an argument, it should somehow be recalculated from all variables, otherwise the result depends on which dataset was set as a Primary Data.

Screenshot 2020-02-24 at 10 20 27

Orange/data/variable.py
```diff
@@ -556,9 +556,11 @@ def repr_val(self, val):
...
 
     def copy(self, compute_value=None, *, name=None, **kwargs):
-        var = super().copy(compute_value=compute_value, name=name,
-                           number_of_decimals=self.number_of_decimals,
-                           **kwargs)
+        var = super().copy(
+            compute_value=compute_value, name=name,
+            number_of_decimals=kwargs.pop("number_of_decimals",
+                                          self.number_of_decimals),
+            **kwargs)
```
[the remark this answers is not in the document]
janezd: You're right. This should go through setter. I changed it to:

```python
    def copy(self, compute_value=None, *, name=None, **kwargs):
        number_of_decimals = kwargs.pop("number_of_decimals")
        var = super().copy(compute_value=compute_value, name=name, **kwargs)
        if number_of_decimals is not None:
            var.number_of_decimals = number_of_decimals
        return var
```

`adjust_decimals` is a flag that tells whether the number of decimals was fixed (e.g. through setter) or is it being adjusted to the largest number of decimals (while reading the file). See function `val_from_str_add_cont`.

Orange/data/variable.py
janezd: True. But this commit is from #4422, on which *this* PR was based. #4422 was already merged with this code.

I agree that `values` should be a property and it should also be a tuple, not a list. But if I remember correctly, I already tried to change this once but ran into problems. At any rate, it's not in this PR, but if you'd like to try it, I support it.

Orange/data/variable.py
janezd: Same as above. I agree, but this is from #4422, which is already merged.

Orange/widgets/data/owconcatenate.py
janezd: Inertia. :)

Orange/widgets/data/owconcatenate.py
janezd: Huh, works for me. Number of decimals is adjusted a few lines earlier,

```python
                if el.number_of_decimals > desc.number_of_decimals:
                    attrs[el] = attrs[el]._replace(
                        original=False,
                        number_of_decimals=el.number_of_decimals)
```

janezd: This is even tested.  In `test_get_unique_vars` -- it just doesn't try different orders of connections. But the first attribute has a smaller number of decimals.

Orange/data/variable.py
janezd: I now implemented this in #4450.

Plus fixing some lint issues for a good measure.

janezd:
@VesnaT, I added a test that tries different orders of adding signals: https://github.com/biolab/orange3/pull/4425/commits/98b0e15e7451d507eca66f7679add8a4140546b5#diff-e8327cf214d283cae48e36d934820db9R351.

Can you pull the PR and run the test? If it passes: can you modify it to replicate what you're doing in canvas, so that it will fail?

PR #4431, opened 2020-02-17
[FIX] Ensure unique var names in file

AndrejaKovacic:
##### Issue

Fixes part of  #4382
File widget allowed renaming vars with no regards for possible duplicates it may create.

##### Description of changes
As in other widgets, duplicates are checked by get_unique_name_duplicates and a warning is shown.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/tests/test_util.py
+++ b/Orange/data/tests/test_util.py
@@ -59,10 +59,39 @@ class TestGetUniqueNames(unittest.TestCase):
             ["x (2)", "x (3)", "x (1)"])
         self.assertEqual(
             get_unique_names_duplicates(["x (2)", "x", "x", "x (2)", "x (3)"]),
-            ["x (2) (1)", "x (1)", "x (4)", "x (2) (2)", "x (3)"])
+            ["x (2) (1)", "x (4)", "x (5)", "x (2) (2)", "x (3)"])
+        self.assertEqual(
+                        get_unique_names_duplicates(["iris", "iris", "iris (1)"]),
+                        ["iris (2)", "iris (3)", "iris (1)"])
+
+        self.assertEqual(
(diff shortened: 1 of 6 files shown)


janezd:
This changes the user input -- and the change, at least on my Qt, isn't visible until I click Apply and then click one of variables.

I would change it so that the user either wouldn't be allowed to use the name that already exists (this would be a typical solution when renaming files) or I would keep the names set by the user as they are, and only rename the output variables, in the background (this is what most widgets do).

I am more inclined towards the latter. We will need this same thing in many other widgets, so perhaps it would make sense to write a function `Orange.data.util.create_safe_domain(attributes, class_vars, variables)` (name is up to discussion) that would get "proposed" names of attributes, class_vars and variables, rename them as needed and return a tuple `(domain, renamed_variables)`.

janezd:
@AndrejaKovacic, I'm so sorry that I did it again. :( After I realized my above comment makes no sense, I started poking around and changed the code here and there to see how this should be done ... and ... then it felt stupid to throw the changes away. I promise to discipline myself in the future.

You changed the code in the right place I think, but there were two problems. One is that it modifies the user's modifications, as I complained above. The other is that `DomainEditor.get_domain` used an attribute (`renamed_variables`) to return result. That is, `renamed_variables` is not a property of the object but just an information that is stored there to be retrieved after calling `get_domain`.

I changed your code so that it returns renamed variables as a part of result and so that it renames variables without changing the model.

AndrejaKovacic:
To break the current iteration, rename the attributes to:
iris (1) (1)
iris (1)
iris (1)
iris 
iris

janezd:
This time, the problem was that I didn't escape a regular expression, so it didn't properly parse names that included parentheses. It would be even more interesting if you named a variable `"unclosed parentheses("`.

PR #4433, opened 2020-02-17
[ENH] Predictions: Update splitter on resize

### LINKED ISSUE #4430: Predictions: smarter resizing
Predictions widget has a very interesting approach to resizing. Main area is uniformly resized, meaning the model (left) part and data (right) part are resized both at the same time. This often makes the model part full of white space and then the user has to drag the parting line around to adjust it.

Resizing the left part would stop once all the learners are visible.
Probably not easy to implement. Perhaps even impossible. But it would be nice.

**Additional context**
Left part too big, right part too small. It is annoying.
Screen Shot 2020-02-17 at 11 59 10

janezd:
##### Issue

Fixes #4430.

Fails (and will continue to fail) at coverage.

##### Description of changes

Update splitter on resize event.
 
##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -590,6 +590,10 @@ class OWPredictions(OWWidget):
             self.report_table("Data & Predictions", merge_data_with_predictions(),
                               header_rows=1, header_columns=1)
 
+    def resizeEvent(self, event):
+        super().resizeEvent(event)
+        self._update_spliter()
+
 
 class PredictionsItemDelegate(QStyledItemDelegate):
     """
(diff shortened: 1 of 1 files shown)


ajdapretnar:
I personally love this fix, but a few more people should probably agree this is the desired behaviour. @markotoplak @lanzagar @PrimozGodec

lanzagar:
I definitely think it is better than before.
But in the (rare?) case when the splitter is manually changed I find the abrupt jump on resize a bit jarring. Maybe it would be best to keep the left size constant on resizes and stretch/shrink just the data view? Not sure if that complicates the implementation?

janezd:
> Maybe it would be best to keep the left size constant on resizes and stretch/shrink just the data view?

Like this?

- When the left part is changed, the splitter would reset to fit the left side. (Like it does now on resize.)
- The user can manually move it.
- When the window is resized, the left side would keep the same (no matter whether it was set automatically or by the user).

janezd:
> Maybe it would be best to keep the left size constant on resizes and stretch/shrink just the data view?

Would this be OK?

- When the left part is changed, the splitter would reset to fit the left side. (Like it does now on resize.)
- The user can manually move it.
- When the window is resized, the left side would keep the same (no matter whether it was set automatically or by the user).

janezd:
I changed it as described above.

lanzagar:
:+1:

PR #4444, opened 2020-02-20
Add function data.util.get_unique_names_domain

janezd:
##### Issue

Widgets that construct new domains (see #4382) may need this function to prepare unique names.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation


AndrejaKovacic:
This is basically a more elegant implementation of _domain_rename_duplicates from Merge data widget. I was thinking we should extract that one to utils and use it everywhere. When this is merged, I think we should change owmergedata.py accordingly.

PR #4448, opened 2020-02-24
Omit all test folders from .coveragerc

JakaKokosar:
Some test folders are inflating code coverage. Should we consider dropping them from coverage report?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,7 +7,7 @@ source =
     Orange
 omit =
     Orange/canvas/*
-    Orange/tests/*
+    */tests/*
     */setup.py
     */*/setup.py
 


.coveragerc
```diff
@@ -7,7 +7,7 @@ source =
     Orange
 omit =
     Orange/canvas/*
-    Orange/tests/*
+     */test*
```
janezd: I agree, but could we use `*/tests/*` instead of `*/test*`? `*/test*` covers `Orange/evaluation/testing.py`, which is an ordinary module.

If `*/tests/*` doesn't suffice, we can also exclude `*/test_*`. I suppose no module will start with `test_`.

JakaKokosar: Good catch. Thanks!

PR #4451, opened 2020-02-25
[ENH] OWCreateClass: Data info displayed in the status bar

aturanjanin:
#### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcreateclass.py
+++ b/Orange/widgets/data/owcreateclass.py
@@ -15,6 +15,7 @@ from Orange.widgets import gui, widget
 from Orange.widgets.settings import DomainContextHandler, ContextSetting
 from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.widgetpreview import WidgetPreview
+from Orange.widgets.utils.state_summary import format_summary_details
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owcreateclass.py
```diff
@@ -255,7 +259,11 @@ def set_data(self, data):
         if not model:
             self.attribute = None
             self.Outputs.data.send(None)
+            self.info.set_input_summary(self.info.NoInput)
```
VesnaT: self.data is not always None when there is no model.
In case of 'Iris' dataset on the input, the input info summary is not correct.

Orange/widgets/data/owcreateclass.py
```diff
@@ -475,6 +483,8 @@ def apply(self):
             domain.attributes, new_class, domain.metas + domain.class_vars)
         new_data = self.data.transform(new_domain)
         self.Outputs.data.send(new_data)
+        self.info.set_output_summary(len(new_data),
```
VesnaT: If class name is deleted, the output info summary is not correct.

PR #4455, opened 2020-02-26
[ENH] OWCorrelations: Data info displayed in the status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcorrelations.py
+++ b/Orange/widgets/data/owcorrelations.py
@@ -25,6 +25,7 @@ from Orange.widgets.utils import vartype
 from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.signals import Input, Output
 from Orange.widgets.utils.widgetpreview import WidgetPreview
+from Orange.widgets.utils.state_summary import format_summary_details
 from Orange.widgets.visualize.utils import VizRankDialogAttrPair
 from Orange.widgets.widget import OWWidget, AttributeList, Msg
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owcorrelations.py
```diff
@@ -352,12 +356,16 @@ def set_data(self, data):
                 cont_data = Table.from_table(Domain(cont_vars), data)
                 remover = Remove(Remove.RemoveConstant)
                 cont_data = remover(cont_data)
+                self.info.set_input_summary(len(data),
```
VesnaT: Info is not updated with one instance on the input.

Orange/widgets/data/owcorrelations.py
```diff
@@ -384,6 +392,7 @@ def commit(self):
             self.Outputs.data.send(self.data)
             self.Outputs.features.send(None)
             self.Outputs.correlations.send(None)
+            self.info.set_output_summary(self.info.NoOutput)
```
VesnaT: This it not correct info when `self.cont_data is None` and `self.data is not None`.

Orange/widgets/data/owcorrelations.py
```diff
@@ -402,6 +415,8 @@ def commit(self):
         corr_table.name = "Correlations"
 
         self.Outputs.data.send(self.data)
+        self.info.set_output_summary(len(self.data),
```
VesnaT: To avoid duplicated code, I'd suggest removing upper to lines and add something like:

```
    def commit(self):
        self.Outputs.data.send(self.data)
        summary = len(self.data) if self.data else self.info.NoOutput
        detail = format_summary_details(self.data) if self.data else ""
        self.info.set_output_summary(summary, detail)

        if self.data is None or self.cont_data is None:
            self.Outputs.features.send(None)
            self.Outputs.correlations.send(None)
            return
```

at the beginning of the function.

PR #4458, opened 2020-02-26
gui: add accidentally removed ValueCallback

markotoplak:
Removing ValueCallback from imports in gui (PR #4158) broke Spectroscopy.

Changing imports, to me, seems the same as renaming a function without deprecation period. I therefore brought it back.

### DIFF
--- a/Orange/widgets/gui.py
+++ b/Orange/widgets/gui.py
@@ -31,7 +31,7 @@ from orangewidget.gui import (
     VisibleHeaderSectionContextEventFilter,
     checkButtonOffsetHint, toolButtonSizeHint, FloatSlider,
     ControlGetter, VerticalScrollArea, ProgressBar,
-    ControlledCallback, ControlledCallFront, connectControl,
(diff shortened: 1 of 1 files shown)


janezd:
The title is misleading. Removal was not accidental. But it may had been hasty, overzealous. :)

Maybe we should also re-add other classes that were removed?

Maybe we should add deprecation warnings? The only (reasoable) way I see is to derive a class from `ValueCallback` and put a warning into its `__init__`.

markotoplak:
That module is full stuff that is imported and exposed just for backward compatibility. Perhaps we should deprecate all of it.

Until we do, why wouldn't we add this one back? Not even pylint complains - probably because I added it to __all__. Wasn't that the reason for the removal?

janezd:
> Until we do, why wouldn't we add this one back? 

Sure, let us. I would already merge this is if I wouldn't wonder whether to add the rest, too.

janezd:
I'm sorry: other removed functions belong to a different category. :) Merging.

PR #4466, opened 2020-02-27
[ENH] OWContinuize: Provide the same options as in Preprocess/Normalize

### LINKED ISSUE #4418: Unify normalization in Preprocess and Continuize widgets
Now that #4379 was merged, Continuize can not normalize sparse data.

Also, Preprocess and Continuize widgets have different normalization options. They should be probably be unified. Some options from Preprocess would also work for sparse data.

janezd:
##### Issue

Resolves #4418.

##### Description of changes

The widget now provides the same normalizations as Preprocess/Normalize.

Option "zero_based" is removed. See discussion in #4418.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcontinuize.py
+++ b/Orange/widgets/data/owcontinuize.py
@@ -1,13 +1,13 @@
 from functools import reduce
+from types import SimpleNamespace
 
 from AnyQt.QtCore import Qt
 
 import Orange.data
 from Orange.util import Reprable
 from Orange.statistics import distribution
-from Orange.preprocess import Continuize, Normalize
-from Orange.preprocess.transformation import \
-    Identity, Indicator, Indicator1, Normalizer
+from Orange.preprocess import Continuize
+from Orange.preprocess.transformation import Identity, Indicator, Normalizer
 from Orange.data.table import Table
 from Orange.widgets import gui, widget
 from Orange.widgets.settings import Setting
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I think this is a good solution. I would just add a warning when using something from Numeric section and no numeric data is present on the input. And same for categorical. Probably it would be nice to let the user know her selection will not have any effect.

janezd:
> I would just add a warning when using something from Numeric section and no numeric data is present on the input.

I guess I would prefer not to do it.

We have a warning of a similar kind in Test & Score, where the widget warns the user if there is Test data on input, but the user didn't choose to use it. But there the user *showed an intention* and the "wrong" setting in the widget *affects the result*.

In Continuize, the setting would probably just be left like it was before (hence *no intention*) and, most importantly, the *setting has no effect on output* in this case.

By adding this warning -- would we warn the user to not use this setting when there is no numeric data? Ehm, no, it doesn't hurt. We would essentially be warning the user that he *showed an intention to normalize numeric features* but fed no numeric features to the widget. If this was the case, he would notice missing numeric features anyway. :)

I would avoid warnings that do not require user's attention, otherwise we make users imune to warnings (think "your connection is not secure" in browsers...)

ajdapretnar:
Ok.

PR #4472, opened 2020-02-28
[ENH] OWPivotTable: Data info displayed in the status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpivot.py
+++ b/Orange/widgets/data/owpivot.py
@@ -25,6 +25,7 @@ from Orange.widgets.settings import (Setting, ContextSetting,
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.widgetpreview import WidgetPreview
+from Orange.widgets.utils.state_summary import format_summary_details
 from Orange.widgets.widget import OWWidget, Input, Output, Msg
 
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owpivot.py
```diff
@@ -821,6 +829,11 @@ def commit(self):
         self.Outputs.pivot_table.send(self.pivot.pivot_table)
         self.Outputs.filtered_data.send(self.get_filtered_data())
 
+        output = self.pivot.pivot_table
```
VesnaT: I'd rather consider `filtered data` here.

aturanjanin: Even though `pivot table` is the default output of the widget and if a user wants filtered data on the output they have to change connection afterwards?

VesnaT: Yes.

Orange/widgets/data/owpivot.py
```diff
@@ -821,6 +829,11 @@ def commit(self):
         self.Outputs.pivot_table.send(self.pivot.pivot_table)
         self.Outputs.filtered_data.send(self.get_filtered_data())
 
+        output = self.get_filtered_data()
```
VesnaT: Since you are saving the filtered data into a temporary variable, you can do it before `self.Outputs.filtered_data.send(self.get_filtered_data())` and use it there as well instead of crating the data twice.

PR #4474, opened 2020-02-28
PyTableModel.__setitem__: properly emit signals

ajdapretnar:
##### Issue

`PyTableModel.__setitem__` didn't properly signal the change: it emitted `dataChanged` for indices in the existing slice, which was empty for empty models.

0. In Time Series, pull https://github.com/biolab/orange3-timeseries/pull/80.
1. Take Datasets, load Philadelphia. 
2. Connect to Aggregate, change Aggregation function for one of the attributes.
3. Disconnect Aggregate
4. Connect again. Previous changes are gone (because they were not saved because the model didn't emit the signal properly).

##### Description of changes

Add proper signals.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -102,6 +102,7 @@ class PyTableModel(AbstractSortTableModel):
     # pylint: disable=missing-docstring
     def __init__(self, sequence=None, parent=None, editable=False):
         super().__init__(parent)
+        self._rows = self._cols = 0
         self._headers = {}
         self._editable = editable
         self._table = None
@@ -109,10 +110,10 @@ class PyTableModel(AbstractSortTableModel):
         self.wrap(sequence or [])
 
     def rowCount(self, parent=QModelIndex()):
-        return 0 if parent.isValid() else len(self)
+        return 0 if parent.isValid() else self._rows
 
     def columnCount(self, parent=QModelIndex()):
-        return 0 if parent.isValid() else max(map(len, self._table), default=0)
+        return 0 if parent.isValid() else self._cols
 
     def flags(self, index):
         flags = super().flags(index)
@@ -214,6 +215,7 @@ class PyTableModel(AbstractSortTableModel):
             del self[row:row + count]
             for rowidx in range(row, row + count):
                 self._roleData.pop(rowidx, None)
+            self._rows = self._table_dim()[0]
             return True
         return False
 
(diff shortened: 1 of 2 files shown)


ajdapretnar:
@janezd Please provide a better description of the changes and PR as I am unable to.

Orange/widgets/utils/itemmodels.py
```diff
@@ -265,15 +265,21 @@ def __delitem__(self, i):
         self.endRemoveRows()
 
     def __setitem__(self, i, value):
+        self._check_sort_order()
         if isinstance(i, slice):
+            self.layoutAboutToBeChanged.emit()
```
ales-erjavec: Is there any reason for layoutAboutToBeChanged/layoutChanged surrounding the begin/endRemoveRows? They are unnecessary (layoutChanged notifies when internal positions of items has changed. e.g. ordering changes, if it is used it should explicitly update persistent indices `self.changePersistentIndex`).

I am not even sure it is supported by views that the model would change dimensions (rowCount/columnCount) between layoutAboutToBeChanged and layoutChanged.

Orange/widgets/utils/itemmodels.py
```diff
@@ -265,15 +265,21 @@ def __delitem__(self, i):
...
+            self.beginResetModel()
             start, stop, _ = _as_contiguous_range(i, len(self))
-            stop -= 1
+            self.removeRows(start, stop - start)
+            self.insertRows(start, len(value))
+            self._table[start:start + len(value)] = value
+            self.dataChanged.emit(
+                self.index(start, 0),
+                self.index(start + len(value) - 1, self.columnCount() - 1))
+            self.endResetModel()
```
janezd: @ales-erjavec, this code now calls `removeRows` and `insertRows` to properly handle `_roledata`, too. These two methods emit (begin|end)(Remove|Insert)Rows. Afterwards it modifies the table and emits `dataChanged`.

But still: if I connect Dataset (with Philadelphia crime) to Aggregate (from Time Series add-on), then remove the connection and connect it again, the view in Aggregate remains empty -- unless I call `beginResetModel` and `endResetModel` . After a day of exploration, I've no clue why. I see that the view does call `rowCount` and `columnCount` (which give correct results), all model's data seems properly initialized, but the view is empty.

Do you see why is this, or should I just remove `dataChanged` and we reset the model? In practice, I doubt it will make a difference because this code is almost always called when we replace the entire model's data (`self.model[:] = ...`).

ales-erjavec: The problem was that `columnsInserted|Removed` were not emitted. The model was cleared (`clear()`) which reset its column count to 0, then when the model was populated with `model[:] = ...` it only ever emitted `rowsInserted`, never `columnsInserted` event though the models `columnCount()` changed  implicitly.

PR #4477, opened 2020-03-01
[ENH] Edit Domain: Option to merge less frequent values

### LINKED ISSUE #4436: An option to group less common values

Sometimes I want to reduce number of different values for a discrete variable. It is helpful when I want to observe it in the box plot, to make cross-validation (without training on values that appears just in few cases), and observe it in the confusion matrix.
Currently, I make it in the python script.

I suggest an option to group values (let say in preprocess widget) that appears:
- less than n times
- reduce the least common values when number of values is less than n

PrimozGodec:
##### Issue

Fixes #4436

##### Description of changes
Added an option to merge less frequent values:
- values that have less than n appearances
- keep only n different values

The current implementation of merging selected values moved in a new dialogue. 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/oweditdomain.py
+++ b/Orange/widgets/data/oweditdomain.py
@@ -11,27 +11,24 @@ from itertools import zip_longest, repeat, chain
 from contextlib import contextmanager
 from collections import namedtuple, Counter
 from functools import singledispatch, partial
-
 from typing import (
     Tuple, List, Any, Optional, Union, Dict, Sequence, Iterable, NamedTuple,
     FrozenSet, Type, Callable, TypeVar, Mapping, Hashable
 )
+
+import numpy as np
+import pandas as pd
 from AnyQt.QtWidgets import (
     QWidget, QListView, QTreeView, QVBoxLayout, QHBoxLayout, QFormLayout,
     QToolButton, QLineEdit, QAction, QActionGroup, QGroupBox,
     QStyledItemDelegate, QStyleOptionViewItem, QStyle, QSizePolicy, QToolTip,
-    QDialogButtonBox, QPushButton, QCheckBox, QComboBox, QShortcut,
-    QStackedLayout
-)
+    QDialogButtonBox, QPushButton, QCheckBox, QComboBox, QStackedLayout,
+    QDialog, QRadioButton, QGridLayout, QLabel, QSpinBox, QDoubleSpinBox)
 from AnyQt.QtGui import QStandardItemModel, QStandardItem, QKeySequence, QIcon
 from AnyQt.QtCore import (
-    Qt, QEvent, QSize, QModelIndex, QAbstractItemModel, QPersistentModelIndex,
-    QRect,
+    Qt, QEvent, QSize, QModelIndex, QAbstractItemModel, QPersistentModelIndex
 )
 from AnyQt.QtCore import pyqtSignal as Signal, pyqtSlot as Slot
-
-import numpy as np
-import pandas as pd
 
 import Orange.data
 
@@ -72,6 +69,20 @@ class _DataType:
 
     def __hash__(self):
         return hash((type(self), super().__hash__()))
+
+    def name_type(self):
+        """
+        Returns a tuple with name and type of the variable.
+        It is used since it is forbidden to use names of variables in settings.
+        """
+        type_number = {
+            "Categorical": 0,
+            "Ordered": 1,
+            "Real": 2,
+            "Time": 3,
+            "String": 4
+        }
+        return self.name, type_number[type(self).__name__]
 
 
 #: An ordered sequence of key, value pairs (variable annotations)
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/oweditdomain.py
```diff
@@ -662,6 +673,168 @@ def on_label_selection_changed(self):
...
+        # if selected attributes available check the first radio button,
+        # it is most probable option
+        if selected_attributes:
+            radio1.setChecked(True)
+        else:
+            radio2.setChecked(True)
+
+        label2 = QLabel("occurrences")
+        label3 = QLabel("occurrences")
+        label4 = QLabel("the most frequent values")
```
lanzagar: Remove "the"

lanzagar:
After you select one of the merge options, set its parameter, apply the transformation and then see the result you might want to change the parameter. It is kind of annoying that opening the merge dialog again resets everything (it would be nicer to see what was selected previously and allow a minor correction).
Would that complicate the code a lot or is it simple to change?

Orange/widgets/data/oweditdomain.py
```diff
@@ -662,6 +673,168 @@ def on_label_selection_changed(self):
...
+        spin3.setDecimals(1)
+        spin3.setSingleStep(0.1)
+        spin3.setMaximum(100)
+        spin3.setValue(10)
+        spin3.setMinimumWidth(self.fontMetrics().width("X") * (2 + 1) + 20)
+        spin3.setSuffix(" %")
+        spin3.valueChanged.connect(partial(self._spin_changed, spin3))
+
+        self.n_values_spin = spin4 = QSpinBox()
+        max_val = min(10, len(variable.categories))
```
janezd: Is this limit needed? Couldn't we just limit to `len(variable.categories)`?

Orange/widgets/data/oweditdomain.py
```diff
@@ -662,6 +673,168 @@ def on_label_selection_changed(self):
...
+        self.frequent_rel_radio = radio3 = QRadioButton(
+            "Group values with less than"
+        )
+        self.n_values_radio = radio4 = QRadioButton(
+            "Group all except"
+        )
+
+        # if selected attributes available check the first radio button,
+        # it is most probable option
+        if selected_attributes:
```
janezd: What about even disabling (but not hiding - let the user know it's there!) the first option when there is no selection?

Orange/widgets/data/oweditdomain.py
```diff
@@ -662,6 +673,168 @@ def on_label_selection_changed(self):
...
+        # third row
+        grid_layout.addWidget(radio3, 2, 0, 1, 2)
+        grid_layout.addWidget(spin3, 2, 2)
+        grid_layout.addWidget(label3, 2, 3)
+        # fourth row
+        grid_layout.addWidget(radio4, 3, 0)
+        grid_layout.addWidget(spin4, 3, 1)
+        grid_layout.addWidget(label4, 3, 2, 1, 2)
+
+        group_box = QGroupBox("Grouping strategy")
```
janezd: "Strategy" sounds quite advanced. :) Maybe just "Grouping"? Or a box without a title?

Orange/widgets/data/oweditdomain.py
```diff
@@ -662,6 +673,168 @@ def on_label_selection_changed(self):
...
+        buttons.rejected.connect(self.reject)
+
+        # join components
+        self.setLayout(QVBoxLayout())
+        self.layout().addWidget(group_box)
+        self.layout().addLayout(name_hlayout)
+        self.layout().addWidget(buttons)
+        self.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
+
+    def _spin_changed(self, spin: QSpinBox) -> None:
```
janezd: Writing three separate functions instead of this one would actually be shorter (excluding docstring, which is not really needed for a one-line function). :)

lanzagar:
Also, don't forget to put an [ENH] tag in the PR title (for the changelog).

Orange/widgets/data/oweditdomain.py
```diff
@@ -662,6 +673,168 @@ def on_label_selection_changed(self):
...
+        grid_layout.addWidget(radio4, 3, 0)
+        grid_layout.addWidget(spin4, 3, 1)
+        grid_layout.addWidget(label4, 3, 2, 1, 2)
+
+        group_box = QGroupBox("Grouping strategy")
+        group_box.setLayout(grid_layout)
+
+        # grouped variable name
+        new_name_label = QLabel("New value name: ")
+        self.new_name_line_edit = n_line_edit = QLineEdit("Others")
```
janezd: I'm not sure, but think it would be better to call this "Other"? Like in zoo, type can be "mammal", "fish", "insect" or "other". Or, for instance, "cause" would be "human error", "mechanical error", "other".

Orange/widgets/data/oweditdomain.py
```diff
@@ -662,6 +673,168 @@ def on_label_selection_changed(self):
...
+                       if c / n_all * 100 < self.frequent_rel_spin.value()]
+        return np.array(self.variable.categories)[indices].tolist()
+
+    def get_merged_value_name(self) -> str:
+        """
+        Returns
+        -------
+        New label of merged values
+        """
+        return self.new_name_line_edit.text() or "Others"
```
janezd: As above.

Orange/widgets/data/oweditdomain.py
```diff
@@ -916,13 +1092,12 @@ def __init__(self, *args, **kwargs):
...
             shortcut=QKeySequence(Qt.ControlModifier | Qt.Key_Equal),
-            shortcutContext=Qt.WidgetShortcut,
-            enabled=False,
+            shortcutContext=Qt.WidgetShortcut
         )
 
         self.add_new_item.triggered.connect(self._add_category)
         self.remove_item.triggered.connect(self._remove_category)
-        self.merge_items.triggered.connect(self._merge_categories)
+        self.merge_items.triggered.connect(self._merge_less_common_categories)
```
janezd: Given that it also merges selected categories, the old name, `_merge_categories`, fits better.

Orange/widgets/data/oweditdomain.py
```diff
@@ -1182,65 +1361,63 @@ def _add_category(self):
...
-        sh.activated.connect(cb.close)
-        cb.setParent(self, Qt.Popup)
-        cb.move(vrect.topLeft())
-
-        cb.addItems(
-            list(unique(str(row.data(Qt.EditRole)) for row in rows)))
-        prows = [QPersistentModelIndex(row) for row in rows]
-
-        def complete_merge(text):
+        dlg = GroupLessFrequentItemsDialog(
```
janezd: Same as above - this dialog could have a more general name, like `GroupItemsDialog`.

janezd:
> It is kind of annoying that opening the merge dialog again resets everything

I also thought about this. Then I considered print dialogs, which usually do not remember page intervals, for instance.

But I agree with @lanzagar - if it's not too complicated to include this as a setting, please do.

Orange/widgets/data/oweditdomain.py
PrimozGodec: Agree. It is was changed since first I taught there will be two dialogues. It is fixed now.

Orange/widgets/data/oweditdomain.py
PrimozGodec: Actually limit was `len(variable.categories)`, it was just a setting for the current number. The name of the variable was misleading.

PrimozGodec: Anyway, it is changed now.

PrimozGodec:
I implemented that the dialogue remembers settings. Settings are not global (are not saved in schema, or context), but stays in the dialogue until data are changed or widget is removed.

PrimozGodec:
Now parameters are saved as a Context variable. The solution does not look nice, but I did not see any better option. If it looks too ugly we can remove it.

Orange/widgets/data/oweditdomain.py
```diff
@@ -662,6 +673,198 @@ def on_label_selection_changed(self):
...
+        )
+        self.frequent_rel_radio = radio3 = QRadioButton(
+            "Group values with less than"
+        )
+        self.n_values_radio = radio4 = QRadioButton(
+            "Group all except"
+        )
+
+        # if selected attributes available check the first radio button,
+        # othervise disable it
```
janezd: -> otherwise

Orange/widgets/data/oweditdomain.py
```diff
@@ -1613,6 +1838,7 @@ def __init__(self):
     @Inputs.data
     def set_data(self, data):
         """Set input dataset."""
+        self._merge_dialog_settings = self._editor.get_merge_context()
```
janezd: This will not save the context when the widget is closed. I believe the proper solution is to define `storeSpecificSettings`:

```python
        self._merge_dialog_settings = self._editor.get_merge_context()
```

PrimozGodec: Thank you, I didn't think about that situation. It is fixed now.

PR #4481, opened 2020-03-02
[ENH] Testing with Tox

### LINKED ISSUE #4324: Testing: switch to tox
`python setup.py test` issues a warning:
`WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.`

We should look into tox and see how to test with it.

JakaKokosar:
##### Issue
Fixes #4324

Invoking ```python setup.py test ``` is now deprecated. Judging by this comment https://github.com/pypa/setuptools/issues/1684#issuecomment-508311518, authors of setuptools have discouraged this for quite some time now. 

They suggest that Tox is used as a new "frontend" for running project tests. While this is not strictly necessary for us to do (we can use unittest/pytest directly) it is nice because we can define the steps for building and testing Orange in one place. This means that by invoking tests with Tox we can run them on any CI (or locally) in a python environment that is defined and created by this tool. Also, it seems that using Tox is becoming a trend in all major python projects.

Note that this adds a bit of overhead. It creates new .tox folder that is used to create new test environments. If one would like to bypass Tox entirely you need to use unittest/pytest or invoking setup.py while it lasts :). 

##### Description of changes

New configuration file tox.ini is added. 

Some of the steps that are defined and ready to use in this PR:
- pip install tox
- running: 
    - ```tox -e py36/py37/py38 ``` will run entire Orange test suit on python 3.6/3.7/3.8 (if it's installed on you machine)
    - ```tox -e coverage``` will run tests and output coverage report.
    - ```tox -e build_doc``` will build documentation and fail if something is broken.
    - ```tox -e add-ons``` will build orange with your changes, install some of the supported add-ons (we can extend this list) and run test discovery in orangecontrib folder.
 
See #4482 how tox can be used in CI configuration.


Orange/widgets/data/tests/test_owoutliers.py
```diff
@@ -47,8 +47,8 @@ def test_outputs(self):
         inliers = self.get_output(self.widget.Outputs.inliers)
         outliers = self.get_output(self.widget.Outputs.outliers)
         data = self.get_output(self.widget.Outputs.data)
-        self.assertEqual(len(inliers), 135)
-        self.assertEqual(len(outliers), 15)
+        self.assertLessEqual(len(inliers), 136)
+        self.assertGreaterEqual(len(outliers), 14)
```
lanzagar: Wouldn't it be better to use `assertIn` here (like you did below)?

lanzagar:
I have some trouble running this locally, so I would go over it together with you when we get a chance.
(`tox -e py38` ended with `ERROR:   py38: commands failed`)

lanzagar:
If anyone else wants to take a look please do. Otherwise I am in favor of merging this.
(my problems with some widget tests are unrelated to the changes here and can be investigated separately)

PR #4484, opened 2020-03-02
[FIX] OWEditDomain: Clear editor when data is disconnected

### LINKED ISSUE #4483: EditDoman doesn't clear editor when data is disconnected
**Describe the bug**

If we give an input to OWEditDomain and then disconnect it the variables remove but the editor keeps the last change.

1. Select Iris in File and connect it to EditDomain
2. Remove connection

**Orange version:**
3.25.dev

**Expected behavior**
Widget clears its state when the input data is disconnected.

**Screenshots**
editdomain

**Operating system:**
Windows 10

aturanjanin:
##### Issue
Fixes #4483 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/oweditdomain.py
+++ b/Orange/widgets/data/oweditdomain.py
@@ -1382,9 +1382,9 @@ class ReinterpretVariableEditor(VariableEditor):
             typecb.activated[int].connect(self.__reinterpret_activated)
             form.insertRow(1, "Type:", typecb)
             # Insert the typecb after name edit in the focus chain
-            nameedit = editor.findChild(QLineEdit, )
-            if nameedit is not None:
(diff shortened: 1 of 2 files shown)


janezd:
Line 1: "should": I meant "shouldn't"! :)

aturanjanin:
What if we add `self.layout().currentWidget().clear()` call into the already existing function `OWEditDomain.clear_editor` and call that function in the `OWEditDomain.clear`, would that be a fine solution maybe?
Also, current widget can only be one of four - Discrete/Continuous/Time/Variable Editor.

janezd:
OK, you can do so, too, I guess.

PR #4491, opened 2020-03-05
Box Plot: Correct mean label rounding

PrimozGodec:
##### Issue

Mean label not rounded correctly when only whole numbers in data (read from file). In this case, the domain sets the formatting string to 0 decimal numbers. Box plot should not use `attr.str_val` for mean since in the described case mean of whole numbers can be a decimal number.

Screenshot 2020-03-04 09 11 22

##### Description of changes

Mean rounded to MAX_NUM_OF_DECIMALS (defined by a variable).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


PrimozGodec:
Closing since @lanzagar will prepare the pull request with a more complete fix.

PR #4499, opened 2020-03-06
[ENH] OWImpute: Data info displayed in the status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/data/owimpute.py
```diff
@@ -346,6 +354,7 @@ def commit(self):
 
         if not self.data or not self.varmodel.rowCount():
             self.Outputs.data.send(self.data)
+            self.info.set_output_summary(self.info.NoOutput)
```
VesnaT: When `self.data is not None` and `self.varmodel.rowCount() == 0`, there actually is data on the output.

PR #4516, opened 2020-03-09
[WIP] OWPaintData: Fix Reset to Input; Data info displayed in the status bar

### LINKED ISSUE #4514: PaintData: Output doesn't reset when Reset to Input Data is pressed
**Describe the bug**
When some file is set as input of Paint Data and if we add more points to the plot and then click `Reset to Input Data`, the plot resets but the actual output doesn't, the widget outputs all file instances plus those added ones. 

Additionally, when `Clear` is pressed the plot is empty and output is None and if then we click `Reset to Input Data`, the plot shows input instances but output is still None. 

**To Reproduce**
1. Connect Iris file to the Paint Data widget
2. Add more points on the plot / Clear the plot
3. Click `Reset to Input Data`
4. Connect Paint Data to Data Table and see the number of output instances

**Orange version:**
3.25.dev

**Expected behavior**
The widget resets and outputs (only) the input data.

aturanjanin:
##### Description of changes
Fixes #4514 .
Input / Output data info displayed in the status bar.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpaintdata.py
+++ b/Orange/widgets/data/owpaintdata.py
@@ -1029,7 +1029,6 @@ class OWPaintData(OWWidget):
         else:
             self.input_data = np.column_stack((X, y))
         self.reset_to_input()
-        self.unconditional_commit()
 
     def reset_to_input(self):
         """Reset the painting to input data if present."""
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owpaintdata.py
```diff
@@ -1029,7 +1037,6 @@ def _check_and_set_data(data):
         else:
             self.input_data = np.column_stack((X, y))
         self.reset_to_input()
-        self.unconditional_commit()
```
VesnaT: The widget should output the data (when opening a saved workflow) even if the `Send Automatically` is off. That is the point of `self.unconditional_commit` and should be kept.

PR #4517, opened 2020-03-09
Update README.md

daikikatsuragawa:
Syntax highlighting adaptation

##### Includes
- [ ] Code changes
- [ ] Tests
- [X]  Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,21 +24,28 @@ Orange requires Python 3.6 or newer.
 
 First, install [Miniconda] for your OS. Create virtual environment for Orange:
 
-    conda create python=3 --name orange3
-
+```Shell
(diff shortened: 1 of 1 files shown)


janezd:
Thanks for the PR.

It doesn't do much, though. The only change that I can spot is that the word "source" in one of the snippets is shown in blue. Given that this is the only blue word, I'd prefer to not change it -- it looks as if it's blue by mistake.

lanzagar:
I like it from the perspective of a better defined markdown - instead of a generic block I like that we mark it as shell code since that is what it is. Even though the effect is minor (I would not worry too much about the single highlighted word being blue or not).

Issue #4276, opened 2019-12-18
Python Script progress log


Hrovatin:
Currently the Python Script widget prints to the console only after it has finished running and there is no indicator of whether it is running or not. 
It should be somehow indicated whether the script is currently running or not. Otherwise it is not clear whether 1.) the script even started running and 2.) why Orange is unresponsive.

irgolic:
Something akin to a radio button next to the `Run` button, that's red when the script is running.

I think Orange would benefit a lot from a strong Python script widget, but I appreciate it's a relatively big project.

Maybe after fixing this issue, we could separate out some smaller tasks, like:
- [ ] Add line numbers
- [ ] Rework the Library view (e.g., changing Update to Save),
- [ ] Run script in parallel thread/process, so stdout shows as the script runs

Also (the biggest task), maybe we could improve the data table (`in_data`/`out_data`) interface in some way. A user in the Discord support channel was confused why `in_data['Date',]= str(datetime.datetime.strptime(date, "%d.%m.%Y"))` wasn't working. The closer the exposed interface is to something users are familiar with, the flatter the widget's learning curve.

(sorry for hijacking your issue)

ajdapretnar:
> A user in the Discord support channel was confused why in_data['Date',]= str(datetime.datetime.strptime(date, "%d.%m.%Y")) wasn't working.

This is because most people are used to working with pandas, but Orange's data structures are not like that. That said, this is something only @janezd can answer.

pavlin-policar:
While the script widget can be improved in lots of ways, the thing that annoys me the most is that stdout is not flushed while running. In this simple example,

```
import time
time.sleep(2)
print('Hello world')
time.sleep(2)
print('Hello world')
time.sleep(2)
print('Hello world')
```

_hello world_ is printed three times when _everything_ finishes, not every two seconds. The fix for this is probably just adding `sys.stdout.flush()` after every line and probably wouldn't require a major rewrite.

Also, in addition to printing actual stdout, I would suggest outputting the commands, like ipython does. So the output for the simple example above would be:

```
>>> import time
>>> time.sleep(2)
>>> print('Hello world')
Hello world
>>> time.sleep(2)
>>> print('Hello world')
Hello world
>>> time.sleep(2)
>>> print('Hello world')
Hello world
```

instead of just

```
Running script
Hello world
Hello world
Hello world
```

janezd:
[...]

PR #4528, opened 2020-03-13
[WIP] PythonScript: Execute in separate thread

janezd:
##### Issue

Attempt at #4276.

WIP, just exploring the principle.

##### Description of changes

I moved execution into a separate thread. It gets commands through a signal (hence they are executed within this thread's event loop). At flush, they call `qApp.processEvents()`.

While writing this description, I checked and realized that `processEvents()` will process the caller's thread and not the main thread. (Yet it still works?!)

@ales-erjavec, does what I'm doing make any sense - and could it be salvaged?

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Thanks. I guess I'm not up to this task.

Issue #4464, opened 2020-02-27
Improve portability of Orange3 workspaces by using relative instead of fixed file paths


c-a-schiller:
**Is your feature request related to a problem? Please describe.**

Orange3 workspaces are not portable because the widgets with load/save functions use fixed/hardcoded file paths. That means if a workspace is moved from one machine to another, all load/save widgets have to be touched again manually, and parameters of load/save widgets have to be set again, which is very cumbersome and time-consuming.

**Describe the solution you'd like**

If all widgets with load/save file functions had an option to store relative paths instead of fixed/hardcoded ones, workspaces would be much more easy to port from machine to machine.

**Describe alternatives you've considered**

I don't see an alternative to my proposed solution right now, except the above-mentioned manual process, which is very slow, cumbersome and error-prone (due to the fact that widget parameters get lost when re-linking to a new file on the new machine.

**Additional context**

n/a

janezd:
Relative to what?

c-a-schiller:
> Relative to what?

Relatively to the location of the workspace (.ows) file.

janezd:
I don't like the an-option-to-store-relative-paths part. I assume this would be a checkbox that many would not understand, and we'd like to avoid such things if possible. Computer scientists understand the concept of "relative path", while most lay people don't.

Always saving relative paths would be a bad idea, too, because the file can be completely elsewhere, and moving the workflow file shouldn't affect the file's path in this case.

Currently, the File and Distance File widgets stores absolute paths, but resolves them later by searching for the file in "possible relative paths". If you, for instance, save the workflow in the same directory as the file, or if the file is in a subdirectory of the directory with the workflow's file, re-loading files should work.

The Load Model widget doesn't do it, but it should.

For widget that save files (Save, Save Model, Save Distance Matrix...), we don't have a useful mechanism, and I don't think we should change this. I suppose it's OK that saving requires user's action.

c-a-schiller:
Well, my usage scenario for Orange is to use it for teaching the basics of machine learning already early in school, starting around 7th grade. For this, I am currently developing a "playful" machine learning scenario. The informatics and/or mathematics teachers I am going to interact with (while discussing how to integrate the machine learning part into their regular school curriculums) are not data scientists. They may have decent IT knowledge, but I still think that it would be beneficial to spare them of any non-necessary manual modifications to a pre-provided Orange Workspace, such as having to modify all Load and Save widgets in the Workflow before being able to use it.

[...]

PR #4532, opened 2020-03-14
[ENH] Save widgets: Store paths relative to workflow directory

janezd:
##### Issue

Fixes part of #4464 by storing paths relative to the workflow directory. 

##### Description of changes

Path is always relative to the workflow directory, and the saved file must be either in the same director or in its subtree. On loading the widget, auto-save is disabled unless a relative path is used to prevent auto saving to arbitrary locations when the workflow is loaded.

Implementation does not use `RecentPath` because it does not fit.

All core widgets for saving (Save, Save Distances, Save Model) are derived from this base class, so the patch should work for them, as well as for all properly written widgets in add-ons (Save Network).

I took care to keep the backward compatibility, though ... please check all potentially affected add-ons.

Unlike codecov, I believe that coverage is 100 %. At least judging by codecov's diff.

##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owsave.py
+++ b/Orange/widgets/data/tests/test_owsave.py
@@ -79,7 +79,6 @@ class TestOWSave(OWSaveTestBase):
 
     def test_initial_start_dir(self):
         widget = self.widget
-        widget.filename = _w("/usr/foo/bar.csv")
         self.assertEqual(widget.initial_start_dir(),
                          _w(os.path.expanduser("~/")))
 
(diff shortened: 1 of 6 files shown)


PrimozGodec:
The implementation works correctly. There is only one thing I would change.

When setting a location to save data the text at the save button changes to `Save as <filname>` if I save data with name `foo.csv` the button has the text `Save as foo.csv`. Then I save the workflow and reopen it. I would like to see a button with the same text again (it would give a user feels that the path to save data is already set). After reopening text at the button is just `Save`

PrimozGodec:
Thank you. It is great now.

PR #4535, opened 2020-03-15
DiscreteVariable: Save color attributes with values

### LINKED ISSUE #4515: Header type annotation (colours) not saved correctly
Colours are not saved/reloaded correctly. I tried to set colours for a categorical feature and then saved the file with header type annotation.  Upon importing the new file into the Orange the data does not have correct colours. The colours from the saved data are loaded correctly in terms of hues, but the categorical instances are mixed between the categories. 
Screenshot from 2020-03-09 13-32-24

janezd:
##### Issue

Discrete variables save user-assigned colors in a list. This fails if the order of values changes.

This PR is intended to fix #4515.

##### Description of changes

Save colors in dictionaries, with values as keys.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/utils/colorpalettes.py
+++ b/Orange/widgets/utils/colorpalettes.py
@@ -7,6 +7,7 @@ import numpy as np
 from AnyQt.QtCore import Qt
 from AnyQt.QtGui import QImage, QPixmap, QColor, QIcon
 
+from Orange.data import Variable, DiscreteVariable, ContinuousVariable
 from Orange.util import Enum, hex_to_color, color_to_hex
 
 NAN_COLOR = (128, 128, 128)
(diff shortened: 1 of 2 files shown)


Hrovatin:
The file is still not loaded properly. I saved it with _Add type annotations to header_, which correctly saved the colours as a dictionary.

markotoplak:
In this PR, `discrete_get_colors` always removes the palette (because it calls `set_colors`), even if that the palette would still be valid (and could be saved as a palette) if the first `if` form `retrieve_colors` is True.

Furthemore, even if the first `if` was false, is setting the var.attributes["color"] really needed if `_colors` are set? If the variable contained enough info to compute the colors, why change how are they stored?

janezd:
> In this PR, `discrete_get_colors` always removes the palette (because it calls `set_colors`), even if that the palette would still be valid (and could be saved as a palette) if the first `if` form `retrieve_colors` is True.

Good one. I'd change it to

```python
        nvalues = len(var.values)
        if var._colors is not None and len(var._colors) == nvalues:
            return var._colors

        colors = retrieve_colors()
        if var._palette is None:
            set_colors(var, colors)
        return colors
```

Or do you favour just replacing `set_color` with setting the color like this:

```python
        nvalues = len(var.values)
        if var._colors is None or len(var._colors) != nvalues:
            var._colors = retrieve_colors()
            var._colors.flags.writeable = False
        return var._colors
```

> Furthemore, even if the first `if` was false, is setting the var.attributes["color"] really needed if `_colors` are set? If the variable contained enough info to compute the colors, why change how are they stored?

Unless I'm mistaken, `var.attributes["color"]` is not set. The function calls `set_colors`, not `discrete_set_colors`.

markotoplak:
I went with (2).

PR #4536, opened 2020-03-15
Feature statistics: Reformat numbers

### LINKED ISSUE #4522: Feature Statistics and other widgets do not report min and max correctly - rounding error
When a variable has low values (e.g. FDR on the below image) some of the widgets (Heatmap annotation legend, Feature Statistics) do not report min and max correctly due to the rounding. This can result in min and max being the same, despite the variable in fact having different values. In such cases scientific notation could be used instead of decimal notation, rounding to N significant instead of N decimal places. 
Screenshot from 2020-03-11 13-43-45

janezd:
##### Issue

Fixes #4522.

##### Description of changes

- Instead of `%.2f`, it now uses `str_val`. `str_val` will show a larger number of decimals in the future (in another pull request).
- Model's `data` method was split into functions for readability and lower cyclomatic complexity.

##### Includes
- [X] Code changes
- [ ] Tests

### DIFF
--- a/Orange/widgets/data/owfeaturestatistics.py
+++ b/Orange/widgets/data/owfeaturestatistics.py
@@ -7,14 +7,13 @@ TODO:
 """
 
 import datetime
-import locale
 from enum import IntEnum
 from typing import Any, Optional, Tuple, List
 
 import numpy as np
 import scipy.stats as ss
 import scipy.sparse as sp
-from AnyQt.QtCore import Qt, QSize, QRectF, QVariant, QModelIndex, pyqtSlot, \
+from AnyQt.QtCore import Qt, QSize, QRectF, QModelIndex, pyqtSlot, \
     QRegExp, QItemSelection, QItemSelectionRange, QItemSelectionModel
 from AnyQt.QtGui import QPainter, QColor
 from AnyQt.QtWidgets import QStyledItemDelegate, QGraphicsScene, QTableView, \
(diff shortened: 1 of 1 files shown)


janezd:
I'm not completely sold on `%g`. Maybe `number_of_decimals + 2` would be better.

Screenshot 2020-03-15 at 21 02 23

janezd:
If `abs(round(x, var.number_of_decimals) - x) > something reasonable, which depends upon number_of_decimals`, `str_val` adds two decimals.

Orange/widgets/data/owfeaturestatistics.py
```diff
@@ -515,60 +505,45 @@ def data(self, index, role):
...
-        elif column == self.Columns.MIN:
-            if role == Qt.DisplayRole:
-                if isinstance(attribute, DiscreteVariable):
-                    if attribute.ordered:
-                        output = attribute.str_val(self._min[row])
-                elif isinstance(attribute, TimeVariable):
-                    output = attribute.str_val(self._min[row])
+                    return format_time_diff(self._min[row], self._max[row])
+                elif isinstance(attribute, DiscreteVariable):
+                    return "%g" % self._dispersion[row]
```
lanzagar: This produces quite a bit more decimals then the previous rule of (always) 2 decimals.
And it usually looks a bit too much.
E.g. dispersion for the iris class (categ. var.) was previously reported as `1.10` and now `1.09861`

Maybe `"%.2g"` or `"%.3g"` would be better?
E.g. for the above iris example .2g produces just `1.1`
1/3 results in `0.33`, 1/3000 = `0.00033`, and 1/30000 = `3.3e-05`

I don't think precision>3 is needed for categ. vars in most cases. But I am not sure about 2 for cases like the `1.1` above (while `0.33` seems more acceptable).

Thoughts?

janezd: You're right. I changed it to `"%.3g"`.

PR #4540, opened 2020-03-16
heatmap: Change tick label display precison

ales-erjavec:
##### Issue

Ref gh-4522

##### Description of changes

Change tick display format for gradient color legend from .2f to .6g

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Hrovatin:
Now the min/max are correct. However, as there are more decimals the numbers overlap so they are not readable. I would suggest using scientific notation as it can be shorter.
Screenshot from 2020-03-19 20-05-39

Hrovatin:
This seems to work now.

PR #4546, opened 2020-03-18
[WIP] state_summary: Modification for missing values and multiple inputs

aturanjanin:
##### Description of changes
Information about missing values added.

`format_multiple_input` function for formatting details string for widgets with multiple inputs. This feature can be used when the next version of orange-widget-base is released.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/state_summary.py
+++ b/Orange/widgets/utils/state_summary.py
@@ -57,10 +57,46 @@ def format_summary_details(data):
         targets = format_variables_string(data.domain.class_vars)
         metas = format_variables_string(data.domain.metas)
 
+        features_missing = missing_values(data.has_missing_attribute()
+                                          and data.get_nan_frequency_attribute())
         n_features = len(data.domain.variables) + len(data.domain.metas)
         details = \
             f'{len(data)} instance{_plural(len(data))}, ' \
             f'{n_features} variable{_plural(n_features)}\n' \
-            f'Features: {features}\nTarget: {targets}\nMetas: {metas}'
-
+            f'Features: {features} {features_missing}\n' \
+            f'Target: {targets}\nMetas: {metas}'
     return details
+
+
+def missing_values(value):
+    if value:
+        return f'({value*100:.1f}% missing values)'
+    else:
+        return '(No missing values)'
+
+
+def format_multiple_inputs(inputs):
+    """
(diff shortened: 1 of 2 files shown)


Orange/widgets/utils/state_summary.py
```diff
@@ -57,10 +57,47 @@ def _plural(number):
...
+    summary for widgets that have more than one input.
+
+    :param inputs: A list of tuples for each input dataset where the first
+    element of the tuple is the name of the dataset and the second is the
+    dataset
+    :type inputs: list(tuple(str, Orange.data.Table))
+
+    :return: A formatted string
+    """
+
```
VesnaT: Inputs where dataset is None are insufficient handled. E.g.: formatted string for a widget with i.e. three inputs where two of those are None, results in horizontal lines with no text in between.

Screenshot 2020-03-20 at 14 31 48

I'd recommend introducing default strings for "None" datasets (i.e. "No data on input" for inputs and "No data on output" for outputs).

Orange/widgets/utils/state_summary.py
```diff
@@ -57,10 +57,47 @@ def _plural(number):
...
+    full_details = ''
+    for (name, data) in inputs:
+        details = new_line(format_summary_details(data))
+        if name == '':
+            full_details = f'{full_details}{details}<hr>'
+        else:
+            full_details = f'{full_details}{name}:<br>{details}<hr>'
+    # remove the last occurrence of the line "<hr>"
+    return full_details[:-4]
```
VesnaT: I'd avoid code like this. Maybe try using `"<hr>".join(details)` to construct the full detail, where `details` is list of 'detail' strings.

Orange/widgets/utils/state_summary.py
aturanjanin: Didn't we say that when there is no data on output there will be just output icon with no summary and/or detail? Or did you maybe mean that we could implement `format_multiple_inputs` function into widgets that have multiple outputs and the user can see data info for all outputs of the widget?

VesnaT: I meant for the widgets with multiple equivalent inputs. If all of them are None, just the icon is seen, without the summary and detail. But if one is not None and two are, then we get something similar as above.

PR #4547, opened 2020-03-18
[ENH] OWOutliers: Data info displayed in the status bar

aturanjanin:
##### Description of changes
Modification of the status bar data info display.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owoutliers.py
+++ b/Orange/widgets/data/tests/test_owoutliers.py
@@ -8,6 +8,7 @@ from Orange.data import Table
 from Orange.classification import LocalOutlierFactorLearner
 from Orange.widgets.data.owoutliers import OWOutliers, run
 from Orange.widgets.tests.base import WidgetTest, simulate
+from Orange.widgets.utils.state_summary import format_summary_details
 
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/tests/test_owoutliers.py
```diff
@@ -130,13 +131,17 @@ def test_nans(self):
 
     def test_in_out_summary(self):
         info = self.widget.info
+        data = self.iris
+        input_sum = self.widget.info.set_input_summary = Mock()
+        output_sum = self.widget.info.set_output_summary = Mock()
```
VesnaT: There is not much point in `self.assertEqual(info._StateInfo__input_summary.brief, "")` lines, since `self.widget.info.set_input_summary` was mocked.

VesnaT: The same goes for the code some lines below.
`info._StateInfo__input_summary.brief` could never have been changed, since mocked, so there is no point of testing it.

You could leave the test as it was, just add the lines to test the detail:
`self.assertEqual(info._StateInfo__input_summary.detail, detail)`

PR #4549, opened 2020-03-19
Data Table: change output when selecting is finished

PrimozGodec:
##### Issue

Table widget output the selection on every change (when the user still making his/her selection). It is annoying since the widget that is connected to the table is constantly performing its calculation or plotting. Some widgets even block the main thread for part of the second or more and it blocks the selection process.

##### Description of changes
After this change, the widget is outputting selection when selection is finished - when the user stop selecting.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -601,7 +601,12 @@ class OWDataTable(OWWidget):
         selmodel = BlockSelectionModel(
             view.model(), parent=view, selectBlocks=not self.select_rows)
         view.setSelectionModel(selmodel)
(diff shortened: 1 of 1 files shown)


PrimozGodec:
@ales-erjavec it can be covered with `keyReleasedEvent` (I added a line with this change), but I am still curious if there is any nicer solution. Is it maybe possible to know if the user is still in the process of selection?

ales-erjavec:
See gh-4552

PrimozGodec:
@ales-erjavec Thank you I will close this PR and merge yours.

PR #4550, opened 2020-03-19
[FIX] Detect duplicate names of variables in projections.

### LINKED ISSUE #4497: Linear Projection: Unique variable names isn't working properly.
**Describe the bug**
In some cases when changing the domain name, the selected data in the linear projection widget doesn't recognize the duplicates in the attribute names. The  `get_unique_names` function from `Orange.data.util ` isn't working properly.

**To Reproduce**
1. Import `iris` dataset
2. Change `petal-width` attribute to `C-y`
3. Use `Linear Projection` widget on the edited dataset and select a subsection of the dataset
4. List the selected subsection in the `Data Table` and we can see that, the duplicates in attribute names aren't detected.

**Orange version:**
3.24.1

**Expected behavior**
The duplicate names in attributes should be detected and properly.

**Screenshots**
Screen Shot 2020-03-06 at 12 10 03
**Operating system:**
OSX

NejcDebevec:
##### Issue
Fixes #4497 

##### Description of changes
Check for duplicates in original domain, not only in the selected one.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/utils/widget.py
+++ b/Orange/widgets/visualize/utils/widget.py
@@ -707,10 +707,18 @@ class OWAnchorProjectionWidget(OWDataProjectionWidget, openclass=True):
     def _get_projection_data(self):
         if self.data is None or self.projection is None:
             return None
+        proposed = [a.name for a in self.projection.domain.attributes]
+        names = get_unique_names(self.data.domain, proposed)
+
+        if proposed != names:
+            attributes = tuple([attr.copy(name=name) for name, attr in
+                                zip(names, self.projection.domain.attributes)])
+        else:
+            attributes = self.projection.domain.attributes
         return self.data.transform(
             Domain(self.data.domain.attributes,
                    self.data.domain.class_vars,
-                   self.data.domain.metas + self.projection.domain.attributes))
+                   self.data.domain.metas + attributes))
 
     def commit(self):
         super().commit()
--- a/Orange/widgets/visualize/tests/test_owlinearprojection.py
+++ b/Orange/widgets/visualize/tests/test_owlinearprojection.py
@@ -5,7 +5,7 @@ import numpy as np
 
 from AnyQt.QtCore import QItemSelectionModel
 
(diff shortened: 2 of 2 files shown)


Orange/projection/base.py
```diff
@@ -145,7 +145,7 @@ def proj_variable(i, name):
     def _get_var_names(self, n):
         postfixes = ["x", "y"] if n == 2 else [str(i) for i in range(1, n + 1)]
         names = [f"{self.var_prefix}-{postfix}" for postfix in postfixes]
-        return get_unique_names(self.orig_domain, names)
+        return names
```
VesnaT: Since `DomainProjection` can be used independently of any widget, I'd leave it as it was (retain the `get_unique_names`)

Orange/widgets/visualize/utils/widget.py
```diff
@@ -704,10 +704,17 @@ def _manual_move_finish(self, anchor_idx, x, y):
     def _get_projection_data(self):
         if self.data is None or self.projection is None:
             return None
+
+        names = get_unique_names(self.data.domain, [a.name for a in
+                                                    self.projection.domain.attributes])
+
+        attributes = tuple([a.copy(name=n) for n, a in
+                      zip(names, self.projection.domain.attributes)])
```
VesnaT: Lint is complaining about the indentation. Speaking of lint, please keep the line length up to 79 characters.

Orange/widgets/visualize/utils/widget.py
```diff
@@ -704,10 +704,17 @@ def _manual_move_finish(self, anchor_idx, x, y):
     def _get_projection_data(self):
         if self.data is None or self.projection is None:
             return None
+
+        names = get_unique_names(self.data.domain, [a.name for a in
```
VesnaT: Maybe the variables should be copied only when necessary, i.e. when the obtained names are different than the proposed.

PR #4551, opened 2020-03-19
[FIX] Paint Data: Send correct output after clicking Reset to Input

### LINKED ISSUE #4514: PaintData: Output doesn't reset when Reset to Input Data is pressed

aturanjanin:
##### Issue
Fixes #4514.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpaintdata.py
+++ b/Orange/widgets/data/owpaintdata.py
@@ -1059,6 +1059,8 @@ class OWPaintData(OWWidget):
         else:  # set_dimensions already calls _replot, no need to call it again
             self._replot()
 
+        self.commit()
+
     def add_new_class_label(self, undoable=True):
 
         newlabel = next(label for label in namegen('C', 1)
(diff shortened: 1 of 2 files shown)


lanzagar:
PR titles (with tags) are used to automatically compile the CHANGELOG.
Users will probably not know what changed when seeing this one in the list of changes.
Maybe something like:
[FIX] Paint Data: Send correct output after clicking Reset to Input
(or something better if anyone has a suggestion)

p.s. commit messages can be more technical as they are intended for developers.
An improvement for this one could be:
"OWPaintData: commit after reset_to_input()"

aturanjanin:
@lanzagar Thanks for your suggestions, I will fix this now.

PR #4569, opened 2020-03-23
state_summary: Add multiple inputs function and missing values info

aturanjanin:
##### Issue
Existing state_summary could not be used for widgets that have multiple inputs and has no information about missing values of input/output.

##### Description of changes
Information about missing values added.

`format_multiple_inputs` function added for formatting details string for widgets with multiple inputs.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/state_summary.py
+++ b/Orange/widgets/utils/state_summary.py
@@ -55,10 +55,49 @@ def format_summary_details(data):
         targets = format_variables_string(data.domain.class_vars)
         metas = format_variables_string(data.domain.metas)
 
+        features_missing = missing_values(data.has_missing_attribute()
+                                          and data.get_nan_frequency_attribute())
         n_features = len(data.domain.variables) + len(data.domain.metas)
         details = \
             f'{len(data)} instance{_plural(len(data))}, ' \
             f'{n_features} variable{_plural(n_features)}\n' \
-            f'Features: {features}\nTarget: {targets}\nMetas: {metas}'
-
+            f'Features: {features} {features_missing}\n' \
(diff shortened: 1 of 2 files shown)


VesnaT:
The `format_multiple_inputs` function should be generalized, so it can be used for outputs as well.

Orange/widgets/utils/tests/test_state_summary.py
```diff
@@ -100,13 +101,17 @@ def make_table(attributes, target=None, metas=None):
 
 
 class TestUtils(unittest.TestCase):
+    def setUp(self):
+        self.maxDiff = None
```
VesnaT: Is this ever used?

Orange/widgets/utils/state_summary.py
```diff
@@ -55,10 +55,50 @@ def _plural(number):
...
+
+    def new_line(text):
+        return text.replace('\n', '<br>')
+
+    full_details = []
+    for (name, data) in data_list:
+        if data:
+            details = new_line(format_summary_details(data))
+        else:
+            details = 'No data on input' if type_io == 'input' else \
```
VesnaT: The IF statement could be shortened:
`details = f'No data on {type_io}.'`

Plus, I'd prefer a dot at the end.

Orange/widgets/utils/tests/test_state_summary.py
aturanjanin: It came up as an error on travis before so I set it, but now it seems fine without it.

PR #4575, opened 2020-03-26
[ENH] Fix printing values with too few decimals

### LINKED ISSUE #4555: Values of numeric variables should sometimes show more decimals
`ContinuousVariable.str_val` shows `number_of_decimals` decimals. This is not OK when the value being printed is, for instance, an average of otherwise discrete values, like the average age in case when the original age was given in years, as integers.

`str_val` could easily detect such situations: if `abs(round(x, var.number_of_decimals) - x)` is larger than, for instance, `10 ** (-number_of_decimals - 5)`, it would use `number_of_decimals + 2` decimals.

janezd:
##### Issue

Fixes #4555.

##### Description of changes

Print two more decimals when the format is not `%g` and the difference from the rounded values is outside numeric error.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/tests/test_variable.py
+++ b/Orange/data/tests/test_variable.py
@@ -498,23 +498,42 @@ class TestContinuousVariable(VariableTest):
 
     def test_decimals(self):
         a = ContinuousVariable("a", 4)
-        self.assertEqual(a.str_val(4.654321), "4.6543")
-        self.assertEqual(a.str_val(4.654321654321), "4.6543")
+        self.assertEqual(a.str_val(4.6543), "4.6543")
+        self.assertEqual(a.str_val(4.25), "4.2500")
         self.assertEqual(a.str_val(Unknown), "?")
         a = ContinuousVariable("a", 5)
         self.assertEqual(a.str_val(0.000000000001), "0.00000")
         a = ContinuousVariable("a", 10)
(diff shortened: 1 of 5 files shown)


markotoplak:
This PR makes printing tables slower, but this operation does not need to be fast. It does not influence saving data, because the writer has its own formatter. Here are the differences between master and this branch:

```
[str_val_decimals] with 10 loops, best of 3:
	min 797 usec per loop
	avg 824 usec per loop
[str_val_g] with 10 loops, best of 3:
	min 681 usec per loop
	avg 728 usec per loop
[print_dense] with 10 loops, best of 3:
	min 1.98 msec per loop
	avg 2.02 msec per loop
[print_dense_decimals] with 10 loops, best of 3:
	min 2.1 msec per loop
	avg 2.16 msec per loop


[str_val_decimals] with 10 loops, best of 3:
	min 2.51 msec per loop
	avg 2.7 msec per loop
[str_val_g] with 10 loops, best of 3:
	min 1.01 msec per loop
	avg 1.04 msec per loop
[print_dense] with 10 loops, best of 3:
	min 2.45 msec per loop
	avg 2.47 msec per loop
[print_dense_decimals] with 10 loops, best of 3:
	min 11.7 msec per loop
	avg 11.9 msec per loop
```

janezd:
Excellent, thanks!

And even these differences are not too large.

PR #4590, opened 2020-03-28
[FIX] Use new access token in cleanup workflow

JakaKokosar:
##### Issue
The cleanup workflow in Github actions sometimes fail.

##### Description of changes
Use personal access token of our biolab-helper Github account. Hopefully, this will work more consistent now.


JakaKokosar:
@janezd, I think this is safe to merge now. I hope this fixes the issues with cleanup workflow.

PR #4591, opened 2020-03-30
[FIX] Nomogram: Fix crash on Python 3.8

lanzagar:
##### Issue

On py3.8 nomogram crashed for logistic regression models with:
`TypeError: only integer scalar arrays can be converted to a scalar index`

##### Description of changes
Iterate over actual coefficient values (numbers) instead of rows of a ndarray (which are still ndarrays) to get a proper boolean value for `reverse`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


PrimozGodec:
Thank you. It works fine now.

PR #4595, opened 2020-03-30
[ENH] SelectByDataIndex: data info displayed in the status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectbydataindex.py
+++ b/Orange/widgets/data/owselectbydataindex.py
@@ -1,10 +1,14 @@
 import numpy as np
 
+from AnyQt.QtCore import Qt
+
 from Orange.data import Table
 from Orange.widgets import widget, gui
 from Orange.widgets.utils import itemmodels
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.utils.widgetpreview import WidgetPreview
+from Orange.widgets.utils.state_summary import format_summary_details, \
+    format_multiple_summaries
 from Orange.widgets.widget import Input, Output
 from Orange.widgets.utils.annotated_data import (create_annotated_table)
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owselectbydataindex.py
```diff
@@ -62,6 +69,18 @@ def set_data_subset(self, data):
...
 
     def handleNewSignals(self):
+        if self.data and self.data_subset:
+            data_list = [('Data', self.data), ('Data subset', self.data_subset)]
+            summary = f"{len(self.data)}, {len(self.data_subset)}"
+            details = format_multiple_summaries(data_list)
+            self.info.set_input_summary(summary, details, format=Qt.RichText)
+        elif not self.data and not self.data_subset:
+            self.info.set_input_summary(self.info.NoInput)
+        else:
```
VesnaT, line 79: The two inputs are considered equivalent, so I would omit this `else` and replace the upper condition `self.data and self.data_subset` with `self.data or self.data_subset`.

In case when only one dataset is on the input, just show 0 in the summary.
Something like:
```
        summary, detail, kwargs = self.info.NoInput, "", {}
        if self.data or self.data_subset:
            n_data = len(self.data) if self.data else 0
            n_data_subset = len(self.data_subset) if self.data_subset else 0
            summary = f"{self.info.format_number(n_data)}, " \
                      f"{self.info.format_number(n_data_subset)}"
            kwargs = {"format": Qt.RichText}
            detail = format_multiple_summaries([
                ("Data", self.data),
                ("Data subset", self.data_subset)
            ])
        self.info.set_input_summary(summary, detail, **kwargs)
```

Orange/widgets/data/owselectbydataindex.py
```diff
@@ -89,6 +108,10 @@ def commit(self):
             matching_output = self.data[row_sel]
             non_matching_output = self.data[~row_sel]
             annotated_output = create_annotated_table(self.data, row_sel)
+
+        summary = len(matching_output) if matching_output else self.info.NoOutput
```
VesnaT, line 112: This is not true. 
`matching_output` evaluates to False when `len(matching_output) == 0`, causing `summary = self.info.NoOutput`, even though dataset with 0 instances was send to the output.

PR #4601, opened 2020-03-31
[FIX] Edit Domain: Improve Text/Categorical to Time conversion

robertcv:
##### Issue
New TimeVariable is initialized considering its parsing of string but actual data is converted by pandas.
This results in two issues:
 * Edit Domain can convert more formats (like m/d/y) but the TimeVariable is not correctly initialized (have_date and have_time)
 * pandas conversion is applied only when converting from Text but not when converting from Categorical.

##### Description of changes
Move pandas transformation into a separate function and apply it accordingly.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/oweditdomain.py
+++ b/Orange/widgets/data/oweditdomain.py
@@ -2527,8 +2527,16 @@ def make_dict_mapper(
 def time_parse(values: Sequence[str], name="__"):
     tvar = Orange.data.TimeVariable(name)
     parse_time = ftry(tvar.parse, ValueError, np.nan)
-    values = [parse_time(v) for v in values]
-    return tvar, values
+    _values = [parse_time(v) for v in values]
+    if np.all(np.isnan(_values)):
+        # try parsing it with pandas (like in transform)
+        dti = pd.to_datetime(values, errors="coerce")
+        _values = datetime_to_epoch(dti)
+        date_only = getattr(dti, "_is_dates_only", False)
+        if np.all(dti != pd.NaT):
+            tvar.have_date = True
+            tvar.have_time = not date_only
+    return tvar, _values
 
 
 as_string = np.frompyfunc(str, 1, 1)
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/tests/test_oweditdomain.py
```diff
@@ -917,6 +918,19 @@ def test_column_str_repr(self):
         d = column_str_repr(v, np.array([0., np.nan, 1.0]))
         assert_array_equal(d, ["00:00:00", "?", "00:00:01"])
 
+    def test_time_parse(self):
+        """parsing additional datetimes by pandas"""
+        date = ["1/22/20", "1/23/20", "1/24/20"]
```
ajdapretnar: How about when dates are 1/1/20? How do you know, which is day and which is month? M/D/Y is not universal.

robertcv: if data is not in iso8601 format then month first is assumed. Trying to parse dates without specifying the format is always tricky. Perhaps an additional field to specify d-m-y order would be the solution. However, that would probably be in a new pr. I would really like to merge this since I need it to pares such dates.

Orange/widgets/data/oweditdomain.py
```diff
@@ -2527,8 +2527,16 @@ def mapper(arr, out=None, dtype=dtype, **kwargs):
 def time_parse(values: Sequence[str], name="__"):
     tvar = Orange.data.TimeVariable(name)
     parse_time = ftry(tvar.parse, ValueError, np.nan)
-    values = [parse_time(v) for v in values]
-    return tvar, values
+    _values = [parse_time(v) for v in values]
+    if np.all(np.isnan(_values)):
+        # try parsing it with pandas (like in transform)
+        dti = pd.to_datetime(values, errors="coerce")
+        _values = str_to_time(np.array(values))
```
lanzagar, line 2534: You extracted some code into a new function `str_to_time` to reuse it here, but then have to call `pd.to_datetime` outside of the new function anyway, to get some extra info...

PR #4602, opened 2020-03-31
[FIX] Classification models output correct shapes

PrimozGodec:
##### Description of changes
- Support for all types of a sparse matrix in models `__call__`
- Make `data_to_model_domain` public
- Small clenup

This PR now includes essential changes that are required for the explain models widget and some cleanup. Other changes will be presented in separate PR.

@VesnaT and @lanzagar, your comments will be taken into account in the separate PR.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -320,6 +320,39 @@ class Model(Reprable):
         new_probs = new_probs / tots[:, None]
         return new_probs
 
+    def data_to_model_domain(self, data: Table) -> Table:
+        """
+        Transforms data to the model domain if possible.
+
+        Parameters
+        ----------
+        data
+            Data to be transformed to the model domain
+
+        Returns
+        -------
+        Transformed data table
+
+        Raises
+        ------
+        DomainTransformationError
+            Error indicates that transformation is not possible since domains
+            are not compatible
+        """
+        if data.domain == self.domain:
+            return data
+
+        if self.original_domain.attributes != data.domain.attributes \
+                and data.X.size \
+                and not all_nan(data.X):
+            new_data = data.transform(self.original_domain)
+            if all_nan(new_data.X):
+                raise DomainTransformationError(
+                    "domain transformation produced no defined values")
+            return new_data.transform(self.domain)
+
+        return data.transform(self.domain)
+
     def __call__(self, data, ret=Value):
         multitarget = len(self.domain.class_vars) > 1
 
@@ -335,21 +368,6 @@ class Model(Reprable):
 
         def fix_dim(x):
             return x[0] if one_d else x
-
-        def data_to_model_domain():
-            if data.domain == self.domain:
-                return data
-
-            if self.original_domain.attributes != data.domain.attributes \
-                    and data.X.size \
-                    and not all_nan(data.X):
-                new_data = data.transform(self.original_domain)
-                if all_nan(new_data.X):
-                    raise DomainTransformationError(
-                        "domain transformation produced no defined values")
-                return new_data.transform(self.domain)
-
-            return data.transform(self.domain)
 
         if not 0 <= ret <= 2:
             raise ValueError("invalid value of argument 'ret'")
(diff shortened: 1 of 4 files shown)


Orange/classification/svm.py
```diff
@@ -36,20 +44,18 @@ def __init__(self, penalty='l2', loss='squared_hinge', dual=True,
...
-
-class NuSVMClassifier(SklModel):
-
-    def predict(self, X):
-        value = self.skl_model.predict(X)
-        if self.skl_model.probability:
-            prob = self.skl_model.predict_proba(X)
-            return value, prob
-        return value
+    def __call__(self, data, progress_callback=None):
```
VesnaT, line 47: The code is duplicated. I'd recommend creating a SVMBaseLearner

Orange/classification/tests/test_base.py
```diff
@@ -186,6 +186,17 @@ def test_no_common_values(self):
             self.assertTrue(np.all(val <= 2))
             np.testing.assert_array_equal(prob, 1 / 3)
 
+    def test_sparse_matrix(self):
+        iris_sparse = self.iris.to_sparse()
+        for lrn in [TreeLearner, LogisticRegressionLearner]:  # skl and non-skl
```
lanzagar, line 191: Change the order so that the comment matches the learners.

Orange/base.py
```diff
@@ -320,12 +320,55 @@ def backmap_probs(self, probs, n_values, backmappers):
...
     def __call__(self, data, ret=Value):
         multitarget = len(self.domain.class_vars) > 1
 
         def one_hot_probs(value):
             if not multitarget:
-                return one_hot(value)
+                probs = one_hot(value)
+                # sometimes a maximum class value(s) are not predicted and
+                # those probs.shape[1] does not match to the length of values
+                # in the domain - here we add missing dimension if required
```
lanzagar, line 364: This comment is very hard to understand even ignoring the grammar issues (singular/plural)...

Orange/tests/test_classification.py
```diff
@@ -190,6 +193,76 @@ def test_incompatible_domain(self):
         with self.assertRaises(DomainTransformationError):
             clf(titanic)
 
+    LEARNERS_CLASSIFICATION = [
```
VesnaT, line 196: There is a function `all_learners()` in `LearnerAccessibility` class. Could that be used for testing instead of hard coding the `learners` list. If we add a new learner, we will most likely forget to add it to the list.

Orange/base.py
```diff
@@ -320,12 +320,55 @@ def backmap_probs(self, probs, n_values, backmappers):
...
+                num_val = (
+                    len(self.original_domain.class_var.values)
+                    if self.original_domain is not None else probs.shape[0]
+                )
+                return np.hstack((
+                    probs, np.zeros((probs.shape[0], num_val - probs.shape[1]))
+                ))
```
lanzagar, line 371: The only other use of `one_hot` I found, had an almost identical (but not the same) fix as here.
It would probably be better to improve `one_hot` itself instead of calling it and adding a fix after that every time.
E.g. an additional optional argument like `max_value` which is used if provided (+ maybe a check that if `max_value` is given it must be `>=` the max found in values).

Orange/classification/svm.py
```diff
@@ -24,6 +24,14 @@ def __init__(self, C=1.0, kernel='rbf', degree=3, gamma="auto",
         super().__init__(preprocessors=preprocessors)
         self.params = vars()
 
+    def __call__(self, data, progress_callback=None):
+        if len(np.unique(data.Y)) > 1:
+            return super().__call__(data, progress_callback)
+        else:
+            warnings.warn("Single class in data, returning Constant Model.")
+            maj = MajorityLearner()
+            return maj(data)
```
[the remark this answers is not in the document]
lanzagar: We have to be careful if we move this check to the `Learner` base class, because we probably expect a specific model as a result in some cases. E.g. logistic regression models can be connected to nomograms and might crash if all of a sudden they happen to be constant models (without coefficients etc).
But I still think this is not an svm-specific issue.

PR #4604, opened 2020-03-31
Added PLS (Scikit Learn) Implementation

realrk69:
##### Description of changes
PLS Algorithm (NIPALS). I can add other cross decomposition methods, but those are mostly unused in today's time. LDA and PCR will be implemented in the coming time. This is minimal reproducible. After your suggestions, I'll make all the changes needed.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


realrk69:
@markotoplak  Hey bro, did you try the widget? Any suggestions?

markotoplak:
@realrk69, thank you for this PR. The code is short and to the point.

I was thinking about it for a while, but I am still not sure what is the best way of putting PLS into Orange. I see two ways of using PLS:

1) As a regression model.
2) As a data transformation utility, in a way similar to PCA.

And honestly, I do not know if these both belong to the same widget. You have implemented (1), which already needs 3 widget outputs. To do (2), the widget would at least need to output rotations and scores for both X and Y: at least 2 or 4 additional outputs.

To see what works best, different users should try to use the widget. Because the PLS widget will likely change substantially in the future, I ask you to make a pull request into https://github.com/biolab/orange3-prototypes . Two reasons: in the main repository we aim for backward compatibility, which is only possible if inputs/outputs do not change meaning, and, it allows for shorter change-release-test cycles.

This implementation seems correct and already does standarnization (in scikit-learn). For now, I suggest staying with the scope of this PR: PLS as a regression model - (1). The only things preventing it to be immediately accepted in orange3-prototypes are the missing coefficients output, some tests (both for the "library", PLS learner/model classes, and for the widget - see Logistic regression widget coefficients tests) and some slight code reformatting so that  it passes lint (tip: some editors, such as PyCharm, already show the warnings).

Minor:
- please replace "Number of Components" with just "Components"; this will make the widget smaller

I am closing this PR hoping that you will reopen a similar PR, with the mentioned issues resolved, in orange3-prototypes. Meanwhile, we can use this PR for discussion. I also invite @stuart-cls and @scott-cls into this discussion.

realrk69:
@markotoplak Thanks for the reply. Here's what I know about PLS honestly:

1. Regression works as is. The problem is outputing coefficients and will be fixed before the weekend.
2. I think you're refering to LDA (sklearn.discriminant_analysis.LinearDiscriminantAnalysis). You can refer to it [here](https://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.LinearDiscriminantAnalysis.html). And I agree 100% it needs a separate widget. It has an implementation of a "lsqr" (least squares solution) which does what you've mentioned. I have refered to this in the Issues and will be making it soon as a separate commit.

I wanted to ask if aside from this, other methods of PLS Regression should be implemented (canonical, svd etc.). They're not used much tbh.

I use Visual Studio Code as an editor, will try developing in PyCharm and creating a new PR in orange-protoypes after issues have been resolved

janezd:
You don't have to use PyCharm if you don't want to (although I use and recommend it, too). There's a lint plugin for Visual Studio Code, too, and I suppose it's the same as the one in PyCharm.

realrk69:
I saw [this](https://github.com/biolab/orange3-prototypes/blob/master/orangecontrib/prototypes/widgets/owlda.py). Isn't this the normal LDA? Its 2 years old and does not have various implementations of solvers. So should I make an LDA widget or is this still under trial by the community?

markotoplak:
> I think you're refering to LDA (sklearn.discriminant_analysis.LinearDiscriminantAnalysis). You can refer to it here.

No, not LDA, I was actually referring to PLS. So, as made possible by this PR, PLS can be used as a regression. But with (2) I meant using intermediate results to gain insight into your data. For example, someone could try to interpret what do individual components mean. Or they could try to transform new data into component scores with exact same meaning as it was done on the train data.

markotoplak:
> I wanted to ask if aside from this, other methods of PLS Regression should be implemented (canonical, svd etc.).

For not, nipals is enough. Let us only add others if a particular need arises.

markotoplak:
> Isn't this the normal LDA? Its 2 years old and does not have various implementations of solvers. So should I make an LDA widget or is this still under trial by the community?

LDA is only better than logistic regression very rarely, and most of times is worse (it uses some assumptions about the data that are usually not met). So, LDA and LR build similarly structured models, and LR is more flexible of the two. Therefore we do not think that we should crowd Orange with LDA.

stuart-cls:
@markotoplak I don't have any specifics about PLS to add but I support your idea of starting in -prototypes.

scott-cls:
@markotoplak - I agree with you that the development of this could be pushed to prototypes, I am just worried it will be lost there.  

I know there is a growing interest in the spectroscopic community to see this type of regression technique implemented - it would help with the visibility of Quasar (Orange).

realrk69:
@scott-cls This is one of the reasons why I've been trying to get this to open source communities rather than building one software of my own (although I did make one in PyQt5, but got to know about Orange and its already well established features). Our company's name is RAAV Techlabs and we assemble spectrometers. Our partners include Texas Instruments, Hamamatsu, HORIBA and Ocean Optics. All spectral calibration making companies charge horrendous amounts of money for something that newbies cannot easily code themselves. It also stops research from evolving in poorer regions of the world.

That is why I initially wanted to commit this to Orange-Spectroscopy, since PLS shows most usage there. And also export the models for prediction on cross platform apps. If you guys think that is better and can be followed up readily (compared to orange-prototypes), then I'll make a pull request there.

markotoplak:
@realrk69, you are right. Given that most users that asked for it are using spectroscopy, perhaps it would be better to make a PR there. I think that it will get more use/testing there.

borondics:
@realrk69, thanks a lot for working on this! Many people have asked already, but we don't have enough time to satisfy every request immediately. I also think this could nicely fit into Orange-Spectroscopy. @stuart-cls referred to our prototypes repository I think (https://github.com/Quasars/orange-spectroscopy-prototypes) but maybe it can already go into the main Orange-Spectroscopy toolbox. @markotoplak, what do you think?

@realrk69, since you are interested in spectroscopy as well, we would be happy to have you as a developer of the package. As a small return for your contributions, I can offer to have your company's logo on the Quasar (a pre-packaged Orange with spectroscopy tools for easy installation) website (https://quasar.codes). Have a look and let me know if you are interested in more coding!

stuart-cls:
@borondics I was actually just referring to the orange-prototypes that Marko had mentioned. I think @markotoplak meant just go straight into orange-spectroscopy. That should be a good spot.

realrk69:
@borondics I would be very glad to join the community. I'm working on PCR, LDA and fixing PLS coef output (various versions with insightful outputs). I'll be creating a pull request at orange-spectroscopy within a few days with the first commit. Thanks for your offer, that would be great.

realrk69:
Hey guys, I'm almost done, PR will be created tomorrow. Just wanted to know if PLS transformations as @markotoplak had mentioned is important from spectroscopy point of view. Should I also create a separate PR for that? Where should that PR go (orange core or orange spectroscopy)? Honestly, I have not found any real world usage scenarios for transforming the data (for spectral). So, I have made it into a separate widget (PLS Transform). 

Thanks all for your support, hope you're keeping healthy. Looking forward to coding with you. 

@stuart-cls @borondics @scott-cls @janezd

PR #4614, opened 2020-04-02
Feature Constructor: Allow decreasing the height

janezd:
##### Issue

Feature Constructor's height should be allowed to be decreased. It would make for nicer screenshots.

##### Description of changes

The problem was the height of the list view. Minimum height is now set to 50.

##### Includes
- [X] Code changes


ajdapretnar:
This is all great but fails on AppVeyor. If all in favour to ignore it, I can merge.

Issue #4553, opened 2020-03-20
Test segfaults


markotoplak:
**Describe the bug**

Test sometimes segfault on Travis and github actions.

I am abusing this PR for a collection of segfault reports so that we (maybe) see what is happening.

markotoplak:
OSX 10.14, Orange master. `tox -e py` For the first three times everything was fine. On the fourth time I got this:

```
test_minimum_size (Orange.widgets.evaluate.tests.test_owtestlearners.TestOWTestLearners) ... ok
test_more_learners (Orange.widgets.evaluate.tests.test_owtestlearners.TestOWTestLearners) ... ok
test_msg_base_class (Orange.widgets.evaluate.tests.test_owtestlearners.TestOWTestLearners) ... ok
test_nan_class (Orange.widgets.evaluate.tests.test_owtestlearners.TestOWTestLearners) ... ok
test_nan_on_comparison (Orange.widgets.evaluate.tests.test_owtestlearners.TestOWTestLearners) ... ERROR: InvocationError for command /Users/marko/dev/orange3/.tox/py/bin/python -m unittest --verbose Orange.tests Orange.widgets.tests (exited with code -11 (SIGSEGV)) (exited with code -11)
```

If I remember correctly, travis tests also sometimes segfault for some Test Learner test, but not always for the same one.

janezd:
MacOS 10.15, Python 3.7, Github actions, #4558.

```
test_comparison_bad_slots (Orange.widgets.evaluate.tests.test_owtestlearners.TestOWTestLearners) ... ERROR: InvocationError for command /Users/runner/runners/2.165.2/work/orange3/orange3/.tox/py/bin/python -m unittest --verbose Orange.tests Orange.widgets.tests (exited with code -11 (SIGSEGV)) (exited with code -11)
```

markotoplak:
OSX 10.15, Orange master.

```
test_comparison_requires_cv (Orange.widgets.evaluate.tests.test_owtestlearners.TestOWTestLearners) ... ok
test_comparison_requires_multiple_models (Orange.widgets.evaluate.tests.test_owtestlearners.TestOWTestLearners) ... Unable to set '<orangewidget.utils.combobox.ComboBox object at 0x144fbd050>' to 1; largest index is -1
ERROR: InvocationError for command /Users/marko/dev/orange3/.tox/py/bin/python -m unittest --verbose Orange.widgets.tests (exited with code -11 (SIGSEGV)) (exited with code -11)
```

[...]

PR #4615, opened 2020-04-03
[FIX] owtestandscore: Add cancelled tasks to dispose queue

ales-erjavec:
##### Issue

Annother stab at gh-4553

##### Description of changes

* Add cancelled tasks to dispose queue until final finished signal.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
Seem to work - I can not crash it anymore. Does this mean we need to add something like this to other widgets that use threads?

ales-erjavec:
I don't know.

PR #4616, opened 2020-04-03
Select Columns: Store settings as Variable

### LINKED ISSUE #4613: Select Columns shouldn't store Variables as string settings
OWSelectColumns stores `domain_role_hints` as strings. Support for storing variables as strings will presumably be dropped in Orange 3.26.

```
/Users/janez/orange3/Orange/widgets/data/owselectcolumns.py:310: UserWarning: Storing variables as strings in settings is deprecated.
Support for this will be dropped in Orange 3.26.
Change domain_role_hints to store an instance of `Variable`.
  self.openContext(data)
```

Don't forget to migrate settings.

ajdapretnar:
Fixes #4613.

There is no migration because settings are saved exactly as they were before. The difference is that before this PR, variables were encoded as (name, type) pair by the widget, hence the settings handler complained about receiving variables as strings. Now, encoding is done by context handler, where it belongs, and everybody should be happy.

One test is removed because it is no longer applicable to the widget -- there are no `required` attributes in the context.

##### Includes
- [X] Code changes
- [X] Tests


ajdapretnar:
I have no idea why tests fail. I tested this in GUI, recreated all the steps of connecting and disconnecting data and features, and the widget looks the same as on master. 🤷‍♀

janezd:
Encoding of variables should happen in the context handler, not in the widget. This widget already has its specific handler, `SelectAttributesDomainContextHandler`. A part of the reason why encoding belongs there is also that matching must be aware of how it's done. As it is now, I'm not sure that matching works properly, which might be related to the failed test (I haven't examined it closely enough).

See `SelectRowsContextHandler` for an example. This handler is also interesting because it avoids the problem of migrating variables stored as strings by not migrating and instead accepting old-style settings (until a deadline that will be enforced by tests).

I would also suggest moving `SelectAttributesDomainContextHandler` to `owselectcolumns.py`. It doesn't belong to the file with general handlers.

ajdapretnar:
I have moved the handler into the widget. However, I still don't understand how these settings actually work. I'd love to learn, but if someone can solve this much quicker than I can, it's probably better.

janezd:
@ajdapretnar, this is also one of those that need some clicking before we're sure they work as intended. As a reward, I can explain how it works, so next time you can do the settings and I'll do the clicking. :)

janezd:
#4616 and #4643 have been merged into #4643 because neither can live while the other dies.

PR #4626, opened 2020-04-06
[ENH] BoxPlot: data info displayed in the status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/visualize/owboxplot.py
```diff
@@ -329,8 +333,11 @@ def set_data(self, dataset):
             self.openContext(self.dataset)
             self.grouping_changed()
             self.attr_changed()
+            self.info.set_input_summary(len(dataset),
```
VesnaT, line 336: If one inputs dataset that has only string variables, no info is shown.
Sure, it is useless to input that kind of data, but at least the user would realise why the plot is empty, after examining the detail.

PR #4629, opened 2020-04-07
[FIX] Fix dimensionality of probabilities from values

PrimozGodec:
##### Issue

SGD (same than some other models) outputs wrong shapes of probabilities when those retrieved from values (when skl_learner do not provide probabilities). The issue is caused by the one_hot encoding function that does not consider a number of values in the domain.

SGD (skl_learner) for some of the losses provides probabilities that are not used by Orange since it uses custom predict function.

##### Description of changes

- ` one_hot` function changed to accept `dim` parameter which tells what must be a resulting number of columns
- `__call__` function for model calls `one_hot` with a `dim` parameter that has a number of values in the domain as a shape.
- SGD model now uses a predict function from SklModel (which correctly retrieve probabilities from skl_learner) if possible.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/util.py
+++ b/Orange/data/util.py
@@ -4,7 +4,7 @@ Data-manipulation utilities.
 import re
 from collections import Counter
 from itertools import chain, count
-from typing import Callable
+from typing import Callable, Union, List, Type
 
 import numpy as np
 import bottleneck as bn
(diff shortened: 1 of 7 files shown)


Orange/data/util.py
```diff
@@ -13,22 +13,36 @@
...
     if len(values) == 0:
-        return np.zeros((0, 0), dtype=dtype)
-    return np.eye(int(np.max(values) + 1), dtype=dtype)[np.asanyarray(values, dtype=int)]
+        return np.zeros((0, dim or 0), dtype=dtype)
+    res = np.eye(
+        int(np.max(values) + 1), dtype=dtype)[np.asanyarray(values, dtype=int)]
+    if dim is not None:
+        assert dim >= res.shape[1], "dim demands less columns than needed for values"
+        res = np.hstack((res, np.zeros((res.shape[0], dim - res.shape[1]))))
+    return res
```
lanzagar, line 45: I find the following a bit easier to read and it avoids dim arithmetics and hstacking:

```py
    dim_values = int(np.max(values) + 1)
    if dim is None:
        dim = dim_values
    elif dim < dim_values:
        raise ValueError("dim must be greater than max(values)")
    return np.eye(dim, dtype=dtype)[np.asanyarray(values, dtype=int)]
```

PR #4638, opened 2020-04-09
Help url: use docs.biolab.si instead of docs.orange.biolab.si

markotoplak:
##### Issue
The old URL, docs.orange.biolab.si, has certificates managed by us, and sometimes something goes wrong and it becomes inaccessible. docs.biolab.si is managed by cloudflare.


janezd:
Include Orange in URL for SEO.

acopar:
This URL now works in the same way: [https://docs.biolab.si/orange/3/visual-programming/](https://docs.biolab.si/orange/3/visual-programming/).

PR #4642, opened 2020-04-10
DataTable: data info displayed in the status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -39,6 +39,7 @@ from Orange.widgets.utils import datacaching
 from Orange.widgets.utils.annotated_data import (create_annotated_table,
                                                  ANNOTATED_DATA_SIGNAL_NAME)
 from Orange.widgets.utils.itemmodels import TableModel
+from Orange.widgets.utils.state_summary import format_summary_details
 
 
 class RichTableModel(TableModel):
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owtable.py
```diff
@@ -884,6 +876,10 @@ def commit(self):
             # Selections of individual instances are not implemented
             # for SqlTables
             if isinstance(table, SqlTable):
+                summary = selected_data.approx_len() if selected_data else \
```
VesnaT, line 879: `selected_data` is set to None at the beginning of the function and never set to anything else.

Orange/widgets/data/owtable.py
```diff
@@ -1028,70 +1027,6 @@ def parts(array, density, col_dist):
         return Summary(n_instances, domain, X_part, Y_part, M_part)
 
 
-def format_summary(summary):
```
VesnaT, line 1031: Try reusing this function (or rewrite it for your needs) to get more accurate results on lengths.

Orange/widgets/data/owtable.py
```diff
@@ -491,6 +485,7 @@ def sizeHint():
         return QSize(800, 500)
 
     @Inputs.data
+    @check_sql_input
```
VesnaT, line 488: I think this decorator was left out intentionally. 
The widget handles big sql tables differently then other widget.

Orange/widgets/data/owtable.py
aturanjanin: What do you think about reusing this function but within the `format_summary_details` function? That function takes the table as the input and it calculates all the lengths itself. Also, in that way we get consistent results for all widgets.

VesnaT: I only meant it for the "length" part of the function. It returns `approx_len` for SqlTable, unless the length has finished calculating. I'm not sure whether we can spot the difference (probably for big sql tables), but I guess we should keep the behavior.

VesnaT:
Can you pleas rebase to master?

VesnaT:
Can you pleas rebase to master?

aturanjanin:
@VesnaT Rebased now, thanks.

PR #4656, opened 2020-04-15
[FIX] Pythagorantrees/forests: change context handler

PrimozGodec:
##### Issue

Pythagorantrees and Pythagorantrees use domain context handler without having any context variable with variables from the domain

##### Description of changes
DomainContextHandler replaced with ClassValuesContextHandler
Besides: Fixed warnings about setting index (for the selected item) when an empty combo box

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
@PrimozGodec, could you rebase this to master, so we can see the tests?

PR #4673, opened 2020-04-19
OWRank: Store selection as variables

### LINKED ISSUE #4672: Rank widget uses DomainContextHandler, but stores indices
Rank widget stores selection as row indices instead of variables, so `DomainContextHandler` will match any data.

janezd:
##### Issue

Fixes #4672.

Existing contexts are not migrated but removed. Old contexts store row indices instead of variables. These could be migrated if we assume that dictionary `context.attributes` is ordered, which is not necessarily true. They could also be migrated in `set_data`.

I however decided to remove them because they would match any domain and thus never really worked.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -1,10 +1,3 @@
-"""
-Rank
-====
-
-Rank (score) features for prediction.
-
-"""
 import warnings
 from collections import namedtuple, OrderedDict
 import logging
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owrank.py
```diff
@@ -198,9 +192,9 @@ class Outputs:
     sorting = Setting((0, Qt.DescendingOrder))
     selected_methods = Setting(set())
 
-    settings_version = 2
+    settings_version = 3
     settingsHandler = DomainContextHandler()
-    selected_rows = ContextSetting([])
+    selected_attrs = ContextSetting([], schema_only=True)
```
VesnaT, line 197: Can `ContextSetting` can be `schema_only`?
I think it is always read from a context and ignores the schema_only flag.

At this point we should decide whether we want the selection as schema_only setting or context dependent.

janezd: I would like it to be a schema-only context setting. :) If the flag is currently ignored, I'd still keep it for the case it is supported some day. But if you're sure it's ignored and you want to remove it, I can do so, too.

PR #4679, opened 2020-04-20
DomainContextHandler: support dictionaries as context settings

### LINKED ISSUE #4664: Let DomainContextHandler handle dict with Variables as keys
**Describe the bug**

Discretize widget uses `DomainContextHandler`, but the context handler does not understand the widget's context setting (a dictionary with attributes as keys), and hence matches all domains.

**To Reproduce**

1. Feed Iris to Discretize, set one variable's method to whatever.
2. Feed heart disease to discretize. (At this point, the widget will re-use the context from Iris!)
3. Feed Iris to Discretize. The setting from the first step was lost.

**Expected behavior**

~~Discretize widget should have a context handler whose~~
- ~~match checks the dictionary~~
- ~~encode and decode handles the dictionary.~~

~~Alternatively,~~ `DomainContextHandler` could be improved to handle dicts whose keys are variables. Discretize is not the only widget that needs it: Select Columns implements its own `SelectAttributesDomainContextHandler` for the same reason.

janezd:
##### Issue

Resolves #4664.

##### Description of changes

- DomainContextHandler now supports dictionaries whose keys are values (as used by Impute and Discretize) as context settings
- Empty matches (that is, domain contexts with no attributes to match) were promoted if they appeared first. Now they match, but with low priority.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -127,6 +127,10 @@ class DomainContextHandler(ContextHandler):
             new_value = [item for item in value
                          if self.is_valid_item(setting, item, attrs, metas)]
             data[setting.name] = new_value
+        elif isinstance(value, dict):
+            new_value = {item: val for item, val in value.items()
+                         if self.is_valid_item(setting, item, attrs, metas)}
+            data[setting.name] = new_value
         elif value is not None:
             if (value[1] >= 0 and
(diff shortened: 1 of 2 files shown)


markotoplak:
Also thanks for the empty match fix!

Issue #4684, opened 2020-04-21
Heatmap row splitting (missing values?)


Hrovatin:
Sometimes when I want to use row splitting in Heatmap I get the below error. If I remove rows with missing values in the feature used in splitting the splitting works.
Last master on linux.
```Traceback (most recent call last):
  File "/home/karin/Documents/git/orange3venv/lib/python3.6/site-packages/orangecanvas/scheme/signalmanager.py", line 936, in __process_next
    if self.__process_next_helper(use_max_active=True):
  File "/home/karin/Documents/git/orange3venv/lib/python3.6/site-packages/orangecanvas/scheme/signalmanager.py", line 974, in __process_next_helper
    self.process_node(selected_node)
  File "/home/karin/Documents/git/orange3venv/lib/python3.6/site-packages/orangecanvas/scheme/signalmanager.py", line 605, in process_node
    self.send_to_node(node, signals_in)
  File "/home/karin/Documents/git/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 792, in send_to_node
    self.process_signals_for_widget(node, widget, signals)
  File "/home/karin/Documents/git/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 826, in process_signals_for_widget
    handler(*args)
  File "/home/karin/Documents/git/orange3/Orange/widgets/visualize/owheatmap.py", line 562, in set_dataset
    self.update_heatmaps()
  File "/home/karin/Documents/git/orange3/Orange/widgets/visualize/owheatmap.py", line 599, in update_heatmaps
    self.construct_heatmaps_scene(parts, self.effective_data)
  File "/home/karin/Documents/git/orange3/Orange/widgets/visualize/owheatmap.py", line 783, in construct_heatmaps_scene
    self.setup_scene(parts, data)
  File "/home/karin/Documents/git/orange3/Orange/widgets/visualize/owheatmap.py", line 806, in setup_scene
    widget.setHeatmaps(parts)
  File "/home/karin/Documents/git/orange3/Orange/widgets/visualize/utils/heatmap.py", line 383, in setHeatmaps
    assert data.shape == (nrows, ncols)
AssertionError

PR #4686, opened 2020-04-22
[FIX] Heatmap: Split by missing values

ales-erjavec:
##### Issue

Fixes gh-4684

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Hrovatin:
This does not crash anymore. However, it is not what I expected from splitting with missing values: The missing values should be grouped in a separate N/A group (as in colour annotation legend). The behavior that was implemented by this fix can be already achieved by using Select Rows -> is defined.

Hrovatin:
Now it works as expected.

ajdapretnar:
This works well for me.

PR #4695, opened 2020-04-24
Revert "TreeViewer: data info displayed in the status bar"

VesnaT:
Reverts biolab/orange3#4682


VesnaT:
Data info should not be displayed for models.

PR #4696, opened 2020-04-24
Revert "PythagoreanTree: data info displayed in the status bar"

VesnaT:
Reverts biolab/orange3#4685


VesnaT:
Data info should not be displayed for models.

PR #4711, opened 2020-04-30
Remove files for building installers

markotoplak:
Installers were moved to orange3-installers. 3.25.0 was already released from installers there. These files are now stale.


vinbrule:
Hello, can you point me to the location of  orange3-installers? I couldn't find it here. Bests

janezd:
Google can't help? And git's search also doesn't find it? :)

ales-erjavec:
> Hello, can you point me to the location of orange3-installers?

https://github.com/ales-erjavec/orange3-installers

vinbrule:
Thank so much

Issue #4608, opened 2020-04-01
Test run failures: fix localisation and rearrange gui's


robertcv:
**Expected behavior**
Local run tests should not fail.

**Versions:**
Orange master, PyQt5==5.13.0
Python 3.6.9,
Ubuntu 18.04.4 LTS, GNOME 3.28.2

**Additional context**
In the spirit of the recent quest for segfault, I run the whole test suite locally (with `tox -e py`) and found that there are some tests failing:

```
======================================================================
FAIL: test_binning (test_discretize.TestTimeBinning)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/robert/git/orange3/.tox/py/lib/python3.6/site-packages/Orange/preprocess/tests/test_discretize.py", line 62, in test_binning
    615254400])]
AssertionError: Lists differ: [('10[677 chars]['75 jan', 'jul', '76 jan', 'jul', '77 jan', '[538 chars]00])] != [('10[677 chars]['75 Jan', 'Jul', '76 Jan', 'Jul', '77 Jan', '[538 chars]00])]

First differing element 4:
('6 months', ['75 jan', 'jul', '76 jan', 'jul', '77 jan', '[537 chars]400])
('6 months', ['75 Jan', 'Jul', '76 Jan', 'Jul', '77 Jan', '[537 chars]400])

Diff is 2834 characters long. Set self.maxDiff to None to see it.

======================================================================
FAIL: test_minimum_size (Orange.widgets.data.tests.test_owcontinuize.TestOWContinuize)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/robert/git/orange3/.tox/py/lib/python3.6/site-packages/orangewidget/tests/base.py", line 181, in test_minimum_size
    self.check_minimum_size(widget)
  File "/home/robert/git/orange3/.tox/py/lib/python3.6/site-packages/orangewidget/tests/base.py", line 501, in check_minimum_size
    self.assertLess(min_size.height(), 700)
AssertionError: 724 not less than 700

======================================================================
FAIL: test_minimum_size (Orange.widgets.data.tests.test_owimpute.TestOWImpute)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/robert/git/orange3/.tox/py/lib/python3.6/site-packages/orangewidget/tests/base.py", line 181, in test_minimum_size
    self.check_minimum_size(widget)
[...]

PR #4717, opened 2020-04-30
Rearrange a few widgets' UX to decrease their heights

janezd:
##### Issue

Fixes (partially) #4608.

##### Description of changes

I changed the widgets roughly along the lines we set we discussing the issue, except for making ROC scrollable. If the user wants smaller graphs, (s)he can hide the control area, like in other widgets, IMHO.

### DIFF
--- a/Orange/widgets/data/owcontinuize.py
+++ b/Orange/widgets/data/owcontinuize.py
@@ -2,6 +2,7 @@ from functools import reduce
 from types import SimpleNamespace
 
 from AnyQt.QtCore import Qt
+from AnyQt.QtWidgets import QGridLayout
 
 import Orange.data
 from Orange.util import Reprable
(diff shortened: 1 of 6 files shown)


ajdapretnar:
- Pivot: both Auto Apply and top box have extra space that doesn't have to be there. Something like `box_label=False`?
- Statistics: I would prefer having the numbers aligned. 😇 Also, try hiding and showing controlArea. The widget behaves in a strange way. How about doing in Continuize style, with stats below each corresponding window? Do stats have to be mainArea?
- Impute: Apply automatically probably doesn't need this super extended box?

Otherwise, I have to say some solutions are quite elegant. I like Continuize. At first it looks strange, but then I started liking it more and more.

ajdapretnar:
Examples of changes for future reference:

**Pivot**
Screen Shot 2020-05-08 at 10 49 32

**ROC Analysis**
Screen Shot 2020-05-08 at 11 10 37

**Continuize**
Screen Shot 2020-05-08 at 11 10 13

**Purge Domain**
Screen Shot 2020-05-08 at 11 11 47

**Impute**
Screen Shot 2020-05-08 at 11 10 16

markotoplak:
I like the changes.

Another minor issue: Impute should also have margins fixed. Margins around the upper two boxes are much bigger than in any other widget. That was already there before the PR.

markotoplak:
Also, in Purge domain the top left and right boxes not valigned.

janezd:
Uh, this looks horrible, as if they are too large and space is inserted everywhere. See my screenshots below.

Impute's margins on my machine are perhaps even too small, and the top two boxes in Purge are aligned.

I suspected it was saved geometry, but I saved a schema with an old master and reloaded it, and they look OK.

I can't debug this. Could you try, please?

purge
pivot
impute

ajdapretnar:
I think it could be an issue with my settings. I am getting strange fonts and extra spaces everywhere for some reason. I noticed my old screenshots did not have this. 🤷

markotoplak:
I see the same things as @janezd. @lanzagar, can you check on Linux?

lanzagar:
I get the same extra spaces and misalignments as @ajdapretnar !

markotoplak:
@lanzagar, and what happens in master?

lanzagar:
Well Purge Domain for example does not have two columns on master so there is no alignment there, while here Features and Statistics boxes are not aligned for me like they are for @janezd (see @ajdapretnar's screenshot).
The empty space at the top of the new Pivot gui is also not there on master, because the box on master has a title, while here it is an untitled box with labels inside. But otherwise this is not this PR's fault - untitled boxes have empty space on top on master and here (not sure when this started happening, maybe it has to do with the qt version?).
The above also applies to auto commit boxes - the extra space is because they are in untitled boxes.

janezd:
So, is there something we can fix here? If so, can you try to fix it -- I can't because it works for me...

ajdapretnar:
I think we saw something similar if Timeseries four weeks ago. I cannot remember the exact fix, it was something like `box_label = False`. Perhaps @robertcv, @irgolic or @AndrejaKovacic remember what it was?

PR #4718, opened 2020-04-30
[ENH] Color widget: Add reset button

### LINKED ISSUE #4665: Color: No way to reset to default
**Describe the bug**
After making some changes in Color widget it should be possible to get back to default colors.
It would probably be enough to allow clearing all settings, but we might discuss undoing even individual changes?

**Orange version:**
3.25

janezd:
##### Issue

Fixes #4665.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcolor.py
+++ b/Orange/widgets/data/owcolor.py
@@ -35,6 +35,9 @@ class AttrDesc:
         self.var = var
         self.new_name = None
 
+    def reset(self):
+        self.new_name = None
+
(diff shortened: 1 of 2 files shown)


ajdapretnar:
This works well, just the button is fairly small compared to the widget.

janezd:
Indeed it is, but Reset button is supposed to be small. :)
(It's also easier to implement it this way.)

PR #4721, opened 2020-05-01
[WIP] [RFC] Setting type checking in Domain and ClassValues context handlers; annotations in projections

janezd:
##### Issue

Uses changes from https://github.com/biolab/orange-widget-base/pull/62.

Changes in projections are given as an example of kinds of annotations that will be required from widgets to avoid warnings and allow storing their settings as json.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/data/owcorrelations.py
```diff
@@ -431,10 +432,6 @@ def migrate_context(cls, context, version):
             sel = context.values["selection"]
             context.values["selection"] = [(var.name, vartype(var))
                                            for var in sel[0]]
-        if version < 3:
-            sel = context.values["selection"]
-            context.values["selection"] = ([(name, vtype + 100)
-                                            for name, vtype in sel], -3)
```
janezd, line 437: This migration is removed because annotated settings are no longer encoded into tuples. `DomainContextHandler` thus performs an exactly opposite migration for all annotated settings. If this migration here is left in, it would "unmigrate" the setting.

I assume such cases like this will be rare. So although this situation could be avoided if `DomainContextHandler` performed the migration after calling `migrate_context`, the downside would be that this would force the future `migrate_context` to work with raw, encoded settings instead of simpler, decoded, and thus force them to "understand" how coding is done - which should be of no concern to migrations.

PR #4738, opened 2020-05-05
[FIX] Test and Score: Improve data errors

### LINKED ISSUE #4734: Test & Score: better warning/error when no features on input
Test&Score shows the following error when no features are present on the input:
`ValueError: Found array with 0 feature(s) (shape=(40, 0)) while a minimum of 1 is required.`

The error is very uninformative for a lay user. Provide a better error, something like
"Data has no features." / "No features present on the input."

PrimozGodec:
##### Issue

- Fixes #4734
- Currently, we only check that class_var has more than 1 class value, but it can happen that this condition is valid but there is only one class present in Y. I think it is better not to handle this case and show an error. The same is done by any model widget in Orange. Another argument to prevent this option is since learners do not support only one class in data. I hope my thinking is ok and that I didn't overlook any special case.

##### Description of changes
- Added error for no features
- Showing error when only one class in data - even class_var.values have more.
- Errors cleanup

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/evaluate/owtestandscore.py
+++ b/Orange/widgets/evaluate/owtestandscore.py
@@ -32,6 +32,7 @@ import Orange.evaluation
 from Orange.evaluation import Results
 from Orange.preprocess.preprocess import Preprocess
 import Orange.regression
+from Orange.statistics.util import unique
 from Orange.widgets import gui, settings, widget
 from Orange.widgets.evaluate.utils import \
     usable_scorers, ScoreTable, learner_name, scorer_caller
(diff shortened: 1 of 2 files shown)


Orange/widgets/evaluate/owtestandscore.py
```diff
@@ -383,25 +381,29 @@ def set_train_data(self, data):
         self.cancel()
         self.Information.data_sampled.clear()
         self.Error.train_data_empty.clear()
-        self.Error.class_required.clear()
-        self.Error.too_many_classes.clear()
-        self.Error.no_class_values.clear()
-        self.Error.only_one_class_var_value.clear()
+        self.Error.data_error.clear()
```
lanzagar, line 384: Maybe you could rename `data_error` to `train_data_error` and use it for `train_data_empty` as well.

PR #4740, opened 2020-05-05
[ENH] Select Rows: Allow partial context matches

### LINKED ISSUE #4737: Select Rows - do not change filter after column change if it is not required
When performing filtering with Select Rows the filters are reset after changing the column on which the filtering is performed. This is annoying when one has multiple similar columns that require the same filters. Could there be an option not to reset the filter columns, except if necessary due to a variable type change?
Additionally, even if a filter is changed from colA to colB and back to colA the previous settings used for colA are not retained.

Version:
Orange 3.26.

janezd:
##### Description of changes

Allow partial matches (but request matching at least one attribute) and remove non-matching conditions.

In case of multiple matches, take the match with the most matching variables.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -80,12 +80,29 @@ class SelectRowsContextHandler(DomainContextHandler):
         conditions = context.values["conditions"]
         all_vars = attrs.copy()
         all_vars.update(metas)
-        # Use this after 2022/2/2:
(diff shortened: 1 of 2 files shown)


janezd:
This PR won't have >= 95 % diff coverage because `conditions` is the only context setting at the moment, so `if setting.name != "conditions"` can never be true, but it must remain there for the case we add some it in the future.

PR #4755, opened 2020-05-08
Select Rows: Keep operator when possible

### LINKED ISSUE #4737: Select Rows - do not change filter after column change if it is not required

janezd:
##### Issue

Fixes #4737, the real #4737, I hope.

##### Description of changes

After changing the attribute reset the choice to the previous, if it still exists.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -339,6 +339,8 @@ class OWSelectRows(widget.OWWidget):
 
     def set_new_operators(self, attr_combo, adding_all,
                           selected_index=None, selected_values=None):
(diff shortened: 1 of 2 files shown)


Hrovatin:
It works as expected.

PR #4757, opened 2020-05-11
KMeans: data info displayed in the status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -570,6 +580,11 @@ def set_data(self, data):
         else:
             self.invalidate(unconditional=True)
 
+    def _set_input_summary(self):
```
VesnaT, line 583: Just a question. Why the `input` has a function and the `output` hasn't?

aturanjanin: I overlooked this, I will add a function for the `output` now.

Issue #4715, opened 2020-04-30
Softmax Regression issues


PrimozGodec:
Softmax Regression has several issues:
1. The shape of the model prediction shape (probability) is wrong when the domain has more values, but not all are present in the training data (iris[:100]). This can be fixed with replacing `fit` function with `fit_storage` - where the domain is passed together with data, or with a general fix in the base class - now a similar fix exists only for skl-learners. I am more for a second option since that way we keep fit function.
2. Softmax regression does not call `__init__` of the base model like other learners do.

PR #4767, opened 2020-05-13
[FIX] Fix and update Softmax regression learner

PrimozGodec:
##### Issue

Fixes https://github.com/biolab/orange3/issues/4715

##### Description of changes
- Updated softmax regression
- Updated base model in parts where it is required for softmax regression to work correctly.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -206,15 +206,12 @@ class Model(Reprable):
     ValueProbs = 2
 
     def __init__(self, domain=None, original_domain=None):
-        if isinstance(self, Learner):
-            domain = None
-        elif domain is None:
-            raise ValueError("unspecified domain")
         self.domain = domain
         if original_domain is not None:
             self.original_domain = original_domain
         else:
             self.original_domain = domain
+        self.used_vals = None
 
     def predict(self, X):
         if type(self).predict_storage is Model.predict_storage:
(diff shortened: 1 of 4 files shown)


Orange/classification/base_classification.py
```diff
@@ -18,27 +16,8 @@ class ModelClassification(Model):
...
-                probs_ext[:, c, cv] = probs[:, c, i]
-        if not self.supports_multiclass:
-            probs_ext = probs_ext[:, 0, :]
-        return values, probs_ext
+    pass
 
 
 class SklLearnerClassification(SklLearner, LearnerClassification):
-    __returns__ = SklModelClassification
+    __returns__ = SklModel
```
lanzagar, line 23: Why was this changed?

Orange/base.py
```diff
@@ -383,6 +380,30 @@ def one_hot_probs(value):
                 probs[:, i, :] = one_hot(value[:, i])
             return probs
 
+        def extend_probabilities(probs):
+            """
+            Since SKLmodels and models with fit function does not guarantee
```
lanzagar, line 385:
```
Since SklModels and models implementing `fit` and not `fit_storage` do not guarantee
```

Orange/base.py
```diff
@@ -383,6 +380,30 @@ def one_hot_probs(value):
...
+            correct prediction dimensionality, extend dimensionality of
+            probabilities when it does not mach to the number of values in
+            domain.
```
lanzagar, line 388: mach to the number of values in domain -> match the number of values in the domain

Orange/classification/base_classification.py
PrimozGodec: It is my mistake. First I wanted to get rid of SklLearnerClassification class, but then I decided to do refactoring in separate (following) PR. It would be too many changes for this PR.

PR #4768, opened 2020-05-13
[FIX] Fix compatibility with scikit-learn 0.23

PrimozGodec:
##### Issue

The new scikit-learn release yesterday caused orange learners to start failing:
- Learners parameters cannot be retrieved with learner.__init__.__code__ anymore since init is not called directly.
- KMeans is significantly slower: https://github.com/scikit-learn/scikit-learn/issues/17208

##### Description of changes
- Use `inspect.signature` instead of `__code__` to get parameters.
- I suggest temporary pin scikit-learn to != 0.23.0.

##### Discussion
Should we release orange with these fixes after it is merged? Since I think we will not make a major release soon, people installing Orange with pip/conda will still get the newest version of scikit-learn. The same is for tests on add-ons. We just need a release for pip and condo. The installed do not need to be updated since it has packages packed together with orange.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -520,11 +520,14 @@ class SklLearner(Learner, metaclass=WrapperMeta):
     def _get_sklparams(self, values):
         skllearner = self.__wraps__
         if skllearner is not None:
-            spec = inspect.getargs(skllearner.__init__.__code__)
+            spec = list(
+                inspect.signature(skllearner.__init__).parameters.keys()
+            )
             # first argument is 'self'
-            assert spec.args[0] == "self"
-            params = {name: values[name] for name in spec.args[1:]
-                      if name in values}
+            assert spec[0] == "self"
+            params = {
+                name: values[name] for name in spec[1:] if name in values
+            }
         else:
             raise TypeError("Wrapper does not define '__wraps__'")
         return params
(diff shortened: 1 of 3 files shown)


janezd:
Doesn't pinning to 0.22 already do the job of fixing the test? Would it make sense to have a (temporary) pinning in a separate PR, so tests we'll pass, and then you can take your time fixing the pickling issue?

PrimozGodec:
> Doesn't pinning to 0.22 already do the job of fixing the test?
Pinning to >=0.22 does not help since the problematic version is 0.23

> Would it make sense to have a (temporary) pinning in a separate PR, so tests we'll pass, and then you can take your time fixing the pickling issue?
The main reason why I pinned is k-means issue https://github.com/scikit-learn/scikit-learn/issues/17208 K-means is extremely slow with the latest release. I think it should be pinned until they fix the issue and make a new release. If that will be soon maybe it is not required to merge pinning, but meanwhile, tests on PRs will fail. 

For pickling, it is funny since I cannot reproduce the issue at my computer anymore. I will try on some other machine.

markotoplak:
@PrimozGodec, so you suggest releasing 3.25.1 and building pypi/conda packages, which would be 3.25 with frozen scikit-learn dependency?

On conda forge we could fix requirements without doing a new release (there we can release new builds), but for pip we would need a new release.

PrimozGodec:
Yep, I think we need to fix versions on both conda and pip (for people who directly download from pip and for CI on addons). I agree with you that on conda we can easily make a new build without rising the version. On pip we can make either 3.25.1 or 3.25.0.post1.

PR #4779, opened 2020-05-15
[FIX] normalize: Adjust number_of_decimals after scaling

pavlin-policar:
##### Issue
Data, when explicitly centered through the preprocess widget, would show up as having mean 1-e16 in the Feature Statistics widget (and potentially elsewhere). E.g. File (iris) > Preprocess (normalize to have mean=0) > Feature Statistics

##### Description of changes
1. Anything that is explicitly zero-centered will have `number_of_decimals=3`.
2. Sometimes things would show up as `-0.000` because of rounding a tiny negative number e.g. 1e-16 to zero. Zeros are now always positive.

##### Possible issues
This is impossible to do through the UI, but someone could potentially call `normalize` with an `sd>>1`, so all the values would be scaled to something tiny, and then `str_val` would print 0 for everything. Again, this cannot happen in Orange, because there is no way to manually specify the standard deviation. Let me know if you think this is worth fixing.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/preprocess/normalize.py
+++ b/Orange/preprocess/normalize.py
@@ -52,8 +52,14 @@ class Normalizer(Reprable):
             compute_val = Norm(var, avg, 1 / sd)
         else:
             compute_val = Norm(var, 0, 1 / sd)
-        num_decimals = var.number_of_decimals + int(np.ceil(np.log10(sd)))
-        num_decimals = max(num_decimals, 0)  # num decimals can't be negative
+
+        # When dealing with integers, and multiplying by something smaller than
+        # 1, the number of decimals should be decreased, but this integer will
+        # likely turn into a float, which should have some default number of
+        # decimals
+        num_decimals = var.number_of_decimals + int(np.round(np.log10(sd)))
+        num_decimals = max(num_decimals, 1)  # num decimals can't be negative
+
         return var.copy(compute_value=compute_val, number_of_decimals=num_decimals)
 
     def normalize_by_span(self, dist, var: ContinuousVariable) -> ContinuousVariable:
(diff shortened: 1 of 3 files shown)


lanzagar:
> Anything that is explicitly zero-centered will have number_of_decimals=3.

Actually, I don't think zero centering should impact the num. of decimals. Imagine small values (e.g. below 1e-5) that are just centered - they are shifted around 0, but still small and need more decimals. And this _can_ be done in the widget as well!
Rather, it is the rescaling of the variable that should reset the num. of decimals - for normalization to std=1 as well as to 0-1 range it should be OK to use 3 decimals.
For just centering (without scaling) I think the correct fix is to keep the number of decimals of the original variable.

lanzagar:
If we want to be even more fancy and complicate things:
The new number of decimals after rescaling could even be computed and adjusted after rescaling.
E.g. original var had `num_of_decimals=5`, then we rescaled by dividing with `std=10` -> the new var should have 6 decimals. If we divided by `std=0.1` then it should have 4 decimals.

@markotoplak, @janezd - What do you think about just setting to 3 after rescaling or computing the needed number of decimals?

pavlin-policar:
I've added this line here
```
num_decimals += int(-np.floor(np.log10(sd)))
```
which automatically adjusts the number of decimals based on the standard deviation. I've also had to add to this
```
        if float(self._format_str % val) == 0:
             return "%.2f" % 0.0
```
because otherwise, on standardized iris, I would get a mix of 0.000 and 0.0000 (4 zeros), which looked strange.

pavlin-policar:
Let me know if you like this solution, or in what other way this could be solved. If we want to go with this, I'll fix the tests.

Issue #4778, opened 2020-05-15
Edit Domain: replace Ordered with Sort numerically


ajdapretnar:
**Is your feature request related to a problem? Please describe.**

Orange sorts labels in categorical variables alphabetically. I often have numeric labels, such as months of the year (1, 2, 3, 4, 5,...11, 12). When using Pivot Table or some visualization, the labels would be ordered in a wrong way, that is 1, 11, 12, 2, 3,...

**Describe the solution you'd like**

Remove Ordered checkbox from Edit Domain. We assume it does not do anything (is not considered anywhere). Check before removing.
Then replace this with a radio button (or something appropriate) that enables switching between numeric and alphabetic sorting. If labels cannot be interpreted as ints or floats (probably doesn't apply to time variable), then always sort alphabetically.

**Describe alternatives you've considered**
When constructing a Data Table we could check if labels are numbers and sort numerically by default.

**Additional context**
Perhaps it would be nice to have this option in Edit Domain to enable the user some flexibility.

PrimozGodec:
I think that numbers should be always sorted as numbers. I don't see the case when numbers should be sorted as ["1", "12", "2", ...]. This is why I propose a solution where numbers would be always sorted as numbers even in the middle of the string. 
E. g.:
```
[ 
    "something1",
    "something12",
    "something29"
    "something17",
    "something2",
    "something25",
]
```
should be sorted as:
```
[ 
    "something1",
    "something2",
    "something12",
    "something17",
    "something25",
    "something29"
]
```
So, in my opinion, we don't need an option to select the type of sorting but just sort all numbers as numbers. If we decide on a solution with two types of sorting as you proposed. Then the second option (sort as numbers) should include the case where we have letters and numbers mixed.

ajdapretnar:
No, I am absolutely fine with always sorting by numbers. I just thought there might be a use case for alphabetical sorting, but I can't think of one.

While doing this, we should check whether 'Ordered' does anything at all and remove it if not.

PrimozGodec:
Great. I can dig into it. :)

Issue #4529, opened 2020-03-13
Heat Map ignores the order given in the domain


ajdapretnar:
**Describe the bug**

Heat Map ignores the order given in the domain.

**To Reproduce**
1. File (heart-disease)
2. Pivot table (rows=gender,columns=chest pain)
3. Edit Domain (place female to be under male for gender)
4. Heat Map (label with gender, female is first)

**Orange version:**
master

**Expected behavior**
Order of a categorical variable should be the same as given in the domain.

**Screenshots**
<img width="1277" alt="Screen Shot 2020-03-13 at 13 01 43" src="https://user-images.githubusercontent.com/12524972/76619832-15fba080-652c-11ea-9685-d4a38d774fba.png">

ajdapretnar:
So after some exploration, the heat map apparently takes the order the data come in. But in case of discrete variables this is not always desired (as is the case of my months). I assumed Edit Domain fixes this, but it does not. Am I missing something? Is this a bug or the way it should work and we just have no widget in Orange to change the order of rows?

janezd:
No, it sorts them alphabetically, which is less random than the order of appearance.

We could do the same trick as macOS Finder does with some file names: if all values of discrete attributes look numeric, we could sort them as numbers. On the first glance, we'd modify this: https://github.com/biolab/orange3/blob/master/Orange/data/io_base.py#L281. Should we?

As for Edit domain, reordering of values works for me. Even without checking "Ordered"; I don't think this property is observed anywhere in Orange.

ajdapretnar:
No, it doesn't sort them alphabetically, but in order of appearance (in my case for sure). I have switched the order in the original data and then it worked. Otherwise, Edit Domain would work and it doesn't.

It does sort them alphabetically at the beginning, when loading the file, which becomes evident in Pivot Table and so on. Edit Domain does help with this one - it returns the order the user wanted.

I often find wanting a simple 'order as numbers' option or a default. Otherwise I have to click hundreds of time to reorganize, say, hours of the day (1, 10, 11, 12.... 2, 20, 21... in my case). One cannot drag them around in Edit Domain, each switch is done once for one variable.

VesnaT:
The Heat Map displays rows in a "dataset order". I guess the issue could be fixed by #4644.

janezd:
[...]

PR #4793, opened 2020-05-19
[ENH] Sort values naturally when reading files

PrimozGodec:
##### Issue

Fixes https://github.com/biolab/orange3/issues/4778

##### Description of changes
When reading files sort values naturally (before alphabetical sort was used).
Natural sort sorts numbers inside string as numbers: `"sth2"` is before `"sth12"`

#### Discussion

#4778 also said that `Ordered` is not used it is partially true. It sets the `ordered` flag in the DiscreteVariable. This flag tells whether values are ordered in the variable so that other widgets would respect that. I think values are in a current situation always ordered variable and all widgets respect this order (at least none expect FeatureStatistics care about this flag). So I think this flag can be removed. It is used only in the future statistics widget where if this flag is set the min and the max values are shown for the discrete variable.

In case we decide to leave this variable as it is we need to write better documentation for this checkbox.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io_util.py
+++ b/Orange/data/io_util.py
@@ -8,6 +8,7 @@ from Orange.data import (
     is_discrete_values, MISSING_VALUES, Variable,
     DiscreteVariable, StringVariable, ContinuousVariable, TimeVariable,
 )
+from Orange.misc.collections import natural_sorted
 
 __all__ = ["Compression", "open_compressed", "detect_encoding", "isnastr",
            "guess_data_type", "sanitize_variable"]
@@ -121,7 +122,7 @@ def guess_data_type(orig_values, namask=None):
     if namask is None:
         namask = isnastr(orig_values)
     if is_discrete:
-        valuemap = sorted(is_discrete)
+        valuemap = natural_sorted(is_discrete)
         coltype = DiscreteVariable
     else:
         # try to parse as float
--- a/Orange/data/io_base.py
+++ b/Orange/data/io_base.py
@@ -22,6 +22,7 @@ from Orange.data.io_util import Compression, open_compressed, \
     isnastr, guess_data_type, sanitize_variable
 from Orange.data.util import get_unique_names_duplicates
 from Orange.data.variable import VariableMeta
+from Orange.misc.collections import natural_sorted
 from Orange.util import Registry, flatten, namegen
 
 __all__ = ["FileFormatBase", "Flags", "DataTableMixin", "PICKLE_PROTOCOL"]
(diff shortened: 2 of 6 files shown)


janezd:
> This flag tells whether values are ordered in the variable so that other widgets would respect that. I think values are in a current situation always ordered variable and all widgets respect this order (at least none expect FeatureStatistics care about this flag). So I think this flag can be removed.

Not all widgets. Heat map doesn't. #4529! If the flag is set, it would have to be observed!

The flag was meant, in ancient times, to change behaviour of some methods like subsetting in binary trees. If variable's values are ordered, tree induction wouldn't try all possible subsets (as it does now), but only all possible cuts.

As it is, I don't think it is observed anywhere, and the order of values is "not ignored" by chance. It seems we can remove it, but let's discuss.

ajdapretnar:
I would just state that Heat Map does consider the order. Split by uses the order set by the user. #4529 is about labels, which have nothing to do with DiscreteVariable as such. The behaviour is the same as in heat map plots in other libraries.

Orange/data/io_util.py
```diff
@@ -1,5 +1,7 @@
 import subprocess
+import re
```
ajdapretnar, line 2: @PrimozGodec Considering we had issues with nltk replacing re in favor of regex, perhaps we should do the same. It could be our users have variables in non-Latin script.

PrimozGodec: Uf, that is true, but I think they have a different case there. I will check if this particular regex works with re on non-Latin. If it works I would go with re since regex still does not provide macOS wheels.

PrimozGodec: @ajdapretnar I checked used regex with texts in more exotic languages from this PR https://github.com/nltk/nltk/issues/2409. It seems regex used here `re.split(r'(\d+)', element)` works well with those characters. In NLTK library have the case with different regex.

ajdapretnar: Perfect! Thanks for checking!

PrimozGodec:
> Just a thought - does natural_sorted indeed belong to io_util or is it more general? (Our "more general functions" are not well-organized and packed, though, so this may not really matter.)

It is a more general function and can be placed somewhere else. Maybe in Orange.utils?

janezd:
You mean `Orange.util`? I'd prefer to avoid it, it could become a kitchen sink of functions.

`Orange.misc` has similar stuff, but organized into modules. What about putting it there? Perhaps into `Orange.misc.collections`? As I said, it was just a though. These "handy functions" were never organized well in Orange.

PrimozGodec:
@janezd I moved the function to `misc`

PR #4798, opened 2020-05-21
[FIX] Select Rows: Fix incorrectly stored values in settings

janezd:
##### Issue

Select Rows stored values of discrete variables as indices. If the data changed so that values were reordered, it chose wrong values. If values were removed, it could crash with index out of range.

To reproduce, construct File -> Select Rows -> Select Rows. Load Iris, obviously.

- In the second Select Rows, choose the second Iris type.
- In the first Select Rows remove the first type and remove unused classes.

After this the second Select Rows will have selected the (originaly) third (now second) Iris type.

Alternatively:

- In the second Select Rows select the last Iris type
- In the first Select Rows remove one type and remove unused classes.

This will crash the second Select Rows widget.

##### Description of changes

- Use `DomainModel` for combos.
- Internally store variables as instances of `Variable`; context handler converts this to the usual `str, int` presentation. With this, the context handler has an instance of variable, which it needs to see the values.
- Context handler stores values as strings. Conversion from and to `int` is done in context handler to minimize the changes to the widget's code.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -1,6 +1,5 @@
 import enum
 from collections import OrderedDict
-from itertools import chain
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owselectrows.py
```diff
@@ -52,24 +51,49 @@ def encode_setting(self, context, setting, value):
...
+            CONTINUOUS = vartype(ContinuousVariable("x"))
             # Use this after 2022/2/2:
-            # for i, (attr, _, op, values) in enumerate(value):
-            for i, condition in enumerate(value):
-                attr = condition[0]
-                op, values = condition[-2:]
-
-                var = attr in domain and domain[attr]
-                if var and var.is_continuous and not isinstance(var, TimeVariable):
+            # for i, (attr, tpe, op, values) in enumerate(value):
```
ajdapretnar, line 74: Is this intentionally left commented?

PR #4799, opened 2020-05-21
owfile: remove an unused setting: xls_sheet

markotoplak:


Orange/widgets/data/owfile.py
```diff
@@ -188,7 +187,8 @@ def __init__(self):
         layout.addWidget(reload_button, 0, 3)
 
         self.sheet_box = gui.hBox(None, addToLayout=False, margin=0)
-        self.sheet_combo = gui.comboBox(None, self, "xls_sheet",
+        self.sheet_index = 0
+        self.sheet_combo = gui.comboBox(None, self, "sheet_index",
```
janezd, line 191: But now it has an attribute `sheet_index` that is never used. Why not just this?

```python
        self.sheet_combo = QComboBox()
        self.sheet_combo.activated[str].connect(self.select_sheet)
```

(In theory, you could also pass `None` instead of `"sheet_index"`. In practice, it doesn't work - it won't connect the callback.)

markotoplak: No other reason than laziness. Changed into what you suggest.

PR #4800, opened 2020-05-21
SelectRows: Add calendar popup for TimeVariables

### LINKED ISSUE #4676: Select Rows: calendar view for TimeVariable

When setting date ranges with Select Rows, one has to type literal date into line edit (e.g. 2020-02-02).

It would be nice to see calendar view similar to the one in Time Slice for TimeVariable.

aturanjanin:
##### Issue
Fixes #4676.

##### Description of changes
Calendar popup added when setting dates for TimeVariables filtering.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectrows.py
+++ b/Orange/widgets/data/owselectrows.py
@@ -36,6 +36,7 @@ from Orange.widgets.utils.state_summary import format_summary_details
 class SelectRowsContextHandler(DomainContextHandler):
     """Context handler that filters conditions"""
 
+    # pylint: disable=arguments-differ
     def is_valid_item(self, setting, condition, attrs, metas):
         """Return True if condition applies to a variable in given domain."""
         varname, *_ = condition
@@ -54,6 +55,7 @@ class SelectRowsContextHandler(DomainContextHandler):
                 # None will match the value returned by all_vars.get
                 encoded.append((attr, None, op, values))
             else:
+                # check for exact match, pylint: disable=unidiomatic-typecheck
                 if type(attr) is ContinuousVariable \
                         and values and isinstance(values[0], str):
                     values = [QLocale().toDouble(v)[0] for v in values]
@@ -68,6 +70,7 @@ class SelectRowsContextHandler(DomainContextHandler):
                 ))
         return encoded
 
+    # pylint: disable=arguments-differ
     def decode_setting(self, setting, value, domain=None, *_args):
         value = super().decode_setting(setting, value, domain)
         if setting.name == 'conditions':
@@ -79,6 +82,7 @@ class SelectRowsContextHandler(DomainContextHandler):
                 if tpe != [None] \
                         or not tpe and attr not in OWSelectRows.AllTypes:
                     attr = domain[attr]
+                # check for exact match, pylint: disable=unidiomatic-typecheck
                 if type(attr) is ContinuousVariable \
                         or OWSelectRows.AllTypes.get(attr) == CONTINUOUS:
                     values = [QLocale().toString(float(i), 'f') for i in values]
@@ -134,6 +138,7 @@ class SelectRowsContextHandler(DomainContextHandler):
 
 
 class FilterDiscreteType(enum.Enum):
+    # pylint: disable=invalid-name
     Equal = "Equal"
     NotEqual = "NotEqual"
     In = "In"
@@ -375,9 +380,9 @@ class OWSelectRows(widget.OWWidget):
         # Disconnect signals to avoid stray emits when changing variable_model
         for row in range(self.cond_list.rowCount()):
             for col in (0, 1):
-                widget = self.cond_list.cellWidget(row, col)
-                if widget:
-                    widget.currentIndexChanged.disconnect()
+                widg = self.cond_list.cellWidget(row, col)
+                if widg:
+                    widg.currentIndexChanged.disconnect()
         self.cond_list.clear()
         self.cond_list.setRowCount(0)
         self.remove_all_button.setDisabled(True)
@@ -505,7 +510,7 @@ class OWSelectRows(widget.OWWidget):
         elif var is not None and var.is_discrete:
             if oper_combo.currentText().endswith(" one of"):
                 if selected_values:
-                    lc = [x for x in list(selected_values)]
+                    lc = list(selected_values)
                 button = DropDownToolButton(self, var, lc)
                 button.var_type = vtype
                 self.cond_list.setCellWidget(oper_combo.row, 2, button)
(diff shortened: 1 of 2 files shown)


janezd:
Please don't merge before #4798.

There will be conflicts and #4798 is more complex than #4800, so I'd appreciate if rebasing is left for #4800.

Orange/widgets/data/owselectrows.py
```diff
@@ -68,7 +67,7 @@ def decode_setting(self, setting, value, domain=None):
                 op, values = condition[-2:]
 
                 var = attr in domain and domain[attr]
-                if var and var.is_continuous and not isinstance(var, TimeVariable):
+                if var and var.is_continuous:
```
janezd, line 70: Be careful when solving the conflict with #4798: there's no `var` but just `attr`, and it can be a `str` or `Variable`, so you probably want something like

```python
if isinstance(attr, ContinuousVariable) \
    or isinstance(attr, str) and OWSelectRows.AllTypes.get(attr) == CONTINUOUS
```

ajdapretnar:
@aturanjanin #4798 is now merged, so please rebase this and I will check. Thanks! :)

aturanjanin:
@ajdapretnar I've just rebased, could you check if everything looks okay? Thanks!

ajdapretnar:
I have check this and I have a couple of comments.

- Data range in calendar should be the data range of actual data. Now, I can select dates that don't exist in the data set.
- I can select from 2010-01-30 to 2010-01-01. I shouldn't be able to select crossing dates. An option for selecting 'between' and 'outside' could be also the quite standard _data range_ of online platforms. So the first click is the first date, the second click marks the end of the range. If this is too difficult to implement, never mind.
- If my time variable is in hours instead of days, I cannot select the hour range.
- _Between_ and _Outside_ options have strange GUI. They get too narrow. If I try to enlarge the window, the middle condition is resized, not the calendar. Middle condition should be of fixed size and condition resizable.
- This change desperately needs tests. :)

ajdapretnar:
@janezd We're in a pickle here and need your expert opinion/suggestion.
If time variable has only hours, how do we handle it in the view?
If TV has both dates and hours and I want to select hours, too, how to handle this?

ATM, the widget adds dummy date or time to variables, e.g. 1970-01-01 if data has only hours and 00:00:00 if it has only dates.

janezd:
> ... need your expert opinion/suggestion

I wonder in what sense am I an expert here? I don't use time variables and I never heard about `QDateEdit` before. :)

On the first glance, you would need `QDateTimeEdit`, not `QDateEdit`. There you would set the desired display format that would show only date or only hours and minutes or both. It is also possible to control the pop-up that is shown by overriding [`QDateTimeEdit.calendarWidget`](https://doc.qt.io/qt-5/qdatetimeedit.html#calendarWidget). The pop-up can probably be replaced with one that shows only hours, minutes and seconds when desired. It may also be possible to extend `CalendarWidget` with line edits for hours, minutes and seconds, when all are needed.

But I'm just reading the docs and guessing -- which about rounds up my expertise. :)

janezd:
@aturanjanin, I added a commit that adds time to the popup. Now I leave to you to actually connect it, that is, set defaults and connect the signals so that date changes when the user changes the numbers.

Orange/widgets/data/owselectrows.py
```diff
@@ -809,5 +880,73 @@ def resizeEvent(self, QResizeEvent):
...
+        QDateTimeEdit.__init__(self, parent)
+
+        self.parent = parent
+        self.format = datetime_format
+        self.have_date, self.have_time = datetime_format[0], datetime_format[1]
+        self.column = parent.data[:, col_idx]
+        self.set_format()
+        self.setSizePolicy(
+            QSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding))
+        self.dateTimeChanged.connect(parent._invalidate_dates)
```
janezd, line 894: Not like this. The idea of signals is that the object that needs to be informed about something "subscribes" to the signal. Remove this line and instead connect to the signal when you construct the widget, e.g.

```python
                w = DateTimeWidget(self, var_idx, datetime_format)
                w.dateTimeChanged.connect(self._invalidate_dates)
```

As it is now, this widget assumes that the parent has a method `_invalidate_dates` (and, besides, it starts with underscore, so it's kind of private). With the signals, method naming is up to parent. And nobody touches its private methods.

Orange/widgets/data/owselectrows.py
```diff
@@ -809,5 +880,73 @@ def resizeEvent(self, QResizeEvent):
...
+        if self.have_date and self.have_time:
+            if datetime and time:
+                self.setDateTime(QDateTime(datetime, time))
+            else:
+                self.setDateTime(datetime if datetime else self.min_datetime)
+        elif self.have_date and not self.have_time:
+            self.setDate(datetime if datetime else self.min_datetime)
+        elif not self.have_date and self.have_time:
+            self.setTime(datetime if datetime else self.min_datetime)
+        self.dateTimeChanged.connect(self.parent.conditions_changed)
```
janezd, line 934: Did you mean to connect the signal or emit it? I suppose you wanted to do this:

```python
self.dateTimeChanged.emit()
```

To call (or connect) the proper method is up to the parent widget.

Orange/widgets/data/owselectrows.py
```diff
@@ -145,6 +142,36 @@ def _plural(s):
...
+        sublay.addStretch(1)
+        sublay.addWidget(QLabel("Time: "))
+        sublay.addWidget(self.timeedit)
+        sublay.addStretch(1)
+        self.layout().addLayout(sublay)
+        self.parent = self.parent()
+        self.timeedit.dateTimeChanged.connect(self.set_time)
+
+    def set_time(self):
+        self.parent.set_datetime(self.parent.date(), self.timeedit.time())
```
janezd, line 162: Don't call parent's methods. It's better to emit signals and let the parent connect to them. This way, the widget doesn't have to make assumptions about its parent and the parent's methods.

Orange/widgets/data/owselectrows.py
```diff
@@ -145,6 +142,36 @@ def _plural(s):
...
+        self.timeedit.setTime(self.parent().min_datetime.time())
+
+        self._time_layout = sublay = QHBoxLayout()
+        sublay.setContentsMargins(6, 6, 6, 6)
+        sublay.addStretch(1)
+        sublay.addWidget(QLabel("Time: "))
+        sublay.addWidget(self.timeedit)
+        sublay.addStretch(1)
+        self.layout().addLayout(sublay)
+        self.parent = self.parent()
```
janezd, line 158: Don't store the parent. First, it is accessible as `self.parent()`. Second, you don't need it if you use signals instead (see below).

Orange/widgets/data/owselectrows.py
```diff
@@ -484,22 +523,49 @@ def add_datetime(contents):
...
-                    box.controls.append(validator(lc[1]))
+                    box.controls.append(add_numeric(lc[1]))
             elif vtype == 3:  # string:
                 box.controls = [add_textual(lc[0])]
                 if oper in [6, 7]:
                     gui.widgetLabel(box, " and ")
                     box.controls.append(add_textual(lc[1]))
+            elif vtype == 4:  # time:
+                datetime_format = (var.have_date, var.have_time)
+                w = DateTimeWidget(self, var_idx, datetime_format)
```
janezd, line 538: Here, connect `w`'s signals to `_invalidate_dates`. (See the bottom of the file for a longer comment.)

Orange/widgets/data/owselectrows.py
```diff
@@ -484,22 +523,49 @@ def add_datetime(contents):
...
+                w.set_datetime(lc[0])
+                box.controls = [w]
+                box.layout().addWidget(w)
+                self._box = box
+                if oper > 5:
+                    gui.widgetLabel(box, " and ")
+                    w_ = DateTimeWidget(self, var_idx, datetime_format)
```
janezd, line 545: Same here.

Orange/widgets/data/owselectrows.py
```diff
@@ -484,22 +523,49 @@ def add_datetime(contents):
...
         if not adding_all:
             self.conditions_changed()
 
+    def _invalidate_dates(self):
+        w = self._box.controls[0]
+        if len(self._box.controls) > 1:
+            w_ = self._box.controls[1]
+            if w.dateTime() > w_.dateTime():
+                w_.setDateTime(w.dateTime())
+                w_.dateTimeChanged.connect(self.conditions_changed)
```
janezd, line 562: Huh, this reconnects the signal at every change?! You wanted to connect the method or to just call it?

Orange/widgets/data/owselectrows.py
```diff
@@ -809,5 +880,73 @@ def resizeEvent(self, QResizeEvent):
         self.set_text()
 
 
+class DateTimeWidget(QDateTimeEdit):
+    def __init__(self, parent, col_idx, datetime_format):
+        QDateTimeEdit.__init__(self, parent)
+
+        self.parent = parent
```
janezd, line 887: If you use signals, you won't need to store `self.parent`.

Orange/widgets/data/owselectrows.py
```diff
@@ -809,5 +880,73 @@ def resizeEvent(self, QResizeEvent):
...
+        self.format = datetime_format
+        self.have_date, self.have_time = datetime_format[0], datetime_format[1]
+        self.column = parent.data[:, col_idx]
```
janezd, line 890: I'd be happier if `column` was passed as an argument, so the widget wouldn't need to assume how the parent stores data.

janezd:
I apologize for doing it instead of showing you what to do.

I just wanted to fix some issues reported by lint. One of them was that attribute `_box` is defined outside `__init__`. Then I realized that `_box` is problematic as such: each row with conditions would need to have its own `_box`. It's not a property of a widget but of a single line. I couldn't construct a case where this would cause problems, though. I don't understand why the widget still worked. :)

However, instead of defining `_box` in `__init__`, I moved `invalidate_datetime` and `datetime_changed` into `set_new_values`. This makes `set_new_values` even longer, but allows `invalide_datetime` and `datetime_changed` to access `w` and `w_` as local variable (within closure). So we no longer need to define `_box`, and it also has a separate `w` and `w_` for each row of conditions.

I recommend that you check this code, you may find the trick useful.

lanzagar:
Orange now crashes (after this PR was merged) when using Select Rows and selecting a Time variable from metas (but not if the time var is a feature or class...).

To reproduce:
Datasets (select Online Retail) -> Select Rows (select variable InvoiceDate)

janezd:
It would be better if you wrote this as an issue, we'll ovelook it here. (I'll do it.)

PR #4809, opened 2020-05-24
[ENH] HeatMap: Allow setting the center when using diverging palettes

### LINKED ISSUE #4789: Heat map center value feature request

When generating Heat Maps, I would like to be able to display the data relative to some central value. For instance, if 1 is a "no effect" value, I'd like to see values >1 light up as red, <1 as green, for instance. 

I'm using the latest Orange available in Anaconda, v3.25.

Scaling in the heat map widget seems to be a moving target. You used to be able to center the colors around 0. That was sometimes useful when you had data that was actually centered around 0. Now that's gone. It would be nice to be able to center around an arbitrary value. For instance, I have some data that's centered around 0, others that is centered around 1.

My only alternative now is to use the sliders that control the High and Low values manually. However, that has to be changed as the data changes.

**Additional context**

janezd:
##### Issue

Resolves #4789.

##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation

### DIFF
--- a/Orange/widgets/utils/colorgradientselection.py
+++ b/Orange/widgets/utils/colorgradientselection.py
@@ -2,11 +2,12 @@ from typing import Any, Tuple
 
 from AnyQt.QtCore import Qt, QSize, QAbstractItemModel, Property
 from AnyQt.QtWidgets import (
-    QWidget, QSlider, QFormLayout, QComboBox, QStyle
-)
+    QWidget, QSlider, QFormLayout, QComboBox, QStyle,
+    QHBoxLayout, QLineEdit, QLabel)
 from AnyQt.QtCore import Signal
+from AnyQt.QtGui import QFontMetrics, QDoubleValidator
 
-from Orange.widgets.utils import itemmodels
+from Orange.widgets.utils import itemmodels, colorpalettes
 
 
 class ColorGradientSelection(QWidget):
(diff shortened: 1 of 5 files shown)


Hrovatin:
This seems to work. 
However, as the centering window shows only after the diverging palette is chosen this might be hard to find for new users. Other heatmap GUI options seem to be always shown, but do not allow selection/changes if they would not make sense. Maybe the same could be done for the _Center at_ field; e.g. grayed out if non-diverging palette is selected. 
Also, other GUI options seem to be aligned more to the left, while _Center at_ is aligned to the right. There could also be a bit more space between _Center at_ and palette selector.

janezd:
> However, as the centering window shows only after the diverging palette is chosen this might be hard to find for new users. Other heatmap GUI options seem to be always shown, but do not allow selection/changes if they would not make sense. Maybe the same could be done for the Center at field; e.g. grayed out if non-diverging palette is selected.

I wanted to minimize the height, which is already large. The benefit of having a disabled checkbox would be to let the user know it is there, but it still wouldn't tell her how to enable it. What do you say about adding a hint, like in, say Confusion, that shows the first time the user opens the widget? Something like "For data with a 'base value', choose a diverging palette and set its center."

> Also, other GUI options seem to be aligned more to the left, while Center at is aligned to the right. There could also be a bit more space between Center at and palette selector.

I wanted the opposite effect. :) I wanted to "tie" this option to the palette by aligning it to the right and putting it as close to palette as I could. If this intention is not clear, I should indeed align it to the left, like normal options.

**Summary**: I would keep hiding, but add a hint. I will align the checkbox to the left.

Orange/widgets/utils/colorgradientselection.py
```diff
@@ -128,12 +129,13 @@ def thresholdHigh(self) -> float:
     def setThresholdHigh(self, high: float) -> None:
         self.setThresholds(min(self.__threshold_low, high), high)
 
-    def center(self) -> float:
+    def center(self) -> Union[float, bool]:
         return self.__center
 
-    def setCenter(self, center: float) -> None:
+    def setCenter(self, center: Union[float, bool]) -> None:
```
ales-erjavec, line 135: Optional[float]?

Also 
`centerChanged = Signal(float)` needs to be changed to `centerChanged = Signal(object)`

janezd: Not really; `Optional[float]` would be equivalent to `Union[float, None]`, while this was really `Union[float, bool]`.

Terrible design. Now constructor gets either a `float` or `None`, in which there is no centering, and can't be re-enable by changing `center`.

Hrovatin:
If the _Center at_ appears only for certain palettes then it is probably not that odd if it is located on the right, close to the palette selection box. (So left or right would be fine by me then.)

janezd:
> If the Center at appears only for certain palettes then it is probably not that odd if it is located on the right, close to the palette selection box. (So left or right would be fine by me then.)

OK, I prefer it on the right side. Moving back.

PR #4810, opened 2020-05-24
Sample Data: Swap output signals for cross-validation

### LINKED ISSUE #4802: Data sampler: cross validation sampling type is inconsistent
**Describe the bug**

Not exactly a bug, just slight inconsistency.

Bootstrap sends the "training" data to the Data Sample output, while Cross Validation sends the single chosen fold, analogous to testing data. The fixed proportion of data with its default of 70% also implies that the training set goes to the Data Sample output.

This is a problem when switching between sampling types, because the outputs need to be reconnected for CV.

Screenshot 2020-05-21 at 20 49 37

We could solve it by changing by swapping the output meaning and, for clarity, changing "Selected fold" into "Remaining fold".

This would break saved workflows, which we could avoid by detecting the old version, put up a warning that shows that the widget is in "compatibility mode", and show label "Sampled fold" instead of "Selected fold".

janezd:
##### Issue

Resolves #4802.

##### Description of changes

- Outputs are swapped.
- Option is renamed "Unused subset".
- Settings migration detects old setting, sets "compatibility_mode" flag (a schema-only seting). In this mode, the labels and outputs are as they were, and an information message is shown.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation

### DIFF
--- a/Orange/widgets/data/owdatasampler.py
+++ b/Orange/widgets/data/owdatasampler.py
@@ -55,18 +55,33 @@ class OWDataSampler(OWWidget):
     number_of_folds = Setting(10)
     selectedFold = Setting(1)
 
+    # Older versions of the widget had swapped outputs for cross validation
+    # Migrations set this to True for compability with older workflows
+    compatibility_mode = Setting(False, schema_only=True)
+
+    settings_version = 2
+
+    class Information(OWWidget.Information):
+        compatibility_mode = Msg(
+            "Compatibility mode\n"
+            "New versions of widget have swapped outputs for cross validation"
+        )
+
     class Warning(OWWidget.Warning):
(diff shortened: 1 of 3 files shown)


markotoplak:
@janezd, why wouldn't detection of older settings version work for the backward-compatibility hack? When could that be fooled? I don't see it...

janezd:
I thought because of migration, but ...

So we would add a schema-only setting "compatibility_mode" (or whatever) which would be set at migration? If the flag is set, the widget shows a warning and swaps the output (and also changes the labels?)?

markotoplak:
@janezd, yes, if we wanted to support old workflows then we would need to such setting. It is not hard, but still makes code uglier...

So do we want to do it? Or do we break compatibility? Or just keep old behavior? I do not know...

PR #4818, opened 2020-05-26
Discrete variable: remove ordered attribute

PrimozGodec:
##### Issue

As we already discussed in https://github.com/biolab/orange3/pull/4793 ordered attribute from the Discrete variable is not respected in most of the widgets anymore (except Feature Statistics), so it can be removed. Order defined in values tuple should be respected in any widget which needs order.

##### Description of changes
- The ordered attribute is deprecated and will stay so until 3.28.0 then it can be removed.
- Checkbox which sets ordered attribute in the edit domain is removed
- Feature statistic will never show min and max value for the discrete variable (before it happened in a case when ordered attribute was True)
- When saving data always write the order of values for discrete in the header (except in the case with only one value)

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -629,21 +629,16 @@ class DiscreteVariable(Variable):
     .. attribute:: values
 
         A list of variable's values.
-
-    .. attribute:: ordered
-
-        Some algorithms (and, in particular, visualizations) may
-        sometime reorder the values of the variable, e.g. alphabetically.
-        This flag hints that the given order of values is "natural"
-        (e.g. "small", "middle", "large") and should not be changed.
     """
 
     TYPE_HEADERS = ('discrete', 'd', 'categorical')
 
     presorted_values = []
 
-    def __init__(self, name="", values=(), ordered=False, compute_value=None,
-                 *, sparse=False):
+    def __init__(
+            self, name="", values=(), ordered=None, compute_value=None,
+            *, sparse=False
+    ):
         """ Construct a discrete variable descriptor with the given values. """
         values = TupleList(values)  # some people (including me) pass a generator
         if not all(isinstance(value, str) for value in values):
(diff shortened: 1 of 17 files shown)


Orange/data/variable.py
```diff
@@ -632,21 +632,16 @@ class DiscreteVariable(Variable):
...
     TYPE_HEADERS = ('discrete', 'd', 'categorical')
 
     presorted_values = []
 
-    def __init__(self, name="", values=(), ordered=False, compute_value=None,
-                 *, sparse=False):
+    def __init__(
+            self, name="", values=(), compute_value=None, sparse=False,
+            **kwargs
+    ):
```
janezd, line 644: This is a bit dangerous: any call that passes `compute_value` as non-keyword argument will (possibly silently) fail. What about keeping `ordered` as an argument, but changing its default to `None` and the test `if "ordered" in kwargs:` to `if ordered is not None`?

Though: this is unlikely and, besides, such calls won't be fixed until the argument is actually removed and the calls will actually fail. We ignore warnings, including those about deprecations.

Orange/tests/test_filter.py
```diff
@@ -444,7 +444,8 @@ def test_reprs(self):
         flid = IsDefined(negate=True)
         flhc = HasClass()
         flr = Random()
-        fld = FilterDiscrete(self.attr_disc, None)
+        fld = \
+            (self.attr_disc, None)
```
janezd, line 448: I don't understand this.

PrimozGodec: It was a mistake.

Orange/data/variable.py
PrimozGodec: You are right. Didn't think about this possibility. This solution was to avoid problems with __repr__ but now I fixed it differently.

Issue #4493, opened 2020-03-06
Customizing plots (a summary of possible customization options)


Hrovatin:
The Orange plots could be improved with the following customisations:
**Changing text:** 

- [X] Font, size, italic
Applicability: General
Importance: High for size, low for others (KH).
- [ ] Orientation
Applicability: Multiple plots, especially plots with text based axes values (e.g. Line Plot, Mosaic Display).
Importance: Medium (KH). 

Currently there is a problem of displaying long axis value labels. This could be solved by (combination of) text size, orientation, and name shortening with full name display on hover (as in Mosaic Display); the latter is not a solution for plot saving. The Line Plot even hides some axis labels if it runs out of space; this is not appropriate.
- [ ] Colour
Applicability: General
Importance: Low (KH).
- [ ] Notation (%, currency, scientific, number of decimal places, short (5k))
Applicability: General.
Importance: Low for addition of signs (%, $ - this can be done for axis title) and medium for others (e.g. short notation would be useful when there are numbers in thousands). (KH)

**Annotations:**

- [X] Graph title and description
Applicability: General.
Importance: Medium (KH) (can be done in the report description for saving).
- [X] Changing axis titles
Applicability: General. Multiple plots are missing axes titles (e.g. Boxplot, Line Plot for y axis).
Importance: High (KH).
- [ ] Annotations on graph (comments/pointers/lines)
Applicability: General.
Importance: Low. Sometimes it is easier to add pointer/text on a graph than describe the trend in the report. For scatter plot this can be done to some extent by selecting points, which is shown in the report. However, as this is more applicable for editing already saved images it should probably be done outside of Orange. (KH)
     - **JD**: This sounds too complicated
     - **VT:** Seems doable if necessary. 
- [ ] Vertical/horizontal lines
Applicability: General, all plots with numerical axes.
Importance: Medium. One or a few lines can be an useful visual guide. It might also be useful if they could be frozen when changing plotted variables (KH).
- [ ] Statistical (stdev, significant differences on box/bar plot - e.g. between individual groups, ...)
Applicability: Bar plot, box plot.
Importance: Low (KH).

**Changing plots:**

- [X] Graph elements (choosing shapes for individual categories of categorical variable, opacity of lines/other elements, line type, symbol orientation in scatter plot (e.g. arrows), ...)
Applicability: Multiple plots (e.g. Scatter Plot, Line Plot).
[...]

PR #4828, opened 2020-05-28
[ENH] Projection plots: Customizable plots

VesnaT:
##### Issue

Partially implements #4493
Needs https://github.com/biolab/orange-widget-base/pull/74

##### Description of changes
Enable setting label (axes, labels, legend) fonts, annotations and figure properties on projection widgets and some other widgets with plots
-  Scatterplot, MDS, tSNE, Radviz, Freeviz, LinearProjection
-  Line plot
- Calibration plot

##### TODO
- [x] tick font resize for Calibration plot
- [X] save settings as schema-only settings

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owcalibrationplot.py
+++ b/Orange/widgets/evaluate/owcalibrationplot.py
@@ -162,6 +162,7 @@ class OWCalibrationPlot(widget.OWWidget):
     fold_curves = settings.Setting(False)
     display_rug = settings.Setting(True)
     threshold = settings.Setting(0.5)
+    visual_settings = settings.Setting({}, schema_only=True)
     auto_commit = settings.Setting(True)
 
     graph_name = "plot"
@@ -564,8 +565,9 @@ class OWCalibrationPlot(widget.OWWidget):
         if self.score != 0:
             self.report_raw(self.get_info_text(short=False))
 
-    def set_visual_settings(self, *args):
-        self.plot.set_parameter(*args)
+    def set_visual_settings(self, key, value):
+        self.plot.set_parameter(key, value)
+        self.visual_settings[key] = value
 
 
 def gaussian_smoother(x, y, sigma=1.0):
--- a/Orange/widgets/evaluate/owcalibrationplot.py
+++ b/Orange/widgets/evaluate/owcalibrationplot.py
@@ -76,37 +76,6 @@ class ParameterSetter(Setter):
         }
     }
 
-    def __init__(self):
-        def update_font_family(**settings):
-            for label, setter in self.setters[self.LABELS_BOX].items():
-                if label != self.FONT_FAMILY_LABEL:
-                    setter(**settings)
-
-        def update_title(**settings):
-            Updater.update_plot_title_font(self.title_item, **settings)
-
-        def update_axes_titles(**settings):
-            Updater.update_axes_titles_font(self.axis_items, **settings)
-
-        def update_axes_ticks(**settings):
-            Updater.update_axes_ticks_font(self.axis_items, **settings)
-
-        def update_title_text(**settings):
-            Updater.update_plot_title_text(
-                self.title_item, settings[self.TITLE_LABEL])
-
-        self.setters = {
-            self.LABELS_BOX: {
-                self.FONT_FAMILY_LABEL: update_font_family,
-                self.TITLE_LABEL: update_title,
-                self.AXIS_TITLE_LABEL: update_axes_titles,
-                self.AXIS_TICKS_LABEL: update_axes_ticks,
-            },
-            self.ANNOT_BOX: {
-                self.TITLE_LABEL: update_title_text,
-            }
-        }
-
     @property
     def title_item(self):
         return self.titleLabel
--- a/Orange/widgets/evaluate/owcalibrationplot.py
+++ b/Orange/widgets/evaluate/owcalibrationplot.py
@@ -7,6 +7,8 @@ from AnyQt.QtWidgets import QListWidget, QSizePolicy
 
 import pyqtgraph as pg
 
+from orangewidget.utils.visual_settings_dlg import VisualSettingsDialog
+
 from Orange.base import Model
 from Orange.classification import ThresholdClassifier, CalibratedLearner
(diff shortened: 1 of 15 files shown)


Orange/widgets/evaluate/owcalibrationplot.py
```diff
@@ -58,13 +60,51 @@
 )]
 
 
+class ParameterSetter(Setter):
+    initial_settings = OrderedDict({
+        Setter.LABELS_BOX:
+            OrderedDict({name: values for name, values in (
```
janezd, line 66: Perhaps you already told me and I forgot -- why do you explicitly need to use `OrderedDict` here?

VesnaT: > why do you explicitly need to use `OrderedDict` here?
To assure ordering in the visual settings dialog. It is not really necessary.

VesnaT:
> Why can't the user change the font size for ticks?

I think the pyqtgraph does not support that (well, it does, but not correctly and the ticks may overlap with the title), so some `AxisItem` functions should probably be overwritten. I just haven't managed to dive into that yet.

> Now that the user can increase it, could we make the default size for labels smaller? 9 looks great on my computer.

If you prefer, sure.

> Do you think it's necessary to have separate font sizes for categorical and numerical legend?

I don't know. I just kept the old functionality. IMO, the smaller size looks better for the numerical one. We can discuss upon that, now that the sizes can easily be changed during the runtime.

> Reset button resets the graph, but not the values in controls, e.g. spin boxes.

Strange, it does on my computer. I'll check that.

> Modified settings are not saved. I think they should be saved as schema-only settings.

I agree. I thought I'd do it in the next iteration, when the idea is confirmed.

VesnaT:
Are you sure about the Reset button? It works fine for me. How about the `VisualSettingsDialog` itself? You can run it independently of the widget.

I'm always up for the killer functionalities :)

borondics:
@markotoplak, is this going to propagate to the Orange Spectroscopy plotting widgets?

Orange/widgets/evaluate/owcalibrationplot.py
janezd: Pure `dict` is already ordered from (C)Python 3.6 onwards.

VesnaT: Is it? Ok, I'll remove the OrderedDict.

markotoplak:
@borondics, yes.

janezd:
This is indeed becoming huge and needs to be reviewed ASAP. It is amazing, though.

Most of the code is in orange-widget-base (which I think is OK). The code here is either "declarative" or takes care of updating. There is not much to check, except that updating indeed updates -- which is to be done manually, I suppose, because part of it is whether there are any pyqtgraph quirks to avoid. Therefore, I think somebody should systematically click through visualizations. Here are comments after my non-systematic clicking:

- Could the line edit for titles and similar be extended over the entire width of the window?
- Button OK is misleading; it doesn't confirm the changes (as opposed to Cancel), it just closes the window. Perhaps rename to Close, or you can eliminate it.
- Is there a reason to include "Antialias"? Would the user ever want to have non-anti-aliased plots?
- The opacity setting is far from obvious. I saw a spin in line plot, changed it and nothing happenned. Then I checked the tooltip to see what it does. But even after this, I didn't realize that the spins for "Range" and "Selected Range" are also opacities - I had to see the tooltip again. Would it be possible to add a label? Also, could these spins have a step of 5?
- Fonts start with a list of (exotic) fonts whose names start with "." (on macOS). Well, one of them is the default, San Francisco (shown as .SF). I propose the following ordering:
   - default family (.SF on macOS), but without the leading period.
   - all fonts that don't begin with period.
   - all fonts that begin with period (assuming they are exotic system fonts).

    Here I of course assume that this order is not imposed by Qt and is possible to change. (I haven't checked the code.)

This being a non-essential part of Orange and unlikely to cause any debilitating bugs, I'd vote for merge early and fix as needed.

ajdapretnar:
I volunteer. I am extremely qualified for systematic clicking. 👑

VesnaT:
> Is there a reason to include "Antialias"? Would the user ever want to have non-anti-aliased plots?

Yes. Selection in Line plot is too slow for antialiased plots.

> The opacity setting is far from obvious. I saw a spin in line plot, changed it and nothing happenned. Then I checked the tooltip to see what it does. But even after this, I didn't realize that the spins for "Range" and "Selected Range" are also opacities - I had to see the tooltip again. Would it be possible to add a label? Also, could these spins have a step of 5?

I agree with the setting not being obvious, but I'm not sure how to include the label. 
The spins do have step of 5.

janezd:
> Yes. Selection in Line plot is too slow for antialiased plots.

OK.

> I agree with the setting not being obvious, but I'm not sure how to include the label.

Let's wait if anybody has a suggestion, otherwise leave.

> The spins do have step of 5.

I had this idea while writing, and haven't checked. Sorry. :)

VesnaT:
> Button OK is misleading; it doesn't confirm the changes (as opposed to Cancel), it just closes the window. Perhaps rename to Close, or you can eliminate it.

Renamed in https://github.com/biolab/orange-widget-base/pull/83

VesnaT:
> Fonts start with a list of (exotic) fonts whose names start with "." (on macOS). Well, one of them is the default, San Francisco (shown as .SF). I propose the following ordering:
-default family (.SF on macOS), but without the leading period.
-all fonts that don't begin with period.
-all fonts that begin with period (assuming they are exotic system fonts).
Here I of course assume that this order is not imposed by Qt and is possible to change. (I haven't checked the code.)

I can change the order (Qt provides a sorted list of font families), but I'd rather not remove the period. If I removed it, I'd need a mapping (back to original) which would complicate the code.

ajdapretnar:
Works great so far! :)

ajdapretnar:
We decided to ignore coverage, because tests would take months to write and in the meantime this PR would get outdated. Wrote an issue #4891 not to forget about it.

PR #4832, opened 2020-05-29
[ENH] Distributions: change Histogram Data output

### LINKED ISSUE #4813: Distribution Widgets passes no data on ist output connections for class variables.
**Describe the bug**
Actually there are several bugs, which are probably closely related.
1) The distribution widget has 3 output connections to other widegts.
"Selected Data", "Data" and "Histogram Data", but none of theses connections delivers data to a table widget when the selected variable is a class widget. 
2) If I use a numerical variable to calculate the historgram The widget delivers data to all three output channels. "Data" contains only data when one bin is selected.
3) "Histogram Data" contains all data with the bin label, but I would expect the frequency data for the variable.

**Here**, provide a clear and concise description of what the bug is.
I would expect that the "Histogram Data" from the bar chart delivers me the frequency values as shown in the barchart for all classes of a variable in tabular form. 

**To Reproduce**

**Orange version:**
Orange - About - version number.
3.24.1

**Expected behavior**
A data table with the histogram data is expected.

**Screenshots**

**Operating system:**
Windows, OSX or Linux? Which version?
OSX

**Additional context**

thocevar:
##### Issue
Fixes #4813

##### Description of changes

Data output (annotated data) is always active. Before it was active only when something was selected. There was an explicit test that the output is empty after loading the data (no idea why).

Histogram data output now contains the bin and frequency data that is displayed in the widget. I don't know what was its purpose before (it was almost the same as Data).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owdistributions.py
+++ b/Orange/widgets/visualize/owdistributions.py
@@ -10,7 +10,7 @@ from AnyQt.QtGui import QColor, QPen, QBrush, QPainter, QPalette, QPolygonF
 from AnyQt.QtCore import Qt, QRectF, QPointF, pyqtSignal as Signal
 import pyqtgraph as pg
 
-from Orange.data import Table, DiscreteVariable, ContinuousVariable
+from Orange.data import Table, DiscreteVariable, ContinuousVariable, Domain
 from Orange.preprocess.discretize import decimal_binnings, time_binnings, \
     short_time_units
 from Orange.statistics import distribution, contingency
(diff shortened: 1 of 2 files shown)


ajdapretnar:
This drastically changes the Histogram output. Before it output the data with the selected variable split into bins as set in the widget.

I think Data could always be on the output (that is the way elsewhere).

thocevar:
Information about bins (for distributions of continuous variables) is now included in the Data output besides the selection indicator.

janezd:
I wanted to add a commit with updated documentation, but then saw that the current documentation is from the period shortly after the elves departed from Middle Earth and needs to be entirely rewritten and rescreenshot.

We should organize an "update the docs" sprint some day.

Issue #4794, opened 2020-05-19
CSV Import interpret discrete variables from the CSV file as StringVariable


PrimozGodec:
**Describe the bug**

CSV Import interprets discrete variables from the CSV file as StringVariable.

**To Reproduce**

1. Place CSV import widget at the canvas
2. Open a CSV file that has columns that should be categorical - Iris column in the Iris dataset.
3. Leave Column Type Automatic
4. Connect data table and see that categorical features are in meta as String variables

**Orange version:**

master

**Expected behavior**

CSV Import interprets discrete variables from the CSV file as DiscreteVariable - the same heuristic can be used as in the File widget.

**The question here is, is it worth to invest time and fix this issue since we are planning to merge this widget with the File widget?**

janezd:
Friday's note: @PrimozGodec will explore whether it is possible to improve detection of variable types without hurting the speed and memory requirements.

ales-erjavec:
Closed by gh-4838

PR #4838, opened 2020-06-02
[ENH] CSV Import: guess data types

PrimozGodec:
##### Issue

Implements #4794

##### Description of changes

Implemented guessing strategy for CSV import which should match the strategy in `io_utils.guess_data_type`. 

This PR adds another iteration over the columns. For each column, it checks for the data type according to data. Complex operations here are:
- unique: Pandas unique is based on a hash table - less complex than Numpy's one
- casting times to date-time

@ales-erjavec is this kind of guessing acceptable or would decrease the performance of the widget too much?

#### TODO
- Tests

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcsvimport.py
+++ b/Orange/widgets/data/owcsvimport.py
@@ -1259,6 +1259,8 @@ def load_csv(path, opts, progress_callback=None):
             float_precision="round_trip",
             **numbers_format_kwds
         )
+        df = guess_types(df, dtypes, columns_ignored)
+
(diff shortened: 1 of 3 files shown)


ales-erjavec:
> is this kind of guessing acceptable or would decrease the performance of the widget too much?

I think it is ok.

PrimozGodec:
Ok. Then I will write a few tests before it is mreged.

Issue #4763, opened 2020-05-13
Wrong GUI scaling


mriza:
**Describe the bug**

Orange3 (version 3.25.0) was installed using conda 4.8.3 under Fedora 32. The display resolution is 1920x1080 with 100% scaling. Orange was displayed in 200% scaling when run using `python -m Orage.canvas` or using Anaconda Navigator.

**To Reproduce**

Steps to reproduce the behavior go **here**:
1. Install anaconda
2. activate base environment 'conda activate base'
3. python -m Orange.canvas    -OR-
4. Click Orange Launch button in Anaconda Navigator

**Orange version:**
Orange - 3.25.0

**Expected behavior**
Orange run in 100% scaling
 
**Screenshots**
![Screenshot from 2020-05-13 14-15-25](https://user-images.githubusercontent.com/1440628/81782636-54a9d900-9524-11ea-9d6c-78d5a0f5e1c4.png)

**Operating system:**
Linux Fedora 32

mriza:
Temporary workaround for this bug is to set `QT_SCALE_FACTOR` and `QT_FONT_DPI` when invoking `python -m Orange.canvas` command.

I succesfully run Orange in acceptable resolution with
```
QT_SCALE_FACTOR = 50%
QT_FONT_DPI = 192 DPI
```

run this command in terminal :
QT_SCALE_FACTOR=0.5 QT_FONT_DPI=192 python -m Orange.canvas
![Screenshot from 2020-05-13 14-24-48](https://user-images.githubusercontent.com/1440628/81783445-85d6d900-9525-11ea-8ced-14b3239c9f17.png)

ajdapretnar:
Similar to #2832.

mriza:
> Similar to #2832.

Yes, similar but different, mine is not even HiDPI nor 4K

ajdapretnar:
I don't think it is a temporary workaround, but a permanent solution. Orange is based on Qt and from what I know, that is their issue, not ours.

mriza:
You're right. I will just create a launcher script for now, until this issue resolved by Qt.

ajdapretnar:
I know it sucks, I was struggling with the same on Linux Mint and this way the only things that solved it. Sorry about that.

mriza:
That's okay, I'll be patient

> You're right. I will just create a launcher script for now, until this issue resolved by Qt.
**Orange Application Launcher with Qt resolution workaround**

create **/opt/orange/orange**  :
```
[...]

PR #4849, opened 2020-06-11
Set PyQt5 as dependency

robertcv:
##### Issue
Older versions of PyQt5 cause problems like #4763. The solution is to set minimum supported version and thereby also setting PyQWt5 as the default binding for Qt (although Orange theoretically  also supports PyQt4, PySide and PySide2)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/requirements-gui.txt
+++ b/requirements-gui.txt
@@ -1,7 +1,8 @@
 orange-canvas-core>=0.1.12,<0.2a
 orange-widget-base>=4.5.0
 
-# PyQt4/PyQt5 compatibility
+PyQt5>=5.12
+PyQtWebEngine>=5.12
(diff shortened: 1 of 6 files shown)


requirements-gui.txt
```diff
@@ -1,7 +1,8 @@
 orange-canvas-core>=0.1.12,<0.2a
 orange-widget-base>=4.5.0
 
-# PyQt4/PyQt5 compatibility
+PyQt5>=5.12,<5.14
```
lanzagar, line 4: Why `<5.14`?
I am using 5.15 and everything seems to be working fine...

PR #4852, opened 2020-06-12
Remove TupleList

PrimozGodec:
##### Issue

Test started to warn that TupleList should be removed in version >=3.27

##### Description of changes
TupleList removed

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/tests/test_variable.py
+++ b/Orange/data/tests/test_variable.py
@@ -8,7 +8,6 @@ import unittest
 import pickle
 import pkgutil
 from datetime import datetime, timezone
-from distutils.version import LooseVersion
 
 from io import StringIO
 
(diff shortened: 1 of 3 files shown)


lanzagar:
This breaks loading pickled data saved in latest Orange (3.26).
We probably need to add the TupleList class back (so that it can be unpickled), but can keep using tuples directly in the rest of the code (as we currently are, after this PR).

We can discuss further before action, this is just a note that this needs to be handled carefully!

janezd:
Could we have somethine like `TupleList = tuple  # backward compatibility`, so unpickle would work, but the class would be no more?

lanzagar:
Seems better to me (and works fine as far as I can tell).

Sidenote (@PrimozGodec ): similar care probably needs to be taken when removing the ordered attribute - I now get a Read Warning in the File widget when opening a pickled dataset that says "ordered is deprecated..."

PrimozGodec:
@lanzagar it is commented in the pull request which will warn to remove ordered and now also newly introduced test in https://github.com/biolab/orange3/pull/4879 will fail if someone removes ordered.

PR #4855, opened 2020-06-17
Add cosine metric for Louvain clustering.

Hrovatin:
##### Issue
Adds cosine distance to Louvain clustering widget. 
I need this for "supplementary code" of a paper. I made the change in my own Orange version, but now it needs to be available to everyone.

##### Description of changes
This simply adds an option to the Louvain widget to use the cosine distance in sklearn's NearestNeighbors. However, NearestNeighbors uses brute force computation with cosine distance, if I understand it correctly (here is the base class https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/neighbors/_base.py with specified VALID_METRIC for each nearest neighbours algorithm). Thus a modification would be required to prevent the use of cosine distance on data with too many samples. (It works fine with my data that has ~1000 samples with ~500 features.)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owlouvainclustering.py
+++ b/Orange/widgets/unsupervised/owlouvainclustering.py
@@ -40,7 +40,7 @@ _MAX_PCA_COMPONENTS = 50
 _DEFAULT_PCA_COMPONENTS = 25
 _MAX_K_NEIGBOURS = 200
 _DEFAULT_K_NEIGHBORS = 30
-
+_MAX_COSINE_ROWS = 2000
 
 METRICS = [("Euclidean", "l2"), ("Manhattan", "l1"), ("Cosine", "cosine")]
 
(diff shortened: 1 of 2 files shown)


janezd:
@Hrovatin, could you experiment with different data sizes (rows and columns) and decide how large is too large?

janezd:
@Hrovatin, I added the check for data size. Please check whether it works correctly, in particular that the error message disappears when it should, and that there is no data on output when clustering failed.

Also, set the maximum to some reasonable value. Currently it's arbitrary, 2000.

pavlin-policar:
If you want to be really efficient you can use this little trick. The ranking of the nearest neighbors using the cosine metric is the same as the ranking using euclidean distance on row-normalized data. So, it is possible to get around using brute-force search and using the scikit-learn ball tree, but the data must be normalized first. This just gives us the ranking, and the actual distances have to be computed later on, but then you only have to calculate N\*k distances instead of N\*N. I did this in openTSNE [here](https://github.com/pavlin-policar/openTSNE/blob/master/openTSNE/nearest_neighbors.py#L90-L124) and it works fine. You can decide if it's worth the hassle though...

janezd:
Closed in favour of #4864.

PR #4862, opened 2020-06-19
tsne: require opentsne>=0.4.3

pavlin-policar:
##### Changes
Bump minimum required version of openTSNE to 0.4.3. This also gets rid of numba. I removed the only mention of numba I was able to find in this repo. There are no mentions of numba in orange-canvas and orange-widget-base, so I think this was the only one.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
Windows installers built with 0.4.3 are 451 MB instead of 467 MB. I hoped for more, but OK.

PR #4864, opened 2020-06-19
[ENH] Louvain Clustering: Add cosine similarity

janezd:
As suggested by @pavlin-policar in a comment to #4855, cosine distance can be computed as Euclidean on row-normalized data, because we care only about ranking. This allows NearestNeighbours to use ball trees.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/clustering/louvain.py
+++ b/Orange/clustering/louvain.py
@@ -44,6 +44,13 @@ def matrix_to_knn_graph(data, k_neighbors, metric, progress_callback=None):
 
     """
     # We do k + 1 because each point is closest to itself, which is not useful
+    if metric == "cosine":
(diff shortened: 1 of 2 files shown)


janezd:
@pavlin-policar, I didn't really read the entire widget's code, but it seems that it would suffice to normalize the data and change the metric in `matrix_to_knn_graph`. As I see, the absolute distances are not needed at all (`return_distance=False`, and I see no other mention of `metric`). 

Am I correct?

pavlin-policar:
> As I see, the absolute distances are not needed at all

Yes, that's right. We build a graph, and the edges are weighted with the Jaccard coefficient of shared nearest neighbors between points. Distances are not used anywhere.

This codecov is doing strange things again...

PR #4879, opened 2020-07-03
[FIX] Re-add TupleList to DiscreteVariable

PrimozGodec:
##### Issue

Fixes https://github.com/biolab/orange3/issues/4875

##### Description of changes
Re-add TupleList and write test

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/data/variable.py
```diff
@@ -622,7 +625,7 @@ def __init__(
             *, sparse=False
     ):
         """ Construct a discrete variable descriptor with the given values. """
-        values = tuple(values)  # some people (including me) pass a generator
+        values = TupleList(values)  # some people pass a generator
```
lanzagar, line 628: Correct me if I'm wrong, but I think this can remain just `tuple` (for creating new instances) as long as the above is there for unpickling of saved older classes.

PrimozGodec: You are right :)

PR #4900, opened 2020-07-13
PCA: clear output when input data is removed

### LINKED ISSUE #4899: PCA does not clear output
PCA widget does not clear output after the input data is removed.

**To reproduce:**
File -> Iris -> PCA -> Data Table -> Remove connection between File and PCA

aturanjanin:
##### Issue
Fixes #4899.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owpca.py
+++ b/Orange/widgets/unsupervised/owpca.py
@@ -116,6 +116,8 @@ class OWPCA(widget.OWWidget):
         self.clear()
         self.information()
         self.data = None
+        if not data:
+            self.clear_outputs()
         if isinstance(data, SqlTable):
             if data.approx_len() < AUTO_DL_LIMIT:
                 data = Table(data)
--- a/Orange/widgets/unsupervised/tests/test_owpca.py
+++ b/Orange/widgets/unsupervised/tests/test_owpca.py
@@ -248,6 +248,10 @@ class TestOWPCA(WidgetTest):
         self.assertEqual(trans.domain.attributes, outdom.metas[1:])
         np.testing.assert_equal(trans.X, output.metas[:, 1:])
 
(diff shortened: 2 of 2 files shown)


Orange/widgets/unsupervised/owpca.py
```diff
@@ -141,6 +141,7 @@ def set_data(self, data):
 
     def fit(self):
         self.clear()
+        self.clear_outputs()
```
janezd, line 144: If data is present, this will first output `None` and then the actual data. Such redundant signals should be avoided because the recepient may have some kind of caching speed-ups.

Method `set_data` already checks whether the table is empty, and clears outputs and exits if so. This would be a proper place to check for `data is None`.

The check in `fit` is needed only for the case when normalization is changed when no data is present.

While you're fixing this: I think that `set_data` should also set `self.data` to `None` in the two cases where it raises errors. This will prevent `fit` from trying to fit when there are no attributes.

aturanjanin: Thank you for your review, I'll fix this.

Just one thing about this last suggestion regarding the `self.data` setting to `None`. Isn't this already handled in the line 118:
https://github.com/biolab/orange3/blob/2733ef0961c326a2310f732ffb9ae2084d7ee33a/Orange/widgets/unsupervised/owpca.py#L118 
and it doesn't change if on input there is data with no attributes. In this case the if condition `if not data.domain.attributes:` is met and `self.data` will remain `None`

janezd: > Just one thing about this last suggestion regarding the self.data setting to None. Isn't this already handled in the line 118:

Of course. I'm sorry.

PR #4904, opened 2020-07-13
Distances: data info displayed in the status bar

aturanjanin:
##### Description of changes

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


PrimozGodec:
The output here is a distance matrix which is N x N matrix. Currently, we just show the length of the matrix. Should we show both the height and width of the matrix instead? I know it is always a square matrix but I think users would have a better feeling if they see something like `150 x 150`.

janezd:
I don't disagree. :) I'd use 150×150 -- no spaces and symbol × instead of letter x.

aturanjanin:
So we should put 150×150 into both, summary and details?

janezd:
Maybe. Let's decide tomorrow.

PrimozGodec:
Note: `self.info.set_output_summary(summary, summary)` is used since in the cases with string summary description is not set automatically to same value than it is done for int summary. I suggest unifying those behaviours.

PR #4908, opened 2020-07-17
[ENH] Rank widget computation in a separate thread

PrimozGodec:
##### Issue

Rank widget blocks Orange while computing on a larger dataset

##### Description of changes
Use `ConcurrentWidgetMixin` in the rank widget.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -1,39 +1,39 @@
-import warnings
-from collections import namedtuple, OrderedDict
 import logging
+import warnings
+from collections import OrderedDict, namedtuple
 from functools import partial
 from itertools import chain
+from types import SimpleNamespace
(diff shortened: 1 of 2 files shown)


markotoplak:
@PrimozGodec, please make sure that run does use any attributes from the object, because they can now be set independently (connecting a new data set to it could do that).

As it is now, run() calls get_scorer_score and get_method_scores which both use self.data. To be safe, could you make run() and all the methods it calls static.

PrimozGodec:
@markotoplak I changed now such that problematic methods get data now and are static. lru_cache uses hashes of arguments to store results.

PrimozGodec:
Documentation test will pass after merging this PR https://github.com/biolab/orange3/pull/5010

I hope it is working now correctly.

PR #4920, opened 2020-07-24
[FIX] MDS: Move lines when points are jittered

### LINKED ISSUE #4898: MDS: pair lines don't follow jittering
When setting jittering on points in MDS, the lines connecting similar pairs don't follow the points. When re-setting the Show similar pairs slider, they get updated.

To reproduce:
Datasets (zoo) - MDS. Jitter the points. Lines are not properly placed after jittering.

janezd:
##### Issue

Fixes #4898 ...

##### Description of changes

... by calling `update_pairs` from `update_jittering` and ...

##### Includes
- [X] Code changes

... but no tests because a test would be trivial (`update_pairs.assert_called`), or it would add yet another test that would run MDS and increase the running time, or it would require lots of mocking (and be trivial at the end, again).

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -102,6 +102,10 @@ class OWMDSGraph(OWScatterPlotBase):
         super().update_coordinates()
         self.update_pairs(reconnect=False)
 
+    def update_jittering(self):
+        super().update_jittering()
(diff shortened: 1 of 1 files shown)


ajdapretnar:
This fixes the problem at hand, so I am merging. However, even on master, things gets slow (even for zoo) when changing sliders (jittering and show similar pairs). Are we calling `on_release()` here?

PR #4922, opened 2020-07-24
[FIX] Edit Domain (and perhaps other widgets) could cause missing data later in the workflow

### LINKED ISSUE #4895: Color node is destructive
Hello !
so

**Explain the bug**
I cant use the color node on my data since i downloaded the last version of orange (3.26.0), the version i was working on before was the 3.24.1.
From now on, if i use the color node on my categorical data, this categorical data will be lost in the following nodes (? appears instead of 6 or 3 for example), but after further researchs, it seems that its my way to make csv that is not well recognized ? (i saved my csv with the save data node, reused it and the color node was working on this new csv

**How to reproduce the problem.** 

[orange_color_node_bug.zip](https://github.com/biolab/orange3/files/4895384/orange_color_node_bug.zip)

To reproduce the problem, open the workspace, put the csv_not_working.csv in a csv node, and look at the node after the color node , there will be no colors and the categorical information is now "?"

the csv_working is a csv made with orange from a csv not working but actually work, so maybe i dont make my csv the good way idk)

**Which version of Orange are you using? How did you install it?**
orange is in 3.26.0, installed with the .exe for windows

**What is your operating system?**
I am on windows 10 updated

here is an image of what i saw : https://cdn.discordapp.com/attachments/633377197196836875/730338897820909680/unknown.png)

### LINKED ISSUE #4896: After merging custumizable plots, tests segfault on Linux
Started happening after #4828 was merged. Linux, Ubuntu 18.04, Python 3.7, PyQt 5.12. I have no problems with reports or help windows in Orange.

If I run `python -m unittest Orange.widgets.visualize.tests.test_owlineplot` I get a segfault with the following backtrace:

```
#0  0x00007fffc4aed970 in QOpenGLContext::isValid() const () from /home/marko/venv37/lib/python3.7/site-packages/PyQt5/Qt/lib/libQt5Gui.so.5
#1  0x00007fffb79bd8b9 in ?? () from /home/marko/venv37/lib/python3.7/site-packages/PyQt5/Qt/lib/libQt5WebEngineCore.so.5
#2  0x00007fffb79be90c in ?? () from /home/marko/venv37/lib/python3.7/site-packages/PyQt5/Qt/lib/libQt5WebEngineCore.so.5
#3  0x00007fffbe60015e in QWebEnginePagePrivate::QWebEnginePagePrivate(QWebEngineProfile*) ()
   from /home/marko/venv37/lib/python3.7/site-packages/PyQt5/Qt/lib/libQt5WebEngineWidgets.so.5
#4  0x00007fffbe600e54 in QWebEnginePage::QWebEnginePage(QObject*) ()
   from /home/marko/venv37/lib/python3.7/site-packages/PyQt5/Qt/lib/libQt5WebEngineWidgets.so.5
#5  0x00007fffbe60e02b in QWebEngineView::page() const ()
   from /home/marko/venv37/lib/python3.7/site-packages/PyQt5/Qt/lib/libQt5WebEngineWidgets.so.5
#6  0x00007fffbe835b59 in meth_QWebEngineView_page () from /home/marko/venv37/lib/python3.7/site-packages/PyQt5/QtWebEngineWidgets.so
#7  0x0000555555684e67 in _PyMethodDef_RawFastCallKeywords ()
#8  0x0000555555684c70 in _PyCFunction_FastCallKeywords ()
#9  0x00005555556f88c0 in _PyEval_EvalFrameDefault ()
#10 0x00005555556f38ce in _PyEval_EvalCodeWithName ()
```

Also some other tests segfault, but Orange seems to be behaving normally.

Happens both  with pyqtgraph 1.0.0 and 1.1.0.

Intentionally disabling WebEngine by breaking the import "fixes" segfaults.

janezd:
##### Issue

Fixes #4895.

Equality of variables also checks equality of their `compute_value`. This can only work if `Transformation` and derived classes have `__eq__` and `__hash__`.

To fix #4895, it would suffice to change `oweditdomain.LookupMappingTransform`. The PR does it for other classes too, to prevent similar future mishaps.

##### Description of changes

Define `__eq__` and `__hash__` where needed.

Any add-ons that don't have will now misbehave in a different way -- before this, `__eq__` always returned `False`, even when transforms were the same. Now it will sometimes return `True` even though they may be different. This will happen if the class defines additional fields that are not checked by the base class.

The base class can neither check children's extra attributes, neither can it raise an exception (e.g. in a meta class), because many derived classes (even some that define extra attributes) do not need `__eq__`.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/preprocess/transformation.py
+++ b/Orange/preprocess/transformation.py
@@ -143,7 +143,7 @@ class Lookup(Transformation):
         :param variable: The variable whose transformed value is returned.
         :type variable: int or str or :obj:`~Orange.data.DiscreteVariable`
         :param lookup_table: transformations for each value of `self.variable`
-        :type lookup_table: np.array or list or tuple
+        :type lookup_table: np.array
         :param unknown: The value to be used as unknown value.
         :type unknown: float or int
         """
@@ -162,14 +162,10 @@ class Lookup(Transformation):
         return np.where(mask, self.unknown, values)
 
     def __eq__(self, other):
-        def nan_equal(x, y):
-            return x == y or np.isnan(x) and np.isnan(y)
-
         return super().__eq__(other) \
-               and len(self.lookup_table) == len(other.lookup_table) \
-               and all(nan_equal(x, y) or np.isnan(x) and np.isnan(y)
-                       for x, y in zip(self.lookup_table, other.lookup_table)) \
-               and nan_equal(self.unknown, other.unknown)
+               and np.allclose(self.lookup_table, other.lookup_table,
+                               equal_nan=True) \
+               and np.allclose(self.unknown, other.unknown, equal_nan=True)
 
     def __hash__(self):
         return hash((type(self), self.variable,
--- a/Orange/preprocess/transformation.py
+++ b/Orange/preprocess/transformation.py
@@ -48,13 +48,6 @@ class Transformation(Reprable):
         raise NotImplementedError(
             "ColumnTransformations must implement method 'transform'.")
 
-
-class Identity(Transformation):
-    """Return an untransformed value of `c`.
-    """
-    def transform(self, c):
-        return c
-
     def __eq__(self, other):
         return type(other) is type(self) and self.variable == other.variable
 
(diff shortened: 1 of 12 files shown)


Orange/preprocess/transformation.py
```diff
@@ -48,25 +48,21 @@ def transform(self, c):
...
 
-
-class Identity(Transformation):
-    """Return an untransformed value of `c`.
-    """
-    def transform(self, c):
-        return c
-
     def __eq__(self, other):
         return type(other) is type(self) and self.variable == other.variable
```
lanzagar, line 52: How about the default `__eq__` for `Transformation` along these lines:
```py
return type(other) is type(self) and vars(self) == vars(other)
```

Seems that that is usually what we want and it would make a lot of trivial overloads of eq unnecessary?

`__hash__` is a bit more tricky since class attributes can often be unhashable objects. We could leave it as it is or try hashing a sorted tuple of `vars(self)`. The latter should work for simple cases like `Indicator` and at least fails more noticably if it should have been overloaded in a subclass but was not!

Orange/preprocess/transformation.py
```diff
@@ -156,3 +160,17 @@ def transform(self, column):
...
+
+    def __eq__(self, other):
+        def nan_equal(x, y):
+            return x == y or np.isnan(x) and np.isnan(y)
+
+        return super().__eq__(other) \
+               and len(self.lookup_table) == len(other.lookup_table) \
+               and all(nan_equal(x, y) or np.isnan(x) and np.isnan(y)
+                       for x, y in zip(self.lookup_table, other.lookup_table)) \
+               and nan_equal(self.unknown, other.unknown)
```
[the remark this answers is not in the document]
janezd: I changed the docstring: lookup_table must obviously be an array. As for checking, I went for `np.allclose(..., ..., equal_nan=True)`.

Orange/preprocess/transformation.py
janezd: I'm hesitant here. It looks like a good idea, but making the parent class to smart might shoot a derived class in the foot. Somebody could add an attribute without realizing it's used in comparisons. Maybe it's better to be explicit.

lanzagar: You convinced me at first, but then again, you can make an equivalent claim for the other side:
Somebody could add an attribute without realizing it's _NOT_ used in comparisons.

Maybe I preferred comparing everything (vs the minimum) because it behaves closer to how it was before - if some class doesn't overload eq it might erroneously return False (like up to now), but a True should be correct.

In the end it does not really matter, if something is not working like it should it's a bug to be fixed in either case.
I don't mind merging it with either default.

PR #4923, opened 2020-07-27
[ENH] Bar Plot: New widget

VesnaT:
##### Issue

Add Bar Plot widget.

##### Description of changes
I'll add the docs when the UI is approved.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/__init__.py
+++ b/Orange/widgets/utils/__init__.py
@@ -5,6 +5,7 @@ from collections import deque
 from typing import (
     TypeVar, Callable, Any, Iterable, Optional, Hashable, Type, Union
 )
+from xml.sax.saxutils import escape
 
 from AnyQt.QtCore import QObject
 
@@ -26,7 +27,7 @@ def vartype(var):
 
 
 def progress_bar_milestones(count, iterations=100):
-    return set([int(i*count/float(iterations)) for i in range(iterations)])
+    return {int(i * count / float(iterations)) for i in range(iterations)}
 
 
 def getdeepattr(obj, attr, *arg, **kwarg):
@@ -35,9 +36,10 @@ def getdeepattr(obj, attr, *arg, **kwarg):
     return deepgetattr(obj, attr, *arg, **kwarg)
 
 
-def to_html(str):
-    return str.replace("<=", "&#8804;").replace(">=", "&#8805;").\
+def to_html(s):
+    return s.replace("<=", "&#8804;").replace(">=", "&#8805;"). \
         replace("<", "&#60;").replace(">", "&#62;").replace("=\\=", "&#8800;")
+
 
 getHtmlCompatibleString = to_html
 
@@ -136,3 +138,24 @@ def enum_get(etype: Type[_E], name: str, default: _T1) -> Union[_E, _T1]:
         return etype[name]
     except LookupError:
         return default
+
+
+def instance_tooltip(domain, row, skip_attrs=()):
+    def show_part(_point_data, singular, plural, max_shown, _vars):
+        cols = [escape('{} = {}'.format(var.name, _point_data[var]))
+                for var in _vars[:max_shown + len(skip_attrs)]
+                if _vars is domain.class_vars
+                or var not in skip_attrs][:max_shown]
+        if not cols:
+            return ""
+        n_vars = len(_vars)
+        if n_vars > max_shown:
+            cols[-1] = "... and {} others".format(n_vars - max_shown + 1)
+        return \
+            "<b>{}</b>:<br/>".format(singular if n_vars < 2 else plural) \
+            + "<br/>".join(cols)
+
+    parts = (("Class", "Classes", 4, domain.class_vars),
+             ("Meta", "Metas", 4, domain.metas),
+             ("Feature", "Features", 10, domain.attributes))
+    return "<br/>".join(show_part(row, *columns) for columns in parts)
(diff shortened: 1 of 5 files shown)


Orange/widgets/utils/__init__.py
```diff
@@ -23,7 +24,7 @@ def vartype(var):
 
 
 def progress_bar_milestones(count, iterations=100):
-    return set([int(i*count/float(iterations)) for i in range(iterations)])
+    return set(int(i*count/float(iterations)) for i in range(iterations))
```
janezd, line 27: I agree that this is certainly an improvement, but
`{int(i * count / float(iterations)) for i in range(iterations)}` could be even better. :)

Orange/widgets/utils/__init__.py
```diff
@@ -32,8 +33,8 @@ def getdeepattr(obj, attr, *arg, **kwarg):
     return deepgetattr(obj, attr, *arg, **kwarg)
 
 
-def to_html(str):
-    return str.replace("<=", "&#8804;").replace(">=", "&#8805;").\
+def to_html(str_):
+    return str_.replace("<=", "&#8804;").replace(">=", "&#8805;").\
         replace("<", "&#60;").replace(">", "&#62;").replace("=\\=", "&#8800;")
```
janezd, line 38: What about `return escape(str_, entities={"<=": "&#8804;", ">=": "&#8805;", "=\\=": "&#8800;"})`?

By the way, for long sequences of replacements I sometimes prefer

```python
reduce(lambda s, t:s.replace(*t), (
    ("<=", "&#8804;"), 
    (">=", "&#8805;"),
    ("<", "&#60;"),
    (">", "&#62;"),
    ("=\\=", "&#8800;")
    )), s)
```

But this is a matter of taste, and `escape` fits better here anyway.

Orange/widgets/utils/__init__.py
```diff
@@ -32,8 +33,8 @@ def getdeepattr(obj, attr, *arg, **kwarg):
     return deepgetattr(obj, attr, *arg, **kwarg)
 
 
-def to_html(str):
```
janezd, line 35: Why not just `s`? :)

Orange/widgets/utils/__init__.py
```diff
@@ -90,3 +91,24 @@ def apply_all(seq, op):
     """Apply `op` on all elements of `seq`."""
     # from itertools recipes `consume`
     deque(map(op, seq), maxlen=0)
+
+
+def instance_tooltip(domain, row, skip_attrs=()):
+    def show_part(_point_data, singular, plural, max_shown, _vars):
+        cols = [escape('{} = {}'.format(var.name, _point_data[var]))
+                for var in _vars[:max_shown + 2]
+                if _vars == domain.class_vars
```
janezd, line 100: I'd maybe prefer `is` instead of `==` here. (Very marginal faster, but) the intent is clearer.

Orange/widgets/utils/__init__.py
```diff
@@ -90,3 +91,24 @@ def apply_all(seq, op):
     """Apply `op` on all elements of `seq`."""
     # from itertools recipes `consume`
     deque(map(op, seq), maxlen=0)
+
+
+def instance_tooltip(domain, row, skip_attrs=()):
+    def show_part(_point_data, singular, plural, max_shown, _vars):
+        cols = [escape('{} = {}'.format(var.name, _point_data[var]))
+                for var in _vars[:max_shown + 2]
```
janezd, line 99: I suppose `+ 2` should become `+ len(skip_attrs)`. The code is copied from a function which skips at most two attributes (x and y).

VesnaT:
> Annotations will almost never be visible, I suppose. 

Why?

Orange/widgets/utils/__init__.py
VesnaT: Come on :) I just did this for the lint's sake.

Orange/widgets/utils/__init__.py
VesnaT: Same as above..

Orange/widgets/utils/__init__.py
VesnaT: Same as above :)

Orange/widgets/utils/__init__.py
VesnaT: True.

janezd:
> > Annotations will almost never be visible, I suppose.
> 
> Why?

I meant, they will almost always overlap.

Orange/widgets/utils/__init__.py
janezd: I know it was for lint. I just wanted to improve the improvement.

Same below. And below. :)

VesnaT:
Almost never. They are displayed vertically and one can set the font size.
I added a `None` option, though.

janezd:
Even with 100 instances and 10 px labels, this already spreads to > 1000 pixels. But maybe this widget won't be used liked that -- I confess I don't see it's actual use (but it's just me).

Adding None suffices for me. :)

lanzagar:
> Even with 100 instances and 10 px labels, this already spreads to > 1000 pixels. But maybe this widget won't be used liked that -- I confess I don't see it's actual use (but it's just me).
> 
> Adding None suffices for me. :)

We were talking about problems when trying to visualize too many bars and how to prevent that. The first idea was to limit it to ~50 rows of data. Then after seeing that the actual bar plot works quite well even for 150 instances, we agreed to limit it to 200 instead. Obviously annotations become a problem in these cases. Since it is now easy to disable them and even to adjust the font sizes I think we can leave the limit at 200, but I would be fine with reducing it to 100 as well.

The main motivation for this widget was to show 3-20 values with a bar chart since that was a common request by people using Orange, who where then forced to use a different tool to produce bar charts of Orange's results. For that scope the annotations should not be a problem and are probably an important feature to have.

janezd:
Alright then. As I said, I didn't know how this was intended to be used.

PR #4926, opened 2020-07-30
Fix tests that fails because of changes in pyqtgraph

PrimozGodec:
##### Issue

- Miscellaneous workflow started to fail because of changes in the latest pyqtgraph (https://github.com/pyqtgraph/pyqtgraph/pull/1276/files).
- There is also an issue with ArrowItem which is fixed by @ales-erjavec in https://github.com/pyqtgraph/pyqtgraph/pull/1336

##### Description of changes
- Reset "mouseRateLimit" for unittests.

##### Includes
- [ ] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/tests/test_owrocanalysis.py
+++ b/Orange/widgets/evaluate/tests/test_owrocanalysis.py
@@ -4,7 +4,7 @@ import unittest
 from unittest.mock import patch
 import copy
 import numpy as np
-
+import pyqtgraph as pg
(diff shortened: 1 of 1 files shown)


PrimozGodec:
The test will still fail until pyqtgraph/pyqtgraph#1336 will be merged

janezd:
... and these sagas will continue until @ales-erjavec some day has enough and writes AnyPyQtGraph. :)

Issue #4914, opened 2020-07-21
Edit Domain - reset of merged values


che9900:
Orange 3.26 / using new EDITDOMAIN

What was the task: group average_price into 3 segments: LOW, MEDIUM, HIGH. 

Ideally this segmentation would be based on the pre-defined values (I tried using DISCRETIZE, but that only offers pre-defined equal frequency / equal width / and some additional options)

I selected multiple values and using [M] key merged the values and added the label LOW. When merging a second group of values (for MED) the first merge (LOW) is removed. The only way at the moment to have multiple merged groups is a manual process for each numerical value. With larger data sets this is obviously not workable solution.

Request: adding a new merge does not remove all other merges/ labelings

ajdapretnar:
You are right, it does remove them. However, I think these are two separate issues.

1.) Provide custom values in Discretize.

2.) Enable multiple merging in Edit Domain (the problem here is to distinguish between different merges).

PR #4929, opened 2020-07-31
[ENH] Discretize: Manual cut points entry

ales-erjavec:
##### Issue

Ref: https://github.com/biolab/orange3/issues/4914#issuecomment-662296325 (1)

##### Description of changes

Provide custom values input in Discretize widget.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


ajdapretnar:
I really like this, but I noticed a couple of things.

1. Manual entry works beautifully, but it would probably be nice to have a tooltip? Something like: "Separate cut-offs with a comma, use point decimal separator." Or perhaps simply: "i.e. 0.5, 1.5, 2.5". 🤷 

2. For some reason, I cannot see the tooltip on CC. Even reading it, I am not sure what it does... 🤔 I assumed it copies manual cut-offs from single variable field to the Manual field at the top. If so, it doesn't work for me.

3. Perhaps not a part of this PR, but I have set two cut-offs, i.e. 1.5 and 2.5. There are no values that fall between 1.5 and 2.5. I then try to find them and merge with Edit Domain (merge values option), but nothing happens. Probably there's a bug somewhere. @PrimozGodec

ales-erjavec:
> For some reason, I cannot see the tooltip on CC

Fixed

> I assumed it copies manual cut-offs from single variable field to the Manual field at the top. If so, it doesn't work for me.

It copies the current (non-manual) cut points to the edit field and at the same time activates the Manual option.

janezd:
Nice.

How difficult would it be to somehow indicate an invalid input? I keep forgetting commas. It would be nice to also indicate that the order is not increasing.

Are commas really necessary? Why not also accepting "1 2 3"?

In case of invalid input (e.g. in default), the output is "single_value". Any chance of renaming this to "invalid_cut_of_points"?

In Iris, I set Entropy-MDL as default, then choose sepal width and click CC. I get "2.95, 3.3499999999999996". It would be nicer to have "2.95, 3.35", which are also the values shown in the list view on the left.

Also, when I click Cc, the line edit does not refresh, but this is probably the problem with my Qt -- it does repaint if I hide and show the widget.

PR #4958, opened 2020-08-27
[ENH] Add an option to set intercept in linear regression to 0

### LINKED ISSUE #4954: Force regression through origin (zero)
**relevant info**
macOS 10.13.6, orange 3.26.0 from Anaconda distribution (conda 4.8.4).

No.

**Here, write a clear and concise description of what the problem is. Ex. I'm always frustrated when [...]**
When using the Canvas, Model > Linear Regression and Educational > Polynomial Regression do not allow forcing of the regression through 0, i.e. Coefficients always include intercept and independent variable terms.

Add GUI elements (e.g., checkbox) to Linear Regression and Polynomial Regression widgets that allow the user to specify the forcing of the regression through the origin.

(Separately, perhaps add a Model > Polynomial Regression feature separate from the Educational add-on?)

**Add a clear and concise description of what you want to happen here**.
Return regression coefficients so fit `y = m_1x^n + m_2x^{n-1} + ... + m_nx + b` is calculated such that `b = 0`. (May only be feasible/valid for polynomials of first order.)

n/a

**Additional context**
n/a

Thanks! (And apologies if this feature exists. However, a search of the GUI and documentation was not fruitful.)

janezd:
Fixes #4954.

For polynomial regression in Educational add-on, I opened an issue there.

### DIFF
--- a/Orange/regression/linear.py
+++ b/Orange/regression/linear.py
@@ -41,6 +41,7 @@ class LinearRegressionLearner(SklLearner, _FeatureScorerMixin):
 class RidgeRegressionLearner(LinearRegressionLearner):
     __wraps__ = skl_linear_model.Ridge
 
+    # Arguments are needed for signatures, pylint: disable=unused-argument
     def __init__(self, alpha=1.0, fit_intercept=True,
                  normalize=False, copy_X=True, max_iter=None,
                  tol=0.001, solver='auto', preprocessors=None):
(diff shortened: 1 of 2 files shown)


Orange/regression/linear.py
```diff
@@ -140,6 +147,7 @@ def __str__(self):
 
 class PolynomialModel(Model):
     def __init__(self, model, polyfeatures):
+        super().__init__()
```
janezd, line 150: This change is suggested by lint but otherwise unrelated to the issue.

PR #4961, opened 2020-08-28
[MNT] Reduce the number of warnings in tests

janezd:
##### Issue

Tests result in too many warnings, so we overlook pending changes in dependencies.

##### Description of changes

Silencing some, fixing problems, replacing deprecated attributes...


janezd:
Why does this fail on Windows?! 

 TypeError: CCompiler_spawn() got an unexpected keyword argument 'env'`

janezd:
Most warnings are now resolved and the unittest output is reduced to just above 200 lines. (It will drop some more after the PR in widget-base is merged.)

The prevalent remaining warning is `RuntimeWarning: sys.excepthook was already patched (is <function pyqt5_qabort_override at 0x7f3fe0ac1830>)(just thought you should know this)`. @ales-erjavec, if it's important: can you resolve it? If not: can I remove it?

Orange/data/io.py
```diff
@@ -300,7 +300,13 @@ class ExcelReader(_BaseExcelReader):
     @property
     def workbook(self) -> openpyxl.Workbook:
         if not self._workbook:
-            self._workbook = openpyxl.load_workbook(self.filename,
+            with warnings.catch_warnings():
+                # We don't care about extensions, but we hate warnings
```
janezd, line 304: These are not file extensions but extensions to .xlsx format. Some internal MS stuff.

ales-erjavec:
> The prevalent remaining warning is `RuntimeWarning: sys.excepthook was already patched (is <function pyqt5_qabort_override at 0x7f3fe0ac1830>)(just thought you should know this)`. @ales-erjavec, if it's important: can you resolve it? If not: can I remove it?

It can be removed.

Orange/widgets/visualize/ownomogram.py
```diff
@@ -1251,7 +1254,11 @@ def get_points_from_coeffs(current_value, coefficients, possible_values):
                sorted_coefficients[i] * sorted_values[i] * (1 - k)
 
     def reset_settings(self):
-        self._reset_settings()
+        with warnings.catch_warnings():
+            # setting target_class_index will trigger this innocent warning
+            warnings.filterwarnings(
+                "ignore", "combo box 'target_class_index' is empty")
+            self._reset_settings()
```
janezd, line 1261: @markotoplak, the warning about invalid index is ignored here.

Nomogram is one of the few widgets that implements `reset_settings`. It sets the settings to default values. If there is no data, the default target class, 0, is invalid, but it's OK and harmless. I don't see a way around it.

If I haven't missed anything, this is the only place where this warning is ignored instead of being fixed.
```suggestion
            self._reset_settings()
```
```suggestion
            self._reset_settings()
```

PR #4971, opened 2020-09-04
TreeLearner: build_tree to private and add docs

ajdapretnar:
##### Issue

`build_tree()` had not reason to be public. It was misleading for users finding it in the documentation and thinking it prints the tree.
Also no documentation explaining how to print a tree.

##### Description of changes
Make `build_tree()` private.
Add an example of how to print a tree.

##### Includes
- [X] Code changes
- [ ] Tests
- [X] Documentation


janezd:
Could you also rename the method in `Orange.regression.tree`?

Issue #4952, opened 2020-08-21
Tests crash miserably with pyqt 5.15


markotoplak:
On master and on all tested operating systems; see #4945.

This is not anything new. On my linux machine, I can also reproduce it with orange-widget-base 4.5.0 and Orange 3.25.1.

Here is the output or `python -m unittest Orange.tests Orange.widgets.tests` with the current masters of oragne-widget-base and orange3.

```
Release of profile requested but WebEnginePage still not deleted. Expect troubles !
Release of profile requested but WebEnginePage still not deleted. Expect troubles !
Release of profile requested but WebEnginePage still not deleted. Expect troubles !
Release of profile requested but WebEnginePage still not deleted. Expect troubles !
Release of profile requested but WebEnginePage still not deleted. Expect troubles !
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
QGuiApplication::font(): no QGuiApplication instance and no application font set.
[...]

PR #4974, opened 2020-09-07
[FIX] Tests: PyQt5 5.15 compatible cleanup

ales-erjavec:
##### Issue

Ref gh-4952

##### Description of changes

* test_owpythonscript: Clear sys.last_{type,value,traceback} 
* test_gui: Remove widget in tearDown
*  test_owscatterplotbase: Reset Mock instances in class namespace 
* owpythonscript: Change shared_namespace to WeakKeyDictionary 
* owsom: Remove back ref to widget from closures in background thread 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpythonscript.py
+++ b/Orange/widgets/data/owpythonscript.py
@@ -4,8 +4,8 @@ import code
 import keyword
 import itertools
 import unicodedata
+import weakref
 from functools import reduce
-from collections import defaultdict
 from unittest.mock import patch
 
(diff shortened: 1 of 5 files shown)


markotoplak:
Thank you! I especially like that Python Script will not leak memory anymore.

@janezd, see how #3840 was supposed to be done. Yes, I did not understand the implications either, I merged #3840.

PR #4975, opened 2020-09-07
Tidy up issue templates

irgolic:
It seems that sometimes people just ignore the issue template and append their request to the end. This change declutters the templates and reworks the feature request template a bit.

##### Description of changes
- Turn all prompts into questions
- Add space in-between prompts
- Move additional information to HTML comments

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,12 +7,25 @@ assignees: ''
 
 ---
 
-**Bug reports that do not sufficiently describe a bug are not helpful and will most likely be ignored.**
+<!-- Thanks for taking the time to report a bug! -->
+<!-- To fix the bug, we need to be able to reproduce it, so please let us know the following... -->
 
-- [ ] **A clear and concise description.** "*I'm getting an error when I open a tree viewer*" is far from enough (what kind of error? what kind of data you give it?). "*I'm getting an error*" is even worse.
+- [ ] **What's wrong?**
+<!-- Be specific, clear, and concise. Include screenshots if relevant. -->
+<!-- If you're getting an error message, copy it, and enclose it with three backticks (```). -->
 
-- [ ] **How to reproduce the problem.** Upload a zip with a workflow and data, or describe the steps (open this widget, click there, then add this ...)
 
-- [ ] **Which version of Orange are you using? How did you install it?**
+
+- [ ] **How can we reproduce the problem?** 
+<!-- Upload a zip with the .ows file and data. -->
+<!-- Describe the steps (open this widget, click there, then add this...) -->
+
+
+
+- [ ] **Which version of Orange are you using? How did you install Orange?**
+
+
 
 - [ ] **What is your operating system?**
+
+
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -7,16 +7,21 @@ assignees: ''
 
 ---
 
-**Is your feature request related to a problem? Please describe.**
+<!-- Thanks for taking the time to submit a feature request! -->
+<!-- For the best chance at our team considering your request, let us know the following... -->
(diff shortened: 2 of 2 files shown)


ajdapretnar:
That's nice, but I don't think the problem lies in the templates, but in the people not bothering to read/use them.

markotoplak:
@irgolic, thanks, I really like how respectfully you come across. I hope this will motivate users to work a tiny bit harder to clearly describe what is happening.

janezd:
I agree with @markotoplak, it's really nicely written. But I even more agree with @ajdapretnar. It angers me to see people who use free software and ask us to spend time fixing bugs and fulfiling their wishes, but they themselves don't bother to read a few lines of simple instructions.

I'm not very hopeful that these changes, though so nicely composed, will help.

borondics:
Maybe using a form for issue submission that would check if all required information is filled in (even on another site if this is not possible with GitHub, which I think is the case) would be a way to force a uniform issue/bug submission. Of course, this would not prevent badly described problems. 

We used to do something similar with Quasar/Orange Spectroscopy.

markotoplak:
Sure, a minority thinks they deserve to have their bugs fixed immediately.

But mostly, I think people are doing us a service when they post a bug. They did free testing for us. And to them, Orange seemed important enough that they did the (non-trivial, for them) work needed to report the bug. They needed to find how to report the bug, maybe even had to create a GitHub account, and then actually write it up. Many have no experience in programming and so can not even imagine what it means to replicate a bug.

Even if their report is vastly insufficient, they at least point out that a component is failing on some (probably unusual) data. Or they point out that we buried a feature somewhere where they could not find it. All of this is a signal. So I can only respect the work they put in.

ajdapretnar:
They've put in all this work yet cannot spend two minutes reading a template?

You say people are doing us a service when they post a bug. True. But we are also doing them a service by providing a free software in the first place. Fixing a bug means they can continue using the software they way they want to. It is a mutual process which requires mutual respect, a lot of which can be shown by reading the actual template. Even if they do not know what replicating a bug means, they can ask or write that they don't know how to do it. An honestly filled out template is better than one that is ignored.

janezd:
> I think people are doing us a service when they post a bug.

Does somebody do you a favour by telling you what to do for him?

"Let me do you a favour: I'm telling you that I'm thirsty." (And then the other person needs to discover what the thirsty one wants to drink, because (s)he didn't bother to read the instructions.)

I understand @markotoplak's point. In commercial software, testing is supposed to be done by developers, while in open source users are involved in the process. But the one who profits in this relation are still just users, not developers, so the relation is asymmetric: only developers do favours.

markotoplak:
> > I think people are doing us a service when they post a bug.
> 
> Does somebody do you a favour by telling you what to do for him?

Depends on what you are trying to achieve. If I aim to build robust and usable software, then absolutely yes.

borondics:
I think this should be a conversation with _at least_ one beer for everyone. (I can only join on Zoom if needed... :)) And even then there might be no solution.

@janezd, I think there is a flaw in your logic. The relationship is surely asymmetric because usually the user benefits from the properly working features a lot more than the developer from the bug report. But do you know of any open source project that is only for users and developers don't use the product at all? In that case a bug report would indeed help only the users. So, if we assume that bugs are not implemented on purpose, the user still does a favour by pointing them out before the developer finds and fixes them.

irgolic:
I'm hesitant to blame users, I think it's a UX issue.

Sure, there's a small minority of self-entitled users. In my experience, this minority is tiny. Most people are incredibly thankful for any response they receive.

Orange's target audience is non-tech-savvy. If someone like that takes the time to create an account on GitHub, it's that they don't *want* to take the time to submit a good issue, it's that it's hard for them. 

Using a new platform like GitHub already feels overwhelming for users. The most we can do for them is present them with an intuitive template, to lessen their cognitive load. If this updated template still doesn't help, I have the idea of adding even more space between the questions, to make it even more obvious where the answers have to be written.

PR #4977, opened 2020-09-09
[ENH] OWColor: Saving and loading color schemata

### LINKED ISSUE #4890: Loading pre-defined colour schemes
- Loading pre-defined colour schemes (e.g. for categorical variables with many groups)
Applicability: General.
Importance: High. Sometimes multiple datasets have the same categorical variables. Colour widget could try to read colours from a HEX column. Then a single table of unique category-colour mapping entries could be merged with data sets that have the categorical variable with same categories to assign the colours (I am not sure if this can be currently done by Merge Data - e.g. if the Data has more rows with the same category to which to append the Extra Data columns). (KH)
    - **JD:** Could this be an enhancement of the Color widget?

janezd:
##### Issue

Resolves #4890.

##### Description of changes

I defined two file formats. One contains all changes done by the user, e.g.

```
{
    "categorical": {
        "gender": {
            "renamed_values": {
                "female": "women",
                "male": "man"
            }
        },
        "fasting blood sugar > 120": {
            "rename": "fasting blood sugar is high",
            "colors": {
                "0": "#21FF06",
                "1": "#996633"
            }
        },
        "exerc ind ang": {
            "colors": {
                "0": "#46BEFA",
                "1": "#996633"
            }
        },
        "slope peak exc ST": {
            "renamed_values": {},
            "colors": {
                "upsloping": "#FD8008",
                "flat": "#ED462F",
                "downsloping": "#AAF22B"
            }
        }
    },
    "numeric": {
        "rest SBP": {
            "colors": "linear_bgyw_20_98_c66"
        },
        "cholesterol": {
            "colors": "diverging_bwr_40_95_c42"
        },
    }
}
```

When loading the file, it's content does not have to exactly match the domain: redundant variables/values are ignored, and data for variables that does not appear in the file is retained as it is in the widget.

The second format contains general definitions of colors, like

```
{
    "female": "#FF0000",
    "1": "#996633",
    "upsloping": "#FD8008"
}
```

When saving, it stores the colors for values that were changed by the user, and only if the change if the same for all variables in the domain (e.g. if all "female" that are changed are changed to the same color). When loading, it changes colors for all values that appear in any variable.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcolor.py
+++ b/Orange/widgets/data/owcolor.py
@@ -63,9 +63,11 @@ class AttrDesc:
     @classmethod
     def from_dict(cls, var, data):
         desc = cls(var)
+        if not isinstance(data, dict):
+            raise InvalidFileFormat
         new_name = data.get("rename")
         if new_name is not None:
-            if not isinstance(desc.name, str):
+            if not isinstance(new_name, str):
                 raise InvalidFileFormat
             desc.name = new_name
         return desc, []
@@ -163,7 +165,7 @@ class DiscAttrDesc(AttrDesc):
                 if value in new_colors:
                     try:
                         color = hex_to_color(new_colors[value])
-                    except (IndexError, ValueError) as exc:
+                    except ValueError as exc:
                         raise InvalidFileFormat from exc
                     colors.append(color)
                 else:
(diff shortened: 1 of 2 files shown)


janezd:
Note to self: remove the "general" format (in a separate commit, so we still have the code in case we need it someday).

The remaining code implements what the issue was about. The second format is not very useful, except as a workaround for copying colors assigned to one variable to all other variables. A better solution for this is to add "Copy to all" to discrete variables, but so far nobody needed it, so we won' t implement the feature, until somebody wants it badly enough.

ajdapretnar:
I tried testing this but I can't seem to get a working file. I tried the first file in the first comment. What am I doing wrong?

janezd:
I may have changed the format (though I think I haven't). But why would you use the file from the comment? You can just change some colors and names, and save the file.

PR #4985, opened 2020-09-14
Impute: Prevent overfitting (and exploding) in default model imputation

### LINKED ISSUE #4982: Will not impute data. Keeps crashing my computers.
##### Orange version

3.26

##### Expected behavior
Impute Data 

##### Actual behavior
It's getting stuck at various percentages while trying to impute the data, mostly in the 80-90% range, freezing, and I must do a forced shutdown for my computer to work. It actually makes my screen go black and I can't do anything. I also tried downloading Orange on another computer of mine but its doing the same thing and it crashed that computer as well.   

Select a file with a large data set with missing variables...this data set has 1302 instances and 19 features. 

image

janezd:
##### Issue

Fixes #4982.

Model-based imputation uses `SimpleTree` with default arguments, which are 2 instances in a leaf and (essentially) unlimited depth. If data includes enough numeric attributes, this led to memory overflow.

##### Description of changes

Cut trees to 10 levels with minimum of 10 instances in leaves.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/classification/simple_tree.py
+++ b/Orange/classification/simple_tree.py
@@ -78,7 +78,7 @@ class SimpleTreeLearner(Learner):
 
     name = 'simple tree'
 
-    def __init__(self, min_instances=2, max_depth=1024, max_majority=1.0,
+    def __init__(self, min_instances=2, max_depth=32, max_majority=0.95,
                  skip_prob=0.0, bootstrap=False, seed=42):
         super().__init__()
         self.min_instances = min_instances
--- a/Orange/widgets/data/owimpute.py
+++ b/Orange/widgets/data/owimpute.py
@@ -162,7 +162,7 @@ class OWImpute(OWWidget):
         super().__init__()
         self.data = None  # type: Optional[Orange.data.Table]
         self.learner = None  # type: Optional[Learner]
-        self.default_learner = SimpleTreeLearner()
+        self.default_learner = SimpleTreeLearner(min_instances=10, max_depth=10)
         self.modified = False
         self.executor = qconcurrent.ThreadExecutor(self)
         self.__task = None


lanzagar:
I have a question and a suggestion, but we can move this to a new issue any time and merge this - I think the new values are definitely an improvement.

Question:
How did the current defaults lead to a memory overflow on data with 1302 instances? Wouldn't the worst case be to split off 2 instances every time and end up with 1302 nodes (half of that leaves and half internal nodes)? Does not sound like that takes up a lot of memory...

Suggestion:
I agree the current defaults are very weird and propose we change them for SimpleTreeLearner as well, not just in Impute's call. I don't think Simple Trees are available directly from any widget (otherwise we would probably notice this earlier), but SimpleRandomForest constructs the trees with the same defaults and I don't think we should have default parameters for two methods that crash the computer on a relatively small dataset.
I think
- `max_majority` could at least be 0.99 or even 0.95
- `max_depth` seems pretty useless at 1024 - we can be permissive with 128, or more cautious with something like 16
- `min_instances` is the one I would be most careful changing. I think unbalanced classes could have problems with this set to e.g. 10 (if only 5% of samples are positive, you probably want to allow splitting off 5 instances into a leaf). If the above is enough I would consider leaving this at 2 or increase it to 5 at most?

janezd:
Random Forest that is exposed in the widget uses an implementation from scikit-learn. SimpleTreeLearner is used only for imputation and in SimpleRandomForest, which is not accessible from a widget.

Regarding question:
you're right, it shouldn't take so much memory. It is interesting, though, that there seem to be no leak. Given enough time, it works and, on the first glance, it frees memory.

Regarding suggestion:
I agree. But I'd still also keep the arguments in imputation.

My suggestion: What about removing SimpleTreeLearner and SimpleRandomForest altogether? Nobody will notice, I guess. Debugging simple tree learner doesn't look appealing. Imputation can use the ordinary TreeLearner.

lanzagar:
> My suggestion: What about removing SimpleTreeLearner and SimpleRandomForest altogether? Nobody will notice, I guess. Debugging simple tree learner doesn't look appealing. Imputation can use the ordinary TreeLearner.

I thought about this too. If I remember correctly, the motivation for our own simple trees and RF was speed compared to scikit-learn. We should probably decide if we want to remove it, or make it more accessible and advertise its speed benefits better.

janezd:
Decision (with @lanzagar): remove the unmaintained `SimpleTreeLearner` and `SimpleRandomForest`. Imputation will use `TreeLearner` (with proper parameters) instead.

PR #4990, opened 2020-09-17
Mosaic Vizrank with fixed number of variables

janezd:
##### Issue

Resolves https://github.com/biolab/orange-widget-base/issues/87.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


ajdapretnar:
I like this. It's not a bad solution. Not at all. :)

janezd:
@ajdapretnar, I've added some tests. If you clicked it through and it works (pay attention to the situation when the widget user switches to Pearson correlation, or it is switched automatically because there is not class; in this case, the minimal number of attributes is 2, never 1), you can merge.

PR #4993, opened 2020-09-18
owtestandscore: Change default k (folds) from 10 to 5

irgolic:
##### Description of changes
The choice of k is arbitrary, so I suggest going with a lower one for faster default performance. @janezd?

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owtestandscore.py
+++ b/Orange/widgets/evaluate/owtestandscore.py
@@ -169,7 +169,7 @@ class OWTestAndScore(OWWidget):
     #: Selected resampling type
     resampling = settings.Setting(0)
     #: Number of folds for K-fold cross validation
-    n_folds = settings.Setting(3)
+    n_folds = settings.Setting(2)
     #: Stratified sampling for K-fold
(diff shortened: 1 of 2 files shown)


janezd:
I don't have a strong opinion.

- I always assumed that 10 is the most common number of folds.
- Search for 10-fold cross validation gives 6.4M hits on Google, while 5-fold gives 1.5M. (This is more than 4-fold difference. 10-fold also seems faster: it took Google 0.5s to search for 10-fold and 0.7s for 5-fold. :)).
- One argument for 10 may be that existing teaching materials typically keeps the default, so keeping it as it is will retain the consistency of screenshots as well some of reported results (in case of fixed seeds and stable methods). But I haven't checked this.
- Scikit uses 5 as default, though.

I'd pass this to @BlazZupan, or to a discussion next Friday.

ajdapretnar:
I saw 10 in most papers. 🤷 But it is strange sklean has 5 - we normally follow their defaults.

PR #4994, opened 2020-09-21
Server Embedder: remove warnings

PrimozGodec:
##### Issue

Server embedded did not close the event loop on exception. 

##### Description of changes
Closing the event loop on exception - this removes warnings

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/misc/server_embedder.py
+++ b/Orange/misc/server_embedder.py
@@ -117,9 +117,13 @@ class ServerEmbedderCommunicator:
 
         loop = asyncio.new_event_loop()
         asyncio.set_event_loop(loop)
-        embeddings = asyncio.get_event_loop().run_until_complete(
-            self.embedd_batch(data, processed_callback)
-        )
+        try:
+            embeddings = asyncio.get_event_loop().run_until_complete(
+                self.embedd_batch(data, processed_callback)
+            )
+        except Exception:
+            loop.close()
+            raise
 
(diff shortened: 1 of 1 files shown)


Orange/misc/server_embedder.py
```diff
@@ -117,9 +117,13 @@ def embedd_data(
...
-        embeddings = asyncio.get_event_loop().run_until_complete(
-            self.embedd_batch(data, processed_callback)
-        )
+        try:
+            embeddings = asyncio.get_event_loop().run_until_complete(
+                self.embedd_batch(data, processed_callback)
+            )
+        except Exception as ex:
+            loop.close()
+            raise ex
```
janezd, line 126: Shouldn't this be just `raise`, not `raise ex`? Pure `raise` re-raises the exception and keeps the original traceback. (In this case, of course, also remove `as ex`.)

While at it, you can also make changes to exception handling in method `_send_request`:

- change `raise EmbeddingConnectionError` to `raise EmbeddingConnectionError from ex` (and add `as ex` a few lines earlier); I believe this would be appropriate in both `raise` clauses;
- change `raise e` to `raise`, because this is just re-raise after logging.

PrimozGodec: Thank you for these suggestions. Now it is changed.

PR #4997, opened 2020-09-22
[ENH] Neighbors: improve exclusion of references, checkbox to (un)limit output data

borondics:
##### Issue
OWNeighbors was limited to output distances only a subset of the input data. This change enables outputting distances to the full dataset.

##### Description of changes
For Orange Spectroscopy we would like to be able to compare the full dataset to a selected reference instance and visualize the distances to it.

##### TODO:
The widget needs some work in my opinion, right now, every instance that is in an arbitrarily defined 1E-5 distance from the reference is excluded from the output. This should be handled dynamically based on the values of the data instances.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owneighbors.py
+++ b/Orange/widgets/data/owneighbors.py
@@ -41,6 +41,11 @@ class OWNeighbors(OWWidget):
     class Outputs:
         data = Output("Neighbors", Table)
 
+    class Info(OWWidget.Warning):
+        removed_references = \
+            Msg("Input data includes reference instance(s).\n"
+                "Reference instances are excluded from the output.")
+
     class Warning(OWWidget.Warning):
         all_data_as_reference = \
             Msg("Every data instance is same as some reference")
@@ -52,8 +57,8 @@ class OWNeighbors(OWWidget):
     distance_index: int
 
     n_neighbors = Setting(10)
+    limit_neighbors = Setting(True)
     distance_index = Setting(0)
-    exclude_reference = Setting(True)
     auto_apply = Setting(True)
 
     want_main_area = False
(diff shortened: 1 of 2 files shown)


janezd:
If the user wants to have all instances, (s)he now has to limit it to a high-enough number. Wouldn't it be more elegant to just put a checkbox in front of the spin?

As for 1e-5, what about this: if there are some instances that are (almost) equal to the reference instance, we would expect the distribution of distances to have a small group of values below 1e-5 (I think this is still reasonably above numeric errors) and then nothing for a long time (but perhaps not all the way to 1e-5). 

On the other hand, we may not need to go through all this: values are normalized, hence 1e-5 is already dynamic in a sense. Have you encountered cases when it removed instances that you'd prefer kept?

I wouldn't expose 1e-5 in the user interface. Beside it looking to technical and taking too much space, the number is meaningless to the user because variables are normalized. I guess.

borondics:
In the current master the number of neighbours on the output was limited to 100, which could be dynamically changed to the number of instances to allow what you suggest, but I don't find it more usable to have to look up the number of instances and to have to type it into the spin box. The arrows would be pretty useless in some cases too as we can easily get a >100000 instances in Spectroscopy.

Also, not having the automatic maximum output would limit the reusability of saved workflows with different datasets as the user would need to look up the size of the dataset each time and update the spinbox when loading new data.

Typing accuracy would limit the usability as well (in case of 124681 instances it is easy to make a typo).

The Jaccard distance always (? needs more testing) gives me the "all instances are the same" message for spectroscopy data. I haven't looked into the actual calculation though. I put 1E-5 to the GUI hoping that the hardcoded comparison will be removed in the new version of the calculation. 

If the values are normalized we should add it as an option to turn it off.

janezd:
> I did put a checkbox in front of the spin

I must be blind. Sorry. .)

The ridiculously high number could be replaced (there's no reason not to...), though it won't change anything.

Jaccard is a different kind of distance, not applicable to spectroscopy. It computes differences between sets.

Discussion results:

- The checkbox for removal of similar instances was a workaround for a bug. We just want to exclude the reference example if it appears in the data. The bug has been fixed, so the plan now is to remove the checkbox entirely. If the user wants to remove instances that are too similar, (s)he can use "Select Rows".
- For normalization, there should be a checkbox similar to the one in Distances.

Would this be OK?

Would you continue working on this or should somebody else take over?

borondics:
@janezd, thanks for the update! 

I'd be happy with any solution as long as we can output the full incoming dataset. For the normalization, it would be great to have it as an option.

I think you should take over because I just saw @BlazZupan just reported something in #5000 that I'm not sure I can fix.

markotoplak:
@janezd, this looks good. The only question is whether we need to keep backward compatibility for this one. As this is more like a bugfix (and the changes would be slight), I'd say there is no need to make a "compatibility mode". What do you think?

janezd:
I think this is too trivial for complications with "compatibility mode". It would be just dead, garbage code.

PR #5013, opened 2020-10-02
Use buttonsArea, UI cleanup

irgolic:
To be merged after https://github.com/biolab/orange-widget-base/pull/130 is merged and released.

##### Description of changes
This PR moves some auto_commit buttons, and data projection controls, from controlArea to buttonsArea. With the widget-base PR, this is displayed below a scrollable controlArea, making it always visible.

Also, widget UIs are redesigned to be more consistent. All widgets except based only in mainArea were moved to controlArea (except Venn Diagram).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpreprocess.py
+++ b/Orange/widgets/data/owpreprocess.py
@@ -1107,10 +1107,20 @@ class OWPreprocess(widget.OWWidget, openclass=True):
         box = gui.vBox(self.controlArea, "Preprocessors")
         gui.rubber(self.controlArea)
 
+        # we define a class that lets us set the vertical sizeHint
+        # based on the height and number of items in the list
+        # see self.__update_list_sizeHint
+
         class ListView(QListView):
+            def __init__(self, *args, **kwargs):
+                super().__init__(*args, **kwargs)
+                self.vertical_hint = None
+
             def sizeHint(self):
                 sh = super().sizeHint()
-                return QSize(sh.width(), 220)
+                if self.vertical_hint:
+                    return QSize(sh.width(), self.vertical_hint)
+                return sh
 
         self.preprocessorsView = view = ListView(
             selectionMode=QListView.SingleSelection,
@@ -1169,6 +1179,8 @@ class OWPreprocess(widget.OWWidget, openclass=True):
                           Qt.ItemIsDragEnabled)
             self.preprocessors.appendRow([item])
 
+        self.__update_list_sizeHint()
+
         model = self.load(self.storedsettings)
 
         self.set_model(model)
@@ -1180,6 +1192,14 @@ class OWPreprocess(widget.OWWidget, openclass=True):
             self.__update_size_constraint()
 
         self.apply()
+
+    def __update_list_sizeHint(self):
+        view = self.preprocessorsView
+
+        h = view.sizeHintForRow(0)
+        n = self.preprocessors.rowCount()
+        view.vertical_hint = n * h + 2  # only on Mac?
+        view.updateGeometry()
 
     def load(self, saved):
         """Load a preprocessor list from a dict."""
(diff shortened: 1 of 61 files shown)


irgolic:
I only went over widgets who explicitly override `buttons_area_orientation`. However, that option defaulted to `Qt.Horizontal`, so I'll take a closer look to see if there's other widgets which should be fixed. The way biolab/orange-widget-base#102 deprecates the GUI layout setting though, it should be fine.

Some mini issues:

OWSilhouettePlot, OWRank: why is controlArea slightly smaller in width than buttonsArea (on Mac)? When I try to debug this by setting a `'background: red;'` stylesheet, the problem disappears. What a heisenbug.

OWContinuize: After moving it down, the 'Apply Automatically' button is gigantic. Moving it below the options is definitely better, but its width needs to be constrained. Perhaps a max width should be set if `want_main_area` is False?

irgolic:
Due to https://github.com/biolab/orange-widget-base/pull/127, this will need to be adjusted to retain the `gui.rubber`s in the controlArea.

Orange/widgets/data/owpreprocess.py
```diff
@@ -1142,8 +1154,8 @@ def mimeData(indexlist):
         self.mainArea.layout().addWidget(self.scroll_area)
         self.flow_view.installEventFilter(self)
 
-        box = gui.vBox(self.controlArea, "Output")
-        gui.auto_apply(box, self, "autocommit", box=False)
+        # autocommit is not in buttonsArea to allow list view stretch
```
markotoplak, line 1157: Is this comment still relevant?

Orange/widgets/data/owpreprocess.py
```diff
@@ -1105,8 +1105,20 @@ def mimeData(indexlist):
         self.preprocessors.mimeData = mimeData
 
         box = gui.vBox(self.controlArea, "Preprocessors")
+        gui.rubber(self.controlArea)
 
-        self.preprocessorsView = view = QListView(
+        class ListView(QListView):
```
markotoplak, line 1110: Why is this needed?

Orange/widgets/gui.py
```diff
@@ -34,6 +34,7 @@
     CalendarWidgetWithTime, DateTimeEditWCalendarTime,
     ControlGetter, VerticalScrollArea, ProgressBar,
     ControlledCallback, ControlledCallFront, ValueCallback, connectControl,
+    is_macstyle
```
markotoplak, line 37: Isn't `Orange/widget/gui.py` only for backward compatibility and Orange-specific additions? If so, the new `is_macstyle` from `widget-base` should not be added here.

Orange/widgets/model/owloadmodel.py
```diff
@@ -30,14 +30,15 @@ class Error(widget.OWWidget.Error):
...
+    buttons_area_orientation = None
     resizing_enabled = False
 
     def __init__(self):
         super().__init__()
         RecentPathsWComboMixin.__init__(self)
         self.loaded_file = ""
 
-        vbox = gui.vBox(self.controlArea, "File", addSpace=True)
+        vbox = gui.vBox(self.controlArea, "File", addSpace=False)
```
markotoplak, line 41: `addSpace` was deprecated thus all uses need to be removed.

Orange/widgets/model/owrandomforest.py
```diff
@@ -52,12 +52,14 @@ def add_main_layout(self):
...
-            callback=self.settings_changed)
+            callback=self.settings_changed,
+            attribute=Qt.WA_LayoutUsesWidgetRect)
         self.weights = gui.checkBox(
             box, self,
             "class_weight", label="Balance class distribution",
             callback=self.settings_changed,
-            tooltip="Weigh classes inversely proportional to their frequencies."
+            tooltip="Weigh classes inversely proportional to their frequencies.",
+            attribute=Qt.WA_LayoutUsesWidgetRect
```
markotoplak, line 62: Could you also mention the `WA_LayoutUsesWidgetRect` trick in the style guide?

markotoplak:
@irgolic, thank you, looks good!

I skimmed went through the changes and I have some (very) minor comments.

Issue #4986, opened 2020-09-15
Unbalanced target variable


robertcv:
If I understand correctly there is no simple way to handle unbalanced target variables in Orange.
One possible solution would be to oversample but then we could end up having the same sample in the training and testing data. Another solution would be to subsample but I cannot use it as I, unfortunately, have a small dataset.
The solution I would propose is to expose `class_weight='balanced'` as an option in the GUI. This parameter can be used on some of the sklearn's models ([LogisticRegression](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html), [RandomForestClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html), [SVM](https://scikit-learn.org/stable/modules/generated/sklearn.svm.LinearSVC.html)) and solves the unbalance in the target variable.

janezd:
We add a checkbox for learners for which this is supported in scikit-learn.

We add warning that this can decrease performance.

janezd:
Fixed via #5026.

zy-tech:
I came to this issuse just now. 

for the unbalanced target variable, could we use random forest with some sampling trick: we use all the minorities in each tree, but we random sample majorities in each tree. So in each tree, we got the balanced data with majority subsample, but for the forest, it is a minority oversample.

janezd:
It is an idea. You can try it...

As for adding it to Orange, we're not to keen about adding experimental stuff. Experience taught us it often ends as dead code.

zy-tech:
I used to use this tech in KNIME, they provide this method in their "Ensemble Learner", I have tried and it works very well. You can have a try it in KNIME

PR #5026, opened 2020-10-09
RF and LogReg: add class_weight parameter

ajdapretnar:
##### Issue

Implements #4986.

##### Description of changes
Add checkbox to consider class weights in Logistic Regression and Random Forest.
Trigger Warning when checking the parameter.
Add tests.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/tests/test_owlogisticregression.py
+++ b/Orange/widgets/model/tests/test_owlogisticregression.py
@@ -112,3 +112,13 @@ class TestOWLogisticRegression(WidgetTest, WidgetLearnerTestMixin):
         self.send_signal("Data", table)
         coef2 = self.get_output("Coefficients")
         self.assertTrue(np.array_equal(coef1, coef2))
+
+    def test_class_weights(self):
+        table = Table("iris")
+        self.send_signal("Data", table)
+        self.assertFalse(self.widget.class_weight)
+        self.widget.controls.class_weight.setChecked(True)
+        self.assertTrue(self.widget.class_weight)
+        self.widget.apply_button.button.click()
+        self.assertEqual(self.widget.model.skl_model.class_weight, "balanced")
+        self.assertTrue(self.widget.Warning.class_weights_used.is_shown())
--- a/Orange/widgets/model/owlogisticregression.py
+++ b/Orange/widgets/model/owlogisticregression.py
@@ -8,6 +8,8 @@ from Orange.widgets import settings, gui
 from Orange.widgets.utils.owlearnerwidget import OWBaseLearner
 from Orange.widgets.utils.signals import Output
 from Orange.widgets.utils.widgetpreview import WidgetPreview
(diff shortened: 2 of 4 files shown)


ajdapretnar:
I don't like tests. They seem useless. Any suggestion?

janezd:
Tests test scikit-learn, not Orange: they test whether the flag has any effect (which happens in sklearn) and not whether the widget actually sets the flag.

Add a test to test_owlogisticregression and test_owrandomforest, where you set and unset the flag and check whether the learner on the output (don't give any data to the widget) has the `weight_class` argument properly set. You can also check whether the warning properly appears and disappears.

The tests that you have added should better be removed. Not only because they don't test our code, but also because they are a bit dangerous.

Second, the label at the checkbox, "Class weight" is not very informative, and the tooltip is a bit misleading -- classes are weighed according to the *inverse* proportion of frequencies. Also, for logistic regression, I'd prefer a separate box because the existing box is focused on regularization parameters.

Screenshot 2020-10-16 at 11 15 35

I experimented with UX changes, so I already have the code and I'll just push a commit. I'd ask you just to fix the tests.

janezd:
I'm so sorry, I accidentally rebased before pushing my commit. Please ask your local git expert how to pull your branch before adding or changing your commits. Squash at will.

ajdapretnar:
I did feel weird about those tests, but just checking a flag seemed... as if I haven't done anything. 😆 But you are right, I shouldn't be checking sklearn's code.

I also welcome the UX changes. My local git expert did his job well. ;)

janezd:
> just checking a flag seemed... as if I haven't done anything.

That's because your code hasn't done anything (almost). The code's job is to set the flag, and the test's job is to verify that the code does it's job. Reacting to the flag is up to sklearn, and testing the reaction is up to sklearn's tests.

Orange/widgets/model/tests/test_owlogisticregression.py
```diff
@@ -112,3 +112,13 @@ def test_target_with_nan(self):
...
+
+    def test_class_weights(self):
+        table = Table("iris")
+        self.send_signal("Data", table)
+        self.assertFalse(self.widget.class_weight)
+        self.widget.controls.class_weight.setChecked(True)
+        self.assertTrue(self.widget.class_weight)
+        self.widget.apply_button.button.click()
+        self.assertEqual(self.widget.model.skl_model.class_weight, "balanced")
+        self.assertTrue(self.widget.Warning.class_weights_used.is_shown())
```
janezd, line 124: I won't nitpick further, but the more important test, IMHO, is that the warning is cleared when necessary. We are good at raising them, but used to have problems clearing.

You also haven't tested that the argument is not set when the checkbox is unchecked. But let's assume that this but is unlikely, so I'm merging. :)

Issue #4918, opened 2020-07-23
'Drag and drop' on canvas


irgolic:
If a loadable file is dropped onto canvas, add a widget (File or Load Distances or Load Model) at that position and load the file.

It would be sensible to implement #3997 before this.

Related to #4917 in terms of reusable code.

janezd:
Approved. 👍 

Widget should declare, whilch file extensions they handle (if any).

borondics:
This would be another great addition!

@markotoplak are there any extra things we would need to do to enable this for Orange Spectroscopy readers like `Multifile` (and any other we might have later)?

Issue #3997, opened 2019-09-02
Merge File and CSV File Import widgets


irgolic:
**Is your feature request related to a problem? Please describe.**
In an effort to reduce the number of widgets, I believe the CSV File Import widget's functionality should be moved to the File widget. It is intuitive to use the File widget to load a CSV file.

**Describe the solution you'd like**
Should a CSV file be chosen in the File widget, the CSV import window opens in the same way it does currently in the CSV File Import widget.

irgolic:
The biggest issue here is that you can't pick a target attribute in CSV File Import, the user to know to use Select Columns.

zy-tech:
I find "csv import" is much faster than 'file" widget, why is that?

janezd:
When merging, keep File's RecentPaths (see #4784).

irgolic:
We'd like to keep CSV File Import, over File.
Missing functionality:
- Built-in domain editor
- URLs
- .xlsx (and other)

PR #5031, opened 2020-10-14
[ENH] Create nodes on canvas drag/drop

ales-erjavec:
##### Issue

Closes biolab/orange3#4918

##### Description of changes

Uses:
* biolab/orange-canvas-core#143 
* biolab/orange-widget-base#106

Implements drop handlers for 'File', 'Load Model', 'Distance File' and 'Python Script' widgets.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/mainwindow.py
+++ b/Orange/canvas/mainwindow.py
@@ -3,6 +3,7 @@ from AnyQt.QtWidgets import (
     QFormLayout, QCheckBox, QLineEdit, QWidget, QVBoxLayout, QLabel
 )
 from orangecanvas.application.settings import UserSettingsDialog, FormLayout
+from orangecanvas.document.interactions import PluginDropHandler
 from orangecanvas.document.usagestatistics import UsageStatistics
 from orangecanvas.utils.overlay import NotificationOverlay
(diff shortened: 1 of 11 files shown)


irgolic:
This is really cool, thanks. Changes like this go a long way, as minor as they may seem on the surface.

How hard would it be to support drag-and-dropping several files at a time to canvas?

As is right now though, most any file I try drag-and-dropping (.csv, .tab, .dst, .py) gives the option of using CSV File Import (alongside another). For .py, .dst, this shouldn't happen.
The point of this change is an intuitive way to load your data into Orange. But by dragging a file (.csv, .tab...) to canvas, a dropdown is (correctly) shown offering either File or CSV File Import – something totally confusing for a new Orange user, REALLY underscoring that #3997 needs to be fixed. I'd even wait with merging this before that's fixed.

Also, a nitpick, it gets laggy when holding a file over the window (this PR checked out on left, latest release on right)
ezgif-4-bc385bff6734
Any chance some caching could come into place for this to only lag the first time the file touches the window? (or not lag at all?)

PR #5032, opened 2020-10-15
Slow horizontal scrolling in `QTableView`s

irgolic:
##### Issue
Horizontal scrolling is too fast when unsynthesized (e.g., using a trackpad).
When scrolling with Shift/Alt+Scroll, it scrolls a page at a time. This seems to be intentional behavior, but I don't like it.

##### Description of changes
Unsynthesized horizontal scrolling is made as fast as vertical.
Shift/Alt+Scroll will scroll in steps, just like vertical.

This applies the change to Data Table, Predictions and Distance Matrix. Is there any other widgets this should be applied to?

##### Notes

Should I move the mixin somewhere else, like a util file? I wasn't sure where to put it.

Off-topic, why is scrolling so slow in table view's with a larger amount of visible cells? Frames per second already drop with 80 cells visible! Is it reloading all of the visible cells on every scroll event? If fetching from the model is the bottleneck, maybe a threaded solution is viable, where cells remain blank until their value is fetched?

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -363,7 +363,7 @@ def table_selection_to_list(table):
 TableSlot = namedtuple("TableSlot", ["input_id", "table", "summary", "view"])
 
 
-class TableView(QTableView):
+class TableView(gui.HScrollStepMixin, QTableView):
     #: Signal emitted when selection finished. It is not emitted during
     #: mouse drag selection updates.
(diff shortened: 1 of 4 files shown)


janezd:
This was one of those things that annoy you, but you subconsciously accept them without thinking about changing them. :)

PR #5033, opened 2020-10-16
[ENH] Create Instance: New widget

VesnaT:
##### Issue
Implement a widget for creating an instance, based on input data.

##### Description of changes
- new widget
image

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation


borondics:
Would it make sense to add the possibility of splitting the dataset according to categorical variables? Right now the category of the output instance can be selected by hand.

markotoplak:
@borondics, what do you mean? Is what you say actually connected to this widget (Create instance)? This widget can only create a single output instance, thus I do not understand what you meant by splitting.

borondics:
Yes, that way the widget would create one instance for each categorical class.

PR #5037, opened 2020-10-17
customizableplot.available_font_families: Fix for non-existent default family

### LINKED ISSUE #5036: `available_font_families` crashes if default font is missing
- [ ] **What's wrong?**

캡처

- [ ] **Which version of Orange are you using? How did you install Orange?**

- [ ] **What is your operating system?**

janezd:
##### Issue

Fixes #5036.

It seems that the default font doesn't always exist.

##### Description of changes

Check whether the default font family exists. If not, just return a sorted list of families, without pushing the default to the top.

 Also: tests for `available_font_family`.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/visualize/utils/customizableplot.py
+++ b/Orange/widgets/visualize/utils/customizableplot.py
@@ -28,8 +28,11 @@ def available_font_families() -> List:
     if not QApplication.instance():
         _ = QApplication(sys.argv)
     fonts = QFontDatabase().families()
-    default = fonts.pop(fonts.index(default_font_family()))
+    default = default_font_family()
+
     defaults = [default]
+    if default in fonts:
+        fonts.remove(default)
 
     guessed_name = default.split()[0]
     i = 0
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/utils/customizableplot.py
```diff
@@ -28,18 +28,23 @@ def available_font_families() -> List:
...
-            defaults.append(fonts.pop(i))
-        else:
-            i += 1
-    return FontList(defaults
-                    + [""]
+
+    top_section = []
+
+    default_family = default_font_family()
+    if default_family in fonts:
```
VesnaT, line 35: I think the `default_family` should be added to fontList even though not present in `fonts`. If not, the combobox may crash when trying to set its text to something missing in its model().

VesnaT (changes requested):

VesnaT (approved):

PR #5038, opened 2020-10-17
OWScatterPlotBase: Fixup selection UX

irgolic:
##### Description of changes
A couple of minor changes to keyboard shortcuts and the accompanying tooltip for selection.

I removed the 'Warning: Selection is applied to unjittered data' tooltip, I appreciate the gesture in explaining that to the user, but it doesn't seem like something somebody would complain to us about. I don't see why someone would want (or expect) to have jittered data selected. (also I can't think of a way to make it look nice)

I also wanted to make the tooltip show on click+hold, but this seemed more difficult/messier to do. The problem was in resetting its visibility, other objects (e.g., ScatterPlotItem) eat the MousePressEvent.

@janezd, could you look into making the first group's color stay the same when adding a second group to selection?

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -7,7 +7,7 @@ from unittest.mock import Mock, patch
 import numpy as np
 import scipy.sparse as sp
 
-from AnyQt.QtGui import QFont
+from AnyQt.QtGui import QFont, QTextDocumentFragment
 from AnyQt.QtCore import QRectF, QPointF
 from AnyQt.QtTest import QSignalSpy
 from AnyQt.QtWidgets import (
@@ -676,11 +676,9 @@ class ProjectionWidgetTestMixin:
 
     def test_dragging_tooltip(self):
         """Dragging tooltip depends on data being jittered"""
-        text = self.widget.graph.tiptexts[0]
+        text = QTextDocumentFragment.fromHtml(self.widget.graph.tiptexts[0]).toPlainText()
         self.send_signal(self.widget.Inputs.data, Table("heart_disease"))
         self.assertEqual(self.widget.graph.tip_textitem.toPlainText(), text)
-        self.widget.graph.controls.jitter_size.setValue(1)
-        self.assertGreater(self.widget.graph.tip_textitem.toPlainText(), text)
 
     def test_sparse_data(self, timeout=DEFAULT_TIMEOUT):
         """Test widget for sparse data"""
--- a/Orange/widgets/utils/plot/owplotgui.py
+++ b/Orange/widgets/utils/plot/owplotgui.py
@@ -279,6 +279,7 @@ class OWToolbar(OrientedWidget):
                         self.buttons.update(s.buttons)
                         self.groups[buttons[i+1]] = s
                         i = j
+                        self.layout().addStretch()
                         break
                     else:
                         state_buttons.append(buttons[j])
@@ -291,7 +292,6 @@ class OWToolbar(OrientedWidget):
             else:
                 self.buttons[buttons[i][0]] = gui.tool_button(buttons[i], self)
             i = i + 1
-        self.layout().addStretch()
 
     def select_state(self, state):
         # SELECT_RECTANGLE = SELECT
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -604,7 +604,9 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
              format("Cmd" if sys.platform == "darwin" else "Ctrl")),
             (Qt.AltModifier, "Alt: Remove")
         ]
-        all_parts = ", ".join(part for _, part in tip_parts)
+        all_parts = "<center>" + \
+                    ", ".join(part for _, part in tip_parts) + \
+                    "</center>"
         self.tiptexts = {
             int(modifier): all_parts.replace(part, "<b>{}</b>".format(part))
             for modifier, part in tip_parts
@@ -616,6 +618,7 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         text.setHtml(self.tiptexts[Qt.ControlModifier])
         text.setPos(4, 2)
         r = text.boundingRect()
+        text.setTextWidth(r.width())
         rect = QGraphicsRectItem(0, 0, r.width() + 8, r.height() + 4)
         rect.setBrush(QColor(224, 224, 224, 212))
         rect.setPen(QPen(Qt.NoPen))
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -540,6 +540,7 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         gui.OWComponent.__init__(self, scatter_widget)
 
         self.subset_is_shown = False
+        self.jittering_suspended = False
 
         self.view_box = view_box(self)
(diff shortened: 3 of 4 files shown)


janezd:
> I removed the 'Warning: Selection is applied to unjittered data' tooltip, I appreciate the gesture in explaining that to the user, but it doesn't seem like something somebody would complain to us about.

I see a point that I try to select. I drag a rectangle around it ... and it's not selected, because it's actually not there -- I see just its jittered position. The point is not where it seems to be. To any user (including me, on the first glance), this would look like a bug.

If we remove the warning, then I believe we should prevent the user from make any selection when the data is jittered.

> @janezd, could you look into making the first group's color stay the same when adding a second group to selection?

To my recollection, we already discussed this. The colors of the groups should match the usual sequence (blue, red, green ...) because these colors would represent those groups in any subsequent visualizations. So if the color stays the same, it would have to be blue. On the other hand, we rarely create multiple groups, so we almost always see just the initial color. The current initial color, orange, sticks out much better than blue, which already appears as one of the points' colors. (I hope I explained the problem well enough.)

I guess we need to discuss this live.

irgolic:
> > I removed the 'Warning: Selection is applied to unjittered data' tooltip, I appreciate the gesture in explaining that to the user, but it doesn't seem like something somebody would complain to us about.
> 
> I see a point that I try to select. I drag a rectangle around it ... and it's not selected, because it's actually not there -- I see just its jittered position. The point is not where it seems to be. To any user (including me, on the first glance), this would look like a bug.
> 
> If we remove the warning, then I believe we should prevent the user from make any selection when the data is jittered.

Ah, I thought it was warning about the data values on the output not being affected by jitter. I get the rationale behind this, it's noisy/misleading to select jittered data.
Although, it feels unintuitive for points not in the rectangle to not be selected. I'd rather a rectangle selects the jittered points within it, than the invisible points it leaves behind.

What if selection worked as expected, but selected points would return to their unjittered position?

> > @janezd, could you look into making the first group's color stay the same when adding a second group to selection?
> 
> To my recollection, we already discussed this. The colors of the groups should match the usual sequence (blue, red, green ...) because these colors would represent those groups in any subsequent visualizations. So if the color stays the same, it would have to be blue. On the other hand, we rarely create multiple groups, so we almost always see just the initial color. The current initial color, orange, sticks out much better than blue, which already appears as one of the points' colors. (I hope I explained the problem well enough.)
> 
> I guess we need to discuss this live.

I see, I understand and agree with you.

janezd:
Yes, anyway you do it, It doesn't work.

If selection is done on points as shown (i.e. based on jittered values):
- you select all points in a certain region, say [3, 4] x [3, 4]. You feed the output to the next scatter plot and it shows some points utside this region;
- and though you thought you selected the region, you later notice some unselected points within.

If selection is done on original data:
- you drag a selection rectangle but the shown selection does not correspond to what you see.

Currently, the widget does the latter (it follows the reality and not the perception), but warns the user why (s)he sees what (s)he sees.

> What if selection worked as expected, but selected points would return to their unjittered position?

This would be nice. Perhaps with also showing very transparent points at their jittered positions?

irgolic:
@janezd Check it out, do you think it's necessary to add the transparent points as well?

ajdapretnar:
IMHO transparent points would add to the confusion, not subtract from it. ;)

irgolic:
Points will now unjitter during selection.

ezgif-7-2ab765bb133a

Selected points are still not jittered. Should this remain the case, or should selected points jitter as usual?

janezd:
I like this a lot from the user perspective (bold parts in tooltips are great!).

But I'm a uncomfortable with implementation. owscatterplotgraph was reimplemented from scratch two years ago. The main reason was that methods had many arguments that would decide whether to update this and that or not; whether to include this and that part of processing or not; ... The current philosophy is very clean: (exposed) get_* and update_* methods have no arguments. The exception, `_keep_selection` in `reset_graph`, was discussed a lot before being added.

I would be very happy if the `doJitter` argument could be avoided -- for no particular reason except that it's a step towards where've been and where we wouldn't want to be again.

What do you think about the patch I pasted below? The `suspend_jittering` method could also be used in other widgets, e.g. MDS during optimization.

```
diff --git a/Orange/widgets/visualize/owscatterplotgraph.py b/Orange/widgets/visualize/owscatterplotgraph.py
index f2fea6390..7dc43cd73 100644
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -540,6 +540,7 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         gui.OWComponent.__init__(self, scatter_widget)
 
         self.subset_is_shown = False
+        self.jittering_suspended = False
 
         self.view_box = view_box(self)
         _axis = {"left": AxisItem("left"), "bottom": AxisItem("bottom")}
@@ -640,8 +641,16 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
                 break
         self.tip_textitem.setHtml(text)
 
-    def update_jittering(self, doJitter=True):
-        x, y = self.get_coordinates(doJitter)
+    def suspend_jittering(self):
+        self.jittering_suspended = True
+        self.update_jittering()
+
+    def unsuspend_jittering(self):
+        self.jittering_suspended = False
+        self.update_jittering()
+
+    def update_jittering(self):
+        x, y = self.get_coordinates()
         if x is None or len(x) == 0 or self.scatterplot_item is None:
             return
         self._update_plot_coordinates(self.scatterplot_item, x, y)
@@ -759,7 +768,7 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         else:
             return np.asarray(data[self.sample_indices])
 
-    def get_coordinates(self, doJitter=True):
+    def get_coordinates(self):
         """
         Prepare coordinates of the points in the plot
 
@@ -787,8 +796,7 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         # a "sample size" slider), points would move around when the sample
         # size changes. To prevent this, jittering should be done before
         # sampling (i.e. two lines earlier). This would slow it down somewhat.
-        if doJitter:
-            x, y = self.jitter_coordinates(x, y)
+        x, y = self.jitter_coordinates(x, y)
         return x, y
 
     def _create_sample(self):
@@ -810,7 +818,7 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         Display coordinates to random positions within ellipses with
         radiuses of `self.jittter_size` percents of spans
         """
-        if self.jitter_size == 0:
+        if self.jitter_size == 0 or self.jittering_suspended:
             return x, y
         return self._jitter_data(x, y)
 
diff --git a/Orange/widgets/visualize/utils/plotutils.py b/Orange/widgets/visualize/utils/plotutils.py
index f6d466f8b..f3a3b46a8 100644
--- a/Orange/widgets/visualize/utils/plotutils.py
+++ b/Orange/widgets/visualize/utils/plotutils.py
@@ -211,13 +211,13 @@ class InteractiveViewBox(pg.ViewBox):
                 self.safe_update_scale_box(ev.buttonDownPos(), ev.pos())
                 if ev.isFinish():
                     self._updateDragtipShown(False)
-                    self.graph.update_jittering(True)
+                    self.graph.unsuspend_jittering()
                     self.rbScaleBox.hide()
                     value_rect = get_mapped_rect()
                     self.graph.select_by_rectangle(value_rect)
                 else:
                     self._updateDragtipShown(True)
-                    self.graph.update_jittering(False)
+                    self.graph.suspend_jittering()
                     self.safe_update_scale_box(ev.buttonDownPos(), ev.pos())
 
         def zoom():
```

irgolic:
You're right, that's much cleaner, thanks @janezd. :)

PR #5039, opened 2020-10-18
owscatterplotgraph: Align point and selection outline

irgolic:
##### Issue
The selection outline of scatterplot points was off-center by a pixel. This is due to the way pyqtgraph optimizes symbol drawing in `SymbolAtlas`, and also the reason for why points look like they jitter by a pixel when changing symbol size.

Before:
Screenshot 2020-10-18 at 16 47 13
After:
Screenshot 2020-10-18 at 16 46 47

##### Description of changes
Is there a reason for the pen to have width one larger than the set selection width? If this is necessary, it should be +2 to avoid the offset issue.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
My absolute favourite nitpicking PR so far! <3

markotoplak:
Looks really good, thanks!

PR #5042, opened 2020-10-20
Add psycopg2-binary as alternative to psycopg2 to docs

PrimozGodec:
##### Issue

`psycopg2` need to be compiled in some cases e.g. MacOS.

##### Description of changes
The user can use (until he/she needs to use system libraries) `psycopg2-binary` instead `psycopg2` since it has already compiled wheels for many pythons-os combinations:

Citing `psycopg2`:
> You can also obtain a stand-alone package, not requiring a compiler or external libraries, by installing the psycopg2-binary package from PyPI

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation

### DIFF
--- a/doc/visual-programming/source/widgets/data/sqltable.md
+++ b/doc/visual-programming/source/widgets/data/sqltable.md
@@ -32,6 +32,12 @@ Install the backend.
 
 Alternatively, you can follow [these instructions](https://blog.biolab.si/2018/02/16/how-to-enable-sql-widget-in-orange/) for installing the backend.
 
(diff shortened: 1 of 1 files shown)


lanzagar:
Are we sure we want this as the dependency, the official docs advise against it?
https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary

Maybe it could just be an alternative for those that can't install the regular package? I don't know if there is an automatic way to to that (that makes sense) or if we just explain it in the readme/comments?

PrimozGodec:
@lanzagar I haven't seen this part of the documentation. Then let's leave it as it is. I suggest including this option in the documentation (will upgrade the docs). There are still windows binaries in `psycopg2`, so `psycopg2-binary` solves only macOS and Linux issues.

PR #5043, opened 2020-10-21
setup: Add sphinx and recommonmark to build time dependencies

ales-erjavec:
##### Issue

Widget help files are not included in source and wheel files.

##### Description of changes

* Add sphinx and recommonmark to build time dependencies
* Build and install widget help documentation files automatically.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,9 @@
 requires = [
     "setuptools>=40.8.0,<50.0",
     "wheel",
-    "oldest-supported-numpy"
+    "oldest-supported-numpy",
+    "sphinx",
+    "recommonmark",
 ]
 
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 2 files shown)


markotoplak:
I tried this out, and both binary and source packages work beautifully. Thanks, @ales-erjavec.

Our pypi are going to be ~20MB after this PR. Now we really need to be careful with screenshot size. Currently we even have some images that are almost 0.5 MB. Some of them are even retina. @ajdapretnar, you work with docs the most, please be mindful of it. Perhaps we could go for stricter image tests (disallowing images that are bigger that 1200x800, or those > 100kb) or something, but it is hard to think of a criterium that makes sense...

@ales-erjavec, would we also need to do something about conda packages or this also applies to them?

PR #5046, opened 2020-10-23
owlineplot: Fix future compatibility

irgolic:
pyqtgraph/pyqtgraph@3b6eb02 makes a test fail due to an invalid mapping of points (I'm not sure whether functionality is also comprised). 

I'm also not sure this is the correct method to use, it seems a correct answer would also be to not map points at all, using the raw point values.

@VesnaT, would you know more?

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owlineplot.py
+++ b/Orange/widgets/visualize/owlineplot.py
@@ -133,8 +133,8 @@ class LinePlotViewBox(ViewBox):
         self.addItem(self.selection_line, ignoreBounds=True)
 
     def update_selection_line(self, button_down_pos, current_pos):
-        p1 = self.childGroup.mapFromParent(button_down_pos)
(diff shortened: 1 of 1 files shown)


VesnaT:
This PR does fix the falling test, but disables the selection.

It looks like the LinePlot has several issues with the pyqtgraph's master branch (lines, legend, visual settings do not work on legend's font, ...). 
- pyqtgraph-0.11.0 branch:
Screenshot 2020-10-26 at 09 16 15
- master branch
Screenshot 2020-10-26 at 09 14 17

irgolic:
@VesnaT Could you fix this, since you'll probably be faster, having written the widget?

VesnaT:
I'll look into it.

PR #5050, opened 2020-10-27
owdataprojectionwidget: Update labels on attr change

irgolic:
##### Issue
File (Iris) -> OWScatterPlot
Enable 'Label only selection and subset', choose a label, choose some points
Change one of the attribute

The points will still be selected, but won't have labels displayed next to them. This is easily fixable by toggling the 'Label only selection and subset' option.

##### Description of changes
Retain labels next to selected points upon attribute change.

I'd love to write a test for this, but I'm honestly not sure where to do it. Could someone advise me?

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/utils/widget.py
+++ b/Orange/widgets/visualize/utils/widget.py
@@ -554,6 +554,8 @@ class OWDataProjectionWidget(OWProjectionWidgetBase, openclass=True):
             self.__pending_selection = None
             self.graph.selection = selection
             self.graph.update_selection_colors()
+            if self.graph.label_only_selected:
+                self.graph.update_labels()
 
     def selection_changed(self):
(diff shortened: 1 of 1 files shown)


ajdapretnar:
I suppose in `test_owprojectionwidget`?

janezd:
You can write a trivial unittest that checks whether `update_labels` is called. Or a test that some labels are shown after changing an attribute. We have lots of such, typically comented with "GH-xxxx". I don't like them because they test for a specific bug and not some behaviour.

We are good at unit tests, but not at integration tests. @VesnaT writes them, but (despite my immense respect to her meticulousness) they are not very systematic. Probably because it is impossible to systematically test all possible scenarios and sequences.

I don't see how to test a general scenario, not this particular bug. I'm merging, but you can add a test later if you wish.

PR #5062, opened 2020-11-03
Distance Matrix: Ignore coloring on string variables

### LINKED ISSUE #5061: Distance Matrix: tries to color by label and crashes on string
- [ ] **What's wrong?**
Distance Matrix crashes when selecting a string label. It also seems to color the headers, which IMHO, doesn't make sense.

Datasets (Grades data set) - Distances - Distance Matrix. Select label = name.

- [ ] **Which version of Orange are you using? How did you install Orange?**
3.27.1

- [ ] **What is your operating system?**
OSX

ajdapretnar:
##### Issue

Fixes #5061. 

##### Description of changes
Ignore coloring by attribute when attribute is string.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


ajdapretnar:
I don't really know how to write the tests here. On master, the tests pass, but the error appears. Here, the tests pass and no error. 🤷

PR #5064, opened 2020-11-03
__main__: Update Orange in-app

irgolic:
##### Issue
As Orange can be updated from the Add-ons dialogue, there seems to be no need to redownload the whole package.

##### Description of changes
On clicking 'Yes' in the update request notification, the Installer (previously Add-ons) dialogue will open, and select Orange3 to be updated. All the user must do is press Ok.

If the user does not have install permissions, the website is opened to prompt a download.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #5071, opened 2020-11-06
__main__: Restart on exit code 96

irgolic:
https://github.com/biolab/orange-canvas-core/pull/153
https://github.com/biolab/orange-widget-base/pull/113

Decoupled from #5064.

This should be merged after https://github.com/biolab/orange-canvas-core/pull/157 is merged and released.

I considered running `execv` with the arguments Orange was run with previously (e.g., open a workflow), but it didn't seem much more intuitive than a feature that would automatically open the previously opened workflow on starting Orange (idea?).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -24,6 +24,7 @@ from unittest.mock import patch
 from urllib.request import urlopen, Request
 
 import pkg_resources
+import psutil
 import yaml
 
 from AnyQt.QtGui import QFont, QColor, QPalette, QDesktopServices, QIcon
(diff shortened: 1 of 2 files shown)


Orange/canvas/__main__.py
```diff
@@ -694,6 +696,17 @@ def show_message(message):
...
+
+    if status == 96:
+        log.info('Restarting via exit code 96.')
+        try:
+            p = psutil.Process(os.getpid())
+            for handler in p.open_files() + p.connections():
+                os.close(handler.fd)
+        except Exception as e:  # pylint: disable=broad-except
+            logging.error(e)
+        return os.execv(sys.executable, ['python', sys.argv[0]])
```
ales-erjavec, line 708: I do not think this approach is correct. Granted that this is close to process exit, but threads could (and do) still run and it is unknown how forcibly closing all fds would affect them.

Similarly plain execv would for the non-main threads be equivalent to SIGKILL, possibly leaving inconsistent persistent state.

I think it would be better to simply start the new process (QProcess.startDetached or just subprocess.Popen), and letting this one exit normally.

irgolic: Good point. I was trying to still run as the 'same process' in the eyes of a command line. I guess this isn't really necessary.

irgolic: Well, QProcess.startDetached was introduced in Qt 5.10 (unviable for Anaconda users on 5.9), and subprocessing.Popen's `DETACHED_PROCESS` flag was introduced in python3.7.

I'll have to do some more reading on subprocessing.Popen, I think it should allow the parent process to exit gracefully regardless of the `DETACHED_PROCESS` flag.

PR #5080, opened 2020-11-10
Cythonize sources at build time

ales-erjavec:
##### Issue

Cythonized sources under source control do no build for Python 3.9.

##### Description of changes

Remove cythonized sources from version, cythonize at build time instead.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Should we add cython to requirements? I suppose no, otherwise you'd add it instead of checking whether it's installed. But would it make sense to add it to requirements-dev?

markotoplak:
Works great for me. I amended the last commit by also adding `_variable.c` to `.gitignore`. 

If I do `pip install -e .` everything works nicely even if I do not have cython installed: pip downloads cython and other build dependecies and puts them into a special folder; it does not pollute my environment. If I do `python setup.py develop`, it complains that I do not have cython. Both scenarios work as expected.

`python setup.py bdist_wheel` also works.

lanzagar:
My frankenstein building with `pip install --no-build-isolation --user -e .` also worked.
It does take noticeably longer now to (re)install, but I guess this is a price we are willing to pay.

PR #5089, opened 2020-11-18
[FIX] SQL: Fix the issue with database collation setting when retrieving column values

PrimozGodec:
##### Issue

The database can have different collation settings (how to compare characters in the group by, distinct, ...). It has an impact on values we get when we retrieve all distinct values from column to create a discrete variable. If a collation is Latin it ignores utf8 characters, it can even be case insensitive. When creating variable we still want to have all possible values in the column since later when we retrieve actual data for a table (SELECT * FROM Table) we get values as they are (collation does not have an effect).

##### Description of changes
This PR implements a workaround for GROUP BY in cases when we want all values to create a variable.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/sql/backend/mssql.py
+++ b/Orange/data/sql/backend/mssql.py
@@ -152,3 +152,16 @@ class PymssqlBackend(Backend):
                     warnings.warn("SHOWPLAN permission denied, count approximates will not be used")
                     return None
                 raise BackendError(parse_ex(ex)) from ex
+
+    def distinct_values_query(self, field_name: str, table_name: str) -> str:
+        field = self.quote_identifier(field_name)
+        return self.create_sql_query(
+            table_name,
+            [field],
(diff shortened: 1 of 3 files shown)


Orange/data/sql/backend/mssql.py
```diff
@@ -152,3 +152,16 @@ def count_approx(self, query):
...
                     return None
                 raise BackendError(parse_ex(ex)) from ex
+
+    def distinct_values_query(self, field_name: str, table_name: str) -> str:
+        field = self.quote_identifier(field_name)
+        return self.create_sql_query(
+            table_name,
+            [field],
+            # workaround for collations that are not case sensitive and
+            # and UTF characters sensitive - in the domain we still want to
```
lanzagar, line 162: and and -> and

PR #5095, opened 2020-11-24
[FIX] Predictions: Fix crash when clicking on empty left pane

### LINKED ISSUE #5074: Predictions: IndexError when clicking outside the table
- [ ] **What's wrong?**

Predictions crashes when clicking outside the table fields.

```
  File "/Users/ajda/orange/orange3/Orange/widgets/evaluate/owpredictions.py", line 1130, in select
    self.store.select(selection, flags)
  File "/Users/ajda/orange/orange3/Orange/widgets/evaluate/owpredictions.py", line 1057, in select
    rows = {selection.model().mapToSource(selection).row()}
AttributeError: 'NoneType' object has no attribute 'mapToSource'
```

Create a prediction workflow with a small test data set. Click outside the table.

- Operating system: OSX
- Orange version: 3.28.dev
- How you installed Orange:

erelin6613:
##### Issue

Fixes #5074.

Predictions widget reported an error if it was given just data, but no models, and the user clicked on the (empty) left pane.

##### Description of changes

`select` method now checks whether the model exists; if not, it set the selection to empty.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -1054,7 +1054,10 @@ class SharedSelectionStore:
                 flags that tell whether to Clear, Select, Deselect or Toggle
         """
         if isinstance(selection, QModelIndex):
-            rows = {selection.model().mapToSource(selection).row()}
+            if selection.model() is not None:
+                rows = {selection.model().mapToSource(selection).row()}
+            else:
+                rows = set()
         else:
             indices = selection.indexes()
             if indices:
@@ -1074,15 +1077,16 @@ class SharedSelectionStore:
             flags (QItemSelectionModel.SelectionFlags):
                 flags that tell whether to Clear, Select, Deselect or Toggle
         """
-        with self._emit_changed():
-            if flags & QItemSelectionModel.Clear:
-                self._rows.clear()
-            if flags & QItemSelectionModel.Select:
-                self._rows |= rows
-            if flags & QItemSelectionModel.Deselect:
-                self._rows -= rows
-            if flags & QItemSelectionModel.Toggle:
-                self._rows ^= rows
+        if len(rows) !=0:
+            with self._emit_changed():
+                if flags & QItemSelectionModel.Clear:
+                    self._rows.clear()
+                if flags & QItemSelectionModel.Select:
+                    self._rows |= rows
+                if flags & QItemSelectionModel.Deselect:
+                    self._rows -= rows
+                if flags & QItemSelectionModel.Toggle:
+                    self._rows ^= rows
 
     def clear_selection(self):
         """Clear selection and emit changeSelection signal to all models"""
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -24,3 +24,4 @@ pandas
 pyyaml
 openpyxl
 httpx~=0.14.0
+Cython==0.29.21


erelin6613:
Minor fix for error handling

Orange/widgets/evaluate/owpredictions.py
```diff
@@ -1074,15 +1077,16 @@ def select_rows(self, rows: Set[int], flags):
...
-        with self._emit_changed():
-            if flags & QItemSelectionModel.Clear:
-                self._rows.clear()
-            if flags & QItemSelectionModel.Select:
-                self._rows |= rows
-            if flags & QItemSelectionModel.Deselect:
-                self._rows -= rows
-            if flags & QItemSelectionModel.Toggle:
-                self._rows ^= rows
+        if len(rows) !=0:
```
janezd, line 1080: Is this necessary?

Without this condition, clicking in the empty area clears the selection, which, I guess, makes sense. I'd prefer to remove this `if`.

If you think we should keep it, I'd suggest using `if rows:`. This is a more Pythonic way to test whether a set is non-empty. I noticed that the above docstring is incorrect (my fault!), so if you'd make this change, you can also fix the docstring --- `selection (set of int)` should be `rows (set of int)`.

requirements-core.txt
```diff
@@ -24,3 +24,4 @@ pandas
 pyyaml
 openpyxl
 httpx~=0.14.0
+Cython==0.29.21
```
janezd, line 27: This is unrelated. It probably came from something else you were doing. Can you remove it, please?

janezd (changes requested):
Thanks for diagnosing and fixing this bug.

I added just a few comments regarding two other parts of the code that you changed. I guess you can just remove those changes and amend the existing commit.

I also edited the PR's description to be more informative. As you can see, I changed the title to "Predictions: Fix crash when clicking on empty left pane" and prefixed it with `[FIX]`, so it will be mentioned in the release notes. I tagged the description with "Fixes #5074", which will make github automatically close issue 5704, when this PR is merged. I also added two sentences of description.

Would you be able to write a test (probably in in `test_owpredictions.SharedSelectionStoreTest`) that would fail without your fix? Writing such tests may be complicated, so feel free to say no, if you don't wish to dive into it.

requirements-core.txt
erelin6613: Okay, I will remove it. I was trying to work with clear virtual environment and code would not run complaining about Cython. Perhaps it is a limitation of the virtualenv, I don't argue.

Orange/widgets/evaluate/owpredictions.py
erelin6613: Otherwise code fails with the issue I was trying to solve. True, `if rows` would be more pythonic, I will it and (if no objection arises) will redo request. Thank you!

janezd: Are you sure? If you make the first fix but skip this one, I can no longer replicate the issue - clicking just disables current selection. Also, the code that is guarded with this `if` seems to be OK with empty sets.

erelin6613: That is odd, It has been crashing for me, in case it helps the trace is as follows
`Traceback (most recent call last):
  File "/home/val/coding/orange3/Orange/widgets/evaluate/owpredictions.py", line 1134, in select
    self.store.select(selection, flags)
  File "/home/val/coding/orange3/Orange/widgets/evaluate/owpredictions.py", line 1068, in select
    self.select_rows(rows, flags)
  File "/home/val/coding/orange3/Orange/widgets/evaluate/owpredictions.py", line 1089, in select_rows
    self._rows ^= rows
  File "/usr/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/home/val/coding/orange3/Orange/widgets/evaluate/owpredictions.py", line 1115, in _emit_changed
    deselected = map_from_source(old_rows - self._rows)
  File "/home/val/coding/orange3/Orange/widgets/evaluate/owpredictions.py", line 1108, in map_from_source
    index = self.proxy.sourceModel().index
AttributeError: 'NoneType' object has no attribute 'index'`
So that is how I wanted to secure it. Could it be an issue I work from specifically created env but with all dependencies installed?

janezd: Uh, I can't reproduce it, but I see, where it crashes -- thanks to your traceback.

`_emit_changed` is a context manager, and so the code inside `finally` will be executed after the block within `with` in the function you patched. That's why it helped when you have put the whole `with` into an `if` close.

But in this case, the code that is wrong is actually in `_emit_changed` and a similar problem will appear anytime `_emit_changed` is used without having all models. So this is what needs to be fixed.

Could you please remove the `if` about which we talk here, and try adding `if self.proxy.sourceModel() is not None` at the beginning of `finally`.  It might be also prudent to add `if model is not None` before `model.emit_selection_rows_changed(selected, deselected)`.

After doing so, please check whether the bug is still present. I'd try it myself, but can't because it already works for me...

Thanks a lot for digging into this - it's turning out to be not as trivial as it seemed, isn't it? :)

> Could it be an issue I work from specifically created env but with all dependencies installed?

I don't think so. We probably just do something slightly differently.

erelin6613: Aha, I see. I did not work with qt in a while so was trying to put my captain hat here. Okay, I will do my best. 

Indeed the code-base alone and work you guys put in it is enormous. Hopefully once I will figure out interactions here I will be able to help even more :)

janezd: Any progress?

janezd:
Continued in #5222.

PR #5105, opened 2020-12-03
Select columns: option to unselect new columns

PrimozGodec:
##### Issue

New columns that come in the select columns widget are placed in selected data. It can sometimes break the workflow. *It is the most problematic when workflows still pass but results can be wrong because of new columns.*

##### Description of changes
I present the option to decide what to do with new columns.

#### To discuss
- ideas for a good label
- since this option says about new/additional features, I decided to take checkbox into account only in cases when context already exists for this domain. In cases with new context, all features are automatically selected. Should we keep like this or should we also take this option into account with a completely new domain? If we decide for later we should change the label.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectcolumns.py
+++ b/Orange/widgets/data/owselectcolumns.py
@@ -1,5 +1,5 @@
 from functools import partial
-from typing import Optional
+from typing import Optional, Dict, Tuple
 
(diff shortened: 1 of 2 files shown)


janezd:
I like the solution, but it requires a very informative label.

Issue #5069, opened 2020-11-06
Distance Matrix is very slow


janezd:
- [x] **What's wrong?**

Distance Matrix freezes on not-unreasonably large data.

- [x] **How can we reproduce the problem?** 

- Download, unzip [moscow oblast.dst.zip](https://github.com/biolab/orange3/files/5500105/moscow.oblast.dst.zip)
- load it with Distance File
- open it in Distance Matrix.

Or construct a distance matrix from heart disease data.

- [x] **Which version of Orange are you using? How did you install Orange?**

Any, including development.

- [x] **What is your operating system?**

macOS

PR #5136, opened 2020-12-17
owdistancematrix: Reduce size hinting computational cost

ales-erjavec:
##### Issue

Fixes gh-5069

##### Description of changes

 Reduce size hinting computational cost

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owdistancematrix.py
+++ b/Orange/widgets/unsupervised/owdistancematrix.py
@@ -177,7 +177,8 @@ class TableView(gui.HScrollStepMixin, QTableView):
         style = self.style()
         sh = style.sizeFromContents(
             QStyle.CT_ItemViewItem, opt, QSize(), self)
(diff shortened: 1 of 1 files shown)


janezd:
How I wish this was fixed two months ago!

It's superfast, but appears to make the columns too narrow.

In the font I have (on Catalina), digits `4` and `6` are wider than `0`. With only `1` and `7` being narrower, this is what I get on heart disease. (Interestingly, some numbers are elided and some are broken into two lines?)

Screenshot 2020-12-18 at 10 38 47

If I change the template to use 6 instead of 0, it is better, but I still found a number which is too wide for its column.

Screenshot 2020-12-18 at 10 42 03

Already in the first screenshot, some numbers are too wide although they don't contain any 4s or 6s. Is there some hidden padding unaccounted for?

PR #5138, opened 2020-12-18
[FIX] ROC shows all points, including the last

### LINKED ISSUE #5119: Error in ROC Analysis plot
**What's wrong?**
The plot in ROC analysis is wrong. The most obvious mistake is that the steps in the x-direction should be in 1/N increments. Instead, the steps currently shown in ROC are in 1(N-1) increments, that is, one instance of non-target class seems to be ignored. While it looks like that to be the one with the lowest target probability, that would need to be verified.

For instance, for the following predicted probabilities:
image
the ROC curve is
image
Notice that, going from (0, 0), the curve takes the right steps (up and right) until the last item with a true class of "0", where the last horizontal step should be extended by one more "stair". The area under the curve from this screenshot is 0.60, whereas Predictions shows the correct value of 0.64. 

Notice also that the used grid in the x-direction from the screenshot is 0.25 (=1/4), whereas it should be 0.2 (=1/5). That is, the first step to the right, starting from (0, 0), should be (0.2, 0.4) and not (0.25, 0.4).

Use the following workflow ([roc-bug.ows.zip](https://github.com/biolab/orange3/files/5688123/roc-bug.ows.zip))

image

Alternatively, choose any set for the Predictions widget where, say, N=5 and P=5 (number of instances in non-target and target class).

**What's your environment?**
- Operating system: osx
- Orange version: 3.28.dev
- How you installed Orange: the bug is both in dmg and on the latest version from github

janezd:
##### Description of changes

- Fixes #5119: The last point on the curve was not shown.
- Fixes the threshold shown at the bottom-left point (now set to 1; used to be 1 + the highest p in the data)
- Adds simpler preview data
- Shows ticks for actual FPR/TPR values, if there are <15 unique values

roc

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owrocanalysis.py
+++ b/Orange/widgets/evaluate/owrocanalysis.py
@@ -825,6 +825,13 @@ def roc_curve_for_fold(res, fold, clf_idx, target):
         drop_intermediate=drop_intermediate
     )
 
+    # skl sets the first threshold to the highest threshold in the data + 1
+    # since we deal with probabilities, we (carefully) set it to 1
+    # Unrelated comparisons, thus pylint: disable=chained-comparison
+    if len(thresholds) > 1 and thresholds[1] <= 1:
+        thresholds[0] = 1
+    return fpr, tpr, thresholds
+
 
 def roc_curve_vertical_average(curves, samples=10):
     if not curves:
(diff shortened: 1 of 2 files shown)


janezd:
@BlazZupan, do you like the changed ticks when the data is small (to be used in particular for teaching)? If you do, you can review and merge if you will. (The failing check is irrelevant.) If not, say so and I'll remove the commit.

BlazZupan:
I have tested this widget on several smaller test sets, where I could draw the ROC curve manually. Now it works fine, and the error was corrected. I also like the changes with the ticks.

PR #5141, opened 2020-12-18
[ENH] Projections: Allow transparent subset

### LINKED ISSUE #5109: Geo Map disables Opacity when data subset signal present
**What's wrong?**

Feeding a data subset into Geo Map disables Opacity. Which is wrong, since opacity allows to see beyond the data points and perhaps read some of the labels on the map, which are now, with disabled opacity setting not readable.

image

In the old version of the widget a similar visualization looked like this:

old-one

Notice also that the opacity in the old version of the widget was such that it was possible to read the map labels.

Here is the workflow. Use the Philadelphia Crime data set.

image

[philly.ows.zip](https://github.com/biolab/orange3/files/5642912/philly.ows.zip)

**What's your environment?**

- Orange version: 3.27

janezd:
##### Issue

Resolves #5109.

##### Description of changes

- No longer disable opacity in presence of data subset
- If opacity is low and subset is present, inform the user (s)he should increase opacity if (s)he can't see the subset

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -531,7 +531,7 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
 
     COLOR_NOT_SUBSET = (128, 128, 128, 0)
     COLOR_SUBSET = (128, 128, 128, 255)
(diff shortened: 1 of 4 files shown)


BlazZupan:
Works fine and the warning comes handy.

PR #5159, opened 2021-01-05
[FIX] Calibration model: Work with numpy data

PrimozGodec:
##### Issue

Calibration models do not support passing NumPy data since their `data_to_model_domain` does not transform data correctly. It should take into account the base model domain. The error is that the calibration model stores the wrong domain. It first sets the base_learners domain, but it is then overwritten with the wrong (not preprocessed) domain.

This code snipet works with other models but not with callibrations:

```
data = Table('heart_disease')
base_learner = LogisticRegressionLearner()
model = CalibratedLearner(base_learner)(data)
model(model.data_to_model_domain(data).X)
```

The correction would support explain the model on calibration learner.

##### Description of changes
This PR fixes the base learner such that it does not overwrite the domain of models that already set it. It only set the domain for other models (mostly SKL models).

If it looks ugly this error could be also resolved by completely removing the line that set the domain. Then all models would need to pass the domain to the model's constructor. It would be a major change in models and learners.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -136,7 +136,11 @@ class Learner(ReprableWithPreprocessors):
         progress_callback(0.1, "Fitting...")
         model = self._fit_model(data)
         model.used_vals = [np.unique(y).astype(int) for y in data.Y[:, None].T]
-        model.domain = data.domain
+        if not hasattr(model, "domain") or model.domain is None:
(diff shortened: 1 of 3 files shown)


janezd:
I agree it's not nice, but it works.

Issue #5107, opened 2020-12-04
Gradient boosted trees: Implementation


VesnaT:
Implement Gradient boosted trees (XGBoost) algorithm.

PR #5160, opened 2021-01-05
[ENH] Gradient Boosting: New widget

VesnaT:
##### Issue

Implements #5107
Alternative to #5116

##### Description of changes
- wrap sklearn gradient boosting classification and regression learners
- wrap catboost classification and regression learners
- wrap xgboost classification and regression learners
- Gradient Boosting widget

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


lanzagar:
Works well for me.
I would suggest:
- being explicit for the default "Gradient Boosting" method as well and changing to "Gradient Boosting (scikit-learn)" to be more consistent with other entries in the dropdown
- changing the default num of trees for catboost from 500 to 100 so it is the same as all the other methods - catboost is already the slowest (might need additional investigation)
- change default value of Replicable training to True. BTW is this really only available from scikit-learn, do xgboost and catboost not allow setting random seeds?

VesnaT:
> * change default value of Replicable training to True. BTW is this really only available from scikit-learn, do xgboost and catboost not allow setting random seeds?

I think so. Both (xgboost and catboost) have 0 as default. If one passes None, it is changed to 0.

PR #5167, opened 2021-01-08
OWSGD: Rerrange UX

janezd:
##### Issue

Tests fail because the OWSGD is too high.

##### Description of changes

Rearrange the GUI into two columns + minor realignments.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owsgd.py
+++ b/Orange/widgets/model/owsgd.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 
 from AnyQt.QtCore import Qt
+from AnyQt.QtWidgets import QHBoxLayout, QVBoxLayout, QGridLayout, QLabel
 
 from Orange.widgets.report import bool_str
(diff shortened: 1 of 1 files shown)


janezd:
Let somebody review and merge this (if OK), so our tests start passing again.

Screenshot 2021-01-08 at 15 29 13

markotoplak:
There is a big margin before the last button on my Linux machine, but I am OK with it.

image

janezd:
Is this the only widget in which this happens? I think I've seen this before, somewhere on someone's machine.

Could you please check whether removing `__init__` helps?

markotoplak:
Removing `__init__` keeps the space on my computer. The problem is in the superclass, because also the other learners behave the same.

markotoplak:
Actually, the problem is in the auto_apply button, because even ScatterPlot has that huge space.

markotoplak:
Actually, the problem may be in the surrounding box. :D

PR #5189, opened 2021-01-17
Better pandas integration

irgolic:
Bases off #5190.

##### Issue
At the moment, our pandas solution copies the X, Y and meta numpy arrays into one contiguous array. It also sets its column `dtypes`. This is inefficient, and loses information like global row IDs, roles, variable info. 

The proposed solution is not an attempt at migration, but rather a seamless integration (both Orange and pandas operate on numpy arrays). This will be used in Python Script, but could also be leveraged by widget code.

##### Description of changes
Added a new `table_to_frames` method, which wraps the three existing numpy arrays into three DataFrames. This is near-instantaneous, as no copying occurs, it simply wraps the preexisting data structures. Also:
- it keeps global row index information by setting pandas indices in `_o1, _o2, _o3` format
- upon conversion back to table, it copies the columns' original variables (setting `compute_value` to `None`)
- it remembers what role the converted table represented (feature/target/meta), and applies the role to all its columns upon conversion back to table

An edge case: how do you handle concatenate when one of the frames has weights and the other doesn't?

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/pandas_compat.py
+++ b/Orange/data/pandas_compat.py
@@ -1,6 +1,12 @@
 """Pandas DataFrame↔Table conversion helpers"""
+from unittest.mock import patch
+
 import numpy as np
+from scipy import sparse as sp
+from scipy.sparse import csr_matrix
 import pandas as pd
+from pandas.core.arrays import SparseArray
+from pandas.core.arrays.sparse.dtype import SparseDtype
 from pandas.api.types import (
     is_categorical_dtype, is_object_dtype,
     is_datetime64_any_dtype, is_numeric_dtype,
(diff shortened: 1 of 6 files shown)


irgolic:
Table stores X, Y, and metas in three separate numpy arrays. DataFrame depends on numpy, just like us. So it's possible to simply wrap each of the tables into a DataFrame.

Orange provides `table_from_frame` and `table_to_frame` functions. When generating a DataFrame from a Table, all three numpy arrays (X, Y, metas) are copied to a contiguous location, and the DataFrame's dtypes are set. When converting back, variables types are inferred. The global row index is lost. This creates a fully-fledged standalone DataFrame, with no Orange footprint.

At some point, we contemplated taking Table and stuffing it into the DataFrame interface. This would be nice, but a rewrite would be better probably. Maybe in Orange4, when we separate server and client.

This PR adds a `table_to_frame` function, which returns three DataFrames: X, Y, and metas. The DataFrame's index is set to `_o1, _o2, _o3` format. In its metadata, it keeps the original orange variables (associated by column name), and the frame's role (attribute/target/meta). When converting back to an Orange table, global row indices are kept. The domain is constructed by copying the original column and setting `compute_value=None` (or guess new columns).

I did this TDD style, so the tests are nice.

I've added some benchmarks too, wherein `table` is an Orange table, `df` is a natural DataFrame, `orangedf` is DataFrame converted from an Orange table. Normalization is tested both with pandas vectorization and numpy vectorization (https://engineering.upside.com/a-beginners-guide-to-optimizing-pandas-code-for-speed-c09ef2c6a4d6). They're all one-liners, check the code if benchmark names are unclear.

```python
cols = 100
rows = 100000
```
```properties
[create_df] with 5 loops, best of 3:
	min 102 usec per loop
	avg 117 usec per loop
[create_orangedf] with 5 loops, best of 3:
	min 103 msec per loop
	avg 104 msec per loop
[create_table] with 5 loops, best of 3:
	min 38.3 msec per loop
	avg 39 msec per loop

[multiply_df] with 5 loops, best of 3:
	min 14.2 msec per loop
	avg 14.6 msec per loop
[multiply_numpy] with 5 loops, best of 3:
	min 20.5 msec per loop
	avg 21 msec per loop
[multiply_orangedf] with 5 loops, best of 3:
	min 14.3 msec per loop
	avg 14.5 msec per loop

[normalize_df] with 5 loops, best of 3:
	min 372 msec per loop
	avg 382 msec per loop
[normalize_df_numpy] with 5 loops, best of 3:
	min 97.6 msec per loop
	avg 99.3 msec per loop
[normalize_orangedf] with 5 loops, best of 3:
	min 386 msec per loop
	avg 393 msec per loop
[normalize_orangedf_numpy] with 5 loops, best of 3:
	min 84.7 msec per loop
	avg 85.6 msec per loop
[normalize_table] with 5 loops, best of 3:
	min 540 msec per loop
	avg 549 msec per loop

[revert_orangedf] with 5 loops, best of 3:
	min 203 msec per loop
	avg 207 msec per loop

```


```python
cols = 10000
rows = 100
```
```properties
[create_df] with 5 loops, best of 3:
	min 128 usec per loop
	avg 130 usec per loop
[create_orangedf] with 5 loops, best of 3:
	min 2.61 msec per loop
	avg 2.78 msec per loop
[create_table] with 5 loops, best of 3:
	min 1.29 msec per loop
	avg 1.36 msec per loop

[multiply_df] with 5 loops, best of 3:
	min 787 usec per loop
	avg 824 usec per loop
[multiply_numpy] with 5 loops, best of 3:
	min 748 usec per loop
	avg 774 usec per loop
[multiply_orangedf] with 5 loops, best of 3:
	min 768 usec per loop
	avg 778 usec per loop

[normalize_df] with 5 loops, best of 3:
	min 5.94 sec per loop
	avg 6.31 sec per loop
[normalize_df_numpy] with 5 loops, best of 3:
	min 8.01 msec per loop
	avg 8.31 msec per loop
[normalize_orangedf] with 5 loops, best of 3:
	min 5.84 sec per loop
	avg 6.03 sec per loop
[normalize_orangedf_numpy] with 5 loops, best of 3:
	min 6.42 msec per loop
	avg 6.57 msec per loop
[normalize_table] with 5 loops, best of 3:
	min 3.04 sec per loop
	avg 3.25 sec per loop

[revert_orangedf] with 5 loops, best of 3:
	min 202 msec per loop
	avg 218 msec per loop
```

irgolic:
The above benchmark is not incorrect, I've just rerun it again. On arrays with a lot of rows, we perform about as well as the DataFrame (x5 times slower on normalize). On arrays with a lot of columns, we're 1000x slower. It seems that we optimize operations for rows, and pandas manages to optimize for both rows and columns. (it could also be just the way normalize is implemented in Orange, please suggest more benchmarks)

If pandas DataFrames were completely interchangeable on the same piece of memory with Orange Tables, you could instantiate a DataFrame from a Table, run the math, and turn it back into a table. We're a couple steps away from that though. `compute_value` is set to `None` during DataFrame -> Table conversion (is this a problem?). More importantly, new df columns (not those which came from a Table -> DataFrame conversion) must sometimes be preprocessed, which creates a whole new numpy array.
I assume they must be preprocessed due to the way `table_from_frame` was already implemented:
```python
for name, s in df.items():
    if _is_discrete(s):
        discrete = s.astype('category').cat
        attrs.append(DiscreteVariable(name, discrete.categories.astype(str).tolist()))
        X.append(discrete.codes.replace(-1, np.nan).values)
    elif _is_datetime(s):
        tvar = TimeVariable(name)
        attrs.append(tvar)
        s = pd.to_datetime(s, infer_datetime_format=True)
        X.append(s.astype('str').replace('NaT', np.nan).map(tvar.parse).values)
    elif is_numeric_dtype(s):
        attrs.append(ContinuousVariable(name))
        X.append(s.values)
    else:
        metas.append(StringVariable(name))
        M.append(s.values.astype(object))
```
The above code creates a new numpy array for each discrete/datetime variable, but otherwise uses the original array. If we were to adjust Table/Variable such that it would be fully compatible with DataFrame's underlying numpy array, it could truly be used interchangeably with an overhead of practically constant complexity.

I've added support for sparse Tables as well. Pandas stores these column by column (coo format), so a new array is created during conversion. In Python Script, if a sparse array is passed in as data and automatic pandas conversion is chosen, a warning will be shown that this is inefficient.
```python
cols = 100
rows = 100000
```
```properties
[create_df] with 5 loops, best of 3:
	min 187 msec per loop
	avg 206 msec per loop
[create_orangedf] with 5 loops, best of 3:
	min 434 msec per loop
	avg 439 msec per loop
[create_table] with 5 loops, best of 3:
	min 10.9 msec per loop
	avg 11.1 msec per loop

[multiply_df] with 5 loops, best of 3:
	min 28.6 msec per loop
	avg 29.2 msec per loop
[multiply_df_numpy] with 5 loops, best of 3:
	min 59.9 msec per loop
	avg 60.7 msec per loop
[multiply_orangedf] with 5 loops, best of 3:
	min 27.3 msec per loop
	avg 27.7 msec per loop
[multiply_orangedf_numpy] with 5 loops, best of 3:
	min 58.9 msec per loop
	avg 60.6 msec per loop
[multiply_table] with 5 loops, best of 3:
	min 12.8 msec per loop
	avg 13.9 msec per loop

[normalize_df] with 5 loops, best of 3:
	min 246 msec per loop
	avg 251 msec per loop
[normalize_df_numpy] with 5 loops, best of 3:
	min 141 msec per loop
	avg 145 msec per loop
[normalize_orangedf] with 5 loops, best of 3:
	min 239 msec per loop
	avg 244 msec per loop
[normalize_orangedf_numpy] with 5 loops, best of 3:
	min 146 msec per loop
	avg 148 msec per loop
[normalize_table] with 5 loops, best of 3:
	min 601 msec per loop
	avg 617 msec per loop

[revert_orangedf] with 5 loops, best of 3:
	min 343 msec per loop
	avg 351 msec per loop

```

irgolic:
[Pandas support](https://trello.com/c/51ZRcAom/5-pandas-support)

Orange/data/pandas_compat.py
```diff
@@ -10,72 +16,310 @@
...
+class OrangeDataFrame(pd.DataFrame):
+    _metadata = ["orange_variables", "orange_weights",
+                 "orange_attributes", "orange_role"]
 
-    Parameters
-    ----------
-    df : pandas.DataFrame
-    force_nominal : boolean
-        If True, interpret ALL string columns as nominal (DiscreteVariable).
+    def __init__(self, *args, **kwargs):
```
markotoplak, line 28: Please make attributes explicit.

Orange/data/table.py
```diff
@@ -1873,3 +1905,31 @@ def assure_domain_conversion_sparsity(target, source):
     target.Y = match_density[conversion.sparse_Y](target.Y)
     target.metas = match_density[conversion.sparse_metas](target.metas)
     return target
+
+
+def vstack(arrs):
```
markotoplak, line 1910: `vstack` and `merge1d` were just moved, but are not used anywhere. Please move them back. If we were to expose them, we would need to properly document and test them.

Orange/data/table.py
```diff
@@ -970,6 +1001,7 @@ def ensure_copy(self):
         """
         Ensure that the table owns its data; copy arrays when necessary.
         """
+
```
markotoplak, line 1004: If you are linting unrelated parts of the code, use a separate commit, please. I would even prefer if you do not lint code you were not changing because then the history is easier to understand.

I think, of all Orange source code, I have been looking into the history of the Table the most. The clearer it is, the better.

markotoplak:
For now, I only skimmed the code and wrote some petty comments. I still did not seriously try it out and think about the interface...

Orange/data/table.py
irgolic: @janezd implemented them as part of Table.concatenate 4683cef2f314781be19d0989ddd88c5dc250936b, they have coverage.

I think I moved them here because they're only used here. I've prefixed them with `_`, similar to five other functions in the module.

markotoplak:
When I was looking into this today I had problems with the setters, with both (1) that the change the table and (2) that they are not atomic (separate change of the domain and then of the array could be problematic for multithreading). Then, in discussions with @VesnaT, we saw that we already have exact same problems with Table's value arrays (see #5303).

So now I think that the proposed interface is fine (#5303 needs to be handled separately). @lanzagar, what do you think of the interface added to the Table?

PR #5191, opened 2021-01-18
[FIX] PCA: Fix rounding error

### LINKED ISSUE #5157: Rounding error in PCA
- [x] **What's wrong?**

There is a rounding error in the PCA widget.
It comes up very rarely but it does appear sometimes.

A bit of patience will produce it with general workflows I think. 

- Orange version: Quasar 1.0.0
- How you installed Orange: Windows installer

image001

VesnaT:
##### Issue

FIxes #5157

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owpca.py
+++ b/Orange/widgets/unsupervised/owpca.py
@@ -211,7 +211,7 @@ class OWPCA(widget.OWWidget):
         if self._pca is not None:
             var = self._cumulative[components - 1]
             if numpy.isfinite(var):
-                self.variance_covered = int(var * 100)
+                self.variance_covered = round(var * 100, 0)
 
         self._invalidate_selection()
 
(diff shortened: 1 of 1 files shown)


markotoplak:
I am not sure that was a bug.

It seems so consistent in the code that it likely intentional. I investigated this a bit. This setting used to be named "Variance covered", and meant: give me enough components to cover the specified variance (and that usually means that you need to go slightly over it).

So these numbers in the left has a different meaning than the one in the graph, which means they will be inconsistent. For example, PCA on iris. If I move the selector, I get this.

image

Then, if I type 73 into examplained varience, I see that I need 2 components to examplain 73% of variance.

image

So I think that previous behavior was intentional and more useful. I suggest closing the issue and PR.

PR #5207, opened 2021-01-22
Fix pyqtgraph==0.11.1 test failures

markotoplak:
##### Issue
Some tests were failing with the new release of pyqtgraph. Our tests currrently pass with 0.11.0 or 0.11.1 or master.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -223,6 +223,27 @@ class ScatterPlotItem(pg.ScatterPlotItem):
             self._z_mapping = np.argsort(z)
             self._inv_mapping = np.argsort(self._z_mapping)
 
+    def setCoordinates(self, x, y):
+        """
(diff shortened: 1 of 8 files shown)


markotoplak:
The second-last commit ran tests with pyqtgraph 0.11.0, where all passed and coverage was OK.

Coverage on 0.11.1, that I enabled in the last commit, is worse, because some 0.11.0 workarounds are not needed there. So I think this coverage problems is not an issue.

VesnaT:
It seems ok, apart from one thing: when I upgrade to pyqtgraph==0.11.1, setting Legend's font size does not work any more.

To reproduce: 
- upgrade pyqtgraph 0.11.1
- Iris -> Scatter Plot -> Set visual settings -> Change Font size for `Categorical legend`.

markotoplak:
@VesnaT, thanks, I found the culprit.

PR #5220, opened 2021-01-29
domain: Deprecate __len__

irgolic:
##### Description of changes
As discussed last week, we deprecate `__len__` before we change its behavior along with `__iter__`'s, to iterate/len metas as well.

After this is released, merge #5190.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/domain.py
+++ b/Orange/data/domain.py
@@ -245,8 +245,13 @@ class Domain:
     def metas(self):
         return self._metas
 
+    @deprecated("len(Domain.variables)")
     def __len__(self):
-        """The number of variables (features and class attributes)."""
+        """The number of variables (features and class attributes).
+
+        The current behavior returns the length of only features and
+        class attributes. In the near future, it will include the
+        length of metas, too, and __iter__ will act accordingly."""
         return len(self._variables)
 
     def __bool__(self):


irgolic:
I've changed it most everywhere, but problem: some things rely on `__len__` (`assertSequenceEqual`, even `zip`). It seems that this only happens in tests (in Orange3), and won't be affected by the deprecation.

janezd:
What about casting to lists, e.g. `self.assertEqual(list(model), list(domain))`? Not elegant, but it works, and I know I already often do this because I forget that I can use `assertSequenceEqual`?

Alternatively, we can write `assertGeneratorEqual` (why doesn't Python already have it?!), but this would require subclassint (or monkey patching) `unittest.TestCase`.

irgolic:
> What about casting to lists, e.g. `self.assertEqual(list(model), list(domain))`? Not elegant, but it works, and I know I already often do this because I forget that I can use `assertSequenceEqual`?
> 
> Alternatively, we can write `assertGeneratorEqual` (why doesn't Python already have it?!), but this would require subclassint (or monkey patching) `unittest.TestCase`.

Good idea, I'll cast to list. Unideal, but we can work with this.

markotoplak:
And then we will deprecate `__iter__`... :D

PR #5221, opened 2021-01-29
[RFC] Use Slider, not QSlider

irgolic:
##### Issue
With the left side scrolling change, slider shouldn't interrupt scrolling anymore. Any instances of QSlider in the interface should be replaced by Slider.

##### Description of changes
Replaces some instances of QSlider with Slider.

Am I doing this right, import from orangewidget? Should I be doing it some other way?

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


irgolic:
Same should be done for spinboxes.

PR #5224, opened 2021-01-29
Multiple targets warning

### LINKED ISSUE #5144: Max 1 target variable per domain

I'm a new user, and I'm picking up data mining as I go. I don't know what meta attributes are, and I'm not quite sure what a target variable is. Can I set multiple target variables? 
Upon doing so and connecting a model, a `Data has no target variable` error appears. Confusing. Rank displays `Data does not have a (single) target variable`, slightly better.

Limit the number of target variables to one. If a new target variable is set, set the previous one to the role it previously played. Make sure this behavior is reflected in the file loading widget's domain editor and Select Columns. Make the Select Columns box 'target variable' box smaller, to fit one variable.

Better, unified error messages.

janezd:
##### Issue

Fixes #5144.

##### Description of changes

- File widget and Select Columns now show warnings when outputting data with multiple targets.
- Widgets that showed potentially misleading error (learners and freeviz) now explicitly state that the data has to have a **single** target.

##### Includes
- [X] Code changes
- [ ] Tests

### DIFF
--- a/Orange/widgets/utils/owlearnerwidget.py
+++ b/Orange/widgets/utils/owlearnerwidget.py
@@ -136,8 +136,14 @@ class OWBaseLearner(OWWidget, metaclass=OWBaseLearnerMeta, openclass=True):
         self.set_input_summary()
 
         if data is not None and data.domain.class_var is None:
-            self.Error.data_error("Data has no target variable.\n"
-                                  "You can set a target variable with the Select Columns widget.")
+            if data.domain.class_vars:
+                self.Error.data_error(
+                    "Data contains multiple target variables.\n"
+                    "Select a single one with the Select Columns widget.")
+            else:
+                self.Error.data_error(
+                    "Data has no target variable.\n"
+                    "Select one with the Select Columns widget.")
             self.data = None
 
         self.update_model()
--- a/Orange/widgets/visualize/owfreeviz.py
+++ b/Orange/widgets/visualize/owfreeviz.py
@@ -139,14 +139,16 @@ class OWFreeViz(OWAnchorProjectionWidget, ConcurrentWidgetMixin):
     graph = settings.SettingProvider(OWFreeVizGraph)
 
     class Error(OWAnchorProjectionWidget.Error):
-        no_class_var = widget.Msg("Data has no target variable")
+        no_class_var = widget.Msg("Data must have a target variable.")
+        multiple_class_vars = widget.Msg(
+            "Data must have a single target variable.")
         not_enough_class_vars = widget.Msg(
-            "Target variable is not at least binary")
+            "Target variable must have at least two unique values.")
         features_exceeds_instances = widget.Msg(
             "Number of features exceeds the number of instances.")
         too_many_data_instances = widget.Msg("Data is too large.")
         constant_data = widget.Msg("All data columns are constant.")
-        not_enough_features = widget.Msg("At least two features are required")
+        not_enough_features = widget.Msg("At least two features are required.")
 
     class Warning(OWAnchorProjectionWidget.Warning):
         removed_features = widget.Msg("Categorical features with more than"
@@ -257,10 +259,12 @@ class OWFreeViz(OWAnchorProjectionWidget, ConcurrentWidgetMixin):
 
         super().check_data()
         if self.data is not None:
-            class_var, domain = self.data.domain.class_var, self.data.domain
-            if class_var is None:
+            class_vars, domain = self.data.domain.class_vars, self.data.domain
+            if not class_vars:
                 error(self.Error.no_class_var)
-            elif class_var.is_discrete and len(np.unique(self.data.Y)) < 2:
+            elif len(class_vars) > 1:
+                error(self.Error.multiple_class_vars)
+            elif class_vars[0].is_discrete and len(np.unique(self.data.Y)) < 2:
                 error(self.Error.not_enough_class_vars)
             elif len(self.data.domain.attributes) < 2:
                 error(self.Error.not_enough_features)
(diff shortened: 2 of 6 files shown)


Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -136,7 +136,7 @@ def set_data(self, data):
         self.set_input_summary()
 
         if data is not None and data.domain.class_var is None:
-            self.Error.data_error("Data has no target variable.\n"
+            self.Error.data_error("Data must have a (single) target variable.\n"
```
irgolic, line 139: Can we be more explicit here? As in, a different message if there's no target vs if there are too many targets?

Something like `'Data must have one, not %d target variables.\n'`

ajdapretnar: Agree. I would put two different errors here. They require similar, but different actions (add / remove).

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -136,8 +136,14 @@ def set_data(self, data):
...
-            self.Error.data_error("Data has no target variable.\n"
-                                  "You can set a target variable with the Select Columns widget.")
+            if data.domain.class_vars:
+                self.Error.data_error(
+                    "Data contains multiple target variables.\n"
+                    "You can select a single one with Select Columns widget.")
+            else:
+                self.Error.data_error(
+                    "Data has no target variable.\n"
+                    "You can select one with Select Columns widget.")
```
irgolic, line 146: the Select Columns widget

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -136,8 +136,14 @@ def set_data(self, data):
         self.set_input_summary()
 
         if data is not None and data.domain.class_var is None:
-            self.Error.data_error("Data has no target variable.\n"
-                                  "You can set a target variable with the Select Columns widget.")
+            if data.domain.class_vars:
+                self.Error.data_error(
+                    "Data contains multiple target variables.\n"
+                    "You can select a single one with Select Columns widget.")
```
irgolic, line 142: the Select Columns widget

Orange/widgets/visualize/owfreeviz.py
```diff
@@ -139,14 +139,16 @@ class OWFreeViz(OWAnchorProjectionWidget, ConcurrentWidgetMixin):
     graph = settings.SettingProvider(OWFreeVizGraph)
 
     class Error(OWAnchorProjectionWidget.Error):
-        no_class_var = widget.Msg("Data has no target variable")
+        no_class_var = widget.Msg("Data must have a target variable.")
+        multiple_class_vars = widget.Msg(
+            "Data must have just a single target variable.")
         not_enough_class_vars = widget.Msg(
-            "Target variable is not at least binary")
+            "Target variable is not at least binary.")
```
irgolic, line 146: Target variable must have at least two unique values?

Orange/widgets/utils/owlearnerwidget.py
janezd: I intentionally removed `the` to make it shorter. But OK, I added it back.

irgolic: Honestly, I didn't open the widget and check. Generally, I'm not too worried about the message length, they're visible on hover/click.

Orange/widgets/visualize/owfreeviz.py
```diff
@@ -139,14 +139,16 @@ class OWFreeViz(OWAnchorProjectionWidget, ConcurrentWidgetMixin):
     graph = settings.SettingProvider(OWFreeVizGraph)
 
     class Error(OWAnchorProjectionWidget.Error):
-        no_class_var = widget.Msg("Data has no target variable")
+        no_class_var = widget.Msg("Data must have a target variable.")
+        multiple_class_vars = widget.Msg(
+            "Data must have just a single target variable.")
```
ajdapretnar, line 144: "Just" is likely redundant. I prefer:
- "Data must have a single target variable."
- "Data must have only one target variable."

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -136,8 +136,14 @@ def set_data(self, data):
         self.set_input_summary()
 
         if data is not None and data.domain.class_var is None:
-            self.Error.data_error("Data has no target variable.\n"
-                                  "You can set a target variable with the Select Columns widget.")
+            if data.domain.class_vars:
+                self.Error.data_error(
+                    "Data contains multiple target variables.\n"
+                    "You can select a single one with the Select Columns widget.")
```
ajdapretnar, line 142: "You can select a single one with the Select Columns widget." --> "Select a single one with Select Columns." (shorter?)

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -136,8 +136,14 @@ def set_data(self, data):
...
+            else:
+                self.Error.data_error(
+                    "Data has no target variable.\n"
+                    "You can select one with the Select Columns widget.")
```
ajdapretnar, line 146: Same as above.

Orange/widgets/utils/owlearnerwidget.py
janezd: "You can" was meant to sound as suggestion, just one of possible options. It obviously didn't, so I removed it.

Orange/widgets/visualize/owfreeviz.py
janezd: OK. I chose the first one, the shorter.

Orange/widgets/utils/owlearnerwidget.py
ajdapretnar: I started from the position that I would like to use the learner, so I *have* to have just a single target variable. 😁

PR #5226, opened 2021-01-29
OWSelectColumns: Fixup UI

irgolic:
Screenshot 2021-02-01 at 21 29 24

(might not look the same with a non-master version of orange-widget-base)
##### Description of changes
- remove Up/Down buttons
- fit target box to contents
- reorganize option and autocommit into buttonsArea
- simplified names of boxes

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owselectcolumns.py
+++ b/Orange/widgets/data/owselectcolumns.py
@@ -195,6 +195,8 @@ class OWSelectAttributes(widget.OWWidget):
 
         self.controlArea = QWidget(self.controlArea)
         self.layout().addWidget(self.controlArea)
+
+        # init grid
         layout = QGridLayout()
         self.controlArea.setLayout(layout)
         layout.setContentsMargins(4, 4, 4, 4)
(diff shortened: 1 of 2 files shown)


irgolic:
Also, I know we've spent hours on the new checkbox

but can I suggest we invert it? `Ignore new variables by default`, something like that? It seems easier to word when it's inverted. The word 'select' doesn't really connect with the UI (it should mention Features).

markotoplak:
I added this to the milestone. As I understood, the checkbox was not released yet, so we do not need to worry about setting migration (please double check this). Perhaps also use a new name for the setting.

PrimozGodec:
> but can I suggest we invert it? Ignore new variables by default, something like that? It seems easier to word when it's inverted. The word 'select' doesn't really connect with the UI (it should mention Features).

Implemented in https://github.com/biolab/orange3/pull/5237

irgolic:
Merged with https://github.com/biolab/orange3/pull/5013

PR #5208, opened 2021-01-24
[ENH] OWPythonScript: Better text editor

irgolic:
Requires https://github.com/biolab/orange3/pull/5526 and https://github.com/biolab/orange3/pull/5530.

##### OWPythonScript PRs:
here -> #5228

##### Description of changes
Screenshot 2021-01-24 at 18 14 36
I've separated the Python Script rewrite into several PRs. Here's the first, and probably biggest. :) Coming soon are a PR replacing the console, a PR replacing the library, and a PR integrating pandas.

Don't fret the number of added lines, there's hundreds of tests.

In this PR, andreikop/Qutepart is ported into Orange and adjusted for the python script widget. 

Also, a fake function signature adorns the editor, showing what variables are on the input. A return statement in the script will be used for variable collection in a later PR, when that's reworked. A later PR will also include completions in the editor.

RFC because please give it a try, let me know if any keyboard shortcuts/vim mode should be adjusted.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #5228, opened 2021-01-31
Pythonscript console

irgolic:
##### OWPythonScript PRs:
#5208 -> here

##### Issue
The console hangs on execution. Also it could be prettier.

##### Description of changes

Screenshot 2021-01-31 at 21 21 32

The console now launches a python kernel in a separate process. 

Some new files:
- python_serialize.py
This is variable injection ZMQ stuff between console and kernel. It facilitates a pipe-based IPC (inter-process communication) solution. Makes a great prototype for client/server architecture in Orange. **I have conflicting information on whether this works on Windows or not. It'd be great if somebody could try it out, pls :)**.
- python_console.py
Uses python_serialize.py's OrangeZMQMixin. Also does a lot of work maintaining current state (transferring variables, executing script), also triggering signals that indirectly change the widget's status message.
- python_kernel.py
Uses python_serialize.py's OrangeZMQMixin. Simplest of the bunch, just make sure not to import Qt or Orange here or it hangs the kernel on startup (making it timeout).

I tried making the kernel fast, but if the timeout an issue, I can extend it. From what I've seen, it only causes a small UI glitch.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


irgolic:
[Better OWPythonScript](https://trello.com/c/uT4ni1qS/4-better-owpythonscript)

markotoplak (changes requested):
We need to ensure that old workflows keep the same functionality. For old workflows, the in-process kernel needs to be used.

To achieve this, the settings version needs to increase in this PR, and then, if old settings are detected, enable the old kernel for that process. If there are other changes to the behavior (I might have missed something), also consider a compatibility mode as in #3810.

This is a serious blocker, but fortunately easy to fix.

PR #5231, opened 2021-02-01
owheatmap: replace item.rotate with proper setTransform

markotoplak:
Before, heatmap was using pyqtgraph's monkey-patched methods, which they now removed (see pyqtgraph/pyqtgraph#1533). This PR uses proper PyQt5 interface.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/utils/heatmap.py
+++ b/Orange/widgets/visualize/utils/heatmap.py
@@ -405,7 +405,7 @@ class HeatmapGridWidget(QGraphicsWidget):
         for i, rowitem in enumerate(parts.rows):
             if rowitem.title:
                 item = QGraphicsSimpleTextItem(rowitem.title, parent=self)
-                item.rotate(-90)
+                item.setTransform(item.transform().rotate(-90))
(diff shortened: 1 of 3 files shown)


Orange/widgets/visualize/utils/heatmap.py
```diff
@@ -405,7 +405,7 @@ def setHeatmaps(self, parts: 'Parts') -> None:
         for i, rowitem in enumerate(parts.rows):
             if rowitem.title:
                 item = QGraphicsSimpleTextItem(rowitem.title, parent=self)
-                item.rotate(-90)
+                item.setTransform(item.transform().rotate(-90))
```
pijyoi, line 408: You could consider also "item.setRotation(-90)" if that's the only transformation needed.

markotoplak:
This fixes tests on pyqtgraphs master and one bug I randomly found in our Scatter Plot. There are probably other bugs. Please use Orange on pyqtgraph's master from time to time. :)

Issue #5108, opened 2020-12-04
Automatically set input/output summary


janezd:
@irgolic said that we could get rid of duplicated code for setting and testing summaries (see #5002) by putting the code for setting the summaries and for calling them into `Orange.widgets.widget.OWWidget`. Discussion (@irgolic, @markotoplak, @janezd) led to the following.

1. Summary is set automatically under the following conditions:
    - widget has the default signal
    - the default signal is of type `Table`
    - widget does not set the flag, named sth like `auto_set_summary` (but better) to `False` (default is `True`).
2. For the sake of backward compatilibility or perhaps also for future convenience, calling `set_input_summary`/`set_output_summary` sets this flag to `False`.
3. The redundant code is removed from widgets, together with corresponding tests (see #5002).

Afterthoughts:

If we do this, we can probably close https://github.com/biolab/orange-widget-base/pull/101?

Output is easy. The obvious place to trigger this for output is `Orange.widgets.widget.OWWidget.send`. 

For input, there is no such spot. `orangewidget.utils.signals.InputSignal.__call__` is a decorator that registers input, but returns the method essentially undecorated. We could derive a new class, `Orange.widgets.widget.Input` that would wrap the method into one that would call `set_input_summary`. However, this solution not only introduces magic, but may also be inconsistent with what happens in the widget: the input signal handler may reject the data (e.g. sth. like `self.data = None`), but the summary would still be set.

janezd:
Done in #5308.

PR #5233, opened 2021-02-03
owtable: Select all rows by default

irgolic:
Bases off #5276 

##### Issue
- [X] **What's your use case?**

It's intuitive to connect `File -> Table -> ...`.

Usage statistics indicate that most Orange users do this, as seen in https://github.com/biolab/orange-canvas-core/pull/154#issuecomment-727269185. I've seen people routinely connect a data table, CTRL+A select all rows, and connect on from there. Some opt to use the Data signal, but that adds an additional column, useless in this use case.

It is true, that it's more correct to connect both Data Tables and other widgets from the data source. But users learn by experimenting. And when you find something that works, you stick with it, and don't even consider finding an alternative. The cognitive load of connecting from data source is greater than that of connecting from Data Table + selecting all.

- [X] **What's your proposed solution?**

If we make it easy to connect from Data Table, we will help flatten the learning curve of Orange (and data science itself).

We can encourage people to always look at their data.

Also in some cases it just looks nicer:

###### _"Correct way"_

Screenshot 2021-02-03 at 00 50 52

###### _"Incorrect way"_

Screenshot 2021-02-03 at 00 51 25

- [X] **Are there any alternative solutions?**

If we insist on Data Table implementing a Visualization/Selection interface, it should be moved to the Visualize category.

##### Description of changes
Adds a `Setting(True)` for sending all data in the Selected Data signal, should no data be selected.

Screenshot 2021-02-03 at 00 56 32

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -739,9 +739,9 @@ class OWDataTable(OWWidget):
                 metas = select_vars(TableModel.Meta)
                 domain = Orange.data.Domain(attrs, class_vars, metas)
 
-            # Avoid a copy if none rows are selected.
+            # Send all data by default
             if not rowsel:
-                selected_data = None
+                selected_data = table
             else:
                 selected_data = table.from_table(domain, table, rowsel)
 
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -196,6 +196,8 @@ class OWDataTable(OWWidget):
     selected_rows = Setting([], schema_only=True)
     selected_cols = Setting([], schema_only=True)
 
+    settings_version = 2
+
     def __init__(self):
         super().__init__()
 
--- a/Orange/widgets/data/tests/test_owconcatenate.py
+++ b/Orange/widgets/data/tests/test_owconcatenate.py
@@ -5,6 +5,8 @@ from unittest.mock import patch, Mock
 
 import numpy as np
 
+from orangewidget.widget import StateInfo
+
 from Orange.data import (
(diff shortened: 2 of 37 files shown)


Orange/widgets/data/owtable.py
```diff
@@ -230,6 +231,8 @@ def __init__(self):
 
         gui.checkBox(box, self, "select_rows", "Select full rows",
                      callback=self._on_select_rows_changed)
+        gui.checkBox(box, self, "send_all_by_default", "Select all by default",
```
janezd, line 234: This label is somewhat misleading. Rows are not selected by default. All rows are sent if none are selected. Though I don't like any of the following two selections much: "Send all data on empty selection", "Send all rows when none are selected".

What would happen if we don't have a checkbox at all, but instead always send all data? (If we have already discussed this, I surely hope the checkbox wasn't my suggestion. :)

irgolic: I'm down, no checkbox is even better. :)

ajdapretnar: Agree, I was just about to say the checkbox is redundant.

irgolic:
https://github.com/biolab/orange3/pull/5002 can be merged as is, but I'm going to rewrite summary tests #5108 before we merge this.

markotoplak:
@janezd, before you merge this, could you please also increment `settings_version` of OWTable?

This will allow us to handle any migrations properly if we need them. If we perhaps discover that users are now having problems with saved workflows or if we perhaps wanted to roll-back this feature. :)

PR #5245, opened 2021-02-07
Properly separate warnings by threads

### LINKED ISSUE #5008: Warnings can get mixed up between widgets
I have a specific configuration where I get an unexpected read warning in the File widget.

To reproduce:
1. File (heart disease)
2. Test & Score, LR as learner
3. Set LOO in T&S.
4. Let T&S run a bit, then change the data set to iris.
5. Read warning occurs.

```
lbfgs failed to converge (status=1):
STOP: TOTAL NO. of ITERATIONS REACHED LIMIT.

Increase the number of iterations (max_iter) or scale the data as shown in:
    https://scikit-learn.org/stable/modules/preprocessing.html
Please also refer to the documentation for alternative solver options:
    https://scikit-learn.org/stable/modules/linear_model.html#logistic-regression
```

Probably something that should be handled more gracefully.

janezd:
##### Issue

Fixes #5008: the File widget seems to be the only widget that uses `catch_warnings` to record and show warnings.

##### Description of changes

The code is a bit of a hack: it's based on checking stack frames, and it patches an internal function in `warnings`. I don't see a way around any of that.

@markotoplak, can you check? If you confirm, I can add tests for context manager (besides the current tests for the widget).

There are two other potential problems

- `catch_warnings(record=True)` is used in some tests. If these are executed in parallel, they may "steal" each other's warnings.
- We call `catch_warnings()` to supress warnings in various places. Two are in code (`data.io` and the Rank widget), others in tests. This is not thread safe, but won't have very noticeable consequences.
- `catch_warnings()` is called in libraries we use. We have no control over that, and can't fix that; Python has to do it.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owfile.py
+++ b/Orange/widgets/data/tests/test_owfile.py
@@ -1,7 +1,8 @@
 # Test methods with long descriptive names can omit docstrings
-# pylint: disable=missing-docstring
+# pylint: disable=missing-docstring,protected-access
 from os import path, remove, getcwd
 from os.path import dirname
+import unittest
 from unittest.mock import Mock, patch
 import pickle
 import tempfile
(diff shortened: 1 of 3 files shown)


Orange/widgets/data/tests/test_owfile.py
```diff
@@ -603,3 +604,34 @@ def test_files_relocated(self):
             self.assertEqual(w.recent_paths[0].relpath, base_name)
         finally:
             remove(file_name)
+
+    @patch("os.path.exists", new=lambda _: True)
+    def test_warning_from_another_thread(self):
+        class AnotherWidget(QObject):
+            @staticmethod
+            def issue_warning():
```
ales-erjavec, line 612:
```suggestion
            def issue_warning(self):
```

janezd:
Wow. I would never notice that. What a trap! Thanks!

PR #5246, opened 2021-02-09
__main__: Debug/Profile in PyCharm without error

irgolic:
This is annoying because crash recovery isn't invoked when debugging/profiling.
(I use crash recovery while developing)
OASYS actually had this one figured out 4 years ago.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Works like Charm.

PR #5251, opened 2021-02-09
Table: Add methods 'join' and 'with_column'

### LINKED ISSUE #5115: API: add function for adding new columns
I wish for an easy way to add new columns in Orange scripting.

Something along the lines of:
```
iris = Table('iris')
iris.add_column(var, values)
```

I was taught the way to add columns in Orange is:
```
from Orange.data import Table, Domain, ContinuousVariable

zoo = Table('zoo')
values = ['foo' for i in range(len(zoo))]
new_var = ContinuousVariable(name='String')
new_domain = Domain(attributes=zoo.domain.attributes,
                    metas=zoo.domain.metas+(new_var,))
out_data = zoo.transform(new_domain)
col, _ = out_data.get_column_view(new_var)
col[:] = values
```

Looking at the script closely, you see that it shouldn't work. But it does (no objections from Orange). As is expected. But the limitation of this method is that one can make silly mistakes as above (I forgot to replace ContVar with StrVar) without noticing. It would be lovely if Orange provided an API for adding new columns, which would check for such inconsistencies.

janezd:
##### Issue

Resolves #5115.

Please comment the method names. 'join' reminds of SQL's join, but it's kind of correct.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -898,11 +898,11 @@ class Table(Sequence, Storage):
     @classmethod
     def concatenate(cls, tables, axis=0):
         """
-        Concatenate tables into a new table, either horizontally or vertically.
+        Concatenate tables into a new table, either vertically or horizontally.
 
-        If axis=0 (horizontal concatenate), all tables must have the same domain.
+        If axis=0 (vertical concatenate), all tables must have the same domain.
 
-        If axis=1 (vertical),
+        If axis=1 (horizontal),
         - all variable names must be unique.
         - ids are copied from the first table.
         - weights are copied from the first table in which they are defined.
@@ -915,12 +915,28 @@ class Table(Sequence, Storage):
         Returns:
             table (Table)
         """
-        if axis == 0:
-            return cls._concatenate_vertical(tables)
-        elif axis == 1:
-            return cls._concatenate_horizontal(tables)
-        else:
+        if axis not in (0, 1):
             raise ValueError("invalid axis")
+        if not tables:
+            raise ValueError('need at least one table to concatenate')
+
+        if len(tables) == 1:
+            return tables[0].copy()
+
+        if axis == 0:
+            conc = cls._concatenate_vertical(tables)
+        else:
+            conc = cls._concatenate_horizontal(tables)
+
+        # TODO: Add attributes = {} to __init__
+        conc.attributes = getattr(conc, "attributes", {})
+        for table in reversed(tables):
+            conc.attributes.update(table.attributes)
+
+        names = [table.name for table in tables if table.name != "untitled"]
+        if names:
+            conc.name = names[0]
+        return conc
 
     @classmethod
     def _concatenate_vertical(cls, tables):
@@ -941,10 +957,6 @@ class Table(Sequence, Storage):
         def collect(attr):
             return [getattr(arr, attr) for arr in tables]
 
-        if not tables:
-            raise ValueError('need at least one table to concatenate')
-        if len(tables) == 1:
-            return tables[0].copy()
         domain = tables[0].domain
         if any(table.domain != domain for table in tables):
             raise ValueError('concatenated tables must have the same domain')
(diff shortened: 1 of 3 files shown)


ajdapretnar:
Thanks for this one! 🎉 

Few comments:
1. I would prefer to name the method _add_column_. _with_column_ really confused me.
2. I would add docstrings. 🤷 (I can do it, if you want.)
3. Why does this fail?
```
iris = Table(iris)
iris1 = iris[:,:2]
iris2 = iris[:,2:]
Table.join(iris1, iris2)
```

janezd:
> 1. I would prefer to name the method _add_column_. _with_column_ really confused me.

Inspired by Kotlin's convention (e.g. `withIndex`). I'm not too happy with this name, but I didn't like `_add_column` and similar, because it suggests that the method adds a column to *this* table. I'll be happy to rename, but the name mustn't mislead.
 
> 2. I would add docstrings. 🤷 (I can do it, if you want.)

Was planned. Now done.

> 3. Why does this fail?
> 
> ```
> iris = Table(iris)
> iris1 = iris[:,:2]
> iris2 = iris[:,2:]
> Table.join(iris1, iris2)
> ```

Wow. I was writing the tests and the code in parallel, and solved some problem by using an empty array of appropriate dimensions, but I accidentally just used the same dimension as the array in tests, so`empty((5, 0))` was hardcoded. :)))

lanzagar:
> so empty((5, 0)) was hardcoded. :)))
Everybody knows that hardcoded things should be fitted to iris. Then nobody complains for a long time, because when they test the code it always works ;)

I also don't particularly like the name `with_column`. Maybe the effort to be so precise is a bit unneeded here, since unlike the differences between sort/sorted and similar we are talking about a method of the Table class which AFAIK we decided should be immutable in all such cases. Maybe that sacrifice at least allows us to use the most obvious name and not be misleading?

On the other hand `join` is a bit misleading for me, because I would definitely expect the option to join based on a unique identifier and not just sequentially. In connection to this: the method above (`concatenate`) used to allow axis=1 which we removed - do we remember the reason? And should we re-enable that now by calling `join` when axis=1? Or to avoid the name `join` just not have a separate method at all and use concatenate(..., axis=1) like in the olden days?
Sidenote: if we do keep two methods, I do not like the inconsistent signatures  `*tables` vs `tables`

And finally: YAY, we are getting a method to add columns :)

janezd:
#### add column

OK, I'll try one last appeal: doesn't this look nice?

```python
amended = iris.with_column(label_var, label_col)
```

As opposed to

```python
amended = iris.add_column(label_var, label_col)
```

In the second case, I'd expect `amended` to be `None`. (But it may be just me. Yesterday I thought for a moment that `ndarray.flatten` works in place!)

If I haven't convinced you, say so and I'm renaming it to `add_column`.

#### join

I wasn't sure about `join`, either. I still decided to have it as analogous to `str.join`. 

But I agree on merging this with `concatenate` with `axis=1` (and obviously using `tables` rather than `*tables`)

I forgot the reason for it's removal. I believe it was a mess. We probably thought it would be cleaner and didn't realize how much we'd miss it.

ajdapretnar:
I see your point, but I am somehow not used to methods describing the results, but rather the operation. For example, `table.to_dense()` returns a dense table. If it described the result, it would be `table.as_dense()`. `concatenate()` and `join()` both describe the action. If they described the result, they would be `concatenated()` and `joined()`. Or am I mistaken? Perhaps @markotoplak and @VesnaT can add their two cents.

As for `join()`, I was equally surprised it didn't accept the key. pandas' `join()` joins by key if given, else it joins by index. Is this the right moment to reiterate how pandas will solve everything?

janezd:
Decision: `concatenate` with `axis`; `add_column`.

Orange/data/table.py
```diff
@@ -897,7 +897,33 @@ def __repr__(self):
 
     @classmethod
     def concatenate(cls, tables, axis=0):
-        """Concatenate tables into a new table"""
+        """
+        Concatenate tables into a new table, either horizontally or vertically.
```
lanzagar, line 901: Maybe change the order here to "vertically or horizontally" to have the default mentioned first and match the order bellow.

Orange/data/table.py
```diff
@@ -942,6 +966,65 @@ def collect(attr):
...
+        for W in Ws:
+            if W.size:
+                break
+
+        parts = all_of(doms, ("attributes", "class_vars", "metas"))
+        domain = Domain(*(tuple(chain(*lst)) for lst in parts))
+        table = cls.from_numpy(domain, Xs, Ys, Ms, W, ids=tables[0].ids)
+        for ta in reversed(table_attrss):
+            table.attributes.update(ta)
+
```
lanzagar, line 998: The name of the new table should also be set, otherwise it is 'untitled'.
The same approach can be used as for axis=0:
```
names = [table.name for table in tables if table.name != "untitled"]
if names:
    table.name = names[0]
```

lanzagar (changes requested):
We should probably think about using the new functions (especially add_column) in the many places where it is done manually now.
We obviously don't need to wait and can merge this beforehand, after the minor changes mentioned (docstring and name).

Orange/data/table.py
janezd: I extracted the common code from `_concatenate_vertical` and `_concatenate_horizontal` to `concatenate`.

All changes are in the second commit.

lanzagar (approved):

PR #5252, opened 2021-02-10
[ENH] Violin Plot: New widget

### LINKED ISSUE #5215: Violin Plot
Following is a proposal for a new widget that displays the violin plot. Notice that the widget has a resemblance to Box Plot, and is indeed an alternative way to show variable distributions. The need for the widget was raised by several important collaborators, including Genialis, and hence this issue and proposal for the creation of the widget.

image

- Variable and Subgroups panes are exactly like those in the Box Plot
- Box plot, Strip plot and Rug plot options trigger additional display of box plot, strip plot and rug plot. If shown together, box plot is on top, and strip plot is on the bottom. See a [guide](https://chartio.com/learn/charts/violin-plot-complete-guide/) for these kind of displays, and use the display style similar to that of [seaborn](https://seaborn.pydata.org/generated/seaborn.violinplot.html). The defaults are as shown in the figure above.
- Ordering subgroups orders the subgroup plot by increasing mean. See also the [guide](https://chartio.com/learn/charts/violin-plot-complete-guide/). This check box is disabled if Subgroups=None.
- Orientation can be vertical or horizontal. See example charts in [seaborn](https://seaborn.pydata.org/generated/seaborn.violinplot.html).
- Density estimations can use "Normal kernel", "Epanechnikov kernel" or "Triangular kernel". In the first approximation of the widget, this could be omitted and a default normal kernel could be used insted.

It would be great if the widget could support data selection. If only one violin is shown, than dragging over the violin can mark a variable range and selected the corresponding data points. If, in case of the subgroups, several violins are displayed, then the user can drag over plots that display the distributions for various subgroups, and modifier keys should be used to select data points from two or more subgroups.

VesnaT:
##### Issue

Fixes #5215

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owviolinplot.py
+++ b/Orange/widgets/visualize/owviolinplot.py
@@ -1,6 +1,6 @@
 from collections import namedtuple
 from itertools import chain, count
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Set
 
 import numpy as np
 from scipy import stats
@@ -8,7 +8,8 @@ from sklearn.neighbors import KernelDensity
 
 from AnyQt.QtCore import QItemSelection, QPointF, QRectF, QSize, Qt, Signal
 from AnyQt.QtGui import QBrush, QColor, QPainter, QPainterPath, QPolygonF
-from AnyQt.QtWidgets import QCheckBox, QSizePolicy, QWidget
+from AnyQt.QtWidgets import QCheckBox, QSizePolicy, QGraphicsRectItem, \
+    QGraphicsSceneMouseEvent, QApplication, QWidget
 
 import pyqtgraph as pg
 
@@ -23,7 +24,6 @@ from Orange.widgets.settings import ContextSetting, DomainContextHandler, \
 from Orange.widgets.utils.annotated_data import ANNOTATED_DATA_SIGNAL_NAME, \
     create_annotated_table
 from Orange.widgets.utils.itemmodels import VariableListModel
-from Orange.widgets.utils.plot import PANNING, SELECT, ZOOMING
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.utils.state_summary import format_summary_details
 from Orange.widgets.visualize.owboxplot import SortProxyModel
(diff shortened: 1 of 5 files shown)


VesnaT:
image

Orange/widgets/visualize/owviolinplot.py
```diff
@@ -0,0 +1,1091 @@
...
+            box, self, "order_violins", "Order subgroups",
+            callback=self.__order_violins_changed,
+        )
+        gui.radioButtons(box, self, "orientation_index",
+                         ["Horizontal", "Vertical"], label="Orientation: ",
+                         orientation=Qt.Horizontal,
+                         callback=self.__orientation_changed)
+
+        box = gui.vBox(self.controlArea, "Density Estimation",
+                       sizePolicy=(QSizePolicy.Minimum, QSizePolicy.Maximum))
```
irgolic, line 832: Why override sizePolicy?

VesnaT: Because:
image

ajdapretnar:
Icon is available once rebased. :)

Orange/widgets/visualize/owviolinplot.py
```diff
@@ -0,0 +1,1092 @@
...
+            self.selection = indices
+            self.commit()
+
+    def _add_controls(self):
+        self._value_var_model = VariableListModel()
+        sorted_model = SortProxyModel(sortRole=Qt.UserRole)
+        sorted_model.setSourceModel(self._value_var_model)
+        sorted_model.sort(0)
+
+        view = self._value_var_view = ListViewSearch()
```
markotoplak, line 780: These ListViewSerach boxes need a greater minimum height. Now, these selectors are useless if the widget is small enough that the controlArea scrollbar is shown. 

image

Why are they so small?

BlazZupan:
Nice widget! When the feature we use for subgrouping that is originally numerical is discretized, the axis labels overlap and become unreadable. Here I used the housing data set and discretized CRIM.

image

BlazZupan:
Labels overlap when subgrouping variable has enough values, say, over five. I am using adult data set here. This is perhaps generic problem for most similar widgets. Labels should be displayed vertically in such cases, and I am not sure if this should be corrected in this widget alone.

image

BlazZupan:
If the variable has only one single value, the plot remains empty. That is, the box plot is too tiny to be displayed (probably the height of the box plot is 0). In such cases, strip plot would display the points, and rug plot would be empty as well (no lines displayed). 

Should we set the minimum height of the box plot? Why is there no rug plot (a single line) in such case?

I am using adult data in this example.

image

markotoplak:
A possible solution for overlapping labels. Saves space compared to verticals and is probably easier to read.

image

VesnaT:
How about something like that?

image

Should not be difficult to apply since already implemented.

Issue #4919, opened 2020-07-23
Store recent paths in QSettings, not .ows


irgolic:
File (and other file loading widgets) show you the most recently chosen files in a dropdown.

This is currently stored in the .ows file itself. I believe this should be user-dependent, saving it in QSettings would make more sense.

janezd:
Only the currently opened file should be in the schema, others in QSettings.

janezd:
Also, @irgolic suggested that absolute path should never be stored in schema. If the recent path is absolute, it should be in QSettings.

janezd:
The widget currently does not store the active sheet name. This is non-trivial to fix because it would have to be a context setting, but contexts currently correspond to domain, not file name.

The fix should, however, be a part of solving this issue because file name should be in `QSettings`.

janezd:
I would really like to have this implemented - I even wrote a script that I use for cleaning the history. I wrote a draft PR,  #7228. The change actually simplifies the file widget a bit, and adds a mix-in that is no more complicated than the existing. (But perhaps breaks something.)

I wonder, though, if we shouldn't just make this a schema-only setting. I can imagine a scenario in which somebody saves a workflow with three data sets to choose from in the file widget, and sends this to a colleague. Having them in a combo is practical.

Options:
1. We store recent files in `QSettings`, as proposed above and implemented in #7228.
2. We change the setting to schema-only.
3. We change the settings to schema-only and add a "Clear recent files history" button.
4. Widget's GUI is somewhat unusual. It would be more common to have just "Open file" (like our "..." button) and "Recent files" selector that would change the open file, but would look like a menu rather than combo. This complicates the code a bit (but doesn't really confuse the user, imho).

I am undecided between 1-3.

PR #5275, opened 2021-02-17
Show correct initial sheet name

### LINKED ISSUE #5127: File: Excel with multiple sheets shows wrong one upon loading
- [ ] **What's wrong?**

When loading an Excel file with multiple sheets, the last sheet is loaded, but the dropdown shows as if the first one is shown.

Load an Excel file with multiple sheets and observe the Sheet value.

[exc-multi-sheets.xlsx](https://github.com/biolab/orange3/files/5695852/exc-multi-sheets.xlsx)

- Operating system: OSX High Sierra
- Orange version: 3.28.dev
- How you installed Orange: conda/pip

janezd:
##### Issue

Fixes #5127.

For xlsx, the initially selected sheet will be the one that is stored as active in the .xlsx. For Xls, it's gonna be the first sheet.

Active sheet should be stored as setting. This is however a bit complicated (see https://github.com/biolab/orange3/issues/4919#issuecomment-780749240)  and not worth fixing because it would have to be reimplemented anyway when fixing #4919.

##### Description of changes

Readers' constructors set `self.sheet`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -414,7 +414,7 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
             self.sheet_combo.setCurrentIndex(idx)
         except ValueError:
             # Requested sheet does not exist in this file
-            self.reader.select_sheet(0)
+            self.reader.select_sheet(None)
             self.sheet_combo.setCurrentIndex(0)
 
     @staticmethod
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -409,14 +409,12 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
         self.sheet_box.show()
 
     def _select_active_sheet(self):
-        if self.reader.sheet:
-            try:
-                idx = self.reader.sheets.index(self.reader.sheet)
-                self.sheet_combo.setCurrentIndex(idx)
-            except ValueError:
-                # Requested sheet does not exist in this file
-                self.reader.select_sheet(None)
-        else:
+        try:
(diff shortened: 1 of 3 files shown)


markotoplak:
I think we can save it. If the sheet is saved as schema-only, we would solve the immediate problem of loading workflows correctly. We just need to reset it when a new file is chosen.

Sure, swapping to another file and back and keeping the sheet would require file-name-based contexts, but that is probably secondary.

janezd:
My mistake: sheets are already saved in recent paths. Nothing to fix there - except for migrating to `QSettings`.

Orange/widgets/data/owfile.py
```diff
@@ -409,14 +409,12 @@ def _update_sheet_combo(self):
...
-            except ValueError:
-                # Requested sheet does not exist in this file
-                self.reader.select_sheet(None)
-        else:
+        try:
+            idx = self.reader.sheets.index(self.reader.sheet)
+            self.sheet_combo.setCurrentIndex(idx)
+        except ValueError:
+            # Requested sheet does not exist in this file
+            self.reader.select_sheet(None)
```
markotoplak, line 417: Your PR used `select_sheet(0)` here. Was that intentional? I added a commit that changes it back to `None` (which is the current value for unknown/default sheet).

markotoplak: @janezd

janezd: (I thought I answered this, but now I can't find the answer. Perhaps I forgot to submit it.)

I guess this was needed at some iteration, but now seems to works without. You can merge with your change.

Problems on macOS are unrelated and also appeared in other PRs. I'll investigate.

PR #5277, opened 2021-02-18
SliderGraph: Disable auto scale button

### LINKED ISSUE #5270: PCA: scree plot formatted strangely when resetting zoom
- [ ] **What's wrong?**

Scree plot in PCA is not nicely formatted upon resetting the zoom. Also the new curve labels fall out of view when resizing the widget.

File (brown-selected).
PCA. Press the reset zoom option in the bottom left corner of the plot. This happens:
Screen Shot 2021-02-17 at 10 48 08
Also, component variance and cumulative variance fall out of view when the widget size is small. Perhaps set them within the two curves, not on the outside.

- Operating system: OSX
- Orange version: 3.28.dev
- How you installed Orange: conda

janezd:
##### Issue

Fixes #5270. Twice.

##### Description of changes

First, it disables the auto scale button. This graph (used here and in dbscan) does not need it.
Second, it fixes the problem that caused the wrong scaling (lines going from -100 instead of from minimal x).

The second fix alone would also suffice, though auto scaling would still change the plot's y scale (I suppose that because of labels).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/slidergraph.py
+++ b/Orange/widgets/utils/slidergraph.py
@@ -36,6 +36,7 @@ class SliderGraph(PlotWidget):
         self.getViewBox().setMouseEnabled(False, False)
         self.showGrid(True, True, alpha=0.5)
         self.setRange(xRange=(0.0, 1.0), yRange=(0.0, 1.0))
+        self.hideButtons()
 
         # tuples to store horisontal lines and labels
         self.plot_horlabel = []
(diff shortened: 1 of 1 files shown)


ajdapretnar:
This works well for the plot, but the issue with component variance and cumulative variance falling out of view is still not solved. Is this intentional - a wontfix?

This is what happens with small window:
Screen Shot 2021-02-19 at 11 54 02

janezd:
> This works well for the plot, but the issue with component variance and cumulative variance falling out of view is still not solved. Is this intentional - a wontfix?

I haven't noticed it, sorry. Please try now.

ajdapretnar:
Works beautifully.

PR #5279, opened 2021-02-18
Datasets: Make some columns narrower

### LINKED ISSUE #5091: OWDatasets: Reduce width
The widget is quite big at the moment, and many of the columns could be narrower by default.

**Current**
Screenshot 2020-11-19 at 17 00 56

**Possible solution**
Screenshot 2020-11-19 at 17 01 11

The sizeHint should not be set on the widget itself, it should be set on the table. To calculate it, each column's preferred width should be calculated by checking its values, and it should be clamped to a maximum value (i.e., long titles should elide).

janezd:
##### Issue

May resolve #5091.

##### Description of changes

After losing nerves and time with different methods that didn't work, I realized that we can pretty well know what the column widths should be, so I just fixed them. If we don't like the widths, we can modify them; but I see no reason to be fancier and resize to contents.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -377,10 +377,13 @@ class OWDataSets(OWWidget):
             self.__on_selection
         )
 
+        scw = self.view.setColumnWidth
+        width = self.view.fontMetrics().width
         self.view.resizeColumnToContents(0)
-        self.view.setColumnWidth(
-            1, min(self.view.sizeHintForColumn(1),
-                   self.view.fontMetrics().width("X" * 37)))
+        scw(self.Header.title, width("X" * 37))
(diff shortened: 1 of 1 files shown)


ajdapretnar:
I don't think this solves the problem - the idea was to make the column narrower, not wider. I was expecting something like the max length of Size, Instances, and Variables columns.

janezd:
Does it make them wider for you?! (Before is above, after is below).

Screenshot 2021-02-19 at 11 22 19

ajdapretnar:
I'll double check. Just to make sure I wasn't looking at the wrong version.

ajdapretnar:
You are right, apparently, I mixed up the two (weird, I was sure I was checking the right version). Will merge.

PR #5283, opened 2021-02-18
[FIX] Create Class: multiple patterns for a class value

markotoplak:
##### Issue
Repeated output class values created a DiscreteVariable with repeated values, which introduced problems later on. This PR fixes  Quasars/orange-spectroscopy#521.

##### Description of changes
This PR ensures that all values of the output DiscreteVariable are unique.

Allowing multiple match patterns for a single value seems like a useful feature (we get something similar to OR conditions). So I decided to support it rather than blocking it on input.

Patterns are still being matched in the same order as before, so this PR should not change any behavior that was valid before the fix. Also, the order of output values is kept.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcreateclass.py
+++ b/Orange/widgets/data/owcreateclass.py
@@ -19,20 +19,27 @@ from Orange.widgets.utils.state_summary import format_summary_details
 from Orange.widgets.widget import Msg, Input, Output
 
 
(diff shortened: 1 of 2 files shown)


janezd:
@markotoplak, can you rebase, please? (Conflict in tests.)

markotoplak:
@janezd, I just rebased.

PR #5309, opened 2021-03-05
Edit Domain: Map bool to bool

### LINKED ISSUE #5307: Edit Domain: consider bool as bool
- [ ] **What's wrong?**

Box Plot doesn't show hidden attributes. But if one changes an attribute from hidden=True to hidden=False in Edit Domain, Box Plot still doesn't show the attribute, even though it was "unhidden".

Corpus - Preprocess Text (output only 5 tokens for ease of debugging) - Bag of Words - Edit Domain (change one attribute's label from hidden=True to hidden=False) - Box Plot.
Box Plot won't show the attribute.

- Operating system: macOS
- Orange version: 3.28.dev
- How you installed Orange: git

ajdapretnar:
##### Issue

Fixes #5307.

##### Description of changes
Map "True" and "False" to respective boolean values.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_oweditdomain.py
+++ b/Orange/widgets/data/tests/test_oweditdomain.py
@@ -217,6 +217,30 @@ class TestOWEditDomain(WidgetTest):
         t2 = self.get_output(self.widget.Outputs.data)
         self.assertEqual(t2.domain["a"].attributes["list"], [1, 2, 4])
 
+    def test_annotation_bool(self):
+        """Check if bool labels remain bool"""
+        a = ContinuousVariable("a")
+        a.attributes["hidden"] = True
+        d = Domain([a])
+        t = Table.from_domain(d)
+
+        self.send_signal(self.widget.Inputs.data, t)
+
+        assert isinstance(self.widget, OWEditDomain)
+        # select first variable
+        idx = self.widget.domain_view.model().index(0)
+        self.widget.domain_view.setCurrentIndex(idx)
+
+        # change first attribute value
+        editor = self.widget.findChild(ContinuousVariableEditor)
+        assert isinstance(editor, ContinuousVariableEditor)
+        idx = editor.labels_model.index(0, 1)
+        editor.labels_model.setData(idx, "False", Qt.EditRole)
+
+        self.widget.commit()
+        t2 = self.get_output(self.widget.Outputs.data)
+        self.assertFalse(t2.domain["a"].attributes["hidden"])
+
     def test_duplicate_names(self):
         """
         Tests if widget shows error when duplicate name is entered.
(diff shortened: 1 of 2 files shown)


ajdapretnar:
@janezd @PrimozGodec While the widget seems to work, the test I have written, fails. Can you spot the issue? Is it the test?

Orange/widgets/data/oweditdomain.py
```diff
@@ -2420,11 +2430,8 @@ def abstract(var):
 
 def _parse_attributes(mapping):
     # type: (Iterable[Tuple[str, str]]) -> Dict[str, Any]
-    # Use the same functionality that parses attributes
-    # when reading text files
-    return Orange.data.Flags([
-        "{}={}".format(*item) for item in mapping
-    ]).attributes
+    # Map to dict to preserve booleans
+    return dict((key, value) for key, value in mapping)
```
ales-erjavec, line 2434: ```Orange.data.Flags``` use here was added to match the .*tab* annotations row parsing. E.g. it parses number strings to integers, ..., but not True/False booleans. The tests fail because it no longer does this mapping due to this change.

I think the boolean mapping/parsing should be moved to the `Orange.data.Flags`. That way it will preserve the value type on save/load.

ajdapretnar: Ok, thanks for the hint! Will have a look.

Orange/widgets/data/tests/test_oweditdomain.py
```diff
@@ -217,6 +217,30 @@ def test_list_attributes_remain_lists(self):
...
+        assert isinstance(self.widget, OWEditDomain)
+        # select first variable
+        idx = self.widget.domain_view.model().index(0)
+        self.widget.domain_view.setCurrentIndex(idx)
+
+        # change first attribute value
+        editor = self.widget.findChild(ContinuousVariableEditor)
+        assert isinstance(editor, ContinuousVariableEditor)
+        idx = editor.labels_model.index(0, 1)
+        editor.labels_model.setData(idx, "False", Qt.EditRole)
```
ajdapretnar, line 238: @ales-erjavec Am I right in assuming one could never set the label to `False`, but always to `"False"`?

ales-erjavec: Yes

PrimozGodec:
@ajdapretnar it looks good to me. Should we also convert strings that are numbers to float as @janezd suggested last Friday?

ajdapretnar:
Does this not already happen in io_base.py, L60? We are still talking about Flags, no?

PrimozGodec:
Sorry, I wrote that before I really checked. 🙈

PR #5314, opened 2021-03-08
[ENH] Transpose: Offload work onto separate thread, remove redundant instance

### LINKED ISSUE #5298: Transpose cannot be cancelled
- [ ] **What's wrong?**

For large data sets, Transpose gets slow. It is impossible to cancel the computation (remove the widget).

Use a large data set (adult?) and transpose it.

- Operating system: OSX
- Orange version: 3.27.1
- How you installed Orange: installer

### LINKED ISSUE #5300: Transpose widget bug?
- [x] **What's wrong?**

Not sure if this is a bug or just counter-intuitive for me. 

When selecting the Generic option in the `Transpose` widget the result is what I would expect.

image

However, when I select a variable to get the Feature names from that variable is not taken out of the `Data Table`. I think in this case the selected variable (row in the transposed Data Table) should be removed.

image

I think this works with any file but in this case I used the [file](https://github.com/Quasars/orange-spectroscopy/files/6055430/sample.zip) provided in https://github.com/Quasars/orange-spectroscopy/issues/522

- Operating system: OS X 11.2.1
- Orange version: GitHub master (also tested it in Quasar 1.0.0, same)

VesnaT:
##### Issue

Fixes #5300
Fixes #5298

##### Description of changes
- Offload work onto a separate thread
- Remove redundant instance from transposed table - one that is used for feature names 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -17,8 +17,6 @@ from scipy import sparse as sp
 from scipy.sparse import issparse, csc_matrix
 
 import Orange.data  # import for io.py
-from Orange.misc.collections import frozendict
-from Orange.util import OrangeDeprecationWarning
 from Orange.data import (
     _contingency, _valuecount,
     Domain, Variable, Storage, StringVariable, Unknown, Value, Instance,
@@ -27,9 +25,11 @@ from Orange.data import (
 from Orange.data.util import SharedComputeValue, \
     assure_array_dense, assure_array_sparse, \
     assure_column_dense, assure_column_sparse, get_unique_names_duplicates
+from Orange.misc.collections import frozendict
 from Orange.statistics.util import bincount, countnans, contingency, \
     stats as fast_stats, sparse_has_implicit_zeros, sparse_count_implicit_zeros, \
     sparse_implicit_zero_weights
+from Orange.util import OrangeDeprecationWarning, dummy_callback
 
 __all__ = ["dataset_dirs", "get_sample_datasets_dir", "RowInstance", "Table"]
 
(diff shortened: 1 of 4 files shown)


Orange/data/table.py
```diff
@@ -1727,12 +1732,16 @@ def transpose(cls, table, feature_names_column="",
...
             names = [str(row[feature_names_column]) for row in table]
+            progress_callback(0.1)
             names = get_unique_names_duplicates(names)
+            progress_callback(0.3)
             attributes = [ContinuousVariable(name) for name in names]
         else:
             places = int(np.ceil(np.log10(n_cols))) if n_cols else 1
             attributes = [ContinuousVariable(f"{feature_name} {i:0{places}}")
                           for i in range(1, n_cols + 1)]
+        progress_callback(0.4)
```
janezd, line 1759: I believe the correct value here is 0.418. Please check your calculations again.

Orange/widgets/data/owtranspose.py
```diff
@@ -33,6 +57,7 @@ class Outputs:
     feature_type = ContextSetting(GENERIC)
     feature_name = ContextSetting("")
     feature_names_column = ContextSetting(None)
+    remove_redundant_inst = ContextSetting(False)
```
markotoplak, line 60: Wouldn't a True be a better default? If we migrate so that it becomes False for already saved workflows we are not breaking anything.

Orange/widgets/data/owtranspose.py
```diff
@@ -69,6 +95,11 @@ def __init__(self):
             "feature_names_column", contentsLength=12, searchable=True,
             callback=self._feature_combo_changed, model=self.feature_model)
 
+        self.remove_check = gui.checkBox(
+            gui.indentedBox(box, gui.checkButtonOffsetHint(button)), self,
+            "remove_redundant_inst", "Remove redundant instance",
+            callback=lambda: self.apply())
```
markotoplak, line 101: Why not just `callback=self.apply`? Why a lambda function? (I am not criticizing, I genuinely want to know if there is something I do not understand.)

PR #5315, opened 2021-03-08
DiscreteVariable: Remove 'ordered' argument

VesnaT:
##### Issue

Ordered is deprecated and should be removed from constructor.

##### Description of changes
Remove 'ordered' argument from DiscreteVariable's constructor.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/variable.py
+++ b/Orange/data/variable.py
@@ -34,7 +34,10 @@ def make_variable(cls, compute_value, *args):
     if compute_value is not None:
         return cls(*args, compute_value=compute_value)
     else:
-        # For compatibility with old pickles
+        # For compatibility with old pickles: remove the second arg if it's
+        # bool `compute_value` (args[3]) can't be bool, so this should be safe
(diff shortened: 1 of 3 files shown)


Orange/data/variable.py
```diff
@@ -35,7 +35,7 @@ def make_variable(cls, compute_value, *args):
         return cls(*args, compute_value=compute_value)
     else:
         # For compatibility with old pickles
-        return cls(*args)
+        return cls(*tuple(x for i, x in enumerate(args) if i != 2))
```
janezd, line 38: What about this? @lanzagar, I suppose this should be OK?

```python
# For compatibility with old pickles: remove the second arg if it's bool
# `compute_value` (args[3]) can't be bool, so this should be safe
if len(args) >= 2 and isinstance(args[2], bool):
    args = args[:2] + args[3:]
return cls(*args)
```

VesnaT: if `len(args) == 2` this raises an IndexError.

PR #5316, opened 2021-03-08
README: Fixup instructions for running Orange

irgolic:
##### Issue
https://github.com/biolab/orange3-text/issues/623
I'm assuming the root of this issue is having installed Orange with pip.

##### Description of changes
Add a Windows warning next to the pip installation instructions.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,23 +70,7 @@ Starting up for the first time may take a while.
 
 ### Installing with pip
 
-Warning: on Windows, Orange cannot be installed with pip properly. Use our [standalone installer](https://orange.biolab.si/download) or conda instead.
-
-To install Orange with pip, run the following.
-
-```Shell
-# Install build requirements via your system's package manager
-sudo apt install virtualenv build-essential python3-dev
-
-# Create an environment for Orange and its dependencies
-virtualenv --python=python3 --system-site-packages orange3venv
-
-# Activate the environment
-source orange3venv/bin/activate
-
-# Install Orange
-pip install orange3
-```
+We recommend using our [standalone installer](https://orange.biolab.si/download) or conda, but Orange is also installable with pip. On Windows, you will also need Microsoft Visual Studio Build Tools.
 
 #### Running
 
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ conda install orange3-<addon name>
 
 ### Installing with pip
 
-We recommend using our [standalone installer](https://orange.biolab.si/download) or conda, but Orange is also installable with pip. On Windows, you will also need Microsoft Visual Studio Build Tools.
+We recommend using our [standalone installer](https://orange.biolab.si/download) or conda, but Orange is also installable with pip. You will need a C/C++ compiler (on Windows we suggest using Microsoft Visual Studio Build Tools).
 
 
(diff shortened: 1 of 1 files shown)


ajdapretnar:
This doesn't fix biolab/orange3-text#623. The OP had a Linux installation, which was segfaulting (likely because of WebView). To solve this, one would have to look into Linux installation instruction and perfect them.

markotoplak:
@irgolic, thanks, but in biolab/orange3-text#623 one person also had a problem on Linux, with both pip and conda. I suspect Qt...

@ajdapretnar, why wouldn't Orange be pip-installable on Windows? Did anyone try it recently? I understand sometimes you could have some libraries that are not compiled, but these are getting rarer.

I think we should keep directing users towards our packages, as for the developers, well, we can let them choose. If they have compilers properly set up I do not see pip being a problem...

ajdapretnar:
> @ajdapretnar, why wouldn't Orange be pip-installable on Windows? Did anyone try it recently? I understand sometimes you could have some libraries that are not compiled, but these are getting rarer.

Probably no one has tested it in a while, but numpy+mkl used to be an issue. As for Text (the issue was reported on Text), it cannot be installed on Windows due to gensim (I think).

markotoplak:
Now I remembered: both Orange and Orange3-Text can work on Windows normally if pip installed (provided compilers are there, maybe even if they are not, depends on the libraries) otherwise our tests would fail (our tests use pip for installation).

irgolic:
So what exactly are the prerequisited for a pip installation to fail?

markotoplak:
In theory )and controlled practice), `pip` into a fresh environment should work well and actually does. It is what we test.

In practice, some packages may only have source packages on PyPI, which would require people to have their compiler chain installed. Most Windows users are not programmers, so they don't. Hence they could see errors because compilers are not found. This can also be a problem on Macs and Linux, but installing a compiler is easier (also, users are different).

I do not know what the current state on packages requiring compilation is, exactly... There was a period when some crucial Windows packages lacked binary packages, but nowadays, most of them actually do have them.

janezd:
Decision at discussion: remove instructions for pip, which make an impression this is easy and works; instead say only that you can install Orange with pip, but you'll need compilers etc, and ... have fun.

README.md
```diff
@@ -56,49 +56,32 @@ conda install orange3-<addon name>
...
-and run either `orange-canvas` or `python3 -m Orange.canvas`. 
-
-Add `--help` for a list of program options.
-
-Starting up for the first time may take a while.
 
 ### Installing with pip
 
-To install Orange with pip, run the following.
+We recommend using our [standalone installer](https://orange.biolab.si/download) or conda, but Orange is also installable with pip. On Windows, you will also need Microsoft Visual Studio Build Tools.
```
thocevar, line 62: You will need a C/C++ compiler (on Windows we suggest using Microsoft Visual Studio Build Tools).

janezd:
@JakaKokosar are this problems with CI related to the experiment with rebasing?

JakaKokosar:
Yes, they are. I will try to finish this as soon as possible :)

PR #5308, opened 2021-03-04
[ENH] Automated and better summaries

janezd:
##### Issue

Ref: #5108

Requires: https://github.com/biolab/orange-widget-base/pull/136.

##### Description of changes

See https://github.com/biolab/orange-widget-base/pull/136.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #5324, opened 2021-03-10
Fix calls to set_domain

### LINKED ISSUE #5323: Merge Data: crash on empty data with domain
- [ ] **What's wrong?**

Merge Data crashes on empty data with domain (I suppose that is the case).

File (iris) - Pivot Table - Merge Data. Now, uncheck all aggregation methods in Pivot Table.
```

  File "/Users/ajda/orange/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 1010, in __process_next
  File "/Users/ajda/orange/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 1048, in __process_next_helper
  File "/Users/ajda/orange/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 679, in process_node
  File "/Users/ajda/orange/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 792, in send_to_node
  File "/Users/ajda/orange/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 826, in process_signals_for_widget
    handler(*args)
  File "/Users/ajda/orange/orange3/Orange/widgets/utils/sql.py", line 31, in new_f
    return f(widget, data, *args, **kwargs)
  File "/Users/ajda/orange/orange3/Orange/widgets/data/owmergedata.py", line 355, in set_data
    self.model.set_domain(data and data.domain)
  File "/Users/ajda/orange/orange3/Orange/widgets/utils/itemmodels.py", line 501, in set_domain
    (domain.attributes, domain.class_vars, domain.metas))
AttributeError: 'Table' object has no attribute 'class_vars'
```

- Operating system: OSX
- Orange version: 3.28.0
- How you installed Orange: git

janezd:
##### Issue

Fixes #5323, and the same in three other widgets.

This happenned because empty table is also considered false, hence `data and data.domain` resulted in passing a `Table` (instead of `None` or an instance of `Domain`) when the table was empty.

I'd prefer not to write tests for this, though they'd be trivial. I don't like tests that just replicate a bug. Of course one could say "but you should test that a widget does not crash on empty table", but then we should add such a test to all widgets, not just the four that happenned to have this problem.

##### Includes
- [X] Code changes
- [ ] Tests

### DIFF
--- a/Orange/widgets/data/owcorrelations.py
+++ b/Orange/widgets/data/owcorrelations.py
@@ -376,7 +376,8 @@ class OWCorrelations(OWWidget):
         self.vizrank.button.setEnabled(self.cont_data is not None)
 
     def set_feature_model(self):
-        self.feature_model.set_domain(self.cont_data and self.cont_data.domain)
+        self.feature_model.set_domain(
+            self.cont_data.domain if self.cont_data else None)
         data = self.data
         if self.cont_data and data.domain.has_continuous_class:
             self.feature = self.cont_data.domain[data.domain.class_var.name]
--- a/Orange/widgets/data/owcreateclass.py
+++ b/Orange/widgets/data/owcreateclass.py
@@ -301,7 +301,7 @@ class OWCreateClass(widget.OWWidget):
         details = format_summary_details(data) if data else ""
         self.info.set_input_summary(summary, details)
         model = self.controls.attribute.model()
-        model.set_domain(data.domain if data else None)
+        model.set_domain(data.domain if data is not None else None)
         self.Warning.no_nonnumeric_vars(shown=data is not None and not model)
         if not model:
             self.attribute = None
--- a/Orange/widgets/data/owcreateclass.py
+++ b/Orange/widgets/data/owcreateclass.py
@@ -301,7 +301,7 @@ class OWCreateClass(widget.OWWidget):
         details = format_summary_details(data) if data else ""
         self.info.set_input_summary(summary, details)
         model = self.controls.attribute.model()
-        model.set_domain(data and data.domain)
+        model.set_domain(data.domain if data else None)
         self.Warning.no_nonnumeric_vars(shown=data is not None and not model)
         if not model:
             self.attribute = None
--- a/Orange/widgets/data/owmergedata.py
+++ b/Orange/widgets/data/owmergedata.py
@@ -352,13 +352,13 @@ class OWMergeData(widget.OWWidget):
     @check_sql_input
     def set_data(self, data):
         self.data = data
-        self.model.set_domain(data and data.domain)
+        self.model.set_domain(data.domain if data else None)
 
     @Inputs.extra_data
     @check_sql_input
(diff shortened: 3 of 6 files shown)


ajdapretnar:
I have a couple of issues, not necessarily with this fix, but let's discuss what needs to be solved where.
1. Merge Data. The first issue is now solved, so when I connect from Pivot Table to Merge Data. But if I connect the second table to Merge Data, a new, but related issue appears.
T```
raceback (most recent call last):
  File "/Users/ajda/orange/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 1010, in __process_next
  File "/Users/ajda/orange/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 1048, in __process_next_helper
  File "/Users/ajda/orange/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 679, in process_node
  File "/Users/ajda/opt/miniconda3/envs/o3/lib/python3.8/site-packages/orangewidget/workflow/widgetsscheme.py", line 792, in send_to_node
  File "/Users/ajda/opt/miniconda3/envs/o3/lib/python3.8/site-packages/orangewidget/workflow/widgetsscheme.py", line 833, in process_signals_for_widget
    widget.handleNewSignals()
  File "/Users/ajda/orange/orange3/Orange/widgets/data/owmergedata.py", line 369, in handleNewSignals
    self.openContext(self.data and self.data.domain,
  File "/Users/ajda/opt/miniconda3/envs/o3/lib/python3.8/site-packages/orangewidget/widget.py", line 1113, in openContext
  File "/Users/ajda/orange/orange3/Orange/widgets/data/owmergedata.py", line 191, in open_context
    self._encode_domain(domain1),
  File "/Users/ajda/orange/orange3/Orange/widgets/data/owmergedata.py", line 210, in _encode_domain
    all_vars = chain(domain.variables, domain.metas)
AttributeError: 'Table' object has no attribute 'variables'
```
2. Correlations. Now it reports 0 instances on the input and a warning "At least two instances are needed". Transpose reports None (-). We should decide on one option, IMO.
3. Create Class. Report None (-) and a warning "Data contains only numeric variables." Shouldn't widgets report "No data on input" or something?

janezd:
1. Fixed. (The general `DomainContextHandler` accepts `Domain` or `Table`; `MergeDataContextHandler` didn't, but now does.)
2. Inconsistencies in summaries will be fixed by auto-summaries (#5308).
3. I fixed the warning.

As for whether the widget should report that the data table is empty: I wouldn't, because in this case every widget would have such a warning (or even an error) with the corresponding additional code ... I think a "0" in the summary should suffice, unless some particular widget needs to show a warning -- for instance a widget that need at least two instance and wants to inform the user that a single instance is not enough.

ajdapretnar:
Ok, let's go with a 0. It is just slightly confusing from a user's perspective, since we denote "empty" signals with a dashed line. Here, the line is full. But I agree, people should get used to using the status bar. ;)

PR #5325, opened 2021-03-10
[ENH] Feature Statistics: Add median

### LINKED ISSUE #5317: Add median to Feature Statistics
I need some statistics about the features of my data, including the median from each feature.
Feature Statistics is very close from what I need but doesn't provide the median.

Adding a column with the median in Feature Statistics would solve my problem. 

I don't know Orange well enough to provide any alternative solutions.

janezd:
##### Issue

Resolves #5317.

##### Description of changes

- Add a column with median
- Show the most frequent value of discrete values in column with medians rather than with means,
- Rename column "Center" to "Mean"

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfeaturestatistics.py
+++ b/Orange/widgets/data/owfeaturestatistics.py
@@ -110,14 +110,16 @@ class FeatureStatisticsTableModel(AbstractSortTableModel):
     HIDDEN_VAR_TYPES = (StringVariable,)
 
     class Columns(IntEnum):
-        ICON, NAME, DISTRIBUTION, CENTER, DISPERSION, MIN, MAX, MISSING = range(8)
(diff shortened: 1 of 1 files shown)


ajdapretnar:
I believe the reason the column was called Center instead of Mean was because it shows "mode" for discrete data.

janezd:
That's it - tests doesn't cover sorting, but that's also the case for most other scores, and there's also not much to test, so I won't add them.

irgolic:
This is cool. I think mean/median is much more transparent and easy to understand than center.


Sidenote:

Dispersion is confusing too, similarly as center was. I had to open the documentation to understand what the values meant. This information could easily be in a tooltip when hovering over the column/column header. Having docs is great, but having out-of-date docs is not. However, keeping them up-to-date by hand is a three-man full-time-job. Let's talk about this over tea.

PR #5326, opened 2021-03-10
Feature Constructor: Fix a problem with some unicode characters

### LINKED ISSUE #5305: Feature Constructor: Handle features with special symbols
- [ ] **What's wrong?**
 The Feature Constructor widget can not handle features with `µ` letter in their names.

Run [test.zip](https://github.com/biolab/orange3/files/6074250/test.zip)

- Orange version: master
- How you installed Orange: from source

janezd:
##### Issue

Fixes #5305.

##### Description of changes

Normalize encoding of variable names as (seemingly) encoded by Python parser.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -14,6 +14,7 @@ import random
 import logging
 import ast
 import types
+import unicodedata
 
 from traceback import format_exception_only
 from collections import namedtuple, OrderedDict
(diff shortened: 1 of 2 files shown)


janezd:
It's also official: "*All identifiers are converted into the normal form NFKC while parsing; comparison of identifiers is based on NFKC*" (https://docs.python.org/3/reference/lexical_analysis.html#identifiers), so this fix should also work in future (backwards compatible) versions of Python.

PR #5350, opened 2021-03-19
Continuize, Discretize: Add keywords for easier searching

### LINKED ISSUE #5249: Rename: Continuize and Discretize

Continuize and Discretize widget names might be confusing for users.
We have also decided to use numeric and categorical instead of continuous and discrete.

Continuize - To Numeric
Discretize - To Categorical

Open to other suggestions.

janezd:
##### Issue

Resolves #5249.

##### Description of changes

Add keywords
- to Continuize: *encode*, *dummy*, *numeric*, *one-hot*, *binary*
- to Discretize: *bin*, *categorical*, *nominal*, *ordinal*

If anybody has anything to add, say so.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/data/owcontinuize.py
+++ b/Orange/widgets/data/owcontinuize.py
@@ -23,7 +23,8 @@ class OWContinuize(widget.OWWidget):
                    "optionally, normalize numeric values.")
     icon = "icons/Continuize.svg"
     category = "Data"
-    keywords = []
+    keywords = ["encode", "dummy", "numeric", "one-hot", "binary",
+                "treatment", "contrast"]
 
     class Inputs:
         data = Input("Data", Orange.data.Table)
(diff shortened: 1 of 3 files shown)


irgolic:
Great PR, thanks. I might need to change how we parse keywords so that 'one-hot' is hit without the dash, and with a space instead of the dash, as well.

Regen/update widget.json, and I'll merge.

Orange/widgets/data/owcontinuize.py
```diff
@@ -23,7 +23,7 @@ class OWContinuize(widget.OWWidget):
                    "optionally, normalize numeric values.")
     icon = "icons/Continuize.svg"
     category = "Data"
-    keywords = []
+    keywords = ["encode", "dummy", "numeric", "one-hot", "binary"]
```
ales-erjavec, line 26: \+ "treatment", "contrast" ?

PR #5360, opened 2021-03-25
Test and score docs: fix comparison interpretation

markotoplak:
##### Issue
Docs assumed that for all measures the higher value means a better model.

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation


markotoplak:
@ajdapretnar, I hope you double-checked that, I was not 100% certain. :D

PR #5365, opened 2021-03-26
Melt: Add widget

janezd:
Moved from prototypes (https://github.com/biolab/orange3-prototypes/pull/242).

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation


janezd:
@ajdapretnar, please put finishing touches to documentation.

ajdapretnar:
One thing, do we want to keep the shopping cart icon? I think it is fun and I see how it relates to the widget, but I think it will leave most of our users confused. :) In this respect, I'd rather wait with the documentation until we decide on the icon.

irgolic:
I think the icon is cool :)

borondics:
I agree with @ajdapretnar. The function relates to the icon but the name doesn't.

markotoplak:
We have too many widgets describing abstract functionality that icon images could be sufficiently descriptive. The next best thing is to make icons recognizable (meaning that users learn to recognize them quickly). Being special and funny likely helps.

borondics:
How about a melting piece of chocolate or ice cream as an icon? That would
be funny enough.

Otherwise the name could be Shopping... :D

irgolic:
Melt is the established name for the operation.
https://pandas.pydata.org/docs/reference/api/pandas.melt.html
https://www.rdocumentation.org/packages/reshape2/versions/1.4.4/topics/melt

But I think, when learning what melt means, the analogy of a shopping cart might help you understand it.

janezd:
I like the icon, too. If we try to draw an icon that will "animate" what happens with the table, it will be just one more among  many poorly recognizable icons.

@irgolic: "Melt" may be established name, but none of us knew about it. :)

@borondics, "Shopping List" was indeed the former name of this widget, thence the icon. :)

irgolic:
> @irgolic: "Melt" may be established name, but none of us knew about it. :)

I'm afraid this may speak to our lack of contact with the rest of the DS community (R, pandas). Empirically, see this random pandas cheat-sheet I found, where melt is shown next to concat and pivot. https://pandas.pydata.org/Pandas_Cheat_Sheet.pdf

lanzagar:
Everything looks OK to me.
I would suggest a change from warning to info for the "No columns with unique values" message, since we support using row numbers and it does not mean something is wrong if you have data like that.

A possible enhancement for the future would be support for "multi-indexes" - currently only one feature can be selected as the Unique Row Identifier.

ajdapretnar:
@janezd I think I successfully debugged the docs.
Once the tests pass, this is ready for a merge.

janezd:
@ajdapretnar, I just did something stupid: I didn't know why checks fail, I assumed this could be helped by rebasing to the latest master and pushing (-f). This removed your commits. I'm to trigger-happy with -f ...

You have the latest version that includes my commits (and yours, of course). Could you please `rebase -i master` and `push -f`?

PR #5401, opened 2021-04-16
[DOC] Clean up development documentation

### LINKED ISSUE #5059: Orange Visual Programming documentation looks unfinished
https://orange3.readthedocs.io/en/latest/?badge=latest
Is this the right link for our documentation?

If so, the Getting Started section reads 'Here we need to copy the getting started guide.'
Is there anything else beside 'Loading you data', which should be there?
Perhaps it'd be nice to somehow link to the other relevant documentation too (development, canvas-core, widget-base)

ajdapretnar:
##### Issue

Partially fixes #5059. 

##### Description of changes
Join two Getting started sections.
Remove concurrent (as it doesn't yield any results).

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation


ajdapretnar:
Note: building documentation fails, because likely some functions were renamed or removed, while they are still called in the docs. Someone should some day check this and clean it up.

markotoplak:
@ales-erjavec, it seems to me that these docs are now part of orange-widgets-base. Could you please confirm that before I do something stupid?

If so, this PR should be closed and the development docs deleted from this repo.

ales-erjavec:
I don't know.

The entire 'tutorial' should probably be here and removed from orange-widget-base, keeping only the module index/api reference.

markotoplak:
@ales-erjavec, I now see that it uses Orange's objects, so the natural place for the current tutorial does indeed seem to be here. So we'd probably need to "synchronize" and remove the duplicated parts.

PR #5413, opened 2021-04-29
[FIX] Speed-up slow table_to_frame

PrimozGodec:
##### Issue

`table_to_frame` is slow on large dataset. On dataset with shape (1M, 12) it took ~20s to transform dataframe to table. The reason for slow error is list comprehension used to find out if there is any nan in numeric column.

##### Description of changes
List comprehension is now replaced with NumPy functions.
Transformation on the previously mention dataset now takes ~0.6s

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/pandas_compat.py
+++ b/Orange/data/pandas_compat.py
@@ -368,8 +368,7 @@ def table_to_frame(tab, include_metas=False):
         elif col.is_continuous:
             dt = float
             # np.nan are not compatible with int column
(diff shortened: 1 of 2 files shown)


borondics:
We started using bottleneck in Orange Spectroscopy. Faster than numpy. Maybe it would also make sense to use that here.
Check out https://github.com/Quasars/orange-spectroscopy/pull/531

PR #5423, opened 2021-05-06
DBSCAN: Retain data ids

### LINKED ISSUE #5422: DBSCAN changes instance ids
The output of DBSCAN looses traceability of data instances. The following workflow should explain the problem:

image

t-SNE should mark all data instances, as the data subset is the same as the data, but instead reports that "No subset data instances appear in input data". 

image

Similar is of course true with any selection of the data that comes from DBSCAN. My original intention was to send the DBSCAN data to the boxplot and then show the selection from the box plot, sat, the clusters with interesting class-distribution, as a data subset to the t-SNE. But then I get, of course, the same warning in t-SNE, reporting the mismatch of the data subset and data provided to this widget.

VesnaT:
##### Issue

Fixes #5422

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owdbscan.py
+++ b/Orange/widgets/unsupervised/owdbscan.py
@@ -10,7 +10,7 @@ from Orange.preprocess import Normalize, Continuize, SklImpute
 from Orange.widgets import widget, gui
 from Orange.widgets.utils.slidergraph import SliderGraph
 from Orange.widgets.settings import Setting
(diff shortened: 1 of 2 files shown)


markotoplak:
Thanks, @VesnaT. This mistake seems to happen often. Perhaps we can showcase common bugs like these at our meetings.

PR #5427, opened 2021-05-06
[FIX] Outlier detection: keep instance ids, make thread safe

markotoplak:
##### Issue

1. Outlier detection did not keep instance ids, so subsets did work with the `Data` output of OWOutliers.
2. Also, it was not thread-safe: multiple calls to  `_OutlierModel.__call__` could result in undefined behavior, because some caching was done at the object level.

##### Description of changes
To fix (1), I just used domain transformations instead of copying. To fix (2), I replaced object-level caching with `SharedComputeValue`.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/classification/outlier_detection.py
+++ b/Orange/classification/outlier_detection.py
@@ -9,43 +9,51 @@ from sklearn.neighbors import LocalOutlierFactor
 from sklearn.svm import OneClassSVM
 
 from Orange.base import SklLearner, SklModel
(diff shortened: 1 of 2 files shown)


VesnaT (approved):
A minor comment. Other than that, I'm happy to merge.

markotoplak:
@VesnaT, thanks!

PR #5428, opened 2021-05-07
DBSCAN: Optional normalization

### LINKED ISSUE #5424: Unknown Preprocessing Steps
- [x] **What's wrong?**

Some widgets, like [DBSCAN](https://github.com/VesnaT/orange3/blob/1f43169ffd55745300c8429f81b95e2494e63072/Orange/widgets/unsupervised/owdbscan.py#L23), do preprocessing without making the user aware of what is happening. I think it should be somehow be made clear as for example spectroscopy we typically wouldn't want to normalize features for example.

All widgets should make the users aware of their operations at least in the documentation. But ideally this information should be shown inside the widget as well. If the preprocessing steps are absolutely necessary to make something work properly, I could imagine adding a "Preprocessing" section somewhere with checkboxes, where the items are grey so they can't be changed.

Maybe it would be good to add the preprocessing steps to the little popup help as well between Inputs and Outputs: 
image

- Operating system: OSX
- Orange version: Git master

VesnaT:
##### Issue

Fixes #5424

##### Description of changes
Add a check box to normalize/not normalize the input dataset. The check box is checked by default.

image

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owdbscan.py
+++ b/Orange/widgets/unsupervised/owdbscan.py
@@ -78,6 +78,7 @@ class OWDBSCAN(widget.OWWidget):
     min_samples = Setting(4)
     eps = Setting(0.5)
     metric_idx = Setting(0)
+    normalize = Setting(True)
     auto_commit = Setting(True)
     k_distances = None
     cut_point = None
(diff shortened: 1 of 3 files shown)


markotoplak:
@VesnaT, thanks, looks good to me!

borondics:
This is a nice addition, but what about the other preprocessors in DBSCAN? The user still doesn't know about them, even if they are necessary...

markotoplak:
The other preprocessors are necessary if you do not want the widget to work at all with missing values or some discrete columns. If your data does not have either they will not do anything anyway.

The UI's are different because one allows distance measurement method setting and the other does not.

janezd:
But preprocessors should be listed in documentation, I guess. For this widget and for others.

irgolic:
Could we automate/standardize preprocessors, to make them a bit more transparent?

PR #5440, opened 2021-05-14
[ENH] Scatter plot: Bring discrete attributes functionality back

### LINKED ISSUE #4385: Visualizing categorical data in scatter plot with Orange (version > 3.18)
**Describe the bug**
It is not possible to visualize categorical data in Orange, version >3.18. Orange official doc states otherwise, showcasing support for both, continuous as well as discrete-valued attributes. 

**To Reproduce**

**Orange version:** 
3.20 - 3.24

**Operating system:** 
Win 10 64-bit, Win 8.1 64-bit,

VesnaT:
##### Issue

Fixes #4385

##### Description of changes
- revert commit f19342036a92e6fb45281aeb856ca99847697aa3
- revert commit 87b67a895f2a96743228fd7d67db69145d10b7da
- apply commit 1418d09ec8630693e720b12c8d150b5c4a294ef3

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -2,6 +2,7 @@ from itertools import chain
 from xml.sax.saxutils import escape
 
 import numpy as np
+from AnyQt.QtWidgets import QGroupBox, QPushButton
 from scipy.stats import linregress
 from sklearn.neighbors import NearestNeighbors
 from sklearn.metrics import r2_score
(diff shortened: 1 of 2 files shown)


markotoplak:
It might be a good idea to bump the widget settings version.

Also, jittering continuous attributes is not a parameter; before any amount of jittering applied to continuous features. So to keep the same behavior for saved workflows we should set it to `True` for old workflows.

lanzagar:
the second bullet in Description of changes says:
> revert commit 3773093

Did you copy the wrong commit hash because that one seems unrelated?

Otherwise I tried it and categorical vars work again :+1:

VesnaT:
@lanzagar yes, the commit was wrong. I fixed that, thanks for noticing.
@markotoplak you are right, thanks for pointing that out. I'll implement the settings migration.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -537,6 +586,9 @@ def migrate_settings(cls, settings, version):
                 settings["auto_commit"] = settings["auto_send_selection"]
             if "selection_group" in settings:
                 settings["selection"] = settings["selection_group"]
+        if version < 5:
```
markotoplak, line 589: Thanks, @VesnaT, but I think this is too strong: `jitter_continous` was a valid parameter somewhere before 5. Shouldn't we only set it to True if it is not in the dictionary?

Issue #5453, opened 2021-05-25
Widgets' controls call `unconditional_commit`


janezd:
For example, `OWAggregate` has

```python
 var_list = gui.listView(
            box, self, "variables", model=self.variable_model,
            callback=self.commit)
```

It should have `callback=lambda: self.comit()` instead. God knows how many widgets have the same problem. Even veteran developers make this mistake; beginners don't even know why the heck should they use lambda here.

As @irgolic once said, it should be the other way around -- the widget should have an `unconditional_commit` and `auto_commit` should add the conditional one. At the same time, `self.commit` should be available at the time of `__init__` and refer to the conditional commit, which is added later. Furthermore, this must be resolved so that it doesn't break backward compatibility and it doesn't introduce to much magic. No meta classes, for instance.

markotoplak:
Thanks, Janez. I was complaining about that multiple times and no one wanted to back me up.

My original complaint was that adding a button renames a function and it was confusing where it came from.

I only work on Orange since 2007 - I have no idea if that counts as a veteran, but this keeps confusing me. A few weeks ago I asked @VesnaT why did she use that unneeded lambda in a callback while I was reviewing her PR. It turned out it was needed.

 I suggest deprecating the current button and just make a new one with the appropriate semantics.

irgolic:
One option is, we could remove autocommits and enhance the UX of freezing the workflow.

Another option, keep the UX as is, and (as Janez suggests) make `self.commit` refer to the conditional commit upon `__init__` time. We could add this to `WidgetMetaclass`, wherein the original commit is delegated to a hidden method. It wouldn't really add any magic that isn't already present. 
@markotoplak and I have talked about this before, I think `self.unconditional_commit` really shouldn't be public. Right now, all widgets manually implement (and test) calling it upon receiving a new data source. That should be made consistent and implemented in OWBaseWidget. Also note, it can be called `unconditional_apply` too if the commit method is called `apply`. We should probably consolidate that too.

Really, both workflow freezing and autocommits are there because there's no way to interrupt processing. We separate orange-canvas-core into a GUI process (purely for canvas GUI - toolbox, connecting widgets) and a worker process (wherein widgets run - widget interfaces, SignalManager).

P.S.: widgets like File really really need to follow the convention and get an autocommit button. I've spent the past few months closely observing users, it genuinely confuses them.

janezd:
Here's an idea that requires minimal changes.

Widget has to decorate the commit function, like this:

```python
[...]

PR #5457, opened 2021-05-26
Aggregate Columns - wrap commits in lambda for callbacks

PrimozGodec:
##### Issue

Fixes temporarily https://github.com/biolab/orange3/issues/5453

##### Description of changes
We wrap self.commit in lambda and so it is not getting called when auto-commit is not selected

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owaggregatecolumns.py
+++ b/Orange/widgets/data/owaggregatecolumns.py
@@ -53,21 +53,22 @@ class OWAggregateColumns(widget.OWWidget):
             order=DomainModel.MIXED, valid_types=(ContinuousVariable, ))
         var_list = gui.listView(
             box, self, "variables", model=self.variable_model,
-            callback=self.commit)
+            callback=lambda: self.commit()  # pylint: disable=W0108
+        )
         var_list.setSelectionMode(var_list.ExtendedSelection)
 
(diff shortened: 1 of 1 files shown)


irgolic:
You need to call the commit within the lambda
`callback=self.commit` -> `callback=lambda: self.commit()`

PrimozGodec:
Thank you for that. It is fixed now. I don't know how it went through when I was testing 🙈

Orange/widgets/data/owaggregatecolumns.py
```diff
@@ -53,21 +53,22 @@ def __init__(self):
...
-            callback=self.commit
+            callback=lambda: self.commit()  # pylint: disable=W0108
         )
         combo.setSizePolicy(QSizePolicy.MinimumExpanding, QSizePolicy.Fixed)
 
         gui.lineEdit(
             box, self, "var_name",
             label="Variable name: ", orientation=Qt.Horizontal,
-            callback=self.commit
+            callback=lambda: self.commi()  # pylint: disable=W0108
```
markotoplak, line 71: Primož. :)

PrimozGodec:
It should be ok now. :)

Issue #5460, opened 2021-05-28
Issue with pagination on widget bar 3.29 on Conda with Ubuntu


EsotericRebel:
- [ ] **What's wrong?**

- [ ] **How can we reproduce the problem?** 

﻿2021-05-28 10:27:17,450:INFO:Orange.canvas.__main__: Running widget discovery process.
2021-05-28 10:27:17,521:INFO:orangecanvas.registry.discovery: Could not import 'Orange.widgets.data.owaggregatecolumns'.
Traceback (most recent call last):
  File "/home/geoffrey/miniconda3/envs/orange/lib/python3.8/site-packages/orangecanvas/registry/discovery.py", line 260, in iter_widget_descriptions
    module = asmodule(name)
  File "/home/geoffrey/miniconda3/envs/orange/lib/python3.8/site-packages/orangecanvas/registry/discovery.py", line 503, in asmodule
    return __import__(module, fromlist=[""])
  File "/home/geoffrey/miniconda3/envs/orange/lib/python3.8/site-packages/Orange/widgets/data/owaggregatecolumns.py", line 9, in <module>
    from Orange.widgets import gui, widget
  File "/home/geoffrey/miniconda3/envs/orange/lib/python3.8/site-packages/Orange/widgets/gui.py", line 17, in <module>
    from orangewidget.utils.itemdelegates import (
ModuleNotFoundError: No module named 'orangewidget.utils.itemdelegates'
2021-05-28 10:27:17,711:INFO:orangecanvas.registry.discovery: Could not import 'Orange.widgets.data.owcolor'.
Traceback (most recent call last):
  File "/home/geoffrey/miniconda3/envs/orange/lib/python3.8/site-packages/orangecanvas/registry/discovery.py", line 260, in iter_widget_descriptions
    module = asmodule(name)
  File "/home/geoffrey/miniconda3/envs/orange/lib/python3.8/site-packages/orangecanvas/registry/discovery.py", line 503, in asmodule
    return __import__(module, fromlist=[""])
  File "/home/geoffrey/miniconda3/envs/orange/lib/python3.8/site-packages/Orange/widgets/data/owcolor.py", line 18, in <module>
    from Orange.widgets import widget, settings, gui
  File "/home/geoffrey/miniconda3/envs/orange/lib/python3.8/site-packages/Orange/widgets/widget.py", line 11, in <module>
    from Orange.widgets.utils.progressbar import ProgressBarMixin
[...]

PR #5461, opened 2021-05-28
Set minimal version of orange-widget-base to 4.13

PrimozGodec:
##### Issue

Fixes https://github.com/biolab/orange3/issues/5460

##### Description of changes
Set minimal version of orange-widget-base to 4.13 since it is the first version that implements `itemdelegates`

##### Includes
- [ ] Code changes
- [ ] Tests
- [ ] Documentation


PrimozGodec:
We should also add CI for the oldest supported version of orange-widget-base and canvas-core or just upgrade both versions when the new release comes.

PR #5466, opened 2021-06-03
table_from_frame - speedups

PrimozGodec:
##### Issue

`table_from_frame` can be slow when transforming big dataframes

##### Description of changes
In this PR I identify places where transformation is slow and proposed changes.

I tested the speedup on the data frame with 6M rows and ~10 columns and succeeded in speedup the whole transformation from 16s to 2.5s.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/pandas_compat.py
+++ b/Orange/data/pandas_compat.py
@@ -2,6 +2,7 @@
 from unittest.mock import patch
 
 import numpy as np
+from pandas.core.dtypes.common import is_string_dtype
 from scipy import sparse as sp
 from scipy.sparse import csr_matrix
 import pandas as pd
@@ -170,9 +171,20 @@ def vars_from_df(df, role=None, force_nominal=False):
         _role = role
 
     # If df index is not a simple RangeIndex (or similar), put it into data
-    if not any(str(i).startswith('_o') for i in df.index) \
-            and not (df.index.is_integer() and (df.index.is_monotonic_increasing
-                                                or df.index.is_monotonic_decreasing)):
+    if (
+        # not range-like index - test first to skip slow startswith(_o) check
+        not (
+            df.index.is_integer()
+            and (df.index.is_monotonic_increasing or df.index.is_monotonic_decreasing)
+        )
+        # check that it does not contain Orange index
+        and (
+            # startswith is slow (for long drs) - firs check if col has strings
+            isinstance(df.index, pd.MultiIndex)
+            or not is_string_dtype(df.index)
+            or not any(str(i).startswith("_o") for i in df.index)
+        )
+    ):
         df = df.reset_index()
 
     Xcols, Ycols, Mcols = [], [], []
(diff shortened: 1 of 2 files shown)


Orange/data/pandas_compat.py
```diff
@@ -203,12 +216,16 @@ def vars_from_df(df, role=None, force_nominal=False):
...
+            def to_cat(s, _):
+                x = s.astype("category").cat.codes
+                # it is same than x.replace(-1, np.nan), but much faster
+                x = x.where(x != -1, np.nan)
+                return np.asarray(x)
+
+            Xexpr.append(to_cat)
         elif _is_datetime(s):
             var = TimeVariable(str(column))
-            s = pd.to_datetime(s, infer_datetime_format=True)
```
PrimozGodec, line 211: I think this line dos do not have an effect? @irgolic is it correct or I missed something?

irgolic: I think this was a remnant of @kernc's table to dataframe implementation; in lieu of backwards compatibility, I tried to keep the behavior as similar as possible

Orange/data/pandas_compat.py
```diff
@@ -170,16 +171,28 @@ def vars_from_df(df, role=None, force_nominal=False):
...
+        not (
+            df.index.is_integer()
+            and (df.index.is_monotonic_increasing or df.index.is_monotonic_decreasing)
+        )
+        # check that it does not contain Orange index
+        and (
+            # startswith is slow (for long drs) - firs check if col has strings
+            isinstance(df.index, pd.MultiIndex)
+            or not is_string_dtype(df.index)
+            or not any(str(i).startswith("_o") for i in df.index)
```
VesnaT, line 185: Is it necessary to go through all the indexes? 
Some cases are not handled anyway, e.g:
`pd.DataFrame({"a": [1, 2], "b": [3, 4]}, index=["a", "o"])` vs. `pd.DataFrame({"a": [1, 2], "b": [3, 4]}, index=["_a", "_o"])`
Could this be speeded up even more by checking only, e.g., the first index (`df.index[0]`)?

Orange/data/pandas_compat.py
```diff
@@ -281,13 +298,19 @@ def table_from_frame(df, *, force_nominal=False):
...
         attributes = df.orange_attributes
-        ids = [int(i[2:]) if str(i).startswith('_o') else Table.new_id()
-               for i in df.index]
+        if isinstance(df.index, pd.MultiIndex) or not is_string_dtype(df.index):
+            # we can skip checking for Orange indices when MultiIndex an when
+            # not string dtype and so speedup the conversion
+            ids = None
+        else:
+            ids = [
+                int(i[2:]) if str(i).startswith("_o") else Table.new_id()
```
VesnaT, line 311: Is it anyhow assured that `i[2:]` can always cast to `int`?

Orange/data/pandas_compat.py
PrimozGodec: This one is actually fine. Since the argument to any is generator it will not be asked for the next value after first True appears.

Orange/data/pandas_compat.py
PrimozGodec: You are right. There is no guarantee for that in case the index with `_o` was not produced with the transformation of Table to frame. It should be fixed now.

PR #5476, opened 2021-06-09
WidgetLearnerTestMixin: improve test

VesnaT:
##### Issue

A test `test_output_model_picklable` passes in case the widget does not output the model, even though the model could not have been pickled.

##### Description of changes
Add a `model is not None` assertion.

##### Includes
- [ ] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
@VesnaT, thanks for the fix.

PR #5495, opened 2021-06-19
Auto commit -> Delayed commit

janezd:
##### Issue

Illustrates the use of decorator `delayed`, introduced by https://github.com/biolab/orange-widget-base/pull/157. 

Requires https://github.com/biolab/orange-widget-base/pull/157.

See #5453 for details.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


pavlin-policar:
This is great. This is much better. I might rename `now` to `force`. `force` is a heavier word, and in almost all cases, we want to defer the commit.

VesnaT (approved):

lanzagar (commented):
I also like the new approach better than the current auto_commit/lambda practice!

Is the plan to expand this PR with updates to all auto_commits across Orange, and accept it soon after the update to widgetbase (to avoid hundreds of warnings)?

janezd:
> I might rename `now` to `force`. `force` is a heavier word, and in almost all cases, we want to defer the commit.

I'd keep `now`. First, it is more to the point and in contrast with `deferred`; forced sounds as if the caller requests a commit that otherwise wouldn't happen. Second, calling `now` is not an exception; it is less common, but it is expected to call it in response to receiving a new input.

irgolic:
> Second, calling now is not an exception; it is less common, but it is expected to call it in response to receiving a new input.

Non sequitur, you mentioned these may be called in threads some time? Could you point out these weird cases? My impression is still that this could (at some point outwith the scope of this PR) be standardized across all widgets in some way that would eliminate calling `.now` individually in widgets.

janezd:
> > Second, calling now is not an exception; it is less common, but it is expected to call it in response to receiving a new input.
>  Non sequitur, you mentioned these may be called in threads some time? Could you point out these weird cases? My impression is still that this could (at some point outwith the scope of this PR) be standardized across all widgets in some way that would eliminate calling `.now` individually in widgets.

We talked about threads with respect to table locking. Have we also discussed them in relation with this?

Deferred methods are meant to defer user actions, for instance to let them change several parameters of a method before it's run. We (should) never defer the widget's reaction upon receiving a new signal, hence `now`. Both calls will thus remain valid for as long as there exist computations that can't be cancelled. Including in add-ons. :)

janezd:
> I also like the new approach better than the current auto_commit/lambda practice!
> 
> Is the plan to expand this PR with updates to all auto_commits across Orange, and accept it soon after the update to widgetbase (to avoid hundreds of warnings)?

Yes, I guess. I wanted to wait for the design to be confirmed. Assuming we have (that is: if we keep `now`), I can start converting other widgets.

irgolic:
> > > Second, calling now is not an exception; it is less common, but it is expected to call it in response to receiving a new input.
> > > Non sequitur, you mentioned these may be called in threads some time? Could you point out these weird cases? My impression is still that this could (at some point outwith the scope of this PR) be standardized across all widgets in some way that would eliminate calling `.now` individually in widgets.
> 
> We talked about threads with respect to table locking. Have we also discussed them in relation with this?
> 
> Deferred methods are meant to defer user actions, for instance to let them change several parameters of a method before it's run. We (should) never defer the widget's reaction upon receiving a new signal, hence `now`. Both calls will thus remain valid for as long as there exist computations that can't be cancelled. Including in add-ons. :)

I was referring to this comment: https://github.com/biolab/orange3/issues/5453#issuecomment-864995825

You're right, widgets should commit when receiving a new signal. And that's the only use for `now`. When is more than this necessary?
```python
def handleNewVars():
    self.commit.now()
```
If threading is a problem, we could queue it in the event loop:
```python
def handleNewVars():
    QTimer.singleShot(self.commit.now)
```
If this is the only use for `now`, maybe it'd be cleaner and simpler to enforce it in a base class. You said there were some weird cases with `now` possibly being called from separate threads. Could you point them out? I'm wondering if these cases legitimate or just weirdly written widgets. Most of the uses I find are as simple as the above examples.

janezd:
I assume that by `handleNewVars` you mean `handleNewSignals`.

I would hate having this in the base class. First, many widgets that override `handleNewSignals` don't call `super`, sometimes for a good reason. Second, widgets that don't override it, call `commit` from individual input signal handlers. Widgets are currently not written like that. Putting this into the base class would be a big change.

And, most importantly, the logic around committing may be very complicated. We must have control over when commit happens. We don't want to have this done automagically in the base class. The proposed solution is explicit and clean.

ales-erjavec:
Workflow:
Screen Shot 2021-07-20 at 11 55 40

Change the selected file in *File* -> only the last added *Unique (2)* updates the output.

Delete the *Unique (2)* and reload selected file in *File* -> a runtime error is raised.

janezd:
Continued in https://github.com/biolab/orange-widget-base/pull/157#issuecomment-883982093.

Orange/widgets/data/tests/test_owunique.py
```diff
@@ -71,22 +71,22 @@ def test_compute(self):
         w.selected_vars = w.var_model[:2]
 
         w.tiebreaker = "Last instance"
-        w.commit()
+        w.commit.now()
```
janezd, line 74: In tests - in particular in places like this - `commit.now` is more appropriate than `commit.deferred`, which relies on auto apply being on.

Orange/widgets/data/owpaintdata.py
```diff
@@ -818,7 +818,7 @@ def __init__(self):
         self.tools_cache = {}
 
         self._init_ui()
-        self.commit()
+        self.commit.now()
```
janezd, line 821: I changed this to `now`, not `deferred`, because it is like getting new data.

janezd:
This changes all core widgets, with a few exceptions described below. I believe this is now ready for review (note that it requires https://github.com/biolab/orange-widget-base/pull/157; I made a few changes to that PR today, so the last version is needed).

A few comments:

- Calibration plot, Distributions: in signal handler, deferred was used instead of now. This further convinced me that calling `commit` should raise an exception instead of defaulting to now or deferred. We were making two mistakes: calling immediate commit because of forgetting to use (weird) lambdas, and not calling `unconditional_commit` where we should have (because we forgot that `commit` was replaced). If `commit` defaulted to anything, we'd still make one mistake or the other.

- I skipped category Model. See #5577.

- I skipped Louvain Clustering, where @aleserjavec, did this:

    ```python
    self.apply_button = gui.auto_apply(
                self.buttonsArea, self, "auto_commit",
                commit=lambda: self.commit(), callback=lambda: self._on_auto_commit_changed()
            )
    ```

    I would like to keep this update trivial, so I'll think about the effect of using lambda here ... later.

The widgets I did not modify will raise warnings for the time being.

PrimozGodec:
> > Do not forget to change `orange-widget-base` version to a version with deferred implementation.
> 
> I did it in requirements-gui.txt. No other places set a specific version, right? (I admit I longer understand all details of our CI.)

Yap, that it is set only in requirements-gui.txt.

PR #5509, opened 2021-06-29
OWPredictions: Migrate to AbstractTableModel

### LINKED ISSUE #5405: Predictions widget sorts numerical columns as text
The table view of the _**Predictions**_ widget - when clicked on column header - sorts numerical values like text instead of number in the _**target**_ and _**prediction**_ columns but NOT in the rest of the data columns. (e.g. "100.0" < "58.0" because it starts with a '1')

image

janezd:
##### Issue

#5405 could be fixed by properly defining the key for sorting. `orangewidget.utils.itemmodels.AbstractTableModel` however already implements some required functionality in a simpler way, besides being faster because it uses numpy's sorting instead of Qt's, which uses Qt's indices and comparison functions in Python. Furthermore, its use of numpy indices for mapping allows for a much simpler sorting of twin tables in owpredictions.

Requires https://github.com/biolab/orange-widget-base/pull/159.

Fixes #5405.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Huh, works for me.

What's the exception? If you can replicate it, do you see what's wrong and can fix it?

VesnaT:
I can replicate it any time.
```
---------------------------- ValueError Exception -----------------------------
  File "/Users/vesna/orange3/Orange/widgets/evaluate/owpredictions.py", line 736, in initStyleOption
    super().initStyleOption(option, index)
  File "/Users/vesna/orange-widget-base/orangewidget/utils/itemdelegates.py", line 397, in initStyleOption
    init_style_option(self, option, index, data, self.roles)
  File "/Users/vesna/orange-widget-base/orangewidget/utils/itemdelegates.py", line 196, in init_style_option
    option.text = delegate.displayText(value, option.locale)
  File "/Users/vesna/orange3/Orange/widgets/evaluate/owpredictions.py", line 725, in displayText
    return self.fmt.format(value=value, dist=dist)
ValueError: Unknown format code 'f' for object of type 'numpy.str_'
```

janezd:
Me too, now. I don't know, why it didn't crash in my previous attempts.

It looks like a simple, but nasty bug -- it seems that the problem is that delegate is updated only after updating the model. If I update the delegate first, it fails because the model is updated before the delegate. Resetting the model first has its own problems.

janezd:
This should do it: https://github.com/biolab/orange3/pull/5509/files#diff-dd77f983a656d8fa3f453496f08aad0ae7e8c5a57e48406d5520836c76ce1690R220.

`_set_class_values` triggers a callback that updates the delegate before the data is ready. Setting the class values later crashes something else, so it's simpler to just remove the model while new data is being set up.

VesnaT:
The above case now works, but the widget crashes when I select a row in the table:
File (Iris) -> Predictions (Select a row).

```
  File "/Users/vesna/orange3/Orange/widgets/evaluate/owpredictions.py", line 1031, in select
    self.store.select(selection, flags)
  File "/Users/vesna/orange3/Orange/widgets/evaluate/owpredictions.py", line 967, in select
    self.select_rows(rows, flags)
  File "/Users/vesna/orange3/Orange/widgets/evaluate/owpredictions.py", line 987, in select_rows
    self._rows ^= rows
  File "/Users/vesna/miniconda3/lib/python3.8/contextlib.py", line 120, in __exit__
    next(self.gen)
  File "/Users/vesna/orange3/Orange/widgets/evaluate/owpredictions.py", line 1016, in _emit_changed
    model.emit_selection_rows_changed(selected, deselected)
  File "/Users/vesna/orange3/Orange/widgets/evaluate/owpredictions.py", line 1061, in emit_selection_rows_changed
    self.selection_from_rows(selected),
  File "/Users/vesna/orange3/Orange/widgets/evaluate/owpredictions.py", line 1043, in selection_from_rows
    index = model.index
AttributeError: 'NoneType' object has no attribute 'index'
```

Issue #5468, opened 2021-06-04
Add explained variance to PCA variable name


robertcv:
I would like to have a checkbox in PCA widget that would add the percentage explained variance of each PCA axis to its variable name:

`PC1` -> `PC1(35%)`

Currently, I have to do it through Edit Domain and manually change the names. I would argue that the explained variance is as important as the actual values when visualizing the PCA in a scatter plot and it should therefore be easier to include it in the plot.

markotoplak:
This could create problems. Feature names are used in context matching. If we did what you propose then viewing the PCA components in a scatterplot and changing anything prior to PCA (or its normalization) would change the variance and invalidate the selected axes in the PCA.

I know you were asking for it as an option, but I'd try to avoid it if we can solve the problem in any other way.

Perhaps we could make it a Variables's attribute.

janezd:
At the meeting we discussed that
- the graph could show explained variance (per component) on a separate test data (this is unrelated to this request),
- could present a table of with explained variances,
- could add this number as an attribute, which would be useful if Transpose widget turns this into columns (does it?).

ajdapretnar:
> which would be useful if Transpose widget turns this into columns (does it?)

It does. For example, "bow-feature" and "hidden" from Text add-on become meta attributes.

PR #5513, opened 2021-07-04
[ENH] PCA: Output variance of components

janezd:
##### Issue

Fixes two thirds of #5468.

To test before https://github.com/biolab/orange-widget-base/pull/162 is merged, change

```python
        self.varmodel[:] = zip(self._variance_ratio[:self.maxp],
                               numpy.cumsum(self._variance_ratio))
```

to 

```python
        self.varmodel.wrap(list(zip(self._variance_ratio[:self.maxp],
                               numpy.cumsum(self._variance_ratio))))
```

##### Description of changes

1. Variances are added as attributes of variables, and as extra meta columns in the components output.

2. The table with variances is shown in control area and can be used as the third mean to select components. Clicking a row selects all components up to that row. (Selecting arbitrary rows is not possible: the widget's code does not support it; interactions the other two controls would be weird; this is not typical for PCA; for atypical uses, there's Select Rows; no a/b testing of this feature will be done, thanks for not asking :).

    Screenshot 2021-07-04 at 13 03 54

3. The third point in the issue was reporting variance on new data. This can be added to the table, but I'd suggest this to be a separate issue/PR.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owpca.py
+++ b/Orange/widgets/unsupervised/owpca.py
@@ -293,22 +293,36 @@ class OWPCA(widget.OWWidget):
                 self._transformed = self._pca(self.data)
             transformed = self._transformed
 
+            if self._variance_ratio is not None:
+                for var, explvar in zip(
+                        transformed.domain.attributes,
+                        self._variance_ratio[:self.ncomponents]):
+                    var.attributes["variance"] = round(explvar, 6)
             domain = Domain(
                 transformed.domain.attributes[:self.ncomponents],
                 self.data.domain.class_vars,
                 self.data.domain.metas
             )
             transformed = transformed.from_table(domain, transformed)
+
             # prevent caching new features by defining compute_value
             proposed = [a.name for a in self._pca.orig_domain.attributes]
             meta_name = get_unique_names(proposed, 'components')
-            dom = Domain(
-                [ContinuousVariable(name, compute_value=lambda _: None)
-                 for name in proposed],
-                metas=[StringVariable(name=meta_name)])
+            meta_vars = [StringVariable(name=meta_name)]
             metas = numpy.array([['PC{}'.format(i + 1)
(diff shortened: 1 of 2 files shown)


janezd:
I added variance on test data. (Not thoroughly checked, yet.) The table is a bit wide, but OK. The plot may look nice

Screenshot 2021-07-05 at 10 27 31

... except when it doesn't:

Screenshot 2021-07-05 at 10 25 52

I hate to think about how to compute the positions of labels in pyqtgraph to prevent overlapping. Any good suggestions?

ajdapretnar:
I love how this widget got from basic to jacked 💪  in no time. 😆 

As for the label, if points for labeling are <= to the height of the font, then plot the higher one above, not below the point it labels? (stupid suggestion probably, just throwing ideas out there)

janezd:
~~The problem is that in pyqtgraph you set the position in plot coordinates, not in pixels. There are several functions that transform coordinates it's all black magic, at least to @VesnaT and me. I'm thus not even sure I can detect that labels are overlapping.~~

Ha! Labels are `QGraphicsTextItem`'s, so Qt tells you if they collide. In case they do, I just set the anchors as you suggested and it works like magic. Sometimes the label would overlap with the curve, which didn't happen before because curves were monotone. But it was good for my grandmother, Lord it's good enough for me!

markotoplak:
I rebased this and am (for now) only merging the first commit. The others are backed-up at https://github.com/markotoplak/orange3/tree/pca-show-variance-backup

In between, a group from Mannheim, particularly @Les-Simon, proposed some changes to the PCA (and other widgets): https://github.com/Charisma-Mannheim/OrangeExtension. They tend to work with PCA a lot, much more than we do. We should consider their suggestions before adding additional complexity to the widget.

PR #5539, opened 2021-08-01
Merge Data: Embiggen + and x buttons

### LINKED ISSUE #5285: OWMergeData: + and x symbols are small and hard to click
Not sure if this happens only on Windows or elsewhere too.

It should be implemented like the buttons are in OWCreateClass.

janezd:
##### Issue

Fixes #5285.

##### Description of changes

As proposed, adapt the code from Create Class.

~~The widget shows empty/disabled buttons, that is, grayed-out squares instead of hiding the button completely. I find it kind-of nice, but can change it.~~

I have also put the + button in a separate row. This looks nicer, though it uses some extra space; the widget is small, anyway. (The previous design was, I guess, OK because the buttons looked more like labels that are shown or hidden. With buttons, it is not.)

##### Includes
- [X] Code changes
- [X] Test changes

### DIFF
--- a/Orange/widgets/data/owmergedata.py
+++ b/Orange/widgets/data/owmergedata.py
@@ -5,8 +5,9 @@ import numpy as np
 
 from AnyQt.QtCore import Qt, QModelIndex, pyqtSignal as Signal
 from AnyQt.QtWidgets import (
-    QWidget, QLabel, QPushButton, QVBoxLayout, QHBoxLayout
+    QWidget, QLabel, QVBoxLayout, QHBoxLayout, QSizePolicy
 )
+
 from orangewidget.utils.combobox import ComboBoxSearch
 
 import Orange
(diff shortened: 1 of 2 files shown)


thocevar:
This is a nice improvement on Windows. Before, the symbols were tiny, partially cut off and did not seem clickable. Buttons fix all of this.

PrimozGodec:
Buttons look ok on macOS

Screenshot 2021-08-06 at 11 10 04

Maybe one note (which was already there before): phrase matches in the first row does not align with other rows. Should we align that too or are we ok with the current solution?

borondics:
MacOS window arrangements have serious issues I think, which is due to some obscure QT stuff (@markotoplak has some stories to tell about this). It would be great to have similar spacing in the layouts as for example in Windows because some widgets become humongous (see Fitting in Orange Spectroscopy) if using multiple text boxes or other elements. In any case, all improvements are much appreciated!

ajdapretnar:
Agree that dropdowns and "matches" should be aligned.

markotoplak:
Alignment is fine on my computer, just the spacing is different from the one in Create Class. I am fine with it being different.

screenshot

PR #5541, opened 2021-08-02
New widget: Group By 

PrimozGodec:
##### Issue

Orange is missing group by - aggregate widget with similar behaviour than Panda groupby - aggregate

##### Description of changes
New groupby widget

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


ajdapretnar:
When the data set contains TimeVariable, I get this error (try with _Banking Crises_ from Datasets):

```
  File "/Users/ajda/opt/miniconda3/envs/o3/lib/python3.8/site-packages/orangecanvas/scheme/signalmanager.py", line 1024, in __process_next
  File "/Users/ajda/opt/miniconda3/envs/o3/lib/python3.8/site-packages/orangecanvas/scheme/signalmanager.py", line 1062, in __process_next_helper
  File "/Users/ajda/opt/miniconda3/envs/o3/lib/python3.8/site-packages/orangecanvas/scheme/signalmanager.py", line 690, in process_node
  File "/Users/ajda/orange/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 806, in send_to_node
  File "/Users/ajda/orange/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 820, in process_signals_for_widget
    process_signals_for_widget(widget, signals, workflow)
  File "/Users/ajda/opt/miniconda3/envs/o3/lib/python3.8/functools.py", line 875, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/ajda/orange/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 921, in process_signals_for_widget
    process_signal_input(input_meta, widget, signal, workflow)
  File "/Users/ajda/opt/miniconda3/envs/o3/lib/python3.8/functools.py", line 875, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/ajda/orange/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 884, in process_signal_input_default
    notify_input_helper(
  File "/Users/ajda/opt/miniconda3/envs/o3/lib/python3.8/functools.py", line 875, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/ajda/orange/orange-widget-base/orangewidget/utils/signals.py", line 643, in set_input_helper
    handler(*args)
  File "/Users/ajda/orange/orange-widget-base/orangewidget/utils/signals.py", line 198, in summarize_wrapper
    method(widget, value)
  File "/Users/ajda/orange/orange3/Orange/widgets/data/owgroupby.py", line 507, in set_data
    {
  File "/Users/ajda/orange/orange3/Orange/widgets/data/owgroupby.py", line 508, in <dictcomp>
    attr: DEFAULT_AGGREGATIONS[type(attr)].copy()
KeyError: <class 'Orange.data.variable.TimeVariable'>
```

ajdapretnar:
Also, wouldn't it make sense to put the grouped by variables in metas for the output data table since they are unique?

ajdapretnar:
Filter doesn't work properly.

Screen Shot 2021-08-04 at 16 35 47

ajdapretnar:
Corpus instances are not retained (i.e. a grouped Corpus is not a Corpus on the output). The problem is (possibly) how the `self.text_variable` is treated in the new data set. We could perhaps just set it as it was if the variable is present in the new data set, else we output a Table?

PrimozGodec:
This PR need https://github.com/biolab/orange3/pull/5547 to work properly for TimeVariables

PrimozGodec:
> Filter doesn't work properly. Also when connecting new data, the filter value should probably be reset (unless perhaps the query matches certain variables). Dunno, just an idea.

It is a bug in ListViewSearch. Will be addressed in the separate PR since it is an issue that also appears elsewhere.

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+        checked or just between unchecked and checked - depending on situation
+        """
+        if self.checkState() == Qt.Checked:
+            # if checked always uncheck
+            self.setCheckState(Qt.Unchecked)
+        else:
+            agg = self.text()
+            selected_attrs = self.parent.get_selected_attributes()
+            types_ = set(type(attr) for attr in selected_attrs)
+            can_be_applied_all = (AGGREGATIONS[agg].types & types_) == types_
```
janezd, line 253: Isn't this the same as `types_ <= AGGREGATIONS[agg].types`?

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+        time. In this case we step between unchecked, partially checked and
+        checked or just between unchecked and checked - depending on situation
+        """
+        if self.checkState() == Qt.Checked:
+            # if checked always uncheck
+            self.setCheckState(Qt.Unchecked)
+        else:
+            agg = self.text()
+            selected_attrs = self.parent.get_selected_attributes()
+            types_ = set(type(attr) for attr in selected_attrs)
```
janezd, line 252: Nitpicking: why `types_` and not `types`?

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+                self.setCheckState(
+                    Qt.Checked if can_be_applied_all else Qt.PartiallyChecked
+                )
+
+
+@contextmanager
+def block_signals(widget):
+    widget.blockSignals(True)
+    yield
+    widget.blockSignals(False)
```
janezd, line 286: I think you must do

```
try:
    yield
finally:
    widget.blockSignals(False)
```

(https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager)

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def __init__(self):
+        super().__init__()
+        ConcurrentWidgetMixin.__init__(self)
+
+        self.data = None
+        self.result = None
+
+        self.gb_attrs_model = DomainModel(
+            valid_types=(DiscreteVariable, ContinuousVariable, StringVariable),
```
janezd, line 318: Why prohibit `TimeVariable`? It is just as discrete as `ContinuousVariable`.

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+        tableview.selectionModel().selectionChanged.connect(self.__rows_selected)
+        tableview.horizontalHeader().setSectionResizeMode(QHeaderView.Stretch)
+
+        vbox = gui.vBox(self.mainArea, " ")
+        vbox.layout().addWidget(tableview)
+
+        # aggregations checkboxes
+        grid_layout = QGridLayout()
+        gui.widgetBox(self.mainArea, orientation=grid_layout, box="Aggregations")
+        itm_col = ceil(len(AGGREGATIONS) / 3)
```
janezd, line 354: Interestingly, I'd write `(len(AGGREGATIONS) + 2) // 3` here. Always.

This is just an observation, not a suggestion. I guess it shows I'm old-school. My first CPUs didn't have floats. :) (Though: the difference is that `ceil` returns a `float`, while I get an `int`.)

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+            grid_layout.addWidget(cb, i % itm_col, i // itm_col)
+
+    ############
+    # Callbacks
+
+    def __rows_selected(self) -> None:
+        """Callback for table selection change; update checkboxes"""
+        selected_attrs = self.get_selected_attributes()
+
+        types_ = set(type(attr) for attr in selected_attrs)
```
janezd, line 369: It is interesting that we never write `list(type(attr) for attr in selected_attrs)` instead of `[type(attr) for attr in selected_attrs]`, but we often write `set(type(attr) for attr in selected_attrs)` instead of `{type(attr) for attr in selected_attrs}`.

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+        active_aggregations = [self.aggregations[attr] for attr in selected_attrs]
+        for agg, cb in self.agg_checkboxes.items():
+            cb.setDisabled(len(types_ & AGGREGATIONS[agg].types) == 0)
```
janezd, line 372: Prehaps `not types_ & AGGREGATIONS[agg].types`?

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+
+            activated = [agg in a for a in active_aggregations]
```
janezd, line 374: Just an observation: it would be fun to use a set here, `activated = {agg in a for a in active_aggregations}`, because the condition would then be `Qt.checked if activated == {True} else (Qt.Unchecked if activated == {False} else Qt.PartiallyChecked)`. :)

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+                for attr in data.domain.variables + data.domain.metas
+            }
+            if data
+            else {}
+        )
+
+        self.openContext(self.data)
+
+        # update selections in widgets and re-plot
+        self.agg_table_model.set_domain(data.domain if data else None)
```
janezd, line 435: Domain is already set above, before opening the context (as it should be). Is there any reason to repeat this?

Orange/data/aggregate.py
```diff
@@ -0,0 +1,117 @@
...
+
+import pandas as pd
+
+from Orange.data import Domain, Table, Variable, table_from_frame, table_to_frame
+from Orange.util import dummy_callback
+
+
+class OrangeTableGroupBy:
+    """
+    The class which object is the result of the groupby operation on Orange's
```
janezd, line 12: A class representing the result of ...

janezd (approved):
Nicely done.

I wrote some minor comments. I'll have more in person. :)

Orange/widgets/data/owgroupby.py
```diff
@@ -0,0 +1,490 @@
...
+
+def concatenate(x):
+    """
+    Concatenate values of series if value is not missing (nan or empty string
+    for StringVariable)
+    """
+    return " ".join(str(v) for v in x if not pd.isnull(v) and len(str(v)) > 0)
+
+
+AGGREGATIONS = {
```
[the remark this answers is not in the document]
ajdapretnar: "List of values" doesn't really sound right when talking about strings I would say.

Orange/widgets/data/owgroupby.py
PrimozGodec: It was already included since it is a subclass of ContinuousVariable. For the purpose of code readability, I added it.

Orange/widgets/data/owgroupby.py
PrimozGodec: In my case, the reason for this is the fact that `{}` produces dict and not set and for an empty set, one must write `set()`. Then sometimes I automatically use set(...) in comprehensions instead of {...}

mw25:
Thank you for this widget! I needed this functionality too and was about to start trying an implementation until I saw this pull request.

I like it very much, but if I may, I would like to add some more ideas:

- How about an aggregation "None" where only grouping is done, but no aggregation. Or am I missing something and this is already easier to do with another widget?

I work with time series data and when I group and aggregate I also always have to select the time range of the group to aggregate over. So from a group or stage always only e.g. the last 10 values or 10 minutes or similar are selected and aggregated, because first a settling process must be waited for. Do you understand what I mean? I think this is a relatively common approach.

- For this the function pandas.core.groupby.GroupBy.tail would be good, where you can specify the last n values.
- For time series data, it's a bit more complicated (The last x seconds/minutes/... of each group). I think you would have to take a combination of pandas.core.groupby.GroupBy.apply and pandas.DataFrame.last, where you can pass arbitrary time periods.  In a python script I do it like this: 
`df.groupby("B").apply(lambda df: df.last('2min').mean())`

For each of these aggregations, an additional parameter is required. Also, you might need to use two groupby widgets in a row, the first for grouping and cutting the range, the second for a subsequent aggregation like mean.

The apply function could perhaps be called "Custom" and allow text input to use other pandas functions. 
See screenshot. (I only extended the GUI, there is no function behind it).

What do you guys think about this?
I hope that was understandable and not too much.
Thanks again for the status so far.

groupby-ideas

Orange/widgets/data/owgroupby.py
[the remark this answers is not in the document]
janezd: This is not a list of possible values but space-separated values that appear in the column (with repetitions).

I agree "List of values" is ambiguous. Concatenate may be better after all, unless we come up with something better.

Orange/widgets/data/owgroupby.py
janezd: If all types are valid, I'd suggest to increase the readability by omitting the parameter. :) Allowing all is what `DomainModel` does by default.

Orange/widgets/data/owgroupby.py
janezd: Yes, that's probably the reason for all of us.

Orange/widgets/data/owgroupby.py
PrimozGodec: `valid_types` is redundant since we include all variables -- I removed it. It was here since I probably want to make aggregation differently in the beginning.

janezd:
> How about an aggregation "None" where only grouping is done, but no aggregation. Or am I missing something and this is already easier to do with another widget?

A widget called *Unique* already does it. Also, if you are not interested in any aggregation, you can use whichever you want and ignore it. :) Though your suggestion is still reasonable.

Orange/widgets/data/owgroupby.py
PrimozGodec: Above we set the domain to `gb_attrs_model` (attribute model for the group by list). Here we set the domain to `agg_table_model` which is used for the table. It is set after the context is open to plot the table with settings opened by the context (aggreagtions). Initializing it before the context is open would require another call to the table to replot after the context is open.

mw25:
> > How about an aggregation "None" where only grouping is done, but no aggregation. Or am I missing something and this is already easier to do with another widget?
> 
> A widget called _Unique_ already does it. Also, if you are not interested in any aggregation, you can use whichever you want and ignore it. :) Though your suggestion is still reasonable.

Thanks for the hint.
To be honest, I can't remember which use case I had in mind with the "None" option...

mw25:
> 
> For each of these aggregations, an additional parameter is required. Also, you might need to use two groupby widgets in a row, the first for grouping and cutting the range, the second for a subsequent aggregation like mean.
> 

Here could be a problem: The normal functions return one value per group, mine suggested usually several. What happens to the output table if "Mean" is selected for one signal and "Tail" for another?
Maybe the functions I suggested (if they are wanted at all) should not be considered as further aggregate functions, but as optional preprocessing.

When selecting a time range (with e.g. last('10s')) the selection of the time channel to be used would also be necessary.

Orange/widgets/data/owgroupby.py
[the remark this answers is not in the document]
PrimozGodec: I also don't have a better idea for now. Will left concatenate and we can change later

PrimozGodec:
> > How about an aggregation "None" where only grouping is done, but no aggregation. Or am I missing something and this is already easier to do with another widget?
> 
> A widget called _Unique_ already does it. Also, if you are not interested in any aggregation, you can use whichever you want and ignore it. :) Though your suggestion is still reasonable.

What unique widget do can be already done with this widget (except the option `Discard non-unique instances`) with using `Frist value`, `Last value`, ...

I think @mw25 had in mind a case where we just group and do not attach any aggregation to it. For example, if we have a table with attributes a, b, c and we group by a and b the table would include only columns a and b (with unique values combinations). **Do I understand it right?** It cannot be done with the unique widget. 

I actually think that Group By widget should return this kind of "aggregation" when turning off all aggregations on all attributes. Currently, the error is raised by Pandas.  I will fix that.

mw25:
> I think @mw25 had in mind a case where we just group and do not attach any aggregation to it. For example, if we have a table with attributes a, b, c and we group by a and b the table would include only columns a and b (with unique values combinations). **Do I understand it right?** It cannot be done with the unique widget.
> 

I tried hard again to remember what I originally wanted =D
If I group my table (columns A, B, C) by A and B I wanted the following result:
The signals stay as they are, so the number of rows does not change. But a new column is added with the grouping. So for example a new numbering (that would be the result of pandas.core.groupby.GroupBy.ngroup) or the values of the columns A and B as text.

groupby

ajdapretnar:
You can already achieve the second option with Feature Constructor and `str(A)+"-"+str(B)` option.

PrimozGodec:
> I work with time series data and when I group and aggregate I also always have to select the time range of the group to aggregate over. So from a group or stage always only e.g. the last 10 values or 10 minutes or similar are selected and aggregated, because first a settling process must be waited for. Do you understand what I mean? I think this is a relatively common approach.
> 
> For this the function pandas.core.groupby.GroupBy.tail would be good, where you can specify the last n values.
> For time series data, it's a bit more complicated (The last x seconds/minutes/... of each group). I think you would have to take a combination of pandas.core.groupby.GroupBy.apply and pandas.DataFrame.last, where you can pass arbitrary time periods. In a python script I do it like this:
> df.groupby("B").apply(lambda df: df.last('2min').mean())

Regarding the other two aggregations:
- tail (and possibly the head) would be confusing. What happens when the user select tail (with 10 values) and mean? Mean returns only one value per group tail multiple. What to output then. It would be actually the complete new widget. 
- custom function is possible but the question is if is not too complicated for users? What do you think @janezd and @ajdapretnar.

mw25:
> You can already achieve the second option with Feature Constructor and str(A)+"-"+str(B) option.

Oh yes that's right. I knew I missed something...

> tail (and possibly the head) would be confusing. What happens when the user select tail (with 10 values) and mean? Mean returns only one value per group tail multiple. What to output then. It would be actually the complete new widget.

@PrimozGodec Yes, I noticed that afterwards, too. Did you see my comment:

> Maybe the functions I suggested (if they are wanted at all) should not be considered as further aggregate functions, but as optional preprocessing.

These preprocessing options would then of course apply to all signals and be listed separately in the widget.

ajdapretnar:
Personally, I'd have a separate widget in Timeseries to handle such use cases. I'd prefer to keep the core Group by widget simple and straightforward.

janezd:
> custom function is possible but the question is if is not too complicated for users? What do you think @janezd and @ajdapretnar.

I didn't want to comment it, but if you ask me directly, I will: I don't like it. For those who want to code, there are Feature Constructor and Python Script. If we add it here, we'll want to add it to the next widget next week and soon many widgets will have inputs that will only be understood by programmers.

mw25:
Thank you for your feedback.

I think we agree that a custom option is not an option.

But what about another box (e.g. in the controlArea under the signal list) with an checkable option "Preprocessing" or "Range selection" or something? There could be a dropdown with the functions tail (n values), head (n values), tail (n seconds), head (n seconds), ... and another input field for n (and maybe another one for the time channel?).

Would this still be possible in the groupby widget? Otherwise I think @ajdapretnar idea to implement this in timeseries is good. 
But what would be the procedure?  Should the groupby widget be derived and simply extended with these elements?
(Of course I would also participate).

mw25:
I have extended OrangeTableGroupBy and Table.groupby() a bit to allow preprocessing / range selection for each group. 

Regardless of whether the preprocessing is placed in the groupby widget or a timeseries widget, or is only available via scripting (or only in a private widget), I would like to propose my extension.

How can I commit or propose my changes?

ajdapretnar:
@mw25 As I think your preprocessing is very much related to timeseries, I would propose you make a PR on the [Timeseries repository](https://github.com/biolab/orange3-timeseries), where we can review it and merge it. The PR should include at least tests and some documentation. Ideally, this would be a widget for timeseries preprocessing called Preprocess Timeseries. That's how I see it.

PR #5547, opened 2021-08-10
pandas_compat: fix conversion of datetime series

PrimozGodec:
##### Issue

- Currently, dates and dimes are converted from pd.DataFrame to the Table via string date-time representation. Conversion between frame and table fails since some string formats that Pandas produce are not compatible with `datetime.strftime` and specified formats (e.g. pandas output time on nanosecond precision while DateTime only support reading of milliseconds precision max, and some other cases). It could be solvable via some string manipulation which is heavily error prune.
- Current conversion is slow (20 s for 1M dates).

##### Description of changes
- New implementation of the conversion that does not use string 
- Speedup (now 0.5 s for 1M dates)
- Deprecating TimeVariable.utc_offset since it is only used in one place in Orange3-timeseries and can be implemented using `timezone` attribute. Storing utc_offset in TimeVariable is also not a good practice since two dates in the same timezone can have two different offsets depending on summer, winter time or time zones changes in history. Remembering the timezone is more suitable.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/data/pandas_compat.py
```diff
@@ -157,13 +157,56 @@ def _is_datetime(s):
...
 
 
+def _convert_datetime(series, var):
+    def col_type(dt):
+        """Test if is date, time or datetime"""
+        dt_nonnat = dt[~pd.isnull(dt)]  # nat == nat is False
+        if (dt_nonnat.dt.floor("d") == dt_nonnat).all():
+            # all times are 00:00:00.0 - pure date
+            return 1, 0
+        elif (dt_nonnat.dt.date == pd.Timestamp("now").date()).all():
```
ajdapretnar, line 175: Does pandas automatically assume today's date if only time is present?

PrimozGodec: yep. When `pd.datetime()` is called on column of times it will be transformed to datetime64 and all dates will be today's dates

PR #5549, opened 2021-08-11
OWScatterPlotBase: Animate point resize for half a second

VesnaT:
##### Issue

Projection widgets animate point resize. The animation should last half a second.

##### Description of changes
Resize points within approximately half a second.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
This can be merged, but ... whose idea is that it should last a second and am I the only one who thinks this is too slow?

VesnaT:
After discussing it with @BlazZupan and @markotoplak, we agreed on half a second long animation.

PR #5555, opened 2021-08-17
OWPredictions: Set proper delegate options when window has no focus

### LINKED ISSUE #5554: Predictions: wrong colors in inactive state
**What's wrong?**

Selection background is black even when the widget has no keyboard focus.

Open the widget and unfocus it.

This happens because the widget's table is composed of two tables. Only one can have focus, but both pretend to have it, otherwise one would have a grey selection. They do so by setting the state to active and has focus without checking whether the widget itself has focus or not.

janezd:
##### Issue

Fixes #5554.

##### Description of changes

- Do not set `option` for `paint` as `Active` when the window doesn't have keyboard focus.
- Refactor the two delegates to use a common base class with `initStyleOption`.

##### Includes
- [X] Code changes


janezd:
Coverage test doesn't and won't pass. To cover the missed line I could mock that the window is active and check that the flag is set correctly, but that's pointless.

Issue #5559, opened 2021-08-23
Orange not Launching


Killorito:
I am trying to use Orange for some of the tasks that i want to complete.

However, when i was trying to install an add-on for orange, an error came up. I disregarded the error, and after a while closed the application. Now, every time i try and launch Orange from Anaconda, it is not working. Anaconda displays a loading sign when i want to launch Orange, yet it stills does not launch after the sign goes away. I have also tried logging on and off of anaconda to see if it works or not.

I have even installed Orange separately from anaconda, yet it still does not open up!

What can i do? Have i done something wrong?

ajdapretnar:
Sorry for the inconvenience. Could you perhaps explain which add-on did you install and on what operating system. How did you install the add-on (via add-on dialogue)?

My assumption is the add-on likely corrupted some libraries, which are now incompatible and thus cause the error.

I would try removing Orange and installing it once again so that the incompatible libraries would be overridden.

However, I am confused about the final statement. You installed Orange with the installer and it is still not launching?

markotoplak:
@Killorito, as @ajdapretnar pointed out, installing some add-ons into your Anaconda environment could break it (incompatible versions of some packages may be installed). Please use the official Orange packages from https://orangedatamining.com/download/

Alternatively, add the conda-forge repository to your Anaconda installation and only then install Orange and the add-ons.

We would very much like to know which add-on did you install to know how to fix your problem in the future. Thanks!

Killorito:
@ajdapretnar Thank you for your response. I am using Windows 10, Orange 3 version 3.26.  I installed the add ons from the add ons dialogue box on the top. I have uninstalled Orange and reinstalled it. I have gone through, deleting all the files that are attributed to orange, yet again, it still does not work.

However, i have made a new windows account, and i see that orange works fine. 

The add-on that i saw had the issue was the "Networking" add on. That is what caused the issue.

Killorito:
@markotoplak I have used the official package as well, yet it still does not load. I installed the networking add on.

Issue #5070, opened 2020-11-06
ValueError: PyCapsule_GetPointer called with incorrect name


liar666:
Hello everybody,

I wanted to give a try to `Orange3`.

I'm running `python 3.8.5` with a `venv` specific for datascience tools.

I've upgraded all my packages in this `venv` and installed orange with `python3 -m pip install orange3`

Everything went OK.

- [ ] **What's wrong?**

When I try to run Orange with `python3 -m Orange.canvas` or the `orange-canvas` script I get:
```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/gmuller/Python/EEGs/lib/python3.8/site-packages/Orange/canvas/__main__.py", line 48, in <module>
    from orangewidget.workflow.errorreporting import handle_exception
  File "/home/gmuller/Python/EEGs/lib/python3.8/site-packages/orangewidget/workflow/errorreporting.py", line 29, in <module>
    from orangewidget.workflow.widgetsscheme import WidgetsScheme
  File "/home/gmuller/Python/EEGs/lib/python3.8/site-packages/orangewidget/workflow/widgetsscheme.py", line 49, in <module>
    from orangewidget.widget import OWBaseWidget
  File "/home/gmuller/Python/EEGs/lib/python3.8/site-packages/orangewidget/widget.py", line 22, in <module>
    from orangewidget.report import Report
  File "/home/gmuller/Python/EEGs/lib/python3.8/site-packages/orangewidget/report/__init__.py", line 4, in <module>
    from .report import *
  File "/home/gmuller/Python/EEGs/lib/python3.8/site-packages/orangewidget/report/report.py", line 13, in <module>
    from orangewidget.io import PngFormat
  File "/home/gmuller/Python/EEGs/lib/python3.8/site-packages/orangewidget/io.py", line 15, in <module>
    from orangewidget.utils.webview import WebviewWidget
  File "/home/gmuller/Python/EEGs/lib/python3.8/site-packages/orangewidget/utils/webview.py", line 30, in <module>
    from AnyQt.QtWebKitWidgets import QWebView
  File "/home/gmuller/Python/EEGs/lib/python3.8/site-packages/AnyQt/QtWebKitWidgets.py", line 14, in <module>
[...]

PR #5566, opened 2021-08-26
[FIX] Remove PyQt5 dependency

markotoplak:
The PyQt5 dependency conflicts with conda, where the same package is called pyqt. Installing pyqt5 inside a conda environment may be catastrophic. Removing the dependency makes pip installation into conda environments safer; such installations frequently occur when users installs an add-on not available in the conda repos that requires a newer version of Orange . The same issue appears when Anaconda users try to update Orange with the Add-on dialog box. Or when someone who installed Orange with conda downloads the master branch and does "pip install -e .".

Some related bug reports: gh-5559. gh-5070

Orange added the PyQt5 dependency about a year ago to make pip installations friendlier (gh-4849). Unfortunately, that caused more problems than it fixed.

For some context: ContinuumIO/anaconda-issues#1554

This needs at least the following addtional work:
- [x] Fix CI
- [x] Update pip installation instructions

But then, I am not sure how we should approach fixing #4763, which should be reopened. Perhaps Orange could print a warning if a non-supported version of Qt is used?

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ conda install orange3-<addon name>
 
 ### Installing with pip
 
-We recommend using our [standalone installer](https://orange.biolab.si/download) or conda, but Orange is also installable with pip. You will need a C/C++ compiler (on Windows we suggest using Microsoft Visual Studio Build Tools).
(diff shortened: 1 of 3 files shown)


janezd:
Maybe try importing PyQt at some early point and instead of raising an exception, print a message that Orange requires PyQt and how to install it.

markotoplak:
Replaced by gh-5593

Issue #5599, opened 2021-09-21
Gradient Boosting kills Orange


VesnaT:
**What's wrong?**

Orange unexpectedly terminates when tuning Gradient Boosting (xgboost) parameters. 

Log: 
[log.txt](https://github.com/biolab/orange3/files/7201197/log.txt)

**How can we reproduce the problem?** 

- Open the attached workflow. 
- Keep clicking (e.g. 10 clicks per second) the `Apply` button in the `Gradient Boosting` widget.

[xgboost_crash.ows.zip](https://github.com/biolab/orange3/files/7201211/xgboost_crash.ows.zip)

**What's your environment?**

- Operating system: macOS Catalina
- Orange version: master
- How you installed Orange: with pip from source

markotoplak:
Can not replicate on macOS Mojave with the 3.30 test build (https://dev.azure.com/markotoplak/markotoplak/_build/results?buildId=25&view=artifacts&pathAsName=false&type=publishedArtifacts)

Can not replicate on my Linux system.

PrimozGodec:
I cannot replicate it on Big Sur (11.5.2) with test build and neither with development version.

markotoplak:
@VesnaT can replicate it reliably both with 3.30 test build (running on Qt 5.15) and 3.29.3 (current official build, uses Qt 5.12). She also updated libomp and it did not help.

markotoplak:
She seems quite proud of her three-clicks-to-crash average.

janezd:
The crash, which occurs in thread 35 in your log, has `xgboost::DMatrix* xgboost::DMatrix::Create<xgboost::data::DenseAdapter>` on the stack, which looks like the same function that crashed in tests on Mac. If this is indeed the same bug, #5613 should provide a temporary workaround and the next release of xgboost should contain a fix.

VesnaT:
The temporary fix works for me.

PR #5608, opened 2021-09-27
[FIX] Fixes for scikit-learn 1.0

janezd:
##### Issue

Tests fail after release of scikit-learn 1.0.

- Scikit's fit checks whether the object has `columns` to distinguish between arrays and pandas frame. `Orange.data.Table` also has `columns`, but with different content, which caused a problem when (incorrectly) passed to scikit.
- ROC has tests for tooltips, which use test results for Titanic with kNN. Scikit changed something in kNN so that all returned probabilities are 0 or 1 (when k=5, this is not entirely unexpected).

##### Changes

- I fixed the call to kNN to pass an array `X`, not the whole `Table`.
- The offending test for ROC now uses artificial test results, not results from an actual model.

**To be considered:** `Orange.data.Table`'s property `columns`, which contains an instance of this: https://github.com/biolab/orange3/blob/03e8adaf41dac18d747c7f47afd7262c493fa951/Orange/data/table.py#L190  is unknown, strange, useless and to my knowledge unused.  I assume it was there for easier access to columns when working with Orange interactively. I would just remove it without any deprecation period.

##### Includes
- [X] Code changes
- [x] Tests
- Documentation N/A

### DIFF
--- a/Orange/widgets/evaluate/tests/test_owrocanalysis.py
+++ b/Orange/widgets/evaluate/tests/test_owrocanalysis.py
@@ -10,6 +10,7 @@ from AnyQt.QtWidgets import QToolTip
 from Orange.data import Table
 import Orange.evaluation
 import Orange.classification
+from Orange.evaluation import Results
 
 from Orange.widgets.evaluate import owrocanalysis
 from Orange.widgets.evaluate.owrocanalysis import OWROCAnalysis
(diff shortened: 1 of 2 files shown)


janezd:
I fixed the failing tests, but can't reproduce core dumps on Mac. Help wanted.

markotoplak:
I merged as is.

I tried reproducing the segfaults locally. At first, I could not, but after I installed xgboost (an optional Orange dependency) it reliably segfaults on my 10.14 Mac. I just need to do the following in a fresh virtualenv:

```
pip install orange3
pip install xgboost
python -m unittest Orange.tests
```

For xgboost to work, you also need libomp. Before trying anything, do `brew install libomp`.

@janezd, can you replicate?

janezd:
I can. `test_all_models_work_after_unpickling` segfaults after installing xgboost (and libomp).

janezd:
Actually, the learner just crashes, always. 

```
>>> XGBLearner()(Orange.data.Table("iris"))
Segmentation fault: 11
```

janezd:
We may be experiencing https://github.com/dmlc/xgboost/issues/7156: the crash happens at the same point, but if during debugging I change the value of variable `threads` from `-1` to `1`, it does not crash here (but crashes later, perhaps for a related reason).

Xgboost 1.5.0rc was released 4 days ago, and release is planned for Oct 11. I suggest we wait it out; till then, we can add a (time-bombed) exclusion for xgboost tests on MacOs.

markotoplak:
That was a got find! Let's hope that 1.5 fixes this segfault and perhaps also #5599.

PR #5613, opened 2021-09-30
XGBoost: Temporary workaround before release of v1.5

janezd:
##### Issue

XGBoost fails on macOS is the number of jobs is left at default (details in https://github.com/dmlc/xgboost/issues/7156). This should be fixed in new release at around Oct 11, 2021.

##### Description of changes

- Set `n_jobs` to `1` on macOS
- Add a test that will fail after Oct 15, to remind us to remove the workaround.

##### Includes
- [X] Code changes
- [X] Tests
- N/A: Documentation

### DIFF
--- a/Orange/classification/xgb.py
+++ b/Orange/classification/xgb.py
@@ -1,5 +1,6 @@
 # pylint: disable=too-many-arguments
 from typing import Tuple
+import sys
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


janezd:
@markotoplak, can you check and merge if OK? We can ignore pylint here.

markotoplak:
Thanks @janezd!

PR #5621, opened 2021-10-02
changed redundant badges into a normal link.

Joe-Sin7h:
##### Description of changes
Removed redundant badges and changed them into normal links, are they were already used in `README.md`. ex : discord badge, it's was used 3 times, which is totally unnecessary, so I converted them into normal links.

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 
 ### Easy installation
 
-For easy installation, [![Download](https://img.shields.io/github/v/release/biolab/orange3?label=download)](https://orange.biolab.si/download) the latest released Orange version from our website. To install an add-on, head to `Options -> Add-ons...` in the menu bar.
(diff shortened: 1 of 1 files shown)


Joe-Sin7h:
One more example 
Capture
:

here download badge is already below `README.md`  header, so I changed this badge into a normal `download` link.

markotoplak:
I wholeheartedly agree. When reading I automatically ignore these (then only interrupt the flow) and it would not occur to me to click on them.

Of course, after a careful consideration I might mouse-over them but, please, don't make me think.

Joe-Sin7h:
> I wholeheartedly agree. When reading I automatically ignore these (then only interrupt the flow) and it would not occur to me to click on them.
> 
> Of course, after a careful consideration I might mouse-over them but, please, don't make me think.

Thank you for your response. Could you merge it?

PR #5626, opened 2021-10-05
Catboost: do not test default learning rate in widget tests

markotoplak:
If learning_rate is None, which is the default, it is automatically defined. The previous version, 0.26.1, and the current version, 1.0, set learning_rate differently. It does not make sense to test the default learner value in the widget: the widget imposes a set value anyway and therefore does neither use now allow the automatic setting.

##### Issue

```
FAIL: test_default_parameters_reg (Orange.widgets.model.tests.test_owgradientboosting.TestCatGBLearnerEditor)
  File "/home/runner/work/orange3/orange3/.tox/orange-released/lib/python3.8/site-packages/Orange/widgets/model/tests/test_owgradientboosting.py", line 298, in test_default_parameters_reg
    self.assertEqual(round(params["learning_rate"], 3), 0.035)
AssertionError: 0.037 != 0.035
```


markotoplak:
This fixes the failing test, but despite merging #5613 we still have segfaults on macs. They are unrelated to this PR and are also happening elsewhere.

PR #5628, opened 2021-10-07
Fix segfault in tests on macOS

janezd:
##### Issue

Tests on macOS segfault in `Orange.widgets.evaluate.tests.test_owpredictions.TestComputeCD.test_compute_cd`

The tests calls `scoring.compute_CD`, which computes an innocent formula, and `scoring.graph_ranks`, which uses matplotlib to plot the graph. The latter looks like the obvious culprit.

##### Description of changes

For now, comment out the call.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


markotoplak:
The problem with segfaults is that the last shown test may not be the culprit... A package I know updated recently is catboost. Try installing/updating it and running tests locally.

PR #5637, opened 2021-10-08
[ENH] Feature Constructor: Evaluate categorical variables to strings

### LINKED ISSUE #5510: Scatter Plot Categorical Misplacement
**What's wrong?**

I have a set of data that goes from 27 -> 45 C, to extract this for a variable I used a "create class" widget to pull them from the file names. This creates a categorical. If the order of these categoricals is not directly sequential then when I plot them I get: 
image
Where both the colour and the axes are ostensibly pulling from the same data set yet display completely different results. And it's not like I can just use the "feature constructor" to generate a feature with the numerical representations as it just gives me some numbers between 1 and 12 instead that aren't the actual values I care about.

[Scatterplot.zip](https://github.com/biolab/orange3/files/6750328/Scatterplot.zip)

All you need to do is open and inspect, if you rewrite the order of the numbers in the class creation widget then it will work just fine. It seems like the axes are recognizing these are numbers and then placing them as they should along the axis, but that's not mirrored in the way the rest of the program treats the categorical variable. This means that when these are plotted the order of the original categoricals in a separate list is maintained but that is not informing the axis construction as it should. Ideally, I'd be able to have it automatically recognize the temperature as an integer value and sort according to the way the axis does.

**What's your environment?**

- Orange version: 3.30.0.dev
- How you installed Orange: command line

janezd:
##### Issue

Fixes #5510. That is, not the original issue but @markotoplak's comment that Feature Constructor should be modified to be more user-friendly.

##### Description of changes

Values of categorical features are now evaluated to strings, not indices.

Indices were not very useful, also because they referred to indices in a list whose order was not necessarily fixed, therefore this was unstable and dangerous. For this reason -- and because this behaviour wasn't documented anyway  (see https://orangedatamining.com/widget-catalog/data/featureconstructor/) -- the only effort towards backward compatibility in this PR is to remove `.value` from expressions. Plus, the change is noted in the tooltip and documentation.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -678,6 +678,7 @@ class OWFeatureConstructor(OWWidget):
         self.closeContext()
 
         self.data = data
+        self.expressions_with_values = False
 
         if self.data is not None:
             descriptors = list(self.descriptors)
--- a/Orange/widgets/data/tests/test_owfeatureconstructor.py
+++ b/Orange/widgets/data/tests/test_owfeatureconstructor.py
@@ -377,7 +377,6 @@ class OWFeatureConstructorTests(WidgetTest):
         self.widget.apply()
         self.assertTrue(self.widget.Error.more_values_needed.is_shown())
 
-
     @patch("Orange.widgets.data.owfeatureconstructor.QMessageBox")
     def test_fix_values(self, msgbox):
         w = self.widget
@@ -423,7 +422,7 @@ class OWFeatureConstructorTests(WidgetTest):
         domain = Domain(v)
         data = Table.from_numpy(domain, [[1, 3.14]])
 
-        settings = {
+        settings_w_discrete = {
             "context_settings":
             [Context(
                 attributes=dict(Ana=1, Cilka=2), metas={},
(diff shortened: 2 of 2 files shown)


janezd:
@lanzagar, I think you'll like this (if it works :).

An "Upgrade Expression" button now removes `.values`, while those without `.values` are replaced with a dictionary and the variable's value use as key. It's not fast, but should work.

Orange/widgets/data/owfeatureconstructor.py
```diff
@@ -330,6 +339,111 @@ def data(self, index, role=Qt.DisplayRole):
             return super().data(index, role)
 
 
+def freevars(exp, env):
```
janezd, line 342: This function has been moved above the class definition: it is called from `migrate_context`, which is called from meta class, which is sets up the class. Function below the class definition are not yet defined at that moment.

Orange/widgets/data/owfeatureconstructor.py
```diff
@@ -709,110 +832,61 @@ def send_report(self):
...
-    elif etype == ast.keyword:
-        return freevars(exp.value, env)
-    else:
-        raise ValueError(exp)
+        def fixer(mo):
+            var = domain[mo.group(2)]
+            if mo.group(3) == ".value":  # uses string; remove `.value`
+                return "".join(mo.group(1, 2, 4))
+            # Uses ints: get them by indexing
+            return "{" + mo.group(1) + \
```
lanzagar, line 855:
```suggestion
            return mo.group(1) + "{" + \
```

PR #5649, opened 2021-10-17
Revert "XGBoost: Temporary workarounds

janezd:
It reverses workaround from pull requests: #5653 and #5613


janezd:
Ha! XGBoost didn't fix it, apparently. :(

PR #5678, opened 2021-11-02
Table: extend aggregate module by preprocessing

mw25:
##### Issue

These are the modifications of OrangeTableGroupBy and Table.groupby() to allow preprocessing / range selection for each group, mentioned in [https://github.com/biolab/orange3/pull/5541#issuecomment-935617593](https://github.com/biolab/orange3/pull/5541#issuecomment-935617593) 

This allows preprocessing via scripting. Optionally, one could make a widget that is based on this.

##### Description of changes
The Table.groupby() functions gets a new parameter with which the preprocessing can be set.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/aggregate.py
+++ b/Orange/data/aggregate.py
@@ -20,6 +20,10 @@ class OrangeTableGroupBy:
     by
         Variable used for grouping. Resulting groups are defined with unique
         combinations of those values.
+    preprocess_settings
+        A DataFrame function to be applied to each group before aggregation.
+        Tuple of the function name, the parameter of the function, and the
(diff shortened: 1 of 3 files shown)


PrimozGodec:
Thank you for your contribution. 

I would prefer it if the implementation of aggregate remains as it is. Since aggregation is a two-step operation one can extend this module when needed and add preprocessing there. 

For example, if that kind of preprocessing is required in a time-series add-on, group-by can be extended there.

janezd:
We discussed this today at live meeting. We agreed with what @PrimozGodec wrote above.

PR #5680, opened 2021-11-03
[ENH] Use palette colors in more places

ales-erjavec:
##### Issue

Some visualizations use hardcoded colors and do not account for the effective color palette.

##### Description of changes

Replace use of hardcoded colors with palette colors.  

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -9,7 +9,7 @@ import numpy as np
 from AnyQt.QtCore import Qt, QRectF, QSize, QTimer, pyqtSignal as Signal, \
     QObject
 from AnyQt.QtGui import QColor, QPen, QBrush, QPainterPath, QTransform, \
-    QPainter
+    QPainter, QPalette
 from AnyQt.QtWidgets import QApplication, QToolTip, QGraphicsTextItem, \
     QGraphicsRectItem, QGraphicsItemGroup
 
@@ -1290,12 +1290,12 @@ class OWScatterPlotBase(gui.OWComponent, QObject):
         if self._too_many_labels or mask is None or not np.any(mask):
             return
 
-        black = pg.mkColor(0, 0, 0)
+        foreground = self.plot_widget.palette().color(QPalette.Text)
         labels = labels[mask]
         x = x[mask]
         y = y[mask]
         for label, xp, yp in zip(labels, x, y):
-            ti = TextItem(label, black)
+            ti = TextItem(label, foreground)
             ti.setPos(xp, yp)
             self.plot_widget.addItem(ti)
             self.labels.append(ti)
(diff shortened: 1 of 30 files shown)


janezd (approved):

ajdapretnar:
The only fault I can find here is the Correspondence Analysis suddenly showing no points at all. 🤷‍♀️



The rest is ok. I hope I have tested it well, I wasn't entirely sure what to check. 😅

Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -91,6 +91,31 @@ def clear(self):
...
+            color.setAlpha(100)
+            pen = QPen(color, 1)
+            pen.setCosmetic(True)
+            return pen
+
+    def brush(self):
+        if self.__brush is not None:
+            return QBrush(self.__brush)
+        else:
+            color = self.palette().color(QPalette.Window)
```
janezd, line 108: I think this is too dark, at least on my machine. I don't have a better suggestion, except, perhaps, `super().brush()` (which, for me, apparently uses a transparent background).

Screenshot 2021-11-19 at 14 04 23

ales-erjavec: Strange its supposed to be: 
Screenshot 2021-11-19 at 14 29 02

ajdapretnar: I think I can finally say: Janez, your Qt is a mess. 😆 
(i.e. My version is ok, too.)

janezd: My `QPalette.Window` is black, which it indeed shouldn't be (in light mode on macOS). Any ideas why?

```
>>> hex(QPalette().color(QPalette.Window).rgb())
'0xff000000'
```

ales-erjavec: Can you post the contents of `~/.config/biolab.si/Orange.ini`in particular `[application-style]` section.

janezd:
```
[application-style]
palette=
style-name=
```

janezd:
I tried in a new environment. It's still (transparent) black.

VesnaT:
image

On:
- macOS Catalina
- python 3.8.3
- PyQt5==5.12.3

janezd:
@ales-erjavec, I downloaded an Orange bundle, replaced the widgets directory with the files from this PR. The legend still has a black background. It looks as something between Qt and Catalina, not my installation of Qt.

janezd:
As suggested by @markotoplak, I tried creating a fresh user. Still black.

janezd:
Found it. Works with pyqtgraph 0.12 (the change that was made just to improve performance, it appears: https://github.com/pyqtgraph/pyqtgraph/pull/1569/commits/fd0e2b65afefdda2ddd08f9ee51a3914c0616b38).

@ales-erjavec, @markotoplak, can we increase the oldest supported version to 0.12?

@ales-erjavec, can you add it to this PR?

ales-erjavec:
> ... can we increase the oldest supported version to 0.12?

I have reset the palette instead so that wont be necessary.

Issue #5425, opened 2021-05-06
increase visibility of data subset in the point-based visualizations


BlazZupan:
In point-based visualisations, like t-SNE, it is very hard to see the data that comes in as a subset. Consider the following figure,

![image](https://user-images.githubusercontent.com/726604/117284448-1dd07f80-ae67-11eb-8c45-6b2b0f34aa24.png)

where a part of the data belongs to the data subset. Even though that cluster is well-defined, it is hard to spot it due to the density of the points around it. The problem becomes even more pronounced with larger data sets.

If the points that represent the selected data items sent to the widget are enlarged, though, they are clearly visible. For instance, following shows the same data as above:

![image](https://user-images.githubusercontent.com/726604/117291051-cf26e380-ae6e-11eb-8cb4-87bfaa40fe83.png)

I would propose a option (checkbox) that says "Enlarge subset instances", or something in this line. If this is implemented, it would be great to animate visualization change upon change in the input data subset, just like this is implemented when we the values in the column that dictates the size of the data points ("Size:").

janezd:
Option 1: Slider for contrast in transparency between subset and others.
Option 2: Transparency is also applied to pen of points not in the subset.

janezd:
Closed via #5684.

PR #5684, opened 2021-11-05
[ENH] Scatterplot: Use opacity for contrast

janezd:
##### Issue

Ref: #5425 

Instead of introducing a new slider, the opacity slider behaves differently in presence of a subset: it still regulates the opacity of the brush of subset, but at the same time it decreases the opacity of the pen of points not in the subset. The slider is still named opacity because I don't think the user will notice that (s)he now essentially regulates the contrast.

For now I implemented this for discrete coloring. If we decide that we like it, I'll change the continuous palette and monochromatic scatterplot in the same fashion.

Comments welcome.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -23,7 +23,6 @@ from Orange.util import utc_from_timestamp
 from Orange.widgets import gui
 from Orange.widgets.settings import Setting
 from Orange.widgets.utils import classdensity, colorpalettes
-from Orange.widgets.utils.plot import OWPalette
 from Orange.widgets.visualize.utils.customizableplot import Updater, \
     CommonParameterSetter
 from Orange.widgets.visualize.utils.plotutils import (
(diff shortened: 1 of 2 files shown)


janezd:
Alternative -- better, I think: decrease opacity for non-subset points faster than for the subset. This solution has a single slider which controls both, opacity and contrast at the same time. Two sliders would, I think, have to have the same effect as this single slider, but with more work from the user.

Screenshot 2021-11-06 at 18 00 43
That's what it looks like:
Screenshot 2021-11-06 at 17 58 49
Screenshot 2021-11-06 at 17 58 55
Screenshot 2021-11-06 at 17 59 06
Screenshot 2021-11-06 at 17 59 12
Screenshot 2021-11-06 at 17 59 18

janezd:
I'd appreciate a quick review (from the point of the user for now) before I update the failing tests. @ajdapretnar or @markotoplak, maybe?

ajdapretnar:
I'm happy to do it.

PR #5709, opened 2021-11-26
Box Plot: Treat nan's as 0's

### LINKED ISSUE #5679: Box Plot: crash on numeric target
**What's wrong?**

Box Plot crashes when given a numeric target and split by categorical variable.

```
  File "/Users/ajda/orange/orange3/Orange/widgets/visualize/owboxplot.py", line 485, in attr_changed
    self._variables_changed(self.apply_group_sorting)
  File "/Users/ajda/orange/orange3/Orange/widgets/visualize/owboxplot.py", line 489, in _variables_changed
    self.compute_box_data()
  File "/Users/ajda/orange/orange3/Orange/widgets/visualize/owboxplot.py", line 541, in compute_box_data
    self.conts = contingency.get_contingency(
  File "/Users/ajda/orange/orange3/Orange/statistics/contingency.py", line 308, in get_contingency
    return Continuous(
  File "/Users/ajda/orange/orange3/Orange/statistics/contingency.py", line 205, in __init__
    self.from_data(dat, col_variable, row_variable)
  File "/Users/ajda/orange/orange3/Orange/statistics/contingency.py", line 231, in from_data
    conts = data._compute_contingency([col_variable], row_variable)
  File "/Users/ajda/orange/orange3/Orange/data/table.py", line 2065, in _compute_contingency
    col_data, W_, classes_ = arr[:, arr_i], W_, row_data
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```

File (heart-disease). Set Age as target (remove Diameter Narrowing as target!). Box Plot (split by gender, select Age).

**What's your environment?**

- Operating system: OSX
- Orange version: 3.31.dev

### LINKED ISSUE #5699: Box Plot: sparse variables are ranked wrong
**What's wrong?**

When analyzing topic distinctions with Box Plot, one would select a categorical variable to group by and use Order by relevance to subgroups to rank the variables by how well they split between categorical values.

In text mining it can happen (but I guess elsewhere too), that there is only one document in a specified group, so Box Plot cannot compute significance, which is fine.

The bug is that this variable is placed before one where significance can be computed. Is this expected?
A bunch of variables where all instances belong to one group are also placed very high. I, of course, expect to see these variables in a list, but probably there could be a checkbox (gosh, another one?) which would hide such variables? Or do we need a separate widget for text mining?

Corpus (book-excerpts) - Preprocess Text (default) - Bag of Words (uncheck Hide bow attributes) - Box Plot. Use Category as Subgroup, then observe the ranked order of variables.

**What's your environment?**

- Operating system: OSX
- Orange version: 3.30.2dev

janezd:
##### Issue

Fixes #5699. Also fixes #5679.

##### Description of changes

- Avoid computing the continuous contingency matrices, which may change the behaviour of the widget.
- #5679 is fixed as a side-effect of not computing contingency matrices for numeric attributes.
- Prepare the data for computation of statistics and showing it in the plot in the same way. There are still some difference in p-values at the third decimal.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -83,23 +83,21 @@ DiscDataRange = namedtuple("DiscDataRange", ["value", "group_value"])
 
 
 class BoxData:
-    def __init__(self, dist, group_val=None):
-        self.dist = dist
-        self.n = n = np.sum(dist[1])
-        if n == 0:
+    def __init__(self, col, group_val=None):
+        self.n = len(col) - np.sum(np.isnan(col))
+        if self.n == 0:
             return
-        self.a_min = float(dist[0, 0])
-        self.a_max = float(dist[0, -1])
-        self.mean = float(np.sum(dist[0] * dist[1]) / n)
-        self.var = float(np.sum(dist[1] * (dist[0] - self.mean) ** 2) / n)
-        self.dev = math.sqrt(self.var)
-        a, freq = np.asarray(dist)
-        q25, median, q75 = _quantiles(a, freq, [0.25, 0.5, 0.75])
-        self.median = median
-        # The code below omits the q25 or q75 in the plot when they are None
-        self.q25 = None if q25 == median else q25
-        self.q75 = None if q75 == median else q75
-        self.data_range = ContDataRange(q25, q75, group_val)
+        self.a_min = np.nanmin(col)
+        self.a_max = np.nanmax(col)
+        self.mean = np.nanmean(col)
+        self.var = np.nanvar(col)
+        self.dev = np.sqrt(self.var)
+        self.q25, self.median, self.q75 = np.nanquantile(col, [0.25, 0.5, 0.75])
+        self.data_range = ContDataRange(self.q25, self.q75, group_val)
+        if self.q25 == self.median:
+            self.q25 = None
+        if self.q75 == self.median:
+            self.q75 = None
 
 
 class FilterGraphicsRectItem(QGraphicsRectItem):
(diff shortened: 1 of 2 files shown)


markotoplak:
Why would sparse arrays even contain nans (unless they are explicit) - they should just have unspecified values. Or do we get nans somewhere in computation?

markotoplak:
Agh, I see it, it comes from here: `col = data.get_column_view(attr)[0].astype(float)`. That is so wrong!

markotoplak:
Perhaps the actual NaNs could be identified before conversion and then, after replacing NaN with 0, set back? I would not want to lose original nans (sometimes that have meaning).

janezd:
@markotoplak, thanks! You're right: missing sparse data is not nan but 0, and there's no need for any special flag.

To speed up the widget, Box plot could internally store sparse data as csc, and then simply use a corresponding slice within `data` in the same way it currently uses `get_column_view` for dense. Right?

Orange/widgets/visualize/owboxplot.py
```diff
@@ -46,60 +46,27 @@ def compute_scale(min_, max_):
     return first_val, step
 
 
-def _quantiles(a, freq, q, interpolation="midpoint"):
```
janezd, line 49: This was needed to compute quartiles from distributions. Now quartiles are computed from data, using `np.nanquartiles`.

Orange/widgets/visualize/tests/test_owboxplot.py
```diff
@@ -281,8 +280,8 @@ def _select_data(self):
                  if isinstance(item, FilterGraphicsRectItem)]
         items[0].setSelected(True)
         return [100, 103, 104, 108, 110, 111, 112, 115, 116,
-                120, 123, 124, 126, 128, 132, 133, 136, 137,
-                139, 140, 141, 143, 144, 145, 146, 147, 148]
+                120, 123, 124, 128, 132, 133, 136, 137, 139,
+                140, 141, 143, 144, 145, 146, 147]
```
janezd, line 284: A slight change because the `np.nanquartiles` does not put the threshold at the same place as `_quantiles`.

ajdapretnar:
Unchecking "Order by relevance to subgroups" is now extremely slow. I suppose this checkbox would now affect only the list of variable? And not involve any specific computation?
Selecting None in Subgroups is also slow.

janezd:
Wasn't it already slow before? Is it (even) worse now?

The checkbox involves computation of statistics for ordering.

ajdapretnar:
Yes, you a right, it was just as bad before. But what I fail to understand why. I have already selected the Subgroup and I see the test statistics. I expect the checkbox to just re-order variables. What am I missing?

janezd:
But you see it just for one variable. When you check the box, this statistics has to be computed for all variables.

ajdapretnar:
Ok, fair enough. I will merge then, the fix is real nice (finally the results make sense!).

PR #5710, opened 2021-11-26
[ENH] Learner widgets: Inform about potential problems when overriding preprocessors

### LINKED ISSUE #5577: OWBaseLearner: mildly broken handling of signals
**What's wrong?**

The widget has two inputs and two outputs.

- `OWBaseLearner.set_preprocessor` calls `self.apply()`, which in turn calls `update_learner()` and `update_model()` that construct and (unconditionally) send the learner and the model.

- `OWBaseLearner.set_data` calls `update_model()`, which updates the model and sends it. Learner is not updated because the data signal does not affect it.

`set_preprocessor` should call unconditional apply, because this is what signal handlers should do ... except that when the widget has multiple inputs affecting the same outputs, apply-like method should be called from `handleNewSignals` to avoid sending outputs before all inputs are present.

I believe this was done to avoid sending a learner when the widget receives just new data.

Similarly, when changing the name of the model, the widget uses

```python
            if self.auto_apply:
                output.send(instance)
```

This is wrong because it does not set the dirty flag; if the user re-enables auto commit, the change won't be propagated.

1. Connect some learner to Test and Score
2. Disable auto apply in the learner
3. Rename the learner
4. Enable auto apply

Test and Score still shows the old name.

But this problem is insignificant. I am reporting this because I prefer that signals are handled properly if possible.

**Proposed solution**

The widget should hold a (weak?) reference to the most recently sent instance of learner and model. Before sending it anew, it should check that the instance indeed changed. With this, we could implement `handleNewSignals` as usual.

**Caveat:** if any derived class overrides `handleNewSignals`, we should ensure that it calls the inherited method. We have a bad habit of not doing it (often for good reasons, though).

I'm skipping this widget in #5495 -- and with it the entire category Model, which is derived from it. Widgets will work despite https://github.com/biolab/orange-widget-base/pull/157, but give warnings, which is desired.

**What's your environment?**

N/A.

### LINKED ISSUE #5703: Learner widgets fail (without warning) without proper preprocessing
**What's wrong?**

When using "Select Relevant Features" preprocessor with a learner, the learner fails with error "Fitting failed. Wrapped scikit-learn methods do not support multinomial variables". Note that when the same preprocessor is given the same data it works, that is, it returns a data set with reduced number of features.

image

image

See the workflow above. I have used the heart_dissease data set.

[preprocessor-bug.ows.zip](https://github.com/biolab/orange3/files/7591004/preprocessor-bug.ows.zip)

**What's your environment?**
- Operating system: mac os x
- Orange version: head (from github)
- How you installed Orange: pip install -e .

**Side note**

The workflow with Test & Score and preprocessed learner fails as well.

image

but when preprocessor is given to the Test & Score, it works. 

image

janezd:
##### Issue

Resolves #5703. Fixes #5577.

##### Description of changes

- Inform when preprocessors are given and the learner has its own default preprocessors, which are thusly overriden.
- Fix invalidation of model and learner, calling of updates and outputting signals.

Missing coverage is mostly in stack learner and was already missed before. This PR won't add tests there.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/model/owstack.py
+++ b/Orange/widgets/model/owstack.py
@@ -33,22 +33,26 @@ class OWStackedLearner(OWBaseLearner):
     @Inputs.learners
     def set_learner(self, index: int, learner: Learner):
         self.learners[index] = learner
+        self._invalidate()
 
     @Inputs.learners.insert
     def insert_learner(self, index, learner):
         self.learners.insert(index, learner)
+        self._invalidate()
 
     @Inputs.learners.remove
     def remove_learner(self, index):
         self.learners.pop(index)
+        self._invalidate()
 
     @Inputs.aggregate
     def set_aggregate(self, aggregate):
         self.aggregate = aggregate
+        self._invalidate()
 
-    def handleNewSignals(self):
-        super().handleNewSignals()
-        self.apply()
+    def _invalidate(self):
+        self.learner = self.model = None
+        # ... and handleNewSignals will do the rest
 
     def create_learner(self):
         if not self.learners:
--- a/Orange/widgets/utils/owlearnerwidget.py
+++ b/Orange/widgets/utils/owlearnerwidget.py
@@ -176,12 +176,19 @@ class OWBaseLearner(OWWidget, metaclass=OWBaseLearnerMeta, openclass=True):
         # invalidate the model so that handleNewSignals will update it
         self.model = None
 
-    def apply(self, *, hard=True):
+    SEND, SOFT, UPDATE = range(3)
+
+    def apply(self, *, level=UPDATE):
         """Applies learner and sends new model."""
-        if hard or self.learner is None:
-            self.update_learner()
-        if hard or self.model is None:
+        if level == self.SEND:
+            self._send_learner()
+            self._send_model()
+        elif level == self.UPDATE:
+            self.update_learnel()
             self.update_model()
+        else:
+            self.learner or self.update_learner()
+            self.model or self.update_model()
 
     def update_learner(self):
         self.learner = self.create_learner()
@@ -189,12 +196,15 @@ class OWBaseLearner(OWWidget, metaclass=OWBaseLearnerMeta, openclass=True):
             self.learner.use_default_preprocessors = True
         if self.learner is not None:
             self.learner.name = self.effective_learner_name()
+        self._send_learner()
+
+    def _send_learner(self):
         self.Outputs.learner.send(self.learner)
         self.outdated_settings = False
         self.Warning.outdated_learner.clear()
 
     def handleNewSignals(self):
-        self.unconditional_apply(hard=False)
+        self.unconditional_apply(level=self.SOFT)
         self.Information.ignored_preprocessors(
             shown=not getattr(self.learner, "use_default_preprocessors", False)
                   and getattr(self.LEARNER, "preprocessors", False)
@@ -216,6 +226,9 @@ class OWBaseLearner(OWWidget, metaclass=OWBaseLearnerMeta, openclass=True):
             else:
                 self.model.name = self.learner_name or self.captionTitle
                 self.model.instances = self.data
+        self._send_model()
+
+    def _send_model(self):
         self.Outputs.model.send(self.model)
 
     def check_data(self):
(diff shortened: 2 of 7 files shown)


markotoplak:
I think the message is not correct for current Orange's behavior. I like what the message implies, so perhaps we should change what Orange does.

Look, from the edited file:

```
    def update_learner(self):
        self.learner = self.create_learner()
        if self.learner and issubclass(self.LEARNER, Fitter):
            self.learner.use_default_preprocessors = True
```

`create_learner()` passes the given preprocessors into the learner, but in `Learner` we have:

```
    @property
    def active_preprocessors(self):
        yield from self.preprocessors
        if (self.use_default_preprocessors and
```

If I understand this code, the default preprocessors will still be used for `Fitter`s, the explicitly input ones are only meant to be used before the fixed ones. 

Logistic regression in Orange is not a `Fitter` so in this specific case the message is correct, but is wrong for `Fitter`s.

Why are `Fitters` (learners that can do both classification or regression) handled differently? No idea.

markotoplak:
Perhaps we can make a different message for `Fitter`s for now and then think about changing the behavior separately?

I have no idea what would that kind of a behavior change mean for backward compatibility, but it could get nasty. :D

Orange/widgets/model/owstack.py
```diff
@@ -33,22 +33,26 @@ def add_main_layout(self):
...
+        self.invalidate()
 
     @Inputs.aggregate
     def set_aggregate(self, aggregate):
         self.aggregate = aggregate
+        self.invalidate()
 
-    def handleNewSignals(self):
-        super().handleNewSignals()
-        self.apply()
```
janezd, line 51: This should have called `unconditional_apply`. Now the base class defines `handleNewSignals` and handles them properly (I suppose).

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -164,12 +173,15 @@ def set_data(self, data):
                     "Select one with the Select Columns widget.")
             self.data = None
 
-        self.update_model()
+        # invalidate the model so that handleNewSignals will update it
+        self.model = None
 
-    def apply(self):
+    def apply(self, *, hard=True):
```
janezd, line 179: Updating only when `None` (i.e. no hard update) would be better, but it would break compatibility with derived widgets counting on it.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -226,8 +245,7 @@ def settings_changed(self, *args, **kwargs):
     def _change_name(self, instance, output):
         if instance:
             instance.name = self.effective_learner_name()
-            if self.auto_apply:
-                output.send(instance)
```
janezd, line 230: This was incorrect because it didn't set dirty flag, but was necessary due to broken signal handling.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -164,23 +176,50 @@ def set_data(self, data):
...
+            self._send_model()
+        elif level == self._UPDATE:
+            self.update_learner()
+            self.update_model()
+        else:
+            self.learner or self.update_learner()
+            self.model or self.update_model()
+
+    def apply_as(self, level, unconditional=False):
+        self.__apply_level.append(level)
```
janezd, line 197: This is horrendous, but I see no way around it (while maintaining compatibility).

This class and derived classes used `apply` to update and send both models; it would be better to update just the invalidated. Furthermore `_change_name` sent outputs by itself if `auto_apply` was `True`, but of course could not set the dirty flag if it was `False`. In some cases, this caused outputing the model twice...

To clean the mess, yet keep the compatilbility, I had a solution in which `apply` got an additional argument with default value. Unfortunatelly, `gui.auto_apply` doesn't pass extra arguments to `apply` -- and we shouldn't modify it to, just to fix this specific case.

To make it even more tricky, the function as defined here is called only when auto apply is on.

Hence, `apply_as` passes an argument by appending the "urgency" to the list. When the apply function is actually called, it finds the maximal "urgency" in the list and acts on it.

Orange/widgets/utils/tests/test_owlearnerwidget.py
```diff
@@ -218,3 +218,73 @@ def check_name(name):
...
+
+        self.send_signal(wno.Inputs.preprocessor, None)
+        self.send_signal(wyes.Inputs.preprocessor, None)
+        self.send_signal(wfit.Inputs.preprocessor, None)
+
+        self.assertFalse(wno.Information.ignored_preprocessors.is_shown())
+        self.assertFalse(wyes.Information.ignored_preprocessors.is_shown())
+        self.assertFalse(wfit.Information.ignored_preprocessors.is_shown())
+
+    def test_multiple_sends(self):
```
janezd, line 274: This test would fail on master because learner would be sent twice.

Issue #5370, opened 2021-03-29
[RFC] Move some widgets from Data, split Data category


markotoplak:
@irgolic just showed me a prototype reorganization for the Data category. Looks promising! But looking at it I wondered why are some widgets even there. I'd suggest:

- Correlations -> Unsupervised
- Neighbours -> Unsupervised, works on distances.
- Outliers -> Unsupervised. I know it can be used to preprocess the data, but used heavy ML machinery and as such does not belong among the basic data-handling widgets.
- Rank -> Model. Not exactly basic data handling; has a pretty specific use. We mostly use it with a class variable, either to score features according to the Model (perfect for the category) or their connection to the class. Sidenote: this widget should maybe split into two widgets, both of which would still be a better fit for the Model category (one doing Gain Ratio, Relief & similar, and perhaps another, "Rank by Model"; stupid name, I know).
- Feature Statistics -> Visualize. Here I have doubts, but isn't this some kind of a data view? It could fit nicely with data Box Plot or Distributions. I assume it is here because we do not have a "Statistics" category...

Could anyone perhaps explain why were these widgets in the Data category in the first place? Could we move any of them?

markotoplak:
Also, I proposed the following heuristic: does the widget use any kind of statistics other than an ordinary high schooler would recognize? If so, not "Data"! :)

ajdapretnar:
> Could anyone perhaps explain why were these widgets in the Data category in the first place? Could we move any of them?

Because it was convenient, I suppose. 😆 

I agree with moving the widgets. One has to be careful to implement remapping for older versions of workflows, though. What about the docs?

borondics:
I think a bit of reorganization is a great idea, I was wondering about this a while ago.

Also, it could be nice if packages, like Spectroscopy, could register widgets into other than their panel. Meaning, if a package has a file reader that would belong to the Data panel it could actually show up there. We had this issue come up recently during a workshop. Maybe this requires some major code changes though.

ajdapretnar:
@borondics I am not entirely against having "data" widgets appear in Data pane, but I would assume such a reader is specific for the domain and would thus probably best fit into <domain> add-on. I would say Data widgets should be very general, while add-on "data" widgets can be specific. A general reader can thus be included easily in the core distribution (is it really easy to add general widget to the core distribution by an external developer?).
But again, I am not against it. Just my 2c. :)

markotoplak:
[...]

PR #5712, opened 2021-11-29
[RFC] Move some widgets from Data to Transform and Unsupervised category

ales-erjavec:
##### Issue

Resolve: gh-5370

(requires https://github.com/biolab/orange-widget-base/pull/179)

##### Description of changes

Move most widgets from Data to Transform or Unsupervised category

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/__init__.py
+++ b/Orange/widgets/__init__.py
@@ -6,11 +6,14 @@ import sysconfig
 
 import pkg_resources
 
-import Orange
+from orangecanvas.registry import CategoryDescription
+from orangecanvas.registry.utils import category_from_package_globals
+import orangewidget.workflow.discovery
 
 
 # Entry point for main Orange categories/widgets discovery
 def widget_discovery(discovery):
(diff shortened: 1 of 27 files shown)


janezd:
I think this is OK to see whether we like this change or not.

But I think that widget discovery should be fixed anyway. If widgets are allowed to specify the category, the registry should still observe category definition from the corresponding `__init__`. The current behaviour, as far as I see, is to reject it (unless it's seen before the widget). I'd propose updating the category definition if the existing definition came from the widget (and is thus mostly blank). Or I've perhaps missed something?

janezd:
When we tried to see this in the morning, I apparently wasn't on orange-widget-base master. With @markotoplak we now said that 

- after releasing Orange 3.31, we provisionally merge this, so we can
- work with the newly split category and see whether we like it. If we do
- we move the files to a new directory (before I realized why this PR doesn't work, I already create a branch with such change),
- fix tests and any other problems,
- and replace the temporary icon.

PR #5714, opened 2021-11-30
Table lock: fix unlocking unpickled tables

markotoplak:
##### Issue

@VesnaT found a pickled table that could not be unlocked, because it was a view when unpickled. I thought this can not happen, but here it is.

While writing tests I also notices that EMPTY `W` from pickled tables could not be unlocked, because when unpickled it was `W.flags.owndata == False`. 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -505,6 +505,9 @@ class Table(Sequence, Storage):
 
         def sync(*xs):
             for x in xs:
+                # no need to make empty arrays writable, as nothing can get written
+                if writeable and x.size == 0:
+                    continue
                 try:
                     undo_on_fail.append((x, x.flags.writeable))
                     x.flags.writeable = writeable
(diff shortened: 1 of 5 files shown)


markotoplak:
The last commit is quite obnoxious. Some arrays have `.base == None` and `.flags.owndata = False`. This happens after unpickling arrays with 0 size; in `Orange.data.Table`, these are the arrays with shape (rows, 0). We use these arrays if there are no columns for a particular part of the `Table`.

For now, numpy allows unlocking (setting `writable = True`) for these arrays, but shows a deprecation warning. To avoid the warning, the last commit disables unlocking of `size=0` tables. The change requires that a bunch of code that was previously just settings something to empty slices is guarded with `if data.X.size`style-conditions. This is uglier than `if data.X is not None` that we were probably trying to avoid by having 0-size arrays. But surprisingly we do not use the "feature" that often...

And ideas for alternative solutions?

I was considering disabling unlocking only if `flags.owndata == False`, but then we would be just moving the problems onto unpickled arrays, where they will be harder to catch. So no to this one.

janezd:
This is so annoying. But it looks like the most straightforward way to fix it.

PR #5736, opened 2021-12-13
[ENH] File: explicit file format choice

### LINKED ISSUE #5724: File widget: make file readers expicit
**What's your use case?**

File loading is a bit frustrating, specifically for the extensions where the content can be of multiple formats and the file is not actively analyzed to determine which specific loader will try to load the file.

Let's say we are trying to load a TXT file which is not of the default TXT loader format. The user has to select the reader in the drop-down menu of the `...` dialog (I guess this window is the OS specific file browser). Then, if the reader is correct, the file gets loaded and we return to the File widget main window. However, if the user wants to load another TXT file with the previous structure, which is reasonable to expect, they have to select again the specific loader, the `...` dialog doesn't remember the previous choice (although the File widget does save this parameter and if the user simply clicks between the different items in the drop-down menu changing back and forth works).

It would be nice to set the file format to the last used in the `...` file selector dialog.

- Allow the file reader to analyze the structure of the file and select the appropriate reader based on the content. This could be fairly fast as the first few lines/bytes usually unambiguously determine the format.
- Try all available readers for that particular extension and use the one that doesn't fail. This might be slow?

markotoplak:
##### Issue
Resolves #5724

##### Description of changes

image

For files, the reader used was already stored in `.file_format`, which could be either a reader's qualified name or `None` (which means that reader is detected depending on extension and priority).

All available readers are shown in a combo box.

For URLs, the reader selection is disabled. That part would have to be refactored a bit so that the chosen reader could be saved.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io_base.py
+++ b/Orange/data/io_base.py
@@ -31,6 +31,11 @@ __all__ = ["FileFormatBase", "Flags", "DataTableMixin", "PICKLE_PROTOCOL"]
 PICKLE_PROTOCOL = 4
 
 
+class MissingReaderException(IOError):
+    # subclasses IOError for backward compatibility
+    pass
+
+
 class Flags:
     """Parser for column flags (i.e. third header row)"""
     DELIMITER = ' '
@@ -551,7 +556,7 @@ class _FileReader:
             if fnmatch(path.basename(filename), '*' + ext):
                 return reader(filename)
 
-        raise IOError('No readers for file "{}"'.format(filename))
+        raise MissingReaderException('No readers for file "{}"'.format(filename))
 
     @classmethod
     def set_table_metadata(cls, filename, table):
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -32,6 +32,8 @@ from Orange.widgets.widget import Output, Msg
 # module's namespace so that old saved settings still work
 from Orange.widgets.utils.filedialogs import RecentPath
 
+DEFAULT_READER_TEXT = "Automatically detect type"
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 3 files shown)


borondics:
I think this is great, but I have two comments:
The error message if the selected reader is not the right one is not too informative:
image

Then, the widget forgets the selected reader for subsequent files. I think it would be nice to save this in the widget settings because one could expect that the same user works with the same type of TXT files. The best solution here would be to first try to autodetect and then fall back to the last selected for that file extension in my opinion. Might be difficult to store all this data though...

markotoplak:
@borondics, I do not want to do any application of the currently selected reader onto a new file into this PR. Added automatics like this can lead to confusion (remember disappearing settings is Select Columns or Feature Constructor) so I would not add it unless absolutely sure. Furthermore, your proposed change would change the current behavior, while this PR aims to keep it exactly as it was, it only introduces the possibility of explicit modifications.

The message is very precise (= not useful for non-developers), just click/mouseover on it. :)

borondics:
OK, I understand that this PR is aimed only for the new placement of the file loader selector. I hope though that in another PR the saving will be also possible.

For the error: yeah, I know that there is a longer message on moouse-over but something for the normal user wouldn't hurt either. I guess there are more users than developers, so it is not absurd to also let them clearly know what is happening. :)

markotoplak:
Pylint has a fun complaint: there are too many public methods (=tests) in the test class. :)

ajdapretnar:
I checked and this PR breaks nothing in Text. 👍 

A small comment. It would be nice(r) if add-on readers would be listed separately. I.e.
.tab
.csv
.pkl
.basket (is this even core?)
-----
.HDF5
.dat
.gsf....

So readers from each add-on would be in a separate "category".

VesnaT:
If I try to open the Iris dataset with an Excel reader, I get a wrong error message:

image

markotoplak:
I tried to address comments by @ajdapretnar and @VesnaT: error reporting is now fixed and the list is sorted per package name.

PR #5750, opened 2021-12-21
Remove deprecated UTC offsets

markotoplak:


markotoplak:
I did not know what to do for "uncomment new implementation for setting timezones in parse method" from the instructions.

@PrimozGodec, could you check this PR?

PrimozGodec:
> I did not know what to do for "uncomment new implementation for setting timezones in parse method" from the instructions.

It is something that I probably left in tests but is not in code anymore.

PR #5752, opened 2021-12-22
Fix some deprecation warnings

markotoplak:
- Deprecated `normalize` sklearn's parameter (leaving it on default value disables the warning)
- Use `squared_error` for criterions where appropriate (and equivalent)
- Fixed some np.float/np.int complaints


janezd:
We seldom specify `dtype`, but when we do, we use `int` and `float`, not `np.float64` and `np.int64`. I don't object to using the latter (particularly in light of the fact that in one operating system (the one that can't properly handle non-ascii letters) `float` resolves to `float32`, not `float64), but I think we should use the same type everywhere.

markotoplak:
@janezd, I changed these to `float` and `int` except for a single test where a specific numpy type seems to make sense.

PR #5760, opened 2022-01-04
[FIX] Support numpy 1.22

markotoplak:
- np.frompyfunc does convert outputs to the out type anymore
- Adapt a locking test for changed numpy internals

### DIFF
--- a/Orange/data/io_util.py
+++ b/Orange/data/io_util.py
@@ -109,7 +109,7 @@ def isnastr(arr, out=None):
     arr = np.asarray(arr)
     if out is None and arr.shape != ():
         out = np.empty_like(arr, dtype=bool)
-    return __isnastr(arr, out=out)
+    return __isnastr(arr, out=out, casting="unsafe")
 
 
 def guess_data_type(orig_values, namask=None):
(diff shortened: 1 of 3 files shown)


markotoplak:
@ales-erjavec, could you please review this?

It seems that `np.frompyfunc` in numpy 1.22 does not do the conversion to the type of its `out=` argument anymore. Thus, the modified `isnastr` will always create an intermediate array. I do not see a clear way to avoid this.

Orange/data/io_util.py
```diff
@@ -107,9 +107,14 @@ def isnastr(arr, out=None):
...
-        out = np.empty_like(arr, dtype=bool)
-    return __isnastr(arr, out=out)
+    na = __isnastr(arr)
+    if out is not None:
+        out[:] = na
+        return out
+    elif arr.shape != ():  # na is an object array
+        return na.astype(bool)
+    else:  # na is not a numpy array
+        return na
```
ales-erjavec, line 117: Maybe just add `casting="unsafe"` to `__isnastr` call? Optionally even check `np.can_cast('bool', out.dtype)`

markotoplak:
@ales-erjavec, thanks, I did not know about that ufunc argument. I changed added casting="unsafe" to all such calls, so this PR is now very simple.

Issue #5738, opened 2021-12-14
Group By: uninformative error message


ajdapretnar:
**What's wrong?**

Group By gives an uninformative error message when asking for a mode aggregation of an empty (all nan) variable.

<img width="902" alt="Screen Shot 2021-12-14 at 12 04 02" src="https://user-images.githubusercontent.com/12524972/145987962-c0b86d9e-752b-47f0-a73a-f9946878bd21.png">

**How can we reproduce the problem?** 

Get any data set with an all nan variable and try to use Mode in Group by for that variable.

**What's your environment?**

- Operating system: OSX
- Orange version: 3.30.2
- How you installed Orange: pip

PR #5763, opened 2022-01-05
[FIX] Group by: compute mode when all values in group nan

PrimozGodec:
##### Issue

Fixes https://github.com/biolab/orange3/issues/5738

##### Description of changes
- Fix mode to work in cases when all values in the group are nan
- test case with only nan values in the group
- additionally: use gui.deferred for autocoomit

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
The reason why I haven't migrated `commit` in this widget is, obviously, that it was still not finished by the time I made that PR. :)

PR #5768, opened 2022-01-07
Curve Fit: Retain expression on data remove

VesnaT:
##### Issue

The Curve Fit widget needs data on the input to initialize the learner (i.e set the expression).
Refreshing the input data clears the expression (and parameters) which is undesirable in most cases.

##### Description of changes
- retain the expression when removing the template data
- enable the expression even when no data on the input

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owcurvefit.py
+++ b/Orange/widgets/model/owcurvefit.py
@@ -396,12 +396,6 @@ class OWCurveFit(OWBaseLearner):
     def set_data(self, data: Optional[Table]):
         self.Warning.data_missing(shown=not bool(data))
         super().set_data(data)
-        self.__clear()
-
-    def __clear(self):
-        self.expression = ""
-        self.__param_widget.clear_all()
-        self.__on_parameters_changed([])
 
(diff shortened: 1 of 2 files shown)


janezd:
@VesnaT, I've noticed this bug today when changing the behaviour of `handleNewSignals` in https://github.com/biolab/orange3/pull/5710/commits/e44754d3aa84e9d981c05218677c17a4ca1da120, when I at first thought I've broken the widget.

The solution you propose here is rather a quick fix. The proper fix would be to use contexts ... unless we think we shouldn't. An argument against using contexts would be that it's too difficult and usafe to guess the names of required variables, and that we on the other hand don't want to use the overly restrictive `PerfectDomainContextHandler`.

markotoplak:
@janezd, we missed you today. In our discussions with @lanzagar and @VesnaT, I advocated for contexts, but we decided not to use them because neither of us knows what exactly do they do when there are multiple good matches. :)

janezd:
> I advocated for contexts,

Oh, I'm sorry I wasn't there to witness this. :)

PR #5769, opened 2022-01-07
Color: Warn about unused variable definitions

### LINKED ISSUE #5742: Color file not correctly loaded?
**What's wrong?**

I may be misunderstanding how this is supposed to work, on the other hand I will argue that the help window for Color does not show the SAVE/LOAD/RESET buttons and accordingly does not explain them :-)  (So the help page, at least, would need to be updated.)

Anyway, in the attached workflow I have a widget Color in which I set the colours for Japan and United States, respectively.  I have pressed SAVE in order to create a file GroupedChartsColourMap.colors.  In another branch of the workflow I want to reuse these colour settings, so in the widget Color (1) I have pressed LOAD in order to read the saved colours.  However, this does not seem to affect the colour settings for Japan and United States in that widget, they retain their defaults.

My assumption was that the SAVE/LOAD pair allows me to reuse colour maps in different workflows, but this does not happen?

The workflow, data, and colors file are attached.  If you open the Color widget, you see the intended setting of country colours.  If you open Color (1), press LOAD and load the GroupedChartsColourMap.colors file, you can see that the colour settings in that widget are not affected.

**What's your environment?**

- Operating system: macOS 12.1
- Orange version: 3.30.2
- How you installed Orange: From disk image at https://orangedatamining.com/download/#macos

[ColourMap.zip](https://github.com/biolab/orange3/files/7735248/ColourMap.zip)

janezd:
##### Issue

Closes #5742.

The issue does not describe a bug, the mistake was on the user's side. But the user should be warned.

##### Description of changes

Add a warning when color definitions describe variables that do not appear in the data.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owcolor.py
+++ b/Orange/widgets/data/owcolor.py
@@ -690,6 +690,7 @@ class OWColor(widget.OWWidget):
 
         # First, construct all descriptions; assign later, after we know
         # there won't be exceptions due to invalid file format
+        unused_vars = []
         both_descs = []
         warnings = []
         for old_desc, repo, desc_type in (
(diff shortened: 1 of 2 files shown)


janezd:
Here's a file that reviewer can use for testing. Open in text editor and remove mock variable definitions if you wish to observe different messages.

[t.colors.zip](https://github.com/biolab/orange3/files/7830724/t.colors.zip)

PR #5770, opened 2022-01-08
Test and Score: Move target class and comparison stat above tables

### LINKED ISSUE #5764: Test and Score: move Target Class and Model Comparison closer the results they influence
Someone wrote me an email asking me why this happens to their data. I tried on `iris` and this is what I get:

Screenshot_2022-01-05_16-50-27

Is anyone aware of any theory why would this happen so regularly?

On `titanic`, I only see CA = Recall, Precision returns something else.

janezd:
##### Issue

Closes #5764.

##### Description of changes

Move combo for target class and combo for score for model comparison closer to what the influence.

##### Includes
- [X] Code changes
- [X] Tests


markotoplak:
I like how the widget is now structured. On my linux computer running dwm it looks like this on master.

master

And like this on this branch.
test-and-score

I already know that nameless boxes look funny on my machine, but the space between the boxes in main area seems excessive. But if it looks good on other platforms I am fine with it.

janezd:
I believe that on @lanzagar's machine it's also as ugly as on yours. On mine, though, it's
 
Screenshot 2022-01-13 at 22 12 58

vs.

Screenshot 2022-01-13 at 22 13 10

The difference is such that guess I'd add the spacing depending on the OS.

lanzagar:
I think the spacing on my system is fine now.
What bothers me is the size of the target dropdown - it is very short when you open an unconnected widget (dropdown is empty), but is not resized when data is connected resulting in unreadable contents.
(it is ok if you already have data connected when first opening the widget).
ts

lanzagar:
The target dropdown is now great.
The model comparison dropdown is still very small (can also be seen on fig above). I would not propose the same solution and set a fixed size (the widget would need to be quite big), but if it is possible it would be nice if it stretches up to a set maximum length (enough to fit "Coefficient of variation of the RMSE"). Not sure how easy that is to do though... (if not, we could go with something else).

PR #5781, opened 2022-01-13
[FIX] Update (and test) minimum dependencies

markotoplak:
##### Issue
Our dependency list is stale. Occasionally tests fail in my valid but old python environment.

##### Description of changes
I added testing with oldest dependencies.

There are still requirements without specified versions. I left them as they were, for now, for better clarity of the changes in this PR (and due to some laziness). These should get defined in another PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,6 +45,10 @@ jobs:
             python-version: 3.8
             tox_env: orange-latest
             name: Latest
+          - os: ubuntu-18.04
+            python-version: 3.7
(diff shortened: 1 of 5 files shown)


markotoplak:
@JakaKokosar, do you know of a better way of solving this problem?

markotoplak:
The last failed test is not connected to this PR.

JakaKokosar:
> @JakaKokosar, do you know of a better way for testing this?

This is how we do it in add-ons. LGTM :)

Issue #5723, opened 2021-12-06
File can not read Google Sheets


ajdapretnar:
**What's wrong?**

Google Drive apparently changed the API to retrieve Google Sheets. Now, one cannot just enter sharable link to load the data (Read error).

We should check the API and see whether we can fix this so that one can easily load the data from the link.

**How can we reproduce the problem?** 

Try to load any data from Google Sheets via link.

**What's your environment?**

- Operating system: OSX
- Orange version: 3.30.2dev
- How you installed Orange: pip

JakaKokosar:
The issue here is that since https://github.com/biolab/orange3/commit/62a7605ba4c08db89703838ec9e767cc400a7a1a we encode URLs, and google docs services don't like this.

Plain URL works fine: https://docs.google.com/spreadsheets/d/1PJhmH0X_N7hZn03le5J70E4IKL4LiqAZenlhg2kWWUU/edit#gid=0

Encoded URL cannot be found:
https://docs.google.com/spreadsheets/d/1PJhmH0X_N7hZn03le5J70E4IKL4LiqAZenlhg2kWWUU/edit%23gid%3D0

This seems to be specific to google services because Dropbox links and links to biolab servers work fine:
http://file.biolab.si/text-semantics/data/elektrotehniski-vestnik-clanki/detektiranje-utrdb-v-%C5%A1ahu-.txt

One solution is to do an additional check for 'docs.google.com' URLs and skip encoding entirely. Any other ideas?

fabform:
This is handy to find google sheets id
[find google sheet id](https://fabform.io/a/how-to-find-google-sheet-id)

lukasholovsky:
Hi, Lukas from AlgOps.com here.

This issue is probably not based on API because in older versions of Orange (3.27 respectively) it worked even though it didn't work in newer versions. A

@JakaKokosar also your plain URL doesn't work fine - I tried this several minutes ago and it returns the same error even though I am able to look at the Sheet in the browser.

Please can you speed up on this issue since a big portion of our customers can't use both AlgOps and Orange due to this?

Thanks!

markotoplak:
The bug is there from merging #5412. So this bug may stem from:
- `quote` being overzealous, at least in that URL part - I do not know the exact URL specification, but quote is written as a general function which does not distinguish between parts of the URL
- dropbox and google docs have a special functions `_trim_*` functions that in docs' case adds a sheet (maybe something else) - maybe quoting prevents it

[...]

PR #5782, opened 2022-01-14
fix UrlReader did not support URL like http://filename.txt?a=1&b=2 .

likier:
##### Description of changes
Fix `UrlReader` did not support URL like `http://filename.txt?a=1&b=2` , witch did quoted to `http://filename.txt%3Fa%3D1%26b%3D2`

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -406,7 +406,14 @@ class UrlReader(FileFormat):
         filename = filename.strip()
         if not urlparse(filename).scheme:
             filename = 'http://' + filename
-        filename = quote(filename, safe="/:")
+
+        # check ? avoid quote redundantly on URL like http://filename.txt?a=1&b=2.
+        mark = filename.find('?')
+        if mark >= 0:
+            filename = quote(filename[:mark], safe="/:") + filename[mark:]
+        else:
+            filename = quote(filename, safe="/:")
+
         super().__init__(filename)
 
     @staticmethod
--- a/Orange/tests/test_url_reader.py
+++ b/Orange/tests/test_url_reader.py
@@ -25,6 +25,15 @@ class TestUrlReader(unittest.TestCase):
                "vestnik-clanki/detektiranje-utrdb-v-šahu-.txt"
         self.assertRaises(OSError, UrlReader(path).read)
 
+    def test_base_url_with_parameters(self):
+        data = UrlReader("https://datasets.biolab.si/core/grades.xlsx?a=1&b=2").read()
+        self.assertEqual(16, len(data))
+
+    def test_special_characters_with_parameters(self):
+        path = "http://file.biolab.si/text-semantics/data/elektrotehniski-" \
+               "vestnik-clanki/detektiranje-utrdb-v-šahu-.txt?a=1&b=2"
+        self.assertRaises(OSError, UrlReader(path).read)
+
 
 if __name__ == "__main__":
     unittest.main()


markotoplak:
Thanks! 

I believe the particular issue, #5723, is a signal of some deeper issue, so the fix should likely be more general. While this may fix the particular issue with Google Sheets, are there any other possibilities where where running `quote` on a string may destroy the address? Judging by documentation of the `quote` functions, they are. Did you look into this also?

```
quote('abc def') -> 'abc%20def'

    Each part of a URL, e.g. the path info, the query, etc., has a
    different set of reserved characters that must be quoted. The
    quote function offers a cautious (not minimal) way to quote a
    string for most of these parts.

    RFC 3986 Uniform Resource Identifier (URI): Generic Syntax lists
    the following (un)reserved characters.

    unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
    reserved      = gen-delims / sub-delims
    gen-delims    = ":" / "/" / "?" / "#" / "[" / "]" / "@"
    sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="

    Each of the reserved characters is reserved in some component of a URL,
    but not necessarily in all of them.

    The quote function %-escapes all characters that are neither in the
    unreserved chars ("always safe") nor the additional chars set via the
    safe arg.
```

Also, could you explain what was your reasoning to handle `?` as an exception and not add to the list of safe characters to `quote`? 

Finally, we need better testing. #5412 was merged because is solved a particular problem, but I fail to understand tests towards its effect. We need to add a URL fails without `quote` but works with current version, just to ensure that we do not break anything additionally.

likier:
@markotoplak 
This problem starts with a download link from MinIO, witch like `http://127.0.0.1:9000/data/d321fa3c6aba/a.txt?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=kier%2F20220114%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20220114T103309Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=f7c67d3280b17255f15949755483e05f6d5d223a1f2aceeb3d2fec31e30ec511`.
Well, The a real reason for this issue and #5723 is that an already encoded URL should not be encoded with `quote` redundantly again. Once a URL is encoded twice, the `http://filename.txt?a=1&b=2` becames `http://filename.txt%3Fa%3D1%26b%3D2`, witch can't be parsed correctly anymore.
The issue #5412 raise because the URL was not well encoded, so that `quote` need to be done.
Sadly, we don't have a very direct way to tell whether a URL has been encoded. As well as the URL should be encoded well as soon as it exists.
Based on existing test cases and following URL components. The code split the path, witch may include special chars and `quote` `path` but leave the rest parts not `quote`.
```
def urlparse(url, scheme='', allow_fragments=True):
    """Parse a URL into 6 components:
    <scheme>://<netloc>/<path>;<params>?<query>#<fragment>
    Return a 6-tuple: (scheme, netloc, path, params, query, fragment).
    Note that we don't break the components up in smaller bits
    (e.g. netloc is a single string) and we don't expand % escapes."""
    url, scheme, _coerce_result = _coerce_args(url, scheme)
    splitresult = urlsplit(url, scheme, allow_fragments)
    scheme, netloc, url, query, fragment = splitresult
    if scheme in uses_params and ';' in url:
        url, params = _splitparams(url)
    else:
        params = ''
    result = ParseResult(scheme, netloc, url, params, query, fragment)
    return _coerce_result(result)
```
Maybe this is not a perfect solution then the `nteloc` or `path` includes  `delims` like %. This is maybe the least significant change to fix above problems.
I added a new tests on UrlReader. Those might help me explain why I did this.

Love & peace; ^_^

likier:
Hi~
I think I've got the whole picture of the problem. I consider making another ”perfect“ solution to solve this problem. Now this PR can be closed as well. I will give another PR as soon as possible. ^_^

PR #5785, opened 2022-01-14
Revert 5412 url reader special chars

ales-erjavec:
##### Issue

Fixes https://github.com/biolab/orange3/issues/5723

##### Description of changes

* Revert gh-5412
* Tolerantly but properly encode url from the user input.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfile.py
+++ b/Orange/widgets/data/owfile.py
@@ -470,6 +470,10 @@ class OWFile(widget.OWWidget, RecentPathsWComboMixin):
             return reader
         else:
             url = self.url_combo.currentText().strip()
+            # tolerant parse, also prevent parsing as local file
+            parsed = QUrl.fromUserInput(url, os.devnull)
+            if parsed.isValid():
+                url = bytes(parsed.toEncoded()).decode("ascii")
             return UrlReader(url)
 
     def _update_sheet_combo(self):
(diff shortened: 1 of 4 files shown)


Orange/widgets/data/owfile.py
```diff
@@ -470,6 +470,10 @@ def _get_reader(self) -> FileFormat:
             return reader
         else:
             url = self.url_combo.currentText().strip()
+            # tolerant parse, also prevent parsing as local file
+            parsed = QUrl.fromUserInput(url, os.devnull)
+            if parsed.isValid():
+                url = bytes(parsed.toEncoded()).decode("ascii")
```
VesnaT, line 476: The problem of this solution is, that the UrlReader itself still doesn't work, so the change would have to be made in every widget that uses it.

I'd rather the fix was made within the UrlReader.

ales-erjavec: No. UrlReader should not accept invalid urls and should not do anything that could break valid urls.

markotoplak: Yes, but valid at which level? Shouldn't we support URLs as a browser would? If there is some non-destructive encoding/decoding happening in the background, why not put it into `UrlReader`?

ales-erjavec: Fine.

PR #5786, opened 2022-01-14
Feature Statistics: Add padding between histogram bars and horizontal line

### LINKED ISSUE #5757: Feature Statistics should have more space at bottom
**What's wrong?**

This is really an æsthetic thing: As can be seen in the attached screen shot, the Feature Statistics widget shows the _x_-axis of the bottom values tightly pressed against the bottom of the window.  To me it gives the impression that the bars have been chopped off.  It would be better if there were at least a few pixels’ worth of space below the _x_-axis.

Screenshot 2021-12-29 at 11 55 35

**What's your environment?**

- Operating system: 12.1
- Orange version: 3.31.0
- How you installed Orange: From the download at https://orangedatamining.com/download/#macos

janezd:
Adds a few pixels between the bars and the horizontal line. Such Tufte-inspired histograms look better.

I agree with @kaimikael the this still does not resolve properly #5757.

Screenshot 2022-01-14 at 12 59 26

### DIFF
--- a/Orange/widgets/data/owfeaturestatistics.py
+++ b/Orange/widgets/data/owfeaturestatistics.py
@@ -562,6 +562,7 @@ class FeatureStatisticsTableModel(AbstractSortTableModel):
                             variable=attribute,
                             color_attribute=self.target_var,
                             border=(0, 0, 2, 0),
+                            bottom_padding=4,
                             border_color='#ccc',
                         )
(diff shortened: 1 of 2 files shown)


kaimikael:
I think I’d like a couple of pixels of space below the bottom horizontal line as well, now it sort of blends in with the window frame.

PR #5788, opened 2022-01-14
Fully support URL with query or fragment like `http://filename.txt?a=1&b=2#c=3`

### LINKED ISSUE #5723: File can not read Google Sheets

likier:
Fully support URL with query or fragment like `http://filename.txt?a=1&b=2#c=3`

##### Issue
Fix #5723

##### Description of changes
Use UTF-8 encoding URL with non-ascii characters, and preserve the role of special characters like ?#&=.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -406,7 +406,13 @@ class UrlReader(FileFormat):
         filename = filename.strip()
         if not urlparse(filename).scheme:
             filename = 'http://' + filename
-        filename = quote(filename, safe="/:")
+
+        # Fully support URL with query or fragment like http://filename.txt?a=1&b=2#c=3
+        def quote_byte(b):
+            return chr(b) if b < 0x80 else '%{:02X}'.format(b)
+
+        filename = ''.join(map(quote_byte, filename.encode("utf-8")))
+
         super().__init__(filename)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


markotoplak:
Thanks!

@ales-erjavec, isn't this solution conceptually similar to yours? Both your solution and this one look good to me.

This one (if correct, I haven't tested it) also allows passing utf8 strings directly to UrlReader, which @VesnaT seems to prefer. Also for me it seems nice if you could copy a link from the browser address bar and insert it directly into the code.

likier:
Well, i saw the Test/lint failing about f-string (since python 3.6). Considering Python compatibility, I didn't use f-string directly. @markotoplak It will be up to the more senior contributors to decide which features are more needed.
I still hope to solve this problem and bring benefits to the customers I recommend Orange.
By the way, we need to use the API as much as we need to use the UI. Just modifying the code in the UI part may not solve the API problem.
^_^

PR #5795, opened 2022-01-18
[FIX] table_from_frame: replace nan with String.Unknown for string variable

PrimozGodec:
##### Issue

1. When transforming pandas data frame to table string columns that contain nan values will 
keep them after the transformation but Orange uses an empty string for the unknown value in the StringVarable. 

   For example:
      df = pd.DataFrame(
           [["a", "b"], ["c", "d"], ["e", "f"], [np.nan, np.nan]],
       )

   will be transformed to the table with two string variables and nan values will be kept even `String.Unknow = ""`

2. When a column is recognized to be a string and has an object type, it still can contain some values that are not strings. Cast column to string.

##### Description of changes
Changed that nan values are transformed to `String.Unknown` for columns that will be transformed to the string variable and values are transformed to strings.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/pandas_compat.py
+++ b/Orange/data/pandas_compat.py
@@ -280,7 +280,12 @@ def vars_from_df(df, role=None, force_nominal=False):
                 raise ValueError("String variable must be in metas.")
             _role = Role.Meta
             var = StringVariable(str(column))
-            expr = lambda s, _: np.asarray(s, dtype=object)
+            expr = lambda s, _: np.asarray(
+                # to object so that fillna can replace with nans if Unknown in nan
+                # replace nan with object Unknown assure that all values are string
+                s.astype(object).fillna(StringVariable.Unknown).astype(str),
+                dtype=object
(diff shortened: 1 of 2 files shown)


markotoplak:
Could you better explain what is the reasoning here?

Does not Orange use the empty string OR nan? I'd lean towards using nan wherever possible. And this goes further from what I'd like to see. :)

PrimozGodec:
> Could you better explain what is the reasoning here?

I added a better explanation of the issue.

> Does not Orange use the empty string OR nan? I'd lean towards using nan wherever possible. And this goes further from what I'd like to see. :)

I am not sure if nan is supported by all widgets (at least it does not work in Orange3-text addon). I can start working on supporting it there. I changed my implementation that it uses StringVariable.Unknow instead of hardcoded "" such that it will still work when we decide to use nan instead of "".

PR #5800, opened 2022-01-21
Translation infrastructure

janezd:
Utilities for translating Orange

### DIFF
--- a/scripts/extract_msgs.py
+++ b/scripts/extract_msgs.py
@@ -1,11 +1,69 @@
-import tokenize
+import os
+
+import libcst as cst
+from libcst.metadata import PositionProvider
+
+class StringCollector(cst.CSTVisitor):
+    METADATA_DEPENDENCIES = (PositionProvider,)
+
+    def __init__(self, ranges):
+        self.module = None
+        self.function_stack = []
+        self.collection = []
+        self.ranges = ranges
+
+    @property
+    def outside(self):
+        if not self.function_stack:
+            return []
+        return self.function_stack[-1].body.children
+
+    def visit_Module(self, node):
+        self.module = node
+
+    def push_context(self, node):
+        blacklist = [
+            child.body[0].value
+            for child in node.body.children
+            if isinstance(child, cst.SimpleStatementLine)
+                and len(child.body) == 1
+                and isinstance(child.body[0], cst.Expr)
+                and isinstance(child.body[0].value, cst.SimpleString)]
+        self.function_stack.append((node, blacklist))
+
+    def pop_context(self):
+        self.function_stack.pop()
+
+    def blacklisted(self, node):
+        return self.function_stack and node in self.function_stack[-1][1]
+
+    def visit_ClassDef(self, node: cst.ClassDef):
+        self.push_context(node)
+
+    def leave_ClassDef(self, _):
+        self.pop_context()
+
+    def visit_FunctionDef(self, node: cst.FunctionDef):
+        self.push_context(node)
+
+    def leave_FunctionDef(self, _):
+        self.pop_context()
+
+    def visit_FormattedString(self, node: cst.FormattedStringExpression):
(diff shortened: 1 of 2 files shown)


janezd:
Running `python setup.py extract_messages -o x.pot` now extracts all string tokens into `x.pot`.

To use ASTs instead of tokens, just modify function `extract`. @markotoplak, enjoy if you will.

- The first obvious improvement should be to ignore docstrings. (Much simpler with AST.)
- Function name is currently also omitted. (Trivial with AST, much much more difficult with tokens.)
- Translators comments should be included. (Probably difficult with AST.)
- We should probably include only `ow*.py` files, hopefully by specifying `[("ow**.py", "scripts.extract_msgs:extract", None), ("**.py", "ignore", None)]` as extractors.

We will need at least two more utilities:
- A utility that replaces strings in the source file with translations found in .po files.
- Translator should be able to mark code-related strings (such as names of settings and string literals used as constants) as "won't translate". A utility should be made to filter out such strings from the pot file, so the translator does not see them next time.

janezd:
Worth looking into: https://pybowler.io/docs/basics-intro. This project is based on 2to3.py and provide '**a grammar-based search "pattern" for finding nodes in the CST as well as methods for transforming nodes that match the pattern.**'

janezd:
Update, TL;DR:
- Utilities that extract messages into YAML and that replace translated strings in Python sources are ready. 
- We need to test it to see whether it works in principle.
- We need a few additional utilities, in particular one that prepares a YAML with untranslated strings and one that merges new translations into existing ones. Plus, it must allow translator to make a string as wont-translate.

At the end of the (yester)day after a long walk I decided against gettext. I believe that the amount of code we'll need to write to do everything from scratch will be comparable with the amount of code needed to (ugly) hack around gettext.

- gettext is opt-in: you mark the strings that need translating. We'd like to have opt-out to keep the code clean. Hence we'd have to write message extraction ourselves in any case.
- gettext's translation function are useless with f-strings. Hence we won't use gettext's function to get translations
- Hence, all we'd have frrom gettext would be a file format that's compatible with its other utilities. However, gettext marks messages with line numbers, which can change. Because we can limit ourselves to Python, we can instead "mark" the message according to its position -- module - class - method - internal function - internal function - ...

With the latter in mind and for the sake of translator-friendliness, I decided to put translations in YAML. Here's a part of the part that translates owtable. Leaves are strings, everything above are names of functions and classes and module. Messages are translated by simply replacing `''` with the translation, as shown below.

```
Orange/widgets/data/owtable.py:
    ApproxSummary: ''
    DenseArray: ''
    M: ''
    NotAvailable: ''
    OWDataTable:
        Data Table: Tabela podatkov
        Inputs:
            Data: ''
        Outputs:
            Selected Data: ''
        View the dataset in a spreadsheet.: ''
        __init__:
            Color by instance classes: Pobarvaj glede na razred
            Info: ''
            Restore Original Order: Začetni vrstni red
            Select full rows: Izbiraj cele vrstice
            Selection: Izbor
            Show rows in the original order: Pokaži vrstice v originalnem vrstnem redu
            Show variable labels (if present): Pokaži oznake (če obstajajo)
            Variables: Spremenljivke
            Visualize numeric values: Histogram številskih vrednosti
            auto_commit: ''
            color_by_class: ''
            select_rows: ''
            show_attribute_labels: ''
            show_distributions: ''
        _info_box_text:
            ' (no missing data)': ' (ni manjkajočih podatkov)'
            No data.: Ni podatkov
            format_part:
                ' ({perc:.1f} % missing data)': ' ({perc:.1f} % manjkajočih podatkov)'
                ' ({tag}, density {dens:.2f} %)': ( ({tag}, density {dens:.2f} %)'
                sparse: ''
                tags: ''
```

Screenshot 2022-01-22 at 17 31 04

janezd:
Continued as separate project, https://github.com/janezd/trubar.

PR #5807, opened 2022-01-26
[FIX] Ensure unlockable sparse arrays after Table.copy

### LINKED ISSUE #5746: Table: copy does not guarantee unlock to suceed
**What's wrong?**

Unlocking sparse data does not work.

Any of the following results in an error (ValueError: 'metas'/'X' is a view into another table and cannot be unlocked)
```
from Orange.data import Table

Table.LOCKING = True
data = Table("zoo")
data = data.to_sparse()
with data.unlocked():
    pass
```

```
from Orange.data import Table

Table.LOCKING = True
data = Table("iris")
data = data.to_sparse()
with data.unlocked():
    pass
data = data.copy()
with data.unlocked():
    pass
```

**What's your environment?**

- Orange version: master

markotoplak:
##### Issue
Fixes  #5746

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1465,9 +1465,10 @@ class Table(Sequence, Storage):
         """
 
         def is_view(x):
-            # Sparse matrices don't have views like numpy arrays. Since indexing on
-            # them creates copies in constructor we can skip this check here.
-            return not sp.issparse(x) and x.base is not None
+            if not sp.issparse(x):
+                return x.base is not None
+            else:
+                return x.data.base is not None
 
         if is_view(self._X):
             self._X = self._X.copy()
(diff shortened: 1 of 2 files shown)


PrimozGodec:
Tests are fixed in #5822

So if I correct @markotoplak:

```
from Orange.data import Table

Table.LOCKING = True
data = Table("zoo")
data = data.to_sparse()
with data.unlocked():  <-- this should still fail since to_sparse use transform and data can be a view
    pass

with data.copy().unlocked():  <-- this does not fail anymore
    pass
```

markotoplak:
@PrimozGodec, yes, you are correct.

PR #5810, opened 2022-01-27
[FIX] ExcelReader: Write roles

### LINKED ISSUE #5798: Excel Writer does not save roles
**What's wrong?**

Excel writer (or at least the widget) does not save roles, e.g. meta, target. This makes it rather useless.

Save zoo data as xlsx and `type` will no longer be a target.

**What's your environment?**
- Operating system: macOS
- Orange version: 3.30

VesnaT:
##### Issue

Fixes #5798

##### Description of changes
Beside headers, also save types and flags.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -277,6 +277,7 @@ class ExcelReader(_BaseExcelReader):
     EXTENSIONS = ('.xlsx',)
     DESCRIPTION = 'Microsoft Excel spreadsheet'
     ERRORS = ("#VALUE!", "#DIV/0!", "#REF!", "#NUM!", "#NULL!", "#NAME?")
+    OPTIONAL_TYPE_ANNOTATIONS = True
 
     def __init__(self, filename):
         super().__init__(filename)
@@ -319,7 +320,7 @@ class ExcelReader(_BaseExcelReader):
             return self.workbook.active
 
     @classmethod
-    def write_file(cls, filename, data):
+    def write_file(cls, filename, data, with_annotations=False):
         vars = list(chain((ContinuousVariable('_w'),) if data.has_weights() else (),
                           data.domain.attributes,
                           data.domain.class_vars,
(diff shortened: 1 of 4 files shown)


markotoplak:
From the code it seems that headers are always written, but this should, instead, respect users' setting in the widget.

Furthermore, for backward compatibility, the add type annotations could get disabled when opening saved workflows that save to Excel.

VesnaT:
> From the code it seems that headers are always written, but this should, instead, respect users' setting in the widget.

I wasn't even aware we have that. I'll fit it.

> Furthermore, for backward compatibility, the add type annotations could get disabled when opening saved workflows that save to Excel.

I'm not sure I want to support backward compatibility for something that didn't work correctly.

markotoplak:
> > Furthermore, for backward compatibility, the add type annotations could get disabled when opening saved workflows that save to Excel.
> 
> I'm not sure I want to support backward compatibility for something that didn't work correctly.

My perspective on this is that for some users, especially for those that saved those workflows, the output was exactly what they needed or they somehow adapted to. Perhaps they just accepted that type setting as something that does not apply to Excel files. If they loaded their workflows after the changes, the saved files are going to change, which can lead to problems later on.

The number of users relying on this is likely small and perhaps we can ignore them and just break things for them... But on the other hand, supporting that here is easy: we could just add the rule on setting open `if setting_version_old and use_excel: value_setting = False`. So no special support of old behaviors, just setting change.

Orange/data/io.py
```diff
@@ -319,7 +320,7 @@ def _get_active_sheet(self) -> openpyxl.worksheet.worksheet.Worksheet:
             return self.workbook.active
 
     @classmethod
-    def write_file(cls, filename, data):
+    def write_file(cls, filename, data, with_annotations=True):
```
markotoplak, line 323: Also this changes the behavior. I'd set it to False. My reasoning is that users who would not want to re-import the edited file back into Orange, would actually prefer it to be without annotations, which only make sense for Orange readers. :)

VesnaT: Fixed

PR #5813, opened 2022-01-28
Predictions: Remove rows with missing target when scoring

### LINKED ISSUE #5793: Scorers and instances with unknown outcomes
**What's wrong?**

When scorers, e.g. in Predictions, encounter data instances with unknown classes, they fail; the widget shows an error in the status bar, saying that some scorers have failed.

This is OK, but I wonder whether scorers could ignore such instances, or, perhaps better, the widget would remove them and show a warning instead of an error.

Train a tree from on this data (select data instances with known class value) and make predictions (using Predictions) for all data.

[Mushrooms.xlsx](https://github.com/biolab/orange3/files/7881889/Mushrooms.xlsx)

VesnaT:
##### Issue

Fixes #5793

##### Description of changes
- show warning when input data has missing targets
- remove instances with missing target while scoring
- set all scores to `NA` when all targets are missing

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -65,6 +65,8 @@ class OWPredictions(OWWidget):
         empty_data = Msg("Empty dataset")
         wrong_targets = Msg(
             "Some model(s) predict a different target (see more ...)\n{}")
+        missing_targets = Msg("Instances with missing targets "
+                              "are ignored while scoring.")
 
     class Error(OWWidget.Error):
         predictor_failed = Msg("Some predictor(s) failed (see more ...)\n{}")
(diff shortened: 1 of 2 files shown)


Orange/widgets/evaluate/owpredictions.py
```diff
@@ -291,19 +293,37 @@ def _update_scores(self):
...
+
+            actual = results.actual
+            predicted = results.predicted
+            probabilities = results.probabilities
+
+            mask = numpy.isnan(results.actual)
+            no_targets = mask.sum() == len(results.actual)
+            results.actual = results.actual[~mask]
+            results.predicted = results.predicted[:, ~mask]
+            results.probabilities = results.probabilities[:, ~mask]
```
janezd, line 305: Although it seems that the only potential source of exceptions is handled, I think I'd feel better if this modifying and restoring of `results` was secured with `try-finally`. In the future, somebody might add more code in between that could raise unhandled exceptions. `try-finally` blocks (especially `finally`) would also visually show what it going on here.

VesnaT: Lint doesn't aggree with that.

janezd: I agree with lint. :) Remove `except`, keep just `finally`. We don't want to mask the exception, the idea was just ensure that `result` is restored.

PR #5814, opened 2022-01-28
[ENH] New icons and splash screen

markotoplak:
From material from biolab/orange-graphics@6e4cdf5093bc5d66d595550aeaaa42ea9a2be0c6

~~Apart from this some things will have to also be replaced in the installer repository. There are some icons there and the DMG screens.~~

I tried to cover everything in this repo, but very likely I missed something...

~~Before merging this test installers should be built.~~

See installers in ales-erjavec/orange3-installers#23


ajdapretnar:
I am so happy this is finally moving in the right direction.

I have some comments, though.
- The resolution of the images, at least on my 4k screen, seems very grainy.
- We should probably replace the colors of core categories, too? I think this was the idea at least.
- Welcome screen is still as it was. Are we keeping it this way or are there any changes planned?

Otherwise, the images are in the right place, so all good. But definitely have to test it with an installer.

markotoplak:
>     * The resolution of the images, at least on my 4k screen, seems very grainy.

Which images? There are:
- It was just the icons, which are the same size and before (the largest being 256x256).
- And the splash screens, which are currently non-retina renders.

>     * Welcome screen is still as it was. Are we keeping it this way or are there any changes planned?

I'd leave it out of these. I do not think we agreed on anything there, for now.

ajdapretnar:
Splash screens. Probably because they are non-retina.

markotoplak:
@ajdapretnar, I do not have retinized splash screens ready, so I think that would be best handled in a different PR. Also the colors.

This PR would be best tested with builds from its companion installer PR: ales-erjavec/orange3-installers#23

I only tested the Mac installers.

markotoplak:
Links for testing (both unsigned):
- Windows: https://ci.appveyor.com/api/buildjobs/3y51eg5j4jeje595/artifacts/installers%2FOrange3-3.32.0.dev0%2B13af216-Miniconda-x86_64.exe
- MacOS: https://dev.azure.com/ales-erjavec/64837369-9704-4f7c-a434-1fe8e78f627a/_apis/build/builds/130/artifacts?artifactName=macOS%20dmg%20installer&api-version=7.0&%24format=zip

ajdapretnar:
I can confirm the .dmg installer works nicely for me.
We should keep in mind we are still "testing" #5712. So before the official release we should agree that we like the changes.

thocevar:
Works fine on Windows as well.

JakaKokosar:
Works fine on macOS Monterey (v12.1)

A minor thing about DMG folder style:

when I open the installer,
Screenshot 2022-02-06 at 13 37 09

expected (make window size match the background image)
Screenshot 2022-02-06 at 13 37 22

Issue #5779, opened 2022-01-13
Edit Domain: define date format for datetime conversion


ajdapretnar:
**What's your use case?**

Dates and timestamps come in all shapes and sizes. It would be nice if one could define the format on-the-fly, similar to the `strptime` function.

**What's your proposed solution?**

Provide a curated list of most common datetime formats to parse the time variable into ISO standard.
I.e.:
dd. mm. yyyy
dd. mm. yy
d. m. yyyy
d. m. yy
mm/dd/yy
mm/dd/yyyy
dd-mm-yy 
dd-mm-yyyy

**Are there any alternative solutions?**
Parse it implicitly, but this could lead to a variety of problems.

ajdapretnar:
I marked it as "needs discussion", because if we are in favour of adding this, we should decide how to expose this in GUI.

PrimozGodec:
We decided to:
- support the custom format in `TimeVariable`. The format is provided when the object is constructed and used for parsing. If it is not defined we use the current default - ISO format.
- add dropdown to select format in File widget
- add dropdown to select format in Edit Domain widget

mw25:
I like this feature. 👍 

But were you only talking about text to datetime conversion, or numeric to datetime conversion as well?
A discussion about numeric to datetime has already been started here: #5648 
Could this issue also be addressed in the current changes?

wvdvegte:
I think this feature would indeed be very useful!

PrimozGodec:
> @PrimozGodec I suppose for numeric variables, we could go with the suggested:
> 
> %Y%m%d%H%M%S
> %Y%m%d
> %Y%j
> %Y
> %H%M%S.%f
> Depending on the type of the feature, we could propose different format options.

@ajdapretnar I would not convert numeric values according to those formats in Edit Domain. The reasons are:
- since numeric values remove leading zeroes, there is no nice way of parsing them. I widget I would need to include a hardcoded format checking and adding leading zeros before conversion, which is ugly, and error prune for some changes in the future
- I think it is not necessary since users will be able to interpret values according to this format in the File widget (I will implement it after I finish with the Edit Domain) 

My proposed solution is that this format is still available for String and Discrete variables (where leading zeros are kept). We convert numeric variables to time as the UNIX epoch (it is done this way currently in Edit Domain).

PR #5819, opened 2022-02-02
[ENH] Datetime format selection

PrimozGodec:
##### Issue

Orange's widgets does not support datatime formats other than ISO.
Fixes https://github.com/biolab/orange3/issues/5779

##### Description of changes
Add dropdown with format selection to Edit domain widget

##### Note
When no year is provided (only day and month) - the last three formats the year will become 1900, since it is the default in python's DateTime and excel too.

##### Next steps
Implement the same functionality in the File widget --- will be implemented in the separate pull request

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/tests/test_variable.py
+++ b/Orange/data/tests/test_variable.py
@@ -13,6 +13,7 @@ from datetime import datetime, timezone
 from io import StringIO
 
 import numpy as np
+import pandas as pd
 import scipy.sparse as sp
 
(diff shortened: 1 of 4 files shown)


ajdapretnar:
This is a great solution and it works well. I have not tested **all possible** cases, but I tested many and they behave as they should.

PR #5820, opened 2022-02-03
Translations: Fixes for plural forms (and similar)

janezd:
##### Issue

If we decide to make Orange translatable (e.g. via https://github.com/janezd/trubar), widgets will need a more systematic approach to plural forms. This PR follows the idea explained at https://github.com/janezd/trubar#plural-forms.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation


Orange/widgets/utils/localization.py
```diff
@@ -0,0 +1,15 @@
+__all__ = ["pl", "plsi"]
+
+
+def pl(n, forms):
```
markotoplak, line 4: Hm, is it really "101 instance"?

Orange/widgets/utils/localization.py
```diff
@@ -0,0 +1,15 @@
...
+    return forms.split("|")[n % 100 != 1]
+
+def plsi(n, forms):
+    forms = forms.split("|")
```
markotoplak, line 8: Slovenian is a wonderful language. Is the rule for 3 and 4 really always the same? I did not know that, but the examples I can think of confirm this. Thanks!

markotoplak (commented):
I like the approach.

Thanks for teaching me grammar. :)

Orange/widgets/utils/localization.py
janezd: Oops. I don't know where I got this from. :) Fixed.

Orange/widgets/utils/localization.py
janezd: Yes, that's the rule. You never knew that you use it all the time, right?

You can see a list of forms for different languages here: https://github.com/translate/l10n-guide/blob/master/docs/l10n/pluralforms.rst. Apparently, only Slovenian, Maltese and variations of gaelic have more than 3 forms (Irish winning with 5). If this list is correct.

It seems that Slavic languages use a different form for plural from 5. (is it always genitive-like?)  "Trije -- kaj, trije stoli?" vs. "Pet -- česa, pet stolov?"). Slovenian then adds an extra one for dual.

Another nice complication is that numbers 11-14 in most Slavic languages use "normal" plural forms, while for n > 20, you need to observe n % 10 because these numbers are pronounced as, for instance, "fifty and one" (pedeset i jedan). In Slovenian, however, this does not apply because 51 is not "fifty and one"  but "one and fifty" (is this because of German influence, ein­und­fünfzig?), so 51 is plural. It applies again for hundred and one, though. See the rules for Serbian or Russian, for instance.

PR #5821, opened 2022-02-03
PyListModelTooltip: Improve handling of tooltip argument

janezd:
##### Issue

The `tooltip` argument for `PyListModelTooltip` has a unsafe default (a mutable list) and does not accept generators. Also, if the number of tooltips is smaller than the number of rows, it gives index error.

##### Description of changes

- Tooltip argument is now `()`.
- To support generators, non-sequences (instances of classes without `__getitem__` are converted to lists. It would be better to convert all to lists (even if already a list), to match the behaviour of `PyListModel` with respect to `DisplayRole`, but I did it this way to maintain backward compatibility with any widgets that might have counted on having a reference to data in the model.
- Index ranges are now checked, and `None` is returned if there are too few tooltips.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -144,13 +144,11 @@ Categorical features are passed as strings
         self.attributescb.setModel(self.attrs_model)
 
         sorted_funcs = sorted(self.FUNCTIONS)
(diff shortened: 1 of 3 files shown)


janezd:
A-ha! Feature Constructor was caught red-handed, changing the value of default argument!

PR #5822, opened 2022-02-03
Remove old settings support in Select Rows

PrimozGodec:
##### Issue

Happy 22/2/2!

Now remove support for version==None settings in
SelectRowsContextHandler.decode_setting and SelectRowsContextHandler.match,
and uncomment OWSelectRows.migrate.
In tests, uncomment test_migration_to_version_1,
and remove test_support_old_settings and this test.
Basically, revert this commit.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owselectrows.py
+++ b/Orange/widgets/data/tests/test_owselectrows.py
@@ -310,37 +310,6 @@ class TestOWSelectRows(WidgetTest):
         self.assertEqual(condition[1], 2)
         self.assertEqual(condition[2], (2, ))  # index of value + 1
 
-    def test_backward_compat_match_values(self):
-        iris = Table("iris")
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/tests/test_owselectrows.py
```diff
@@ -310,37 +310,6 @@ def test_partial_match_values(self):
         self.assertEqual(condition[1], 2)
         self.assertEqual(condition[2], (2, ))  # index of value + 1
 
-    def test_backward_compat_match_values(self):
```
PrimozGodec, line 313: If I understand correctly from `# After 2022/2/2, use just the expression in else clause` indices instead of values in context values also not supported anymore. Removing this test.

Issue #5818, opened 2022-02-02
FR: Neural Network without Regularization


dsanalytics:
I know that very small alpha is essentially no regularization, but would it be possible to have zero for alpha as the left-most value on that slider, plus to have the label change to "Regularization: Off" if alpha set to zero? It'd just look more professional and easier to check visually. See 2 screenshots for what I mean. Thank you in advance.

![image](https://user-images.githubusercontent.com/32625394/152096369-2fe961fb-f7cc-4916-957a-2b00c1056e77.png) ![image](https://user-images.githubusercontent.com/32625394/152096476-76e5955f-a337-4ea7-891a-ca550e5ead8f.png)

Env: Windows + Orange 3.31

janezd:
Resolved in #5824.

PR #5824, opened 2022-02-04
Neural Network: include no layers and no regularization

### LINKED ISSUE #5817: FR: Single-Layer Neural Network possible?
I could be misreading NN widget help file [1], but is it possible to specify a single-layer NN - i.e. no hidden layers? If I enter 0 for the number of hidden layers neurons, it gives an error (screenshot no 1), while if I leave it blank, it defaults to 10. As an example, for a NN with 4 inputs, one output with logistic, what would I need to input in the neurons count box to get it (screenshot no 2)?

image image

[1] Orange Data Mining - Neural Network Widget
https://orangedatamining.com/widget-catalog/model/neuralnetwork/

Env: Windows + Orange 3.31

ajdapretnar:
##### Issue

Resolves #5817 and #5818. 

##### Description of changes
Enable no layers and no regularization.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owneuralnetwork.py
+++ b/Orange/widgets/model/owneuralnetwork.py
@@ -14,6 +14,7 @@ from AnyQt.QtCore import pyqtSlot as Slot, pyqtSignal as Signal
 from Orange.data import Table
 from Orange.modelling import NNLearner
 from Orange.widgets import gui
+from Orange.widgets.widget import Msg
 from Orange.widgets.settings import Setting
 from Orange.widgets.utils.owlearnerwidget import OWBaseLearner
 
@@ -81,11 +82,11 @@ class OWNNLearner(OWBaseLearner):
     activation_index = Setting(3)
     solver_index = Setting(2)
     max_iterations = Setting(200)
-    alpha_index = Setting(0)
+    alpha_index = Setting(1)
     replicable = Setting(True)
-    settings_version = 1
+    settings_version = 2
 
-    alphas = list(chain([x / 10000 for x in range(1, 10)],
+    alphas = list(chain([0], [x / 10000 for x in range(1, 10)],
                         [x / 1000 for x in range(1, 10)],
                         [x / 100 for x in range(1, 10)],
                         [x / 10 for x in range(1, 10)],
@@ -93,6 +94,11 @@ class OWNNLearner(OWBaseLearner):
                         range(10, 100, 5),
                         range(100, 200, 10),
                         range(100, 1001, 50)))
+
+    class Warning(OWBaseLearner.Warning):
+        no_layers = Msg("ANN without hidden layers is equivalent to logistic "
+                        "regression with worse fitting.\nWe recommend using "
+                        "logistic regression.")
 
     def add_main_layout(self):
         # this is part of init, pylint: disable=attribute-defined-outside-init
@@ -184,10 +190,10 @@ class OWNNLearner(OWBaseLearner):
                 ("Replicable training", self.replicable))
 
     def get_hidden_layers(self):
+        self.Warning.no_layers.clear()
         layers = tuple(map(int, re.findall(r'\d+', self.hidden_layers_input)))
         if not layers:
-            layers = (10,)
-            self.hidden_layers_input = "10,"
+            self.Warning.no_layers()
         return layers
 
     def update_model(self):
@@ -306,6 +312,8 @@ class OWNNLearner(OWBaseLearner):
             if alpha is not None:
                 settings["alpha_index"] = \
                     np.argmin(np.abs(np.array(cls.alphas) - alpha))
+        elif version < 2:
+            settings["alpha_index"] = settings.get("alpha_index", 0) + 1
 
 
 if __name__ == "__main__":  # pragma: no cover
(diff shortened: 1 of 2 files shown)


Orange/widgets/model/tests/test_owneuralnetwork.py
```diff
@@ -21,3 +21,14 @@ def test_migrate_setting(self):
...
+    def test_no_layer_warning(self):
+        self.widget.apply_button.button.click()
+
+        self.widget.hidden_layers_input = ""
+        self.widget.apply_button.button.click()
+        self.assertTrue(self.widget.Warning.no_layers.is_shown())
+
+        self.widget.hidden_layers_input = "10,"
+        self.widget.apply_button.button.click()
+        self.widget.apply_button.button.click()
```
janezd, line 34: Did you want to check the widget's resilience to double clicks or did you mean `self.assertFalse(self.widget.Warning.no_layers.is_shown())`? :)

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -19,6 +19,7 @@
 
 from Orange.widgets.utils.concurrent import ThreadExecutor, FutureWatcher
 from Orange.widgets.utils.widgetpreview import WidgetPreview
+from orangewidget.widget import Msg
```
janezd, line 22: Move this line higher, above all imports from Orange. (See pylint.)

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -81,11 +82,11 @@ class OWNNLearner(OWBaseLearner):
     activation_index = Setting(3)
     solver_index = Setting(2)
     max_iterations = Setting(200)
-    alpha_index = Setting(0)
+    alpha_index = Setting(1)
```
VesnaT, line 85: Have you considered settings migration, since alpha_index = 0 now represents alpha = 0?

ajdapretnar: Oh boy, you are right. Thank you for noticing. Will add migration.

Orange/widgets/model/tests/test_owneuralnetwork.py
ajdapretnar: Definitely wanted to make sure the button is certainly clicked. 😆 

Will fix it.

Orange/widgets/model/owneuralnetwork.py
ajdapretnar: I fixed it. Would you mind checking it so that it is not a complete disaster?

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -306,6 +312,10 @@ def migrate_settings(cls, settings, version):
             if alpha is not None:
                 settings["alpha_index"] = \
                     np.argmin(np.abs(np.array(cls.alphas) - alpha))
+        elif version < 2:
+            alpha = settings.pop("alpha", None)
```
VesnaT, line 316: Considering the comment below, I'd go for something like:
`settings["alpha_index"] = settings.get("alpha_index", 0) + 1`

Orange/widgets/model/tests/test_owneuralnetwork.py
```diff
@@ -21,3 +21,18 @@ def test_migrate_setting(self):
         settings = dict(alpha=103)
         OWNNLearner.migrate_settings(settings, None)
         self.assertEqual(OWNNLearner.alphas[settings["alpha_index"]], 100)
+
+        settings = dict(alpha=0.0001)
```
VesnaT, line 25: `alpha` does not appear in the `settings` with `version = 1`, so this should be replaced with `settings = dict(alpha_index=0)` which does not pass.

Orange/widgets/model/tests/test_owneuralnetwork.py
```diff
@@ -21,3 +21,18 @@ def test_migrate_setting(self):
         settings = dict(alpha=103)
         OWNNLearner.migrate_settings(settings, None)
```
VesnaT, line 22: For simplicity purposes this could be changed to `OWNNLearner.migrate_settings(settings, 0)`. 
That's what actually happens, when there is no version.

Orange/widgets/model/tests/test_owneuralnetwork.py
ajdapretnar: Is it possible that alpha is in `version = 1`, but not in a `version is None`? Because now we are at `settings_version = 2`, which was upped from before. And I did not introduce alpha in this version, it was already there.

VesnaT: alpha could (probably) be in version = 1, only if the workflow was build upon version = 0, which is not even correct, because alpha is no longer a setting (replaced by alpha_index)

Orange/widgets/model/owneuralnetwork.py
```diff
@@ -184,10 +190,10 @@ def get_learner_parameters(self):
                 ("Replicable training", self.replicable))
 
     def get_hidden_layers(self):
+        self.Warning.no_layers.clear()
         layers = tuple(map(int, re.findall(r'\d+', self.hidden_layers_input)))
         if not layers:
-            layers = (10,)
-            self.hidden_layers_input = "10,"
+            self.Warning.no_layers()
```
janezd, line 196: This is OK, but you could also skip clearing and do `self.Warning.no_layers(shown=not layers)`.

PR #4846, opened 2020-06-08
Check that Domain is unique

AndrejaKovacic:
##### Issue

Fixes  #4804

##### Includes
- [X] Code changes
- [X] Tests

PR #5828, opened 2022-02-04
[FIX] Linear Projection: Fix LDA

### LINKED ISSUE #5359: LDA not working in Linear Projection widget ?
LDA crashes with a python error in 3.28 (Windows) : "_y should be a 1d array, got an array of shape (150,0) instead_".

image
image

Using iris dataset, target defined as "iris". The classes are properly recognized.

In the Linear Analysis Widget, I can do a PCA or a circular projection but get an empty window and an error message with LDA :

image
image

Unless I've done something wrong (what ?), this looks like a bug in the widget.

lanzagar:
##### Issue

Fixes #5359

##### Description of changes
I thought about removing the overriden `effective_data` (which also fixes LDA), but then saw that it was done based on a talk between @janezd and @VesnaT (https://github.com/biolab/orange3/pull/4846#issuecomment-643242265) who must have had a good reason for it.
So I just added the class to effective_data, but @janezd or @VesnaT should confirm that that makes sense.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owlinearprojection.py
+++ b/Orange/widgets/visualize/owlinearprojection.py
@@ -329,7 +329,10 @@ class OWLinearProjection(OWAnchorProjectionWidget):
 
     @property
     def effective_data(self):
-        return self.data.transform(Domain(self.effective_variables))
+        cvs = None
+        if self.placement == Placement.LDA:
+            cvs = self.data.domain.class_vars
+        return self.data.transform(Domain(self.effective_variables, cvs))
 
     def __vizrank_set_attrs(self, attrs):
         if not attrs:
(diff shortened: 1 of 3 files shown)


VesnaT:
Adding the class to effective_data makes sense, but the _invalidated/_domain_invalidated flags should be changed accordingly.

The following example does not work correctly:
1 . File (Iris) -> Select Columns (Reset) -> Linear Projection (LDA)
2. In the Select Columns move Iris to Ignored -> In Linear Projection the Circular Placement is selected   and LDA is disable but the plot is not changed.
3. In the Select Columns move Iris back to Target -> In Linear Projection the Circular Placement in selected and LDA enabled but the plot is still not changed.

Orange/widgets/visualize/owlinearprojection.py
```diff
@@ -329,7 +329,8 @@ def effective_variables(self):
 
     @property
     def effective_data(self):
-        return self.data.transform(Domain(self.effective_variables))
+        return self.data.transform(Domain(self.effective_variables,
+                                          self.data.domain.class_vars))
```
VesnaT, line 333: I just realised this wont work because class variable is duplicated when selected in effective variables.

PR #5833, opened 2022-02-07
Line Plot: Plot instances with missing values

### LINKED ISSUE #5741: Line Plot unnecessarily strict about missing values
**What's your use case?**

If a row of data has even one missing data item, Line Plot refuses to plot it.  I think this is unnecessarily strict—it should be possible to plot the line up to and including the item before the missing item, and then restart after the missing item(s).  Clearly I can manually remove undesirable columns, as shown in the attached workflow, so why shouldn’t Line Plot be able to do so automatically?  (The advantage would also be that in an automatic solution there would be visible gaps in the diagram, in my example the missing data are just treated is if there was no year 1964 to begin with.)

[LinePlot.zip](https://github.com/biolab/orange3/files/7728446/LinePlot.zip)

Line Plot should be able to plot data with missing values.

VesnaT:
##### Issue

Fixes #5741

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


ajdapretnar:
I like the way it looks and most of it works really well. There's just one extreme case I'm unsure about.
A row with all nan but one value. A fully dashed line is drawn, but I cannot select it not does it get highlighted in selection. Is this intended?

Screen Shot 2022-02-11 at 13 47 34

When a line gets drawn (has at least two consecutive values), I can select it and highlight it. If this is intentional, I can merge.

VesnaT:
Not being able not select a dashed line is intentional, but the vertical line isn't. With a row full of nans but one no line should appear, because it's impossible to know where to connect. I'll look into that. Thanks for noticing.

VesnaT:
Fixed.

PR #5835, opened 2022-02-08
[ENH] Server embedder: use queue, handle unsuccessful requests at the end

PrimozGodec:
##### Issue

Server embedder retries sending unsuccessful requests immediately after the fauliture. 
Since it is possible that the server needs longer to compute (the request was unsuccessful because of timeout) the results (the server will cache the result when finished) it is better to retry later (there is a higher chance that the server already finished processing).

##### Description of changes
Implemented described behaviour. To achieve it I needed to change embedders to use Queue instead of just gathering the results. 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/misc/server_embedder.py
+++ b/Orange/misc/server_embedder.py
@@ -3,6 +3,7 @@ import json
 import logging
 import random
 import uuid
+from collections import namedtuple
 from json import JSONDecodeError
 from os import getenv
 from typing import Any, Callable, List, Optional
@@ -10,22 +11,22 @@ from typing import Any, Callable, List, Optional
 from AnyQt.QtCore import QSettings
 from httpx import AsyncClient, NetworkError, ReadTimeout, Response
 
-from Orange.misc.utils.embedder_utils import (EmbedderCache,
-                                              EmbeddingCancelledException,
-                                              EmbeddingConnectionError,
-                                              get_proxies)
+from Orange.misc.utils.embedder_utils import (
+    EmbedderCache,
+    EmbeddingCancelledException,
+    EmbeddingConnectionError,
+    get_proxies,
+)
 
 log = logging.getLogger(__name__)
+TaskItem = namedtuple("TaskItem", ("id", "item", "no_repeats"))
 
 
 class ServerEmbedderCommunicator:
     """
     This class needs to be inherited by the class which re-implements
-    _encode_data_instance and defines self.content_type. For sending a table
-    with data items use embedd_table function. This one is called with the
-    complete Orange data Table. Then _encode_data_instance needs to extract
-    data to be embedded from the RowInstance. For images, it takes the image
-    path from the table, load image, and transform it into bytes.
+    _encode_data_instance and defines self.content_type. For sending a list
+    with data items use embedd_table function.
 
     Attributes
     ----------
@@ -69,14 +70,14 @@ class ServerEmbedderCommunicator:
             ) or str(uuid.getnode())
         except TypeError:
             self.machine_id = str(uuid.getnode())
-        self.session_id = str(random.randint(1, 1e10))
+        self.session_id = str(random.randint(1, int(1e10)))
 
         self._cache = EmbedderCache(model_name)
 
         # default embedding timeouts are too small we need to increase them
         self.timeout = 180
-        self.num_parallel_requests = 0
-        self.max_parallel = max_parallel_requests
+        self.max_parallel_requests = max_parallel_requests
+
         self.content_type = None  # need to be set in a class inheriting
 
     def embedd_data(
@@ -111,8 +112,7 @@ class ServerEmbedderCommunicator:
         EmbeddingCancelledException:
             If cancelled attribute is set to True (default=False).
         """
-        # if there is less items than 10 connection error should be raised
-        # earlier
+        # if there is less items than 10 connection error should be raised earlier
         self.max_errors = min(len(data) * self.MAX_REPEATS, 10)
 
         loop = asyncio.new_event_loop()
@@ -121,11 +121,9 @@ class ServerEmbedderCommunicator:
             embeddings = asyncio.get_event_loop().run_until_complete(
                 self.embedd_batch(data, processed_callback)
             )
-        except Exception:
+        finally:
             loop.close()
-            raise
 
-        loop.close()
         return embeddings
 
     async def embedd_batch(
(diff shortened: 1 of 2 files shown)


Orange/misc/server_embedder.py
```diff
@@ -153,32 +158,80 @@ async def embedd_batch(
...
+                """Callback called on every successful embedding"""
+                callback(next(progress_items))
+
+        else:
+            warnings.warn(
+                "process_callback is deprecated and will be removed in version 3.33, "
+                "use callback instead",
+                FutureWarning,
+            )
+            success_callback = processed_calback
```
markotoplak, line 175: This won't work due to a typo. :) If you are still supporting it (even for 1 version), please add a test.

Orange/misc/server_embedder.py
```diff
@@ -111,25 +118,25 @@ def embedd_data(
...
-            raise
 
-        loop.close()
         return embeddings
 
     async def embedd_batch(
-            self, data: List[Any], proc_callback: Callable[[bool], None] = None
+        self,
+        data: List[Any],
+        processed_calback: Optional[Callable] = None,
```
markotoplak, line 138: Typo. :) Now I see that you mistyped both uses so the thing actually works!

Why are you renaming it? Renaming makes calls as a kwarg (previously valid) invalid.

Orange/misc/server_embedder.py
```diff
@@ -111,25 +118,25 @@ def embedd_data(
...
+        callback: Callable = dummy_callback,
```
markotoplak, line 139: Perhaps force using callback as a keyword argument (with `*`)? That will prevent unintentional mixups when the deprecated parameter is removed. Then you won't be able to use the default values though, but that could be handled internally. So perhaps `embedd_batch(self, data, *proc_callback, callback=dummy_callback)`, which after deprecation becomes   `embedd_batch(self, data, *, callback=dummy_callback)`.

Orange/misc/server_embedder.py
```diff
@@ -197,63 +250,75 @@ async def _encode_data_instance(
...
+        client: AsyncClient,
+        queue: asyncio.Queue,
+        results: List,
+        proc_callback: Callable[[bool], None] = None,
+    ):
         """
-        Function get an data instance. It extract data from it and send them to
-        server and retrieve responses.
+        Worker that embedds data. It is pulling items from the until the queue
+        is empty. It is canceled by embedd_batch all tasks are finished
```
markotoplak, line 261: Probably WHEN all tasks are finished?

Orange/misc/server_embedder.py
```diff
@@ -197,63 +250,75 @@ async def _encode_data_instance(
...
+                    proc_callback()
+            elif num_repeats < self.MAX_REPEATS:
+                log.debug("Embedding unsuccessful - reading to queue:  %s", cache_key)
+                # if embedding not successful put the item to queue to be handled at
+                # the end - the item is put to the end since it is possible that  server
+                # still process the request and the result will be in the cache later
+                # repeating the request immediately may result in another fail when
+                # processing takes longer
+                queue.put_nowait(TaskItem(i, data_instance, no_repeats=num_repeats))
+            queue.task_done()
```
markotoplak, line 321: Shouldn't task_done be called only when the data_instance was correctly processed, so in the `if emb is not None` part? Or do I miss something?

Orange/misc/server_embedder.py
```diff
@@ -197,63 +250,75 @@ async def _encode_data_instance(
...
-            # maximally for MAX_REPEATS time
-            for i in range(1, self.MAX_REPEATS + 1):
-                self.__check_cancelled()
+        while not queue.empty():
+            # remove in 3.33
+            self.__check_cancelled()
+
+            # get item from the queue
+            i, data_instance, num_repeats = await queue.get()
+            num_repeats += 1
```
markotoplak, line 284: This does not increase `num_repeats` within the namedtuple, so effectively, `num_repeats` is always 1.

Orange/misc/server_embedder.py
```diff
@@ -153,32 +158,80 @@ async def embedd_batch(
...
-                    raise EmbeddingCancelledException()
-                requests.append(self._send_to_server(p, client, proc_callback))
+            tasks = self._init_workers(client, queue, results, success_callback)
 
-            embeddings = await asyncio.gather(*requests)
-        self._cache.persist_cache()
-        assert self.num_parallel_requests == 0
+            # wait for the queue to complete or one of workers to exit
+            queue_complete = asyncio.create_task(queue.join())
+            await asyncio.wait(
```
markotoplak, line 191: I am not sure I completely understand this stopping criterion. Just the double check the intent on this wait: you wait for completion OR for one of the tasks to raise an EmbeddingConnectionError and thus crash?

Orange/misc/server_embedder.py
markotoplak: Oh, no, my mistake: I missed that an element is put into the queue beforehand if the computation was not successful, so ignore the previous comment.

markotoplak:
@PrimozGodec, I like the concept and structuring. I left you some minor comments and suggestions.

In the end, I think I found:
- 2 imaginary serious bugs, where I was only complaining because I did not understand the code properly. 
- 1 bug that I still believe in. :)

Orange/misc/server_embedder.py
PrimozGodec: Before I put it back in the queue I create a new tuple so it is increased (line 320). If everything is OK (it does not need to be sent back to the server) the value does not matter.
I increase it here since it is also used in URL and need to be increased before.

Orange/misc/server_embedder.py
PrimozGodec: As you noticed it was not successful processed but the new item is in the queue and the current one is done.

Orange/misc/server_embedder.py
markotoplak: You are correct, sorry for bugging you!

Orange/misc/server_embedder.py
PrimozGodec: It was a mistake since first I wanted to fix it one way and then I changed back.

Orange/misc/server_embedder.py
PrimozGodec: It is a good idea to force callback to be keyword arg. I add * after proc_callback since it has the default parameter (also before it had).

PrimozGodec:
@markotoplak it should be fixed now

Orange/misc/server_embedder.py
markotoplak: Ha, I did not know you can also have it in the middle. :)

PR #5840, opened 2022-02-12
[FIX] Scatter plot: Fix rotation of regression line label

### LINKED ISSUE #5721: Negative regressions are printed upside-down
**What's wrong?**

As shown in the attached screenshot, if a regression line has a negative slope, the _r_ value is printed upside-down.  I really can’t imagine that it would be common for viewers to prefer that.  The second screenshot also demonstrates that the positioning of the label does not take the markers in the view into account, so the text can be completely lost.

The attached worksheet demonstrates the issue.  Just open the Scatterplot widget.

**What's your environment?**

- Operating system: macOS 12.0.1
- Orange version: 3.30.2
- How you installed Orange: From the disk image at https://download.biolab.si/download/files/Orange3-3.30.2-Python3.8.8.dmg

Screenshot 2021-12-06 at 01 21 32
Screenshot 2021-12-06 at 01 27 57
[Municipalities, Sweden.csv](https://github.com/biolab/orange3/files/7657129/Municipalities.Sweden.csv)
[AreaPopulationScatterPlot.zip](https://github.com/biolab/orange3/files/7657130/AreaPopulationScatterPlot.zip)

janezd:
##### Issue

Fixes #5721.

I've no idea why I though this needs to be rotated. It's been four years; perhaps something changed in pyqtgraph?

##### Description of changes

Remove rotation.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -244,17 +244,12 @@ class OWScatterPlotGraph(OWScatterPlotBase):
         slope, intercept, rvalue, _, _ = linregress(x, y)
         angle = np.degrees(np.arctan(slope))
         start_y = min_x * slope + intercept
-        rotate = 135 < angle % 360 < 315
-        l_opts = dict(color=color, position=abs(rotate - 0.85),
+        l_opts = dict(color=color, position=0.85,
                       rotateAxis=(1, 0), movable=True)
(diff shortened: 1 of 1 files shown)


ajdapretnar:
Tested on pyqtgraph==0.12.3. Even with negative correlation it was not reversed. Tried iris, housing, heart-disease. From this, I will assume it works.

PR #5848, opened 2022-02-17
[ENH] Enable multitarget problem types for OWTestAndScore and OWPredictions

JakaKokosar:
##### Issue
In survival analysis, it is expected to have two target variables. First is the duration of time until the event of interest, and the second is the indicator of censorship.

Preferably, the Survival Analysis add-on would then use the same infrastructure for testing and scoring survival models as it is currently in place for classification and regression related problems. To achieve this, we need to loosen the constraint of a single target variable for the input data.

##### Description of changes

With this pull request, the task was not to change the interface to accommodate for all future tasks one would like to support but to:

- allow the use of multitarget data and make no breaking changes to the existing codebase (do not change the default behaviour),
- for a start, make survival models/scorers available in Orange.
<strike>
Since the registration of Scorers is already implemented, the next step was how to find the usable scorers given the input data. The Scorer base class now holds additional [information](https://github.com/biolab/orange3/compare/master...JakaKokosar:multi_target?expand=1#diff-ebac791194327a764153704a5e2567261585ad3971623c2138be81e8c02b8da5R69) to recognise Scorers that are built-in and those implemented through add-ons.

**If Scorer is defined as built-in, nothing changes**. For non-built-in scorers, we look into Table attributes to determine the 'problem type' of input data. For example, for the survival analysis, As Survival data widget will set class variables to the output table and set attributes of the table as follows:
```python
{
	..., 
	'problem_type': 'time_to_event'
}
```
Usable scorers are those that match the same problem_type with input data. This is not necessarily the best solution and could use further debate. At this stage, no significant changes to the code-base were needed. In theory, everything else should be handled by Learners, Models and Scorers defined for related tasks https://github.com/biolab/orange3-survival-analysis/pull/27. 
<strike/>

Some examples:
Screenshot 2022-02-17 at 15 37 02

Screenshot 2022-02-17 at 15 37 15

Screenshot 2022-02-17 at 15 38 15

Screenshot 2022-02-17 at 15 57 55

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -3,12 +3,12 @@ import itertools
 from collections.abc import Iterable
 import re
 import warnings
-from typing import Callable, Dict
+from typing import Callable, Dict, Optional
 
 import numpy as np
 import scipy
 
-from Orange.data import Table, Storage, Instance, Value
+from Orange.data import Table, Storage, Instance, Value, Domain
 from Orange.data.filter import HasClass
 from Orange.data.table import DomainTransformationError
 from Orange.data.util import one_hot
@@ -86,6 +86,9 @@ class Learner(ReprableWithPreprocessors):
     #: A sequence of data preprocessors to apply on data prior to
     #: fitting the model
     preprocessors = ()
+
+    # Note: Do not use this class attribute.
+    #       It remains here for compatibility reasons.
     learner_adequacy_err_msg = ''
 
     def __init__(self, preprocessors=None):
@@ -95,6 +98,7 @@ class Learner(ReprableWithPreprocessors):
         elif preprocessors:
             self.preprocessors = (preprocessors,)
 
+    # pylint: disable=R0201
     def fit(self, X, Y, W=None):
         raise RuntimeError(
             "Descendants of Learner must overload method fit or fit_storage")
@@ -106,8 +110,23 @@ class Learner(ReprableWithPreprocessors):
         return self.fit(X, Y, W)
 
     def __call__(self, data, progress_callback=None):
-        if not self.check_learner_adequacy(data.domain):
-            raise ValueError(self.learner_adequacy_err_msg)
+
+        for cls in type(self).mro():
+            if 'incompatibility_reason' in cls.__dict__:
+                incompatibility_reason = \
+                    self.incompatibility_reason(data.domain)  # pylint: disable=assignment-from-none
+                if incompatibility_reason is not None:
+                    raise ValueError(incompatibility_reason)
+                break
+            if 'check_learner_adequacy' in cls.__dict__:
+                warnings.warn(
+                    "check_learner_adequacy is deprecated and will be removed "
+                    "in upcoming releases. Learners should instead implement "
+                    "the incompatibility_reason method.",
+                    OrangeDeprecationWarning)
+                if not self.check_learner_adequacy(data.domain):
+                    raise ValueError(self.learner_adequacy_err_msg)
+                break
 
         origdomain = data.domain
 
@@ -175,6 +194,11 @@ class Learner(ReprableWithPreprocessors):
 
     def check_learner_adequacy(self, _):
         return True
+
+    # pylint: disable=no-self-use
+    def incompatibility_reason(self, _: Domain) -> Optional[str]:
+        """Return None if a learner can fit domain or string explaining why it can not."""
+        return None
 
     @property
     def name(self):
--- a/Orange/evaluation/clustering.py
+++ b/Orange/evaluation/clustering.py
@@ -32,6 +32,10 @@ class ClusteringResults(Results):
 class ClusteringScore(Score):
     considers_actual = False
 
+    @staticmethod
+    def is_compatible(domain) -> bool:
+        return True
+
     # pylint: disable=arguments-differ
     def from_predicted(self, results, score_function):
         # Clustering scores from labels
--- a/Orange/regression/base_regression.py
+++ b/Orange/regression/base_regression.py
@@ -6,15 +6,13 @@ __all__ = ["LearnerRegression", "ModelRegression",
 
 class LearnerRegression(Learner):
 
-    def check_learner_adequacy(self, domain):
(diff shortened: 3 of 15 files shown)


janezd (commented):
I read the code to see the idea. My comments refer to what I spotted, and do not mean I like the idea. (I don't. :)

The problem is not the idea itself. I dislike is that it is rather a patch over bad overall design. We should consider some deeper changes, though I fear they lead towards discussing finally moving to pandas.

I think that in light of survival analysis and similar problems, we should consider multiple roles of variables. Currently, a variable can be an independent variable, a dependent variable or a meta, and they are stored in different matrices. We had another type (a weight), which was supposed to be unique (e.g. you cannot choose between multiple weight variables on the same data). You propose to add another role...

Going pandas would mean abandoning X, Y and metas as permanently materialized, and instead having column-based representation. At the same time, every column could be assigned a role. `class_var` would then be a property that would return the variable that is assigned a "target role".

We need to decide whether to continue patching or bite into pandas.

Orange/base.py
```diff
@@ -173,8 +173,8 @@ def active_preprocessors(self):
                 self.preprocessors is not type(self).preprocessors):
             yield from type(self).preprocessors
 
-    def check_learner_adequacy(self, _):
```
markotoplak, line 176: This is dangerous. You are changing a signature of a Learner base class method. That would need deprecation.

Perhaps make a new separate method that just returns the error, like `explain_inadequacy(domain)`. If it returns None, there is no error. :) Then `check_learner_adequacy()` may be just `return explain_inadequacy(..) == None`.

And if we are making new functions, I really dislike the current term, "adequate". I'd replace the current `check_learner_adequacy` with `fits` or `can_fit`: `learner.fits(domain)`, isn't it nicely readable? Maybe "can_fit" would promise too much - what if fitting crashed? :) Perhaps `is_compatible`...

The error message (or None) function might rather be something like `explain_incompatibility`.

@janezd, which names would you choose?

Orange/regression/base_regression.py
```diff
@@ -5,10 +5,14 @@
...
-    learner_adequacy_err_msg = "Numeric class variable expected."
 
     def check_learner_adequacy(self, domain):
-        return domain.has_continuous_class
+        err_msg = ""
+        if len(domain.class_vars) > 1:
+            err_msg = "Too many target variables."
+        elif not domain.has_continuous_class:
+            err_msg = "Numeric class variable expected."
+        return not err_msg, err_msg
```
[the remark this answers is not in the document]
markotoplak: Not OK for me. :) I just commented above.

Orange/base.py
```diff
@@ -86,7 +86,6 @@ class Learner(ReprableWithPreprocessors):
     #: A sequence of data preprocessors to apply on data prior to
     #: fitting the model
     preprocessors = ()
-    learner_adequacy_err_msg = ''
```
markotoplak, line 89: You can not just delete such a public attribute. This needs to stay and be deprecated if not used anymore.

janezd: Ooops. In my first review, I incorrectly assumed that @JakaKokosar added this attribute. I'm sorry, Jaka. :(

I still believe that having such attributes is a bad idea. (I hope it wasn't me who added it.) Please, let us deprecate it. Or make it a property.

Orange/evaluation/clustering.py
```diff
@@ -32,6 +32,10 @@ def get_fold(self, fold):
 class ClusteringScore(Score):
     considers_actual = False
 
+    @staticmethod
+    def is_adequate(domain) -> bool:
```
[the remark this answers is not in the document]
janezd: Yes.

Orange/base.py
markotoplak: `incompatility_reason` is fine.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -246,8 +246,9 @@ def check_data(self):
         self.Error.sparse_not_supported.clear()
         if self.data is not None and self.learner is not None:
             self.Error.data_error.clear()
-            if not self.learner.check_learner_adequacy(self.data.domain):
-                self.Error.data_error(self.learner.learner_adequacy_err_msg)
+            incompatibility_reason = self.learner.incompatibility_reason(self.data.domain)
```
markotoplak, line 249: @JakaKokosar, the thing we just spoke about (old learners still using deprecated functions) also applies here.

Orange/base.py
```diff
@@ -106,8 +110,23 @@ def fit_storage(self, data):
...
+            if 'incompatibility_reason' in cls.__dict__:
+                incompatibility_reason = \
+                    self.incompatibility_reason(data.domain)  # pylint: disable=assignment-from-none
+                if incompatibility_reason is not None:
+                    raise ValueError(incompatibility_reason)
+                break
+            if 'check_learner_adequacy' in cls.__dict__:
+                warnings.warn(
+                    "check_learner_adequacy is deprecated and will be removed "
+                    "in upcoming releases. Please use incompatibility_reason "
```
markotoplak, line 124: Rephrase to something like "Learners should instead implement..." because the user does not actually call these functions, he just implements them.

Orange/widgets/utils/owlearnerwidget.py
```diff
@@ -246,8 +248,26 @@ def check_data(self):
...
+            for cls in type(self.learner).mro():
+                if 'incompatibility_reason' in cls.__dict__:
+                    # pylint: disable=assignment-from-none
+                    incompatibility_reason = \
+                        self.learner.incompatibility_reason(self.data.domain)
+                    break
+                if 'check_learner_adequacy' in cls.__dict__:
+                    warnings.warn(
+                        "check_learner_adequacy is deprecated and will be removed "
+                        "in upcoming releases. Please use incompatibility_reason "
```
markotoplak, line 262: As in the previous comment

PR #5857, opened 2022-02-24
SOM: Recompute only when effective data changes

VesnaT:
##### Issue

SOM recalculates when new data appears on the input, even though the data might be the same.

##### Description of changes
- check if data has change since the last plotting and recalculate only if necessary
- consider only continuous variables

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owsom.py
+++ b/Orange/widgets/unsupervised/owsom.py
@@ -14,6 +14,7 @@ from AnyQt.QtWidgets import \
     QGraphicsPathItem
 
 from Orange.data import Table, Domain
+from Orange.data.util import array_equal
 from Orange.preprocess import decimal_binnings, time_binnings
 from Orange.projection.som import SOM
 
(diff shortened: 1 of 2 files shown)


markotoplak:
I find this pattern problematic. It requires data storage within the widget.

What was your particular use case? Did you use Select Columns before that? Or is there some malfunctioning widget that repeatedly sends the same outputs?

I imagine this is the best solution for now, but we should try to think about how to avoid this.

Orange/widgets/unsupervised/owsom.py
```diff
@@ -345,6 +344,12 @@ def set_warnings():
             else:
                 prepare_data()
 
+        invalidated = not (cont_x is not None and self.cont_x is not None
+                           and array_equal(cont_x, self.cont_x))
```
janezd, line 348: I would find `cont_x is None or self.cont_x is None or not array_equal(cont_x, self.cont_x)` more readable.

I understand what you meant ("it is not invalidated if both are non-None and equal"), but I find this difficult to process.

There's also an edge case: if both are `None`, `invalidated` should be `False`. This could be covered by `(cont_x is None) != (self.cont_x is None) or (cont_x is not None and array_equal(cont_x, self.cont_x))`, but there's no need for because nothing bad happens in that case.

janezd (approved):

Orange/widgets/unsupervised/owsom.py
VesnaT: Done.

PR #5860, opened 2022-02-25
[FIX] SOM: Fix crash when color is constant

### LINKED ISSUE #5858: SOM: IndexError when setting color with constant variable
**What's wrong?**

SOM crashes with an IndexError when selecting a variable with constant values in Color combo box

Run the following test case:
```
    def test_colored_circles_with_constant(self):
        with self.iris.unlocked():
            self.iris.X[:, 0] = 1
        self.send_signal(self.widget.Inputs.data, self.iris)
        simulate.combobox_activate_index(self.widget.controls.attr_color, 4)
```

**What's your environment?**

- Operating system: OSX
- Orange version: master

janezd:
##### Issue

Fixes #5858.

If the variable determining the color was constant, the binning set two thresholds, which were both removed, thus creating an empty list of thresholds and labels. This created problems later in the code.

##### Description of changes

Set a threshold and label.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/unsupervised/owsom.py
+++ b/Orange/widgets/unsupervised/owsom.py
@@ -833,6 +833,17 @@ class OWSOM(OWWidget):
                 binning = decimal_binnings(col, min_bins=4)[-1]
             self.thresholds = binning.thresholds[1:-1]
             self.bin_labels = (binning.labels[1:-1], binning.short_labels[1:-1])
+            if not self.bin_labels[0] and binning.labels:
+                # Nan's are already filtered out, but it doesn't hurt much
+                # to use nanmax/nanmin
+                if np.nanmin(col) == np.nanmax(col):
+                    # Handle a degenerate case with a single value
+                    # Use the second threshold (because value must be smaller),
+                    # but the first threshold as label (because that's the
+                    # actual value in the data.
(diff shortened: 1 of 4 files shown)


VesnaT:
The fix works, except when the constant is NaN:

image

Orange/widgets/unsupervised/owsom.py
```diff
@@ -210,6 +211,8 @@ class Warning(OWWidget.Warning):
         ignoring_disc_variables = Msg("SOM ignores categorical variables.")
         missing_colors = \
             Msg("Some data instances have undefined value of '{}'.")
+        no_defined_colors = \
+            Msg("'{}' has node defined values.")
```
VesnaT, line 215: A typo?

janezd: yes

VesnaT:
Similarly, the plot can't be colored by a Time Variable.

image

janezd:
> Similarly, the plot can't be colored by a Time Variable.

This problem was actually in `time_binning` and would happen in any widget using this for discretization.

Which, by the way, reminds me: all widget should use `time_binning`/`decimal_binning` for discretization. E.g. Mosaic and Sieve, as well as all widgets that show color legends.

PR #5872, opened 2022-03-05
Add GitHub URL for PyPi 

andriyor:
##### Description of changes

Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help the automation tool find the source code for Requests.

Example: [pypi.org/pypi/requests/json](https://pypi.org/pypi/requests/json)
Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,10 @@ LONG_DESCRIPTION_CONTENT_TYPE = 'text/markdown'
 AUTHOR = 'Bioinformatics Laboratory, FRI UL'
 AUTHOR_EMAIL = 'info@biolab.si'
 URL = 'http://orange.biolab.si/'
+PROJECT_URLS = {
+    'Documentation': 'https://orange3.readthedocs.io',
+    'Source': 'https://github.com/biolab/orange3',
+}
 LICENSE = 'GPLv3+'
 
 KEYWORDS = [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -57,6 +57,10 @@
 AUTHOR = 'Bioinformatics Laboratory, FRI UL'
 AUTHOR_EMAIL = 'info@biolab.si'
 URL = 'http://orange.biolab.si/'
+PROJECT_URLS = {
+    'Documentation': 'https://orange3.readthedocs.io',
```
lanzagar, line 61: Maybe the doc link should point to https://orangedatamining.com/docs
The one above might change in the future and covers only part of the documentation.

lanzagar (commented):
@andriyor thanks for spotting this!
@JakaKokosar we could think about adding the Funding and Tracker links too?
Also, the URL field above should be updated to https://orangedatamining.com

JakaKokosar:
@lanzagar I agree.

@andriyor I think we can expand this list, for example:
Homepage = https://orangedatamining.com/
Documentation = https://orangedatamining.com/docs
Source Code = https://github.com/biolab/orange3
Issue Tracker = https://github.com/biolab/orange3/issues
Donate = https://github.com/sponsors/biolab

@ajdapretnar What do you think, should we also include socials?
YouTube = https://www.youtube.com/c/OrangeDataMining
Chat = https://discord.gg/FWrfeXV

PR #5885, opened 2022-03-10
CI: Enable tests for python 3.10

PrimozGodec:
Bottleneck should now support python 3.10. Let's try if Orange works on Py3.10

Changes:
- deprecated windows-2016 are replaced with widows-2019 (and latest with windows-2022)
- python 3.10 is added to the matrix
- fixed the hash error in Lookup (in Python 10 hashes of nan are not guaranteed to be constant)
- some smaller fixes in CI to make it working

### DIFF
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -4,7 +4,7 @@ scipy>=1.3.1
 # scikit-learn version 1.0.0 includes problematic libomp 12 which breaks xgboost
 # https://github.com/scikit-learn/scikit-learn/pull/21227
 scikit-learn>=1.0.1
-bottleneck>=1.3.2
+bottleneck>=1.3.4
 # Reading Excel files
 xlrd>=0.9.2
 # Writing Excel Files
--- a/tox.ini
+++ b/tox.ini
@@ -28,8 +28,10 @@ setenv =
     COVERAGE_FILE = {toxinidir}/.coverage
     COVERAGE_RCFILE = {toxinidir}/.coveragerc
 deps =
-    pyqt5==5.12.*
-    pyqtwebengine==5.12.*
+    pyqt5==5.12.*;platform_system=="Windows" and python_version<'3.10'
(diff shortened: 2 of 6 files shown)


requirements-core.txt
```diff
@@ -4,7 +4,7 @@ scipy>=1.3.1
 # scikit-learn version 1.0.0 includes problematic libomp 12 which breaks xgboost
 # https://github.com/scikit-learn/scikit-learn/pull/21227
 scikit-learn>=1.0.1
-bottleneck>=1.3.2
+bottleneck>=1.3.4
```
PrimozGodec, line 7: bottlencek>=1.3.4 supports Python 3.10

tox.ini
```diff
@@ -28,8 +28,10 @@ setenv =
     COVERAGE_FILE = {toxinidir}/.coverage
     COVERAGE_RCFILE = {toxinidir}/.coveragerc
 deps =
-    pyqt5==5.12.*
-    pyqtwebengine==5.12.*
+    pyqt5==5.12.*;platform_system=="Windows" and python_version<'3.10'
+    pyqt5==5.15.*;platform_system!="Windows" or python_version>='3.10'
+    pyqtwebengine==5.12.*;platform_system=="Windows" and python_version<'3.10'
+    pyqtwebengine==5.15.*;platform_system!="Windows" or python_version>='3.10'
```
PrimozGodec, line 34: macOS installer user Python 3.15.*, so why not to test with pyqt5==5.15.*

tox.ini
```diff
@@ -74,8 +76,7 @@ commands_pre =
     # freeze environment
     pip freeze
 commands =
-    coverage run -m unittest -v Orange.tests Orange.widgets.tests Orange.canvas.tests
-    coverage run -m unittest discover -v Orange.canvas.tests
```
PrimozGodec, line 78: Canvas tests are already discovered in `coverage run -m unittest -v Orange.tests Orange.widgets.tests`, so it is not required to repeat this test.

PR #5888, opened 2022-03-10
Updated Group By icon

larazupan:


markotoplak:
Lara, thanks! Will you do pa PR with Agreggate too? That is the one that rendered funnily on my computer.

PR #5891, opened 2022-03-12
Indicator: Keep missing values

janezd:
Class `Indicator` converts values of a variable to 0 or 1, depending on whether it matches the given constant or not. The class hasn't checked for nan's and hence converted them to 0, which is unexpected. Missing values should remain missing.

Same for `Indicator1`.

I am not sure what this is supposed 

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/preprocess/tests/test_transformation.py
+++ b/Orange/preprocess/tests/test_transformation.py
@@ -1,10 +1,11 @@
 import unittest
 
 import numpy as np
+import scipy.sparse as sp
 
 from Orange.data import DiscreteVariable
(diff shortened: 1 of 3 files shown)


markotoplak:
@janezd, some nomogram test fails, but I have no idea how it is related.

janezd:
> @janezd, some nomogram test fails, but I have no idea how it is related.

I fear I might have. By replacing nans with zeros, one-hot encoding also served as imputation. I'll look into this. Tomorrow.

PR #5910, opened 2022-03-21
fixes in correlations widget

needNewUsername:
set openclass flag for interactions widget
set searchable flag for better usability with large datasets

### DIFF
--- a/Orange/widgets/data/owcorrelations.py
+++ b/Orange/widgets/data/owcorrelations.py
@@ -274,7 +274,7 @@ class OWCorrelations(OWWidget, openclass=True):
         box = gui.vBox(self.controlArea)
         self.correlation_combo = gui.comboBox(
             box, self, "correlation_type", items=CorrelationType.items(),
-            orientation=Qt.Horizontal, callback=self._correlation_combo_changed,
+            orientation=Qt.Horizontal, callback=self._correlation_combo_changed
         )
 
         self.feature_model = DomainModel(
--- a/Orange/widgets/data/owcorrelations.py
+++ b/Orange/widgets/data/owcorrelations.py
@@ -232,7 +232,7 @@ class CorrelationRank(VizRankDialogAttrPair):
         self.master.setStatusMessage("")
 
 
-class OWCorrelations(OWWidget):
+class OWCorrelations(OWWidget, openclass=True):
     name = "Correlations"
     description = "Compute all pairwise attribute correlations."
     icon = "icons/Correlations.svg"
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owcorrelations.py
```diff
@@ -232,7 +232,7 @@ def _set_empty_status(self):
         self.master.setStatusMessage("")
 
 
-class OWCorrelations(OWWidget):
+class OWCorrelations(OWWidget, openclass=True):
```
janezd, line 235: You mustn't do that. This flag is not here to be arbitrarily overridden.

By setting this to `True`, you promise that this widget will not make any backward-incompatible changes to its non-private methods and attributes. We never do this in "final" widgets.

(I just discovered `OWPreprocess` does it, but I believe this is the only exception. There, it's done to support preprocessing for text, I believe; the two are closely related and will be changed in tandem. But when we change Correlations, nobody will think about Interactions or want to maintain the two together.)

We have had a conglomerate of Mosaic, Sieve and perhaps a few others importing functions (that should be methods) from each other. After cleaning up that mess, no widget module does (well, at least no widget module should) import another widget module.

As it looks currently, Interactions is hacked atop of Correlations, though Correlations is not designed to be used that way.

The proper was to do this is to extract methods that are common to Correlations and your widget into another, base widget and put it into a separate file. Only then can you set the `openclass` flag (of the base widget) to `True`.

Also, think about what this base class would mean semantically. Finding a good name for it, for instance, would be a step toward it.

janezd (changes requested):

Orange/widgets/data/owcorrelations.py
needNewUsername: Thank you for pointing this out, I will remove it and only keep it in my local code as I continue working on Interactions.

ajdapretnar: @janezd Just a comment. Preprocess Text does not depend on Preprocess (the pipeline is vastly different). Perhaps it has something to do with Spectoscopy, but just a note - it should be investigated.

janezd: Thanks for this comment, it demonstrates the problem really well. All I know is that `OWPreprocess` is open and that @VesnaT opened it. Perhaps I knew why and the reasoning was sound and I agreed with it ... but as things stand now, we don't know who relies on it and what will fail if we decide to change something.

... two minutes later: I discovered I merged this, here https://github.com/biolab/orange3/pull/4521. We had good reasons, but I regret it, we should have refactored it. SCOrange can explode at any change of Preprocess without anybody noticing.

janezd: Emm, sorry, single cell.

PR #5913, opened 2022-03-22
[FIX] Orange.data.Table: deprecated is_view and is_copy

### LINKED ISSUE #5878: Orange.data.Table's is_view and is_copy do not work for sparse matrices: deprecate them
`Table.is_view` and `Table.is_copy` do not work for sparse matrices.

I do not think they are that useful because they are a conjunction of so many conditions. I think we should deprecate them instead of fixing them. PyCharm only finds one usage in the Orange3 repo, and even that in the tests.

So, I would deprecate and remove in Orange 3.33.

needNewUsername:
##### Issue
fixes #5878 

##### Description of changes
deprecated is_view and is_copy as requested

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/tests/test_table.py
+++ b/Orange/data/tests/test_table.py
@@ -1,10 +1,12 @@
 import pickle
 import unittest
 import os
+from distutils.version import LooseVersion
 
 import numpy as np
 import scipy.sparse as sp
 
+import Orange
 from Orange.data import (
     ContinuousVariable, DiscreteVariable, StringVariable,
     Domain, Table, IsDefined, FilterContinuous, Values, FilterString,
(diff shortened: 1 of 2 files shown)


Orange/data/tests/test_table.py
```diff
@@ -380,8 +380,10 @@ def test_unpickled_owns_data(self):
             Table.LOCKING = default
 
         unpickled = pickle.loads(pickle.dumps(table))
-        self.assertFalse(unpickled.is_view())
-        self.assertTrue(unpickled.is_copy())
+        self.assertFalse(unpickled.is_sparse())
+        self.assertTrue(unpickled.X.shape[-1] or unpickled.X.base is None)
```
markotoplak, line 384: What is the reason that you are testing additional properties for X? If this is indeed needed, please make a comment in the code that explains the reasoning.

needNewUsername: Just to stay consistent with what is_view and is_copy returned. I assumed there was a reason I just couldn't think of one.
If there is no reason and all we need is to check if `unpickled` actually owns it's data then as far as I can tell it would be enough to only check all `base` attributes.

markotoplak: Yes, you should probably only do the latter...

markotoplak:
Thanks! Could you now rebase into a single commit?

PR #5920, opened 2022-03-28
server_embedder - replace run_until_complete with asyncio.run

PrimozGodec:
##### Issue

Since Orange now support python>=3.7 we do not need to use `run_until_complete` and custom loop handling in embedders.

##### Description of changes
User `asyncio.run` instead which is supported python>=3.7

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/misc/server_embedder.py
+++ b/Orange/misc/server_embedder.py
@@ -123,16 +123,9 @@ class ServerEmbedderCommunicator:
         # if there is less items than 10 connection error should be raised earlier
         self.max_errors = min(len(data) * self.MAX_REPEATS, 10)
 
-        loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(loop)
-        try:
-            embeddings = asyncio.get_event_loop().run_until_complete(
-                self.embedd_batch(data, processed_callback, callback=callback)
-            )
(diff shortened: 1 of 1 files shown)


Orange/misc/server_embedder.py
```diff
@@ -123,16 +123,9 @@ def embedd_data(
...
-        loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(loop)
-        try:
-            embeddings = asyncio.get_event_loop().run_until_complete(
-                self.embedd_batch(data, processed_callback, callback=callback)
-            )
-        finally:
-            loop.close()
-
-        return embeddings
```
janezd, line 135: To anyone interested -- while irrelevant for this PR, it may be useful in the future: `finally` is always executed, thus this code could have `return` statement within `try`. 

Not only is `finally` executed even when the `try` block contains `return`, a `return` in `finally` block even "overrides" a `return` in `try`, as shown in [this example](https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions) from Python documentation:

```python
def bool_return():
    try:
        return True
    finally:
        return False
```

This function returns `False`.

PR #5932, opened 2022-04-01
owsavebase: fix path comparison under windows

### LINKED ISSUE #5868: Save Data: autosave is deactivated
**What's wrong?**

The Autosave-Checkbox in Save Data deactivates itself at every restart of Orange, though workflow and target path are in the same directory.

I think it is a Windows-Issue:

Here:
https://github.com/biolab/orange3/blob/5f06ada600eac42138e362bba566824e69e7cd3a/Orange/widgets/utils/save/owsavebase.py#L116-L120

workflow_dir is a path with backslashes: "C:\\\\Users\\\\test\\\\Desktop\\\\TestDir"
absolute_path is a path with slashes: "C:/Users/test/Desktop/TestDir"
So `absolute_path.startswith(workflow_dir.rstrip("/"))` will always be false and the absolute path is written to `self.stored_path`.

**What's your environment?**

- Orange version: 3.32.0 dev
- How you installed Orange: clone from github

mw25:
##### Issue

Fixes #5868

##### Description of changes
I wanted to use `pathlib` to check if the `absolute_path` is under the `workflow_dir`.
Because the comparison via `startswith()` is error prone:
absolute_path: /some/folder_other/file.csv
workflow_dir: /some/folder 

Nevertheless `_absolute_path` and `stored_path` should be stored in the same format as before.
This made the whole thing a bit uglier.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/save/owsavebase.py
+++ b/Orange/widgets/utils/save/owsavebase.py
@@ -1,6 +1,7 @@
 import os.path
 import sys
 import re
+from pathlib import Path
 
 from AnyQt.QtWidgets import QFileDialog, QGridLayout, QMessageBox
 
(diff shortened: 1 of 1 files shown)


thocevar:
This doesn't solve the referenced issue for me. I'll investigate further.

thocevar:
Hm, I don't know what went wrong previously. Maybe some old stored settings on my end. Anyway, #5981 should solve this where I've also added some tests.
Thanks again for reporting and fixing this.

PR #5949, opened 2022-04-20
[ENH] Feature Constructor: Vectorize when possible

### LINKED ISSUE #5911: Performance Issue with Feature Constructor
**What's wrong?**

It's not about a bug, rather a performance issue.

I have already noticed that Feature Constructor takes a little longer with more data.
But now I want to use the construct_variables function in my own widget as well, so I took a closer look:

I analyzed my minimal example (see below) with py-spy and saw that the function `__call__` of FeatureFunc is the bottleneck. Especially collecting the data line by line, it seems to me.
I could imagine saving a lot of time by first reading out the data column-by-column (and then performing the operation row-by-row again, if necessary). 
Of course, I don't know what impact that would have on the further logic and design of the functions in the widget....

Adding two 100,000 long variables took a little less than 2 seconds in my case.

What do you think? Could you imagine to improve the performance?

```
    def test_performance(self):
        from codetiming import Timer
        timer = Timer()

        var_names = ['Var' + str(i) for i in range(100)]
        variables = [ContinuousVariable(name=var_name) for var_name in var_names]
        domain = Domain(variables)
        test_data = np.random.rand(100_000, len(variables))
        table = Table.from_numpy(domain, test_data)

        expr = 'Var0 + Var1'
        desc = ContinuousDescriptor(expr, expr, number_of_decimals=None)
        constr_vars = construct_variables([desc], table)
        new_domain = Domain(table.domain.attributes + tuple(constr_vars))

        timer.start()
        new_table = table.transform(new_domain)
        timer.stop()
```

Console Output: 
`Elapsed time: 1.8722 seconds`

The result of py-spy:

image

**What's your environment?**

- Orange version: '3.32.0.dev0+2b857a3'
- How you installed Orange: git clone

janezd:
##### Issue

Resolves #5911.

##### Description of changes

- Try to compute the function on columns; fall back to per-instance if this fails.
- Import functions from `np` instead of from `math` when available.
- Show all functions in combo (avoid compiling a separate list of functions)
- Remove `cumsum`, `cumprod`, which created two-dimensional data instead of a column

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -15,6 +15,7 @@ import logging
 import ast
 import types
 import unicodedata
+import warnings
 
 from traceback import format_exception_only
 from collections import namedtuple, OrderedDict
@@ -108,7 +109,10 @@ Categorical features are passed as strings
 
 """.lstrip()
 
-    FUNCTIONS = dict(chain([(key, val) for key, val in math.__dict__.items()
+    # import functions from math - but if the function with the same name
+    # exists in np, import it from there
+    FUNCTIONS = dict(chain([(key, np.__dict__.get(key, val))
+                            for key, val in math.__dict__.items()
                             if not key.startswith("_")],
                            [(key, val) for key, val in builtins.__dict__.items()
                             if key in {"str", "float", "int", "len",
@@ -999,7 +1003,9 @@ def bind_variable(descriptor, env, data, use_values):
             values = {name: i for i, name in enumerate(values)}
             descriptor = descriptor._replace(values=values)
 
-            def cast(x):  # pylint: disable=function-redefined
+            def cast(x): # pylint: disable=function-redefined
+                if isinstance(x, np.ndarray):
+                    return np.array([values.get(y, nan) for y in x])
                 return values.get(x, nan)
 
         else:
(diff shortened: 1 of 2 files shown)


ajdapretnar:
This works for my limited amount of use cases.

markotoplak:
I was looking at math vs numpy docs and did not see any cases where I would predict conflicts. That does not mean they aren't any.

@janezd also raised the question about import everything from `math`. I would rather name a subset of functions, perhaps those that are already available in the python 3.10, so that we do not break any current workflows. Imports should be "soft" since not all of them are available on oldest supported python.

janezd:
@markotoplak, I checked the functions in `math` and all seem useful, so I'd keep it the selection it is.

As for reporting warnings and errors in expressions: the problem here is that the user can write an expression that cannot be evaluated by numpy, in which case we fall back to per-instance evaluation. To be on the safe side, I'd fall back at any sign of trouble, that is, at any warning or error. When this happen, the execution will be slower (that is: as slow as before this PR) and any errors will be properly reported.

ales-erjavec:
The random functions (gauss, normalvariate, ...) no longer work on per instance basis. Single generated value is used for the whole column (e.g. `sepal_length + gauss(0, 1)` or just `gauss(0, 1)`)

janezd:
Closed in favor of #5975.

PR #5957, opened 2022-04-27
Predictions: Show errors in regression

### LINKED ISSUE #5901: Show error by color in Predictions
**What's your use case?**

I want to observe regression predictions in Predictions widget. Currently, I only see numbers. It would be nice to have a visual signal in Predictions for how close the prediction is to the true value.

Perhaps give each instance a line below, showing the error. Currently, classification problems display a line underneath with probabilities. For regression, the longer the line, the greater the error (would this be confusing, because for classification longer lines mean higher probability?).

/

janezd:
##### Issue

Fixes #5901

##### Description of changes

- Refactor item delegates into two classes, one for classification and one for regression (to avoid checks in methods)
- Add information about the actual class/outcome to `PredictionModel`
- Show actual class by using narrower lines in distribution bar
- Show the line for regression and mark the true value
- **Fix** a bug in assigning of indices for sorting by probabilities: if any models failed and their respective columns were not shows, the indices for sorting by that column were nevertheless created and inserted. Therefore, if the 3rd model has failed, the 3rd column showed the 4th model, but sorting by that column would sort by the 3rd model (and probably fail).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -1026,7 +1026,7 @@ class RegressionItemDelegate(PredictionsItemDelegate):
             self.span = 1
         else:
             self.offset = minv
-            self.span = maxv - minv
+            self.span = maxv - minv or 1
 
     def initStyleOption(self, option, index):
         super().initStyleOption(option, index)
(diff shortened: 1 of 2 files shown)


janezd:
Like this? Alternative suggestions can be made in form of patches for `RegressionItemDelegate.drawBar` (https://github.com/biolab/orange3/pull/5957/files#diff-dd77f983a656d8fa3f453496f08aad0ae7e8c5a57e48406d5520836c76ce1690R1022)

Screen Shot 2022-04-27 at 21 31 27

ajdapretnar:
It took me a moment or two to realize what the lines show, so is my understanding correct: the blue line shows the predicted value until true value, the circle is the true value, the red line shows if the predicted value goes over the true value?

I like it. One can visually quite quickly inspect which are the "underpredicted" and which the "overpredicted" instances, which was the point of the issue report.

VesnaT:
There is a problem, when removing a class variable:
image
In the above workflow, go to Select Columns and move the Target to Ignored.

janezd:
@ajdapretnar, @VesnaT complained that predicting larger values is indicated in red and thus looks as a "more serious error" than undershooting. We propose the design below. Complaints?

Screen Shot 2022-05-06 at 16 46 35

Thick line indicates predictions, red color indicates the error. One's eyes can easily decide to observe one or the other (that is, concentrat on bars and ignore tiny red dots, or observe only the red parts disregarding its shape (bar or line)).

janezd:
> There is a problem, when removing a class variable:

I fixed and tested that -- and another related problem, which occurred if target class mismatched (resulting in all bars having the same length).

ajdapretnar:
I am, honestly, quite confused. Is this just because red is perceived as "error" and blue as "not an error"? Then you can just switch to another color, if that's the problem. With the current solution it took me a long time to understand what the blue line means, the red dashed line (why is it red?) and the full red line. I don't think it makes visual inspection easy.

I was thinking more in line of having all the true values aligned (are we really interested in the exact value since it is written in the next column on the right?). Then there is a line going left for underestimated and right for overestimated values. We can go with pink if red is too misleading.

Screenshot 2022-05-12 at 15 28 12

janezd:
- Remove red color and dotted line
- Add a checkbox to show/hide error; for regression it can show the bars proposed by @ajdapretnar 
- For regression, error can also be relative
- Also, for categorical data show error as 1 - probability of correct class
- Sorting by error

PR #5962, opened 2022-05-06
Silhouette: Show silhouette score instead of group size

### LINKED ISSUE #5938: Silhouette plot should show statistics for groups
**What's your use case?**

A student who analyzed some data about habits and preferences of students wrote something to the effect of "all men are alike, while women differ from each other". This could be verified by Silhouette if it showed silhouette scores for each group.

Given the widget's current layout, the simplest place to show this would be in control area. The problem is that this cannot be copied, so don't forget to include this info in report.

**Alternative solutions?**

Observe the silhouette as it this. It would seem that the opposite is true, though the difference is small.

janezd:
##### Issue

Resolves #5938.

##### Description of changes

Group sizes are useful, but shown by other widgets. Silhouette plot is about silhouette scores.

##### Includes
- [X] Code changes


PrimozGodec:
I slightly updated the documentation. First I just planed to update such that new change would be present there but then I found out if I replace the image I also need to update text.

janezd:
Thanks, Primož!

PR #5964, opened 2022-05-07
Distributions: Remember number of bins

### LINKED ISSUE #5924: Distributions: changing bin width
**What's wrong?**

I'll call a bug on this one.
When I open Distributions for the first time, Age attribute is selected and it has three bins. Then I change to a different variable and back to Age. Now Age has 10 bins.

https://user-images.githubusercontent.com/12524972/160869888-55876278-06a9-4610-b779-bd009a41a9a2.mov

Datasets (Employee Attrition) - Distributions.
Switch to some other variable, then back to Age.

**What's your environment?**

- Operating system: OSX
- Orange version: 3.32.dev

janezd:
##### Issue

Fixes #5924.

When changing variable, the widget uses the stored number of bins for that variable (not stored in settings, just temporarily). If the variable is shown for the first time, it just keeps the same number of bins as shown currently.

Number of bins was stored for each variable only when the user releases the slider. This thus didn't save the initial setting.

The problem occurred when 
- the widget was showing a numeric variable (the widget initially chooses a sensible number of bins),
- user **does not** move the slider (hence the number of bins is not stored),
- user switches to categorical variable (bins are not used, but the number of bins is reset to maximum due to implementational details)
- user switch back to numeric variable: the number of bins for that variable was not set, hence keeping the current setting, which is the maximum.

##### Description of changes

Store the setting when changing the variable.

##### Includes
- [X] Code changes
- No tests. I could write a test to check for this specific bug, but it's unlikely to reoccur.

### DIFF
--- a/Orange/widgets/visualize/owdistributions.py
+++ b/Orange/widgets/visualize/owdistributions.py
@@ -903,6 +903,7 @@ class OWDistributions(OWWidget):
         self.controls.number_of_bins.setMaximum(max_bins)
         self.number_of_bins = min(
             max_bins, self._user_var_bins.get(self.var, self.number_of_bins))
+        self._user_var_bins[self.var] = self.number_of_bins
         self._set_bin_width_slider_label()
 
     @staticmethod


janezd:
@ajdapretnar, I think this should fix it. I can no longer replicate the problem. Can you check, please?

PR #5966, opened 2022-05-08
Violin Plot: Add grid

### LINKED ISSUE #5929: Violin Plot: add grids for easier interpretation
**What's your use case?**

Looking at Violin Plot it is sometimes hard to know precisely the min and the max of the data points.

Perhaps we could add a grid that would simplify interpretation of the results? Similar to the grid in Scatter Plot.

Hover on data points.

janezd:
##### Issue

Resolves #5929.

##### Description of changes

Add checkbox and grid.

**To do:**

- [ ] tests, of course
- [ ] arrange check boxes into two columns?

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owviolinplot.py
+++ b/Orange/widgets/visualize/owviolinplot.py
@@ -376,7 +376,8 @@ class ViolinPlot(PlotWidget):
 
     def __init__(self, parent: OWWidget, kernel: str, scale: int,
                  orientation: Qt.Orientations, show_box_plot: bool,
-                 show_strip_plot: bool, show_rug_plot: bool, sort_items: bool):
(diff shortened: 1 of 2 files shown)


janezd:
@VesnaT, here's a puzzle for you.

- Try changing orientation back and forth. Everything works as before.
- Show grid and now try changing orientation back and forth. Still works.
- Hide grid and change orientation -> segfault. Something about calling pure virtual function.

This is (at least on my machine) perfectly reproducible. All I need to do is open widget, show grid, hide grid, change orientation. Bang. Why?!

I discovered that if I comment out the call to `self._clear_axes()` in `set_orientation`, this no longer happens, but also grid no longer show properly. It is certainly related to axes, somehow.

(I have pyqtgraph 0.12.3)

PR #5967, opened 2022-05-08
Scatterplot VizRank: Set random_state for ReliefF

### LINKED ISSUE #5900: Find informative projections: Keep reproducible order of ranked variable pairs
**What's wrong?**

Running Find informative projections sometimes returns different order of variable pairs. This happens when two or more pairs have the same score and they are ranked randomly. This is a little confusing when the order on two machines is not the same.

Ideally, we would have reproducible ranking all the time. The easy way is to order variable pairs with same scores alphabetically.

Datasets (Auto MPG) - Scatter Plot. Find informative projections.
Screenshot 2022-03-17 at 08 55 46
Screenshot 2022-03-17 at 08 55 50

**What's your environment?**

- Operating system: OSX
- Orange version: 3.32.dev

janezd:
##### Issue

Fixes #5900.

##### Description of changes

Set random state for ReliefF.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -94,7 +94,8 @@ class ScatterPlotVizRank(VizRankDialogAttrPair):
         data = self.master.data.transform(domain)
         relief = ReliefF if isinstance(domain.class_var, DiscreteVariable) \
             else RReliefF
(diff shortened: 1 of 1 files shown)


janezd:
@ajdapretnar, can you check whether you can still reproduce this?

PR #5988, opened 2022-05-23
Nomogram: Elide long variable names

### LINKED ISSUE #5865: Nomogram: Elide long variable names
**What's wrong?**

Nomogram is a bit problematic with long variable names. When resizing, variable names don't get resized, but the scales do. If variables names are long, this would be problematic. I propose to elide long variable names and show full names in tooltips instead.

This example shows and extreme case where variable names are long and widget is quite resized. Variable names being to overlap with scales and the below probabilities shrink to zero, because they size of the text on the left depends on variable names.
Screen Shot 2022-03-03 at 20 29 21

Datasets (HDI), predict continent (or Cluster if easier). Naive Bayes - Nomogram.

**What's your environment?**

- Operating system: OSX
- Orange version: 3.32.dev

VesnaT:
##### Issue

Fixes #5865

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/ownomogram.py
+++ b/Orange/widgets/visualize/ownomogram.py
@@ -14,7 +14,7 @@ from AnyQt.QtWidgets import (
     QGraphicsEllipseItem, QGraphicsLinearLayout, QGridLayout, QLabel, QFrame,
     QSizePolicy
 )
-from AnyQt.QtGui import QColor, QPainter, QFont, QPen, QBrush
+from AnyQt.QtGui import QColor, QPainter, QFont, QPen, QBrush, QFontMetrics
 from AnyQt.QtCore import Qt, QRectF, QSize, QPropertyAnimation, QObject, \
     pyqtProperty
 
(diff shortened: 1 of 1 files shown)


Orange/widgets/visualize/ownomogram.py
```diff
@@ -949,11 +949,18 @@ def update_scene(self):
         attr_inds, attributes = zip(*self.get_ordered_attributes()[:n_attrs])
         self.Outputs.features.send(AttributeList(attributes))
 
-        name_items = [QGraphicsTextItem(attr.name) for attr in attributes]
         point_text = QGraphicsTextItem("Points")
+        metric = QFontMetrics(point_text.font())
+
+        def text_item(text):
+            elided_text = metric.elidedText(text, Qt.ElideRight, 200)
+            return QGraphicsTextItem(elided_text)
```
janezd, line 957: A tooltip would be useful here, like

```python
        def text_item(text):
            elided_text = metric.elidedText(text, Qt.ElideRight, 200)
            item = QGraphicsTextItem(elided_text)
            item.setToolTip(text)
            return item
```

VesnaT: added

PR #5995, opened 2022-05-27
Create Instance: Unique source variable name

### LINKED ISSUE #5994: Create Instances duplicates attribute names
**What's wrong?**

Variable "Source Id" is added even if the attribute with the same name already exists.

Put two Create Instances widgets one after another.

**What's your environment?**

N/A

VesnaT:
##### Issue

Fixes #5994

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcreateinstance.py
+++ b/Orange/widgets/data/owcreateinstance.py
@@ -16,6 +16,7 @@ from orangewidget.gui import Slider
 
 from Orange.data import DiscreteVariable, ContinuousVariable, \
     TimeVariable, Table, StringVariable, Variable, Domain
+from Orange.data.util import get_unique_names
 from Orange.widgets import gui
 from Orange.widgets.utils.itemmodels import TableModel
 from Orange.widgets.settings import Setting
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owcreateinstance.py
```diff
@@ -646,12 +647,34 @@ def _create_data_from_values(self) -> Table:
...
+    def _append_to_data(self, instance: Table) -> Table:
         assert self.data
-        assert len(data) == 1
+        assert len(instance) == 1
+        source_label = "__source_widget"
 
-        var = DiscreteVariable("Source ID", values=(self.data.name, data.name))
-        data = Table.concatenate([self.data, data], axis=0)
+        data = Table.concatenate([self.data, instance], axis=0)
+        for var in self.data.domain.variables + self.data.domain.metas:
```
janezd, line 656: What about replacing all those `if`-s with `get_column_view`? I think (and also tried) that this should work:

```python
        for var in chain(self.data.domain.variables, self.data.domain.metas):
            if var.attributes.get(source_label) == OWCreateInstance:
                with data.unlocked():
                    data.get_column_view(var)[0][-1] = 1
                return data
```

VesnaT: done

PR #6000, opened 2022-06-02
OWHeatMap: don't show hidden attributes

### LINKED ISSUE #5306: Heat Map: don't show bow attributes if flag hidden=True

Some visualization widgets will ignore bow attributes if they have a hidden=True flag (in attributes of attributes). Heat Map would benefit from the same setup: ignoring bow attributes (and showing just sentiment, topics, etc.).

Add the same option as Box Plot has it for hiding bow attributes.

Keep it as is.

ajdapretnar:
##### Issue

Fixes #5306.

##### Description of changes
Hide hidden attributes.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owheatmap.py
+++ b/Orange/widgets/visualize/owheatmap.py
@@ -229,8 +229,8 @@ class OWHeatMap(widget.OWWidget):
         #: The original data with all features (retained to
         #: preserve the domain on the output)
         self.input_data = None
-        #: The effective data striped of discrete features, and often
-        #: merged using k-means
+        #: The effective data stripped of discrete features and hidden
+        #: attributes, and often merged using k-means
         self.data = None
         self.effective_data = None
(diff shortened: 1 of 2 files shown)


ajdapretnar:
This was more of a last minute PR, so I am happy to redo the tests.

Orange/widgets/visualize/owheatmap.py
```diff
@@ -618,6 +618,17 @@ def set_dataset(self, data=None):
                 self.Information.discrete_ignored(
                     ndisc, "s" if ndisc > 1 else "")
 
+        if data is not None and any(var.attributes.get('hidden', False)
```
janezd, line 621: A block of code above this patch exclude discrete variables. Couldn't you just simply add condition there, so it would exclude discrete and hidden variables?

PR #6012, opened 2022-06-09
[ENH] Predictions: Show errors

### LINKED ISSUE #5971: Show errors in Predictions
**What's your use case?**

#5957 ended with adding a bar delegate to predicted values. As discussion about that feature has shown, visualizing an error is a separate issue, requiring a separate column.

 We (probably) want to:

- [x] Add a checkbox to show/hide error; for regression it can show the bars proposed by @ajdapretnar in #5957
- [x] For regression, error can also be relative; think about how to incorporate this option without burdening the GUI (which just got rid of control area not long ago)
- [x] Also do this for categorical data: show error as `1 - probability of assigned to the actual class`
- [x] The user should of course be able to sort by that column as well.

Implementing all this may not be so simple because of the way `PredictionModel` in this widget works.

Errors can be computed in Feature Constructor and explored in other widgets.

This is not the best alternative because Feature Constructor is not the simplest widget to use.

janezd:
##### Issue

Closes #5971.

##### Description of changes

Add columns with errors for each predictor.

Please comment on functionality and interface. Checking and tests will come afterwards.

Try it on housing and heart disease; errors on iris are too small.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -1,3 +1,4 @@
+import math
 import warnings
 from contextlib import contextmanager
 from functools import partial
@@ -8,7 +9,7 @@ from typing import Set, Sequence, Union, Optional, List, NamedTuple
 import numpy
 from AnyQt.QtWidgets import (
     QTableView, QSplitter, QToolTip, QStyle, QApplication, QSizePolicy,
-    QPushButton)
+    QPushButton, QAbstractItemDelegate)
 from AnyQt.QtGui import QPainter, QStandardItem, QPen, QColor, QBrush
 from AnyQt.QtCore import (
     Qt, QSize, QRect, QRectF, QPoint, QPointF, QLocale,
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I fully approve this functionality! It enables selecting the most underpredicted, overpredicted and best predicted instances and showing them in the scatter plot. I also like the many options for error computation, discretely placed at the top.

Orange/widgets/evaluate/owpredictions.py
```diff
@@ -1094,11 +1260,40 @@ def rowCount(self, parent=QModelIndex()):
...
+            if error_column:
+                if self._actual is None:
+                    return None
+                actual = self._actual[row]
+                if numpy.isnan(actual):
+                    return None
+                elif self._probs[column].size:
+                    return 1 - self._probs[column][row, int(actual)]
+                else:
+                    diff = self._values[column // 2][row] - actual
```
lanzagar, line 1281: `column` was already divided by 2 above (line 1271)!

I must say, there is quite a lot of `%2==1` and `//2` in this file... :)

janezd: > column was already divided by 2 above (line 1271)!

It is disturbing that (seemingly thorough) tests didn't catch this because they used only a single predictor... :(

I fixed this and modified the tests to use two predictors.

> I must say, there is quite a lot of %2==1 and //2 in this file... :)

Apparently (at least) one too many. :)

PR #6021, opened 2022-06-13
[ENH] Feature Constructor: Option to put new variable into metas

### LINKED ISSUE #5635: Feature Constructor: allow creating meta variables directly
**What's your use case?**
Sometimes a user wants to construct a numeric feature that is not used for modeling but only for display. I needed that feature quite a few times.

A user could choose where to put the feature (attributes, metas, classes) in the Feature constructor. That would require an additional combo box.

Currently, a Select Columns can be used after Feature Constructor to move the feature.

janezd:
##### Issue

Fixes #5635.

##### Description of changes

In #5635 I wrote about putting the checkbox into table. I disagree with my ex-self: editor for modifying the feature is at the top, not in the table. If we do change the widget to have controls in the table (including all combos?!), then we'd put the checkbox there, too.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -26,9 +26,10 @@ from typing import List, Dict, Any, Mapping, Optional
 import numpy as np
 
 from AnyQt.QtWidgets import (
-    QSizePolicy, QAbstractItemView, QComboBox, QFormLayout, QLineEdit,
+    QSizePolicy, QAbstractItemView, QComboBox, QLineEdit,
     QHBoxLayout, QVBoxLayout, QStackedWidget, QStyledItemDelegate,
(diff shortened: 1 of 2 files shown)


markotoplak:
@janezd, thanks, this looks and works well. I see also that for strings the meta checkbox is always enabled, which is also a helpful user indicator (before they might have been confused by the position).

In #5635 we decided against constructing classes. I have some minor doubts: I still think the class construction would be helpful, especially if we have a data set where we'd like to make a class from existing meta columns. But even being able to make only metas that is already a big improvement.

markotoplak:
I edited this for "forward" compatibility (so that settings saved with master can be opened with this version). This was enabled by moving the new `meta` descriptor to the end of Descriptors and giving it a default value. @janezd, please check if my changes are OK, and perhaps rebase and merge.

PR #6022, opened 2022-06-13
Apply Domain and Select By Index: Replace info by description

### LINKED ISSUE #5574: Select Data By Index: miserably looking widget
**What's your use case?**

It's not only that it looks miserable: the new tooltip (window at the bottom right) already shows more information than what's in the widget. The tooltip for the output is even larger.

Screenshot 2021-08-28 at 20 13 11

None, but it'd be good to think what to put into this widget instead. Given its mysteriousness, there could simply be a short label explaining what the widget does, to inform the user who is randomly exploring the widgets.

janezd:
##### Issue

Fixes #5574.

##### Description of changes

This is very unusual and I'd hate it to become the norm. But there's nothing reasonable we can put there. Even five-row preview is pointless after #5956.

On the other hand, these descriptions may indeed be useful.

Screen Shot 2022-06-13 at 17 28 31

Screen Shot 2022-06-13 at 17 28 42

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/data/owselectbydataindex.py
+++ b/Orange/widgets/data/owselectbydataindex.py
@@ -45,34 +45,29 @@ class OWSelectByDataIndex(widget.OWWidget):
         self.extra_model_unique = itemmodels.VariableListModel()
         self.extra_model_unique_with_id = itemmodels.VariableListModel()
 
-        box = gui.hBox(self.controlArea, box=None)
-        self.infoBoxData = gui.label(
-            box, self, self.data_info_text(None), box="Data")
-        self.infoBoxExtraData = gui.label(
-            box, self, self.data_info_text(None), box="Data Subset")
+        box = gui.widgetBox(self.controlArea, True)
+        gui.label(
+            box, self, """
+Data rows keep their identity even when some or all original variables
+are replaced by variables computed from the original ones.
+
+This widget gets two data tables ("Data" and "Data Subset") that
+can be traced back to the same source. It selects all rows from Data
+that appear in Data Subset, based on row identity and not actual data.
+""".strip(), box=True)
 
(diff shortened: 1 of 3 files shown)


Orange/widgets/data/owtransform.py
```diff
@@ -36,42 +36,21 @@ def __init__(self):
...
-                len(self.template_data.domain.attributes))
-        self.template_label.setText(text)
-
-    def set_output_label_text(self, data):
-        text = ""
-        if data:
-            text = "Output data includes {:,} features.".format(
-                len(data.domain.attributes))
-        self.output_label.setText(text)
+This includes selecting a subset of variables as well as
```
markotoplak, line 45: These include column selection as well as transformations, like ...

Orange/widgets/data/owselectbydataindex.py
```diff
@@ -45,35 +45,31 @@ def __init__(self):
...
 
-        box = gui.hBox(self.controlArea, box=None)
-        self.infoBoxData = gui.label(
-            box, self, self.data_info_text(None), box="Data")
-        self.infoBoxExtraData = gui.label(
-            box, self, self.data_info_text(None), box="Data Subset")
+        box = gui.widgetBox(self.controlArea, True)
+        gui.label(
+            box, self, """
+Even in a situation in which some or all original variables are
```
markotoplak, line 51: I'd turn it around: Data rows keep their identity even when all original variables ...

markotoplak:
I like this a lot. I proposed some text changes but I am not really sure if they are any better. Feel free to merge as-is.

PrimozGodec:
It looks OK to me. Like @markotoplak sugestions.

PR #6050, opened 2022-07-01
Fix test_fss for numpy 1.23

markotoplak:
##### Issue
`np.ndarray` (which `contingency.Discrete` subclasses) of a single float is not an instance of float anymore.

### DIFF
--- a/Orange/tests/test_fss.py
+++ b/Orange/tests/test_fss.py
@@ -3,7 +3,9 @@
 
 import unittest
 
-from Orange.data import Table, Variable
+import numpy as np
+
+from Orange.data import Table
 from Orange.preprocess.score import ANOVA, Gini, UnivariateLinearRegression, \
     Chi2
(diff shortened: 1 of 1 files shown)


markotoplak:
This fixes some recent test failures.

The outstanding ones are now non-deterministic Nomogram fails and pyqtgraph/pyqtgraph#2352

janezd:
> The outstanding ones are now non-deterministic Nomogram fails and [pyqtgraph/pyqtgraph#2352](https://github.com/pyqtgraph/pyqtgraph/issues/2352)

For nomograms, I started investigating in #6018; it now waits for @VesnaT.

Pyqtgraph with its backward incompatibilities is becoming a liability. Really annoying. I understand that *we* don't care so much about keeping Orange's code backward compatible, but Orange isn't meant to be a dependency, while pyqtgraph is and its developers should care.

PR #6056, opened 2022-07-10
Aggregate Columns: Add additional options for selection

### LINKED ISSUE #5643: Aggregate Columns default selection
**What's your use case?**
I am using this widget to calculate signature scores by choosing different sets of genes. I want to do this more interactively but currently, I can't as I need to manually select genes to be included in the aggregation. I am doing the selection in Gene Sets widget and I would like for the Aggregate Columns widget to use all columns if none are selected. Also, I would exclude meta fields from the list as they are normally excluded from computations in Orange.

janezd:
##### Issue

Fixes #5643.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owaggregatecolumns.py
+++ b/Orange/widgets/data/owaggregatecolumns.py
@@ -1,18 +1,28 @@
-from typing import List
+from itertools import chain
+from typing import List, NamedTuple, Callable
 
 import numpy as np
 
-from AnyQt.QtWidgets import QSizePolicy
+from AnyQt.QtWidgets import QSizePolicy, QStyle, \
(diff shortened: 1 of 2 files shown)


markotoplak:
Seems fine. There is some seemingly inconsistent  behaviour with the Features input:
1. with 2 numerical features, the widget goes into the features input mode and disables the rest.
2. with 1 numerical and 1 categorical feature, the widget goes into the features input mode, disables the rest, and shows a warning,
3. with 2 categorical features, the widget DOES NOT go into the features input mode, but still shows a warning.

With (3) I'd expect that the widget goes into the features input mode (because of the input), and an error was shown.

PR #6058, opened 2022-07-12
[ENH] Table: Add methods get_column and set_column

janezd:
##### Issue

There are two differences between `data.get_column_view(x)[0]` and `data.transform(Domain([x]).X[:, 0]`.

1. The former is simple and short.
2. The latter works. 

Screen Shot 2022-07-12 at 11 55 39

If using Text to Columns used `get_column_view`, Apply Domain could not apply the second transformation because the required attribute (discretized petal length) is not in the domain.

Using `transform` works because it invokes `compute_value`.

##### Description of changes

Add `Table.get_column` which should be preferred over `Table.get_column_view`, and *must* be used in places where the variable is not guaranteed to exist (in particular, in `compute_value` functions).

If the column exists, it calls `get_column_view`, otherwise it uses `compute_value`.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1436,7 +1436,7 @@ class Table(Sequence, Storage):
         domain = Domain(attrs, classes, metavars)
         new_table = self.transform(domain)
         with new_table.unlocked(new_table.metas if to_metas else new_table.X):
-            new_table.get_column(variable, view=True)[:] = data
+            new_table.set_column(variable, data)
         return new_table
 
     def is_sparse(self):
@@ -1565,7 +1565,7 @@ class Table(Sequence, Storage):
         self.W = self.W[ind]
         self.ids = self.ids[ind]
 
-    @deprecated("Table.get_column(..., view=True)")
+    @deprecated("Table.get_column (or Table.set_column if you must)")
     def get_column_view(self, index: Union[Integral, Variable]) -> np.ndarray:
         """
         An obsolete function that was supposed to return a view with a column
@@ -1613,7 +1613,7 @@ class Table(Sequence, Storage):
         else:
             return self.metas[:, -1 - index]
 
-    def get_column(self, index, copy=False, view=False):
+    def get_column(self, index, copy=False):
         """
         Return a column with values of `index`.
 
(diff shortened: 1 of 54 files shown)


ajdapretnar:
Would it be possible to have something similar to pandas' `data["sepal length"]`? This would not require to name the method, but it would likely require too much messing with the existing Table structure.

markotoplak:
@ajdapretnar, that would be more messing with the data structure. :) Especially in this case because the actual variable (with its compute value) is important, not the name.

Now I think `get_column` is fine, especially since then programmers who do not completely understand what is going on will find it first. The function you wrote is still as fast as `get_column_view` when the latter would be appropriate, so I see no loss.

I'd deprecate `get_column_view`.

Slightly related: do we ever enforce what shape functions doing `compute_value` return? I remember we have some code that reshapes `compute_value` outputs within table transforms, but that might be something stale? Is the shape of the output array always the same?

janezd:
You're right, the result of `compute_value` should be reshaped. `compute_value` is currently called in domain conversion, which (for dense arrays) goes through `assure_column_dense`, which calls `reshape(-1)`.

What about sparse data?

We could also request `compute_value` to return either a vector or a sparse matrix. For a few years, returning anything else would trigger a warning.

> I'd deprecate `get_column_view`.

Me too, but its everywhere and we'll never remove it. We can just recommend not using it.
- One argument is that we (in particular: I) use(d) it where we shouldn't.
- Another: `get_column_view` does not even always return a view! In '19, that @janezd guy added code that maps between different ordering of categorical values!

There may be cases when one needs a view, e.g. because you just created and unlocked a table. As current function does not really always give a view, but it has a good name that we can't reuse: the function `get_column` that I propose here has an argument `copy` that, if set to `True`, forces a copy. We could add `view=False`; if set to `True`, it would return a view and fail if it can't.

(Setting both `view` and `copy` to `True` would always return a sparse matrix, because the joker who calls it this way deserves it. 🤪)

Orange/data/table.py
```diff
@@ -208,8 +208,19 @@ def __init__(self, domain):
             setattr(self, v.name.replace(" ", "_"), v)
 
 
-class _ArrayConversion:
+def compute_column(func, *args, **kwargs):
```
markotoplak, line 211: This function should be private.

markotoplak: Done in the added commit.

markotoplak:
@janezd, thanks for this. All the new usages look much simpler, especially when the values needed to be set. Merging as soon as the tests pass.

PR #6062, opened 2022-07-15
Deprecate CD diagrams

markotoplak:
CD diagrams crash headless tests on Linux. Because Orange does not use these functions we should just remove them.

### DIFF
--- a/Orange/evaluation/scoring.py
+++ b/Orange/evaluation/scoring.py
@@ -11,6 +11,7 @@ Examples
 """
 
 import math
+import warnings
 
 import numpy as np
 import sklearn.metrics as skl_metrics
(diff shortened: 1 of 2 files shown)


markotoplak:
This was not the only cause for failing tests, but I still think that removing this code is beneficial. :)

janezd:
This is probably the only function in Orange that plots something and is intended to be called outside a widget. I always found it strange and wholeheartedly agree it should be removed.

But I suppose tests are failing only because of `graph_ranks`? I'm not fan of `compute_cd`, but unlike `graph_ranks`, `compute_cd` is not an exception so it could stay if it doesn't cause trouble.

markotoplak:
Yea, `compute_cd` could stay, but as nothing in Orange uses it, I thought it was best to remove it.

PR #6071, opened 2022-07-26
[ENH] Faster normalization, support Dask Tables

### LINKED ISSUE #5219: Normalization is inefficient in memory and time
- [ ] **What's wrong?**
Normalization is really slow. I did normalization and PCA on the 100000x1000 data set, and PCA was much faster than normalization. Here is what it looks like (memory use/against time from #5218):

myplot

Orange's normalization, which, by default only needs the mean and standard deviation, computes these very inefficiently. As intermediate result it builds a distribution for each variable, which is, for continuous variables, mostly a sorted list of values. 

Because normalization is the default for some learners and unsupervised methods, we should make it faster. This should be huge speedup for some Orange's learners, k-means and PCA.

markotoplak:
##### Issue
Normalization does not work for dask arrays. The computation of distributions is slow (it requires sorting). Fixes #5219

##### Description of changes
Use "fast stats" for Normalization.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -355,8 +355,8 @@ def stats(X, weights=None, compute_variance=False):
         return np.column_stack((
             np.nanmin(X, axis=0),
             np.nanmax(X, axis=0),
-            np.nanmean(X, axis=0) if not weighted else weighted_mean(),
-            np.nanvar(X, axis=0) if compute_variance else \
+            nanmean(X, axis=0) if not weighted else weighted_mean(),
+            nanvar(X, axis=0) if compute_variance else \
                 np.zeros(X.shape[1] if X.ndim == 2 else 1),
             nans,
             X.shape[0] - nans))
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -361,7 +361,7 @@ def stats(X, weights=None, compute_variance=False):
             nans,
             X.shape[0] - nans))
     elif is_sparse and X.size:
-        if compute_variance:
+        if compute_variance and weighted:
             raise NotImplementedError
 
         non_zero = np.bincount(X.nonzero()[1], minlength=X.shape[1])
@@ -370,7 +370,8 @@ def stats(X, weights=None, compute_variance=False):
             nanmin(X, axis=0),
             nanmax(X, axis=0),
             nanmean(X, axis=0) if not weighted else weighted_mean(),
-            np.zeros(X.shape[1]),      # variance not supported
+            nanvar(X, axis=0) if compute_variance else \
+                np.zeros(X.shape[1] if X.ndim == 2 else 1),
             X.shape[0] - non_zero,
             non_zero))
     else:
--- a/Orange/data/dask.py
+++ b/Orange/data/dask.py
@@ -3,6 +3,7 @@ import pickle
 import h5py
 import dask.array as da
 import numpy as np
+import pandas
 
 from Orange.data import Table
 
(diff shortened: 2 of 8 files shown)


markotoplak:
@noahnovsak, could you check this one, please? It is related to what you were looking into.

Orange/statistics/util.py
```diff
@@ -355,22 +355,24 @@ def weighted_mean():
         return np.column_stack((
             np.nanmin(X, axis=0),
             np.nanmax(X, axis=0),
-            np.nanmean(X, axis=0) if not weighted else weighted_mean(),
-            np.nanvar(X, axis=0) if compute_variance else \
+            nanmean(X, axis=0) if not weighted else weighted_mean(),
+            nanvar(X, axis=0) if compute_variance else \
```
noahnovsak, line 359: `column_stack` is still not supported in dask. This means it will compute everything immediately and raise an ugly FutureWarning. If this is something that bothers us we could use something like `np.stack(..., axis=1)` instead.

Also if I am not mistaken `nanmean` calls bottleneck's function for non sparse arrays, which causes trouble with dask arrays, right?

While I'm lookimg at this function `weighted_mean` also calls bottleneck's `nansum` (same concern as above) as well as `np.c_[weights]` (I do not see the point of this).

Orange/statistics/util.py
```diff
@@ -355,22 +355,24 @@ def weighted_mean():
...
         return np.column_stack((
             nanmin(X, axis=0),
             nanmax(X, axis=0),
-            nanmean(X, axis=0) if not weighted else weighted_mean(),
-            np.zeros(X.shape[1]),      # variance not supported
+            mean,
+            nanvar(X, axis=0) if compute_variance else \
+                np.zeros(X.shape[1] if X.ndim == 2 else 1),
             X.shape[0] - non_zero,
             non_zero))
```
noahnovsak, line 377: This seems very similar to the non sparse output. Could it be unified?

noahnovsak (commented):
This speeds things up a bit for normal Tables, but I have some concerns about dask compatibility.
Running it on a 10k x 10k DaskTable used up more memory than it should have (normally dask makes sure memory use stays around 5%, but this went up to 20%), also it didn't actually finish. Probably had to do with transformation, unrelated to normalization.

Orange/statistics/util.py
markotoplak: The variance for sparse matrices does not support weights. If it did, the code would be the same. But hmm, if we put the condition out then, yeah, maybe we could unify. I'll try.

markotoplak:
@noahnovsak, yes, the transformation for dask arrays is slow and takes more memory. This happens because each new column becomes a new computation recipe, which is then stored in memory. This needs to be optimized so that we have a single computation recipe which is then overlayed onto a part of the table.

noahnovsak:
As a side note, here are the performance results of the new and old normalization approaches on different-sized tables.

| data size | dist normalizer | stats normalizer | baseline |
| - | - | - | - |
| 100x100 | 25.1 ms ± 470 µs | 16 ms ± 444 µs | 12.5 ms ± 252 µs |
| 1000x1000 | 332 ms ± 6.07 ms | 190 ms ± 7.97 ms | 143 ms ± 2.96 ms |
| 10000x10000 | 16.6 s ± 1.46 s | 9.18 s ± 1.58 s | 4.47 s ± 164 ms |

Baseline represents the time used to apply any transformation on the table without including any computation of statistics.

Orange/statistics/util.py
markotoplak: Can't unify due to #6074. :) Otherwise I could.

PR #6075, opened 2022-07-27
[ENH] Box Plot: better dask support

noahnovsak:
##### Changes
- samples table for attribute and group sorting
- BoxData handles nan values, and avoids calling compute multiple times
- dask tests prototype

### DIFF
--- a/Orange/data/dask.py
+++ b/Orange/data/dask.py
@@ -164,7 +164,6 @@ class DaskTable(Table):
             return 0
         return np.isnan(self.X).sum().compute() / self.X.size
 
-
     def unlocked(self, *parts):
         # table locking is currently disabled
         return contextlib.nullcontext()
--- a/Orange/data/dask.py
+++ b/Orange/data/dask.py
@@ -153,6 +153,18 @@ class DaskTable(Table):
     def _update_locks(self, *args, **kwargs):
         return
 
+    def ensure_copy(self):
+        self._X = self._X.copy()
+        self._Y = self._Y.copy()
+        self._metas = self._metas.copy()
+        self._W = self._W.copy()
+
+    def get_nan_frequency_attribute(self):
+        if self.X.size == 0:
+            return 0
+        return np.isnan(self.X).sum().compute() / self.X.size
+
+
     def unlocked(self, *parts):
         # table locking is currently disabled
         return contextlib.nullcontext()
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -17,7 +17,6 @@ import scipy.special
 from scipy.stats import f_oneway, chi2_contingency
 
 from Orange.data import Table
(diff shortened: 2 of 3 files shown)


Orange/data/dask.py
```diff
@@ -104,6 +104,17 @@ def _filter_values(self, filter):
     def _update_locks(self, *args, **kwargs):
         return
 
+    def ensure_copy(self):
```
markotoplak, line 107: ensure_copy should be removed

markotoplak: Ok, sorry, that wss probably too early. So, could you explain again why was it needed?

noahnovsak: Normally `ensure_copy` makes a copy of `_X`, `_Y`, `_W` and `_metas` if their `base` attributes are not `None`.
Dask arrays don't have that attribute though, so this method always makes a copy just in case. I'm not sure if there is a better way to do this or if it is even necessary...

Orange/widgets/visualize/owboxplot.py
```diff
@@ -322,6 +290,11 @@ def set_data(self, dataset):
 
         self.dataset = dataset
         if dataset:
+            sample_at = 1000
+            if len(self.dataset) > sample_at:
```
markotoplak, line 294: I think this needs an additional condition: `isinstance(self.dataset, DaskTable)`. The current code would likely fail for big normal tables as `self.sample.compute()`.

markotoplak:
I am merging this (if the tests pass).

There are still some things that need investigating:
- Some tests with Dask are slow. We need to investigate why.
- The new code show the following warning often (on the master branch it is only shown once):
```
orange-widget-base/orangewidget/utils/itemmodels.py:585: DeprecationWarning: In future, it will be an error for 'np.bool_' scalars to be interpreted as an index
  self.dataChanged.emit(index, index)```

I edited your 2nd commit so that I kept the code as similar to the master branch as possible (some of your changes were stylistic). I think we should try to minimize changes to existing codebase.

markotoplak:
@noahnovsak, furthermore, there is no test that specifically tests sampling of DaskTables to order features. Please add it to the dask test class (in a new PR). You need to verify that the warning is shown, that it is cleared if new (smaller) data is input, and that the demanding operations actually use the sample.

noahnovsak:
I have opened #6081 to add the requested tests, but I have not managed to figure out the other issues yet.
1) I fear dask might just be that much slower than numpy as I cannot find any clear way to improve performance. As far as I can tell the main speed difference seems to be with discrete attributes. They calculate stats via `contingency` and `distribution`, which I believe still need to be updated to work nicely with dask?
2) The "np.bool_" warning is also something I am unable to trace. The only thing I've managed to figure out is, that the error disappears if we add `float()` in`_sort_list` like so:
`source_model.setData(source_model.index(i), float(key(attr)), Qt.UserRole)`

PR #6076, opened 2022-07-27
[ENH] Scatter Plot, Violin plot: basic (sampled) view of DaskTables

JakaKokosar:
This makes basic view into the data possible. The sampling is too aggressive. In the future we'd like to handle selections on the full data set.

 Also, when zooming in the zoomed-in region should be resampled.


Orange/data/dask.py
```diff
@@ -8,6 +9,33 @@
 import pandas
 
 from Orange.data import Table
+from Orange.widgets.utils.messages import UnboundMsg
```
markotoplak, line 12: This code should not know (import) anything about the widget. Please move this code somewhere else.

Orange/data/dask.py
```diff
@@ -150,6 +178,12 @@ def compute(self) -> Table:
         table.name = self.name
         return table
 
+    def sample(self, size: int = DEFAULT_SAMPLE_SIZE):
```
markotoplak, line 181: Instead of size use something more concrete, like rows or instances. Perhaps the same function could also sample columns in the future. Or have a parameter that specifies the max bytesize.

Orange/data/dask.py
```diff
@@ -150,6 +178,12 @@ def compute(self) -> Table:
...
+        # TODO: how is the random seed handled in Orange elsewhere?
```
markotoplak, line 182: Usually we want to have workflows reproducible, so set a random seed (like you already did).

I usually instantiate it is a `np.random.RandomState` and did not even know about the `default_rng` function.

Orange/tests/test_dasktable.py
```diff
@@ -111,3 +111,14 @@ def test_domain_basic_stats(self):
             self.assertEqual(c.max, 50)
             self.assertAlmostEqual(c.mean, 22.5328063)
             self.assertAlmostEqual(c.var, 84.4195562)
+
+    def test_sample(self):
+        zoo = Table('zoo')
+        with named_file('', suffix='.hdf5') as fn:
+            DaskTable.save(zoo, fn)
+            dzoo = DaskTable.from_file(fn)
+            dzoo_sample = dzoo.sample(10)
```
markotoplak, line 120: Please try to be specific. In this case `dzoo.sample(size=10)`, but according to another comment up there, rename `size`. :)

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -445,6 +446,7 @@ def _vizrank_color_change(self):
         self.vizrank_button.setToolTip(err_msg)
 
     @OWDataProjectionWidget.Inputs.data
+    @sample_dask_table
```
markotoplak, line 449: That won't suffice in general. Widgets will need to the aware of both (1) the sampled data, for display, and (2) the original data, for the outputs.

Most widgets allow selection types that can be applied onto the original data. Otherwise we seriously degrade the usefulness of the outputs of these widgets that make samples.

markotoplak:
@noahnovsak, @lanzagar, please look at this proposal and my comments. We can discuss it later. Or comment here.

markotoplak:
We need to ensure that widgets are aware of both (1) the sampled data, for display, and (2) the original data, for the outputs.

Most widgets allow selection types that can be applied onto the original data. Otherwise we seriously degrade the usefulness of the outputs on big data. For example, in a Violin Plot users can select ranges. For DaskTables, everything belonging to these ranges should go to the output, not just a subselection from the sample.

Orange/widgets/visualize/owscatterplot.py
JakaKokosar: I agree, and I did this to enable this visualization on sampled data with as few changes as possible.

I guess we have to decide on how we approach this in general. 

I think every widget should be able to access in-memory data from Dask Table and not hold this information on its own. You would do something like this self.data.sample to access in-memory data. And I would avoid doing this separately in every widget if possible.

JakaKokosar: I guess my question here is, do we accept the current solution as temporary until we decide on the mechanics or do we tackle this problem in this PR?

markotoplak:
Sorry, I forgot, I thought this was already part of some previous PR. Please ignore.

PR #6098, opened 2022-08-18
Rank: Move TableView and TableModel to gui

ajdapretnar:
##### Issue

Rank's TableView and TableModel are not independent modules that can be called for other widgets.

##### Description of changes
Move the two classes to gui.py, so other widgets can use them.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/tests/test_owrank.py
+++ b/Orange/widgets/data/tests/test_owrank.py
@@ -21,7 +21,7 @@ from Orange.classification import LogisticRegressionLearner
 from Orange.regression import LinearRegressionLearner
 from Orange.projection import PCA
 from Orange.widgets.data.owrank import OWRank, ProblemType, CLS_SCORES, \
-    REG_SCORES, TableModel
+    REG_SCORES, BarRatioTableModel
 from Orange.widgets.tests.base import WidgetTest, datasets
 from Orange.widgets.widget import AttributeList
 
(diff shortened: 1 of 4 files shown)


janezd:
I agree, but I fear we already have a plentitude of table models and views. Quick check showed it may not be so bad; details below. Still, please

- [ ] Rename `TableModel` into something more meaningful, e.g. `BarTableModel` (still, it is not the only model to provide bars)
- [ ] Add a docstring
- [ ] I'm not sure `TableView` adds anything that warrants moving it from `owrank` to `widgets.gui`. All visual settings and selection-related behaviour may be specific to the rank widget.

Now for my "brief analysis":

## Models

### General table models

<dl>
<dt>orangewidget.utils.itemmodels.AbstractSortTableModel:</dt>
<dd>A sorting proxy table model that sorts its rows in fast numpy, avoiding potentially thousands of calls into `QSortFilterProxyModel.lessThan()` or any potentially costly reordering of original data.
Override `sortColumnData()`, adapting it to your underlying model.</dd>

<dt>Orange.widgets.utils.itemmodels.AbstractSortTableModel(AbstractSortTableModel):</dt>
<dd>Add two deprecated methods for backward compatibility</dd>

<dt>Orange.widgets.utils.itemmodels.PyTableModel(AbstractSortTableModel):</dt>
<dd>A model for displaying python tables (sequences of sequences) in `QTableView objects`.</dd>

<dt>Orange.widgets.utils.itemmodels.TableModel(AbstractSortTableModel):</dt>
<dd>An adapter for using Orange.data.Table within Qt's Item View Framework. Also adds VariableStatsRole (min, max, etc.) and DomainRole (attr, class, meta)</dd>
</dl>


### Table models defined in widget modules (excluding specific models, such as those in Feature Statistics, Distance Matrix, Predictions...)

<dl>
<dt>Orange.widgets.data.owtable.RichTableModel(TableModel):</dt>
<dd>Add BarRole and information for headers (attributes of attributes)</dd>

<dt>Orange.widgets.data.owrank.TableModel:</dt>
<dd>Add a BarRatioRole that returns Data, normalized between the extremes. In sorting, put NaNs last.</dd>
</dl>

## Views

<dl>
<dt>Orange.widgets.utils.tableview.TableView(QTableView):</dt>
<dd>A QTableView subclass that is more suited for displaying large data models. Provides a `selectionFinished` signal that is emitted when selection is finished, but not during selecting.</dd>

<dt>orangewidget.gui.TableView(QTableView):</dt>
<dd>An auxiliary table view for use with PyTableModel in control areas. Essentially about sizing, and a delegate that may print some text in bold.

    **I believe we should deprecate or at least rename this class** because it is being used instead of `QTableView`. It may be indeed useful in evaluation widgets, though I'm not sure - it may also be redundant there. There are some eclatant misuses, such as importing it to get an enum, which is actually defined in `QTableView`.</dd>
</dl>

ajdapretnar:
@janezd I did as suggested. The alternative is to simply copy the code from Rank into OWCollocations (https://github.com/biolab/orange3-text/pull/782), which would also simplify release-making (Collocations would depend on releasing master first).

I am fine with either option, whatever you think is best.

Orange/widgets/data/tests/test_owrank.py
```diff
@@ -556,7 +556,7 @@ def test_concurrent_cancel(self):
 class TestRankModel(GuiTest):
```
janezd, line 556: Could you, please, also move this test to `Orange.widgets.tests.test_gui`?

ajdapretnar: Done. I just wasn't sure whether to rename the test...

janezd: I apologize for being such a nazi. We, that is pylint and I, think that you should also clean and rearrange your imports.

ajdapretnar: I know. I was waiting for hints from the lint run. I don't know how to run it locally so.... 🤷‍♀️

PR #6101, opened 2022-08-19
Confusion Matrix: Add Sum of probabilities

VesnaT:
##### Issue

Confusion matrix currently presents and evaluates the predicted classes. When one wants to evaluate the probability outputs it is less useful. For example, a majority class could be predicted for all instances, but the probabilities of several minority classes are still important (and differ among predictions).

##### Description of changes
Introduce `Sum of probabilities` Show option, which uses the sum of probabilities of the predicted class (instead of predicted class counter) when constructing the matrix.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owconfusionmatrix.py
+++ b/Orange/widgets/evaluate/owconfusionmatrix.py
@@ -101,6 +101,11 @@ class OWConfusionMatrix(widget.OWWidget):
                   "Proportion of predicted",
                   "Proportion of actual",
                   "Sum of probabilities"]
+    qu_tooltips = ["Number of correctly and incorrectly classified instances",
+                   "Proportion of predicted",
+                   "Proportion of actual",
+                   "Number of instances, distributed across columns "
+                   "according to predicted probabilities"]
 
     settings_version = 1
     settingsHandler = ClassValuesContextHandler()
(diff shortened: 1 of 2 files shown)


janezd:
At the first glance, I thought that these numbers would have to be somehow normalized. It took me some ten minutes to realize that they are exactly what we need, without normalization. For this, I had to think about them differently: this is, essentially, a confusion matrix in which a data instance is not put into one column but spread across columns according to assigned probabilities. Technically, it's a sum of probabilities. Conceptually, it's more like "Fuzzy confusion matrix". I think we need to find a more proper name for the option. "Fuzzy number of instances"?! "Number of instances, split according to probabilities" (too long!). Or "Instances, split by probabilites" -- and this option would be put below "Number of instances", to somehow indicate that the word "instances" referers to their number?

I really like this feature, it's going to be really useful!

Orange/widgets/evaluate/owconfusionmatrix.py
```diff
@@ -445,7 +446,8 @@ def _isinvalid(x):
 
         # Update the displayed confusion matrix
         if self.results is not None and self.selected_learner:
-            cmatrix = confusion_matrix(self.results, self.selected_learner[0])
+            learner_index = self.selected_learner[0]
+            cmatrix = confusion_matrix(self.results, learner_index)
```
janezd, line 450: In case `self.selected_quantity == 3`, we do not need to compute `cmatrix` (unused), `colsum` (wrong, recomputed later). `rowsum`, heh, is correctly computed here (the computation below should give the same result), but it's computed from `cmatrix`, thus it needs to be computed below (like it is now).

janezd:
Tooltip:
- Number of instances, distributed across columns according to predicted probabilities

PR #6107, opened 2022-08-25
Minor improvements in use of NumPy

maldil:
##### Issue

Not an issue

##### Description of changes
This is PR is to make use of NumPy APIs, `np.linalg.multi_dot` and `np.mean`. What do you think about the change.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/cur.py
+++ b/Orange/projection/cur.py
@@ -93,7 +93,7 @@ class CUR(Projector):
         if self.compute_U:
             pinvC = np.linalg.pinv(self.C_)
             pinvR = np.linalg.pinv(self.R_)
-            self.U_ = np.dot(np.dot(pinvC, X), pinvR)
+            self.U_ = np.linalg.multi_dot([pinvC, X, pinvR])
         else:
             self.U_ = None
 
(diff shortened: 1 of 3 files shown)


janezd:
Thanks for this.

The use of `multi_dot` is interesting: it will indeed speed up the computation when the number of features is much larger than the number of rows (or the other way around?) because it optimizes the order of multiplication of matrices.

As for the other change: OK, but I find `sum(correct) / len(correct)` cleaner than `mean(correct)`. Of course it's the same, but the former reads as the proportion of correct among all, and the other is ... just the mean. But this is a matter of opinion. And, `np.mean` is of course faster than `sum` - it should had been `np.sum`.

I'll merge, but we first need to see about the failing tests (which are unlikely to have anything to do with your PR).

janezd:
I've rebased your branch to the current master. Some tests still fail, but those are know to have problems. I'm merging your PR. Thanks!

Just a comment: you added commits to your master, pushed them to your github repository and made a pull request. Modifying your master and pushing from there will almost always cause problems, because you'll be unable to pull from github to your master. If your changes are rejected or, as in this case, if somebody (= I) rebases your PR, it could be even worse. Perhaps your best way out of this is now to remove the repository from your local disk and run `git clone`. Or, in the worst case remove your fork from github and fork again. Or it may be simpler -- I'm no git expert and I never needed to do this before.

Next time you decide to make a pull request: create a branch, commit changes to that branch, push it to github and then make a pull request. Here's one page that briefly describes it: https://opensource.com/article/19/7/create-pull-request-github.

PR #6044, opened 2022-06-22
[FIX] Fix tests cleanup

ales-erjavec:
##### Issue

After https://github.com/biolab/orange-widget-base/pull/204 the test will start failing/segfaulting.

##### Description of changes

Fix a call too commit in concurrent_example
Fix OWWidget cleanup in tests.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

PR #6112, opened 2022-08-29
Fix errors visible after removing pyqtgraph's except hook override

markotoplak:
##### Issue
Merging https://github.com/biolab/orange-widget-base/pull/204 showed an error in `-latest` tests.


markotoplak:
Fortunately, there was just a single error we caught this way, and even this one was in a widget only developers see.

The other (ubuntu) failures are unrelated to this PR

markotoplak:
Closed in favor of #6044.

Issue #6125, opened 2022-09-05
Group by: Standard deviation and Sum of TimeVariable


mw25:
**What's wrong?**
In the Group by widget an error message appears when calculating the aggregations "Standard deviation" or "Sum" for a time variable.

**How can we reproduce the problem?** 

- Load Dataset with TimeVariable (e.g. "Banking Crises")
- Select in Group by widget the Aggregations "Standard deviation" or "Sum" for the TimeVariable.

![image](https://user-images.githubusercontent.com/67997227/188425338-5fec3472-5182-4203-86db-cf790fb5571e.png)
![image](https://user-images.githubusercontent.com/67997227/188425370-97230546-72d5-4005-b8a9-e6a751d66691.png)

**What's your environment?**
- Operating system: Win10
- Orange version: '3.32.0.dev0+94958aa'
- How you installed Orange: git clone

PrimozGodec:
Thanks for reporting the issue. It seems like a bug. I will take a look.

PrimozGodec:
It is fixed in #6133

PR #6133, opened 2022-09-09
[FIX] Group By - fix std and sum for TimeVariable

PrimozGodec:
##### Issue

Fixe https://github.com/biolab/orange3/issues/6125

##### Description of changes
- use Series's std instead of GroupBy one since GB aggregations do not support DateTime data https://github.com/pandas-dev/pandas/issues/48481
- for variance, first, transform DateTime to Unix epoch since Panda's variance does not support DateTime
- std and span transform TimeDelta to seconds (Orange does not have TimeDelta type)
- I disabled the sum for TimeVariable since it doesn't make sense to compute data/times' sums.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owgroupby.py
+++ b/Orange/widgets/data/owgroupby.py
@@ -26,6 +26,7 @@ from orangewidget.utils.listview import ListViewSearch
 from orangewidget.utils.signals import Input, Output
 from orangewidget.utils import enum_as_int
 from orangewidget.widget import Msg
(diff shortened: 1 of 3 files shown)


VesnaT:
There is a problem with saved settings. If the Sum had been checked, it cannot be unchecked (it's disabled). This should be handled in settings migration.

PR #6135, opened 2022-09-12
Remove deprecated colorbrewer and colorpalette

PrimozGodec:
##### Issue
I just randomly found the `colorbrewer` and saw it is deprecated for a very long time. I think it is time to remove it.

##### Description of changes
Since `colorbrewer` and `colorpalette` are deprecated for long enough I propose deleting them.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
I welcome removal of `colorpalette` because I always have to check whether to use `colorpalette` or `colorpalettes`.

I hope, though, that no-addons use it. Have you checked? (I don't have single cell and a few others.)

PrimozGodec:
I have believed it is already removed from add-ons since `colorpalettes` has been deprecated for two years, but that is not the case. It is still used in [geo](https://github.com/biolab/orange3-geo/pull/151) (tests only), bioinformatics, and image analytics. So this PR will need to sit here for a while before it can be merged.

I am working on replacing it with colorpalettes.

janezd:
> I have believed it is already removed from add-ons since colorpalettes has been deprecated for two years, but that is not the case.

A s' ti nov tuki?

Still, it's a good that you created this PR, so that we show that deprecation is meant seriously and is coming. :)

PrimozGodec:
By the way, I think that failing the pyqt6 tests are caused by changes in orangewidgetbase (it started to fail after the release). I cannot reproduce it localy.

PrimozGodec:
I created a release of Image Analytics with colorpalette changes that users start to update to the new version. Geo does not need to be released -- only tests were fixed there. At Bioinformatic, fixes are merged, but it must be released before we merge this PR.

After Bioinformatics is released, I suggest waiting a bit before merging this PR so that more users update the addons (maybe we merge this PR after Orange 3.34 is released).

PR #6136, opened 2022-09-16
[FIX] owboxplot: Fix an error in widget cleanup during tests

ales-erjavec:
##### Issue

Since https://github.com/biolab/orange-widget-base/pull/204 was released, `Orange.widgets.tests.test_workflows` fails with
```
  File "/home/runner/work/orange3/orange3/.tox/pyqt6/lib/python3.9/site-packages/Orange/widgets/visualize/owboxplot.py", line 257, in eventFilter
    if obj is self.box_view.viewport() and \
AttributeError: 'OWBoxPlot' object has no attribute 'box_view'
ERROR: InvocationError for command /home/runner/work/orange3/orange3/.tox/pyqt6/bin/python -m unittest Orange.widgets.tests (exited with code -6 (SIGABRT)) (exited with code -6)
```

##### Description of changes

Change order of conditions in eventFilter to short-circuit access to widget members which can no longer exist (empty `self.__dict__`).

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/utils/pythoneditor/tests/test_api.py
+++ b/Orange/widgets/data/utils/pythoneditor/tests/test_api.py
@@ -19,11 +19,13 @@ class _BaseTest(WidgetTest):
     """
 
     def setUp(self):
+        super().setUp()
         self.widget = self.create_widget(SimpleWidget)
         self.qpart = self.widget.qpart
 
     def tearDown(self):
         self.qpart.terminate()
+        super().tearDown()
 
 
 class Selection(_BaseTest):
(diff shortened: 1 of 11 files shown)


janezd:
This looks bad. Do I understand correctly that some methods can be invoked when the widget is being destroyed and already has an empty `__dict__`?

Are there any other methods (that we use) in which we need to be careful?

Is this limited to tests (which have no screen) or can it also happen to the user?

PrimozGodec:
Tests are failing due to changes in pandas. I propose a temporary solution since I think it is a bug in pandas https://github.com/biolab/orange3/pull/6145

ales-erjavec:
> Do I understand correctly that some methods can be invoked when the widget is being destroyed and already has an empty `__dict__`?

Yes.

> Are there any other methods (that we use) in which we need to be careful?

`event()`, but other virtual event handlers could have the same problem.

> Is this limited to tests (which have no screen) or can it also happen to the user?

I guess it can happen to the user also.

markotoplak:
I am fixing the oldest deps tests in a separate PR - no need to worry about that.

PR #6144, opened 2022-09-20
Scatterplot: Fix crash for a single time tick

### LINKED ISSUE #6086: Scatter Plot: Crash with time data
**What's wrong?**

```
  File "/Users/vesna/pyqtgraph/pyqtgraph/graphicsItems/AxisItem.py", line 636, in paint
    specs = self.generateDrawSpecs(painter)
  File "/Users/vesna/pyqtgraph/pyqtgraph/graphicsItems/AxisItem.py", line 955, in generateDrawSpecs
    tickLevels = self.tickValues(self.range[0], self.range[1], lengthInPixels)
  File "/Users/vesna/orange3/Orange/widgets/visualize/owscatterplotgraph.py", line 354, in tickValues
    spacing = min(b - a for a, b in zip(ticks[:-1], ticks[1:]))
ValueError: min() arg is an empty sequence
```

Database (Online Retail) -> resize the ScatterPlot widget to the minimum width -> ScatterPlot (set x axis to InvoiceDate) -> crash

image

janezd:
##### Issue

Fixes #6086.

If there is only a single tick or no ticks, `min` that computed the min difference crashed.

##### Description of changes

Provide `default` argument. It equals the difference between the min and the max value, so `tickStrings` knows how to formulate the label.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -302,7 +302,10 @@ class AxisItem(AxisItem):
             step = int(np.ceil(float(len(ticks)) / max_steps))
             ticks = ticks[::step]
 
-        spacing = min(b - a for a, b in zip(ticks[:-1], ticks[1:]))
+        # In case of a single tick, `default` will inform tickStrings
+        # about the appropriate scale.
+        spacing = min((b - a for a, b in zip(ticks[:-1], ticks[1:])),
+                      default=maxVal - minVal)
         return [(spacing, ticks)]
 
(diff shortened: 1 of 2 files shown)


janezd:
`test_pandas` fails. Not my fault, or is it?

PrimozGodec:
> `test_pandas` fails. Not my fault, or is it?

It isn't your fault. It may be due to the latest pandas release.

PrimozGodec:
Here is the PR that solves the issue https://github.com/biolab/orange3/pull/6145

PR #6145, opened 2022-09-21
Temporally skip pandas version 1.5.0 

PrimozGodec:
##### Issue

Pandas may have a bug (or it is just partially implemented new rule for DateTime conversion) that cause our test to fail in the latest version https://github.com/pandas-dev/pandas/issues/48678

##### Description of changes
Skip pandas version 1.5.0 until they decide whether it is a bug or feature.
If they fix/change the issue solution will probably be available in 1.5.1, so nothing needs to be changed.
If it is not a bug, we need to adapt tests.

##### Includes
- [ ] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -22,7 +22,7 @@ python-louvain>=0.13
 requests
 openTSNE>=0.6.1
 baycomp>=1.0.2
-pandas>=1.3.0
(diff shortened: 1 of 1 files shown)


PrimozGodec:
I can confirm that it is resolved. The error will be fixed in the next version; no action is required

PR #6146, opened 2022-09-23
Hierarchical Clustering: Add Subset signal

### LINKED ISSUE #6126: Hierarchical Clustering needs Data Subset input channel
It would be great if Hierarchical Clustering could show a subset of data. Say, I find some data instances in t-SNE plot, and I would like to know if they have clustered together in HC as well.

One way to depict the subset is to print the labels in bold. If there are no labels, mark the subsetted leaves with a dot.

janezd:
##### Issue

Closes #6126 and more.

- Labels for the subset are shown in bold.
- Adds option to show additional icons, colored by values of some variable; if subset is present, only squares for the rows in the subset are filled.
- If there are no labels and no icons, subset is indicated by small dots.
- Refactors most of `TextListWidget`'s code into a base class (`TextListBase`) and derives a new class, `TextListView` that gets data from a model. Hierarchical Clustering than uses it instead of `TextListWidget`. Distance Matrix and Silhouette can do the same.

- Since class is not used as default for coloring, the first string attribute is used for annotation. If there are none, it uses class, otherwise the first non-None option.
- Enumeration is now shown only if there are not other options (except "None"); enumeration served no purpose, imho.
- Fix this

     ```python
  File "/Users/janez/orange3/Orange/widgets/unsupervised/owhierarchicalclustering.py", line 143, in data
    return self.__colors[index.row()]
IndexError: index 101 is out of bounds for axis 0 with size 101
```

    This probably happens when data is changed and colors are set before labels or vice versa.
Draft: ready to use for demonstration purposes, but needs polishing and testing. In particular, I need to check that resizing (of widget, of fonts ...) is handled properly.

To do:
- [ ] Check what these changes do to other widgets the use `TextListWidget`
- [ ] Change default linkage to Ward (otherwise unrelated to this PR)

Screen Shot 2022-09-24 at 20 20 33

Screen Shot 2022-09-24 at 20 20 51

Screen Shot 2022-09-24 at 20 21 00

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -163,6 +163,8 @@ class OWHierarchicalClustering(widget.OWWidget):
     top_n = settings.Setting(3)
     #: Dendrogram zoom factor
     zoom_factor = settings.Setting(0)
+    #: Show labels only for subset (if present)
+    label_only_subset = settings.Setting(False)
 
     autocommit = settings.Setting(True)
 
@@ -212,6 +214,10 @@ class OWHierarchicalClustering(widget.OWWidget):
         self.connect_control("annotation", on_annotation_changed)
 
         box.layout().addWidget(self.label_cb)
+        gui.checkBox(
+            box, self, "label_only_subset", "Show labels only for subset",
+            disabled=True,
+            callback=self._update_labels, stateWhenDisabled=False)
 
         box = gui.radioButtons(
             self.controlArea, self, "pruning", box="Pruning",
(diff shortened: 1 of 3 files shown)


VesnaT:
Distance Map: 
- [ ] Annotations - all columns/rows have the same value
image

Silhouette Plot: 
- [ ] Annotations - Uncheck "Show in groups" - all rows have the same value

Heat Map:
- [ ] Annotations - Set "Split by rows" to None - all rows have the same value

Hierarchical Clustering
- [ ] I assume the annotations are incorrect when pruned:
image
image

VesnaT (approved):
The subset labels are still bold, despite pruning and warning. If this is intentional, I'm happy to merge.
image

PR #6153, opened 2022-09-30
[FIX] Remove pyqt5 install magic

markotoplak:
Now there is also pyqt6 and we still force our users to also install PyQt5. This trick we had is also dangerous because it behaves differently with wheels and source installs.

Reverts #5593.

After merging this PR please also consider: https://github.com/biolab/orange3/pull/6153


PrimozGodec:
:raised_hands::skin-tone-2:

PR #6154, opened 2022-09-30
Remove PyQt5 requirements from setup

PrimozGodec:
##### Issue

PyQt5 requirement disables users from using PyQt6 
I tried to check PyQt more generally but noticed the following things:
- `pip install .` will build the wheel in a separate isolated process and will not see whether PyQt is in the requirement (so PyQt will always be installed independently from the environment)
- `pip install orange3` installs from wheels that always have pyqt5 requirements inside (and setup.py is not there)

So checking whether PyQt is installed or not does make sense. PyQt will always be put in dependencies. 

##### Description of changes
To avoid confusion in Conda environments (having PyQt both from PyPi and conda), I am removing this dependency.

##### Includes
- [ ] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,14 +34,6 @@ try:
 except ImportError:
     have_cython = False
 
-try:
-    import PyQt5.QtCore  # pylint: disable=unused-import
-    have_pyqt5 = True
-except ImportError:
-    have_pyqt5 = False
-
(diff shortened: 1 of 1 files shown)


markotoplak:
Duplicate of #6153 .

PR #6158, opened 2022-10-01
Feature Statistics: Adapt output to types of variables

### LINKED ISSUE #6151: "Statistics" output of Feature statistics is inconsistent
**What's wrong?**
For categorical variables, Feature Statistics determines the Mode as the central tendency and puts it in the Median column. That's OK with me, although it could be described a bit better in the help popup. A separate Mode column would even be better.
But if I want to put the feature statistics in a table (or save them in a file) the categories in the Median column are converted to integers. This is inconsistent, and unclear because the mapping from a category to the corresponding number isn't obvious. 

See attached zip file and screenshot

**What's your environment?**
- Operating system: Mac OS 12.6 on Silicon
- Orange version: 3.32.0
- How you installed Orange: from DMG
[Feature statistics inconsistent output.ows.zip](https://github.com/biolab/orange3/files/9674384/Feature.statistics.inconsistent.output.ows.zip)
Screen Shot 2022-09-29 at 13 54 27

janezd:
##### Issue

Fixes #6151.

##### Description of changes

- The most common value for discrete variables is now shown in separate column of output. All columns that don't apply (including min and max) are set to undefined.
- Output columns adapt to types of variables: if there are no categorical variables, the majority column is removed. If there are only categorical, the columns that do not apply are removed.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owfeaturestatistics.py
+++ b/Orange/widgets/data/owfeaturestatistics.py
@@ -290,52 +290,45 @@ class FeatureStatisticsTableModel(AbstractSortTableModel):
             time_f=lambda x: ut.nanmedian(x, axis=0),
         )
 
-    def get_statistics_matrix(self, variables=None, return_labels=False):
-        """Get the numeric computed statistics in a single matrix. Optionally,
-        we can specify for which variables we want the stats. Also, we can get
-        the string column names as labels if desired.
+    def get_statistics_table(self):
+        """Get the numeric computed statistics in a single matrix."""
+        if self.table is None or not self.rowCount():
+            return None
 
-        Parameters
-        ----------
-        variables : Iterable[Union[Variable, int, str]]
-            Return statistics for only the variables specified. Accepts all
-            formats supported by `domain.index`
-        return_labels : bool
-            In addition to the statistics matrix, also return string labels for
-            the columns of the matrix e.g. 'Mean' or 'Dispersion', as specified
-            in `Columns`.
+        # don't match TimeVariable, pylint: disable=unidiomatic-typecheck
+        contivars = [type(var) is ContinuousVariable for var in self.variables]
+        if any(contivars):
+            def c(column):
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owfeaturestatistics.py
```diff
@@ -290,52 +290,45 @@ def __mode(x, *args, **kwargs):
...
+            metas = np.vstack((names, majorities)).T
+            meta_attrs = [StringVariable('Feature'), StringVariable('Majority')]
+        else:
+            metas = np.atleast_2d(names).T
+            meta_attrs = [StringVariable('Feature')]
 
-        return matrix
+        domain = Domain(attributes=attrs, metas=meta_attrs)
+        statistics = Table.from_numpy(domain, x, metas=metas)
+        statistics.name = '{self.data.name} (Feature Statistics)'
```
lanzagar, line 330: It's better to:
1) put an f before this string
2) leave this line in `OWFeatureStatistics.commit`, where self.data.name actually exists

:bug: :bug: :bug: 
If you move it outside of this function, don't forget that this can return None in which case you should not set the name...

lanzagar (changes requested):
This works and can be merged after the table name fix.
Some additional considerations:
- We might want to call the new column Mode instead of Majority
- After this change the widget contents and the output table will no longer be in sync. To make it more consistent, I would vote to use Mode in the widget as well - would make even more sense if that column is used for all variables (C and N).
-  Although not related to this change, I noticed these in the log:
  `UserWarning: decorate OWFeatureStatistics.commit with @gui.deferred and then explicitly call commit.now or commit.deferred`

Orange/widgets/data/owfeaturestatistics.py
lanzagar: I just noticed, that you can probably just change self.data to self.table and leave the line where it is.

lanzagar (approved):

PR #6178, opened 2022-10-21
[FIX] Hierarchical Clustering: Color by metas

VesnaT:
##### Issue

The widget crashes when selected `Color by` variable is from metas (and therefore has type object).

##### Description of changes
- cast column to float

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Fixed in #6058.

PR #6192, opened 2022-11-05
CI: Raise required oldest sicpy to 1.9

janezd:
##### Issue

Tests on oldest fail after #6174, "Feature Statistics: Use deferred commit + **minor warning fix** due to the "minor warning fix" part. In scipy 1.9, `scipy.stats.mode` introduced argument `keepdims` and warns if it's not set. In Scipy 1.8 this argument did not exist.

We overlooked this because we got so used to some test configurations failing that we don't even check. Me very included.

##### Description of changes

We could
- ignore this failing configuration (obviously: no),
- remove `keepdims` argument and live with warning, but this would hunt us some day,
- do something fancy to pass the argument only when necessary,
- raise required version of Scipy from 1.3 to 1.9, which would then cascade to some other libraries (numpy 1.17 -> 1.19.5)

I see no reason against the latter. Does anybody?

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -38,8 +38,8 @@ requirements:
   run:
     - python
     - setuptools >=41.0.0
-    - numpy >=1.17.3
-    - scipy >=1.3.1
+    - numpy >=1.19.5
+    - scipy >=1.9
     - scikit-learn >=1.0.1
     - bottleneck >=1.3.4
     - chardet >=3.0.2
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -1,6 +1,6 @@
 pip>=18.0
-numpy>=1.17.3
-scipy>=1.3.1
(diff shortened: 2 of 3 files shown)


PrimozGodec:
I do not see any problem with using a newer version of Scipy. Also, a Numpy change would not hurt; 1.19 is still quite old. Other options would bring just some complications in our code.

If we keep this change, just do not forget to make changes in requirements.txt

PrimozGodec:
Now we already have a known problem with reathedocs: https://github.com/readthedocs/readthedocs.org/issues/9623
They, by default, still use python 3.7 to build the documentation, and we cannot directly use reathdocs.yaml file in our repository since we build multiple subprojects.

I wanted to deal with this but didn't have time yet.

PrimozGodec:
This PR will fix the documentation: https://github.com/biolab/orange3/pull/6197

PR #6200, opened 2022-11-14
Avoid counting unknowns for big tables

markotoplak:
##### Issue
For big tables counting unknowns takes too long (especially important for table summaries that all widgets have). Therefore we opted to skip it for big tables.

##### Description of changes
The original code was also slightly rewritten to avoid a separate NaN check, which does not seem to gain anything (just an additional pass through the data).

This PR also fixes a bug with `get_nan_frequency_attribute` and sparse tables.

This comes from the `dask` branch.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1532,16 +1532,18 @@ class Table(Sequence, Storage):
         """Return `True` if there are any missing class values."""
         return bn.anynan(self._Y)
 
-    def get_nan_frequency_attribute(self):
-        if self.X.size == 0:
+    @staticmethod
+    def __get_nan_frequency(data):
+        if data.size == 0:
             return 0
-        vals = self.X if not sp.issparse(self.X) else self.X.data
-        return np.isnan(vals).sum() / self.X.size
+        dense = data if not sp.issparse(data) else data.data
+        return np.isnan(dense).sum() / np.prod(data.shape)
+
+    def get_nan_frequency_attribute(self):
+        return self.__get_nan_frequency(self.X)
 
     def get_nan_frequency_class(self):
-        if self.Y.size == 0:
-            return 0
-        return np.isnan(self._Y).sum() / self._Y.size
+        return self.__get_nan_frequency(self.Y)
 
     def checksum(self, include_metas=True):
         # TODO: zlib.adler32 does not work for numpy arrays with dtype object
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1535,7 +1535,8 @@ class Table(Sequence, Storage):
     def get_nan_frequency_attribute(self):
         if self.X.size == 0:
             return 0
-        return np.isnan(self.X).sum() / self.X.size
+        vals = self.X if not sp.issparse(self.X) else self.X.data
+        return np.isnan(vals).sum() / self.X.size
 
     def get_nan_frequency_class(self):
         if self.Y.size == 0:
--- a/Orange/widgets/data/tests/test_owfile.py
+++ b/Orange/widgets/data/tests/test_owfile.py
@@ -76,6 +76,23 @@ class TestOWFile(WidgetTest):
     def tearDown(self):
         dataset_dirs.pop()
 
+    def test_describe(self):
+        table = Table("iris")
+        description = f"<p><b>Iris flower dataset</b><br/>" \
+                      f"Classical dataset with 150 instances of Iris setosa, Iris virginica and Iris versicolor.</p>" \
+                      f"<p>150 instance(s)<br/>" \
+                      f"4 feature(s) (no missing values)<br/>" \
+                      f"Classification; categorical class with 3 values (no missing values)<br/>" \
+                      f"0 meta attribute(s)</p>"
+        self.assertEqual(description, self.widget._describe(table))
+
+        table = Table.from_numpy(domain=None, X=np.random.random((10000, 1000)))
+        description = f"<p>10000 instance(s)<br/>" \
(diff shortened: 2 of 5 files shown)


Orange/data/table.py
```diff
@@ -1535,7 +1535,8 @@ def has_missing_class(self):
     def get_nan_frequency_attribute(self):
         if self.X.size == 0:
             return 0
-        return np.isnan(self.X).sum() / self.X.size
+        vals = self.X if not sp.issparse(self.X) else self.X.data
+        return np.isnan(vals).sum() / self.X.size
```
janezd, line 1539:
```
>>> sp.csc_matrix([[1, 2, 0], [0, 0, 1]]).size
3
```

This code assumes it's 6, right?

janezd (changes requested):

janezd:
https://github.com/biolab/orange3/pull/5820/files#diff-83b0a0d5553c7fdad2b2a942047ff6ee5812cec82457e1995514363e4a3196db, which was merged recently, changed some messages in the file widget. I suppose this is the reason for failed tests.

Orange/data/table.py
noahnovsak: I thought I had an answer, but looking at it again the concept of nan frequency in sparse arrays bugs me. What is the number we actually care about here?

janezd: In sparse matrices, the absent values are zeros, not missing. Hence the correct denominator is `self.X.shape[0] * self.X.shape[1]`, I think.

markotoplak: I think that instead of the size, as Janez pointed out, we should use the thing that returns "6". :)

Orange does not count NaNs consistently. The code by Niko usually returns number of zero elements. I'd prefer it to return the number of actual "NaN" values. So I'd go with the actual number of NaNs.

janezd: As far as I checked - and I did check -, there seems to be no thing that returns 6.

noahnovsak: `vals.size` doesn't return 6?

Orange/data/table.py
```diff
@@ -1535,7 +1535,8 @@ def has_missing_class(self):
     def get_nan_frequency_attribute(self):
         if self.X.size == 0:
             return 0
-        return np.isnan(self.X).sum() / self.X.size
+        vals = self.X if not sp.issparse(self.X) else self.X.data
+        return np.isnan(vals).sum() / np.prod(self.X.shape)
 
     def get_nan_frequency_class(self):
```
janezd, line 1541: I think you should do the same for `get_nan_frequency_class`. The problem there is much smaller - while `Domain` and `Table` support multiple classes, many widgets don't so we don't recommend that and almost never use it. Still; it's a trivial change, and it may hunt us or somebody else some time in the future.

Orange/widgets/data/tests/test_owfile.py
```diff
@@ -76,6 +76,23 @@ def setUp(self):
     def tearDown(self):
         dataset_dirs.pop()
 
+    def test_describe(self):
```
janezd, line 79: You do not intend to test this exact wording, right? This test is meant to check that `get_nan_frequency_attribute` is not called for large data. Hence, explicitly test that it's not called. Do this by patching the function (like in the test bellow) and `assert_not_called` in the second case.

Also, rename the test into what it tests, e.g. `test_describe_not_calling_get_nans`.

I know that we have tests that test the exact wording and I may had been writing them, too -- or at least I merged some. I think they are a bad idea, though. A proper test for the description itself would be that it includes numbers 150 and 4 and the word Iris.

Orange/widgets/utils/tests/test_state_summary.py
```diff
@@ -183,6 +184,15 @@ def test_details(self):
                   f'Target: —'
         self.assertEqual(details, format_summary_details(data))
 
+        data = Table.from_numpy(domain=None, X=np.random.random((10000, 1000)))
+        details = f'{len(data):n} instances, ' \
+                  f'{len(data.domain.variables)} variables\n' \
+                  f'Features: {len(data.domain.variables)} numeric \n' \
+                  f'Target: —'
+        with patch.object(Table, "get_nan_frequency_attribute") as mock:
+            self.assertEqual(details, format_summary_details(data))
```
janezd, line 193: I should have the same objection as above. But these tests are already written the wrong (imho) way, so ... let it stay.

Orange/widgets/utils/tests/test_state_summary.py
```diff
@@ -183,6 +184,15 @@ def test_details(self):
...
+            self.assertFalse(mock.called)
```
janezd, line 194: The usual way to test this is `mock.assert_not_called()` because it gives a better diagnostic message - if you test for `False`, it just tells you that `True` is not `False`...

Orange/data/table.py
noahnovsak: Upon copying the changes to `get_nan_frequency_class` I decided I could also just merge the two into a single function that handles both `X` and `Y` (and possibly `metas` or `W` in the future). I am however curious if they should remain separate for some reason?

janezd: No, I see no reason for separate functions given that array/matrix can be an argument.

markotoplak: If you'd make a new `get_nan_freqency` function that can handle any part, keep in mind that we can not delete the old functions immediately. We'll need to deprecate them first.

PR #6204, opened 2022-11-16
[FIX] Weighted mean computation in Orange.statistics.util.stats

markotoplak:
##### Issue
The tests added in the first commit of this PR fail. The new tests test a combination of unit (ones) weights and NaNs, which should give exactly the same results as unweighted samples. Well, they do not, because we have been sloppy: the current code also divided with non-used weights (corresponding to NaN data entries).

##### Description of changes
The `nanmean` function was easy to extend with weights. For sparse I just needed to add a parameter. For dense I did an extension. Now, `stats` uses `nanmean` internally.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -342,20 +342,14 @@ def stats(X, weights=None, compute_variance=False):
     is_numeric = np.issubdtype(X.dtype, np.number)
     is_sparse = sp.issparse(X)
     weighted = weights is not None and X.dtype != object
-
-    def weighted_mean():
-        if is_sparse:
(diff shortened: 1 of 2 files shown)


markotoplak:
The bug originates from 8a76580ec839cf6fff9364ff1f4794e86834cbe6 by @nikicc. I nominate @pavlin-policar, who worked on similar code, to review this PR. :)

pavlin-policar:
This looks good to me. I've re-run the failing tests, but, in case they don't pass, can I just ignore them? The failures have nothing to do with this code here, and this looks ready to merge.

PR #6210, opened 2022-11-18
Performance Curve: Compute area under the curve

VesnaT:
##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owliftcurve.py
+++ b/Orange/widgets/evaluate/owliftcurve.py
@@ -1,5 +1,5 @@
 from enum import IntEnum
-from typing import NamedTuple, Dict, Tuple
+from typing import NamedTuple, Dict, Tuple, List
 
 import numpy as np
 from sklearn.metrics import precision_recall_curve
@@ -202,6 +202,9 @@ class OWLiftCurve(widget.OWWidget):
 
         gui.rubber(self.controlArea)
 
+        box = gui.vBox(self.controlArea, box="Area under the curve")
+        self._area_info = gui.label(box, self, "/", textFormat=Qt.RichText)
+
         gui.auto_apply(self.buttonsArea, self, "auto_commit")
 
         self.plotview = GraphicsView()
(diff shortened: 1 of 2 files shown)


Orange/widgets/evaluate/owliftcurve.py
```diff
@@ -388,6 +398,29 @@ def _plot(points, pen, kwargs):
         self.plot.curve_items.append(_plot(points, wide_pen, line_kwargs))
         return True
 
+    def _update_info(self, is_valid: bool):
```
janezd, line 401: `bool` -> `List[bool]`

Orange/widgets/evaluate/owliftcurve.py
```diff
@@ -388,6 +398,29 @@ def _plot(points, pen, kwargs):
...
+        self._area_info.setText("/")
+        if is_valid:
```
janezd, line 403: See above. :)

Orange/widgets/evaluate/tests/test_owliftcurve.py
```diff
@@ -394,6 +394,12 @@ def test_points_from_results_precision_recall():
         np.testing.assert_almost_equal(res.respondents, respondents)
         np.testing.assert_almost_equal(res.thresholds, thresholds)
 
+    def test_area(self):
+        x = np.arange(5)
+        y = np.arange(5)[::-1]
+        area = compute_area(x, y)
+        self.assertEqual(area, 10)
+
```
janezd, line 402: This test is not very strong - points start at 0, with equal space of `1`, the curve is symmetric. I can think of many functions that would pass this. :)

Here's a stronger test. With a nice ring to it, too. :)

```python
    def test_area(self):
        x = np.array([5, 8, 9, 11])
        y = np.array([7, 14, 8, 0])
        area = compute_area(x, y)
        self.assertEqual(area, 51)
```

janezd (changes requested):

PR #6211, opened 2022-11-19
Translation fixes: Pivot and Edit Domain

janezd:
##### Issue

Pivot used enum names as checkbox labels. This can't be translated. It also pickled enums, which is not recommended.

Edit Domain had a dialog that made assumptions about widths of labels (it simplified the laying out by assuming that two labels are much longer than the third; in Slovenian translation, this cause some ugly white space)

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owcreateclass.py
+++ b/Orange/widgets/data/owcreateclass.py
@@ -5,7 +5,7 @@ from itertools import count
 import numpy as np
 
 from AnyQt.QtWidgets import QGridLayout, QLabel, QLineEdit, QSizePolicy, QWidget
-from AnyQt.QtCore import QSize, Qt
+from AnyQt.QtCore import Qt
 
 from Orange.data import StringVariable, DiscreteVariable, Domain
 from Orange.data.table import Table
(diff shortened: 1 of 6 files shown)


VesnaT (approved):

Orange/widgets/data/owcreateclass.py
```diff
@@ -562,23 +564,30 @@ def send_report(self):
...
+                else:
+                    return f"all {n_total} matching instances are {aca}"
+            elif not patt:
+                return f"{n_matched} {pl(n_matched, 'instance')}"
             else:
-                return "{} matching instances".format(n_matched)
+                m = f"{n_matched} matching {pl(n_matched, 'instance')}"
+                if n_matched < n_total:
+                    n_already = n_total - n_matched
+                    m += f" (+{n_already} that {pl(n_already, 'is|are')} {aca}"
```
VesnaT, line 589: Ok, you've handled the missing `' '`, but can you handle the missing `)`

janezd: Not yet. You must understand that developing a solution for missing spaces took years and years of programming experience. The problem of missing parentheses is at least as hard as missing spaces, so it will require an even greater moment of inspiration.

For now, I just humbly added it.

PR #6213, opened 2022-11-19
Add option InheritEq and fix several missing __eq__s and __hash__s

janezd:
##### Issue

Missing `__eq__` and `__hash__` in classes used for `compute_value` seems to have become the large source of warnings. In many cases, `__eq__` and `__hash__` are not needed, but the tests we have still require providing the two methods (which then just call inherited methods).

This cannot be avoided automatically, I think. The warning cannot detect whether the instance's attributes were added by instances actual type or by its parent. Furthermore, in theory, the parent's attributes can be used differently and hence require reimplementing hash.

##### Description of changes

The class can now add a class attribute `InheritEq = True`. If set, it supresses the warning. The attribute is checked for in type's `__dict__` and is thus not inheritable.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/filter.py
+++ b/Orange/data/filter.py
@@ -31,6 +31,12 @@ class Filter(Reprable):
     def __call__(self, data):
         return
 
+    def __eq__(self, other):
+        return type(self) is type(other) and self.negate == other.negate
+
+    def __hash__(self):
+        return hash(self.negate)
+
 
 class IsDefined(Filter):
     """
@@ -53,7 +59,7 @@ class IsDefined(Filter):
 
     def __init__(self, columns=None, negate=False):
         super().__init__(negate)
-        self.columns = columns
+        self.columns = tuple(columns) if columns is not None else None
 
     def __call__(self, data):
         if isinstance(data, Instance):
@@ -69,6 +75,12 @@ class IsDefined(Filter):
         if self.negate:
             r = np.logical_not(r)
         return data[r]
+
+    def __eq__(self, other):
+        return super().__eq__(other) and self.columns == other.columns
+
+    def __hash__(self):
+        return super().__hash__() ^ hash(self.columns)
 
 
 class HasClass(Filter):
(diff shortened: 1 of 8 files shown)


Orange/data/filter.py
```diff
@@ -70,6 +76,12 @@ def __call__(self, data):
             r = np.logical_not(r)
         return data[r]
 
+    def __eq__(self, other):
+        return super().__eq__(other) and self.columns == other.columns
+
+    def __hash__(self):
+        return super().__hash__() ^ hash(self.columns)
```
markotoplak, line 83: Why xor and not putting it within one `hash()`?

markotoplak:
I strongly recommend merging this.

Orange/data/filter.py
janezd: It would actually be much better to use a hash of a tuple. Tuple's hash is much smarter than xor: https://github.com/python/cpython/blob/main/Objects/tupleobject.c#L289

```
>>> a = "Ana"
>>> b = "Berta"
>>> hash(a) ^ hash(a)
0
>>> hash(b) ^ hash(b)
0
```

Obviously.

Whereas:

```
>>> hash((a, a))
3546956001028672602
>>> hash((b, b))
6385707759775101969
```

markotoplak: I know it is probably not significant, but I changed it anyway.

Orange/preprocess/impute.py
```diff
@@ -207,6 +207,14 @@ def __call__(self, data):
         column[mask] = predicted
         return column
 
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
markotoplak: Well, as long as it does not crash it is probably "fine". :) I did fix another bug while playing with this: `__hash__` was wrong because superclass (Reprable) `__hash__` does something strange.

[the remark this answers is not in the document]
janezd: Objects are equal if they are same.

```
>>> a = A()
>>> b = A()
>>> a == b
False
>>> a == a
True
```

In case of models, this suffices. Equality of models (beyond sameness) could be difficult to check, so let's not go there. Note that Python's functions behave the same way.

```
>>> (lambda: 42) == (lambda: 42)
False
```

PR #6208, opened 2022-11-18
[FIX] Data Sampler: Fix crash when requesting an empty sample

janezd:
##### Issue

Fixes #6207.

##### Description of changes

Widget already treated a 100 % sample as special case. The same code is now also used for empty sample.

Related change: when setting the proportion, widget allowed to set 0 %, but when setting the number of instances, it enforced a minimum of 1. Now it also allows 0.

##### Includes
- [X] Code changes
- [X] Tests

PR #6221, opened 2022-11-25
[FIX] stats: Handle empty array

VesnaT:
##### Issue

Fixes https://github.com/biolab/orange3/pull/6208#issuecomment-1327209138

##### Description of changes
Compute stats for an empty array / Table.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -370,6 +370,8 @@ def stats(X, weights=None, compute_variance=False):
             X.shape[0] - non_zero,
             non_zero))
     else:
+        if X.ndim == 1:
+            X = X[:, None]
         nans = (pandas.isnull(X).sum(axis=0) + (X == "").sum(axis=0)) \
             if X.size else np.zeros(X.shape[1])
         return np.column_stack((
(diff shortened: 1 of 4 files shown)


janezd:
@VesnaT, I don't think this is correct fix. I'd even dare putting `assert X.ndim == 2` here. I think we should always safely assume that `X` is two-dimensional, where `X.shape[1]` equals `len(data.domain.attributes)`.

The problem is probably somewhere in class `Table`, which - as far as I see for now - constructs a zero-dim array when selecting zero rows. The problem must be fixed there; unless we prevent zero-dim arrays, everything will crash.

markotoplak:
@janezd, I dare to say you forgot about our beautiful `data.Y`, that can certainly be 1-dimensional (and plays the role of `X` here).

janezd:
@markotoplak, you're wrong. I didn't forget that; at the time of writing I haven't gone deep enough yet, and I thought this is just about `X` (as in `data.X`). I didn't yet know this `X` could also be our beautiful shape-shifting `Y`. :)

PR #6223, opened 2022-11-25
[MNT] WidgetTest: Add tests on nasty data

janezd:
##### Issue

Widget are supposed to survive data all-nans or without rows, attributes, classes... Many don't because we don't test it.

##### Description of changes

`WidgetTest`, the base class for all widget tests now adds a test that sends nasty data to all inputs that accept `Table`.

Suggestions welcome.

This should be fun.

##### Includes
- [ ] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/tests/base.py
+++ b/Orange/widgets/tests/base.py
@@ -14,6 +14,7 @@ from AnyQt.QtWidgets import (
     QComboBox, QSpinBox, QDoubleSpinBox, QSlider
 )
 
+from orangewidget.utils.signals import MultiInput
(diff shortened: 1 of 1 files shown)


i18n/si/msgs.jaml
```diff
@@ -6479,6 +6479,8 @@ widgets/data/owmelt.py:
             Data: Podatki
         class `Outputs`:
             Data: Podatki
+        class `Error`:
+            No features to melt: Ni stolpec za raztapljanje
```
VesnaT: stolpec -> stolpcev

PR #6234, opened 2022-12-02
MDS: Timing of updates, step button, visual improvements

### LINKED ISSUE #6228: Change Refresh in MDS widget to Delay
**What's your use case?**

We used the Refresh option with MDS for educational purposes to show that MDS performs optimization and is an embedding method. The points moved in every step of optimization, and one could see the convergence to the final solution. Not that this is too enlightening or required, but it was just an extra feature that could help when introducing MDS.

image

Since computers got faster, and for small data sets such as the zoo dataset, setting the refresh even to Every does not help. The optimization converges instantly, and the movement of points can not be tracked.

Replace the Refresh with Delay, with options 0.1s, 0.2s, 0.5s, 1s, which defines that max delay between two successive iterations.

Remove the refresh option altogether.

janezd:
##### Issue

Fixes #6228.

##### Description of changes

- Optimization ensures that at least 0.1 s passes between each refresh.
- Added step button, reorganize layout of buttons.
- I thought the widget was not updating. It was, but it was very slow because of cosmetic lines with widths different than 1. Enabling `segmentedLineMode` (available from latest pyqtgraph) makes it smooth. Drawing lines during optimization is no longer disabled.
- Graph now recomputes closest pairs only when necessary.

This PR also raises required versions
- pyqtgraph to 0.13.1; segmented line mode may also be useful for speeding up other widgets.
- (oldest) pyqt >=15.2 (I don't know what was this needed; it used to be 15.* and pip couldn't resolve it?!)
- (oldest) numpy ==  1.20, because pyqtgraph is not compatible with 1.19.5, which was pinned previously

https://user-images.githubusercontent.com/2387315/205372895-ef01cbe1-95ca-46bd-b1f6-35faf8903ca9.mov

https://user-images.githubusercontent.com/2387315/205374233-2f27cc14-073b-461e-9f8b-8f0536d15973.mov

All tests still pass and given that changes are related to speed of drawing, timing of updates and layout, I don't see what to test or how to do it.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -1,14 +1,16 @@
 # pylint: disable=too-many-ancestors
+import time
 from types import SimpleNamespace as namespace
 
 import numpy as np
 import scipy.spatial.distance
 
-from AnyQt.QtCore import QTimer
-from AnyQt.QtTest import QSignalSpy
-from AnyQt.QtWidgets import QSizePolicy, QGridLayout, QLabel
+from AnyQt.QtGui import QIcon
+from AnyQt.QtWidgets import QSizePolicy, QGridLayout, QLabel, QPushButton
 
 import pyqtgraph as pg
+
+from orangecanvas.gui.svgiconengine import SvgIconEngine
 
 from Orange.data import ContinuousVariable, Domain, Table, StringVariable
 from Orange.data.util import array_equal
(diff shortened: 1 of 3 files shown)


BlazZupan:
Works as intended. I have also tested it on slightly bigger data sets. @VesnaT, if everything ok codewise, please merge.

VesnaT:
I'd be happy to merge. Maybe just wait for the tests to pass.

BlazZupan:
In trying this again, the animation works sometimes, but then it stops working alltogether. That is, I press Randomize, then choose Every iteration, press Start and then widget does not display any animation but just shows the final visualisation. I used the zoo data set from Datasets widget for this.

janezd:
@BlazZupan , in the meantime, I changed the way I freeze the thread. I used to use `QTimer` and @VesnaT pointed out that Qt complained about this. The current version now uses `sleep`. Could you please check whether the code you are running contains `time.sleep(wait)`? If not, please checkout the latest version.

PR #6241, opened 2022-12-07
Introduce DaskRowInstance

markotoplak:
##### Issue
When starting with dask we hacked RowInstance so that we could print dask tables

##### Description of changes
A nicer implementation.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/dask.py
+++ b/Orange/data/dask.py
@@ -1,5 +1,7 @@
 import contextlib
 import pickle
+import warnings
+from numbers import Integral
 
 import h5py
 import dask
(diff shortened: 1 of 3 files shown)


markotoplak:
@lanzagar, remember when we first opened and printed out the dask table? We mixed dask into our ordinary tables and I am trying to remove this. Because you were with me when we wrote the original could, could you review this PR, please?

lanzagar:
This has two calls of `compute` - one in init and one in `_str_array`. The latter is probably unnecessary since the already computed values could be used. This needs some changes on Orange first however.
As discussed, let's wait for the refactoring there first?

Orange/data/dask.py
```diff
@@ -1,13 +1,29 @@
...
+from Orange.data import Table, RowInstance, Instance
+
+
+class DaskRowInstance(RowInstance):
+
+    def __init__(self, table, row_index):
+        """
+        Construct a data instance representing the given row of the table.
+        """
+        warnings.warn("Creating Instances from DaskTables is inefficient", stacklevel=2)
```
lanzagar, line 21: If we raise warnings every time someone uses this class we should make it clear it exists only for fallback/compatibility reasons. In that case, I would prefer Orange's warnings that fail in tests so we can make sure our existing and new code will not be using this. Intended uses should be covered in tests with explicit assertWarns.

PR #6245, opened 2022-12-08
Distance File: Read Excel files

### LINKED ISSUE #6227: Allow Distance File to read Excel matrices
**In a nutshell**

I here propose that Distance File should read square matrices from excel that contain row and column labels. I propose that this would be more user-friendly than the current solution and support lay audience with more accessible form and load the distance data.

**What's your use case?**

The [Distance File](https://orange3.readthedocs.io/projects/orange-visual-programming/en/latest/widgets/unsupervised/distancefile.html) widget should read the files that specify distances. Currently, the only input format that Orange supports is a private Orange format whose description I only find in the code of dismatrix.py's [from_file](https://github.com/biolab/orange3/blob/master/Orange/misc/distmatrix.py#L107) class. One can obtain an example of such a file through the following workflow, where I have used the subset of the grades data,

image

This workflow creates an Orange distance, .dst file, that I can load with the Distance File widget. Displayed in the text editor, the matrix is straightforward, but its creation and use of .dst format are not for novice users.

image

**The proposal**

I propose to modify both the widget and [from_file](https://github.com/biolab/orange3/blob/master/Orange/misc/distmatrix.py#L107) class to support loading distance matrices from Excel files. I propose these matrices should be in a simple format with row and column labels. An example of such a distance matrix is [travel distances between cities in Europe](https://www.mapcrow.info/european_travel_distance.html), which I could copy to Excel and save as a file. 

image

Currently, there is no way I can load such a distance matrix in Orange, and it would take at least some work and conversion of the file types to prepare the file from this data that Orange could load.

janezd:
##### Issue

Distance file format is unfriendly. Orange should be able to read Excel files. Fixes #6227.

##### Description of changes

So far, a provisional reader. It expects labels in the first row and column, a rectangle of numbers, and doesn't check for errors. Non-resilient and untested.

**To do**
- [x] treat empty cells as zeros
- [x] treat triangular matrices as symmetric
- [x] detect symmetricity and 
- [x] error reporting: when reading fails for a particular cells, report its coordinates
- [ ] ~~sheet selection? (I hate that)~~ Not in this PR.
- [ ] ~~store data directly into numpy, not in python lists~~ Not really doable.
- [x] detect presence of labels
- [x] if labels are given only in rows/columns, and the matrix is symmetric, assume these are also column/row labels

**and also, not necessarily in this PR**
- [x] Excel writer
- [x] ~~input/output previewer for distance matrices~~ In #6263.

##### Includes
- [X] Code changes
- [x] Tests
- [x] Documentation


VesnaT:
If I open something like:
image


```
  File "/Users/vesna/orange3/Orange/widgets/utils/itemdelegates.py", line 76, in sizeHint
    t2 = self.template(vmax, self.ndecimals)
  File "/Users/vesna/orange3/Orange/widgets/utils/itemdelegates.py", line 61, in template
    ndigits = int(math.ceil(math.log10(abs(value) + 1)))
ValueError: cannot convert float NaN to integer
```

VesnaT (changes requested):

PR #6249, opened 2022-12-09
[FIX] pca: n_features_ attribute of decomposition.PCA is deprecated in favor of n_features_in_

### LINKED ISSUE #6248: API change in scikit-learn latest release
Release notes for scikit-learn == 1.2.0 : https://scikit-learn.org/stable/whats_new/v1.2.html

> API Change The n_features_ attribute of [decomposition.PCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html#sklearn.decomposition.PCA) is deprecated in favor of n_features_in_ and will be removed in 1.4. [#24421](https://github.com/scikit-learn/scikit-learn/pull/24421) by [Kshitij Mathur](https://github.com/Kshitij68).

Error msg when running tests for Louvain Clustering
```

  File "/Users/jakakokosar/Documents/dev/orange/orange3/Orange/widgets/unsupervised/owlouvainclustering.py", line 301, in __on_done
    result = future.result()
  File "/Users/jakakokosar/.pyenv/versions/3.10.8/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/Users/jakakokosar/.pyenv/versions/3.10.8/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/Users/jakakokosar/.pyenv/versions/3.10.8/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/jakakokosar/Documents/dev/orange/orange3/Orange/widgets/unsupervised/owlouvainclustering.py", line 613, in run_on_data
    data = res.pca_projection = pca(data)(data)
  File "/Users/jakakokosar/Documents/dev/orange/orange3/Orange/projection/base.py", line 50, in __call__
    clf = self.fit(data.X, data.Y)
  File "/Users/jakakokosar/Documents/dev/orange/orange3/Orange/projection/pca.py", line 268, in fit
    proj = proj.fit(X, Y)
  File "/Users/jakakokosar/.pyenv/versions/orange/lib/python3.10/site-packages/sklearn/decomposition/_pca.py", line 435, in fit
    self._fit(X)
  File "/Users/jakakokosar/Documents/dev/orange/orange3/Orange/projection/pca.py", line 142, in _fit
    return self._fit_truncated(X, n_components, self._fit_svd_solver)
  File "/Users/jakakokosar/Documents/dev/orange/orange3/Orange/projection/pca.py", line 204, in _fit_truncated
    self.n_samples_, self.n_features_ = n_samples, n_features
AttributeError: can't set attribute 'n_features_'

```

JakaKokosar:
##### Issue

Fixes #6248

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/pca.py
+++ b/Orange/projection/pca.py
@@ -99,12 +99,12 @@ class ImprovedPCA(skl_decomposition.PCA):
     # pylint: disable=too-many-branches
     def _fit(self, X):
         """Dispatch to the right submethod depending on the chosen solver."""
-        X = check_array(
+        X = self._validate_data(
             X,
-            accept_sparse=["csr", "csc"],
             dtype=[np.float64, np.float32],
-            ensure_2d=True,
-            copy=self.copy,
+            reset=False,
+            accept_sparse=["csr", "csc"],
+            copy=self.copy
         )
 
         # Handle n_components==None
@@ -201,7 +201,7 @@ class ImprovedPCA(skl_decomposition.PCA):
                 random_state=random_state,
             )
 
-        self.n_samples_, self.n_features_ = n_samples, n_features
+        self.n_samples_ = n_samples
         self.components_ = V
         self.n_components_ = n_components
 
@@ -221,12 +221,12 @@ class ImprovedPCA(skl_decomposition.PCA):
     def transform(self, X):
         check_is_fitted(self, ["mean_", "components_"], all_or_any=all)
 
-        X = check_array(
+        X = self._validate_data(
             X,
             accept_sparse=["csr", "csc"],
             dtype=[np.float64, np.float32],
-            ensure_2d=True,
-            copy=self.copy,
+            reset=False,
+            copy=self.copy
         )
 
         if self.mean_ is not None:


Orange/projection/pca.py
```diff
@@ -201,7 +201,7 @@ def _fit_truncated(self, X, n_components, svd_solver):
                 random_state=random_state,
             )
 
-        self.n_samples_, self.n_features_ = n_samples, n_features
+        self.n_samples_ = n_samples
```
pavlin-policar, line 204: I think this works because:

Most of the code in `ImprovedPCA` is copied directly from `sklearn.decomposition.PCA`. This was done because it wasn't possible to add the randomized algorithm that enables PCA on sparse data without modifying these functions.
Until recently, sklearn's PCA used `self.n_features_`. However, they recently switched over to using `self.n_features_in`, which is more generally used in sklearn, and is part of their `BaseEstimator` class. The `self.n_features_in` attribute is set in [`BaseEstimator._check_n_features`](https://github.com/scikit-learn/scikit-learn/blob/0266481aeb736e5f2c9bf07c4ebb016845d50eb9/sklearn/base.py#L318-L362), which is, in turn, called by [`BaseEstimator._validate_data`](https://github.com/scikit-learn/scikit-learn/blob/0266481aeb736e5f2c9bf07c4ebb016845d50eb9/sklearn/base.py#L453-L560). In scikit-learn's PCA, this method is called in [`PCA._fit`](https://github.com/scikit-learn/scikit-learn/blob/0266481aeb736e5f2c9bf07c4ebb016845d50eb9/sklearn/decomposition/_pca.py#L485-L487), which ensures that the `self.n_features_in` attribute is set on sklearn's implementation.

They also deprecated `self.n_features` and replaced it with a [readonly property](https://github.com/scikit-learn/scikit-learn/blob/0266481aeb736e5f2c9bf07c4ebb016845d50eb9/sklearn/decomposition/_pca.py#L408-L414) which just returns `self.n_featuers_in` anyways. This is still the same functionality as before, but you have to trace it through these different methods. So, this is still backwards compatible.

So, we can do the same thing in our `ImprovedPCA` class. We can get rid of `self.n_features_ = features`, since this will be set when we call `self._validate_data`, which is inherited from `sklearn.decomposition.PCA <- sklearn.BaseEstimator`. So, the other change we need is to replace our previous call to [`check_array`](https://github.com/biolab/orange3/blob/master/Orange/projection/pca.py#L102-L108) to `self._validate_data`. And I believe this should ensure the exact same functionality as before.

pavlin-policar: The `transform` method should also probably be updated to reflect how [sklearn is doing things now](https://github.com/scikit-learn/scikit-learn/blob/0266481aeb736e5f2c9bf07c4ebb016845d50eb9/sklearn/decomposition/_base.py#L120).

So, this would mean changing

```python
        X = check_array(
            X,
            accept_sparse=["csr", "csc"],
            dtype=[np.float64, np.float32],
            ensure_2d=True,
            copy=self.copy,
        )
```

on lines 224-230 to

```python
X = self._validate_data(
    X, dtype=[np.float64, np.float32], reset=False, check_params=dict(accept_sparse=["csr", "csc"])
)
```

markotoplak: @pavlin-policar, thanks for the explanation.

markotoplak:
Before merging this please ensure that new code also properly runs with the oldest supported scikit-learn. The `-oldest` tests do not fail, so it likely works, but double-check, please.

For now, I would avoid raising the dependency to 1.2. I am OK with raising it to 1.1 if needed.

markotoplak:
I couldn't wait. :) Please, add a commit that reverts #6255 when merging.

pavlin-policar:
The PCA changes in this PR look good to me, but tests related to AdaBoost are failing somewhat consistently. Could this be due to a change in the newer version of scikit-learn? If so, we probably want to include that here as well.

markotoplak:
The remaining tests are failing because of scikit-learn/scikit-learn/issues/26241. I also intend to do a PR that fixes the issue.

I'll remove the last commit (allowing scikit-learn 1.2) and merge this as is and then we can fix adaboost separately.

PR #6267, opened 2022-12-21
[ENH] Faster TableModel for Dask tables (through caching)

noahnovsak:
##### Issue
Currently the data is read one row at a time, which causes issues when the dataset is very wide.

##### Description of changes
Updated `RichTableModel` with a new mechanism to read only the data currently in view and cache it. This only gets used for `DaskTables`. Otherwise if the data is stored in `numpy` arrays it reads each element from them directly, and if the data is sparse it will default to the old behaviour of using `RowInstance`. This is done because the entire row must be read in order for the "basket_formatter" to work. I haven't tested or considered what should happen for `SQLTable`, but I'm thinking we also default to the old behaviour?

also a small change in `test_owtable` because the sorting test was failing.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -787,6 +787,7 @@ class ModelActionsWidget(QWidget):
         return self.insertAction(-1, action, *args)
 
 
+# pylint: disable=abstract-method
 class ModelCache(QObject, ConcurrentMixin):
     def __init__(self, model):
         super().__init__()
@@ -800,7 +801,7 @@ class ModelCache(QObject, ConcurrentMixin):
                      "height": 60,
                      "width": 15}
 
-        self.delayed_request = QTimer()
+        self.delayed_request = QTimer(self)
         self.delayed_request.timeout.connect(self.update)
         self.last_cache_time = 0
 
@@ -860,9 +861,6 @@ class ModelCache(QObject, ConcurrentMixin):
             self.update_view(item)
             self.fetch_data()
             return None
-
-    def on_partial_result(self, result):
-        pass  # can dask report progress?
 
     def on_done(self, result):
         top, left, bottom, right = result
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -1,3 +1,4 @@
+from time import perf_counter
 from numbers import Number, Integral
 from math import isnan, isinf
 
@@ -12,7 +13,7 @@ from xml.sax.saxutils import escape
 
 from AnyQt.QtCore import (
     Qt, QObject, QAbstractListModel, QModelIndex,
-    QItemSelectionModel, QItemSelection)
+    QItemSelectionModel, QItemSelection, QTimer)
 from AnyQt.QtCore import pyqtSignal as Signal
 from AnyQt.QtGui import QColor, QBrush
 from AnyQt.QtWidgets import (
@@ -21,6 +22,7 @@ from AnyQt.QtWidgets import (
 
 import numpy
 
+from orangewidget.utils.cache import LRUCache
 from orangewidget.utils.itemmodels import (
     PyListModel, AbstractSortTableModel as _AbstractSortTableModel,
     LabelledSeparator, SeparatorItem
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owtable.py
```diff
@@ -69,24 +74,157 @@ def __init__(self, sourcedata, parent=None):
...
+        value = self.get_value(row, col)
+        if value is None:
+            return
+        coldesc = self.columns[col]
+
+        if role == Qt.DisplayRole:
+            return coldesc.var.str_val(value)
+        elif role in (self.ValueRole, Qt.EditRole) and isinstance(coldesc, self.Column):
+            return Value(coldesc.var, value)
+        elif role == Qt.BackgroundRole:
```
markotoplak, line 200: I understand you need to re-implement this function for DisplayRole and ValueRole and all roles defined in the previous version, but couldn't you call the super, but couldn't you call the superclass for the others?

Or are the other reasons why you put them here? Perhaps because you'd like to move this code into TableModel?

noahnovsak: As soon as I call super().data() it will read the entire row though. So I need to avoid it entirely and rewrite all of the code or find a clean way to implement it into the base class (with tests and all).

Orange/widgets/data/owtable.py
```diff
@@ -46,9 +46,14 @@
...
 from Orange.widgets.utils.state_summary import format_summary_details
+from Orange.widgets.utils.concurrent import ConcurrentMixin
+from orangewidget.utils.cache import LRUCache
+
 
+Viewport = namedtuple("Viewport", ("top", "left", "bottom", "right"))
 
-class RichTableModel(TableModel):
+
+class RichTableModel(TableModel, ConcurrentMixin):
```
markotoplak, line 56: This class now inherits ConcurrentMixin. I think you should instead make a new class that handles concurrent stuff and then you encapsulate it into the class as an attribute, which you call. Reading this code is now slightly harder because (in many cases optional) concurrency handling functions (which you even can not rename but to how the Mixin works) are mixed with others.

markotoplak: I suggest also including the cache within that new class.

Orange/widgets/data/owtable.py
```diff
@@ -69,24 +74,157 @@ def __init__(self, sourcedata, parent=None):
...
 
+        self.Y_dim = len(self.domain.class_vars)
+        self.M_dim = len(self.domain.metas)
+
+        self.cache = LRUCache(10000)
+        self.viewport = Viewport(0, 0, 0, 0)
+
+        self.view = view  # ugly but I need it to know what to cache
+
+        self.timer = QTimer()
```
markotoplak, line 85: this timer should have a better name :)

Orange/widgets/data/owtable.py
```diff
@@ -69,24 +74,157 @@ def __init__(self, sourcedata, parent=None):
...
+                else:
+                    self.cache[(view_row, view_col)] = X[chunk_row][chunk_col - Y_inview - M_inview]
+
+    def request_data(self):
+        cache_interval = 0.02
+        elapsed = perf_counter() - self.last_cache_time
+        self.timer.stop()
+        self.timer.start(max(1, int((cache_interval - elapsed) * 1000)))
+
+    def on_timeout(self):
```
markotoplak, line 145: Needs a better name.

Orange/widgets/data/owtable.py
noahnovsak: I can't come up with anything more descriptive. Any ideas?

markotoplak: Perhaps `delayed_update`?

Orange/widgets/data/owtable.py
markotoplak: Cool, this looks much more self-contained. Thanks!

noahnovsak:
waiting on #6271

markotoplak:
We'd also need tests for this one. :)

noahnovsak:
I'm having some trouble writing tests for this one. While it would be trivial to manually call each function and make sure it does what it's supposed to I feel that would be pointless, right?
What I really want is to emulate a TableView and have it scroll back and forth to see if and when the appropriate methods are called. But as I've already said I don't quite know how to do this.

markotoplak:
Yes, testing code like this is hard.

The tests you describe are not pointless. If tests cover most of new the code, we'll know if (or better, when) we break stuff (like if we change DaskTable class and a breaking tests indicates we also need to change something here).  Do not spend too much time on it though.

If you still think tests like this are superfluous, then I'll respect your decision.

PR #6271, opened 2022-12-23
TableModel avoid RowInstance

noahnovsak:
##### Description of changes
Trying to avoid the use of `RowInstance` in anticipation of upcoming dask integration (see #6267).
Includes preliminary tests.

### DIFF
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -844,7 +844,7 @@ class TableModel(AbstractSortTableModel):
 
         def make_basket(vars, density, role):
             return TableModel.Basket(
-                vars, TableModel.Attribute, brush_for_role[role], density,
+                vars, role, brush_for_role[role], density,
                 make_basket_formatter(vars, density)
             )
 
@@ -903,8 +903,7 @@ class TableModel(AbstractSortTableModel):
         if isinstance(coldesc, self.Basket):
             # `self.source[row:row + 1]` returns Table
             # `self.source[row]` returns RowInstance
-            if coldesc.role is self.ClassVar:
-                return self.source[row:row + 1].Y
+            # We only worry about X and metas, as Y cannot be sparse
             if coldesc.role is self.Meta:
                 return self.source[row:row + 1].metas
             if coldesc.role is self.Attribute:
--- a/Orange/widgets/utils/itemmodels.py
+++ b/Orange/widgets/utils/itemmodels.py
@@ -819,8 +819,6 @@ class TableModel(AbstractSortTableModel):
         self.X_density = sourcedata.X_density()
         self.Y_density = sourcedata.Y_density()
         self.M_density = sourcedata.metas_density()
-        self.Y_len = len(domain.class_vars)
-        self.M_len = len(domain.metas)
 
         brush_for_role = {
             role: QBrush(c) if c is not None else None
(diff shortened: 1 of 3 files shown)


noahnovsak:
@ales-erjavec could you take a look please? Would changes like this work or am I missing something?

Orange/widgets/utils/itemmodels.py
```diff
@@ -819,39 +819,35 @@ def __init__(self, sourcedata, parent=None):
...
 
-        def format_sparse_bool(vars, datagetter, instance):
-            data = datagetter(instance)
+        def format_sparse_bool(vars, data):
+            data = data.tocsr()
             return ", ".join(vars[i].name for i in data.indices)
 
-        def format_dense(var, instance):
-            return str(instance[var])
+        def format_dense(var, data):
```
markotoplak, line 839: instead of `data` a likely better name would be `value`

Orange/widgets/utils/itemmodels.py
```diff
@@ -819,39 +819,35 @@ def __init__(self, sourcedata, parent=None):
...
 
         brush_for_role = {
             role: QBrush(c) if c is not None else None
             for role, c in self.ColorForRole.items()
         }
 
-        def format_sparse(vars, datagetter, instance):
-            data = datagetter(instance)
-            return ", ".join("{}={}".format(vars[i].name, vars[i].repr_val(v))
+        def format_sparse(vars, data):
```
markotoplak, line 830: instead of `data` a likely better name would be `row`

Orange/widgets/utils/itemmodels.py
```diff
@@ -819,39 +819,35 @@ def __init__(self, sourcedata, parent=None):
...
+            data = data.tocsr()
+            return ", ".join("{}={}".format(vars[i].name, vars[i].str_val(v))
                              for i, v in zip(data.indices, data.data))
 
-        def format_sparse_bool(vars, datagetter, instance):
-            data = datagetter(instance)
+        def format_sparse_bool(vars, data):
```
markotoplak, line 835: instead of `data` a likely better name would be `row`

Orange/widgets/utils/itemmodels.py
```diff
@@ -910,6 +901,25 @@ def row_instance(index):
...
+        elif self.Y_len <= col < self.Y_len + self.M_len:
+            if self.M_density == Storage.DENSE:
+                return self.source.metas[row, col - self.Y_len]
+            else:
+                return self.source.metas[row]
+        else:
+            if self.X_density == Storage.DENSE:
+                return self.source.X[row, col - self.Y_len - self.M_len]
+            else:
+                return self.source.X[row]
```
ales-erjavec, line 921: AFAIK accessing X|Y|metas in case of an SQLTable will download a limited number (10000) rows into memory on first access.  Indexing the arrays after that with row indices larger then the download limit (10000+) will raise an IndexError. @lanzagar is SQL still even supported?

markotoplak:
@noahnovsak, SQL tests work! :)

Orange/widgets/utils/tests/test_itemmodels.py
```diff
@@ -528,5 +530,61 @@ def tip(i):
         self.assertEqual(tip(1), "footip")
 
 
+class TestTableModel(unittest.TestCase):
```
markotoplak, line 533: I suggest that TestTableModel gets a test of SQL data too. Just to ensure that these two play well together.

markotoplak:
Now that you became our local SQL master, I suggest adding a SQL data test to tests of OWTable, because the widget does a bit more than just show the item model.

noahnovsak:
@lanzagar can you take another look at this not that some of the minor changes have been merged?

Issue #6265, opened 2022-12-20
Feature constructor: allow use of temporary variables


wvdvegte:
**What's your use case?**
I've been experimenting with [ChatGPT](https://openai.com/blog/chatgpt/) to help me construct code to solve difficult stuff with the one-line Python syntax in Feature Constructor. Sometimes it comes up with solutions that I can use straight away, like
<img width="365" alt="image" src="https://user-images.githubusercontent.com/55989717/208628146-7878689a-494d-486f-83eb-762b783ede23.png">
but sometimes it comes with solutions introducing an intermediate variable. For instance, if I want to separate the number from a string like "28.665 °C" it suggests solutions like

number := ''.join(filter(lambda c: c.isdigit() or c == '.', Temperature))

which cannot be processed by Feature Constructor because of the unknown variable c.

**What's your proposed solution?**
Would it be possible to allow the use of such intermediate variables in Feature Constructor?

**Are there any alternative solutions?**
As always, use Python Script

PR #6272, opened 2022-12-23
[ENH] Feature Constructor: Enable comprehensions and lambdas in expressions

ales-erjavec:
##### Issue

Fixes https://github.com/biolab/orange3/issues/6265

##### Description of changes

Enable comprehensions and lambdas in expressions

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfeatureconstructor.py
+++ b/Orange/widgets/data/owfeatureconstructor.py
@@ -1182,9 +1182,9 @@ __ALLOWED = [
     "bin", "bool", "bytearray", "bytes", "chr", "complex", "dict",
     "divmod", "enumerate", "filter", "float", "format", "frozenset",
     "getattr", "hasattr", "hash", "hex", "id", "int", "iter", "len",
-    "list", "map", "memoryview", "next", "object",
+    "list", "map", "max", "memoryview", "min", "next", "object",
     "oct", "ord", "pow", "range", "repr", "reversed", "round",
-    "set", "slice", "sorted", "str", "tuple", "type",
+    "set", "slice", "sorted", "str", "sum", "tuple", "type",
     "zip"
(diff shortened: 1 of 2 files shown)


janezd:
It works for `sum([x ** 2 for x in range(5)])`, but `sum(x ** 2 for x in range(5))` fails with 

```
  File "/Users/janez/orange3/Orange/data/table.py", line 780, in from_table
    out = array_conv.get_columns(source, row_indices, n_rows,
  File "/Users/janez/orange3/Orange/data/table.py", line 317, in get_columns
    col_array = match_density(_compute_column(col, sourceri))
  File "/Users/janez/orange3/Orange/data/table.py", line 213, in _compute_column
    col = func(*args, **kwargs)
  File "/Users/janez/orange3/Orange/widgets/data/owfeatureconstructor.py", line 1267, in __call__
    return self.__call_table(table)
  File "/Users/janez/orange3/Orange/widgets/data/owfeatureconstructor.py", line 1291, in __call_table
    return np.asarray(y, dtype=self.dtype)
TypeError: float() argument must be a string or a number, not 'generator'
```

If this is difficult to fix, I'm OK with it - say so and I merge.

ales-erjavec:
This is a problem with `sum` definition in the

https://github.com/biolab/orange3/blob/master/Orange/widgets/data/owfeatureconstructor.py#L1214 
```python
    ...
    "sum": lambda *args: np.sum(args),
```

i.e.
```python
sum = lambda *args: np.sum(args)
print(sum(x ** 2 for x in range(5)))
<generator object <genexpr> at 0x1359e5770>
```

janezd:
> I guess the builtin overrides should be removed?

Yes.

PR #6276, opened 2023-01-04
[FIX] Pythagorean Forest: Fix report

### LINKED ISSUE #6273: Pythagorean Forest: report does not work
**What's wrong?**

Clicking the report button produces an error. The problem is that widget declares `graph_name = "scene"`, although it doesn't have a "scene" but uses a left-to-right wrapping list view instead.

Give it some data and click the report button.

**What's your environment?**

Irrelevant.

VesnaT:
##### Issue

Fixes #6273

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owpythagoreanforest.py
+++ b/Orange/widgets/visualize/owpythagoreanforest.py
@@ -3,11 +3,13 @@ from math import log, sqrt
 from typing import Any, Callable, Optional
 
 from AnyQt.QtCore import Qt, QRectF, QSize, QPointF, QSizeF, QModelIndex, \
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owpythagoreanforest.py
```diff
@@ -23,6 +25,53 @@
...
+
+@media screen and (max-width: 600px) {
+  .forest_model_col {
+    flex: 100%;
+    max-width: 100%;
+  }
+}
+</style>
+"""
+
```
janezd, line 73: Fancy!

PR #6277, opened 2023-01-05
Dask: KMeans

noahnovsak:
Add `dask_ml.cluster.KMeans` as an alternative to `sklearn.cluster.KMeans` for dask arrays.

Probably worth mentioning, dask_ml adds the "k-means||" init option but defaults to the sklearn implementation in the case of "k-means++" or "random" initialization. So using "k-means||" is what enables working with larger datasets at all. However, smaller datasets are still processed much faster just using sklearn directly as there seems to be a lot of overhead with dask_ml.

### DIFF
--- a/Orange/clustering/kmeans.py
+++ b/Orange/clustering/kmeans.py
@@ -20,6 +20,8 @@ class KMeansModel(ClusteringModel):
         self.k = projector.get_params()["n_clusters"]
 
     def predict(self, X):
+        if isinstance(X, da.Array):
+            X = X.rechunk({0: "auto", 1: -1})
         return self.projector.predict(X)
 
 
@@ -48,7 +50,10 @@ class KMeans(Clustering):
                 import dask_ml.cluster
 
                 del params["n_init"]
-                assert params["init"] == "k-means||"
+                if params["init"] != "k-means||":
+                    warnings.warn(f"Initializing with {params['init']} defaults"
+                                  f" to sklearn. Using k-means|| instead.")
+                    params["init"] = "k-means||"
 
                 X = X.rechunk({0: "auto", 1: -1})
                 __wraps__ = dask_ml.cluster.KMeans
--- a/Orange/clustering/kmeans.py
+++ b/Orange/clustering/kmeans.py
@@ -1,5 +1,8 @@
 import warnings
+from typing import Union
 
+import numpy as np
+import dask.array as da
 import sklearn.cluster
 
 from Orange.clustering.clustering import Clustering, ClusteringModel
@@ -36,6 +39,24 @@ class KMeans(Clustering):
         super().__init__(
             preprocessors, {k: v for k, v in vars().items()
                             if k != "compute_silhouette_score"})
+
+    def fit(self, X: Union[np.ndarray, da.Array], y: np.ndarray = None):
+        params = self.params.copy()
+        __wraps__ = self.__wraps__
+        if isinstance(X, da.Array):
+            try:
+                import dask_ml.cluster
+
+                del params["n_init"]
+                assert params["init"] == "k-means||"
+
+                X = X.rechunk({0: "auto", 1: -1})
+                __wraps__ = dask_ml.cluster.KMeans
+
+            except ImportError:
+                warnings.warn("dask_ml is not installed. Using sklearn instead.")
+
+        return self.__returns__(__wraps__(**params).fit(X))
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


Orange/clustering/kmeans.py
```diff
@@ -27,15 +28,20 @@ class KMeans(Clustering):
 
     def __init__(self, n_clusters=8, init='k-means++', n_init=10, max_iter=300,
                  tol=0.0001, random_state=None, preprocessors=None,
-                 compute_silhouette_score=None):
+                 compute_silhouette_score=None, wrap_dask_ml=False):
```
markotoplak, line 31: Ideally, the same class would handle Table and DaskTable without any new parameters, but sometimes this is hard.

I think we should avoid adding parameters for dask. Perhaps a subclass could be a better solution in this case?

I think we could also check whether to go for a solution Orange often uses when same methods can handle both classification and regression (which are seperate in scikit-lean). We often have a "main" class which then calls separate "worker" classes as needed, depending on whether it is given a classification or regression problem. Could you please  study how `modelling.RandomForestsLearner` works, just to get a feel for it?

I think that the first step to a solution would be to make a subclass for Dask and then we could perhaps check how to join them.

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -584,8 +599,18 @@ def set_data(self, data):
...
+                self.init_method.insertItem(0, self.EXTRA_INIT_METHODS[0][0])
+                self.init_method.setCurrentIndex(self.smart_init)
+            elif old_data is not None:
+                self.INIT_METHODS = self.BASE_INIT_METHODS
+                self.init_method.removeItem(0)
+                self.smart_init = self.init_method.currentIndex()
+
         # Do not needlessly recluster the data if X hasn't changed
-        if old_data and self.data and array_equal(self.data.X, old_data.X):
+        if old_data and self.data and array_equal(self.data.X, old_data.X):  # could be an issue for dask
```
markotoplak, line 613: This is a good find. I remember you telling me about it, but then I forgot. I do not have a good solution for now, but we would need to ensure (and test) at the minimum that `array_equal` does not load the whole data set into memory.

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -466,6 +471,16 @@ def invalidate(self, unconditional=False):
         else:
             self.commit.deferred()
 
+    def on_init_changed(self):
+        self.invalidate()
+
+        self.reruns.setEnabled(True)
+        if isinstance(self.data, DaskTable):
+            if self.INIT_METHODS[self.smart_init][1] != "k-means||":
+                self.Warning.default_to_sklearn(self.INIT_METHODS[self.smart_init][1])
```
markotoplak, line 480: A warning is good, but I'd rather have the option to come here at all disabled. Initialization is probably not that big of a deal that we'd want a specific method so badly.

janezd:
1. In Sample Data we have an checkbox `Stratify (if possible)`. Is this an option here? Combos (I would actually prefer radio buttons, but this may be my current temporary preference) and the one that is not always applicable would indicate what it defaults to. At least in a tooltip.

    Option "O" would remain chosen, but a warning would indicate that it was not used.

Another thing: this is only available for Dask, so it will be off for majority of users. If so, I would disable it when inapplicable. If chosen, I would keep it chosen (but disabled!), and have a warning etc. So the user may manually choose another option, or keep the disabled one, which won't work anyway.

An alternative to above would be to hide the option if it is unavailable and not chosen.

Orange/widgets/unsupervised/owkmeans.py
```diff
@@ -133,18 +146,20 @@ class Warning(widget.OWWidget.Warning):
...
     resizing_enabled = False
 
+    settingsHandler = TableTypeContextHandler()
     k = Setting(3)
     k_from = Setting(2)
     k_to = Setting(8)
     optimize_k = Setting(False)
     max_iterations = Setting(300)
     n_init = Setting(10)
-    smart_init = Setting(0)  # KMeans++
```
markotoplak, line 147: This change, if we go with it, also requires settings migration. Just leaving it here as a note.

Orange/widgets/unsupervised/owkmeans.py
noahnovsak: Turns out np.allclose works fine on dask. Comparing an 8GB array against itself took a little over two seconds (this is worst case scenario, runing in an ipython shell that doesn't play well with dask) and memory usage stayed very much within reasonable limits.

markotoplak:
@noahnovsak, could you please rebase this to the now-rebased dask branch? The easiest way is to start with that branch and just move the new commits there (otherwise there will be conflicts).

markotoplak:
Thanks. We need at least some basic tests for this.

PR #6279, opened 2023-01-05
Fixes for translations

janezd:
##### Description of changes

Fix messages that can't be translated (mostly because of plural forms) and some other minor related corrections.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/report/report.py
+++ b/Orange/widgets/report/report.py
@@ -86,16 +86,19 @@ def describe_domain(domain):
     :rtype: OrderedDict
     """
 
-    def clip_attrs(items, s):
-        return clipped_list([a.name for a in items], 1000,
-                            total_min=10, total=" (total: {{}} {})".format(s))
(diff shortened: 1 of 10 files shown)


janezd:
"Needs discussion": Fail on "Oldest dependencies" (orange-canvas' localization)

Orange/widgets/evaluate/owpredictions.py
```diff
@@ -136,7 +136,7 @@ def __init__(self):
             gui.checkBox(
                 predopts, self, "show_probability_errors",
                 "Show classification errors",
-                tooltip="Show 1 - probability assigned to the correct class",
+                tooltip="Show probability assigned to the correct class",
```
lanzagar, line 139: Looked correct to me the way it was before?
The error is the sum of all incorrect probs i.e. 1 - prob of correct.
The wording however could be improved.

PR #6282, opened 2023-01-06
[ENH] Improve scorer selection in Test and Score and predictions

### LINKED ISSUE #6131: Test and Score - intuitive way to add new score
##### Issue

In Test and Score, the menu to add additional scores is hidden (regular users cannot easily find it):
Screenshot 2022-09-09 at 10 26 18

##### Suggestion

Add a sign that would suggest to the user that it is possible to add additional scores. For example, tree dots:
Screenshot 2022-09-09 at 10 26 181

janezd:
##### Issue

- `ScoreTable`, which is used in Test and Score, and in Predictions, has a hard coded list of scorers, which defines the (default) visibility and the order. New scores are thus invisible by default.
- The names of classes are used for names of columns. It is thus impossible to translate them. A `name` attribute exists, but is not defined and defaults to class name.
- The user may not be aware that (s)he can choose additional scores by right-clicking the header.

##### Description of changes

- Scorers define default visibility and priority as class attributes.
- Visibility settings are stored as "hints", so new scorers (e.g. from add-ons) may be visible if they that's their default.
- The `name` attribute is now defined, but in settings the scorers are referred to by their class names, so settings "survive" the change of language.
- Add an indicator that can be clicked and opens a popup for selection of scores (fixes #6131).

    Although currently in `Orange.widgets.evaluate.utils`, the class `SelectableColumnsHeader` is general and can be moved somewhere else and used, e.g. in `OWRanks`.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/evaluation/scoring.py
+++ b/Orange/evaluation/scoring.py
@@ -66,6 +66,9 @@ class Score(metaclass=ScoreMetaType):
     name = None
     long_name = None  #: A short user-readable name (e.g. a few words)
 
+    default_visible = True
+    priority = 100
+
     def __new__(cls, results=None, **kwargs):
         self = super().__new__(cls)
         if results is not None:
@@ -137,6 +140,7 @@ class RegressionScore(Score, abstract=True):
 class CA(ClassificationScore):
     __wraps__ = skl_metrics.accuracy_score
     long_name = "Classification accuracy"
+    priority = 20
 
 
 class PrecisionRecallFSupport(ClassificationScore):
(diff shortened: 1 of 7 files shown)


janezd:
The last commit is questionable. At the very least, it must be checked on different systems and versions of Qt. Possible problems are
- the hack used for `PersistentMenu`
- clicking `...` might also trigger sorting. This can be avoided by disabling sorting and calling sort on click. Unfortunately, Qt enables sorting as soon as there is a visible sorting indicator.
- tests currently fail on clipboard copying. We can change the test and tolerate an additional (empty) column.
-   
Alternative is to put a kind of button on top of the header view. The problem with this solution is that it requires repositioning on resize, I think.

PR #6283, opened 2023-01-07
Scorer column selection

### LINKED ISSUE #6131: Test and Score - intuitive way to add new score

janezd:
##### Issue

Fixes #6131.

Based on #6282.

##### Description of changes

Adds a general class `SelectableColumnsHeader` that can also be used in other places, e.g. in the Rank widget.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/evaluation/scoring.py
+++ b/Orange/evaluation/scoring.py
@@ -66,6 +66,9 @@ class Score(metaclass=ScoreMetaType):
     name = None
     long_name = None  #: A short user-readable name (e.g. a few words)
 
+    default_visible = True
+    priority = 100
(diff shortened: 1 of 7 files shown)


janezd:
Cherry-picked into #6282.

PR #6284, opened 2023-01-08
Resolve Edit Domain (datetime to epoch) timebomb

PrimozGodec:
##### Issue

In Edit Domain has a timebomb suggesting increasing the minimum version of Pandas and removing a workaround for features not supported in older versions.

##### Description of changes
- Removing timebomb
- Removing extra code in the Edit domain and the Group By widgets that changed time to UTC is no longer required since Pandas from version 1.4 supports subtracting different timezones.
- Fix XGB test since `get_params` changed https://github.com/dmlc/xgboost/pull/8634

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/tests/test_owgradientboosting.py
+++ b/Orange/widgets/model/tests/test_owgradientboosting.py
@@ -1,3 +1,4 @@
+import json
 import unittest
 import sys
 from typing import Type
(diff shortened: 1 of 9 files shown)


requirements-core.txt
```diff
@@ -22,7 +22,7 @@ python-louvain>=0.13
 requests
 openTSNE>=0.6.1
 baycomp>=1.0.2
-pandas>=1.3.0,!=1.5.0
+pandas>=1.4.0,!=1.5.0
```
markotoplak, line 25: This should be changed in meta.yaml too.

Orange/widgets/model/tests/test_owgradientboosting.py
```diff
@@ -220,37 +223,39 @@ def test_default_parameters_cls(self):
         booster = XGBRFClassifier()
         model = booster(data)
         params = model.skl_model.get_params()
+        booster_params = json.loads(model.skl_model.get_booster().save_config())
```
markotoplak, line 226: Hahaha, this is a interesting way to get the parameters. I guess this is the only way now. I think we should make sure this also works with the oldest supported xgboost. I see that the "oldest" tests do not test that. Could you also ensure that that gets tested?

requirements-core.txt
PrimozGodec: Done.

Orange/widgets/model/tests/test_owgradientboosting.py
markotoplak: Thanks!

Issue #6229, opened 2022-11-29
Qt6: stored OWRank settings are incompatible


markotoplak:
**What's wrong?**
```
/home/marko/venv310/lib/python3.10/site-packages/orangewidget/settings.py:500: UserWarning: Could not read defaults for widget <class 'Orange.widgets.data.owrank.OWRank'>
The following error occurred:

No module named 'sip'
```

So the problem is with unpickling settings.

I suspect this line:
`sorting = Setting((0, Qt.DescendingOrder))`

**How can we reproduce the problem?** 

Make a Qt5 environment, use Rank. Make a Qt6 environment, open Rank.

**What's your environment?**
Python 3.10, Qt 6.4

PR #6301, opened 2023-01-17
[FIX] Rank - make sorting setting PyQt6 compatible

PrimozGodec:
##### Issue

Probably fixes https://github.com/biolab/orange3/issues/6229

##### Description of changes
Store the sorting setting as an integer and transform it to Qt.SortOrder when passed to model/wiev

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -230,7 +230,7 @@ class OWRank(OWWidget, ConcurrentWidgetMixin):
     sorting = Setting((0, enum2int(Qt.DescendingOrder)))
     selected_methods = Setting(set())
 
-    settings_version = 3
+    settings_version = 4
     settingsHandler = DomainContextHandler()
     selected_attrs = ContextSetting([], schema_only=True)
     selectionMethod = ContextSetting(SelectNBest)
--- a/Orange/widgets/data/owrank.py
+++ b/Orange/widgets/data/owrank.py
@@ -29,6 +29,7 @@ from Orange.widgets.settings import (
     ContextSetting, DomainContextHandler, Setting
 )
 from Orange.widgets.unsupervised.owdistances import InterruptException
+from Orange.widgets.utils import enum2int
 from Orange.widgets.utils.concurrent import ConcurrentWidgetMixin, TaskState
 from Orange.widgets.utils.sql import check_sql_input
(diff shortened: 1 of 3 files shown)


PrimozGodec:
@markotoplak I cannot reproduce the issue you reported, but I fixed it to be PyQt6 compatible now. Can you try if the issue is solved for you?

Orange/widgets/data/owrank.py
```diff
@@ -496,9 +497,11 @@ def on_done(self, result: Results) -> None:
             sort_column, sort_order = self.sorting
             if sort_column < len(labels):
                 # adds 2 to skip the first two columns
-                self.ranksModel.sort(sort_column + 2, sort_order)
+                # in PyQt6 Qt.SortOrder is not int, and so sort_order need to be
+                # transformed to Qt.SortOrder
```
markotoplak, line 501: Could you also comment that here Qt.SortOrder(sort_order) puts whatever this was to the correct thing? :)

PrimozGodec: Done

markotoplak:
This PR does not fix the problem. I still get

```
/home/marko/venv310/lib/python3.10/site-packages/orangewidget/settings.py:500: UserWarning: Could not read defaults for widget <class 'Orange.widgets.data.owrank.OWRank'>
The following error occurred:

No module named 'sip'
  warnings.warn("Could not read defaults for widget {0}\n"
```

after opening Orange with PyQt6 onto settings saved in Orange and PyQt5. But it did fix the opposite problem (saving settings with PyQt6 and opening with PyQt5).

Then I checked what the created defaults file is like. I see `sorting<94>K^@<8c>^Csip<94><8c>^N_unpickle_enum<94><93><94><8c>^LPyQt5.QtCore<94><8c> SortOrder` in it....

I tried making this int even more "inty" by doing  `return int(enum) if isinstance(enum, int) else int(enum.value)` (one additional int statement), but no, no change.

markotoplak:
@PrimozGodec, thanks! I added a commit that just bumps the `settings_version`. Perhaps not needed, but I like settings structure changes noted somewhere.

PrimozGodec:
> @PrimozGodec, thanks! I added a commit that just bumps the `settings_version`. Perhaps not needed, but I like settings structure changes noted somewhere.

I agree with this change.

PR #6305, opened 2023-01-20
Distances: improved GUI

### LINKED ISSUE #6243: Distances widget needs improvements in documentation and interface
**What's your use case?**

When using the Distances widget to compute the correlation between features (columns), the computation returns values inconsistent with the documentation. For instance, the documentation says that for the spearman correlation, the value of the Spearman correlation is scaled to the interval of [0, 1]. It does not say that Orange also substracts the resulting value from 1. If $r$ is the correlation coefficient, Distances returns the matrix where every entry corresponds to $1-(r+1)/2$.

For the user, it may not be clear what is on the output of this widget if we use correlation to estimate distances. We can solve this by:
1. Correcting the documentation (the easiest)
2. Changing the widget settings so Orange can output either the (a) raw scores or (b) scores transformed to distances. (we need to correct the documentation in this case as well)

I would vote for 2, as this is the only widget from which we can get the correlation matrix and then visualize it by, say, Distance Matrix, MDS, or Hierarchical Clustering.

janezd:
##### Issue

Fixes #6243.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owdistances.py
+++ b/Orange/widgets/unsupervised/owdistances.py
@@ -1,6 +1,7 @@
 from typing import NamedTuple, Dict, Type, Optional
 
 from AnyQt.QtWidgets import QButtonGroup, QRadioButton
+from AnyQt.QtCore import Qt
 from scipy.sparse import issparse
 import bottleneck as bn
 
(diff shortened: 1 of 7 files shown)


Orange/widgets/unsupervised/owdistances.py
```diff
@@ -62,18 +101,14 @@ class Inputs:
         data = Input("Data", Orange.data.Table)
 
     class Outputs:
-        distances = Output("Distances", Orange.misc.DistMatrix, dynamic=False)
+        distances = Output("Matrix", Orange.misc.DistMatrix, dynamic=False,
+                           replaces="Distances")
```
VesnaT, line 105: `replaces` parameter should be a list.

PR #6308, opened 2023-01-23
Preprocess.RemoveSparse: Fix defaults and signals

### LINKED ISSUE #6297: Preprocess widget Remove sparse features does nor remove rows with sparse features
**What's wrong?**

On Orange3 3.32 and 3.34 the Preprocess widget, Select Relevant Features dows NOT removes rows with null values.
Moreover, if you have two instances of Select Relevant Features  in the same Preprocess widget, one with zeros and the other with null,
and save the ows and reload it, you find the two instances with null values.

Run Orange3, open error_walmart.ows , in the FeatureFile widget open file features.csv .
Look at FeatureStatistics and find that there are five features with null values more than 50% (Markdown1 - Markdown5)
I build a feature X1 with all values zero
Go to the Preprocess widget and you find two instances with null values. The first was zero and the second was null. Put the first at zero, save the ows, close and reload ad you find again the two at null: the widget does NOT save setting correctly or does NOT reload 
the setting correctly.
Now set again the first at zero and the second at null, and in Feature_datatable(2) you will find that the X1 variable with zero values HAS BEEN deleted while the Markdown1 - Markdown5 columns with more than 50% of null values HAVE NOT been deleted.
**What's your environment?**

- Operating system: UBUNTU LINUX 10.04 LTS 20.04 LTS
- Orange version: 3.32 3.34
- How you installed Orange: with ANACONDA
[error_walmart.zip](https://github.com/biolab/orange3/files/10417305/error_walmart.zip)

janezd:
##### Issue

Fixes #6297.

##### Description of changes

Defaults were mismatched:
- initial button states did not reflect actual parameters, and defaults for the same option were not the same everywhere
- clicking radio button for filter did not emit a change signal
- setting parameter `filter0` did not change the state of the button

This is just about default and signals, so I see no propser way to unit test this.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/data/owpreprocess.py
```diff
@@ -307,10 +307,12 @@ def filterSettingsClicked(self):
 
     def filterByClicked(self):
         self.setFilter0(self.filter_buttons.checkedId())
+        self.edited.emit()
```
VesnaT, line 310: This line is redundant. `setFilter0()` emits the signal, if `self.filter0` has changed.

PR #6309, opened 2023-01-23
[ENH] MDS: Show Kruskal stress

### LINKED ISSUE #6070: MDS: amount of explained variance
**What's your use case?**

I got a comment from a reviewer, that it would be useful if MDS showed the amount of explained variable. Given that most MDS projections start with PCA initialization, this could perhaps be useful.

Add info on amount of explained variance. Perhaps to the status bar or to the info box below parameter settings.

Look for the same info with PCA widget?

janezd:
##### Issue

Fixes #6070.

##### Description of changes

To reviewer: please check that I indeed called `update_stress` from all places where it's needed -- and that it's not called multiple times.

I don't know where to show the stress. Optimization box was about the best place. I don't want to put it on the graph, because then we'd need to at least allow the user to move it and ... no.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -200,6 +200,7 @@ class OWMDS(OWDataProjectionWidget, ConcurrentWidgetMixin):
 
         self.embedding = None  # type: Optional[np.ndarray]
         self.effective_matrix = None  # type: Optional[DistMatrix]
+        self.stress = None
 
         self.size_model = self.gui.points_models[2]
         self.size_model.order = \
@@ -241,6 +242,8 @@ class OWMDS(OWDataProjectionWidget, ConcurrentWidgetMixin):
                 sizePolicy=(QSizePolicy.MinimumExpanding, QSizePolicy.Fixed),
                 callback=self.__refresh_rate_combo_changed),
             1, 1)
+        self.stress_label = QLabel("Kruskal Stress: -")
+        grid.addWidget(self.stress_label, 2, 0, 1, 3)
 
     def __refresh_rate_combo_changed(self):
         if self.task is not None:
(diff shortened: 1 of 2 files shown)


Orange/widgets/unsupervised/owmds.py
```diff
@@ -200,6 +200,7 @@ def __init__(self):
 
         self.embedding = None  # type: Optional[np.ndarray]
         self.effective_matrix = None  # type: Optional[DistMatrix]
+        self.stress = None
```
lanzagar, line 203: As shown by failing tests, this conflicts with the static method with the same name.
This results in a crash when setting the size of points to "Stress".

lanzagar (changes requested):

Orange/widgets/unsupervised/owmds.py
janezd: It's not that I don't warn students about this exact mistake.

Orange/widgets/unsupervised/owmds.py
```diff
@@ -392,17 +395,32 @@ def on_partial_result(self, result: Result):
...
+        stress_val = "-" if self.stress is None else f"{self.stress:.3f}"
+        self.stress_label.setText(f"Kruskal Stress: {stress_val}")
+
+    def _compute_stress(self):
+        if self.embedding is None or self.effective_matrix is None:
+            return None
+        actual = scipy.spatial.distance.pdist(self.embedding)
+        actual = scipy.spatial.distance.squareform(actual)
+        return np.sqrt(np.sum((actual - self.effective_matrix) ** 2)
+                       / (np.sum(self.effective_matrix ** 2) or 1))
```
lanzagar, line 422: It would probably be better to call `get_stress` and normalize the returned result.

janezd: Done.

PR #6310, opened 2023-01-24
OWBoxplot: fix failing tests

noahnovsak:
The sampling added for initial dask compatibility sometimes caused differences in the order of grouping and variable relevance.


markotoplak:
One of design goals for our widgets is that the user will get exactly the same output (or widget behavior) when the save / load the workflow. Therefore the sampling itself should be made in a reproducible way,

markotoplak:
@noahnovsak, could you rebase your commits into one so that we can merge this? Thanks?

PR #6314, opened 2023-01-25
[FIX] stats.utils: Don't count zeros as nans

### LINKED ISSUE #6074: Orange.statistics.utils.stats() counts defined values instead of NaNs
Orange.statistics.utils.stats() returns number of unknown (NaN) values for dense tables and the number of nonzero values for sparse tables. This wrongly equates undefined values and unknown values.

I'd change that. I think we had similar issues recently. Does anyone remember them and what did we do there?

janezd:
##### Issue

Fixes #6074.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/statistics/util.py
+++ b/Orange/statistics/util.py
@@ -344,31 +344,22 @@ def stats(X, weights=None, compute_variance=False):
 
     if X.size and is_numeric:
         if is_sparse:
+            nans = countnans(X, axis=0)
             X = X.tocsc()
+        else:
+            nans = np.isnan(X).sum(axis=0)
         if compute_variance:
             means, vars = nan_mean_var(X, axis=0, weights=weights)
         else:
             means = nanmean(X, axis=0, weights=weights)
             vars = np.zeros(X.shape[1] if X.ndim == 2 else 1)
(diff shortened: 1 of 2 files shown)


Orange/statistics/util.py
```diff
@@ -344,31 +344,22 @@ def stats(X, weights=None, compute_variance=False):
 
     if X.size and is_numeric:
         if is_sparse:
+            nans = countnans(X, axis=0)
```
markotoplak, line 347: Did not measure, but wouldn't 't it be better to count nans it columns after `tocsc`? Or, if this does not changes anything, we could remove `tocsc` altogether.

janezd: I moved it there because `countnans` wants csr:

```
        arr = arr.tocsr()
        counts = _count_nans_per_row_sparse(arr, weights, dtype=dtype)
```

And I kept conversion to csc because `nanmean` calls `mean_variance_axis` which is smart wrt combining axis and sparse matrix type.

markotoplak: Smart! Thanks.

PR #6317, opened 2023-01-26
[FIX] Fix unpickling domains: do not pickle indices (which can cause problems)

markotoplak:
##### Issue

Fixes  #6202, where the domain could not be unpickled if data included something based on SharedComputeValue (that redefined `__hash__`) and was further preprocessed before learning (thus only scikit-based learners failed but TreeLearner did not).

Unpickling dictionaries that include objects that redefine `__hash__` as keys is sometimes problematic (when said objects do not have `__dict__` filled-in yet in but are used as keys in a restored dictionary).

See the following for a problem description. https://petrounias.org/articles/2014/09/16/pickling-python-collections-with-non-built-in-type-keys-and-cycles/

##### Description of changes

Orange's `Domain` includes `_indices` for name-based or var-based indexing. That is a dictionary with `Variables` as keys, and these can cause problems during unpickling. Because `_indices` can be recreated, I decided not to pickle it and therefore managed to avoid this specific problem.

This will makes new pickled domain incompatible with older Orange versions.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/domain.py
+++ b/Orange/data/domain.py
@@ -164,16 +164,38 @@ class Domain:
         if not all(var.is_primitive() for var in self._variables):
             raise TypeError("variables must be primitive")
 
-        self._indices = dict(chain.from_iterable(
-            ((var, idx), (var.name, idx), (idx, idx))
-            for idx, var in enumerate(self._variables)))
-        self._indices.update(chain.from_iterable(
-            ((var, -1-idx), (var.name, -1-idx), (-1-idx, -1-idx))
-            for idx, var in enumerate(self.metas)))
+        self._indices = None
 
         self.anonymous = False
 
         self._hash = None  # cache for __hash__()
+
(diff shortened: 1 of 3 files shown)


markotoplak:
To see the problems, just run the new `test_all_models_work_after_unpickling_pca` without the actual fix.

markotoplak:
@janezd, falling -latest tests are independent of this PR.

janezd:
> @janezd, falling -latest tests are independent of this PR.

Ah, you've seen nothing yet. Wait for lazy signals!

janezd:
> @janezd, falling -latest tests are independent of this PR.

Should be fixed in #6320.

janezd:
> Fixes https://github.com/biolab/orange3/pull/6202

You meant #6206, right?

PR #6320, opened 2023-01-27
WidgetLearnerTestMixin: Fix for clicks of disabled buttons

janezd:
##### Issue

Tests fail after https://github.com/biolab/orange-widget-base/pull/228 because they click disable buttons.

##### Description of changes

Call `clicked.emit()` directly.


markotoplak:
Thanks!

PR #6331, opened 2023-02-07
[FIX] Concatenate - preserve table names when compute value ignored

PrimozGodec:
##### Issue

Coconcate widget does not preserve table names when compute value is ignored, so `unknown` appears in the `Source ID` column instead of the correct name.

##### Description of changes
Preserve table name also when ignore compute value is ticked. 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
Fixes #6332.

PR #6340, opened 2023-02-20
Avoid openTSNE 0.7.0 due to segfaults

markotoplak:
##### Issue
Tests fail after the new openTSNE release.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -57,7 +57,7 @@ requirements:
     - python-louvain >=0.13
     - requests
     - matplotlib-base >=3.2.0
-    - openTSNE >=0.6.1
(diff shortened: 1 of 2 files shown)


markotoplak:
Thanks @noahnovsak for finding the issue. @pavlin-policar, perhaps check what went wrong in openTSE?

pavlin-policar:
Oh, I've tracked down the bug and fixed it before releasing 0.7.1. It wasn't a segfault, but a memory error, which occurred due to a minor change in one the approximation schemes. The entire problem arises because we try to embed points with identical feature values, e.g., in titanic. This is a pretty unusual use-case and wasn't covered in my openTSNE tests. It should be covered now.

Issue #5720, opened 2021-12-05
Lazy outputs (... honour sorting in Data Table)


kaimikael:
**What's your use case?**

In the attached workflow I input a file with the populations of all Swedish municipalities.  They are a bit too many to comfortably display in a Bar Plot, but I thought I could use a Data Sampler to cut them down to size, as it were.  I turns out that the Data Sampler does not retain the original order of the data.  That is not a problem, I thought, I insert a Data Table where I can sort the data rows by population, but it seems the sorting is not retained when sending on to the Bar Plot, see attached screenshots.  This is inconvenient.  

**What's your proposed solution?**

I would like the Bar Plot to honour the sorting of rows done in the Data Table.

[Municipalities, Sweden.csv](https://github.com/biolab/orange3/files/7656761/Municipalities.Sweden.csv)
[PopulationPlot.zip](https://github.com/biolab/orange3/files/7656762/PopulationPlot.zip)
<img width="798" alt="Screenshot 2021-12-05 at 21 10 48" src="https://user-images.githubusercontent.com/10616212/144762803-410b4d7e-105b-4677-a76f-db0d53b4e7ef.png">
<img width="1132" alt="Screenshot 2021-12-05 at 21 11 02" src="https://user-images.githubusercontent.com/10616212/144762818-d3b95d3a-66be-4273-977f-3c2a7eba141f.png">

**Are there any alternative solutions?**

An alternative solution would be to allow me to drag the bars in the bar plot to put them in any order, but it would probably be very tedious for any larger data set.

ajdapretnar:
The issue is explained here: https://github.com/biolab/orange3/issues/5655
It seems this cannot be solved easily.

kaimikael:
Ah, I see, one has to explicitly _select_ the rows to get them in the right order to the next widget.  That’s not too onerous, but maybe it should be explicitly explained in the documentation for Data Table, it feel as if it should be a fairly common use case.

janezd:
Thanks for this suggestion, because it revived the discussion about #5655 and resulted in an interesting idea: outputs could be lazy. A widget could offer a function to compute the output, but the function would be called (and the output computed) only when actually needed by some input.

We'll try the concept. I'm renaming the issue to make it easier to find.

janezd:
Lazy signals do not seem difficult to implement. A quick and dirty solution (which probably breaks something) is to have

```python
class LazyValue:
    NoValue = object()

    def __init__(self, callable):
        self.callable = callable
[...]

PR #6348, opened 2023-02-24
[ENH] Lazy signals for Hierarchical Clustering

janezd:
##### Issue

Ref. #5720.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/utils/state_summary.py
+++ b/Orange/widgets/utils/state_summary.py
@@ -1,10 +1,11 @@
 from datetime import date
 from html import escape
+from typing import Union
 
 from AnyQt.QtCore import Qt
 
 from Orange.widgets.utils.localization import pl
-from orangewidget.utils.signals import summarize, PartialSummary
+from orangewidget.utils.signals import summarize, PartialSummary, LazyValue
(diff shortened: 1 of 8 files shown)


markotoplak:
@janezd, could you rebase this? Also, biolab/orange-widget-base/pull/238 was released in widget-base 4.21.0, so if we change requirements, this PR can be merged.

Orange/widgets/utils/state_summary.py
```diff
@@ -129,11 +131,11 @@ def b(s):
 
 def missing_values(value):
     if value:
-        return f'({value*100:.1f}% missing values)'
+        return f' ({value*100:.1f}% missing values)'
```
markotoplak, line 134: I assume no one calls this externally, and moving a space inside won't be a problem. But I am curious why did you do so.

janezd: Because of space-nazism. We had `f'Features: {features} {features_missing}'` where `features_missing` could be an empty string, resulting in a space at the end of the line. I wouldn't be able to sleep knowing that.

PR #6353, opened 2023-02-27
sqltable compute nan frequency

noahnovsak:
##### Issue
`__get_nan_frequency' uses X and Y arrays, so SqlTables need an alternative implementation.

### DIFF
--- a/Orange/data/sql/table.py
+++ b/Orange/data/sql/table.py
@@ -656,11 +656,13 @@ class SqlTable(Table):
         return np.nan
 
     def __get_nan_frequency(self, columns):
-        query = self._sql_query([" + ".join([f"COUNT(*) - COUNT({col.to_sql()})"
-                                             for col in columns])])
-        with self.backend.execute_sql_query(query) as cur:
-            nans = cur.fetchone()
-        return nans[0] / (len(self) * len(columns))
+        try:
+            query = self._sql_query([" + ".join([f"COUNT(*) - COUNT({col.to_sql()})"
+                                                 for col in columns])])
+            with self.backend.execute_sql_query(query) as cur:
+                return cur.fetchone()[0] / (len(self) * len(columns))
+        except BackendError:
+            return None
 
     def get_nan_frequency_attribute(self):
         return self.__get_nan_frequency(self.domain.attributes)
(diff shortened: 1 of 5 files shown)


markotoplak:
@lanzagar or @PrimozGodec: can one of you please check this PR? Thanks!

Orange/tests/sql/test_sql_table.py
```diff
@@ -764,6 +764,20 @@ def test_list_tables_with_schema(self):
             with self.backend.execute_sql_query("DROP SCHEMA IF EXISTS orange_tests CASCADE"):
                 pass
 
+    @dbt.run_on(["postgres"])
```
PrimozGodec, line 767: It should be also tested on mssql

```suggestion
    @dbt.run_on(["postgres", "mssql"])
```

Orange/data/sql/table.py
```diff
@@ -654,3 +654,19 @@ def _execute_sql_query(self, query, param=None):
 
     def checksum(self, include_metas=True):
         return np.nan
+
+    def __get_nan_frequency(self, columns):
+        nans = 0
+        for column in columns:
+            col_field = column.to_sql()
+            query = self._sql_query(["COUNT(*)"],
+                                    filters=[f"{col_field} IS NULL OR {col_field} = 'NaN'"])
```
PrimozGodec, line 663: Checking NULL is sufficient. Checking `"nan"` even fail on MSSQL when column is float.
```suggestion
        query = self._sql_query(["COUNT(*)"], filters=[f"{col_field} IS NULL"])
```

PrimozGodec: Anyway, can we make it in one query to speed the process up? For example with this: https://stackoverflow.com/a/16528738/3551700

PrimozGodec (changes requested):

PrimozGodec:
Additionally (not connected with your changes but required to insert NULL values in database correctly) `Orange.test.base` lines 280-285 should be changed to:
```
        insert_values = ", ".join(
            "({})".format(
                ", ".join(
                    "NULL" if pd.isna(v) else "'{}'".format(v)
                    for v, t in zip(row, sql_column_types)
                )
            )
            for row in data
        )
```
I replaced v is None with pd.isna(v) to correctly handle None and nan values.

PrimozGodec:
@noahnovsak thank you for patience.

PR #6357, opened 2023-03-02
Table: Fix ensure_copy to also check ids

janezd:
##### Issue

`Table.copy()` does not copy ids, but uses a reference to existing ids, as detected by tests in Image Analytics (https://github.com/biolab/orange3-imageanalytics/actions/runs/4314962156/jobs/7529016431).

##### Includes
- [X] Code changes
- [X] Tests


markotoplak:
`ids` are not supposed to be changed manually. But yes, let us make them so that they can be. It does not cost anything.

PR #6358, opened 2023-03-06
[ENH] Datasets: Add language selection

janezd:
##### Issue

Data sets in the repository now have a language flag, and Data Sets widget allows user to filter files by language.

Default is English, but translations of Orange to other languages can change this default.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation


janezd:
> Please also change the icon that shows that the data set is local but not loaded.

This will be another PR because it requires changing (and then releasing) orange-widget-base.

PR #6362, opened 2023-03-11
Datasets: Nicer indicator

janezd:
##### Issue

Fixes https://github.com/biolab/orange3/pull/6358#issuecomment-1463500026.

~~Requires https://github.com/biolab/orange-widget-base/pull/233.~~

##### Description of changes

Gray indicators (data set is downloaded but not active) are now hollow. Green indicators have gray pen.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -308,7 +308,7 @@ class OWDataSets(OWWidget):
         self.view.setItemDelegate(UniformHeightDelegate(self))
         self.view.setItemDelegateForColumn(
             self.Header.islocal,
-            UniformHeightIndicatorDelegate(self, role=Qt.DisplayRole, indicatorSize=4)
+            UniformHeightIndicatorDelegate(self, indicatorSize=4)
         )
         self.view.setItemDelegateForColumn(
             self.Header.size,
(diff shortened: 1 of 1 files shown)


Orange/widgets/data/owdatasets.py
```diff
@@ -456,20 +457,24 @@ def set_model(self, model, current_index):
                 QItemSelectionModel.ClearAndSelect | QItemSelectionModel.Rows)
             self.commit()
 
+    def indicator_state_for_info(self, info):
+        localinfo = list_local(self.local_cache_path)
```
lanzagar, line 461: This is called in a for loop so it reads all info files from disk and parses the json info for each dataset.
Quadratic complexity for cpu operations and in-ram data would not be a problem, but we should probably avoid excessive disk access.

Orange/widgets/data/owdatasets.py
```diff
@@ -456,20 +457,24 @@ def set_model(self, model, current_index):
...
             info = item.data(Qt.UserRole)
-            is_local = info.file_path in localinfo
-            is_current = (is_local and
-                          os.path.join(self.local_cache_path, *info.file_path)
-                          == self.current_output)
-            item.setData(" " * (is_local + is_current), Qt.DisplayRole)
-            item.setData(self.IndicatorBrushes[is_current], Qt.ForegroundRole)
+            state = self.indicator_state_for_info(info)
+            item.setData(state, UniformHeightIndicatorDelegate.IndicatorRole)
+            item.setData(self.IndicatorBrushes[bool(state)], Qt.ForegroundRole)
```
lanzagar, line 477: The previous spaces were used for sorting, which now does not work anymore.

lanzagar (changes requested):

Orange/widgets/data/owdatasets.py
janezd: This function is used in two places, both time in a loop. I change it to require `localinfo` as an argument, which can be computed before the loop.

Someday, we should refactored. Instead of this loop, we should use a widget-specific model instead of `QStandardItemModel`.

Orange/widgets/data/owdatasets.py
janezd: Fixed, and also added a comment.

PR #6369, opened 2023-03-24
Keywords: changed from list to string

anjamejac:
##### Issue
Keywords were in list, which caused us translation complications.

##### Description of changes
Now they are written in string.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd (approved):
Adding the name of the widget as keyword (unless this is the only keyword and we need a placeholder) is unnecessary. English Orange also searches for name when typing in canvas, and translations can add the English name. But it doesn't hurt either, so you can leave them.

I'd prefer not to merge this before we release Slovenian Orange. If we merge, existing translations of keywords will stop working (unless you update the translations - like tomorrow morning).

PR #6381, opened 2023-03-31
Mosaic tests: Speed up the tests, increase timeout where necessary

janezd:
##### Issue

Tests timeout on Macs.

##### Description of changes

Decrease the number of combinations scored by VizRank. In the tests that still takes 0.5 s on my machine, I increased the timeout from 5 s to 10 s, just to be sure.

##### Includes
- [X] Tests


janezd:
@PrimozGodec, the only tests that still fail are related to pythoneditor/Python Script. We're counting on you. (Consider this a gentle encouragement to skip them asap. :)

PR #6382, opened 2023-03-31
Run workfows in tests (except for the coverage test)

### LINKED ISSUE #6379: Our tests fail to test workflows
...except the Qt6 tests.

#6181 had a compatibility issue which we almost did not catch because only Qt6 tests were having problems.

The cause for this is that `tox.ini` for `testenv` includes:
```
    # Skip loading of example workflows as that inflates coverage
    SKIP_EXAMPLE_WORKFLOWS=True
```

These worfklows should be skipped only for the run that is actually used for coverage. So we should do the opposite, run example workflows everywhere except on a one specific (coverage) run.

markotoplak:
##### Issue
Fixes #6379.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,7 +54,6 @@ jobs:
             tox_env: pyqt6
             name: PyQt6
 
-
     services:
       postgres:
         image: orangedm/postgres:11
@@ -91,6 +90,12 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install --upgrade tox
 
+      - name: Skip testing workflows at coverage
+        if: |
+          matrix.python-version == '3.8' && matrix.tox_env == 'orange-released'
+        run: |
+          echo 'SKIP_EXAMPLE_WORKFLOWS=1' >> $GITHUB_ENV
+
       - name: Run Tox
         run: catchsegv xvfb-run -a -s "$XVFBARGS" tox -e ${{ matrix.tox_env }}
         env:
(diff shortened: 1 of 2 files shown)


markotoplak:
When I tested with the code below, everything but the python 3.8 released test on linux failed (the one used for coverage). So I think this is OK.

```
diff --git a/Orange/widgets/tests/test_workflows.py b/Orange/widgets/tests/test_workflows.py
index 12a07bc9e..1bc593d16 100644
--- a/Orange/widgets/tests/test_workflows.py
+++ b/Orange/widgets/tests/test_workflows.py
@@ -38,6 +38,7 @@ class TestWorkflows(GuiTest):
         and also those placed "workflows" subfolder.
         GH-2240
         """
+        self.assertEqual("TEST WORKFLOWS WAS RUNNING", "YES")
         reg = registry()
         test_workflows = chain(
             discover_workflows(dirname(workflows.__file__)),
diff --git a/tox.ini b/tox.ini
index 6b23e235a..43ec089ed 100644
--- a/tox.ini
+++ b/tox.ini
@@ -71,12 +71,13 @@ deps =
     oldest: catboost==1.0.1
 
 commands_pre =
+    python -c "import os; print(os.environ)"
     # Verify installed packages have compatible dependencies
     pip check
     # freeze environment
     pip freeze
 commands =
-    coverage run -m unittest -v Orange.tests Orange.widgets.tests
+    coverage run -m unittest -v Orange.widgets.tests.test_workflows
     coverage combine
     coverage report
 
@@ -93,7 +94,7 @@ deps =
     PyQt6-WebEngine-Qt6==6.3.*
 
 commands =
-    python -m unittest -v Orange.widgets.tests
+    python -m unittest -v Orange.widgets.tests.test_workflows
 
 [testenv:add-ons]
 deps =
-- 
2.25.1
```

Issue #6041, opened 2022-06-20
Merge data with multiple matches


Chocoder:
**What's your use case?**
Hello, I realised that the merge widget gives a warning when multiple matches are found...

[https://docs.microsoft.com/fr-fr/power-query/merge-queries-left-outer](https://docs.microsoft.com/fr-fr/power-query/merge-queries-left-outer)

[https://docs.microsoft.com/fr-fr/power-query/merge-queries-inner](https://docs.microsoft.com/fr-fr/power-query/merge-queries-inner)

In my use case I have a first table with a CountryID that could exists several time (ex: 1 1 3 4). In my second table those ID exist only one time (1 2 3). So I would like to merge - left or inner - the first table with the second one. Left comportment will matches 3 times (and let one empty cell in the Country column), inner will matches 3 times too (and let no empty cells in the Country column.

Is there a trick to use the existing Merge widget to apply this kind of table merging operation?

**What's your proposed solution?**
Use the same merge widget, with a warning if empty cells may appear, but authorizes the merge option.

**Are there any alternative solutions?**
Merge outside of Orange?
Design a widget with Python raw code?

Chocoder:
@janezd hello! Any clue on that topic? You helped me on another one so I tried again! Thanks,

janezd:
Are you sure you have not just reversed your input signals? As I understand your description, this is exactly what Merge is made for - in its current form.

janezd:
Works for me: [Archive.zip](https://github.com/biolab/orange3/files/11132456/Archive.zip)

PR #6389, opened 2023-04-02
Merge Data: Allow non-unique values in left-join, if they're not used

janezd:
##### Issue

Left join is possible even with non-unique values on the right if these values are not used, i.e. do not appear on the right.

For instance, when adding geo coordinates to Slovenian towns, I used data for 6002 Slovenian places with some 800 duplicates, but these were small places that were not in the left table, so merge would be possible - but the widget rejected it.

This may also be related to #6041.

**To discuss:** we could also soften restrictions for the other two types of join to consider only used data, but I believe this wouldn't look OK for outer join because the duplicated rows would actually appear in the output data, but these same rows would become prohibited at the moment they're actually used. For inner join ... I'm not sure.

##### Description of changes

In left join, issue an error only if duplicates prevent merging, and just a warning in case of harmless duplicates. The warning imho makes sense to alert the user that the right data table may not always be useful for the merge.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation


Orange/widgets/data/owmergedata.py
```diff
@@ -294,21 +294,30 @@ class Outputs:
...
+            "multiple rows.")
 
     class Error(widget.OWWidget.Error):
         matching_numeric_with_nonnum = Msg(
             "Numeric and non-numeric columns ({} and {}) cannot be matched.")
         matching_index_with_sth = Msg("Row index cannot be matched with {}.")
         matching_id_with_sth = Msg("Instance cannot be matched with {}.")
+        nonunique_left_matched = Msg(
+            "Some combinations of values on the left appear in multiple rows."
+            "\nEvery matched combination may appear at most once.")
```
VesnaT, line 311: I'd rather be more consistent:
`left` -> `Data`
`right` -> `Extra Data`

VesnaT (approved):

PR #6390, opened 2023-04-02
[ENH] Select Columns: Accept partially correct drops

### LINKED ISSUE #5897: Select Columns: ignore moving string to features automatically
**What's your use case?**

Sometimes there are many variables in the data, some string, other numeric, discrete. I'd like to be able to Ctrl+A ignored features to features, where Orange automatically takes only valid variables for features, leaving string variables behind in the ignored list. Currently, the widget just blocks moving the variables altogether.

Can Orange cleverly filtering the variables and move just relevant variables to the right and ignore string vars, even when they are selected?

/

janezd:
##### Issue

Fixes #5897.

Drag & Drop between VariablesListViews rejects drops that contain any variables of unacceptable type.

##### Description of changes

- Reimplement drag procedure to accept partially correct drops.
- Same for buttons: the are enabled if there is anything to move, and move only what they may

##### Includes
- [X] Code changes
- [x] Tests
- N/A: Documentation

### DIFF
--- a/Orange/widgets/data/owselectcolumns.py
+++ b/Orange/widgets/data/owselectcolumns.py
@@ -1,12 +1,12 @@
 from functools import partial
 from typing import Optional, Dict, Tuple
 
-from AnyQt.QtWidgets import QWidget, QGridLayout
-from AnyQt.QtWidgets import QListView
 from AnyQt.QtCore import (
     Qt, QTimer, QSortFilterProxyModel, QItemSelection, QItemSelectionModel,
     QMimeData, QAbstractItemModel
 )
+from AnyQt.QtGui import QDrag, QDropEvent
+from AnyQt.QtWidgets import QWidget, QGridLayout, QListView
 
 from Orange.data import Domain, Variable
 from Orange.widgets import gui, widget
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owselectcolumns.py
```diff
@@ -97,7 +102,76 @@ def dropMimeData(self, mime, action, row, column, parent):
         if row < 0:
             row = self.rowCount()
 
+        if self.primitive and not all(var.is_primitive() for var in variables):
+            variables = [var for var in variables if var.is_primitive()]
+            self[row:row] = variables
+            mime.setProperty("_moved", variables)
```
janezd, line 108: @ales-erjavec, MIME is passed from source of the drag to its destination. In this PR, destination may accept only some data, so it must have a way to pass the list of accepted variables back to the source. `drag.exec` does not return values. Here I treated mime data as a shared, two-way object. It works, but is it OK? Documentation neither mentions nor prohibits it.

ales-erjavec: Probably fine.

PR #6396, opened 2023-04-05
Minor f-string changes

anjamejac:
##### Issue
Translation (si) problems with `.format`.

##### Description of changes
Minor f-string changes.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owhierarchicalclustering.py
+++ b/Orange/widgets/unsupervised/owhierarchicalclustering.py
@@ -17,6 +17,7 @@ from AnyQt.QtCore import (
 )
 from AnyQt.QtCore import pyqtSignal as Signal, pyqtSlot as Slot
 
+from Orange.widgets.utils.localization import pl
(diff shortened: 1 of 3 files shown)


Orange/widgets/unsupervised/owhierarchicalclustering.py
```diff
@@ -1044,7 +1045,7 @@ def send_report(self):
         elif self.selection_method == 1:
             sel = "at {:.1f} of height".format(self.cut_ratio)
         else:
-            sel = f"top {self.top_n} clusters"
+            sel = f"top {self.top_n} {pl(self.top_n, 'cluster')}"
```
janezd, line 1048: This change make sense for English but, by the way, in case you didn't know: even if it remained as it was, could still use `plsi` in Slovenian translation.

Even if the string is initially not an f-string (this one is), trubar changes it to f-string, if necessary.

PR #6401, opened 2023-04-05
Outliers: changes for report

anjamejac:
##### Issue
Missing possibility to translate metrics and then have them translated in report.

##### Description of changes
Added list of metrics names and reorganizing code for better report function.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owoutliers.py
+++ b/Orange/widgets/data/owoutliers.py
@@ -6,6 +6,7 @@ import numpy as np
 from AnyQt.QtCore import Signal, Qt
 from AnyQt.QtWidgets import QWidget, QVBoxLayout
 
+from orangewidget.report import bool_str
 from orangewidget.settings import SettingProvider
 
 from Orange.base import Learner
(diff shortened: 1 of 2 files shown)


Orange/widgets/data/owoutliers.py
```diff
@@ -317,32 +338,12 @@ def send_report(self):
...
-                 ("Metric", params["metric"])))
-        elif self.outlier_method == self.IsolationForest:
-            self.report_items(
-                "Detection",
-                (("Detection method", "Isolation Forest"),
-                 ("Contamination", params["contamination"])))
+        if (self.outlier_method == self.OneClassSVM) or \
+                (self.outlier_method == self.Covariance) or \
+                (self.outlier_method == self.LOF) or \
+                (self.outlier_method == self.IsolationForest):
```
janezd, line 344: You can remove this `if` and `else` (together with `raise`).

I think that @VesnaT wanted to raise an exception is somebody adds a new method and forgets to add it to report. But now that you implemented `get_report_parameters` this test is no longer necessary. If somebody adds a new editor and forgets to implement `get_report_parameters`, the base class' will raise this exception. So just calling `self.current_editor.get_report_parameters` suffices.

(By the way, a more Pythonic way for this expression is `if self.outlier_method in (self.OneClassSVM, self.Covariance, self.LOF, self.IsolationForest)`.)

janezd (changes requested):
While your at it: current `send_report` does not report anything if there is no data. It could still report parameters.

I suggest changing report so that if there is no data, you just skip the first `report_items` instead of returning without report.

Issue #6315, opened 2023-01-25
Data Table: single tab, bug fixes and subset input


janezd:
Changing Data Table to single input will require refactoring settings, so this change can do several things at once (not doing so would require changing code that would be later thrown away):

- [ ] Change the input signal to a single table
- [ ] Preserve sorting (#6311)
- [ ] Preserve selection (#6312)
- [ ] Add Data Subset input and indicate the subset in the table (#6128)

janezd:
> Now what will happen to the selection and the sorting when the input data changes? I hope that if the selection was
> set to "all", still all the data remains selected and the sorting order remains.

@wvdvegte, if nothing is selected, it has the same effect as if everything is selected.

... except that in this case sorting is not reflected in the output. Fixing this efficiently will require solving #5720, which looks simple but hides some nasty details.

janezd:
Note: @ales-erjavec proposed to keep the existing widget but have it in a hidden category, so old workflows will keep working. For new workflows, only the new widget would be available.

PR #6405, opened 2023-04-07
[ENH] Data Table: Subset input

ales-erjavec:
##### Issue

Closes: https://github.com/biolab/orange3/issues/6315

Add Data Subset input and indicate the subset in the table

Screenshot 2023-04-07 at 15 14 53

Uses opacity of non subset data to emphasize subset to match other visualization widgets, and also adds an indicator to the row header (I don't like either).

##### Description of changes

Add subset input 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


VesnaT:
It looks good to me. 

One thing, that could also be handled is to retain the selection when connecting/disconnecting the subset.
Beside that, resending the output on connecting/disconnecting the subset seems unnecessary.

PR #6409, opened 2023-04-13
[FIX] tableview: Fix errors with sparse basket columns

ales-erjavec:
##### Issue

Since #6370 errors appear in data table when a sparse basket column in on input.

When clicking 'Show variable labels':
```
  File "/home/ales/devel/orange3/Orange/widgets/data/utils/tableview.py", line 175, in __headerDataChanged
    text += "\n".join(key for key, _ in items)
  File "/home/ales/devel/orange3/Orange/widgets/data/utils/tableview.py", line 175, in <genexpr>
    text += "\n".join(key for key, _ in items)
ValueError: not enough values to unpack (expected 2, got 1)
```

When trying to sort a basket column:
````
  File "/home/ales/devel/orange3/Orange/widgets/data/owtable.py", line 341, in _on_sort_indicator_changed
    self.stored_sort.append((var.name, order))
AttributeError: 'NoneType' object has no attribute 'name'

````

##### Description of changes

Fix errors in handling sparse basket column in tableview.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -1,6 +1,6 @@
 import concurrent.futures
 from dataclasses import dataclass
-from typing import Optional, Union, Sequence, List, TypedDict, Tuple, Dict
+from typing import Optional, Union, Sequence, List, TypedDict, Tuple
 
 from scipy.sparse import issparse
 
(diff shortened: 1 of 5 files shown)


VesnaT:
I can temporarily merge this to avoid the crash, but I don't agree with the solution. 
The table could be sortable by Y or metas even though it has sparse X.

markotoplak:
To me, this seems like removing a feature (being able to sort such table on metas) for the occasional, user-avoidable :) bug. In such cases, I prefer buggy features to no features.

PR #6421, opened 2023-04-19
Add message `Table length` and `Limit`

Mikhaylov-yv:
##### Description of changes

Made a message about the limit. Database table length and actual limit

Now the message looks like this:
```
Data appears to be big. Do you really want to download it to local memory?
Table length: 1,000,100. Limit 1,000,000
```

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


lanzagar:
Thank you for the contribution!

PR #6426, opened 2023-04-24
[ENH] Edit Domain: Change types of multiple variables

### LINKED ISSUE #6088: Edit Domain: enable setting the type for several variables at once
**What's your use case?**

Sometimes, one would like to convert several categorical variables to numeric (or vice-versa). This is not possible to do at the moment. One has to click every variables separately and set it.

Enable selecting several variables in Edit Domain and setting the type. I've checked and if the conversion cannot be done (i.e. transform text to numeric), the widget will output missing values for the erroneous variable. Which is quite smooth.

/

janezd:
##### Issue

Fixes #6088.

Based on #6415 to avoid conflicts.

##### Description of changes

Allows changing the variable type for multiple variables at once.

As discussed with @markotoplak, editing other properties, even labels, would be confusing because controls would have to be disabled if selected variables differ in those properties. Also, the widget already has over 3000 lines of complex code.

The only feature we could easily add would be a tri-state checkbox for unlinking.

The widget has four separate editors for different variable types, which are stacked into a single editor. This PR implements a  fifth editor, which is blank (except for the combo with type) and can handle multiple variables. Existing editors are kept as they are, that is, they accept a single variable. The "top editor" has separate methods for single and multiple variables. *One might think* that the four editors could accept lists instead of single variables to avoid having duplicated methods in the top editor. I thought so for long time, too, but the code was just getting more and more complex. This solution is simpler.

##### Includes
- [X] Code changes
- [x] Tests
- N/A Documentation

### DIFF
--- a/Orange/widgets/data/oweditdomain.py
+++ b/Orange/widgets/data/oweditdomain.py
@@ -12,7 +12,7 @@ from collections import namedtuple, Counter
 from functools import singledispatch, partial
 from typing import (
     Tuple, List, Any, Optional, Union, Dict, Sequence, Iterable, NamedTuple,
-    FrozenSet, Type, Callable, TypeVar, Mapping, Hashable, cast
+    FrozenSet, Type, Callable, TypeVar, Mapping, Hashable, cast, Set
 )
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


markotoplak:
@janezd, just to double check: if I understood this code correctly, it does not in any way change how settings are stored (because they are always saved per variable), and we thus do not need to increase the settings version.

Issue #6422, opened 2023-04-23
Select Rows by "does not contain"


cevznriny:
**What's your use case?**

Selecting rows that do not contain a string.

**What's your proposed solution?**

Add "does not contain" to the list of conditions.

**Are there any alternative solutions?**

No.

wvdvegte:
Do you expect the string that you want to check for to be always in the same column? In that case there is a simple alternative solution: use Feature Generator to create a categorical variable called Unselect:

Unselect := 'Y' if 'xxx' in str(Name) else 'N' 

with 'xxx' your string and Name the name of the column where to look for it.

Then put a Select Rows after Feature Constructor with condition

Unselect | is not | Y

cevznriny:
Thanks. That works. It might be a bit tricky to work out for a non-coder.

I added a "Does Not Contain" to Select Rows in my fork. It works but I don't know if I've found everything.
Could you see if it is worth submitting for a pull request.
https://github.com/cevznriny/orange3/commit/023f668de70962fd26ee65fc09daf873d1b77619

wvdvegte:
I'm not a coder either, but I do know some basic Python. Everything you can pull together in one line in an "=" assignment (shown as ":=" in Orange) can be used in Feature Constructor - especially "... if ... else ..." is very useful. In fact, these are pieces of Python code that you can ask ChatGPT to write for you ("produce one-line Python code that does the following:").
Since I'm not a coder, I have no idea how pull requests on GitHub work.

PR #6430, opened 2023-04-25
added "Does Not Contain" to "Select Rows"

cevznriny:
##### Issue

Fixes https://github.com/biolab/orange3/issues/6422

##### Description of changes
Added a "Does Not Contain" string filter to the Select Rows widget as the opposite of Contains.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
Thanks for this contribution.

We discussed this at the internal meeting. There may be nothing very special about "does not contain", compared to other negations. So we said that we either add all or none. Nobody was opposing the former (=adding), so, could you also add "doesn't begin with", "doesn't end with" and "is not defined"?

Also, could you add "is not defined" for numeric and categorical variables, so the set is complete?

cevznriny:
This is going to take a bit longer than expected. I think I need to work out how to use progresql to test the functions in filter.py. The unit tests don't really seem to test the functionality either, just whether they can be selected in the dialog.

janezd:
I think we can forget about postgresql: it's probably already broken elsewhere and we'll drop this functionality anyway.

Thanks for implementing this!

PR #6465, opened 2023-06-06
Make MAPE available to Test & Score

### LINKED ISSUE #6464: Please make "Mean Absolute Percentage Error" (MAPE) available in the "Test and Score" widget
Mean Absolute Percentage Error (MAPE) is a common metric that fits in well with the other options available in the Test and Score widget. This is available in [scikit-learn](https://scikit-learn.org/stable/modules/model_evaluation.html) and it appears it would be easy to add. 

There is some older Orange3 documentation that makes it look like it was available at one point but doesn't seem to be now.

mode80:
##### Issue
Closes #6464

##### Description of changes
Straightforward addition of the MAPE metric to Test & Score by delegating to scikit-learn. 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
Sure, why not. Thanks!

PR #6472, opened 2023-06-09
Dask: compute Table chunks if unknown

noahnovsak:
Calling len(Table) keeps causing trouble all over Orange, so maybe computing it would be reasonable to compute chunks. AFAIK it only needs to be done once.

### DIFF
--- a/Orange/tests/test_dasktable.py
+++ b/Orange/tests/test_dasktable.py
@@ -189,3 +189,9 @@ class TableTestCase(unittest.TestCase):
         self.assertEqual(150, len(lines))
         self.assertEqual("[[5.1, 3.5, 1.4, 0.2 | Iris-setosa],", lines[0])
         self.assertEqual(" [5.9, 3.0, 5.1, 1.8 | Iris-virginica]]", lines[-1])
+
+    def test_table_len(self):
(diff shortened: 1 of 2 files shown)


markotoplak:
Thanks! Can you also make a test for this?

Orange/tests/test_dasktable.py
```diff
@@ -189,3 +189,8 @@ def test_str(self):
         self.assertEqual(150, len(lines))
         self.assertEqual("[[5.1, 3.5, 1.4, 0.2 | Iris-setosa],", lines[0])
         self.assertEqual(" [5.9, 3.0, 5.1, 1.8 | Iris-virginica]]", lines[-1])
+
+    def test_table_len(self):
+        iris = temp_dasktable("iris")
+        iris = iris[iris.X[:, 0] > 5.5]  # make sure length is unknown
```
markotoplak, line 195: Can you also verify here that len is unknown? So instead of having a comment we should check something dask-internal.

markotoplak: As a general principle it is always better to write an assertion instead of putting it in a comment.

Another option would be to make sure that len() internally computes indices.

PR #6486, opened 2023-06-21
Add wrapper for testing Tables and DaskTables in tandem

noahnovsak:
Added a wrapper to help write tests.


noahnovsak:
Not sure if it's better to just write up new test cases like we've been doing so far or wrap existing ones, in the same way we handle sparse data. Just putting this out there so we can decide.

markotoplak:
Looks good, can come in handy.

PR #6488, opened 2023-06-21
[FIX] Fix classification trees for data with repeated feature values

markotoplak:
##### Issue

I started seeing this on github tests (only on Ubuntu).

```
FAIL: test_full_tree (Orange.tests.test_orangetree.TestClassifier)
  File "/home/runner/work/orange3/orange3/.tox/orange-released/lib/python3.9/site-packages/Orange/tests/test_orangetree.py", line 37, in test_full_tree
    self.assertTrue(np.all(table.Y.flatten() == pred))
```

##### Description of changes
The bug ran deeper. See my last comment: `find_threshold_entropy` skipped computing too many entropies.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/classification/_tree_scorers.pyx
+++ b/Orange/classification/_tree_scorers.pyx
@@ -79,7 +79,7 @@ def find_threshold_entropy(const double[:] x, const double[:] y,
             curr_y = <int>y[idx[i]]
             distr[curr_y] -= 1
             distr[n_classes + curr_y] += 1
-            if curr_y != y[idx[i + 1]] and x[idx[i]] != x[idx[i + 1]]:
+            if x[idx[i]] != x[idx[i + 1]]:
                 entro = (i + 1) * log(i + 1) + (N - i - 1) * log(N - i - 1)
                 for j in range(2 * n_classes):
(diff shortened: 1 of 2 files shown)


markotoplak:
Hmm, the trees are built differently.

When the test fails I see the following splitting process.
```
SA 150
split petal length [50. 50. 50.] 1.9
ending 50 [50.  0.  0.]
SA 100
split petal length [ 0. 50. 50.] 4.7
...
```

And when it does not:
```
SA 150
split petal length [50. 50. 50.] 1.9
ending 50 [50.  0.  0.]
SA 100
split petal width [ 0. 50. 50.] 1.7
...
```

So the second split differs (but both have the same quality).

markotoplak:
During debugging, I saw that `np. resort (the default, unstable variant) worked differently across platforms. This should not be a problem if we did not have a bug in `find_threshold_entropy`, which aimed to skip computing needless entropies.

It avoided too many: it skipped if the next class value was the same or the next value was the same. That was a problem when feature and class values could both (interchangeably) repeat.

PR #6506, opened 2023-07-10
Remove use of deprecated distutils

ales-erjavec:
##### Issue

Distutils stdlib is [deprecated](https://docs.python.org/3/whatsnew/3.10.html#distutils-deprecated)

##### Description of changes

Remove import and use of distutils except in setup.py where setuptools provides it.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/tests/test_doctest.py
+++ b/Orange/tests/test_doctest.py
@@ -1,10 +1,6 @@
 import sys
 import os
-import unittest
 from doctest import DocTestSuite, ELLIPSIS, NORMALIZE_WHITESPACE
-
-import pkg_resources
-import numpy
 
 SKIP_DIRS = (
     # Skip modules which import and initialize stuff that require QApplication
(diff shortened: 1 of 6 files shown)


Orange/tests/test_doctest.py
```diff
@@ -58,8 +58,8 @@ def suite(package):
     # numpy 1.14 changed array str/repr (NORMALIZE_WHITESPACE does not
     # handle this). When 1.15 is released update all docstrings and skip the
     # tests for < 1.14.
-    npversion = LooseVersion(numpy.__version__)
-    if npversion >= LooseVersion("1.14"):
+    npversion = pkg_resources.parse_version(numpy.__version__)
+    if npversion >= pkg_resources.parse_version("1.14"):
```
janezd, line 62: Is this still needed at all? We require numpy >= 1.20, I think.

PR #6513, opened 2023-07-18
Dask: Linear Regression

noahnovsak:
LinearRegressionLearner and the corresponding widget now support DaskTables.


Orange/base.py
```diff
@@ -443,7 +443,7 @@ def fix_dim(x):
         # Call the predictor
         backmappers = None
         n_values = []
-        if isinstance(data, (np.ndarray, scipy.sparse.csr_matrix)):
```
markotoplak, line 446: Do you remember who calls __call__ directly with matrices and not with data? Did you need to make this change or did you do it for safety?

If it was me I'd deprecate this base functionality.

Orange/regression/linear.py
```diff
@@ -27,19 +30,50 @@ def score(self, data):
...
+                return dask_ml.linear_model.LinearRegression(**params)
+            except ImportError:
+                warnings.warn("dask_ml is not installed, using sklearn instead.")
+        return self.__wraps__(**self.params)
+
     def fit(self, X, Y, W=None):
-        model = super().fit(X, Y, W)
-        return LinearModel(model.skl_model)
+        clf = self._initialize_wrapped(X, Y)
+        Y = Y.reshape(-1)
```
markotoplak, line 65: If someone else ensures that this class does not work with multiple targets then reshaping to (-1) is fine. Otherwise, we'd rather select the first column and then (if needed) reshape it.

Orange/statistics/util.py
```diff
@@ -562,7 +562,7 @@ def nanmode(x, axis=0):
 def unique(x, return_counts=False):
     """ Equivalent of np.unique that supports sparse or dense matrices. """
     if not sp.issparse(x):
-        return np.unique(x, return_counts=return_counts)
+        return np.unique(np.asarray(x), return_counts=return_counts)
```
markotoplak, line 565: Do you remember who called unique?

Orange/base.py
noahnovsak: I can't seem find why I put this in anymore. Guess I'll take it out for now and hope nothing breaks.

Orange/statistics/util.py
noahnovsak: in owlearnerwidget there's a `check_data` method, that makes sure that (among other things) Y isn't constant. I can move the computation up there instead if that makes more sense.

PR #6515, opened 2023-07-19
FreeViz: Allow setting ratio btw attractive and repulsive forces

### LINKED ISSUE #6479: Bringing back more control in FreeViz
I brought this up in #5501 earlier and still comes up from time to time in my work, so I'll post it here as a feature request.

Older FreeViz (orange v2.4 or 2.6, I think) gave a lot of control to the user over the attractive and repulsive forces of the optimization process. I often found that a temporary increase in repulsive forces can push it out of a local minimum when it's stuck, achieving much better results (in some cases to almost perfect separation of classes). Would it be possible to bring back the force settings and add an option to disable auto-stopping?

janezd:
##### Issue

Fixes #6479.

##### Description of changes

Adds an option to control the strength of attractive and repulsive forces.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owfreeviz.py
+++ b/Orange/widgets/visualize/owfreeviz.py
@@ -5,7 +5,7 @@ from types import SimpleNamespace as namespace
 import numpy as np
 
 from AnyQt.QtCore import Qt, QRectF, QLineF, QPoint
-from AnyQt.QtGui import QPalette
+from AnyQt.QtGui import QPalette, QFontMetrics
 from AnyQt.QtWidgets import QSizePolicy
 
 import pyqtgraph as pg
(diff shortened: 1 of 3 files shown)


processo:
Just wanted to add that with this new code changing _Initialization_ while _Gravity_ checked either does nothing or gives the error:
```
line 293, in init_projection
  gravity=self.gravity if self.balance else None
AttributeError: 'OWFreeViz' object has no attribute 'gravity'
```

Orange/widgets/visualize/owfreeviz.py
```diff
@@ -159,6 +163,7 @@ class Warning(OWAnchorProjectionWidget.Warning):
     def __init__(self):
         OWAnchorProjectionWidget.__init__(self)
         ConcurrentWidgetMixin.__init__(self)
+        self.__optimized = False
```
VesnaT, line 166: I suppose the flag should be reset, when a new data appears on the input.

VesnaT:
```
  File "/Users/vesna/orange-widget-base/orangewidget/gui.py", line 2261, in __call__
    self.func(**kwds)
  File "/Users/vesna/orange3/Orange/widgets/visualize/owfreeviz.py", line 231, in __init_combo_changed
    self.init_projection()
  File "/Users/vesna/orange3/Orange/widgets/visualize/owfreeviz.py", line 295, in init_projection
    gravity=self.gravity if self.balance else None)
AttributeError: 'OWFreeViz' object has no attribute 'gravity'
```

VesnaT (commented):
Adding some tests for the new code would be nice.
Also, lint does not pass.

PR #6520, opened 2023-07-20
[ENH] SubarrayComputeValue for faster domain transformation 

markotoplak:
##### Issue
Orange does domain transformation per column. Even columns computed with `SharedComputeValue` are filled in by column. This was especially slow for Dask.
 
##### Description of changes
This PR tries to assign column groups when it can. When just copying columns between tables, chunks of columns are now used instead of single ones. This PR introduces `SubarrayComputeValue` that computes a subset of columns at once. This is a `SharedComputeValue` with a limitation that shared results can not be further post-processed (as`SharedComputeValue` was actually used most of the time).

I implemented normalization and imputation with `SubarrayComputeValue`. Speedups with numpy tables are around 2x, and like 10x+ for dask tables.

I did not test performance on sparse arrays though. I have a feeling that would have to be optimized.

Most code here is independent of dask and could be merged straight into master, but I think it is best tested on something less popular. :)

##### Includes
- [X] Code changes
- [X] (Some) Tests
- [ ] Documentation


markotoplak:
After consultation with @noahnovsak and @lanzagar I am merging this into the `dask` branch.

PR #6521, opened 2023-07-20
Change setup.py to work with Cython 3.0

### LINKED ISSUE #6519: New version of cython was released, installation does not work with it
On 17 Jul Cython 3.0 was released. Orange can not be installed from source with it.

In long term, we should fix it. For now, I changed the installer to require older Cython (#6518).

janezd:
##### Issue

Fixes #6519.

Cython 3.0 with current setup.py cythonizes pyx -> c, but doesn't call compiler.

##### Description of changes

Call `cythonize`. Whyever.

I checked that there should be no problems with new default behaviour of `/` ("true division", producing floats). In most pyx files we specifiy `cdivision=True` anyway, so the behaviour of Cython remained the same).

If this is OK, we need to
- [ ] also merge https://github.com/biolab/orange3-network/pull/253
- [ ] check other officially supported extensions that use Cython; I checked those I have have installed)

Interestingly, with addon Associate we distribute .cpp. When was the last time anybody touched that add-on?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
     "setuptools>=41.0.0,<50.0",
     "wheel",
-    "cython<3.0",
+    "cython>=3.0",
     "oldest-supported-numpy",
     "sphinx",
     "recommonmark",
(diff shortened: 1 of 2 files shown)


markotoplak:
If we call cythonize manually I see no reason to use cython's `build_ext`.

I have no idea why this PR changes anything. @ales-erjavec, could you comment on this PR?

ales-erjavec (approved):

PR #6528, opened 2023-08-11
[ENH] PCA widget runs a separate thread

markotoplak:
##### Issue
For larger data PCA takes longer. Moving computation to a separate thread allows canvas to still be responsive.

##### Description of changes
The first two commits only refactor the widget to lose some attributes.

This PR only improves the responsiveness of canvas. I did not find any easy ways to interrupt PCA so that is not done here. 
*Edit*: I added a commit that allows for cancelling within preprocessing. Code was stolen from owoutliers and the current implementation in `Learner`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owpca.py
+++ b/Orange/widgets/unsupervised/owpca.py
@@ -63,7 +63,6 @@ class OWPCA(widget.OWWidget):
         self._transformed = None
         self._variance_ratio = None
         self._cumulative = None
-        self._init_projector()
 
         # Components Selection
         form = QFormLayout()
@@ -138,8 +137,6 @@ class OWPCA(widget.OWWidget):
                 self.clear_outputs()
                 return
 
-        self._init_projector()
-
         self.data = data
         self.fit()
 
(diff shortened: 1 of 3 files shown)


Orange/widgets/unsupervised/owpca.py
```diff
@@ -251,9 +244,15 @@ def _update_normalize(self):
...
-        self._pca_projector = PCA(n_components=MAX_COMPONENTS, random_state=0)
-        self._pca_projector.component = self.ncomponents
+    def _create_projector(self):
+        projector = PCA(n_components=MAX_COMPONENTS, random_state=0)
+        projector.component = self.ncomponents  # for use as a Scorer
+        if self.normalize:
+            projector.preprocessors = \
+                PCA.preprocessors + [preprocess.Normalize(center=False)]
+        else:
+            projector.preprocessors = PCA.preprocessors
```
noahnovsak, line 280: does this do anything?

markotoplak: Khm, it does not. Thanks

Orange/widgets/unsupervised/owpca.py
```diff
@@ -57,8 +58,9 @@ class Error(widget.OWWidget.Error):
 
     def __init__(self):
         super().__init__()
-        self.data = None
+        ConcurrentWidgetMixin.__init__(self)
```
noahnovsak, line 61: I'm still pretty sure this will initialize `ConcurrentWidgetMixin` twice, but if this is the more correct way of doing it I'm happy to start writing my code like this.

noahnovsak (approved):

Orange/widgets/unsupervised/owpca.py
markotoplak: It does not call `__init__` twice - I tested. :) Also, f I delete that line the widget stops working.

Issue #6367, opened 2023-03-23
Widgets fail with a dataset with hidden variables on the input.


PrimozGodec:
##### Description 

Some widgets fail when a dataset with hidden variables at the input. It can be reproduced with data from the Image Embedding widget. Currently, the following widgets are discovered to fail:

- [ ] Scatter Plot fail when I use Find informative projections, including hidden variables. It fails when I click on a combination that includes the hidden variable. See screenshot. My suggestion is to exclude hidden variables from the informative projections search.
![Screenshot 2023-03-23 at 16 10 06](https://user-images.githubusercontent.com/6421558/227249042-ca9f6d4d-87ab-47d5-921e-4af4ecc1ae2a.png)

- [ ] Group By fail immediately when I connect the data with hidden variables.

          Traceback (most recent call last):
            File "/Users/primoz/python-projects/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 1047, in __process_next
              if self.__process_next_helper(use_max_active=True):
            File "/Users/primoz/python-projects/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 1085, in __process_next_helper
              self.process_node(selected_node)
            File "/Users/primoz/python-projects/orange-canvas-core/orangecanvas/scheme/signalmanager.py", line 713, in process_node
              self.send_to_node(node, signals_in)
            File "/Users/primoz/miniconda3/envs/orange3/lib/python3.10/site-packages/orangewidget/workflow/widgetsscheme.py", line 806, in send_to_node
              self.process_signals_for_widget(node, widget, signals)
            File "/Users/primoz/miniconda3/envs/orange3/lib/python3.10/site-packages/orangewidget/workflow/widgetsscheme.py", line 820, in process_signals_for_widget
              process_signals_for_widget(widget, signals, workflow)
            File "/Users/primoz/miniconda3/envs/orange3/lib/python3.10/functools.py", line 889, in wrapper
              return dispatch(args[0].__class__)(*args, **kw)
            File "/Users/primoz/miniconda3/envs/orange3/lib/python3.10/site-packages/orangewidget/workflow/widgetsscheme.py", line 921, in process_signals_for_widget
              process_signal_input(input_meta, widget, signal, workflow)
            File "/Users/primoz/miniconda3/envs/orange3/lib/python3.10/functools.py", line 889, in wrapper
              return dispatch(args[0].__class__)(*args, **kw)
[...]

PR #6531, opened 2023-08-15
[FIX] Scatter Plot - Handle input features that are hidden in data

PrimozGodec:
##### Issue

Connected to #6367 
When some features are hidden in the data, and those features are present on Scatter Plot's feature input widget fails.

##### Description of changes
Handle describe scenario and do not plot anyting

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -526,6 +526,7 @@ class OWScatterPlot(OWDataProjectionWidget):
     def can_draw_regresssion_line(self):
         return self.data is not None and \
                self.data.domain is not None and \
+               self.attr_x is not None and self.attr_y is not None and \
                self.attr_x.is_continuous and \
                self.attr_y.is_continuous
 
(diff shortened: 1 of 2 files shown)


Orange/widgets/visualize/owscatterplot.py
```diff
@@ -571,12 +572,13 @@ def handleNewSignals(self):
...
-                all(attr in self.data.domain for attr
-                        in self.attribute_selection_list):
-            self.attr_x, self.attr_y = self.attribute_selection_list[:2]
+                self.data.domain is not None:
             self.attr_box.setEnabled(False)
             self.vizrank.setEnabled(False)
+            if all(attr in self.xy_model for attr in self.attribute_selection_list):
+                self.attr_x, self.attr_y = self.attribute_selection_list
+            else:
+                self.attr_x, self.attr_y = None, None
```
janezd, line 581: Perhaps it would be better to set `attr_x` and `attr_y` to the first two variables in the model, as default? Or did I miss something?

PrimozGodec: With this change, I wanted to explicitly show that the user provided the wrong features on the input (so nothing is plotted). Do you think it is better to plot the first two features and ignore the features input in this case?

PR #6533, opened 2023-08-16
Test & Score - fix data error conditions

noahnovsak:
Compute nans, number of values and number of features in training data.

### DIFF
--- a/Orange/widgets/evaluate/owtestandscore.py
+++ b/Orange/widgets/evaluate/owtestandscore.py
@@ -414,12 +414,12 @@ class OWTestAndScore(OWWidget):
                     "Train data input requires a target variable.",
                     not data.domain.class_vars
                 ),
-                ("Target variable has no values.", np.isnan(data.Y).all()),
+                ("Target variable has no values.", np.asarray(np.isnan(data.Y).all())),
                 (
                     "Target variable has only one value.",
-                    data.domain.has_discrete_class and len(unique(data.Y)) < 2
+                    data.domain.has_discrete_class and len(np.asarray(unique(data.Y))) < 2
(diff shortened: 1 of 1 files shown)


Orange/widgets/evaluate/owtestandscore.py
```diff
@@ -414,12 +414,12 @@ def set_train_data(self, data):
                     "Train data input requires a target variable.",
                     not data.domain.class_vars
                 ),
-                ("Target variable has no values.", np.isnan(data.Y).all()),
+                ("Target variable has no values.", np.asarray(np.isnan(data.Y)).all()),
```
markotoplak, line 417: So dask can not do `.all()`? If it can, then perhaps that could be  `np.asarray(np.isnan(data.Y).all())`?

noahnovsak: It can but .all() returns a boolean. Your suggestion would return an array with a Boolean inside. We could do .all().compute() but that won't play nice with numpy.

noahnovsak: Or np.asarray(...)[0] would also work, and would probably have a slightly smaller memory impact (but it seems like a case of premature optimization).

markotoplak: What I suggest returns a 0D-array which can be computed to bool. I know is premature optimization, but having a needless array of booleans in memory is ugly. This works for me in numpy and probably should also in dask:

```
>>> Y = np.ones(10)
>>> np.asarray(np.isnan(Y).all())
>>> bool(np.asarray(np.isnan(Y).all()))
False
```

PR #6542, opened 2023-08-20
[ENH] SOM: output columns with coordinates and errors

### LINKED ISSUE #6540: Add output coordinates to SOM
Improve on #6316.

Both output tables in SOM would get four additional columns:
- cell id (perhaps a strings like `"y5x2"`),
- two columns with cell coordinates,
- quantization error

#### Technical notes for whomever (probably me)

- Quantization error is computed in `get_winners` and `get_winners_sparse`; just store `min_diff` in another array and return it.
- The additional columns should be computed via `compute_value`, which must be a shared value (for all three columns), so they can also be used to projecting new data into an existing SOM
- Make the signal lazy. Perhaps implement a general function `lazy_table_from_domain`  that does a lazy domain transform. This almost (but not quite) belongs to `Orange.widgets.utils.annotated_data` or `Orange.widgets.utils.signals`.

janezd:
##### Issue

Resolves #6540.

##### Description of changes

Added columns, with lazy values and everything.

**To discuss**:
- `som_cell` column type.
- split into fitter and model or leave as is? (I'd keep it until (if ever) we really need separate classes)

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owsom.py
+++ b/Orange/widgets/unsupervised/owsom.py
@@ -1,10 +1,9 @@
 from collections import defaultdict, namedtuple
 from contextlib import contextmanager
-from typing import Optional
+from typing import Optional, Union
 from xml.sax.saxutils import escape
 
 import numpy as np
-import scipy.sparse as sp
 
 from AnyQt.QtCore import Qt, QRectF, pyqtSignal as Signal, QObject, QThread, \
     pyqtSlot as Slot
@@ -15,8 +14,10 @@ from AnyQt.QtWidgets import \
     QGraphicsItem, QGraphicsRectItem, QGraphicsItemGroup, QSizePolicy, \
     QGraphicsPathItem
 
-from Orange.data import Table, Domain
-from Orange.data.util import array_equal
+from Orange.widgets.utils.signals import lazy_table_transform
+
+from Orange.data import Table, Domain, ContinuousVariable, DiscreteVariable
+from Orange.data.util import array_equal, SharedComputeValue, get_unique_names
 from Orange.preprocess import decimal_binnings, time_binnings
 from Orange.projection.som import SOM
 
@@ -28,7 +29,8 @@ from Orange.widgets.settings import \
 from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.widgetpreview import WidgetPreview
 from Orange.widgets.utils.annotated_data import \
-    create_annotated_table, create_groups_table, ANNOTATED_DATA_SIGNAL_NAME
+    add_columns, group_values, \
+    ANNOTATED_DATA_SIGNAL_NAME, ANNOTATED_DATA_FEATURE_NAME
 from Orange.widgets.utils.colorpalettes import \
     BinnedContinuousPalette, LimitedDiscretePalette, DiscretePalette
 from Orange.widgets.visualize.utils import CanvasRectangle, CanvasText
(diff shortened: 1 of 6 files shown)


Orange/widgets/unsupervised/owsom.py
```diff
@@ -184,6 +187,96 @@ def disconnected_spin(spin):
...
+        winners[mask], distances[mask] = self.model.winners(cont_x)
+        winners += 1
+        return winners, distances
+
+    # `offsets` and `scales` are ndarray's (and `model` contains ndarray's)
+    # Unless we __eq__ by identity, we can't properly define hash.
+    def __eq__(self, other):
+        return type(self) is type(other) and \
+               self.domain == other.domain \
+               and self.model is other.model \
```
markotoplak, line 221: Could we call `==` and go with whatever it does? Usually they are not implemented, so in effect we get `is`.

Orange/widgets/unsupervised/owsom.py
```diff
@@ -184,6 +187,96 @@ def disconnected_spin(spin):
...
+               and np.array_equal(self.offsets, other.offsets) \
+               and np.array_equal(self.scales, other.scales)
+
+    def __hash__(self):
+        if self.__hash is None:
+            self.__hash = hash(
+                (hash(self.domain),
+                 id(self.model),
```
markotoplak, line 229: Instead of `id(self.model)` I'd just write self.model, and let __hash__ do whatever it does by default - probably something related to ID. :)

Also, no need to make individual hashes of the tuple element, calling just `hash((self.domain, self.model, ...` should suffice.

Orange/widgets/unsupervised/owsom.py
```diff
@@ -184,6 +187,96 @@ def disconnected_spin(spin):
...
+                 hash(tuple(self.offsets)), hash(tuple(self.scales))))
```
markotoplak, line 230: `__hash__` can be called often so unless these arrays are always small, this line can be slow.

markotoplak: I'd just use some elements from the arrays for speed and not worry about hash conflicts, which can happen anyway. Can't hurt much if they come up a bit more often.

Orange/widgets/unsupervised/owsom.py
```diff
@@ -184,6 +187,96 @@ def disconnected_spin(spin):
...
+            if a is not None:
+                a = a.copy()
+                a.flags.writeable = False
+            return a
+
+        self.domain = domain
+        self.model = model
+        self.offsets = immutable(offsets)
+        self.scales = immutable(scales)
+        self.__hash = None
```
markotoplak, line 205: When I cache hashes I try to care that the result is not pickled. Especially if hashing transient things like ids.

markotoplak (commented):
`__eq__` and `__hash__`

Orange/widgets/unsupervised/owsom.py
janezd: Some time ago I used to have habit of writing `return hash(...) ^ hash(...) ^ hash(...)` until I learned from other's code that this isn't OK. These individual hashes probably came from subconscious. :)

Orange/widgets/unsupervised/owsom.py
janezd: The size of the array equals the number of numeric attributes. SOM itself has a time complexity of #variables * #instances * #iterations, hence the computation of hash should be negligible in comparison of computing SOM. Also, the value is cached and hence computed only once per a SOM model - and we won't have many of them.

Fun fact: you can't use the `@cache` decorator on `__hash__`. :)

```
>>> from functools import cache
>>> class A:
...     @cache
...     def __hash__(self):
...         return 42
...
>>> a = A()
>>> hash(a)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

Orange/widgets/unsupervised/owsom.py
```diff
@@ -184,6 +187,100 @@ def disconnected_spin(spin):
...
+
+class SomErrorCompute(SharedComputeValue):
+    InheritEq = True
+
+    def compute(self, _, shared_data):
+        return shared_data[1]
+
+
+class GetGroups:
+    InheritEq = True
```
markotoplak, line 272: This one needs its own `__eq__` and `__hash__`.

Orange/widgets/unsupervised/owsom.py
```diff
@@ -184,6 +187,100 @@ def disconnected_spin(spin):
...
+                a.flags.writeable = False
+            return a
+
+        self.domain = domain
+        self.model = model
+        self.offsets = immutable(offsets)
+        self.scales = immutable(scales)
+        self.__hash = None
+
+    def __getstate__(self):
```
markotoplak, line 207: This class needs `__setstate__` too.

Orange/widgets/unsupervised/owsom.py
```diff
@@ -851,30 +949,72 @@ def rescale(self):
...
-                    indices[rows] = self.selection[x][y]
+                    group = self.selection[x][y]
+                    indices[rows] = group
+                    if group > 0:
+                        for id_ in self.data.ids[rows]:
+                            id_to_group[id_] = group
+
+        cont_domain = self._cont_domain(self.data)
+        shared_compute = SomSharedValueCompute(cont_domain, self.som, self.offsets, self.scales)
+        cell = StringVariable(
```
markotoplak, line 972: Is this what we discussed? If so, this should become a `DiscreteVariable`.

PR #6545, opened 2023-08-22
[ENH] Naive Bayes for Dask

noahnovsak:
Reimplement `_compute_contingency` on `DaskTable` to enable the naive Bayes learner (other uses of this function have not been tested and instead raise an error).

### DIFF
--- a/Orange/base.py
+++ b/Orange/base.py
@@ -444,7 +444,7 @@ class Model(Reprable):
         # Call the predictor
         backmappers = None
         n_values = []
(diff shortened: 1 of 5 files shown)


markotoplak:
We just talked about this, just posting here: there is a refactor in here that would also benefit master, and we should commit it there so that we'd get fewer merge conflicts.

markotoplak:
I merged #6545. Can you rebase?

markotoplak:
@noahnovsak, you probably forgot one import.

PR #6552, opened 2023-08-25
Table: Assure unique table.ids when unpickling

VesnaT:
##### Issue

Different Tables could have the same ids when unpickled.

##### Description of changes
Create new table.ids in `__setstate__`.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/sql/table.py
+++ b/Orange/data/sql/table.py
@@ -1,6 +1,7 @@
 """
 Support for example tables wrapping data stored on a PostgreSQL server.
 """
+import contextlib
 import functools
 import logging
 import threading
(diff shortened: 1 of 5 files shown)


markotoplak:
This is crazy! Changing setstate seems to have broken linux tests. Linux runs SQL tests too and our tests just freeze after a while.

markotoplak:
I figured out why tests froze. There was a deadlock on unpickling, when fresh ids were assigned for SqlTable. The old version of `_init_ids` locked `Lock` when accessing `X.shape`. That was problematic for SqlTables, because this downloaded the data, which tried locking again to obtain instance ids, which it could not. Therefore: deadlock!

I fixed the deadlock. Then I noticed that @VesnaT's original ids fix always caused data download for SqlTables, which I hope I fixed in the following commits.

In doing so I also had to free SqlTables from locking, because pickling tables with downloaded data did not work.

PR #6555, opened 2023-08-28
[ENH] IO - Change origin attribute when not find on system

PrimozGodec:
##### Issue
For example, files that Orange loads (with the File widget) may contain paths in one or more columns—for example, the path to images for image analysis. One way to store paths may be to keep them as origin prefixes in attributes of attribute and the last part of the path as column values. Paths (origin + column values) are absolute paths. When the table is transferred to another computer, the path may not be valid anymore. 

##### Description of changes
If the dataset's author provides files besides the CSV file, they may be discovered, and the origin can be fixed.

Additionally, I replaced StringIO reader inputs with actual files in the test. The code should not be adapted to cases only possible in tests.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io_util.py
+++ b/Orange/data/io_util.py
@@ -1,17 +1,27 @@
+import os.path
 import subprocess
 from collections import defaultdict
+from typing import Tuple, Optional
 
 import numpy as np
+import pandas as pd
 from chardet.universaldetector import UniversalDetector
 
 from Orange.data import (
     is_discrete_values, MISSING_VALUES, Variable,
-    DiscreteVariable, StringVariable, ContinuousVariable, TimeVariable,
+    DiscreteVariable, StringVariable, ContinuousVariable, TimeVariable, Table,
 )
 from Orange.misc.collections import natural_sorted
 
-__all__ = ["Compression", "open_compressed", "detect_encoding", "isnastr",
-           "guess_data_type", "sanitize_variable"]
+__all__ = [
+    "Compression",
+    "open_compressed",
+    "detect_encoding",
+    "isnastr",
+    "guess_data_type",
+    "sanitize_variable",
+    "update_origin",
+]
 
 
 class Compression:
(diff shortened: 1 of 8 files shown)


Orange/data/io_util.py
```diff
@@ -207,3 +216,65 @@ def mapvalues(arr):
         values = [_var.parse(i) for i in orig_values]
 
     return values, var
+
+
+def _extract_new_origin(attr: Variable, table: Table, lookup_dirs: Tuple[str]) -> Optional[str]:
+    # origin exist
```
janezd, line 222: -> exists

Orange/data/io_util.py
```diff
@@ -207,3 +216,65 @@ def mapvalues(arr):
...
+    # last dir of origin in lookup dirs
+    dir_ = os.path.basename(os.path.normpath(attr.attributes["origin"]))
+    for ld in lookup_dirs:
+        new_dir = os.path.join(ld, dir_)
+        if os.path.isdir(new_dir):
+            return new_dir
+
+    # all column paths in lookup dirs
+    for ld in lookup_dirs:
+        if all(os.path.exists(os.path.join(ld, v)) for v in table.get_column(attr)):
```
janezd, line 235: Maybe we would skip unknown values here, e.g. by adding `if v`?

Orange/data/io_util.py
```diff
@@ -207,3 +216,65 @@ def mapvalues(arr):
...
+
+
+def update_origin(table: Table, file_path: str):
+    """
+    When a dataset with file paths in the column is moved to another computer,
+    the absolute path may not be correct. This function updates the path for all
+    columns with an "origin" attribute.
+
+    The process consists of two steps. First, we identify directories to search
+    for files, and in the second step, we check if paths exit.
```
janezd, line 248: -> exist

Orange/data/io_util.py
```diff
@@ -207,3 +216,65 @@ def mapvalues(arr):
...
+
+    Lookup directories:
+    1. The directory where the file from file_path is placed
+    2. The parent directory of 1. The situation when the user palaces dataset
```
janezd, line 252: -> places

Orange/data/io_util.py
```diff
@@ -207,3 +216,65 @@ def mapvalues(arr):
...
+    table
+        Orange Table to be updated if origin exits in any column
+    file_path
+        Path of the loaded dataset for reference. Only paths inside datasets
+        directory or its parent directory will be considered for new origin.
+    """
+    file_dir = os.path.dirname(file_path)
+    parent_dir = os.path.dirname(file_dir)
+    # if file_dir already root file_dir == parent_dir
+    lookup_dirs = tuple({file_dir, parent_dir})
```
janezd, line 275: We probably want to look into `file_dir` first, and only then into `parent_dir`? Sets are unordered.

If you want to keep it short, use `tuple({file_dir: 0, parent_dir: 0})`. :)

Orange/data/io_util.py
```diff
@@ -207,3 +216,65 @@ def mapvalues(arr):
...
+    for attr in table.domain:
```
janezd, line 276: Why not just `table.domain.metas`?

Not only because of efficiency; I'm never sure whether a loop over domain includes metas or not. :)

Orange/data/io_util.py
```diff
@@ -207,3 +216,65 @@ def mapvalues(arr):
...
+        if "origin" in attr.attributes:
```
janezd, line 277: You could add `if attr.is_string` as precaution. Is someone uses `origin` for something else this could prevent some false positives.

janezd (approved):
In general, I'm not very happy about this guessing, but something has to be done and I have no better idea. It would probably do the work.

Orange/data/io_util.py
PrimozGodec: You are right. Fixed

Orange/data/io_util.py
PrimozGodec: Good idea. Also image analytics only consider metas https://github.com/biolab/orange3-imageanalytics/blob/02356b8c14b2ec2d63f3a2f697ce69e47f05c4fa/orangecontrib/imageanalytics/utils/image_utils.py#L15-L33

Orange/data/io_util.py
PrimozGodec: I extended this condition on discrete variables also since Image Analytics also search in those: https://github.com/biolab/orange3-imageanalytics/blob/02356b8c14b2ec2d63f3a2f697ce69e47f05c4fa/orangecontrib/imageanalytics/utils/image_utils.py#L15-L33

PR #6570, opened 2023-09-12
Gradient boosting - Adapt tests to version xgboost=2.0.0

PrimozGodec:
##### Issue

Xgboost released version 2.0.0, which changes some internal interfaces and breaks our tests.

##### Description of changes
Adapt tests to 2.0.0.

Also, now [Xgboost packs Libomp](https://github.com/dmlc/xgboost/pull/7621) in wheels for Macos (I think it was a primary reason for keeping Xgbosot and Catboost optional dependency). I suggest including both as a mandatory dependency. They are packed in Installers and part of a conda-forge package already/anyway.

##### Includes
- [ ] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
Does new xgboost only break tests or also widgets? If the latter, we will perhaps need 3.36.1.

markotoplak:
Oh, I now see there were just the tests.

PrimozGodec:
@markotoplak, no need for 3.36.1. It breaks only tests.

markotoplak:
Can you also update meta.yaml?

Conda-forge only has xgboost 1.7 for now, but that is likely not much of a problem. Anything that relies on building Orange though `conda build` would fail until they release 2.0. Maybe that is not much of a problem.

PrimozGodec:
I updated meta.yaml. I didn't fix the version there. I think it should work for now.

PR #6577, opened 2023-09-15
[ENH] Predictions: Output errors

### LINKED ISSUE #6572: Predictions: include error in 'Predictions' output
**What's your use case?**
When applying the Predictions widget, I'd like to visualize the prediction error for each record, for instance in a scatter plot or violin plot.

Prediction already calculates the error (in the case of regression, various flavors of it) and shows it in its dialog box. Should be no problem to also include it in the 'Predictions' output. Perhaps even all of the available error types in separate columns.
For classification, it may also be useful to have the probabilities per class in the output.

Yes, create an extra variable "error" using Formula (formerly Feature Constructor). But it's in fact redoing a simple calculation that was already made.

VesnaT:
##### Issue

Fixes #6572

##### Description of changes
Attach prediction errors to output when the errors are shown in the table.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -265,6 +265,12 @@ class OWPredictions(OWWidget):
     def is_discrete_class(self):
         return bool(self.class_var) and self.class_var.is_discrete
 
+    @property
+    def shown_errors(self):
+        return self.class_var and (
+            self.show_probability_errors if self.is_discrete_class
+            else self.show_reg_errors != NO_ERR)
+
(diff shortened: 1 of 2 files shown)


markotoplak:
@janezd, this PR seem like a cool addition. Initially I was against (should we really output things that people can easily compute in Formula?), but I changed my mind in discussion with Vesna. The single error column per predictor seems like a sensible idea (and users can also select error types in the widget, which was the thing that convinced me).

Maybe this could also go into 3.36.1, if you perhaps have to review it?

PR #6582, opened 2023-09-19
Daskreader: don't pickle Domain

noahnovsak:
Avoid pickling domain when writing to hdf5.


Orange/data/dask.py
```diff
@@ -2,15 +2,17 @@
 import pickle
 import warnings
 from numbers import Integral
+from os import path
+import json
 
 import h5py
-import dask
+import dask.delayed
```
markotoplak, line 9: Perhaps unused import?

Orange/data/dask.py
```diff
@@ -122,24 +127,52 @@ def from_file(cls, filename, sheet=None):
         """
         h5file = f = h5py.File(filename, "r")
 
-        def read_format_da(name):
+        def read(sub):
```
markotoplak, line 130: Needs a better name. Perhaps `read_domain`? :)

Orange/data/dask.py
```diff
@@ -160,9 +193,27 @@ def __getitem__(self, key):
...
         # the following code works with both Table and DaskTable
+        def parse(attr):
+            params = (attr.name, attr.TYPE_HEADERS[1], {"attributes": attr.attributes})
+            if isinstance(attr, DiscreteVariable):
+                params[2].update(values=attr.values)
+            elif isinstance(attr, TimeVariable):
+                params[2].update(have_date=attr.have_date,
+                                 have_time=attr.have_time)
+            elif isinstance(attr, ContinuousVariable):
+                params[2].update(number_of_decimals=attr.number_of_decimals)
```
markotoplak, line 204: Doesn't ContinuousVariable have more attributes on how to display stuff than the number of decimals? Maybe I mis-remember.

noahnovsak: I believe there is one more, but it's calculated from `number_of_decimals`.

Orange/data/dask.py
markotoplak: Cool, thanks!

Orange/tests/test_dasktable.py
```diff
@@ -74,7 +74,7 @@ def _wrapper(self):
         test_case(self, lambda table: table)
         self.tearDown()
         self.setUp()
-        test_case(self, lambda table: temp_dasktable(table))
```
markotoplak, line 77: Why did you need to change this? (I forgot why there were lambdas here)

noahnovsak: It seemed like the lambda doesn't need to be there, temp_dasktable is already callable.
I can't remember if there was a reason to write it like that in the first place either.

markotoplak: Ah, I see, forget it.

markotoplak:
I rebased some fixes out so that they are easier to move and then joined last two commits. I added some compatibility functions with the old format so that we can also use our old files.

PR #6584, opened 2023-09-21
Gradient Boosting - Remove try/catch around imports

PrimozGodec:
##### Issue

After https://github.com/biolab/orange3/pull/6570, try/catch around xgboost/catboost methods is not required anymore since packages are always installed

##### Description of changes
Remove try/catch around imports, skips in tests and simplify part of the code

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
Come on, let's wait a bit with this so that we see if the added requirements create problems for anyone. I am making this a draft.

PrimozGodec:
I think it can be reviewed and merged now.

markotoplak:
I'd wait with this a little bit now that newest xgboost version started not packing libomp again. Perhaps we should even think about dropping the requirement...

PR #6585, opened 2023-09-21
Allow scikit-learn 1.3

markotoplak:
scikit-learn 1.3 fixes a bug introduced by AdaBoost deprecations in 1.2

Bug: https://github.com/scikit-learn/scikit-learn/issues/26241

Fix was released in 1.3 (from changelog):
Fix deprecation of base_estimator in [ensemble.AdaBoostClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.AdaBoostClassifier.html#sklearn.ensemble.AdaBoostClassifier) and [ensemble.AdaBoostRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.AdaBoostRegressor.html#sklearn.ensemble.AdaBoostRegressor) that was introduced in [#23819](https://github.com/scikit-learn/scikit-learn/pull/23819). [#26242](https://github.com/scikit-learn/scikit-learn/pull/26242) by [Marko Toplak](https://github.com/markotoplak).

But 1.3 also introduced some changes to the `fit()` internals, therefore old "magic" that was supposed to find out whether `fit()` supports the weights too stopped working. I adapted what we had to 1.3 and also wrote some tests for it. I think the best way to go forward would actually be to remove that magic and hardcode it per-learner.

### DIFF
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -17,7 +17,7 @@ pip>=18.0
 python-louvain>=0.13
 pyyaml
 requests
-scikit-learn>=1.1.0,<1.2.0
+scikit-learn>=1.1.0,!=1.2.*  # ignoring 1.2.*: scikit-learn/issues/26241
 scipy>=1.9
 serverfiles		# for Data Sets synchronization
 setuptools>=51.0.0
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -17,7 +17,7 @@ pip>=18.0
 python-louvain>=0.13
 pyyaml
 requests
-scikit-learn>=1.1.0,!=1.2.*  # ignoring 1.2.*: scikit-learn/issues/26241
+scikit-learn>=1.1.0,!=1.2.*,<1.4  # ignoring 1.2.*: scikit-learn/issues/26241
 scipy>=1.9
 serverfiles		# for Data Sets synchronization
 setuptools>=51.0.0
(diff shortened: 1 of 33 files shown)


markotoplak:
The most of this PR is adapting a hack from 2016 to scikit-learn 1.3.

noahnovsak:
The fix seems fine. I feel like a warning or deadline would be nice though, to scare us into hardcoding the values instead.

markotoplak:
@noahnovsak, if we decide against this magic, then the way to do it would be to deprecate the baseclass property with OrangeDeprecationWarning and manually set it as an attribute in derived classes. Then, if tests still pass, we are happy. No need for deadlines.

requirements-core.txt
```diff
@@ -17,7 +17,7 @@ pip>=18.0
 python-louvain>=0.13
 pyyaml
 requests
-scikit-learn>=1.1.0,<1.2.0
+scikit-learn>=1.1.0,!=1.2.*  # ignoring 1.2.*: scikit-learn/issues/26241
```
PrimozGodec, line 20: Is there any specific reason why meta.yaml includes `<1.4` version scikit-learn constraint while core doesn't?

PrimozGodec (commented):
Besides a comment it looks good to me.

requirements-core.txt
markotoplak: Yes. 1.4 deprecates some things that are still used but unfixed by this PR.

markotoplak:
@PrimozGodec, I was betting on you to complain that the deprecation does not have an associated time bomb. :)

PrimozGodec:
> @PrimozGodec, I was betting on you to complain that the deprecation does not have an associated time bomb. :)

Yep, didn't think about that before, but it would be nice. :)

PR #6601, opened 2023-10-20
Distance File: Don't accept dropped .xlsx files

janezd:
##### Issue

When dropping an .xlsx file onto canvas, the user has to choose between Distance File and File, which is confusing (because these are widget names). The user almost always wants to load an .xlsx with File.

##### Description of changes

Distance File's functionality now no longer accepts .xlsx files.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/unsupervised/owdistancefile.py
+++ b/Orange/widgets/unsupervised/owdistancefile.py
@@ -124,6 +124,8 @@ class OWDistanceFile(widget.OWWidget, RecentPathsWComboMixin):
                 err = str(exc)
                 self.Error.invalid_file(" \n"[len(err) > 40] + err)
             else:
+                # If you add any other checks before accepting the file,
+                # you should probably mirror them in canDropFile
                 if distances.shape[0] != distances.shape[1]:
                     self.Error.non_square_matrix()
                 else:
@@ -158,7 +160,14 @@ class OWDistanceFileDropHandler(SingleFileDropHandler):
         return {"recent_paths": stored_recent_paths_prepend(self.WIDGET, r)}
 
     def canDropFile(self, path: str) -> bool:
-        return os.path.splitext(path)[1].lower() in (".dst", ".xlsx")
+        if os.path.splitext(path)[0] == ".dst":
+            return True
+        try:
+            distances = DistMatrix.from_file(path)
+        except Exception:  # pylint: disable=broad-except
+            return False
+        else:
+            return distances.shape[0] == distances.shape[1]
 
 
 if __name__ == "__main__":  # pragma: no cover
--- a/Orange/widgets/unsupervised/tests/test_owdistancefile.py
+++ b/Orange/widgets/unsupervised/tests/test_owdistancefile.py
@@ -35,9 +35,13 @@ class TestOWDistanceFile(WidgetTest):
 
 class TestOWDistanceFileDropHandler(unittest.TestCase):
     def test_canDropFile(self):
+        def ren(name):
+            return os.path.join(os.path.split(Orange.tests.__file__)[0], name)
+
         handler = OWDistanceFileDropHandler()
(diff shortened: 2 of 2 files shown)


Orange/widgets/unsupervised/owdistancefile.py
```diff
@@ -158,7 +160,12 @@ def parametersFromFile(self, path):
         return {"recent_paths": stored_recent_paths_prepend(self.WIDGET, r)}
 
     def canDropFile(self, path: str) -> bool:
-        return os.path.splitext(path)[1].lower() in (".dst", ".xlsx")
+        try:
+            distances = DistMatrix.from_file(path)
+        except Exception:  # pylint: disable=broad-except
+            return False
+        else:
+            return distances.shape[0] == distances.shape[1]
```
VesnaT, line 168: Have we agreed on that?

Orange/widgets/unsupervised/owdistancefile.py
```diff
@@ -158,7 +160,12 @@ def parametersFromFile(self, path):
         return {"recent_paths": stored_recent_paths_prepend(self.WIDGET, r)}
 
     def canDropFile(self, path: str) -> bool:
-        return os.path.splitext(path)[1].lower() in (".dst", ".xlsx")
+        try:
```
VesnaT, line 163: You could keep the old condition (only .dst) for better performance.

Orange/widgets/unsupervised/owdistancefile.py
```diff
@@ -158,7 +160,12 @@ def parametersFromFile(self, path):
...
+            distances = DistMatrix.from_file(path)
```
VesnaT, line 164: This may take too long for big files.

Orange/widgets/unsupervised/owdistancefile.py
janezd: Yes. The widget also tests this (and shows an error) when loading the file.

Orange/widgets/unsupervised/owdistancefile.py
janezd: What was the file dimension again? Can you tell me which file it was - or send it? I tried 600x1000 and takes a few seconds.

What I discovered was that, unfortunately, most time is spent in opening the Excel file. Hence, the File widget will take more or less the same time as well.

Of course, with this PR, the file is read twice and we must prevent this. I'm considering decorating functions that open excel files with a lru cache of size 1 that would remember the last read excel workbook - of course checking not only the file name but also the timestamp. In this way - and given that opening the file takes majority of time - the time spent in owdistancefile would be saved in owfile because it would reuse the workbook object.

VesnaT: It was 1000x1000 .xlsx.

janezd:
Note: taking a file and dragging it around the canvas triggers an incessant, continuous stream of calls to `canDropFile`, which freezes the app (and possible also Finder or another part of (another) OS). This posed no problem when `canDropFile` wasn't sniffing the file, but just looked at its name. For this PR to work, this has to be fixed, too.

janezd:
@ales-erjavec, can we assume that `dragMoveEvent` is always preceded by `dragEnterEvent`?

Could we avoid incessant calling of `canHandleDrop` in [`orangecanvas.document.interations.DropAction.dragMoveEvent`](https://github.com/biolab/orange-canvas-core/blob/124c559535f586b95beaf656aac3a4b811761638/orangecanvas/document/interactions.py#L2161), by changing  `if self.canHandleDrop(event)` to `if self.__designatedAction is not None`?

ales-erjavec:
canHandleDrop should not do anything computationally expensive. Only check filename, possibly read a limited amount of the file (e.g. magic header byte[s])...

PR #6604, opened 2023-10-20
Edit Domain: allow the (existing) checkbox to also control removal of newly created compute value

### LINKED ISSUE #6566: EditDomain: Unlink variable from its source variable
**What's wrong?**
In the edit domain, the "Unlink variable from its source" checkbox is always disabled (or so it seems).

Screenshot 2023-09-06 at 11 40 29

This creates a problem later when we want to apply transformations. I think this workflow is enough to illustrate what the problem is:

Screenshot 2023-09-06 at 11 50 36

Essentially, Orange sees 'Time' as two different variables because we performed transformations in the edit domain. However, in this case, one would like to treat those two variables as if they are the same.

What needs discussion:
- Is this checkbox disabled for a reason or is this a bug?
- Should we make this option checked by default (for those who know whats up, they will disable it)?
- What about apply domain? Should we support something similar there?

File -> Edit Domain -> try to click on "Unlink variable from its source"

janezd:
##### Issue

Fixes #6566. See in particular https://github.com/biolab/orange3/issues/6566#issuecomment-1772746527.

##### Description of changes

The first commit does the job and is trivial: checkbox is no longer disabled, and having `compute_value` is no longer required for unlinking.

The second commit is not needed for fixing #6566. However, now that `compute_value` is no longer required for unlinking, we do not need to check it in hint matching. I thus removed it from the key in hint dictionary and consequently from settings. This required a zillion small changes in tests (removals of the additional argument), settings migration and tests for settings migration.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/oweditdomain.py
+++ b/Orange/widgets/data/oweditdomain.py
@@ -107,8 +107,7 @@ class Categorical(
     _DataType, NamedTuple("Categorical", [
         ("name", str),
         ("categories", Tuple[str, ...]),
-        ("annotations", AnnotationsType),
-        ("linked", bool)
+        ("annotations", AnnotationsType)
     ])): pass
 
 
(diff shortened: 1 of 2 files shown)


JakaKokosar:
LGTM, thanks! :)

PR #6612, opened 2023-10-27
Orange.utils: Add decorator 'allot'

janezd:
##### Issue

When animating an optimization like laying out a network (and supposedly also MDS, tSNE, freeviz) frequent updates of the graph can make the gui unresponsive.

In particular, when optimizing a network layout in the network add-on, the user can temporarily decrease the gravity and then increase it again, using a slider. The slider is unresponsive when the network has a lot of connections.

##### Changes

Decorator `allot(p)` allocates a given proportion of time to the function. I tried it in the Network Explorer, decorating `on_partial_result` with `allot(0.02)` makes the widget responsive on graphs with 10000 nodes and edges.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/util.py
+++ b/Orange/util.py
@@ -1,11 +1,14 @@
 """Various small utilities that might be useful everywhere"""
 import logging
 import os
+import time
 import inspect
 import datetime
 import math
+import functools
 from contextlib import contextmanager
-from typing import TYPE_CHECKING, Callable
+from typing import TYPE_CHECKING, Callable, Union, Optional
+from weakref import WeakKeyDictionary
 
 import pkg_resources
 from enum import Enum as _Enum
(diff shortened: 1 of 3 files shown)


Orange/util.py
```diff
@@ -180,6 +183,135 @@ def wrapper(*args, **kwargs):
...
+            decorator = cls(
+                self.allotted_time,
+                overflow=bound_overflow,
+                _bound_methods=self.__bound_methods)
+            self.__bound_methods[inst] = decorator(self.func.__get__(inst, cls))
+
+        return self.__bound_methods[inst]
+
+    def __call__(self, *args, **kwargs):
+        if time.perf_counter() < self.no_call_before:
```
markotoplak, line 298: Why not time.process_time() or even thread_time() (if available on that system)? Maybe a bit more useful in multitasking environment. But these do not count time speed in sleep so tests would need to be changed. :)

janezd: > Why not time.process_time() or even thread_time() (if available on that system)? Maybe a bit more useful in multitasking environment. But these do not count time speed in sleep so tests would need to be changed. :)

OK. I've put the change into a separate commit. If you merge, squash them.

markotoplak:
Thanks for the tests. These were beautiful.

PR #6614, opened 2023-10-29
Start Vizrank when opened, pause when closed

janezd:
##### Issue

Fixes #6504. Also fixes #6624.

##### Description of changes

- Vizrank starts immediately, as the user presses "Suggest Features", and closing it **pauses** it. 
- Re-opening after closing restarts it.
- Changing the settings (number of attributes) also pauses it. The user can restart with the new setting or continue with the old.

(Rather) unrelated, I disabled `duplicate-code` issue in pylint, as we discussed a few weeks ago. This is a new test, it is too strict, and also unnecessary in our context: duplicated code was never an issue for us, so we will get just plenty of pointless issues. Including here, which is why I've had enough in this PR.

List of commits:

- Commit 1: Auto-start/pause for Linear Projection.
- Commit 2: Fixes #6624.
- Commit 3: Minor change in the runner. I think this is cleaner, plus it reduce the block enclosed in try-except .
- Commit 4: Refactoring: 
    - `VizRankDialogNAttrs`, a VizRank dialog in which user sets the number of variables, with functionality shared between Linear projection and Radviz
    - `CloseVizrankMixin` that should be mixed into all widgets with vizranks so that they stop vizrank when they are closed
    - `auto_start: bool` argument for `VizRankDialog.add_vizrank`, which causes computation to start when the vizrank button is pressed, and pauses it when dialog is closed.
- Commit 5: Auto-start/pause for Radviz.
- Commit 6: Auto-start/pause for Mosaic.
- Commit 7: Sieve and Scatter plot.
- Commit 8: Disable `duplicate-code` in pylintrc

##### Includes
- [X] Code changes
- [X] Some tests

PR #6627, opened 2023-11-07
Implement __eq__ and __hash__ for PCA (and family)

markotoplak:
##### Issue
There are too many warnings when using PCA.

Having worked with this code, I saw that just about everything is wrong with this class hierarchy - it is much too circular. These classes need to be rewritten - at first I though I won't even be able to implement `__eq__` and `__hash__`, but then I somehow remembered the trick in the second commit.

I also added a tiny test. I could theoretically test all classes in the chain separately by initializing them manually and call corresponding functions, but am not sure that would make sense. Do I have to?

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/base.py
+++ b/Orange/projection/base.py
@@ -107,7 +107,8 @@ class Projector(ReprableWithPreprocessors):
 
 class Projection:
     def __init__(self, proj):
-        self.__dict__.update(proj.__dict__)
+        if proj is not None:
+            self.__dict__.update(proj.__dict__)
         self.proj = proj
 
     def transform(self, X):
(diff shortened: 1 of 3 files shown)


janezd:
"(...) for PCA", and for other classes derived from `Projection`, one of which is Linear Projection.

Thank you! I disabled these warnings just to be able to normally work on  #6614!

markotoplak:
@janezd, I did what I could to make warnings go away. It is not pretty (see the second commit), but for a straightforward implementations the whole `Orange.projection` would need refactoring. I decided first to solve the immediate issue (warnings) and leave refactoring for some other time - who know when that would finish.

PR #6637, opened 2023-11-17
adaboost: adapt to scikit-learn's 1.4 deprecation of base_estimator

markotoplak:
##### Issue
/home/runner/work/orange3/orange3/.tox/orange-latest/lib/python3.11/site-packages/sklearn/ensemble/_base.py:156: FutureWarning: `base_estimator` was renamed to `estimator` in version 1.2 and will be removed in 1.4.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/i18n/si.jaml
+++ b/i18n/si.jaml
@@ -2010,14 +2010,18 @@ distance/distance.py:
         def `compute_distances`:
             hamming: false
 ensembles/ada_boost.py:
+    def `base_estimator_deprecation`:
+        '`base_estimator` is deprecated (to be removed in 3.39): use `estimator` instead.': false
     SklAdaBoostClassificationLearner: false
     SklAdaBoostRegressionLearner: false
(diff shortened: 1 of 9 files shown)


i18n/si.jaml
```diff
@@ -2010,14 +2010,18 @@ distance/distance.py:
         def `compute_distances`:
             hamming: false
 ensembles/ada_boost.py:
+    def `base_estimator_deprecation`:
+        '`base_estimator` is deprecated (to be removed in 3.39): use `estimator` instead.': false
```
lanzagar, line 2014: This should not be `false`. I think you either provide a translation or leave it `null` for someone else to translate it later.

janezd: It's OK, this message is not intended to be seen by the user. It should be in English.

PR #6648, opened 2023-11-21
[CI] Test with scientific python nightly wheels 

markotoplak:
##### Issue
Sometimes Orange starts crashing when newer scikit-learn & company get released. We could at least anticipate that.

~~Needs #6637 to be merged first.~~ I remembered doing something like this when I was removing `scikit-learn<1.4` limit in #6637.

##### Description of changes
Adds a nightly build repository to one of our tests.


markotoplak:
@PrimozGodec, can you check this one, please?

PR #6661, opened 2023-12-01
VizRank: Reimplementation

### LINKED ISSUE #6504: Linear projection plot scoring keeps running after window is closed
Not sure if this is intended or not, but when I start a search in _Linear projection_ (Suggest features) if I don't pause it it will keep running even when the _Linear projection_'s window is closed.

How to reproduce:
1. open _Linear projection_, _Suggest features_, start
2. close _Score plots_ window
3. close _Linear projection_ window
4. it is still running in the background (with no progress indicator around the widget icon)
5. you can open _Linear projection_ again, open _Suggest features_ and see it running

OS: Windows 10 x64
Orange: 3.35.0

### LINKED ISSUE #6624: Linear Projection: VizRank skips variables with missing data
**What's wrong?**

VizRank skips variables with missing data although they may be useful.

- Take heart disease and try vizrank with 5 variables. "major vessels colored" won't appear.
- Try 6 variables and you get nothing (because any combination would include "major vessels colored".

Exception is thrown by `fit` in `knn = NearestNeighbors(n_neighbors=n_neighbors).fit(ec)`.

A solution would be to mask out the rows with missing data.

janezd:
##### Issue

Fixes #6504. Also fixes #6624.

#6614 would fixed the above problems, but poking the VizRank code that didn't age well introduced new bugs, and testing exposed some existing bugs.

This PR implements a new, cleaner and leaner VizRank.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation

### DIFF
--- a/Orange/widgets/visualize/owlinearprojection.py
+++ b/Orange/widgets/visualize/owlinearprojection.py
@@ -3,32 +3,33 @@ Linear Projection widget
 ------------------------
 """
 
-from itertools import islice, permutations, chain
-from math import factorial
+from itertools import islice, permutations, chain, combinations
+from math import factorial, comb
 
 import numpy as np
 
 from sklearn.neighbors import NearestNeighbors
 from sklearn.metrics import r2_score
 
-from AnyQt.QtGui import QStandardItem, QPalette
-from AnyQt.QtCore import Qt, QRectF, QLineF, pyqtSignal as Signal
+from AnyQt.QtGui import QPalette
+from AnyQt.QtCore import QRectF, QLineF
 
 import pyqtgraph as pg
 
-from Orange.data import Table, Domain
+from Orange.data import Table, Domain, IsDefined
 from Orange.preprocess import Normalize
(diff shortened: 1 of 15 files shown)


VesnaT:
**Mosaic**
- rename Vizrank dialog to *Score Plots* (for consistency)
- select the first row when computation finishes
- progress > 100%
image
- the orange lines are missing on the top rows
image

**Sieve**
- select the first row when computation finishes

**Linear Projection**
- select the first row when computation finishes
- progress > 100% 

**Radviz**
- select the first row when computation finishes
- progress > 100% 
- used to show scores

**Scatter Plot**
- select the first row when computation finishes

janezd:
I now change the code so that
- if the user closes the dialog and reopens it, ranking is restarted, but
- if the user pauses ranking and closes the dialog, and the reopens it, ranking is not restarted.

This is in a separate commit and not explicitly tested. Perhaps I may be persuaded to add some tests, though they'd be rather stupid.

PR #6668, opened 2023-12-07
[FIX] Fix impute.Model for derived domains

### LINKED ISSUE #6644: predictions displaying error that the target is not in domain
**What's wrong?**

 the predictions is displaying that error 
 Some predictor(s) failed (see more ...)
- Random Forest: Random Forest: 'fetal_health' is not in domain
image
image
image

[fetal_health_data.csv](https://github.com/biolab/orange3/files/13403821/fetal_health_data.csv)

**What's your environment?**

- Operating system: windows 11
- Orange version: 3.31.1
- How you installed Orange: downloaded from the website

markotoplak:
##### Issue
Fixes #6644

##### Description of changes
The compute_value was missing transformation into the variable space it was working upon.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/preprocess/impute.py
```diff
@@ -207,6 +206,9 @@ def __call__(self, data):
         column[mask] = predicted
         return column
 
+    def transform(self, c):
+        pass  # abstract in Transformation, never used here
```
janezd, line 210: Wouldn't it be better to `assert False, "abstract in Transformation, never used here"`, so that any erroneous calls are caught immediately??

PR #6670, opened 2023-12-08
[FIX] Writers: Store nans in StringVariable as empty strings instead of 'nan'

### LINKED ISSUE #6527: Save Data widget - can not same Excel files when unknows are present
When I use the SAVE DATA widget, this does not respond, it doesn't save anything and doesn't export any spreadsheet or any format that it's supposed to be generated by using this widget. I have removed the program and re-installed it, also I have tried to use the widget in other terminals. Didn't work. 
ORANGE VERSION 3.35

janezd:
##### Issue

Fixes #6527.

##### Description of changes

The function that creates a string representation for storing values in file didn't expect that a `StringVariable` can be `np.nan`. Now it does.

Our handling of StringVariable unknowns is a mess; the Table widget will apparently sometimes show `nan`. This must be solved in general. The present PR only fixes the crash in saving of Excel files.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/data/io_base.py
+++ b/Orange/data/io_base.py
@@ -15,6 +15,7 @@ from fnmatch import fnmatch
 from glob import glob
 
 import numpy as np
+import pandas
 
 from Orange.data import Table, Domain, Variable, DiscreteVariable, \
     StringVariable, ContinuousVariable, TimeVariable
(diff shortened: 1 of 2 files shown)


markotoplak:
I think this should be made specific to CSV writer. Any reason you did not make it that way?

Tab writers work well as they are and this fix would there only introduce new ambiguities.

markotoplak:
Ignore the previous comment. I need to check what did it do in write/read combination. If the old .tab behavior was already broken, I have no objections.

markotoplak:
Our tab saving was also broken, because it saved a string `nan` in such cases. Therefore, merging.

PR #6677, opened 2023-12-14
Minor fixes in Mosaic and Sieve vizrank

janezd:
##### Issue

Minor improvements over #6661.

##### Description of changes

**Mosaic:**
- Mosaic: disable vizrank for less than two variables
- Mosaic: disable the option to select three or four variables when the number of variables is smaller than that.
- Mosaic: don't declare vizrank dialog as setting provider because it no longer is, and can't be.

**Sieve:**

- Remove unused `caption_title`.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/visualize/owmosaic.py
+++ b/Orange/widgets/visualize/owmosaic.py
@@ -20,7 +20,7 @@ from Orange.preprocess.score import ReliefF
 from Orange.statistics.distribution import get_distribution, get_distributions
 from Orange.widgets import gui, settings
 from Orange.widgets.settings import (
-    Setting, DomainContextHandler, ContextSetting, SettingProvider)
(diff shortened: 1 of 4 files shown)


Orange/widgets/visualize/owmosaic.py
```diff
@@ -395,7 +404,7 @@ def set_data(self, data):
 
     def init_vizrank(self):
         if self.data is not None and len(self.data) > 1 \
-                and len(self.data.domain.attributes) >= 1:
+                and len(self.data.domain.attributes) >= 2:
```
VesnaT, line 407: The condition should be applied to `len(self.data.domain.variables + self.data.domain.metas)`. (The mosaic is build upon all variables in the domain).

PR #6718, opened 2024-01-26
Predictions: Output annotated table

### LINKED ISSUE #6711: Predictions should have output with annotated data
**What's your use case?**

1. Train a model, test it in Predictions.
2. In Predictions, select all rows for which the predicted probability of the positive outcome is above certain threshold.
3. Feed *the entire data set* into Distributions (or another widget) to analyze predicted vs. actual class distributions by plotting the outcome and grouping by selection (or vice versa, depending on the intent of visualization).

One can't do the last step because Prediction's only output is the selected data. 

Predictions is similar to Table, so it could have a similar output with annotated data. The current output would be renamed to Selection.

One can connect Predictions to Table and select rows there. This doesn't look so elegant (Table is not formatted specifically for predictions) and involves an extra step that could be avoided without any cost to gui or to implementation of the Predictions widget - especially if the output signal is lazy.

janezd:
##### Issue

Fixes #6711.

##### Description of changes

Add a (potentially lazy) output with annotated predictions.

I haven't made the existing output lazy, although this should be done at some point -- but perhaps as a general change in other widgets as well.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -18,6 +18,7 @@ from AnyQt.QtCore import (
 
 from orangecanvas.utils.localization import pl
 from orangewidget.utils.itemmodels import AbstractSortTableModel
+from orangewidget.utils.signals import LazyValue
 
 import Orange
 from Orange.evaluation import Results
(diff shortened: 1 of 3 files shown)


Orange/widgets/evaluate/owpredictions.py
```diff
@@ -855,12 +859,17 @@ def _commit_predictions(self):
...
-            predictions = predictions[datamodel.mapToSourceRows(...)]
-        self.Outputs.predictions.send(predictions)
+            selected = predictions[rows]
+            annotated_data = lazy_annotated_table(predictions, rows)
+        else:
+            if datamodel.sortColumn() >= 0 \
+                    or predmodel is not None and predmodel.sortColumn() > 0:
+                predictions = predictions[datamodel.mapToSourceRows(...)]
+            selected = predictions
+            annotated_data = predictions
```
VesnaT, line 869: `Annotated data` is missing the `Selected` attribute when no rows are selected.

Orange/widgets/evaluate/owpredictions.py
```diff
@@ -855,12 +859,17 @@ def _commit_predictions(self):
...
+        self.Outputs.selected_predictions.send(selected)
+        self.Outputs.annotated.send(annotated_data)
+
```
VesnaT, line 872: While at it, you can also remove the redundant new line :)

janezd:
@VesnaT. there was another bug: if the user sorted the table, the output was sorted only when nothing was selected.

I fixed all in a separate commit; please squash when merging.

PR #6723, opened 2024-02-02
Stack: new table instead of transform

### LINKED ISSUE #6665: Stacking Widget: "Stack failed with an error" when using Time Slice as Data source
**What's wrong?**
When using the Stacking widget with Time Slice as Data source and Test and Score widget on the other side, a "Stack failed with an error" message appears (on the Test and Score widget).
There is no error if Data Sampler (or no widget) is used to slice the data instead the Time Slice widget.

The log shows repeated errors like this one, not sure if this is THE one:
```
TransformDomain should define __eq__ and __hash__ to be used for compute_shared
ComputeValueProjector should define __eq__ and __hash__ to be used for compute_value
or set InheritEq = True if inherited methods suffice
```

Attached is a screenshot of the problem reproduces with "ParlaMint" dataset, in order to have a time field to Time Slice with I used the From and To variables as meta:
image

Attached is a zip of the .ows file also appears in the screenshot (where all relevant widget windows are open):
[Stacking with Time Slice problem report.ows.zip](https://github.com/biolab/orange3/files/13557015/Stacking.with.Time.Slice.problem.report.ows.zip)

**What's your environment?**
OS: macOS 12.7
Orange version: 3.36.2
Orange installed using the .dmg in the download page

ajdapretnar:
##### Issue

Fixes #6665.

##### Description of changes
Stacking transformed the data to the same data type as the input. However, Timeseries has an inbuilt assert for time variable, which was missing from stacked_data.

Tests will be added to Timeseries add-on.

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/ensembles/stack.py
+++ b/Orange/ensembles/stack.py
@@ -81,7 +81,7 @@ class StackedLearner(Learner):
         dom = Domain([ContinuousVariable('f{}'.format(i + 1))
                       for i in range(X.shape[1])],
                      data.domain.class_var)
-        stacked_data = data.transform(dom).copy()
+        stacked_data = Table.from_table(dom, data)
         with stacked_data.unlocked_reference():
             stacked_data.X = X
             stacked_data.Y = res.actual
--- a/Orange/tests/test_stack.py
+++ b/Orange/tests/test_stack.py
@@ -1,7 +1,7 @@
 import unittest
 
 from Orange.data import Table
-from Orange.ensembles.stack import StackedFitter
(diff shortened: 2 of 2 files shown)


ajdapretnar:
[note to self] Converted to draft. Add a test creating a dummy Table class, then mocking self.aggregate from stack.py.

ajdapretnar:
@janezd We went back and forth with Lan on this one. This was the simplest solution we could find.

PR #6729, opened 2024-02-09
[CI] Test with Python 3.12

PrimozGodec:
##### Description of changes
- Test Orange on Python 3.12, which has been out for some time now.
- Replace `distutils`, which was deprecated and removed in Python 3.12 form tests.
- Replace wrongly used called_with (should be assert_called_with) in test_owpredictions. Since it was not even a valid check (selection model change when the model is set to None in the widget), I replaced it with just assert_called_once, which checks that `unregister` was called.

##### Includes
- [ ] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: '3.10'
 
       - name: Install Tox
         run: |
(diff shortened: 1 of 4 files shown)


PrimozGodec:
I think the issue should be solved when Catboost provides Python 3.12 wheels. I asked them if they could provide them.

PrimozGodec:
Waiting for https://github.com/pavlin-policar/openTSNE/pull/255

.github/workflows/test.yml
```diff
@@ -147,7 +147,7 @@ jobs:
             name: PyQt6
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v3
```
ales-erjavec, line 150: Why this downgrade?

PrimozGodec: Thanks for checking. It probably remained the same from the previous version of PR. Fixed now.

PR #6734, opened 2024-02-16
[ENH] PLS: Move from Orange-spectroscopy

VesnaT:
##### Issue

Move PLS widget from Orange-spectroscopy.

##### Includes
- [X] Code changes
- [X] Tests
- [x] Documentation


borondics:
Thanks guys!!!

Orange/regression/pls.py
```diff
@@ -0,0 +1,190 @@
...
+            return ContinuousVariable(name,
+                                      compute_value=PLSProjector(transformer,
+                                                                 i))
+
+        n_components = self.skl_model.x_loadings_.shape[1]
+
+        var_names_X = self._get_var_names(n_components, self.var_prefix_X)
+        var_names_Y = self._get_var_names(n_components, self.var_prefix_Y)
+
+        domain = Domain(
```
markotoplak, line 107: Didn't we (with @lanzagar) speak about including the original data in this output too on Monday? Or did you change your mind?

Orange/regression/pls.py
```diff
@@ -0,0 +1,190 @@
...
+    def __call__(self, data):
+        if data.domain != self.pls_model.domain:
+            data = data.transform(self.pls_model.domain)
+        if len(data.Y.shape) == 1:
+            Y = data.Y.reshape(-1, 1)
+        else:
+            Y = data.Y
+        return self._transform_with_numpy_output(data.X, Y)
+
+    def __eq__(self, other):
```
markotoplak, line 49: Here you are offloading comparisons to `pls_model`, which makes sense, so it is cool.

But these `__eq__`  and `__hash__` need to be tested in a way that actually tests the comparison of two build domains (so, in turn, also `PLSProjector`). See `test_eq_hash` in `test_pca.py`.

markotoplak:
I particularly don't like how ellipsis is implemented because it is hardcoded into PLS and forces the first two components.

I saw that people wish for something similar also for PCA, and, also sometimes they are interested in other components, for example, 2nd vs 3rd. 

If I understand the idea behind it, a simple multivariate (ok, 2-variate) distribution is fitted on the data set, and then the ellipse of a certain percentage can be drawn. The code by @VesnaT only needs two columns of data, and I doubt it requires any assumptions only covered by PLS. So it should be general.

I see two ways of implementing it:
1. A separate widget that allows users to select two features and then adds the column.
2. Add these (general) ellipses to the Scatter Plot.

I would go for the second option. I was also asked quite a few times why we don't have this. Also, these ellipses fit into scatter plot visualization at least as much as regression lines (and even better than those that do not treat variables as independent).

markotoplak:
We discussed this and decided on adding the ellipse to the scatter plot widget.

PR #6737, opened 2024-02-16
[FIX] Fix failing tests on Python 3.11.8 on Windows

PrimozGodec:
##### Issue

Recently, some tests started to fail on Windows. I assume the issue is that Actions upgraded the Python 3.11 version to 3.11.8, released two weeks ago.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


PrimozGodec:
It seems that closing is not working in openpyxl as it worked in older versions of Python. I reported it: https://foss.heptapod.net/openpyxl/openpyxl/-/issues/2149
Will try to continue debugging next week.

markotoplak:
Here, you monkey patch ExcelReader. Does this code need to be in `Orange.__init__`? I know it will be removed, but I'd at least put this into a new file. Then, I'd import it with try catch.

If possible, I'd also rather import it where the reader is implemented (of course, it that does not work, forget it).

PR #6740, opened 2024-02-19
Context handler that respects position in Domain

### LINKED ISSUE #6721: DomainContextHandler does not handle feature "position" within the domain
For remembering states of combo boxes where we choose features we tend to use contexts. We frequently use `DomainContext`, which works well most of the time.

It works so well that I did not notice, until now, that it is not appropriate if combo boxes within the widget are limited to a different subset of (attributes, class_vars, metas).

Here is an example:

```python
from Orange.data import Table, Domain
from Orange.widgets.tests.base import WidgetTest
from Orange.widgets.widget import OWWidget
from Orange.widgets.gui import comboBox
from Orange.widgets.settings import \
    ContextSetting, DomainContextHandler, PerfectDomainContextHandler

from orangewidget.utils.signals import Input


class OWProblem(OWWidget):
    name = "Context Problem"

    v1 = ContextSetting(None)
    v2 = ContextSetting(None)
    settingsHandler = DomainContextHandler()

    class Inputs:
        data = Input("Data", Table, default=True)

    def __init__(self):
        super().__init__()

        m1 = DomainModel(DomainModel.SEPARATED,
                         valid_types=DomainModel.PRIMITIVE)
        comboBox(self.controlArea, self, "v1", model=m1, label="All")
        m2 = DomainModel(DomainModel.METAS | DomainModel.CLASSES,
                         valid_types=DomainModel.PRIMITIVE)
        comboBox(self.controlArea, self, "v2", model=m2, label="Metas + Classes")

        self.contextAboutToBeOpened.connect(lambda x: self.init_interface(x[0]))

    def init_interface(self, data):
        domain = data.domain if data is not None else None
        self.controls.v1.model().set_domain(domain)
        self.controls.v2.model().set_domain(domain)

    @Inputs.data
    def set_data(self, data):
        self.closeContext()
        self.openContext(data)


class TestOWProblem(WidgetTest):

    def setUp(self):
        self.widget = self.create_widget(OWProblem)

    def test_v2_invalid_after_reloading(self):
        iris = Table("iris")
        petals_in_metas = Domain(iris.domain.attributes[:2],
                                 iris.domain.class_var,
                                 metas=iris.domain.attributes[2:])
        self.send_signal(self.widget.Inputs.data,
                         iris.transform(petals_in_metas))
        self.widget.v1 = iris.domain["sepal width"]
        # v2 can have only metas and classes thus this will be invalid
        # context for original iris
        self.widget.v2 = iris.domain["petal width"]
        self.send_signal(self.widget.Inputs.data, iris)  # crashes
```

Running the test gives:
```
  File "/home/marko/dev/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owproblem.py", line 42, in set_data
    self.openContext(data)
  File "/home/marko/dev/orange-widget-base/orangewidget/widget.py", line 1350, in openContext
  File "/home/marko/dev/orange3/Orange/widgets/settings.py", line 125, in open_context
    super().open_context(widget, domain, *self.encode_domain(domain))
  File "/home/marko/dev/orange-widget-base/orangewidget/settings.py", line 833, in open_context
    self.settings_to_widget(widget, *args)
  File "/home/marko/dev/orange-widget-base/orangewidget/settings.py", line 946, in settings_to_widget
    _apply_setting(setting, instance, value)
  File "/home/marko/dev/orange-widget-base/orangewidget/settings.py", line 204, in _apply_setting
    setattr(instance, setting.name, value)
  File "/home/marko/dev/orange-widget-base/orangewidget/gui.py", line 194, in __setattr__
    callback(value)
  File "/home/marko/dev/orange-widget-base/orangewidget/gui.py", line 2320, in __call__
    self.action(*args)
  File "/home/marko/dev/orange-widget-base/orangewidget/gui.py", line 2407, in action
    raise ValueError("Combo box does not contain item " + repr(value))
ValueError: Combo box does not contain item ContinuousVariable(name='petal width', number_of_decimals=1)
```

Is this a bug in DomainContextHandler? No, it is how it was designed; it is just that I was using it wrongly. For this kind of situations, I'll need a new type of context. `PerfectDomainContextHandler` does not exhibit the same issue but is too restrictive. I'll need something in between.

It is a bug that contexts that match and could not be opened effectively (for any reason) block you from using the widget with that data? Probably. :) When this happens, the only solutions is to clear the settings.

markotoplak:
##### Issue

Resolves #6721

##### Description of changes

Added a new context handler that respects context settings' attributes `exclude_attributes`, `exclude_metas` and `exclude_class_vars`.  This context handler avoids storing the whole domain into settings and thus perfect matches do not work.

I searched for possible instances of #6721 in this repo and found only 3 cases where a limited `DomainModel` was used for a context settings, out of which only 1 was problematic:
1. owcorrelations, `feature_model`: not a problem, because it is applied to a processed data set where every possible feature is moved into `.attributes`.
2. owtestandscore, `feature_model` (for CV by Feature): not the same problem, because it uses `PerfectDomainHandler`.
3. owheatmap, `row_side_color_model`: finally, a bug. It this case, because of heavy customization, there were no crashes but UI was in an inconsistent state.

In Heat Map, a combo box for choosing row annotation color is limited to metas and class vars, and thus, when a selected feature was moved from metas to attributes, an incompatible context would match. That did not trigger any crashes because the combo box was updated through special functions, but it did leave the UI in an inconsistent state: an element with index of -1 was selected in the combo (""), and, the saved color was still shown in the map. Fixed by using the proposed ContextHandler.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/settings.py
+++ b/Orange/widgets/settings.py
@@ -138,7 +138,7 @@ class DomainContextHandler(ContextHandler):
             domain = domain.domain
         super().open_context(widget, domain, *self.encode_domain(domain))
 
-    def filter_value(self, setting, data, domain, *args):
+    def _filter_value(self, setting, data, *args):
         value = data.get(setting.name, None)
         if isinstance(value, list):
             new_value = [item for item in value
@@ -151,6 +151,9 @@ class DomainContextHandler(ContextHandler):
         elif self.is_encoded_var(value) \
                 and not self._var_exists(setting, value,  *args):
             del data[setting.name]
+
+    def filter_value(self, setting, data, domain, *args):
+        self._filter_value(setting, data, *args)
 
     @staticmethod
     def encode_variable(var):
@@ -219,9 +222,9 @@ class DomainContextHandler(ContextHandler):
     def match(self, context, domain, attrs, metas):
         if context.attributes == attrs and context.metas == metas:
             return self.PERFECT_MATCH
-        return self._match(context, domain, attrs, metas)
+        return self._match(context, attrs, metas)
 
-    def _match(self, context, domain, *args):
+    def _match(self, context, *args):
         matches = []
         try:
             for setting, data, _ in \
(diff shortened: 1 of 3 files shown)


markotoplak:
I tried adding dummy `.attributes` and `.metas` to the `Context`, hoping that new saved contexts would also work on older versions. Testing showed this does not work: (new) saved settings can get overridden by an (old) saved perfect match.  

So I can't make these compatible, at least not that simple.

PR #6752, opened 2024-03-04
[ENH] Scatter Plot: Add ellipse/s

VesnaT:
##### Issue

Implements #6734

##### Description of changes
Add ellipse (similar to regression line).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscatterplot.py
+++ b/Orange/widgets/visualize/owscatterplot.py
@@ -302,7 +302,9 @@ class OWScatterPlotGraph(OWScatterPlotBase):
     def _add_ellipse(self, x: np.ndarray, y: np.ndarray, color: QColor) -> np.ndarray:
         # https://stats.stackexchange.com/questions/577628/trying-to-understand-how-to-calculate-a-hotellings-t2-confidence-ellipse
         # https://stackoverflow.com/questions/66179256/how-to-check-if-a-point-is-in-an-ellipse-in-python
-        f = ss.f.ppf(0.95, 2, len(x) - 2)
+        n = len(x)
+        f = ss.f.ppf(0.95, 2, n - 2)
+        f = np.sqrt(f * 2 * (n - 1) / (n - 2))
         m = [np.pi * i / 100 for i in range(201)]
         cx = np.cos(m) * np.std(x) * f + np.mean(x)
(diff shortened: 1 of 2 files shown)


VesnaT:
https://github.com/biolab/orange3/assets/11465003/44f2d794-10f9-49fa-878c-4deb95f3daec

thocevar:
```
vals, vects = np.linalg.eig(cov)
...
cx = np.cos(m) * np.sqrt(vals[0] * f)
cy = np.sin(m) * np.sqrt(vals[1] * f)
```
Previously, the stretch of the ellipse was `cov[0, 0]` which is the variance along each coordinate axis while the stretch should reflect the variance in the direction of the computed eigenvectors. This results in more appropriate shapes. However, the regression line will not be completely in line with the ellipse, because it's minimizing vertical errors and not perpendicular projections.

janezd:
Lines can also minimize perpendicular distance if you set the plot to treat variables as independent.

thocevar:
By using the option "treat variables as independent" the regression line actually lines up with the ellipse. Do we want to prioritize this?

janezd:
By making it a default? Probably not because users expect a regression line. Yet, scatter plot shows relation between two variables, so perpendicular projection *should* be the default. We never found a good solution to make users aware that there are two possibilities and that they must decide what they want.

Orange/widgets/visualize/owscatterplot.py
```diff
@@ -274,8 +298,41 @@ def update_regression_line(self):
...
+        angle = math.atan2(vects[1, 0], vects[0, 0])
+        matrix = np.array([[np.cos(angle), -np.sin(angle)],
+                           [np.sin(angle), np.cos(angle)]])
+
+        n = len(x)
+        f = ss.f.ppf(0.95, 2, n - 2)
+        f = f * 2 * (n - 1) / (n - 2)
+        m = [np.pi * i / 100 for i in range(201)]
+        cx = np.cos(m) * np.sqrt(cov[0, 0] * f)
+        cy = np.sin(m) * np.sqrt(cov[1, 1] * f)
```
lanzagar, line 319: Use @thocevar 's fix.

lanzagar (commented):
After the fix everything looks ok.
As discussed we should maybe move this to the base class.

PR #6763, opened 2024-03-19
[FIX] OWSom: Fix crash for data without class variable

### LINKED ISSUE #6750: Bugs in SOM
SOM seems to have a few problems. See https://github.com/biolab/orange3/files/14397182/log.txt.

Anamari-ctrl:
##### Issue
Fixes #6750

##### Description of changes
Fixed TypeError: 'DiscreteVariable' object is not iterable.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/tests/test_owsom.py
+++ b/Orange/widgets/unsupervised/tests/test_owsom.py
@@ -723,6 +723,13 @@ class TestOWSOM(WidgetTest):
         restart_button.click()
         self.assertFalse(w.Information.modified.is_shown())
 
+    def test_make_domain_without_class_vars(self):
+        widget = self.widget
+        self.send_signal(self.widget.Inputs.data, self.iris)
+        widget.data.domain.class_vars = None
+        widget.update_output()
+        self.assertIsNotNone(self.get_output(widget.Outputs.annotated_data))
+
 
 class TestComputeValues(unittest.TestCase):
     def test_eq_hash(self):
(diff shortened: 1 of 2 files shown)


Orange/widgets/unsupervised/tests/test_owsom.py
```diff
@@ -723,6 +723,13 @@ def test_modified_info(self):
         restart_button.click()
         self.assertFalse(w.Information.modified.is_shown())
 
+    def test_make_domain_without_class_vars(self):
+        widget = self.widget
+        self.send_signal(self.widget.Inputs.data, self.iris)
+        widget.data.domain.class_vars = None
+        widget.update_output()
+        self.assertIsNotNone(self.get_output(widget.Outputs.annotated_data))
```
janezd, line 731: This doesn't test much. You may want to test two things. First, if you want to test that the widget no longer crashes, this line is unnecessary -- it would crash in the previous line (prior to your fix). But a better test would be that the widget constructs a correct domain. The proper test is

```python
    def test_make_domain_without_class_vars(self):
        widget = self.widget
        data = self.iris.transform(Domain(self.iris.domain.attributes))

        domain = self.get_output((widget.Outputs.annotated_data)).domain
        self.assertEqual(domain.attributes, data.domain.attributes)
        self.assertEqual(domain.class_var.name, ANNOTATED_DATA_FEATURE_NAME)
        self.assertEqual([var.name for var in domain.metas],
                         ["som_cell", "som_row", "som_col", "som_error"])
```

janezd:
You incorrectly copied the test that I suggested. Please fix the indent.

PR #6771, opened 2024-03-25
PLS: Add rotations to components output

VesnaT:
##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
I like it, if we need rotations too then this is the most direct way I can think of.

lanzagar:
After the coefficients output was expanded, I don't think this is needed.
If someone would like to see the rotations as well, we can reopen this and add them.

PR #6791, opened 2024-04-25
Orange Table-specific HDF5Reader

stuart-cls:
##### Issue
Enable saving/loading the Orange Table data structure from the binary HDF5 container.
Based on the implementation used in the dask branch, but with the dask parts removed.

Related: #6356

##### Description of changes
 - Add `HDF5Reader` with both `read` and `write_file` methods.
 - Add round-trip test for HDF5Reader
 - Add `h5py` to requirements
 - Use existing `.metadata` sidecar file to ~store~ load `table.attributes` if present.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io.py
+++ b/Orange/data/io.py
@@ -617,18 +617,22 @@ class HDF5Reader(FileFormat):
 
     @classmethod
     def write_table_metadata(cls, filename, data):
-        try:
-            dump_dict = {key: value if isinstance(value, str) else json.dumps(value)
-                        for key, value in data.attributes.items()}
-        except TypeError:
-            # some attribute is not JSON serializable, fall back to pickle .metadata file
-            super().write_table_metadata(filename, data)
-        else:
-            with h5py.File(filename, 'r+') as f:
-                metadata_group = f.require_group('metadata')
-                str_dtype = h5py.string_dtype()
-                for key, value in dump_dict.items():
-                    metadata_group.create_dataset(key, data=value, dtype=str_dtype)
+        dump_dict = {}
+        for key, value in data.attributes.items():
+             if isinstance(value, str):
+                 dump_dict[key] = value
+             else:
+                 try:
+                     dump_dict[key] = json.dumps(value)
+                 except TypeError:
+                     # value is not JSON serializable, fall back to pickle
+                    dump_dict[key] = pickle.dumps(value, protocol=PICKLE_PROTOCOL).hex()
+
+        with h5py.File(filename, 'r+') as f:
+            metadata_group = f.require_group('metadata')
+            str_dtype = h5py.string_dtype()
+            for key, value in dump_dict.items():
+                metadata_group.create_dataset(key, data=value, dtype=str_dtype)
 
     @classmethod
     def set_table_metadata(cls, filename, data):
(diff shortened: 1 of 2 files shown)


Orange/data/io.py
```diff
@@ -520,3 +524,95 @@ def _suggest_filename(self, content_disposition):
...
+            return new_var
+
+        def read_hdf5(name, as_str=False):
+            if name in f:
+                if as_str:
+                    return f[name].asstr()[:]
+                return f[name]
+            return None
+
+        assert 'domain' in f
```
stuart-cls, line 563: It might be better to put some attribute in the HDF5 to explicitly identify it as coming from Orange (could also specify which version) rather than just sniffing for "domain". This would improve discoverability, which is always an issue for HDF5 files.

markotoplak: Yes, absolutely. We should have an ID field and probably also a version identifier.

stuart-cls: I added some attributes to the root following the NeXus convention (creator, Orange_version, HDF5_Version, h5py_version) Weird capitalization of "HDF5_Version" is apparently intentional.

stuart-cls:
I couldn't find a satisfactory solution to the `.attributes` problem (for the use case I care about) so I re-used the `.metadata` sidecar files for now, which is better than nothing.

markotoplak:
Comments from @stuart-cls (from his email, just so that they do not get lost):

- Table.attributes -> I could try again to properly store this in the HDF5, instead of the .metadata sidecar. I got stuck trying to get the round-trip on a visible image to work :)
- Compatibility with dask branch: both opening files saved with that branch, and updating the branch to be compatible.

Orange/data/io.py
```diff
@@ -520,3 +524,101 @@ def _suggest_filename(self, content_disposition):
...
+
+        with h5py.File(filename, 'w') as f:
+            f.attrs['creator'] = "Orange"
+            f.attrs['Orange_version'] = ORANGE_VERSION
+            f.attrs['HDF5_Version'] = h5py.version.hdf5_version
+            f.attrs['h5py_version'] = h5py.version.version
+            for subdomain in ['attributes', 'class_vars', 'metas']:
+                parsed = [parse(feature) for feature in getattr(data.domain, subdomain)]
+                domain = np.array([[name, header] for name, header, _ in parsed], 'S')
+                domain_args = np.array([json.dumps(args) for *_, args in parsed], 'S')
```
markotoplak, line 611: For `domain_args` we should certainly use the variable length h5py string data type. Probably for `domain` too.

stuart-cls: Done.

stuart-cls:
Regarding the previous comments:
> Table.attributes -> I could try again to properly store this in the HDF5, instead of the .metadata sidecar. I got stuck trying to get the round-trip on a visible image to work :)

In general this format isn't doing anything clever with nested dictionaries (see domain_args for example). It would be a lot of work to map this to HDF5, and this is the same problem with `Table.attributes`.

> Compatibility with dask branch: both opening files saved with that branch, and updating the branch to be compatible.

I've tested both ways, it works fine. The new reader checks for "Orange" in the "creator" attribute, but falls back to checking that the "domain" group is there.

stuart-cls:
@markotoplak I added the possibility to store the `Table.attributes` dictionary in the file, with pickle `.metadata` file fallback. 
Otherwise, we have been using this quite happily for some time, it is night and day performance improvement for large datasets.

Is there anything I'm missing to get this merged in?

stuart-cls:
@markotoplak Changed the last use of `.metadata` sidecar to store the pickled values in the hdf5 file as we discussed.

Issue #5636, opened 2021-10-08
Feature constructor forgets everything when variable change in upstream widgets


clsandt:
The feature constructor widget deletes all the new variable definitions when you change variables in upstream widgets, it deletes even new variables that do not use the changed variables.

For example, I use an Integrate spectra widget to create variables that I then ratio against each other in the Feature constructor. If I modify or delete one of the integrated variables in teh upstream widget, all variables creatd in the Feature constructor disappear, even those that do not use the modified variable.

I uploaded a worflow that can be used to reproduce the bug. It uses data from the repository accessed with the Dataset widget.

I use Orange '3.29.3' in Quasar 1.2 on Windows X 
Quasar was installed from the installer.
[Bug report feature constructor workflow.zip](https://github.com/biolab/orange3/files/7311808/Bug.report.feature.constructor.workflow.zip)

Thanks

PR #6801, opened 2024-05-10
[FIX] Formula: Remove context settings

ales-erjavec:
##### Issue

Fixes: gh-5636

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


janezd:
I like the idea, it will solve a lot of problems.

I wrote two formulas, one with petal_length and one with petal_length and petal_width. Then I removed petal_width. All formulas persisted, the widget reported the error (via `transform_error`, I believe) and *did not update the output*, so it still output the data from the previous input. I think the widget should do one of the following:
- remove formulae that miss some variables from the model, but keep them in settings as "hints" so they reappear if another input data again contains them,
- show a warning and mark the formulae that cannot be computed (or report them in the warning description) due to missing variables or another problem, but compute the other formulae and update the output,
- report an error (as it does currently, just perhaps without outputting the exception in the console, if possible) and set output to `None`.

PR #6804, opened 2024-05-15
[FIX] Distances: Spearman actually computes Spearman instead of Pearson

VesnaT:
##### Issue

Metric definition for Spearman uses PearsonR instead of SpearmanR.

##### Description of changes
Replace PearsonR with SpearmanR.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/unsupervised/owdistances.py
```diff
@@ -58,7 +58,7 @@ class MetricDef(NamedTuple):
                   distance.PearsonRAbsolute),
         MetricDef(Spearman, "Spearman",
                   "Pearson correlation; distance = 1 - ρ/2",
-                  distance.PearsonR),
+                  distance.SpearmanR),
```
lanzagar, line 61: Let's fix the above string as well
`"Spearman correlation; distance = 1 - ρ/2"`

PR #6815, opened 2024-05-31
Remove Orange implementation of randomized PCA

pavlin-policar:
This PR basically reverts #3532.

At the time this was implemented, scikit-learn did not support PCA on sparse matrices without first densifying them. It has since added this support.

##### Issue
PCA failing on scikit-learn 1.5 due to PCA internal API changes.

##### Description of changes
The reason for our own custom PCA implementation (proper sparse support) is now invalid, since scikit-learn properly handles sprase data now. Therefore, I just removed our implementation of PCA and we now rely purely on scikit-learn. Our implentation was always meant to be temporary.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/pca.py
+++ b/Orange/projection/pca.py
@@ -1,4 +1,5 @@
 import numpy as np
+import scipy.sparse as sp
 from sklearn import decomposition as skl_decomposition
 
 import Orange.data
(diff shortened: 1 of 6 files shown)


pavlin-policar:
@markotoplak Was this the only thing blocking scikit-learn=1.5.0? If so, should I also change the dependency?

pavlin-policar:
Better sparse support was added to scikit-learn in 1.4.0 (https://scikit-learn.org/dev/auto_examples/release_highlights/plot_release_highlights_1_4_0.html#improved-memory-and-runtime-efficiency-for-pca-on-sparse-data), hence the new minimum version bump.

pavlin-policar:
I've sifted through some of the failing tests.

On macos, we now have the following curious error. scikit-learn now supports sparse data, but uses the arpack solver ([here](https://github.com/scikit-learn/scikit-learn/blob/5491dc695/sklearn/decomposition/_pca.py#L519-L520)), which requires that the number of components is less than the number of rows/columns in the data. This is now a problem, because in Orange, we allow the user to request as many principal components as there are columns, hence the failure.

Some windows and ubuntu failures seem unrelated to the changes here.

PR #6817, opened 2024-05-31
[ENH] ScoringSheet and ScoringSheetViewer widgets added

ZanMervic:
##### Issue
Added Scoring Sheet and Scoring Sheet Viewer widgets (transferred from the prototypes addon).

##### Description of changes
- Transferred two new widgets from the prototypes addon: the Scoring Sheet and Scoring Sheet Viewer.
- Included documentation and tests for both widgets.

##### Notice
One of the additions is the FasterRisk source code, copied from the following repository: https://github.com/jiachangliu/FasterRisk.

This inclusion serves as a temporary solution to address compatibility and functionality issues arising from the strict requirements of the original package. It will remain in place until the original maintainer updates the package to resolve these issues.

##### Includes
- [X] Code changes
- [X] Tests
- [X] Documentation


markotoplak:
@ZanMervic, could you check and fix the test errors? For starters, this learner needs to be usable without parameters (just make up some defaults) and needs to be imported in the correct module:

https://github.com/biolab/orange3/actions/runs/9328865142/job/25680578428?pr=6817

It should also be somehow obvious that `fasterrisk` is not another module with Orange learners. I would suggest making a `classification.utils` and put `fasterrisk` there.

markotoplak:
With one failing tests I noticed that this learner is not reproducible (the same learner can yields different results in multiple runs). I fixed one test, but this problem could also appear elsewhere.

To make this learner behave as other Orange learners, we need to make it reproducible. I tried looking at it, but could not find a quick way to do it. This probably stems from the fasterrisk library.

markotoplak:
Test coverage is relatively low and should be improved in the future.

PR #6821, opened 2024-06-07
[FIX] PCA: ensure tests pass on sklearn 1.4 and 1.5, which can return different results

pavlin-policar:
##### Issue
#6815 broke a few things that went under the radar, primarily due to scikit-learn 1.5.0 changes.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/projection/pca.py
+++ b/Orange/projection/pca.py
@@ -48,6 +48,13 @@ class PCA(SklProjector, _FeatureScorerMixin):
         if sp.issparse(X) and params["n_components"] == min(X.shape):
             X = X.toarray()
 
+        # In scikit-learn==1.4.0, only the arpack solver is supported for sparse
+        # data and `svd_solver="auto"` doesn't auto-resolve to this. This is
+        # fixed in scikit-learn 1.5.0, but for the time being, override these
+        # settings here
+        if sp.issparse(X) and params["svd_solver"] == "auto":
+            params["svd_solver"] = "arpack"
+
         proj = self.__wraps__(**params)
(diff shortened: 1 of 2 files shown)


pavlin-policar:
@markotoplak As far as I can tell, none of the remaining failures have anything to do with https://github.com/biolab/orange3/pull/6815. This should be good to merge.

pavlin-policar:
For posterity's sake: PCA can produce different results between scikit-learn versions 1.4.0 and 1.5.0 in that the directions of the vectors may be flipped. While they do ensure a consistent output within each version of scikit-learn using their `svd_flip` function [here](https://github.com/scikit-learn/scikit-learn/blob/5491dc695dbe2c9bec3452be5f3c409706ff7ee7/sklearn/decomposition/_pca.py#L646), the `u_based_decision` was changed to `False` in 1.5.0, while it was `True` (the default) in prior versions of scikit-learn. Here is the [commit](https://github.com/scikit-learn/scikit-learn/commit/311c6e2beff577afc84c4328dc5dfa00be24a238#diff-b17877cd9b0663deb819cce9f4cc84533c4ca88ca0ebd2380f9c8fc5864acf26R646) that makes this change.

This change, which causes inconsitencies between scikit-learn 1.4.0 and 1.5.0 was not discussed in their PR, but it was apparently necessary to make the outputs of PCA consistent between the different solvers in 1.5.0 (original PR and comment [here](https://github.com/scikit-learn/scikit-learn/pull/27491#issue-1917553935)).

PR #6848, opened 2024-07-05
Edit Domain: add custom format

### LINKED ISSUE #6841: Edit Domain interpret as time: add more dd-mm options
**What's your use case?**
image
Currently, manual options for conversion of text to date-time include some options with the month before the day and others with the day before the month. Some options with day before month that are popular in certain European countries are missing, e.g. 25-11-2025 00:00:00 and with "/" or "." instead of "-" as well as yy instead of yyyy are missing

Add the missing options to the drop-down menu

The alternative would be to apply string manipulations using Formula before Edit Domain

ajdapretnar:
##### Issue

Fixes #6841.

##### Description of changes
Add custom format option to Edit Domain.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/data/oweditdomain.py
```diff
@@ -1555,6 +1565,18 @@ def get_data(self):
             tr.insert(0, self.format_cb.currentData())
         return var, tr
 
+    def _on_custom_change(self):
+        self.custom_text = self.custom_edit.text()
+        date_pat = r"%(-?)d|%(b|B)|%(-?)m|%(y|Y)|%(-?)j|%(-?)U|%(-?)W|%(a|A)|%w"
+        time_pat = r"%(-?)H|%(-?)I|%p|%(-?)M|%(-?)S|%f"
```
janezd, line 1571: If we allow arbitrary format, we could also allow arbitrary strings, like: "This happened in Year: %Y, Month: %m, Day: %d". I'd use `self.have_date = any(f"%{x}" in self.custom_text for x in "dBmyYUWwaA"`.

Orange/widgets/data/oweditdomain.py
```diff
@@ -1555,6 +1565,18 @@ def get_data(self):
...
+        pattern1 = re.compile(date_pat)
+        pattern2 = re.compile(time_pat)
```
janezd, line 1573: Compiling if out of fashion. You can just use `re.search(pattern1, self.custom_text)` (if you keep re's at all). We use to compile them because otherwise `re.search` would compile them anew at each call. But Python now caches the compiled re, so compilation is no longer necessary.

Orange/widgets/data/oweditdomain.py
```diff
@@ -1555,6 +1565,18 @@ def get_data(self):
...
+        self.have_date = int(bool(pattern1.search(self.custom_text)))
+        self.have_time = int(bool(pattern2.search(self.custom_text)))
+        if self.format_cb.currentText() != "Custom format":
```
janezd, line 1576: Define a constant `CUSTOM_FORMAT = "Custom format"` and use it when adding the item to combo and here. Otherwise, one can change the combo label and forget to change it here (and elsewhere, as necessary).

Orange/widgets/data/oweditdomain.py
```diff
@@ -1555,6 +1565,18 @@ def get_data(self):
...
+            self.format_cb.setCurrentIndex(-1)
+        self.format_cb.emit()
```
janezd, line 1578: The proper call is `self.format_cb.currentIndexChanged.emit()`. But don't do this. It's better to call `self.variable_changed` directly, so the flow is more transparent. Also, `self.variable_changed` perhaps does something that shouldn't be done if the change is triggered by the user editing this line edit, in which case we'd split `self.variable_changed` in two functions and only call the second part here.)

Orange/widgets/data/oweditdomain.py
janezd, line 1576: Define a constant `CUSTOM_FORMAT = "Custom format"` and use it when adding the item to combo and here. Otherwise, one can change the combo label and forget to change it here (and elsewhere, as necessary).

(This also simplifies translation because you translate it only once and there's no chance to accidentally have two different translations.)

Orange/widgets/data/oweditdomain.py
janezd: By the way, I think that you can't/shouldn't emit this signal. `setCurrentIndex` will already emit this signal (see https://doc.qt.io/qt-6/qcombobox.html#currentIndex-prop). That is, the proper code would be

```
    self.variable_changed()
```

but this will call `self.variable_changed()` twice (once because of the change in combo and once because of the change here), so you should probably write

```
if self.format_cb.currentText() != CUSTOM_FORMAT:
    self.format_cb.setCurrentIndex(self.format_cb.count() - 1)
else:
    self.variable_changed()
```

Orange/widgets/data/oweditdomain.py
```diff
@@ -1555,6 +1565,18 @@ def get_data(self):
...
+    def _on_custom_change(self):
+        self.custom_text = self.custom_edit.text()
+        date_pat = r"%(-?)d|%(b|B)|%(-?)m|%(y|Y)|%(-?)j|%(-?)U|%(-?)W|%(a|A)|%w"
+        time_pat = r"%(-?)H|%(-?)I|%p|%(-?)M|%(-?)S|%f"
+        pattern1 = re.compile(date_pat)
+        pattern2 = re.compile(time_pat)
+        self.have_date = int(bool(pattern1.search(self.custom_text)))
+        self.have_time = int(bool(pattern2.search(self.custom_text)))
+        if self.format_cb.currentText() != "Custom format":
+            self.format_cb.setCurrentIndex(-1)
```
janezd, line 1577: Does `-1` work? I don't think Qt supports negative indices (does it); I'd write `format_cb.count() - 1`.

Orange/widgets/data/oweditdomain.py
```diff
@@ -1526,14 +1528,22 @@ class TimeVariableEditor(VariableEditor):
...
         form = self.layout().itemAt(0)
+        self.custom_text = ""
+        self.have_date = 0
+        self.have_time = 0
 
         self.format_cb = QComboBox()
         for item, data in [("Detect automatically", (None, 1, 1))] + list(
             Orange.data.TimeVariable.ADDITIONAL_FORMATS.items()
-        ):
+        ) + [("Custom format", (self.custom_text, self.have_time,
```
janezd, line 1538: `(self.custom_text, self.have_time, self.have_date)` will be copied into the tuple and won't be changed if the user changes them.

ajdapretnar:
I think this should work now. The lineEdit should handle also weird/random characters. I would test, but I cannot get the test to work. I've tried everything I could think of. See test for reference.

TODO:
- Add a working test/tests.
- Add report.
- Update documentation.
- Remove some (now unnecessary) formats.
- Add a translation.

Orange/widgets/data/oweditdomain.py
```diff
@@ -3080,4 +3132,5 @@ def column_str_repr_string(
 
 
 if __name__ == "__main__":  # pragma: no cover
-    WidgetPreview(OWEditDomain).run(Orange.data.Table("iris"))
+    WidgetPreview(OWEditDomain).run(Orange.data.Table(
+        "/Users/ajda/Desktop/test.csv"))
```
janezd, line 3136: Don't forget to remove this! I suggest you remove it now, and when you're making commits, add with `add -p` and skip this chunk.

janezd (commented):
Please also add a test with a format the includes hours, minutes or seconds (`have_time`), and perhaps some other formats, too.

Orange/widgets/data/tests/test_oweditdomain.py
```diff
@@ -340,6 +340,34 @@ def test_time_variable_preservation(self):
...
+            ["2024-032"],
+            ["2024-150"],
+            ["2024-365"]
+        ]
+        table = Table.from_list(Domain([], metas=[time_variable]), data)
+        self.send_signal(self.widget.Inputs.data, table)
+        model = self.widget.domain_view.model()
+        index = self.widget.variables_view.model().index
+        self.widget.variables_view.setCurrentIndex(index(0))
+
```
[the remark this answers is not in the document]
ajdapretnar: I know the variable didn't change (it stayed StringVariable). I said I tried everything, we even had a look at the code with Lan, but couldn't figure out how to write a proper call.

wvdvegte:
While this is still work in progress, could it perhaps be extended (if the option not included already) to cover #7078 add support for (conversion of) datetime stored as Unix format?

ajdapretnar:
@ales-erjavec Thanks for working on this! I've added a commit which removes a call to local test data.
One issue remains. I probably haven't foreseen it. I can successfully edit categorical variables (which are then converted to datetime), but I cannot do this for numeric variables. Example. I have a Year variable, which is recognized as numeric. When I convert it to datetime, the "Detect automatically" remains blocked, precluding me from providing custom format.

I have solved this locally in L1575, by removing the `Real` check. Not sure whether this breaks anything else, so I didn't push the changes.

I have one more issue, but it probably doesn't belong here. It is related to https://github.com/biolab/orange3/issues/6936. Even currently, using a format %Y doesn't work for years <1970. Even 1899. Any way to handle this?

ales-erjavec:
>... Example. I have a Year variable, which is recognized as numeric. When I convert it to datetime, the "Detect automatically" remains blocked, precluding me from providing custom format.
>
> I have solved this locally in L1575, by removing the Real check. Not sure whether this breaks anything else, so I didn't push the changes.

This would require much bigger change.

>I have one more issue, but it probably doesn't belong here. It is related to https://github.com/biolab/orange3/issues/6936. Even currently, using a format %Y doesn't work for years <1970. Even 1899. Any way to handle this?

For **\<1970**? This works for me.

ajdapretnar:
I think typically, years or miliseconds would be recognized as numeric variables. So having a custom format that applies only to categorical variables is marginally useful. :(
I tried converting years to categorical first, but the issue is that 1899 gets converted to 1899.0, so one has to manually remove the decimal points. Then I added the second Edit Domain and the year conversion worked. So the issue is really with converting numeric data.
I will test and merge this issue, but I think we should work on allowing custom format for numeric data, too. I will make a separate issue for this.

Issue #6861, opened 2024-07-24
Edit Domain: no way to get rid of warning "categories mapping for [variable] does not apply to current input" after change in upstream Formula


wvdvegte:
**What's wrong?**
When changing the mapping of categories of a categorical variable in Edit Domain, Edit Domain displays a warning  "categories mapping for [variable] does not apply to current input" once changes in an upstream Formula produce different or additional categories. Even when adapting the category mappings to the new inputs, the warning persists.

**How can we reproduce the problem?** 
[edit domain category mappings.ows.zip](https://github.com/user-attachments/files/16359530/edit.domain.category.mappings.ows.zip)
In the Formula in the attached workflow, change 'America' in the if statement to 'North America'. Edit Domain will show the warning, although in its dialog box the category name has already been updated. Even when explicitly defining an new mapping based on the new category names, the warning doesn't go away.

**What's your environment?**

- Operating system: Mac OS 14.5
- Orange version: 3.37
- How you installed Orange: from DMG followed by updates using the internal installer within Orange

janezd:
Thank you for another great bug report! If we offered bounties, you'd be rich by now. :)

PR #6865, opened 2024-08-02
[FIX] oweditdomain: Hide "categories mapping" warning on change

ales-erjavec:
##### Issue

Fixes: #6861

##### Description of changes

* Hide/show warning on change
* Indicate items with a warning with a yellow color

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/oweditdomain.py
+++ b/Orange/widgets/data/oweditdomain.py
@@ -1580,6 +1580,9 @@ TransformRole = Qt.UserRole + 42
 #: Any warnings applying to the transform (`list[tuple[Msg, str]]`)
 RestoreWarningRole = TransformRole + 1
 
+#: Hint key that was used to load stored settings.
+RestoreHintKey = RestoreWarningRole + 1
+
 
 class VariableEditDelegate(QStyledItemDelegate):
     ReinterpretNames = {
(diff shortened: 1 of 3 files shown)


ajdapretnar:
I apologise but I don't properly see the difference after the fix. I do like the highlight of the changed variable, although it would be better probably to use yellow as a background highlight (yellow letter are difficult to see on white). But the warning behaviour is still identical for me (or am I missing something)?

The user says <"A" if 0 else "B">. Then she changes this to <"C" if 0 else "B">. The warning says the mapping doesn't apply to the current input. But it *does* (i.e. should be updated in the background). The only thing that has changed was how I want to name my value ("C" instead of "A")? So in my opinion, there should be no warning. Even the widget Edit Domain show the mapping worked (it changes "A" to "C").

It is entirely possible I don't understand the issue here.

Orange/widgets/data/oweditdomain.py
```diff
@@ -1624,9 +1630,16 @@ def initStyleOption(self, option, index):
...
+        def set_color(palette: QPalette, color):
+            palette.setBrush(QPalette.Text, QBrush(color))
+            palette.setBrush(QPalette.HighlightedText, QBrush(color))
+
         if isinstance(multiplicity, int) and multiplicity > 1:
-            option.palette.setBrush(QPalette.Text, QBrush(Qt.red))
-            option.palette.setBrush(QPalette.HighlightedText, QBrush(Qt.red))
+            set_color(option.palette, Qt.red)
+        elif warnings_:
+            set_color(option.palette, Qt.yellow)
```
janezd, line 1642: Just a minor comment: yellow looks nice in dark mode, but is difficult to see on white background. Could you change it to a color that is visible in both modes, or choose different colors for different modes?

Screenshot 2024-09-20 at 13 57 03

wvdvegte:
I agree with @ajdapretnar (if I understand her correctly): it should be possible to change the name of a category on the input of Edit Domain, or even add new categories, and this should be updated in the background. Actually, in current Orange 3.37.0 it _is_ updated in the background - and everything works fine, even if new categories are added: the changes appear in the output of Edit Domain, as intended (I just checked). There just shouldn't be a warning!

PR #6878, opened 2024-08-21
DiscreteVariable: Raise an exception if values are not unique

### LINKED ISSUE #6877: Continuize: multiple attributes with the same name
**What's wrong?**

This issue is related to the issue with Discretization #6876.
If the input to the Continuize widget has attributes with multiple values with the same "name"/"value" (see Issue #6876 for a better explanation), the One-hot encoding will create multiple attributes with the same name which results in an exception.

Workflow I used (an extension of the workflow from issue #6876):
image

Exception:

```
  File "C:\Users\zanme\work\orange3\orange3\Orange\widgets\data\owcontinuize.py", line 458, in _on_radio_clicked
    self.commit.deferred()
  File "C:\Users\zanme\miniconda3\envs\orange3\Lib\site-packages\orangewidget\gui.py", line 2006, in conditional_commit
    do_commit()
  File "C:\Users\zanme\miniconda3\envs\orange3\Lib\site-packages\orangewidget\gui.py", line 2014, in do_commit
    commit.call()
  File "C:\Users\zanme\miniconda3\envs\orange3\Lib\site-packages\orangewidget\gui.py", line 1879, in call
    acting_func(instance)
  File "C:\Users\zanme\work\orange3\orange3\Orange\widgets\data\owcontinuize.py", line 517, in commit
    self.Outputs.data.send(self._prepare_output())
                           ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\zanme\work\orange3\orange3\Orange\widgets\data\owcontinuize.py", line 534, in _prepare_output
    return self.data.transform(Domain(attrs, class_vars, metas))
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\zanme\work\orange3\orange3\Orange\data\domain.py", line 154, in __init__
    raise Exception('All variables in the domain should have'
Exception: All variables in the domain should have unique names.
```

Screenshot of the raised exception and the two attributes with the same name:

image

**Note**

Because of this issue, a test was failing for the ScoringSheet widget. I have temporarily excluded the widget from the test, but it should be included again when the issue is resolved.

Test: ```Orange.tests.test_classification.LearnerAccessibility.test_all_models_work_after_unpickling_pca```

Zip of the workflow: [continuize_bug.zip](https://github.com/user-attachments/files/16691362/continuize_bug.zip)

To reproduce the problem, set the PCA components to 8 in the provided workflow.
image

**What's your environment?**

- Orange version: 3.38
- How you installed Orange: Using pip in a conda environment

janezd:
##### Issue

Closes #6877.

##### Description of changes

Check that all values are unique; raise an exception if not.

This does not solve any bugs, but reports them earlier.

##### Includes
- [X] Code changes
- [X] Tests


markotoplak:
This bug catcher is so good that it successfully reported 2 bugs on the first attempt.

janezd:
I'm so proud of my dilligent work. :)

However, as I unintentionally wrote "This does not solve any bugs, but reports them earlier.". :)

- Now I reimplemented the code that creates labels for discretized variables so that it increases the number of digits when necessary. When the number of digits is given as an argument (as opposed to using `ContinuousVariable.str_val`), it also rounds the interval threshold so that they match the printed labels.
- The other bug was due to an invalid test.

Orange/preprocess/discretize.py
```diff
@@ -49,38 +49,87 @@ def transform(self, c):
...
+            else:
+                if len(values) == len(set(values)):
+                    to_sql = BinSql(var, points)
+                    return points, values, to_sql
+
+        mindigits = max(ndigits or 0,
+                        int(-np.log10(mindiff)))
+        maxdigits = np.finfo(npoints.dtype).precision + 2
+        for digits in range(mindigits, maxdigits + 1):
+            npoints = np.round(points, digits)
```
markotoplak, line 111: This assumes that np.round and floating point formatter do the same thing. They probably do, but why did you not rather make strings directly and count these?

Orange/preprocess/discretize.py
```diff
@@ -49,38 +49,87 @@ def transform(self, c):
...
+            mindiff = np.min(npoints[1:] - npoints[:-1])
+            if mindiff == 0:
+                raise ValueError("Some interval thresholds are identical")
+        else:
+            mindiff = 1  # prevent warnings
 
+        if ndigits is None or len(points) == 1:
+            try:
+                values = cls._get_labels(var.str_val, points)
+            except ValueError:
```
markotoplak, line 100: A comment that this exception happens when upper and lower threshold are formatted the same would be nice.

Orange/preprocess/discretize.py
markotoplak: Actually, see https://numpy.org/doc/2.0/reference/generated/numpy.round.html, they are not the same. :)

markotoplak:
Ok, I found some nitpicky issues. I can fix it myself if you agree with my suggestions.

markotoplak:
I replaced `np.round` with `round`. The results are a bit different, but no implementation was wrong (see tests).

I could not make a test case up where the old implementation would have actual problems, for example, `points` that would not correspond exactly to `limits,` but I can not guarantee one does not exist either. So, this seems safer, but I can not really say that the previous implementation was wrong.

markotoplak:
The difference between Python's and numpy's rounding.

```
n = 2.3455; npn = np.float_(n)
round(n, 3)
Out[17]: 2.345
round(npn, 3)  # this one is fun
Out[18]: 2.346
print("%0.3f" % n)
2.345
print("%0.3f" % npn)  # :)
2.345
```

PR #6882, opened 2024-08-23
Boxplot: Add axis label

### LINKED ISSUE #6863: Box Plot: add attribute name to the plot
**What's your use case?**
Trying to export an image in Box Plot, there is no attribute name reported. E.g. I see no "sepal length" anywhere in the plot, leading me to guess what was the variable used.
iris

Add variable name at the top of the plot. Hopefully, the category names reveal enough about the subgroup used. If not, we would have to think about this, too.
p.s. All other visualization widget include variable name.

Use Inkscape to add it, but this is a bit silly.

janezd:
##### Issue

Fixes #6863, but I'm not sure I like it.

##### Description of changes

Add a label to the axis.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owboxplot.py
+++ b/Orange/widgets/visualize/owboxplot.py
@@ -972,9 +972,10 @@ class OWBoxPlot(widget.OWWidget):
         if self.stat_test:
             label = QGraphicsSimpleTextItem(self.stat_test)
             brect = self.box_scene.sceneRect()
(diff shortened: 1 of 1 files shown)


janezd:
Would this be OK?

Screenshot 2024-09-12 at 15 51 37

markotoplak:
It looks wonderful.

PR #6895, opened 2024-09-20
test_pandas: Remove a failing test caused by pytz

janezd:
##### Issue

pytz 2024.2 seems to have changed treatment of DST and of timezones, which broke our tests.

##### Description of changes

The failing assertion could be fixed, but we prefer removing it: it was a test of pytz regressions rather than of our code.


markotoplak:
Today i learned that time zones are more complicated than I thought.

PR #6903, opened 2024-09-28
CI: Install pymssql separately

janezd:
An attempt at fixing failing tests on macOS with Python >= 3.11.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -158,6 +158,15 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install --upgrade tox
 
+      - name: Install pymssql on macOS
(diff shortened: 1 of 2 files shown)


janezd:
I give up. I guess we should abandon tests on macOS that requires pymssql until pymssql installation is fixed. I hope we distribute a working version of pymssql in released Orange?

markotoplak:
I think (did not check) we don't even distribute pymssql. And if we do, we have tests there anyway that would catch if something was wrong.

janezd:
> we have tests there

Where?

markotoplak:
> > we have tests there
> 
> Where?

When we create packages, we run all the tests on the installed package.

PR #6906, opened 2024-10-03
GroupBy: Avoid guessing variable types

janezd:
##### Issue

The Group By widget will fail on Pandas 3.0.

Group by currently computes aggregations, puts them into a pandas data frame and then guesses data types - at least in part. In particular, it tries to interpret columns as time and creates a `TimeVariable` if conversion succeeds. In tests, we have a concatenation of numeric data that results in value "1.0 1.0". Pandas 3.0 now converts this to January 1st 1970 (I think), so the column becomes a time variable and appears among primitive attributes, not metas(!).

##### Description of changes

I think a proper solution is to define how to construct variables with aggregations. It can be
- a fixed type (e.g. `StringVariable` for concatenation, `ContinuousVariable` for Span...)
- a copy of the existing variable (e.g. for Mode, Min value, Random value...)
- a variable of the same type as the existing variable (Mean...)

The difference between the latter two cases is that it resets the number of decimals.

I modified one test. Apparently it assumes that Count and Count Defined for string variables will be metas, but as far as I see, they are attributes. This PR keeps the behaviour, so I don't understand why tests were failing before.

##### Includes
- [X] Code changes
- [X] Tests


janezd:
@VesnaT, can you review this?

markotoplak:
@janezd, now that I merged the other PR, please rebase this one.

VesnaT (approved):

PR #6910, opened 2024-10-09
Adaboost: Remove deprecated argument 'algorithm'

janezd:
##### Issue

In Sklearn 1.6, the only acceptable value for `AdaBoost's argument `algorithm` is `SAMME`. In 1.8, the argument will be removed. (https://github.com/scikit-learn/scikit-learn/pull/29997)

##### Description of changes

- I removed the argument. This doesn't break any tests, so it apparently didn't have any effect (or we didn't test it properly). Any potential changes are however unavoidable because the current algorithm is removed from SkLearn (for the reasons, see https://github.com/scikit-learn/scikit-learn/issues/26784).
- I removed the setting from the widget. I haven't provided any migration, not even showing a warning because the user would have no way to disable the warning. Besides, it probably doesn't matter.
- Some tests depend on random seed. I change one of them to `4` so that results in scikit-learn=1.6.dev are the same as in earlier version(s).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/tests/test_ada_boost.py
+++ b/Orange/tests/test_ada_boost.py
@@ -33,7 +33,7 @@ class TestSklAdaBoostLearner(unittest.TestCase):
         self.assertLess(ca, 0.99)
 
     def test_adaboost_estimator(self):
-        np.random.seed(0)
+        np.random.seed(4)
         stump_estimator = SklTreeLearner(max_depth=1)
         tree_estimator = SklTreeLearner()
         stump = SklAdaBoostClassificationLearner(
--- a/Orange/widgets/model/owadaboost.py
+++ b/Orange/widgets/model/owadaboost.py
@@ -1,4 +1,5 @@
 from AnyQt.QtCore import Qt
+from AnyQt.QtWidgets import QFormLayout, QLabel
 
 from Orange.base import Learner
 from Orange.data import Table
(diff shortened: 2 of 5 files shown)


Orange/tests/test_ada_boost.py
```diff
@@ -33,7 +33,7 @@ def test_adaboost(self):
         self.assertLess(ca, 0.99)
 
     def test_adaboost_estimator(self):
-        np.random.seed(0)
+        np.random.seed(4)
```
janezd, line 36: If random seed is `0`, the result today is not the same as it was the last week. I discovered the `4` seems OK.

Scikit and pandas are moving targets. :(

Orange/widgets/model/owadaboost.py
```diff
@@ -46,36 +44,36 @@ class Error(OWBaseLearner.Error):
 
     def add_main_layout(self):
         # this is part of init, pylint: disable=attribute-defined-outside-init
-        box = gui.widgetBox(self.controlArea, "Parameters")
+        grid = QFormLayout()
+        box = gui.widgetBox(self.controlArea, box=True, orientation=grid)
```
VesnaT, line 48: `box` variable is unused.

VesnaT (commented):
With the a new version of scikit-learn a default value of unsettable (within Orange) parameter will change (`algorithm`).

Should we set the `algorithm` parameter to `SAMME` to assure the same behaviour as when the scikit 1.6 is released?

janezd:
@VesnaT I put the `algorithm` argument back, with the default value of `SAMME`. As I understand, our wrapper for scikit learners will skip it when scikit-learn will no longer accept it (in version 1.8).

In the meantime, we're back in red because somebody decided it would be a terribly good idea to change the behaviour of AUC (https://github.com/scikit-learn/scikit-learn/pull/27412). But this is something for the next PR... 

Why do we depend on unstable packages like pandas, sklearn and scipy? No better alternatives? Shame, really.

Sorry, I'm pissed. I thought I'd be doing something useful today, not hunting another change in sklearn.

PR #6914, opened 2024-10-15
Test and Score: Adapt test to change in sklearn

### LINKED ISSUE #6913: Broken test due to a change in scikit-learn's AUC
Hard-working maintainers of Python scientific stack decided to change the result of AUC when only one class is present in test data: https://github.com/scikit-learn/scikit-learn/pull/27412, merged two days ago (here: https://github.com/scikit-learn/scikit-learn/pull/27412/files#diff-4eb3c023f8a3f088d62208f6adbd02b6df5196de2257ccd228dffc972c964634R383).

Yes, changing `None` to `0` indeed does not make a huge difference, does it? Well, except for breaking our tests and taking me half an hour to discover who's responsible (scikit? sklearn? pandas?) and what changed and why. Now I only need to make our tests version-dependent and we're fine again. Thank you, thank you, thank you.

I do not have a habit of swearing, but I might start if I keep trying to keep the pace with the ever-changing scipy, sklearn and pandas.

janezd:
##### Issue

Closes #6913.

Sklearn did it again.

##### Description of changes

Expect different result from sklearn >= 1.6.

It seems that we do not have to change anything else but this test? Behaviour of the Test and score widget seems to be the same as before.

##### Includes
- [ ] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/tests/test_owtestandscore.py
+++ b/Orange/widgets/evaluate/tests/test_owtestandscore.py
@@ -2,8 +2,10 @@
 # pylint: disable=protected-access
 import unittest
 from unittest.mock import Mock, patch
+from packaging.version import parse as parse_version
 
 import numpy as np
+from sklearn import __version__ as sklearn_version
 from AnyQt.QtCore import Qt
 from AnyQt.QtTest import QTest
 from AnyQt.QtWidgets import QApplication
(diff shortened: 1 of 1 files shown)


markotoplak:
This should wait for scikit-learn/scikit-learn/pull/30103.

markotoplak:
The comment in tests reminded me of someone. :)

PR #6917, opened 2024-10-16
[FIX] Calibrated Learner: Prevent in place modification of base learner

### LINKED ISSUE #6915: Calibrated learner changes model names the wrong way
1. **Calibrated Learner** adds the calibration names to the model name in **Test and Score** but turning them off does not remove them. So changing calibration and optimization on and off keeps expanding the model names.

2. **Calibrated Learner** extends the name of the calibrated model (that goes through it) and the uncalibrated too (does not go through it). The original learner is separate and not calibrated so its name should not include any of that.

calibrated learner

Orange: 3.37.0
OS: Windows 10 x64

ales-erjavec:
##### Issue

Fixes #6915

##### Description of changes

* Never return `base_learner` instance from `create_learner`
* Replace IdentityWrapper (which did not work) with copy.deepcopy of the instance.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/model/owcalibratedlearner.py
+++ b/Orange/widgets/model/owcalibratedlearner.py
@@ -1,3 +1,5 @@
+import copy
+
 from Orange.classification import CalibratedLearner, ThresholdLearner, \
     NaiveBayesLearner
 from Orange.data import Table
@@ -65,7 +67,6 @@ class OWCalibratedLearner(OWBaseLearner):
         self.learner = self.model = None
 
     def _set_default_name(self):
-
         if self.base_learner is None:
             self.set_default_learner_name("")
         else:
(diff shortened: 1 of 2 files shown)


Orange/widgets/model/owcalibratedlearner.py
```diff
@@ -95,8 +92,11 @@ def fit_storage(self, data):
                                        self.ThresholdMap[self.threshold])
         if self.preprocessors:
             if learner is self.base_learner:
-                learner = IdentityWrapper()
+                learner = copy.deepcopy(learner)
             learner.preprocessors = (self.preprocessors, )
+        if learner is self.base_learner:
+            learner = copy.deepcopy(learner)
+        assert learner is not self.base_learner
```
janezd, line 99: Am i missing something or could this be

```python
        if learner is self.base_learner:
            learner = copy.deepcopy(learner)
        if self.preprocessors:
            learner.preprocessors = (self.preprocessors, )
        assert learner is not self.base_learner
```

PR #6920, opened 2024-10-18
[ENH] Datasets: Add domain field; respect "Unlisted"

janezd:
##### Issue

- When reimplementing the repository of data sets, we added a field "Domain". Its current values are "core" and "sc", but the plan was to add (a very limited) number of others, such as "Education", and to allow the user to select a domain in the Data sets widget. Domains were not meant to be like tags but rather to allow the user select a single specific domain.

- With the new repository, the Datasets widget started showing single-cell data sets that were previously (if I remember correctly) accessible only in the single cell's widget. This PR reverts the behaviour to the more sensible original: by default, it only shows "core" data sets.

- With @BlazZupan we discussed that it would be handy if a data set could be unlisted - shown only if the user types the first five characters of the dataset's name into the filter. This would not be used to prevent the user from seeing the data set, but just to avoid polluting the list with data sets used at occasional workshops. The functionality has been implemented and deployed in the data sets editor (https://github.com/biolab/orada/commit/ccd980e5f05131a80ac33031d6d87009d80a2b21).

##### Description of changes

- Add a combo for selection of domains.
- Show the "core" domain as "(general)". It is the default and does not include sc data sets.
- Don't show data sets whose visibility is set as "unlisted" unless the user types the first five character into the filter.

To test the latter, try typing `luxem`.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -35,6 +35,12 @@ from Orange.widgets.widget import OWWidget, Msg
 
 log = logging.getLogger(__name__)
 
+# These two constants are used in settings (and in the proxy filter model).
+# The corresponding options in the combo box are translatable, therefore
+# the settings must be stored in language-independent form.
+GENERAL_DOMAIN = None
+ALL_DOMAINS = ""  # The setting is Optional[str], so don't use other types here
+
 
 def ensure_local(index_url, file_path, local_cache_path,
                  force=False, progress_advance=None):
@@ -124,7 +130,7 @@ class Namespace(SimpleNamespace):
         self.seealso = []
         self.tags = []
         self.language = "English"
-        self.domain = "core"
+        self.domain = None
         self.publication_status = self.PUBLISHED
 
         super(Namespace, self).__init__(**kwargs)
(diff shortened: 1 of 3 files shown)


borondics:
Would it be possible to make it so that if there is a specific addon installed which has datasetes available on the servers then this domain is also added automatically? Basically the addon would register the domain in the datasets widget...?

janezd:
This is not "domain" in the sense of internet domain. It's more like a field, a category ... The genesis of "domains" is that we have too many data sets that we need for teaching, so we want to separate them from other data sets. We also have data sets for the single-cell add-on, which are probably not useful for the majority of users.

The widget still shows only data sets served from a single server. If you have some data sets to add - from a specific domain -, we can upload them.

But your question is probably rather about adding other servers? We could let the user (or an add-on) add servers, but servers need to provide a json file in a particular not-so-well-documented format. I doubt many would set it up; if only the "official" add-ons would do so, it's simpler to put their corresponding data on the existing server. I think.

borondics:
I think I understood "domain" the way you originally meant it for the new addition.

We have some orange-spectroscopy datasets on the biolab server and they might be excluded with this new PR so I was wondering if we could enable the orange-spectroscopy addon (or any other) to include those files ("domain") if it is installed.

However, adding new servers sounds interesting too...

markotoplak:
We are going to move datasets between domains. Thus, it is critical that, as we just discussed, output the saved selected data set even if it was moved. And yes, we are both aware that this is going to be cumbersome.

Orange/widgets/data/owdatasets.py
```diff
@@ -423,6 +486,16 @@ def _on_language_changed(self):
             self.language = combo.currentText()
         self.view.model().setLanguage(self.language)
 
+    def _on_domain_changed(self):
+        combo = self.domain_combo
+        if combo.currentIndex() == combo.count() - 1:
```
markotoplak, line 491: All domains does not always exist. (just making a note)

markotoplak: I would just change the order... :)

PR #6921, opened 2024-10-21
[ENH] Parameter Fitter: Basic implementation

VesnaT:
##### Issue

Implement a Parameter Fitter widget:
- The initial version can only fit integer parameters.
- to fit a parameter, the Learner has to implement a `fitted_parameters` method.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


Orange/base.py
```diff
@@ -87,6 +88,10 @@ class Learner(ReprableWithPreprocessors):
     #: A sequence of data preprocessors to apply on data prior to
     #: fitting the model
     preprocessors = ()
+    FittedParameter = namedtuple(
```
janezd, line 91: Is there a reason for using `namedtuple` and not `NamedTuple`? (I prefer the latter, also because of type annotation.)

Orange/base.py
```diff
@@ -87,6 +88,10 @@ class Learner(ReprableWithPreprocessors):
...
+        "FittedParameter",
+        ["parameter_name", "label", "tick_label", "type", "min", "max"]
```
janezd, line 93: `parameter_name` -> `name`?

Orange/base.py
```diff
@@ -1,9 +1,10 @@
 import inspect
 import itertools
+from collections import namedtuple
 from collections.abc import Iterable
 import re
 import warnings
-from typing import Callable, Dict, Optional
+from typing import Callable, Dict, Optional, List
```
janezd, line 7: You don't need this. Python >3.9 supports, for instance, `list[int]`.

Orange/base.py
```diff
@@ -179,6 +184,9 @@ def active_preprocessors(self):
                 self.preprocessors is not type(self).preprocessors):
             yield from type(self).preprocessors
 
+    def fitted_parameters(self, *args, **kwargs) -> List:
```
janezd, line 187: `List` -> `list[FittedParameter]` (we can start using lowercase list from since we've dropped support for Python 3.8).

Orange/classification/random_forest.py
```diff
@@ -1,6 +1,8 @@
+from typing import List
```
janezd, line 1: You can use `list` (see above).

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -0,0 +1,509 @@
+from typing import Optional, Tuple, Callable, List, Dict, Iterable, Sized
```
janezd, line 1: See above.

janezd (commented):
I haven't really reviewed the PR - I haven't even checked it out from the repo. Here are just two things I incidentally noticed. Feel free to ignore.

Orange/base.py
```diff
@@ -88,9 +88,15 @@ class Learner(ReprableWithPreprocessors):
...
-        ["parameter_name", "label", "tick_label", "type", "min", "max"]
+    FittedParameter = NamedTuple(
+        "FittedParameter", [
+            ("name", str),
+            ("label", str),
+            ("tick_label", str),
+            ("type", Type),
+            ("min", Union[int, NoneType]),
+            ("max", Union[int, NoneType]),
+        ]
```
janezd, line 99: What I had in mind was

```python
class FittedParameter(NamedTuple):
    name: str
    label: str
    tick_label: str = ""
    type_: type = int
    min: Optional[int] = None
    max: Optional[int] = None
```

No need to change it; I find this easier on the eye, but it's a matter of taste.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -54,8 +55,8 @@ def _search(
         data: Table,
         learner: Learner,
         fitted_parameter_props: Learner.FittedParameter,
-        initial_parameters: dict,
-        steps: Sized,
+        initial_parameters: dict[str, Any],
```
janezd, line 58: Currently, we'd have `dict[str, int]`, but eventually we'd have `dict[str, Union[int, float]]`. On the other hand, what if the learner provides a list of string arguments? So I put `dict[str, Any]`. We can put a humble `int` there, though, for the time being.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -68,15 +69,15 @@ def _search(
         params[name] = value
         result = _validate(data, type(learner)(**params), scorer)
         scores.append((value, result))
-    return scores, scorer.name, fitted_parameter_props.tick_label
+    return scores, scorer.name, fitted_parameter_props.label
```
janezd, line 72: I think it's nicer to write a full label as the axis label, and put just number in the ticks.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -145,9 +146,9 @@ class FitterPlot(PlotWidget):
 
     def __init__(self):
         super().__init__(enableMenu=False)
-        self.__bar_item_tr: pg.BarGraphItem = None
-        self.__bar_item_cv: pg.BarGraphItem = None
-        self.__data: list[ScoreType] = None
+        self.__bar_item_tr: Optional[pg.BarGraphItem] = None
```
janezd, line 149: I swear I'm not nitpicking, it's just because PyCharm marks this `None` as invalid. :)

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -223,15 +224,15 @@ def help_event(self, ev: QGraphicsSceneHelpEvent) -> bool:
         text = ""
         if index is not None:
             _, scores = self.__data[index]
-            text = f"<table align=left>" \
-                   f"<tr>" \
-                   f"<td><b>Train:</b></td>" \
+            text = "<table align=left>" \
```
janezd, line 227: I didn't see the lint output, but it usually complains about unnecessary f-strings.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -252,6 +253,54 @@ def __get_index_at(self, point: QPointF) -> Optional[int]:
         return None
 
 
+class RangePreview(QWidget):
```
janezd, line 256: This is for eliding in the middle of the list, so that preview is shown as, e.g. `10, 20, 30, 40, 50, 60, 70, ..., 120`, not `10, 20, 30, 40, 50, 60, 70, 80, ...`. It also simplifies indenting and aligning.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -270,16 +319,18 @@ class Inputs:
...
     parameter_index = Setting(DEFAULT_PARAMETER_INDEX, schema_only=True)
     FROM_RANGE, MANUAL = range(2)
-    type = Setting(FROM_RANGE)
-    minimum = Setting(DEFAULT_MINIMUM, schema_only=True)
-    maximum = Setting(DEFAULT_MAXIMUM, schema_only=True)
-    manual_steps = Setting("", schema_only=True)
+    type: int = Setting(FROM_RANGE)
+    minimum: int = Setting(DEFAULT_MINIMUM, schema_only=True)
+    maximum: int = Setting(DEFAULT_MAXIMUM, schema_only=True)
+    manual_steps: str = Setting("", schema_only=True)
```
janezd, line 325: Again, I'm not nitpicking. Just want to help PyCharm (and other tools).

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -318,93 +368,150 @@ def _add_plot(self):
...
+        buttons = gui.radioButtons(None, self, "type",
+                                   callback=self.__on_type_changed)
+        button = gui.appendRadioButton(buttons, "Range:")
+        layout.addWidget(button, 1, 0)
+
+        kw = dict(minv=-MIN_MAX_SPIN, maxv=MIN_MAX_SPIN,
+                  alignment=Qt.AlignRight,
+                  callback=self.__on_min_max_changed)
+        box = gui.hBox(None)
+        self.__spin_min = gui.spin(box, self, "minimum", label="From:", **kw)
```
janezd, line 387: This label is slightly to high. I'll take care of this if we decided to keep this layout.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -318,93 +368,150 @@ def _add_plot(self):
...
+        edit = gui.lineEdit(None, self, "manual_steps",
+                            placeholderText="10, 20, 30",
+                            callback=self.__on_manual_changed)
+        layout.addWidget(edit, 4, 1)
+
+        # gui.lineEdit's connect does not call the callback on return pressed
+        # if the line hasn't changed.
+        @edit.returnPressed.connect
+        def _():
+            self.type = self.MANUAL
```
janezd, line 408: The thing with this line edit is: if the user changes it and presses return, we want a recomputation. The `callback` takes care of this.

Now imagine that the user enters some data into the line edit. Then he clicks the first radion (`FROM_RANGE`). Now he clicks on the line edit. Nothing happens (as it shouldn't). Then he presses return (without changing anything). The callback is not triggered because nothing is change. But we still want to change the type to manual. Hence this additional connect.

We need to check, though, that recomputation it is not triggered twice.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -318,93 +368,150 @@ def _add_plot(self):
...
+        self._settings_changed()
+
+    def _settings_changed(self):
         self._update_preview()
         self.commit.deferred()
 
     @property
     def fitted_parameters(self) -> list:
-        if not self._learner or not self._data:
+        if not self._learner:
```
janezd, line 435: I refactored this because `self._data` is not needed if `self._learner` is not an instance of fitting. If this function can work without data, let it. The computation should be (and can be and is) stopped elsewhere.

The reason I like this better is because we may want to output a `Learner` from this widget. (That is, a Learner that has a base learner and performs fitting of arguments internally.) In this case, the widget will also have an active output without data.

Revert if you disagree. :)

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -318,93 +368,150 @@ def _add_plot(self):
...
+            if not self._data:
+                return []
+            return self._learner.fitted_parameters(self._data)
+        else:
+            return self._learner.fitted_parameters()
 
     @property
     def initial_parameters(self) -> dict:
-        if not self._learner or not self._data:
+        if not self._learner:
```
janezd, line 446: Same as above.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -318,93 +368,150 @@ def _add_plot(self):
...
-                return []
+            return self._steps_from_manual()
+
+    def _steps_from_range(self) -> tuple[int]:
+        if self.maximum < self.minimum:
+            self.Error.min_max_error()
+            return ()
+
+        if self.minimum == self.maximum:
+            return (self.minimum, )
```
janezd, line 469: I think this is reasonable, but we can remove it and extend the above condition from `<` to `<=`.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -318,93 +368,150 @@ def _add_plot(self):
...
+
+        diff = self.maximum - self.minimum
+        # This should give between 10 and 15 steps
```
janezd, line 472: Well, 16. :)

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -318,93 +368,150 @@ def _add_plot(self):
...
+                *range((self.minimum // step + 1) * step, self.maximum, step),
+                self.maximum)
+
+    def _steps_from_manual(self) -> tuple[int]:
+        steps = self.manual_steps
+        if not steps:
+            return ()
+        try:
+            steps = tuple(sorted(map(int, steps.split(","))))
+            self.manual_steps = ", ".join(map(str, steps))
```
janezd, line 485: We can discuss this. I don't see a reason, though, why anybody would want unsorted numbers. It is more likely that one will want to insert a number, so this feature is more useful than annoying.

janezd:
I focused on the gui layout and changed a few minor things I encountered.

In the future, we probably want to support:
- optimization of multiple arguments,
- non-int arguments (at least floats),
- outputting a learner with internal fitting.

So, I thought about making some other changes, but decided not to, because they would probably be changed again in that future.

Orange/widgets/evaluate/owparameterfitter.py
VesnaT: Fine. 

But even more useful I find checking that `manual_steps` are within `fitted_parameter.min` and `fitted_parameter.max`.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -252,6 +253,54 @@ def __get_index_at(self, point: QPointF) -> Optional[int]:
...
+        painter = QPainter(self)
+        metrics = QFontMetrics(self.font())
+        style = self.style()
+        rect = self.rect()
+
+        # Indent by the width of the radio button indicator
+        rect.adjust(style.pixelMetric(style.PM_IndicatorWidth)
+                    + style.pixelMetric(style.PM_CheckBoxLabelSpacing), 0, 0, 0)
+
+        last_text = f", {self.__steps[-1]}"
```
VesnaT, line 288: image
I'd remove the `,` for a single value.

VesnaT (commented):
Two more things I'd suggest changing:

- Steps preview does not refresh when key down/up in spin box

- If I select `Manual` and there is no range in the line edit, the behaviour is incorrect.
image

Orange/widgets/evaluate/owparameterfitter.py
janezd: I forgot about this. But now that you mentioned: what about letting the user go out of range, and just show a warning? Just an idea.

janezd:
I think I've addressed all comments except for 

> Steps preview does not refresh when key down/up in spin box

This looks like a problem in `gui.spin`. Open an arbitrary widget, e.g. Tree Learner and key up/down on some spin. It doesn't recompute.

Orange/widgets/evaluate/owparameterfitter.py
VesnaT: That's ok with me, as long as showing the warning also implies to not fitting the model.

markotoplak:
You likely noticed this already, but this PR is not Python 3.9 compatible. Or are we planning to drop support for 3.9? Numpy already did according to https://numpy.org/neps/nep-0029-deprecation_policy.html.

janezd:
I added rudimentary documentation. The widget is bound to significantly change in the future, so it's not worth spending too much time on this.

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -0,0 +1,634 @@
...
+FitterResults = tuple[list[ScoreType], str, str]
+
+
+def _validate(
+        data: Table,
+        learner: Learner,
+        scorer: type[Score]
+) -> tuple[float, float]:
+    # dummy call - Validation would silence the exceptions
+    learner(data)
```
lanzagar, line 48: I don't like this.
A better solution would be to change the evaluation classes below so that they raise the learner's error instead of silently suppressing them (at least optionally, controlled with an argument?)

Orange/regression/pls.py
```diff
@@ -256,6 +255,10 @@ def incompatibility_reason(self, domain):
                     reason = "Only numeric target variables expected."
         return reason
 
+    def fitted_parameters(self, _) -> list[Learner.FittedParameter]:
+        return [self.FittedParameter("n_components", "Number of components",
```
lanzagar, line 259: I think we should strive towards using the same name for a parameter in Parameter Fitter as it is labelled in the learner widget. In this case, PLS has "Components" and Param. fitter has "Number of components".

lanzagar (commented):
- The widget crashes when given a dataset without a class.
- I think the controls in Parameter Fitter should be tied to the Learner input. One should be able to select a parameter to fit and the range of values before connecting the data input (since that already starts the fitting and I would like it to use the correct settings). Also, refreshing the data or even inputting a completely different dataset should not reset the settings in Parameter Fitter. I think it is acceptable if they are reset when the Learner input is refreshed, but ideally it would even try to keep the settings when e.g. I just change the name of the learner or some of the other learner parameters (maybe we don't reset to default if the class of the learner stays the same?).

Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -0,0 +1,644 @@
...
+        if not self._learner:
+            return []
+        return self._learner.fitted_parameters
+
+    @property
+    def initial_parameters(self) -> dict:
+        if not self._learner:
+            return {}
+        if isinstance(self._learner, Fitter):
+            return self._learner.get_params(self._data or "classification")
```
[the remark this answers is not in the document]
VesnaT: No, it is needed to get the parameters.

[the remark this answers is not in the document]
markotoplak: This `params` are very scikit-learn specific and I do not think our learners use them in general (and @pavlin-policar did not consider this either in 2017). So accessing them here is likely bad design.

I think they should be implemented as explicitly as `fitted_parameters` (either combined with them or somehow analogously).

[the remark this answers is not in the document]
markotoplak: We discussed this today and this seems to be the best solution.

PR #6930, opened 2024-11-14
[ENH] Datasets: Let the filter override domain and language

### LINKED ISSUE #6931: Datasets: Wrong data set after reloading workflow
**What's wrong?**

After reloading, workflow loads the data set that was last selected (e.g. clicked), not the one that was loaded (e.g. double-clicked).

- Double-click data set A (A is output).
- Click on data set B (still A on the output, as it should).
- Save the workflow.
- Load the workflow.

The loaded workflow has B on the output.

The problem is that `selected_id`, which contains the current selection, is a setting.

**What's your environment?**

Current master.

janezd:
##### Issue

Adding the domain and language combo made it more difficult to load educational data sets.

Also fixes #6931.

##### Description of changes

Make the filter override the domain (and language) setting. The number of data sets is large, so users always filter them anyway.

The fix may also provide an alternative to 16bbfdfeecff0bb6fc6de562014d1faca1e8905d? @markotoplak 

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owdatasets.py
+++ b/Orange/widgets/data/owdatasets.py
@@ -41,6 +41,9 @@ log = logging.getLogger(__name__)
 GENERAL_DOMAIN = None
 ALL_DOMAINS = ""  # The setting is Optional[str], so don't use other types here
 
+# The number of characters at which filter overrides the domain and language
+FILTER_OVERRIDE_LENGTH = 4
+
 
 def ensure_local(index_url, file_path, local_cache_path,
                  force=False, progress_advance=None):
@@ -180,7 +183,7 @@ class SortFilterProxyWithLanguage(QSortFilterProxyModel):
         data = source.index(row, 0).data(Qt.UserRole)
         in_filter = (
             self.__filter is not None
-            and len(self.__filter) >= 4
+            and len(self.__filter) >= FILTER_OVERRIDE_LENGTH
             and self.__filter in data.title.casefold()
         )
         published_ok = data.publication_status == Namespace.PUBLISHED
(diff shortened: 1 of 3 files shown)


janezd:
@markotoplak, can you take a look at this?

### A better fix

Your fix for unpublished data sets works, but it's rather a hack. The reason why you had to do it is, basically, that `selected_id` depends on the filter, but `selected_id` is a setting while filter is not. On the other hand, we don't want to add the filter to setting.

Your trick wouldn't work in this PR. I'd either have to change (= override the saved) language and/or domain combo, or add filter to settings. So I added a "filter hint", which sets the filter when necessary.

### Why it still won't work

There is another problem, related to an existing inconsistency (essentially from the previous PR):
- use the filter to override the domain/language/unpublished,
- double-click a data set that would not be shown without the filter; the data is downloaded and output
- clear the filter.

The data set is still output, but `selected_id` is set to `None`. The user doesn't notice anything strange, but if you save the workflow and reload it, no data set is selected (because `selected_id` is saved as `None`).

Perhaps the best fix is that `selected_id` overrides the filter: if a data set is selected, it is shown, no matter what any filter and domain and language and unpublished say...

markotoplak:
At first I immediately agreed that selected_id should override the filter. But if it does, then searching for something new will always keep the selected data showing, too, which can be confusing. No idea what is good.

When I was reviewing the previous PR, I found the same issue: output != selection. So what is saved is the actual selection, not the output. It can also be problematic in other ways:  you output D1 then select (single click) D2, save. In the loaded workflow, we get D2.

I think the current `selected_id` should represent the output, which should be decoupled from the selection (we probably do not need a setting for it).

Issue #6911, opened 2024-10-10
Predictions puts "?" regression error values on output while a regular numeric value is shown in its own visualization/dialog box


wvdvegte:
**What's wrong?**
**How can we reproduce the problem?**
In the attached workflow using user reviews from Kaggle (included), open Select Columns, and drag 3 out of 4 features to 'Ignored' so that 'compound' is left as the only feature. Now for several rows, Predictions will produce a "?" on its output for 'Linear Regression (error)' while a regular numeric value is shown in its own visualization/dialog box. For instance, for row 347, the error is shown as 2 in the Predictions visualization/dialog box, but instead Data Table (1) shows a "?"

[Review Mining.zip](https://github.com/user-attachments/files/17330611/Review.Mining.zip)

**What's your environment?**

- Operating system: Mac OS Sequoia
- Orange version: 3.37.0
- How you installed Orange: from dmg; updates through add-ons menu

processo:
I can confirm this. When the _Absolute difference_ output of Predictions would be exactly 2 it is turned into missing (?), no matter the model or data.

janezd:
I cannot reproduce the error. I also tried with a simpler workflow, in which I just create the data, but still no luck.

To make sure we're trying this on the same data: could you upload the data the comes from Select Columns?

markotoplak:
Orange 3.38.1 with the fix for this problem was just released. Download it here: https://orangedatamining.com/download

PR #6933, opened 2024-11-20
[FIX] owpredicttions: Remove special magic value 2

ales-erjavec:
##### Issue

Fixes gh-6911

@janezd what was the point of using 2 as invalid value. I cannot find any place where it would be used/checked, 
except where it is replaced with NaN on output.

##### Description of changes

Remove special magic value 2. Just use NaN.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -915,7 +915,6 @@ class OWPredictions(OWWidget):
             name = f"{slot.predictor.name} (error)"
             newmetas.append(ContinuousVariable(name=name))
             err = self.predictionsview.model().errorColumn(index)
-            err[err == 2] = numpy.nan
             newcolumns.append(err)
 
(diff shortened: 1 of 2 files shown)


janezd:
> @janezd what was the point of using 2 as invalid value.

I'm guessing: the error column contained `2` as the error in probability estimation for discrete problems. In discrete data, this was used only for sorting. I probably wanted to put those values last, hence `2` made sense.

This became a problem after 2375dc78a280fa5a1484d602a0705868ae4f4732, which started using `errorColumn` for output.

This fix is OK if we can assume that `numpy` treats `nan`'s as larger than ordinary numbers and will continue to do so, although I can't find this in documentation.

An alternative would be to do `err[err == 2] = numpy.nan` only for classification.

But the assumption is probably going to hold and your fix makes the code simpler.

PR #6934, opened 2024-11-20
[ENH] Scatter Plot: Error Bars

### LINKED ISSUE #6428: Add error bars to the scatter plot
**What's your use case?**

Data Science. Error Bars are crucial. 

Add possibility to add Error Bars

No

VesnaT:
##### Issue

Add Error Bars to points.

Closes #6428.

##### Description of changes
- introduce a dialog for variable selection
- add error bars to plot

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
I had to change glasses to see what's actually in the button.

Screenshot 2024-12-04 at 12 32 20

If any, I think it should be delta, not epsilon. But what I had in mind would be an icon like this.

Screenshot 2024-12-04 at 12 21 21

It takes a bit more space (which I'm sorry for), but it's better to take space that to look weird. :)

I'd also slightly modify the dialog, like this:

Screenshot 2024-12-04 at 12 30 50

I think it was not very clear what "Difference" and "Absolute" meant. Now it may be a bit better, though a tooltip would help.

One minor comment: "Upper" and "Lower" make more sense for y. For x, the order is also weird -- there I'd expect to set the lower limit first.

I cannot push a new commit to this PR, so I'm attaching a diff and the icons.


[interval-horizontal](https://github.com/user-attachments/assets/f3f05e68-4db0-4658-aa75-b0df853ca9b4)
[interval-vertical](https://github.com/user-attachments/assets/96df70f3-e87a-44cb-a4ba-0e5b86a8f93e)
[addicons.patch](https://github.com/user-attachments/files/18007446/addicons.patch)

PR #6955, opened 2024-12-13
[FIX] Orange can be imported without Qt (fixed importing localization)

### LINKED ISSUE #6952: Orange wheel requires Qt
**What's wrong?**

I created a new environment with Orange, without Qt, because I only needed a headless Orange (use case: Orada). Importing orange results in the following error.

```
/Users/janez/miniforge3/envs/orada/bin/python /Users/janez/dev/orada/backend/app.py 
  File "/Users/janez/dev/orada/backend/dataset.py", line 11, in <module>
    import Orange
  File "/Users/janez/miniforge3/envs/orada/lib/python3.12/site-packages/Orange/__init__.py", line 4, in <module>
    from Orange import data
  File "/Users/janez/miniforge3/envs/orada/lib/python3.12/site-packages/Orange/data/__init__.py", line 4, in <module>
    from .variable import *
  File "/Users/janez/miniforge3/envs/orada/lib/python3.12/site-packages/Orange/data/variable.py", line 1, in <module>
    from orangecanvas.utils.localization.si import plsi, plsi_sz, z_besedo
  File "/Users/janez/miniforge3/envs/orada/lib/python3.12/site-packages/orangecanvas/utils/__init__.py", line 12, in <module>
    from AnyQt.QtCore import Qt
  File "/Users/janez/miniforge3/envs/orada/lib/python3.12/site-packages/AnyQt/QtCore.py", line 1, in <module>
    from . import _api
  File "/Users/janez/miniforge3/envs/orada/lib/python3.12/site-packages/AnyQt/_api.py", line 116, in <module>
    raise ImportError("PyQt4, PyQt5, PySide or PySide2 are not available for import")
ImportError: PyQt4, PyQt5, PySide or PySide2 are not available for import

Process finished with exit code 1
```

This happens because `pip instal orange3` runs Orange through `trubar`, which translates it. Translated orange imports from `orangecanvas.utils.localization.si`, which executes `orangecanvas/utils/__init__.py`, which in turn imports Qt. Since `orangecanvas` depends on Qt (though not through `requirements`, to allow using different versions of Qt), it has every right to import it.

I am reporting the problem here, though the first solution that comes to mind is to move `localization` module from `orangecanvas.utils.localization` to `orangecanvas.localization` to avoid importing Qt.

This will affect anybody who wants to run headless Orange. This "anybody"  probably doesn't mean "manybody", hence no need to rush into 3.38.1.

Create an environment, install Orange, import it.

**What's your environment?**
- Operating system: macOS
- Orange version: 3.38

janezd:
##### Issue

Fixes #6952. Requires https://github.com/biolab/orange-canvas-core/pull/320 to be merged and released.

##### Description of changes

Import localization from a place that doesn't require Qt.

The most important change is the one in configuration of Trubar - this will allow running Orange without Qt.

The other two changes in `[Orange/widgets/utils/localization/__init__.py](https://github.com/biolab/orange3/compare/master...janezd:fix-headless-orange?expand=1#diff-3bf6fd4d089b6924794714d11768d90dbac35b1b370e9cfc788c6cd6f216eece)` are not strictly necessary, but  prevent a warning about importing localization from a deprecated module.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owpredictions.py
+++ b/Orange/widgets/evaluate/owpredictions.py
@@ -16,7 +16,6 @@ from AnyQt.QtCore import (
     QModelIndex, pyqtSignal, QTimer,
     QItemSelectionModel, QItemSelection)
 
-from orangecanvas.utils.localization import pl
 from orangewidget.utils.itemmodels import AbstractSortTableModel
 from orangewidget.utils.signals import LazyValue
 
@@ -39,6 +38,7 @@ from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.utils.state_summary import format_summary_details
 from Orange.widgets.utils.colorpalettes import LimitedDiscretePalette
 from Orange.widgets.utils.itemdelegates import TableDataDelegate
+from Orange.widgets.utils.localization import pl
 
 # Input slot for the Predictors channel
 PredictorSlot = NamedTuple(
--- a/Orange/widgets/utils/localization/__init__.py
+++ b/Orange/widgets/utils/localization/__init__.py
@@ -1,3 +1,3 @@
-from orangecanvas.utils.localization import pl
+from orangecanvas.localization import pl
 
 __all__ = ['pl']
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -32,7 +32,7 @@ requirements:
   run:
     - python
     # GUI requirements
-    - orange-canvas-core >=0.2.4,<0.3a
+    - orange-canvas-core >=0.2.5,<0.3a
     - orange-widget-base >=4.25.0
     - anyqt >=0.2.0
     - pyqt >=5.12,!=5.15.1,<6.0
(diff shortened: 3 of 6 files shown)


markotoplak:
I tested these two PR (installed, so translated, at first I forgot about this) and they did fix the reported issue for me .

But then, when testing, if I only remove pyqt* in my environment, I still get complains from Orange's `__init__.py`:

```
  File "<string>", line 1, in <module>
  File "/home/marko/dev/orange3/Orange/__init__.py", line 30, in <module>
    from AnyQt.QtCore import QObject
  File "/home/marko/venv312/lib/python3.12/site-packages/AnyQt/QtCore.py", line 274, in <module>
    from PyQt5.QtCore import *
ModuleNotFoundError: No module named 'PyQt5.QtCore'
```

I also had to remove AnyQt for them to disappear. Did you have to do that too?

PR #6956, opened 2024-12-13
LogisticRegressionLearner: Remove deprecated argument 'multi_class'

janezd:
##### Issue

As per https://scikit-learn.org/1.5/modules/generated/sklearn.linear_model.LogisticRegression.html,

```
Deprecated since version 1.5: multi_class was deprecated in version 1.5 and will be
removed in 1.7. From then on, the recommended ‘multinomial’ will always be used
for n_classes >= 3. Solvers that do not support ‘multinomial’ will raise an error. Use
sklearn.multiclass.OneVsRestClassifier(LogisticRegression()) if you still want to use OvR.
```

##### Description of changes

Remove the argument.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/classification/logistic_regression.py
+++ b/Orange/classification/logistic_regression.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy as np
 import sklearn.linear_model as skl_linear_model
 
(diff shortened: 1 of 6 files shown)


markotoplak:
In `_initialize_wrapped` we'd also have to switch the l1 solver to "saga", because "liblinear" does not support "multiclass".

PR #6959, opened 2024-12-20
[FIX] 1 line headers: only allow single-letter types and flags

### LINKED ISSUE #6945: Problems with Excel/csv reader in Orange 3.38
**What's wrong?**

#6862 fixed reading of flags in 1-line header format, but it "recognizes" flags that are not.

The attached file has an attribute *1999-12-01 Towards a nuclear-weapon-free world:#the need for a new agenda*. The part before `#` is recognized as flag (I don't know which flags is it supposed to contain) and reading fails when converting the column data into floats.

I think that the code should be more strict about what is recognized as flag. Probably it has to be something like `[cm]?[CDST]?#` and that's it. 

Load the file [un-resolutions-1.csv](https://github.com/user-attachments/files/17970212/un-resolutions-1.csv) with the File widget. (CSV Reader works.)

**What's your environment?**

Oranga 3.38 - dmg or from github.

markotoplak:
##### Issue
Fixes #6945

##### Description of changes
This fix creates specific 1-line header matching rules (does not reuse the 3-line header ones).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/io_base.py
+++ b/Orange/data/io_base.py
@@ -169,17 +169,27 @@ class _TableHeader:
              e.g. d#sex,c#age,cC#IQ
         """
 
-        def is_flag(x):
-            return bool(cls._type_from_flag([x])[0] and
-                        _RE_TYPES.match(cls._type_from_flag([x])[0]) or
-                        Flags.RE_ALL.match(cls._flag_from_flag([x])[0]))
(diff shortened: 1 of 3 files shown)


janezd:
I named the groups. Otherwise OK; merge if you agree with my changes.

markotoplak:
Thanks, naming groups is nice!

PR #6977, opened 2025-01-05
Updated Paint Data widget documentation

borondics:
##### Description of changes
Added new screenshots and modified the text accordingly.

I'm not a fan of the example though. It talks more about the clustering than Paint Data itself and I think it is not easy to see what the difference between k-means and hierarchical clustering is in this case with this layout. In any case, I'll leave it as is for now, with an updated image.

##### Includes
- [ ] Code changes
- [ ] Tests
- [X] Documentation


markotoplak:
The images are too big for what they are for. They are made in retina resolution, which is probably unnecessary for the docs.

markotoplak:
@borondics, we discussed this in the lab. So we will allow retina images, BUT we won't allow any image resizing. Therefore, screenshots should be added in exactly the same size as they were taken, and any necessary resizing should go into the .md file (where we have to somehow specify that retina images are shown in 50% scale).

markotoplak:
Please open a corresponding PR on the submodule.

Issue #6971, opened 2024-12-27
Paint Data is buggy - ill defined


borondics:
**What's wrong?**

I have some not very well definable behaviour in Paint Data. I was recreating the documentation example for new screenshots and first I had a couple of crashes with segfault, so not very informative, although I reproduced it twice. Then, after creating a new workflow from scratch it started working but it visibly slows down every time I paint many points to the state that it keeps painting after the mouse is released.

**How can we reproduce the problem?** 

Paint a lot of data (might depend on the computer how much a lot is) with different classes, just like in the help of the widget. Something like this
<img width="858" alt="image" src="https://github.com/user-attachments/assets/c5aeb12a-caaa-4735-b139-845a4cbb0c35" />

**What's your environment?**

- Operating system: MacOS 14.6.1
- Orange version: 3.39.0dev master
- How you installed Orange: manually from the master branch

ales-erjavec:
> ... first I had a couple of crashes with segfault, ...

Can you open *~/Library/Logs/DiagnosticReports* folder and look for *Python-${DATE}-some-gibberish.ips* file (with $DATE when this occurred if you still remember) open it and post the contents.

> Then, after creating a new workflow from scratch it started working but it visibly slows down every time I paint many points to the state that it keeps painting after the mouse is released.

I cannot reproduce this. What version of PyQt and pyqtgraph do you have installed?

borondics:
The Python crash files are not there in the logs...

I have: 
pyqtgraph==0.13.7
AnyQt==0.2.0
PyQt6==6.7.1

ales-erjavec:
> .. first I had a couple of crashes with segfault, ...

Did you have Hierarchical clustering widget connected to the Paint widget somewhere downstream? I found this #6989 while trying to reproduce this.

> visibly slows down every time I paint many points to the state that it keeps painting after the mouse is released.

Did you have other widgets connected downstream and had Send Automatically enabled? How many points exactly is many?

janezd:
I can't replicate this - despite trying hard and despite using this widget a lot. Also, @ales-erjavec did something in @6990 that could have helped. Let us close this.

PR #6990, opened 2025-01-16
[ENH] Paint Data: Slight optimization

ales-erjavec:
##### Issue

Ref: #6971

##### Description of changes

Slight optimization. Most of the time still spent in ScatterPlotItem constructor.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owpaintdata.py
+++ b/Orange/widgets/data/owpaintdata.py
@@ -1074,19 +1074,14 @@ class OWPaintData(OWWidget):
 
         self.commit.deferred()
 
-    def add_new_class_label(self, undoable=True):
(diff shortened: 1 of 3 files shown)


borondics:
@ales-erjavec, this seems to fix things. I don't get a slowdown anymore. I tried `Paint Data` alone and also when connected to `Distances` and `Hierarchical Clustering` with auto commit on up to about 1000 points. The widget is now responsive. 

Previously, when there were many (~500 points) the painting didn't catch up with the mouse action, meaning that when I released the button, the points still kept coming in. 

Thanks!

PR #6997, opened 2025-01-19
setup: Import trubar locally

### LINKED ISSUE #6992: trubar and setup.py
Currently you can't run setup.py without installing trubar first. The import `from trubar import translate` should be moved to some local scope, where translate is actually used or handled in a similar manner as sphinx, which is imported in a try block.

janezd:
##### Issue

Fixes #6992.

##### Description of changes

Editable install (pip install -e, python setup.py develop) does not run trubar. This PR imports it locally to avoid the unnecessary dependency.

Add-ons already do so. We made a mistake only here.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,8 +11,6 @@ from distutils.command.build_ext import build_ext
 from distutils.command import config, build
 from distutils.core import Extension
 
-from trubar import translate
-
 try:
     import numpy
     have_numpy = True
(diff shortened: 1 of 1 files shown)


markotoplak:
Thanks.

Now I understand how this slipped through. I was only testing editable installs with `pip install -e .`, which worked properly even with old version (takes care of install dependencies separately).

This PR fixes `python setup.py develop`

PR #6998, opened 2025-01-19
Create Class: Add regular expressions

### LINKED ISSUE #6962: “regular expression” in “Create class” does not support full RE syntax
**What's wrong?**

The documentation for “Create class” says: “Substring: regex-defined substring that will match the values from the above-defined attribute”
I took this to mean that you can define an arbitrary regular expression in the Substring field, but as can be seen in the attached screenshot, when I entered ``.*``—which should match _any_ string—I got no matches at all.  I have played around with the input and it seems that what you enter is actually treated as a literal substring to be matched.  If this is the intended behaviour, then please treat this as a feature request:
I’d like the Substring field to support the full Python regular expression syntax, including ``\number``, so that matched substrings can be used in the Name section.  In my concrete case I want to enter the Substring ``(.*) \([12]\)`` and use ``\1`` in the Name field, so that the item ``Belgium (2)`` is renamed ``Belgium``.

Failing that, “regex-defined” should be deleted from the documentation text.

Input any data.  Connect to a “Create Class” widget.  Select any string feature.  Enter ``.*`` as Substring.  Note that no strings are matched.

**What's your environment?**

- Operating system: macOS 15.2
- Orange version: 38.8.0
- How you installed Orange: From https://download.biolab.si/download/files/Orange3-3.38.0-Python3.10.11-x86_64.dmg
Screenshot 2024-12-20 at 11 45 44

janezd:
##### Issue

Closes #6962.

##### Description of changes

Regular expressions and a bit of housekeeping (e.g. `__eq__` and `__hash__`).

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owcreateclass.py
+++ b/Orange/widgets/data/owcreateclass.py
@@ -1,6 +1,7 @@
 """Widget for creating classes from non-numeric attribute by substrings"""
 import re
 from itertools import count
+from typing import Optional, Sequence
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


Orange/widgets/data/owcreateclass.py
```diff
@@ -90,26 +134,14 @@ def transform(self, c):
...
 
-    def __hash__(self):
-        return hash((type(self), self.variable,
-                     tuple(self.patterns),
-                     self.case_sensitive, self.match_beginning,
-                     self.map_values))
-
-
-class ValueFromDiscreteSubstring(Lookup):
+class ValueFromDiscreteSubstring(Lookup, _EqHashMixin):
```
VesnaT, line 144: I suppose this should be the other way around:
`ValueFromDiscreteSubstring(_EqHashMixin, Lookup)`

Orange/widgets/data/owcreateclass.py
```diff
@@ -401,23 +453,44 @@ def class_labels(self):
...
-            if not self.case_sensitive:
-                pattern = pattern.lower()
-            indices = np.char.find(strings, pattern.strip())
-            return indices == 0 if self.match_beginning else indices != -1
-
-        def _lower_if_needed(strings):
-            return strings if self.case_sensitive else np.char.lower(strings)
+        if self.regular_expressions:
+            def _matcher(strings, pattern):
+                re_pattern = re.compile(
```
VesnaT, line 473: How about refactoring the `map_by_substring` and reusing it here?

PR #7010, opened 2025-01-30
[ENH] K-means: clusters can be inferred for new data

markotoplak:
##### Issue
@borondics had a so big data set that k-means was too slow, so he tried doing it on a sample and then modelling the clustering with a classifier. But for k-means this should not be needed because we could use the means/medoids to "predict" the cluster directly.

With one additional line we could make the Cluster useful with Apply domain.

image(3)

Why don't we already do it? All the machinery is already there, in `ClusteringModel`, which does seems unused though. Does 

##### Includes
- [X] Code changes
- [x] Tests
- [ ] Documentation

### DIFF
--- a/Orange/clustering/clustering.py
+++ b/Orange/clustering/clustering.py
@@ -13,7 +13,11 @@ class ClusteringModel:
         self.projector = projector
         self.domain = None
         self.original_domain = None
-        self.labels = projector.labels_
+
+    @property
+    def labels(self):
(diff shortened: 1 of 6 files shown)


markotoplak:
@janezd, I finally finished `__eq__` and `__hash__`. To make it work, I needed to change some attributes into read-only properties (which also made me rewrite a test that relied on changing them).

VesnaT (approved):

PR #7019, opened 2025-02-05
[FIX] Save Data: fix when saving to different drives on Windows

### LINKED ISSUE #7018: File save causes exception if the workflow and destination are not on the same filesystem
**What's wrong?**

Hi! 

On a Windows machine, when a selected Save Data destination path and the workflow save directory are not on the same filesystem (e.g. one or both are mapped network drives), os.path.relpath on [this line](
https://github.com/biolab/orange3/blob/854212293cac175625e381c47e72de8bbfe36141/Orange/widgets/utils/save/owsavebase.py#L127) erros out with 

```
ValueError: path is on mount '<drive letter>', start on mount '<drive letter>'
```

Image

Create a new Workflow. Save it to a location on a certain drive. Add e.g. a Multifile widget and a Save Data widget. Set the path on the Save Data widget to a location on a different drive

(On the stated version the file picker comes back up after the exception, so it's not obvious there was an exception - on a colleague's newer version an exception box comes up.)

**What's your environment?**

- Orange version: Quasar 1.10.1
- How you installed Orange: Windows installer package

Sorry for the trouble - there's a very simple workaround, of course; just save to the same file system. I figured this'd probably be worth submitting anyway.

Thanks for this great software!

markotoplak:
##### Issue
Fixes #7018 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/utils/save/owsavebase.py
+++ b/Orange/widgets/utils/save/owsavebase.py
@@ -124,9 +124,13 @@ class OWSaveBase(widget.OWWidget, openclass=True):
         self.stored_path = absolute_path
         workflow_dir = self.workflowEnv().get("basedir", None)
         if workflow_dir:
-            relative_path = os.path.relpath(absolute_path, start=workflow_dir)
-            if not relative_path.startswith(".."):
-                self.stored_path = relative_path
+            try:
+                relative_path = os.path.relpath(absolute_path, start=workflow_dir)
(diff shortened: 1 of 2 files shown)


markotoplak:
I added some paths to owsavebase windows tests in the first commit. I've verified that they were able to reproduce the issue on Github Actions.

The modified code does not exhibit the same issue anymore.

PR #7023, opened 2025-02-11
[FIX] ScoringSheet deterministic results

### LINKED ISSUE #6924: Scoring sheets model is not reproducible
Scoring sheets return different results with the same data and parameters (I verified this manually by looking at the models). This makes the workflows non-reproducible.

I saw this at first because tests occasionally fail, like here: https://github.com/biolab/orange3/actions/runs/11673277446/job/32503650572?pr=6914

ZanMervic:
##### Issue
Fixes #6924 

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/classification/scoringsheet.py
+++ b/Orange/classification/scoringsheet.py
@@ -56,7 +56,9 @@ class ScoringSheetLearner(Learner):
         if preprocessors is None:
             self.preprocessors = [
                 *self.preprocessors,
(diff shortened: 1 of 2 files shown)


markotoplak:
Oh, that was where the randomness was coming from? Thanks!

Could you also make a test which tests for actual parameter value (hardcoded in the tests)?

PR #7025, opened 2025-02-11
[FIX] ScoringSheetViewer computing target class values from base values

### LINKED ISSUE #6996: Scoring Sheet Viewer: Refactor
**What's wrong?**

_class_combo_changed (https://github.com/biolab/orange3/blob/master/Orange/widgets/visualize/owscoringsheetviewer.py#L446C9-L446C29) checks whether the class indeed changed and if so, they (indirectly) call https://github.com/biolab/orange3/blob/master/Orange/widgets/visualize/owscoringsheetviewer.py#L459, which just negates some coefficients and subtracts risks from 100. I don't think this is very safe. 

Switching back and forth can easily go wrong. The widget should remember the values for one target and use them to compute that values for the shown target. I suspect this may be the reason for failing test (see e.g. #6995).

Test fails randomly, but apparently only on github CI, not locally.

ZanMervic:
##### Issue
Fixes #6996 

##### Description of changes
- Refactors the ScoringSheetViewer widget so the displayed values are computed from base values which never change.
- Sorts the attributes by importance (absolute value of coefficients), so the most important attributes are displayed on top.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owscoringsheetviewer.py
+++ b/Orange/widgets/visualize/owscoringsheetviewer.py
@@ -338,8 +338,11 @@ class OWScoringSheetViewer(OWWidget):
         self.instance = None
         self.instance_points = []
         self.classifier = None
-        self.coefficients = None
+        self._base_coefficients = None
+        self._base_all_scores = None
+        self._base_all_risks = None
         self.attributes = None
+        self.coefficients = None
         self.all_scores = None
         self.all_risks = None
         self.domain = None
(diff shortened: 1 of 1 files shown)


Orange/widgets/visualize/owscoringsheetviewer.py
```diff
@@ -510,15 +515,32 @@ def _extract_data_from_model(self, classifier):
         all_risks = 1 / (1 + np.exp(-all_scaled_scores))
 
         self.attributes = attributes
+        self._base_coefficients = coefficients
+        self._base_all_scores = all_scores.tolist()
+        self._base_all_risks = (all_risks * 100).tolist()
         self.coefficients = coefficients
```
janezd, line 521: Are these three lines still necessary? These attributes are set in `self._adjust_for_target_class`, which is called a few lines later.

Orange/widgets/visualize/owscoringsheetviewer.py
```diff
@@ -510,15 +515,32 @@ def _extract_data_from_model(self, classifier):
...
         # set for the wrong target class. This is a workaround to fix that.
+        self._sort_attributes_by_importance()
         self._adjust_for_target_class()
 
+    def _sort_attributes_by_importance(self):
+        """
+        Sorts the attributes in descending order of importance (absolute coefficient value)
+        by combining the attributes and coefficients into tuples and sorting them.
+        """
+        combined = list(
```
janezd, line 536: No need to call `list` here. You can pass a `zip` object to `sorted`; `sorted` will work on any iterable.

janezd:
I simplified the initialization of coefficients from data -- some lines probably remained from the way it was implemented before. Can someone (@ZanMervic or @VesnaT) check it? Thanks!

PR #7027, opened 2025-02-12
[ENH] Parameter Fitter: more responsive when changing settings

markotoplak:
Callbacks (and thus interruptions) also go into CV.

I was checking why are Parameter Fitter tests so slow. This does not change much for slow tests, but does improve the responsiveness of setting changes with slow learners a lot.

I then added a commit that tries making tests run faster and thus not timeout on Actions. Mainly by actively cancelling running computations when results are not needs.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/owparameterfitter.py
+++ b/Orange/widgets/evaluate/owparameterfitter.py
@@ -72,7 +72,6 @@ def _search(
     scorer = AUC if data.domain.has_discrete_class else R2
     name = fitted_parameter_props.name
     for i, value in enumerate(steps):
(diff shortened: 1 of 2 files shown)


Orange/widgets/evaluate/owparameterfitter.py
```diff
@@ -68,7 +75,8 @@ def _search(
         progress_callback(i / len(steps))
```
VesnaT, line 75: This line could now be removed.

VesnaT (approved):

PR #7038, opened 2025-02-21
MAPE: multiply by 100

### LINKED ISSUE #7032: Predictions / Test & Score: MAPE of linear regression returns fraction rather than percentage
**What's wrong?**
The Mean Average Percentage Error column recently added to Predictions and Test & Score (#6988) always returns a number between 0 and 1. As it has "percentage" in its name, this outcome should be multiplied by 100 in order to reflect a true percentage. I agree that a number between 0 and 1 may be more 'scientific', but a percentage is a percentage ...

**What's your environment?**
- Orange version: 3.38.1

ajdapretnar:
##### Issue

Fixes #7032.

##### Description of changes
Multiply MAPE by 100.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/tests/test_evaluation_testing.py
+++ b/Orange/tests/test_evaluation_testing.py
@@ -10,9 +10,10 @@ from Orange.classification import NaiveBayesLearner, MajorityLearner
 from Orange.evaluation.testing import Validation
 from Orange.regression import LinearRegressionLearner, MeanLearner
 from Orange.data import Table, Domain, DiscreteVariable
-from Orange.evaluation import (Results, CrossValidation, LeaveOneOut, TestOnTrainingData,
+from Orange.evaluation import (Results, CrossValidation, LeaveOneOut,
+                               TestOnTrainingData,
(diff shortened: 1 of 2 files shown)


Orange/tests/test_evaluation_testing.py
```diff
@@ -191,6 +192,11 @@ def test_continuous(self):
             self.housing, [LinearRegressionLearner()])
         self.assertLess(RMSE(res), 5)
 
+    def test_mape_percentage(self):
+        res = CrossValidation(k=3)(
+            self.housing, [LinearRegressionLearner()])
+        self.assertGreater(MAPE(res), 1)
```
janezd, line 198: Not necessarily true in general, but apparently true for housing. :)

ajdapretnar: Absolutely, however, it is very difficult to test whether something is truly in the range of 0-100 and not 0-1. :)

Issue #7043, opened 2025-03-04
Red connection: hint how to resolve


ajdapretnar:
**What's your use case?**
Feedback from the user: it was difficult to understand how to resolve the red connection between widgets.

<img width="278" alt="Image" src="https://github.com/user-attachments/assets/37f38825-2b74-4f1e-b4a4-064e3d7a1a29" />

**What's your proposed solution?**
Perhaps a hint over the connection on how to resolve? Where to put it? On any Text widget (if not Corpus, then raise hint?)?
Need discussion.

**Are there any alternative solutions?**
Add to documentation? Explain in a tutorial?

PR #7045, opened 2025-03-07
[FIX] Disable dynamic signals where applicable

ales-erjavec:
##### Issue

Ref: https://github.com/biolab/orange3/issues/7043

CSV Import in particular should not have declared dynamic output.

##### Description of changes

Disable dynamic outputs where applicable.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owmergedata.py
+++ b/Orange/widgets/data/owmergedata.py
@@ -256,7 +256,8 @@ class OWMergeData(widget.OWWidget):
     class Outputs:
         data = Output("Data",
                       Orange.data.Table,
-                      replaces=["Merged Data A+B", "Merged Data B+A", "Merged Data"])
+                      replaces=["Merged Data A+B", "Merged Data B+A", "Merged Data"],
+                      dynamic=False)
 
     LeftJoin, InnerJoin, OuterJoin = range(3)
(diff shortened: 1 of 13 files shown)


Orange/widgets/data/owmergedata.py
```diff
@@ -256,7 +256,8 @@ class Inputs:
     class Outputs:
         data = Output("Data",
                       Orange.data.Table,
-                      replaces=["Merged Data A+B", "Merged Data B+A", "Merged Data"])
+                      replaces=["Merged Data A+B", "Merged Data B+A", "Merged Data"],
+                      dynamic=False)
```
ajdapretnar, line 260: This one I'm unsure of. Could I not merge two corpora? I do this often (append metadata to a corpus).

ales-erjavec: From what I can gather this will always create a plain Table
https://github.com/biolab/orange3/blob/3.38.1/Orange/widgets/data/owmergedata.py#L630-L632
```python
        table = Orange.data.Table.from_numpy(domain, X, Y, metas)
        table.name = getattr(self.data, 'name', '')
        table.attributes = getattr(self.data, 'attributes', {})
```

PR #7057, opened 2025-03-28
setup.cfg: Remove dashes in options

janezd:
##### Issue

We've probably been bitten by https://setuptools.pypa.io/en/stable/history.html#deprecations-and-removals -- all tests fail in build.

##### Description of changes

Change `with-` with `with_`, i.e. `with-htmlhelp` to `with_htmlhelp`.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [config]
 # applies to sdist and build commands
-with-htmlhelp=1
+with_htmlhelp=1
 
 [aliases]
 # build a sdist and a wheel release with included widget help


janezd:
... and then we get

```
distutils.errors.DistutilsGetoptError: invalid long option name 'with_htmlhelp' (must be letters, numbers, hyphens only
Backend: run command build_sdist with args {'sdist_directory': '/Users/runner/work/orange3/orange3/.tox/.pkg/dist', 'config_settings': None}
```

`setup.cfg` wants `_`, not `-`, and `distutils` only accept dashes?!? I must be overlooking something?!

janezd:
Recommended reading: the uproar in https://github.com/pypa/setuptools/pull/4911/files#r2010652906.

TL;DR: After causing some chaos, the change in setuptools was reverted/postponed because it also breaks the installation of (possibly unmaintained) upstream dependencies, which is supposedly not a problem because:

- it only[!] breaks the upstream dependencies that don't publish a wheel,
- you can pin setuptools [forever?],
- you can contact the packages you depend on to do the change,
- and even, you can  "organize funding to update upstream dependencies or find alternatives or fork them".

The best angry reply was: "**If the plan is to try this again in 2026, the package to be forked will be setuptools.**", and a more polite one reads "**learn to pick and choose your battles, and something like underscores-vs-hyphens usually isn't worth the fight.**"

Fixing Orange (similar for other packages) requires changing a single `-` into `_`.

The humanity learned nothing from [Fillioque](https://en.wikipedia.org/wiki/Filioque).

PR #7066, opened 2025-04-18
tSNE: Add compute_value to tSNE variables

VesnaT:
##### Issue

The tSNE output cannot be used to apply the transformation to new instances, but it should be.

##### Description of changes
Add compute value to the output tSNE variables.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


markotoplak:
I added a test that failed before and a suggestion for a fix. Fix free to modify.

PR #7083, opened 2025-05-16
Tree viewer label

janezd:
##### Issue

Add a list of instance labels at the top of the node tooltips.

##### Description of changes

Added.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owtreeviewer.py
+++ b/Orange/widgets/visualize/owtreeviewer.py
@@ -1,5 +1,6 @@
 """Widget for visualization of tree models"""
 from html import escape
+from typing import Optional
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


janezd:
Sorry, @VesnaT, @markotoplak made me do it. :)

Screenshot 2025-05-17 at 12 08 15

In most cases, the user won't notice a thing (except the new combo). Only in data with a string attribute that has at least 80 % unique values, it will be selected as default in the combo and shown in the tree.

The combo's label, "Node labels:", should now be clear and its effect is immediate.

I admit that this feature is useless outside the classroom, but so is the widget itself. In the classroom, though, this feature will be appreciated.

janezd:
@VesnaT, I changed it as I proposed above: in case any label (in that node!) looks numeric, it shows `(+ n more)`, otherwise it still shows just `+ n`. The test is limited to a node so it could, in principle, show `(+ n more)` in some and `+ n` in others. I could check the entire column instead, but then I'd have to cache this ... and I don't consider this improbably situation worth the effort. :)

janezd:
Dots instead of numbers.

PR #7085, opened 2025-05-20
Feature Statitics: Rename to Variable Statistics

### LINKED ISSUE #7084: Feature Statistics: address inadequate naming
The Feature statistics widget does not only provide statistics of features, but also of targets and metas. Since this is useful, it would make sense to rename it to Variable Statistics. At the same time, the inconsistency in column naming between its visualization window and the Statistics output can be addressed. In the visualization window, the header of the column is Name, while in the Statistics output, its header is Feature. It would be better to use Variable in both cases.
As a use case, consider this workflow I created for collaborative recommendation from user ratings of books (i.e., finding users with similar taste, based on their ratings, then recommend the books they rated highly but haven't already been rated by the reference) for teaching purposes:

Image

Here, in the Select Columns before merge, I have to tell my students to set "Feature" as meta variable, which is totally confusing ...

janezd:
##### Issue

Closes #7084.

##### Description of changes

- Rename Feature Statistics to Variable Statistics.
- Rename the column from "Name" to "Variable" in the Variable Statistics table.
- Rename the output variable from "Feature" to "Variable".

Backward compatibility (non-)issues:

- Renaming the output variable will break existing workflows that ... huh, that ... filtered by this variable?! Such workflows are unlikely, and the fix would introduce some ugly code, so let us not worry about it.
- Since only the name (and not the qualified name) has changed, existing workflows won't be affected.
- I added a keyword so that typing "feature" still finds the widget.

##### Includes
- [X] Code changes
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfeaturestatistics.py
+++ b/Orange/widgets/data/owfeaturestatistics.py
@@ -119,7 +119,7 @@ class FeatureStatisticsTableModel(AbstractSortTableModel):
         @property
         def name(self):
             return {self.ICON: '',
-                    self.NAME: 'Name',
+                    self.NAME: 'Variable',
                     self.DISTRIBUTION: 'Distribution',
                     self.CENTER: 'Mean',
                     self.MODE: 'Mode',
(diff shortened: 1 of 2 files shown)


markotoplak:
Column Statistics?

wvdvegte:
Column Statistics would also be better than Feature Statistics

janezd:
Decision: Column.

lanzagar:
Merging, but the description has an open _To do: documentation_, that should not be forgotten.
Since the widget name itself was changed, it will be hard to find the right docs for new users
https://github.com/biolab/orange3-doc-visual-programming/blob/master/source/widgets/data/featurestatistics.md

PR #7092, opened 2025-05-23
[ENH] OWSilhouettePlot: displays average silhouette

BlazZupan:
##### Issue

Silhouette Plot would display silhouette score per group, but not the average silhouette score which may serve as a reference to judge if the group is below or above the average.

image

Also, learning about average silhouette score may be beneficial when judging on the quality of some clustering that precedes Silhouette Plot widget, say, a Hierarchical Clustering or DBSCAN and Silhouette Plot combination.

##### Description of changes

This pull request introduces an info box that reports on the average silhouette score:

image

The style of reporting (the label in bold and the number in normal text) follows that of an Info widget. When input data is missing, the box displays N/A.

Unit tests should most likely be added. I have currently tested the widget by coupling it in the workflow with k-means, that also reports on the average silhouette:

image

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owsilhouetteplot.py
+++ b/Orange/widgets/visualize/owsilhouetteplot.py
@@ -177,6 +177,11 @@ class OWSilhouettePlot(widget.OWWidget):
         ibox.setFixedWidth(ibox.sizeHint().width())
         warning.setVisible(False)
 
+        # box displaying the average silhouette score
+        avg_box = gui.vBox(self.controlArea, "Average Silhouette Score")
+        self.avg_silhouette_label = gui.widgetLabel(avg_box, "")
+        self._update_avg_silhouette()
+
         gui.rubber(self.controlArea)
 
         gui.auto_send(self.buttonsArea, self, "auto_commit")
@@ -292,6 +297,7 @@ class OWSilhouettePlot(widget.OWWidget):
         self._clear_scene()
         self.Error.clear()
         self.Warning.clear()
+        self._update_avg_silhouette()  # Update the average silhouette display
 
     def _clear_scene(self):
         # Clear the graphics scene and associated objects
@@ -340,6 +346,15 @@ class OWSilhouettePlot(widget.OWWidget):
             else:
                 assert False, "invalid state"
 
+    def _update_avg_silhouette(self):
+        # update the average silhouette score
+        if self._silhouette is not None and len(self._silhouette) > 0:
+            avg_score = np.mean(self._silhouette)
+            self.avg_silhouette_label.setText(
+                f"<b>Silhouette:</b> {avg_score:.4f}")
+        else:
+            self.avg_silhouette_label.setText("<b>Silhouette:</b> N/A")
+
     def _update(self):
         # Update/recompute the effective distances and scores as required.
         self._clear_messages()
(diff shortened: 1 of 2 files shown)


janezd:
Hm, what about putting this on the graph, say at the top left, in two lines?

Silhouette scores for groups are already there. The second reason is that if it is a part of the figure, it is saved together with the figure.

Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -340,6 +346,15 @@ def _ensure_matrix(self):
...
                 assert False, "invalid state"
 
+    def _update_avg_silhouette(self):
+        # update the average silhouette score
+        if self._silhouette is not None and len(self._silhouette) > 0:
+            avg_score = np.mean(self._silhouette)
+            self.avg_silhouette_label.setText(
+                f"<b>Silhouette:</b> {avg_score:.4f}")
+        else:
+            self.avg_silhouette_label.setText("<b>Silhouette:</b> N/A")
```
janezd, line 356: Must it be bold? I don't think we use bold anywhere else (or at least not often).

Orange/widgets/visualize/owsilhouetteplot.py
```diff
@@ -384,6 +399,8 @@ def _update(self):
                 self.Warning.nan_distances(
                     count_nandist, s="s" if count_nandist > 1 else "")
 
+        self._update_avg_silhouette()  # Update the average silhouette display
```
janezd, line 402: Is this comment necessary?

I know: copilot and similar tools add heaps and heaps of comments to explain the code, but they often just state the obvious. [PEP8 says they're distracting](https://peps.python.org/pep-0008/#comments:~:text=Inline%20comments%20are%20unnecessary%20and%20in%20fact%20distracting%20if%20they%20state%20the%20obvious.%20Don%E2%80%99t%20do%20this%3A).

Orange/widgets/visualize/owsilhouetteplot.py
BlazZupan: We used bold in Data Info widget, I copied the style from there. But I agree it sticks out, and will correct. Let me work on this in the following days.

ajdapretnar:
I was working on this in parallel with Blaž. Here is my proposed solution: #7106

BlazZupan:
The score is now displayed in the Info box at the top of the control area, using the same style as in the Data Table.

image

janezd:
@BlazZupan, for some reason, github doesn't allow me to push new commits to your branch. Could you, please, update the translation file? Here are the changes (be careful with indentation).

```
             Ignoring categorical features: Ignoriram kategorične spremenljivke
         def `__init__`:
+            Info: Podatki
             Distance: Razdalja
             distance_idx: false
             Grouping: Skupine
@@ -15095,6 +15096,9 @@ widgets/visualize/owsilhouetteplot.py:
             Input matrix does not have associated data: Matrika razdalj ne vsebuje podatkov.
         def `_ensure_matrix`:
             invalid state: false
+        def `_update_avg_silhouette`:
+            'Average Silhouette: {avg_score:.4f}': 'Povprečna silhueta: {avg_score:.4f}'
+            'Average Silhouette: N/A': 'Povprečna silhueta:'
         def `_update`:
```

PR #7097, opened 2025-06-02
[FIX] Update check and notifications: avoid compression

markotoplak:
##### Issue

Update check failed for me because the response was compresses, and `urllib` (if I understand it) does not decompress it automatically.

```
2025-06-02 13:30:08,020:ERROR:Orange.canvas.__main__: Failed to check for updates Traceback (most recent call last):
  File "/home/marko/dev/orange3/Orange/canvas/__main__.py", line 97, in run
    contents = urlopen(request, timeout=10).read().decode()
````

##### Description of changes

Force `identity` encoding.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/canvas/__main__.py
+++ b/Orange/canvas/__main__.py
@@ -90,7 +90,7 @@ def check_for_updates():
                     request = Request('https://orange.biolab.si/version/',
                                       headers={
                                           'Accept': 'text/plain',
-                                          'Accept-Encoding': 'gzip, deflate',
+                                          'Accept-Encoding': 'identity',
                                           'Connection': 'close',
                                           'User-Agent': ua_string()})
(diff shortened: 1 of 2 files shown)


markotoplak:
There is other place in `orange3` code base where we could get a similar problem: in `UrlReader`. Does anyone remember any problems with it?

Here, I think we had a bug because we even requested a gzip, and the server happily started obliging. But I am not sure what is the default value of                                   'Accept-Encoding' in urllib and how is that handled on servers.

markotoplak:
If this was the problem, users will not be notified of a new version.

@lanzagar suggested setting the server not to compress, if this is possible (this goes through Cloudflare).

PR #7110, opened 2025-06-12
Test with Qt 6.8

markotoplak:


markotoplak:
Qt6 tests pass, merging.

PR #7118, opened 2025-07-04
SklAdaBoostClassificationLearner: deprecate `algorithm`

markotoplak:
##### Issue
Scikit-learn deprecated `algorithm` in 1.6 (to expire in 1.8).

Depends on #7117

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
@markotoplak, could you please rebase this to the current master? I would like to mark the new string as not needing a translation (or you can do it, if you will), but it would make a bit of a mess (possibly conflicts) if I do it on the old messages file.

PR #7119, opened 2025-07-04
Remove explicit pip requirement (oldest tests now force 20.3)

markotoplak:
##### Issue
Oldest test fails. Now they still fail (timeout when the tests are being executed), but at least they get past this:

```pip._vendor.pytoml.core.TomlError: /tmp/pip-install-x83ndwu_/libcst/pyproject.toml(51, 1): array-type-mismatch```

##### Description of changes
require pip>=20.3 (for manylinux_x_y packages like libcst)

See https://github.com/pypa/manylinux

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -54,7 +54,6 @@ requirements:
     - openTSNE >=0.6.2,!=0.7.0
     - pandas >=1.4.0,!=1.5.0,!=2.0.0
     - packaging
-    - pip >=19.3
     - python.app  # [osx]
     - python-louvain >=0.13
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: conda-recipe/meta.yaml [hunk 1]
```diff
@@ -21,7 +21,6 @@ requirements:
     - {{ compiler('cxx') }}
   host:
     - python
-    - pip
```
ales-erjavec, line 24: This should remain in `host` section. It is a build dependency used in `script` above.

markotoplak: Thank you for noticing! Fixed.

PR #7124, opened 2025-07-12
[ENH] Tree Viewer: correctly show the first category (escape class values)

### LINKED ISSUE #7004: Tree viewer doesn't show first category of discretized numeric variable as target
**What's wrong?**
If Tree is used for classification of a discretized numeric variable, the end leafs displayed in Tree Viewer do not show the predicted category if it is the first category (i.e., the one representing the smallest numbers):
Image
In this example (workflow attached), the red leaves show that they represent the category 15.75-19.95 but the blue ones don't show that they represent the category <15.75 - although they do in the tool tip.

See workflow:
[tree viewer bug.ows.zip](https://github.com/user-attachments/files/18536966/tree.viewer.bug.ows.zip)

**What's your environment?**

- Operating system: Mac OS 15.2
- Orange version: 3.38.1
- How you installed Orange: From dmg, updates through add-ons menu

janezd:
##### Issue

Fixes #7004.

@wvdvegte noticed that "Tree viewer doesn't show first category of discretized numeric variable", which is true, and also interestingly misleading. :) I assumed that it has something to do with the index 0 being falsy, while in fact the problem was that the label of the first value of categorized variable starts with a character `<`, and we output this as HTML.

##### Description of changes

Escape the label.

##### Includes
- [X] Code changes
- [ ] Tests

### DIFF
--- a/Orange/widgets/visualize/owtreeviewer.py
+++ b/Orange/widgets/visualize/owtreeviewer.py
@@ -488,7 +488,7 @@ class OWTreeGraph(OWTreeViewer2D):
         else:
             modus = np.argmax(distr)
             tabs = distr[modus]
-            text = f"<b>{self.domain.class_vars[0].values[int(modus)]}</b><br/>"
+            text = f"<b>{escape(self.domain.class_vars[0].values[int(modus)])}</b><br/>"
         if tabs > 0.999:
             text += f"100%, {total}/{total}"
         else:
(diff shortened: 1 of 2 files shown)


ajdapretnar:
Again, I approve this changes, but would prefer the tests passed.
Also, would it make sense to add a small test ensuring we are properly escaping things or not?

PR #7125, opened 2025-07-12
Table: Add 'Clear Selection', disable 'Restore order' when order is original

### LINKED ISSUE #6969: Data Table modifications
I think some of the phrasing and behaviour of Data Table could be improved. Since it is a very heavily used widget I think it would be great to do (some) of the following.

**What's your use case?**

- [ ] The "Variables" section could be renamed to "Visualization", which is more descriptive to me.
- [ ] The Restore Original Order button is active even if instances are in the original order. Deactivating it (similar to the Send Selection button behaviour) would be nice in those cases.
- [ ] The "Send Selection" button changes its text to "Send Automatically" when the checkbox is activated. I think this is confusing. I would simply change the text to "Sending Selection".
- [ ] Deselecting a selected row is not intuitive (needs CTRL-leftclick or CMD-leftclick currently). Could it be changed to simply leftclick?

If this proposal is accepted (and after #6968 is merged) I can work on these (and also update the documentation again).

No.

janezd:
##### Issue

Closes #6969.

##### Description of changes

I think we agreed (in the core group meeting) that the title of the box stays Variables, because it is about variables. We also said that the behaviour of the output button is consistent with other widgets in Orange and that selection behaviour is consistent with common GUIs.

I thus fixed the second point, disabling of reorder button.

As a bonus, I added a button to clear selection (which is also disabled when there's no selection). We have enough space, but its placement doesn't look nice. Perhaps move it to next to "Restore Original Order"?

Screenshot 2025-07-12 at 17 25 19

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtable.py
+++ b/Orange/widgets/data/owtable.py
@@ -239,17 +239,20 @@ class OWTable(OWWidget):
                      callback=self._on_distribution_color_changed)
 
         box = gui.vBox(self.controlArea, "Selection")
-
+        self.clear_button = gui.button(
+            box, self, "Clear Selection", callback=self.clear_selection,
+            autoDefault=False, enabled=False)
         gui.checkBox(box, self, "select_rows", "Select full rows",
                      callback=self._on_select_rows_changed)
 
         gui.rubber(self.controlArea)
 
-        gui.button(self.buttonsArea, self, "Restore Original Order",
-                   callback=self.restore_order,
-                   tooltip="Show rows in the original order",
-                   autoDefault=False,
-                   attribute=Qt.WA_LayoutUsesWidgetRect)
+        self.restore_button = gui.button(
+            self.buttonsArea, self, "Restore Original Order",
+            callback=self.restore_order,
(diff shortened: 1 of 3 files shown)


janezd:
Write tests.

Orange/widgets/data/owtable.py
```diff
@@ -509,7 +515,7 @@ def __restore_sort(self) -> None:
         assert self.input is not None
         sort = self.__pending_sort
         self.__pending_sort = None
-        if sort is None:
+        if not sort:
```
janezd, line 518: Here, it happenned that `sort` was `[]`, not `None`. The remaining code had no effect (as far as I see), but caused the button to be enabled. I could use `self.restore_button.setEnabled(bool(sort))` below, but I think this is a better fix.

janezd:
Apparently, we decided it's OK and that I should "Write tests" (as I wrote on Friday, Aug 29).

Tests are written and this is ready for review. I am volunteering @VesnaT. :)

PR #7127, opened 2025-07-13
CSVImport: Add hint about right-clicking the header rows

### LINKED ISSUE #6788: File widget: read file without header
**What's your use case?**

I have files with no header: the first row of my files are data, so the workaround suggested in https://github.com/biolab/orange3/issues/6675 (right click on the row index to skip it) does not work: the first row is lost forever. 

Add a number field _Number of row for Header_.

- If it is 0, take arbirary names for columns (1,2,3,...), and rename them later.
- If it greater than 1, it will solve [issue-6675, comment-1854075827](https://github.com/biolab/orange3/issues/6675#issuecomment-1854075827).
- Compatibility with previous uses is ensured if the default value is 1.

Do it manually with an editor.

janezd:
##### Issue

Fixes #6788.

##### Description of changes

Adds a hint. If we like it. Otherwise, close the PR and the issue.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/utils/textimport.py
+++ b/Orange/widgets/utils/textimport.py
@@ -704,7 +704,16 @@ class CSVImportWidget(QWidget):
         self.column_type_edit_cb.setCurrentIndex(-1)
 
         form.addRow(QFrame(frameShape=QFrame.HLine))
-        form.addRow("Column type", self.column_type_edit_cb)
+        cb_hint = QHBoxLayout()
+        cb_hint.addWidget(self.column_type_edit_cb)
+        hint = QLabel(
+            "Hint: right click on the row or column header for additional options.")
+        font = hint.font()
+        font.setPointSizeF(0.9 * font.pointSizeF())
+        hint.setFont(font)
+        cb_hint.addWidget(hint)
+        form.addRow("Column type", cb_hint)
+
         layout.addWidget(self.dataview)
         # Overlay error message widget in the bottom left corner of the data
         # view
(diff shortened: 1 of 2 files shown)


Orange/widgets/utils/textimport.py
```diff
@@ -704,7 +704,16 @@ def __init__(self, *args, **kwargs):
         self.column_type_edit_cb.setCurrentIndex(-1)
 
         form.addRow(QFrame(frameShape=QFrame.HLine))
-        form.addRow("Column type", self.column_type_edit_cb)
+        cb_hint = QHBoxLayout()
+        cb_hint.addWidget(self.column_type_edit_cb)
+        hint = QLabel(
+            "Hint: right click on the row or column header for additional options.")
```
ajdapretnar, line 710: right click --> right-click

i18n/si/msgs.jaml
```diff
@@ -13171,6 +13171,7 @@ widgets/utils/textimport.py:
             separator: false
             Ignore: Izpusti
             The column will not be loaded: Stolpca ni mogoče naložiti
+            'Hint: right click on the row or column header for additional options.': Namig: desni klik na glavo vrstice ali stolpca pokaže dodatne možnosti.
```
ajdapretnar, line 13174: right click --> right-click

ajdapretnar:
Frankly, I think the placement of the hint is somewhat strange. It also makes the widget window wider than it was. Perhaps on hover?

janezd:
It is not wider for me, but I agree that the placement is strange.

There are two problems with tooltips, though. The first is that they do not show immediatelly.  They are useful in contexts where the user wants more info and expects to get it by hovering over some element, but not for exposing unexpected functionality. The secon problem is - hover over what?

PR #7128, opened 2025-07-13
CN2 Rules: Add restriction to == for categorical variables

### LINKED ISSUE #7120: Option to disable "not equals" conditions in CN2 Rule Induction
Hi Orange3 team,

I'm using the CN2 Rule Induction model in Orange3 and really enjoying its ability to generate human-readable rules.

However, I’d like to avoid rules that contain "not equals" (≠) conditions. In my use case, such rules are less actionable, and I’d prefer the model to explore alternative rules using "equals" (=) conditions, even if they cover different subsets of the data.

Unfortunately, filtering out ≠ rules after the fact isn’t a good solution because many times those are the only rules returned, and removing them leaves me with nothing. What I’d really like is a way to steer the model to prefer or restrict itself to = conditions during rule induction, ideally via a configuration option or parameter.

Is this currently possible in Orange3? If not, would you consider supporting it?

Thanks so much for your work and for considering this feature!

janezd:
##### Issue

Fixes #7120.

##### Description of changes

I haven't written any tests -- because there are none so far, so I would consider this as a separate project.

As for documentation: can do, but after we agree about this PR.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation


Orange/widgets/model/owrules.py
```diff
@@ -349,6 +357,7 @@ def get_learner_parameters(self):
             ("Covering algorithm", self.storage_covers[self.covering_algorithm]),
             ("Gamma", self.gamma),
             ("Evaluation measure", self.storage_measures[self.evaluation_measure]),
+            ("Restrict equality", self.restrict_equality),
```
markotoplak, line 360: I think here we should have "Restrict to equality" (which is analogous to what is in the Slovenian translation).

markotoplak:
Some tests are in `Orange/tests/test_rules.py` and `Orange/widgets/model/tests/test_owrulesclassification.py`

markotoplak:
I think I may have a more elegant solution to the original problem. When the best rule among a set of rules is chosen, it is, among the ones with the same quality score, always the first.

Now, if we just change the rule generation order... Currently it loops per value, and then per operator. If we change it to loop per operator and then per value, the selection algorithm downstream will prefer the `==` operator.

This approach gives the same result for the `zoo` dataset as with the new option selected. Here is my changed code:

```
-                for val in np.unique(X[:, i]):
+                for op in disc_operators:
                     possible_selectors += (
                         Selector(column=i, op=op, value=val)
-                        for op in disc_operators)
+                        for val in np.unique(X[:, i]))
```

The bad thing is that this approach it is not backward compatible (different rules would be induced with the same settings).

janezd:
Thanks for a great find. I'd keep both.

- Per OP, rules using inequality are less "actionable"; even if the attribute has five values, we should let the user prohibit inequality.
- But your suggestion would give priority to equality, and should always be enabled. I don't care too much about the lost backward compatibility here.

markotoplak:
>     * But your suggestion would give priority to equality, and should always be enabled. I don't care too much about the lost backward compatibility here.

I also don't think compatibility is an issue here.

PR #7150, opened 2025-08-19
Add Cochran rule indicator to Sieve + tests

gmolledaj:
##### Issue NONE
- Adds a language-independent Cochran indicator (green/red square) below χ² and p.
- Exposes an internal flag (_cochran_ok) for testing.
- Adds tests for both cases: condition satisfied and violated.
- No changes to existing color scheme or behavior of Sieve cells.

##### Description of changes
This pull request introduces the following changes:

Cochran’s rule indicator in Sieve widget

Adds a small green/red square below the χ² and p-value labels.

The indicator turns green when Cochran’s rule is satisfied and red otherwise.

A new internal attribute (_cochran_ok) stores the evaluation result.

Unit tests

Added tests to ensure the indicator behaves correctly:

test_cochran_indicator_passes checks that the indicator is green when conditions are satisfied.

test_cochran_indicator_fails checks that the indicator is red when conditions are violated.

These changes improve the statistical robustness of the Sieve visualization and provide users with a quick visual check of Cochran’s rule.

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/visualize/owsieve.py
+++ b/Orange/widgets/visualize/owsieve.py
@@ -516,7 +516,28 @@ class OWSieveDiagram(OWWidget, VizRankMixin(SieveRank)):
                   0, bottom)
         # Assume similar height for both lines
         text("N = " + fmt(chi.n), 0, bottom - xl.boundingRect().height())
+        
+        # Cochran condition check
+        expected = chi.expected
+        total_cells = expected.size
+        num_lt1 = int((expected < 1).sum())
+        num_lt5 = int((expected < 5).sum())
+        cochran_ok = (num_lt1 == 0) and (num_lt5 <= 0.2 * total_cells)
 
+        bottom += 35
+        label_item = text("Cochran:", 0, bottom, Qt.AlignLeft | Qt.AlignVCenter)
+
+        # Green/red square indicator
+        rect_size = 10
+        # Since the text starts at x=0, its width is enough to position the square on the right
+        x_ind = label_item.boundingRect().width() + 6
+        y_ind = bottom - rect_size / 2
+
+        color = QColor("#2ecc71") if cochran_ok else QColor("#e74c3c")
+        indic = CanvasRectangle(self.canvas, x_ind, y_ind, rect_size, rect_size, z=0)
+        indic.setBrush(QBrush(color))
+        indic.setPen(QPen(color))
+      
     def get_widget_name_extension(self):
         if self.data is not None:
             return "{} vs {}".format(self.attr_x.name, self.attr_y.name)
--- a/Orange/widgets/visualize/owsieve.py
+++ b/Orange/widgets/visualize/owsieve.py
@@ -54,7 +54,6 @@ class ChiSqStats:
         self.p = chi2.sf(
             self.chisq, (len(self.probs_x) - 1) * (len(self.probs_y) - 1))
 
-
 class SieveRank(VizRankDialogAttrPair):
     sort_names_in_row = True
 
@@ -124,6 +123,7 @@ class OWSieveDiagram(OWWidget, VizRankMixin(SieveRank)):
         self.mainArea.layout().addWidget(self.canvasView)
         self.canvasView.setVerticalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
         self.canvasView.setHorizontalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
+        self._cochran_ok = None
 
     def sizeHint(self):
         return QSize(450, 550)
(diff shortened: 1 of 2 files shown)


gmolledaj:
Hi, just to clarify the current state of this PR:

When I opened it, I started getting CI errors from other parts of the codebase not related to my changes. While I tried to fix them, it created even more unrelated warnings and errors, although I believe the core changes I introduced are correct (example: refactor comprehensions,  dict creation and filter import).

Regarding translations: there is only one new string, currently "Cochran:". This should more precisely be "Cochran’s rules:", and ideally it should go through the i18n system.

If you prefer, I can close this PR and reopen a clean one with only the original intended changes, without the extra attempts to fix unrelated issues.

Thanks for the review and guidance!

janezd:
@gmolledaj, thanks for your patience.

Some tests crash at random; this is annoying but impossible to debug and fix. We would, hopefully, one day. For now, ignore them.

As for Slovenian translations, it obviously has to be done by somebody else (probably: me) after @thocevar or I review your PR. (We were the obvious choice so we volunteered, but we're both too busy right now.)

gmolledaj:
> @gmolledaj, thanks for your patience.
> 
> Some tests crash at random; this is annoying but impossible to debug and fix. We would, hopefully, one day. For now, ignore them.
> 
> As for Slovenian translations, it obviously has to be done by somebody else (probably: me) after @thocevar or I review your PR. (We were the obvious choice so we volunteered, but we're both too busy right now.)

Thanks for the update! I completely understand that you’re busy right now.
In the meantime, I’m attaching a small dataset I prepared — it produces a case where Cochran’s rule fails (indicator turns red). It might be useful for testing or validation later.
[cochran_fail_example.csv](https://github.com/user-attachments/files/21948302/cochran_fail_example.csv)

janezd:
Warning the user about inadequate data size is a good idea, but the meaning of the red/green square is not self evident. I understand your wish to make it language-independent --- but I don't think it works; there has to be some text to explain it.

@thocevar proposed to make this a warning in the status line, e.g.

```
    class Warning(OWWidget.Warning):
        inadequate_data_size = Msg("Small data size.\n{}")
```

and the you can trigger it with `self.Warning.inadequate_data_size(reason)`, where `reason` is a string that explains which criterion isn't met. It will be shown on mouse hover. Don't forget to clear the warning.

Doing it this way is more in line with other Orange widgets, and it also doesn't take additional space. It also simplifies the code because you don't have to paint the square, and you don't need the extra attribute `_cochran_ok`.

janezd:
In Slovenian, we would say that I have reflexes like a dead horse. I'm buried in work on a project, catching deadlines.

I'd ask you for two changes. The thing is that nobody in our group hasn't heard about the Cochran rule before, so I assume it is not well known.

One is to change `cochran = Msg("Does not meet Cochran's rule")` to `cochran = Msg("Does not meet Cochran's rule\n{}")`. Then please trigger the warning with

```
self.Warning.cochran(msg)
```

where the message describes what is wrong, e.g. "empty cells in contingency table" or whatever. This way, the user will know what the rule requires (and what wasn't met). To get the message, I would change `_cochran_ok` to `_check_cochran` and return either `None` (signaling that everything is OK) or a string with an error message.

Clear the warning with

```
self.Warning.cochran.clear()
```

The second is: please remove `Information`. It is used very sparingly, and even when it is, it means a mild warning. (I just reviewed the uses: most often, we could change it to warning). The reason for avoiding it is that when a widget triggers it, the canvas shows a small icon above it. Its purpose it to attract the user's attention, but we don't want to attract attention when everything is OK. By that logic, any widget that check something (and all widgets do that) would show an info that everything is OK.

This second comment is also related to the fact that most people don't know about the Cochran criterion, so seeing a message that it is met, won't tell them anything.

gmolledaj:
Because “Cochran’s rule” is not widely known, here is a short explanation and references:
Cochran’s rule is a widely used guideline (heuristic) for the validity of the chi-square test in contingency tables. It states that:
* no expected frequency should be less than 1, and
* no more than 20 % of expected frequencies should be less than 5
If these conditions are violated, the χ² approximation may be unreliable (i.e., the Type I error rate may deviate).
References:
· https://www.tandfonline.com/doi/full/10.1080/00031305.2017.1286260 : 
· https://www.statsdirect.com/help/chi_square_tests/chisq_tests.htm : "Cochrane's rule states that no expected frequency should be less than 1 and at least 80% of expected frequencies should be greater than 5 ([Bland, 2000])."
· https://openpublishing.library.umass.edu/pare/article/1485/galley/1436/view/
Cochran, W. (1952). The χ2 test of goodness of fit. Annals of Mathematical Statistics, 23, 315-345. doi.org/10.1214/aoms/1177729380

janezd:
Huh, I've seen your message but overlooked the commit. Sorry. Again. :(

Thanks for your patience and all the changes you've made, including the extra tests.

I added a commit in which I made some cosmetic changes.
- `self.Warning.cochran.clear()` must be called earlier because function has two early exits. In particular, it the widget is showing a warning and then we sent `None` as data, the warning would remain shown. I also added a test for this.
- I undid the change to combo arguments; calling `dict` instead of using string literals has simpler syntax for this case and it doesn't add new (untranslatable) strings for translation.
- I added Slovenian translations.

If (relevant) tests pass, I'll then merge the PR.

Issue #7179, opened 2025-11-05
Test with "oldest dependencies" consistently fail


janezd:
```
    File "/home/runner/work/orange3/orange3/.tox/orange-oldest/lib/python3.10/site-packages/pip/_vendor/toml/decoder.py", line 1002, in load_array
      a[b] = a[b] + ',' + a[b + 1]
  IndexError: list index out of range
  ----------------------------------------
ERROR: Command errored out with exit status 2: /home/runner/work/orange3/orange3/.tox/orange-oldest/bin/python 
```

~~Do we intentionally test on Python 3.10?~~

~~We also have `PYTHON_REQUIRES = ">=3.10"` in setup.py. Is this intentional or forgotten?~~

markotoplak:
Oldest dependencies test should test the oldest supported Python. What is wrong with the setup.py line? Did we stop supporting 3.10?

janezd:
I thought so. It is no longer in the test matrix except for oldest-dependencies.

I changed the issue description; it would still be nice to fix the failing test -- it fails in package installation.

janezd:
Could this be related to updating Ubuntu runners #7096?

The bug happens when the `toml.load` is unable to parse some the pyproject.toml of some dependency (see https://github.com/uiri/toml/issues/270). Based on the logfile, it could be MarkupSafe, but I can't reproduce this on macOS. Can somebody try to install `MarkupSafe` with Python 3.10 in an Ubuntu environment similar to what runs on github?

PR #7188, opened 2025-11-12
[MNT] Drop Python 3.10

ales-erjavec:
##### Issue

Drop Python support 3.10 

Fixes: #7179

##### Description of changes

Drop Python support 3.10 

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -61,7 +61,7 @@ deps =
     oldest: scikit-learn==1.5.1
     oldest: scipy==1.10
     # oldest: serverfiles
-    oldest: xgboost==1.7.4
+    oldest: xgboost==2.1.0
     oldest: xlrd==1.2.0
     # oldest: xlsxwriter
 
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -33,37 +33,35 @@ deps =
...
+    oldest: pandas==1.5.1
     oldest: python-louvain==0.13
     # oldest: pyyaml
     # oldest: requests
     oldest: scikit-learn==1.5.1
-    oldest: scipy==1.9
+    oldest: scipy==1.10
     # oldest: serverfiles
-    oldest: xgboost==1.7.4
+    oldest: xgboost==2.1.0
```
markotoplak: Why did you need to update xgboost? The older version does not work with python 3.11 anmore on linux?

And the rest, were they lifted for compatibility with python 3.11?

Should we also update requirements accordingly?

ales-erjavec: > Why did you need to update xgboost? The older version does not work with python 3.11 anmore on linux?

tox fails in `pip check` step with: `xgboost 1.7.4 is not supported on this platform`.



> And the rest, were they lifted for compatibility with python 3.11?

Mostly oldest available wheels for 3.11

PR #7195, opened 2025-11-20
[FIX] Save dialogs: Avoid weird behavior with extensions on macOS

### LINKED ISSUE #7156: Save widget adds and rejects a suffix
**What's wrong?**

The Save widget sets a default name that includes the extension, e.g. `.tab`. When the user confirms the name, a new dialog appears, saying `You cannot save this document with extension .tab at the end of the name. The required extension is ".tab"`.

This happens because the native dialog adds another .tab and then rejects the double extension.

The user must manually delete the extension set by Orange, which is not only annoying but also quite non-obvious because of the weird error message.

**Supposed fix**

For a start, `initial_start_dir` should probably strip the extension if running on macOS.

The problem would persist: if the user clicks a file (e.g. to overwrite an existing file), the dialog copies the entire file name into the line edit, but then complains about the extension.

Another solution could be to drop the extension from the filter on macOS and add some logic after getting the filename from the dialog. Note that this part is already heuristic and iffy.

Note that Orange doesn't use `setDefaultSuffix` because it doesn't support double extensions, e.g. `tar.gz`. Even if we did, it probably wouldn't help.

File -> Save, click Save As, confirm.

**What's your environment?**

- Operating system: macOS
- Orange version: 3.39.0
- How you installed Orange: DMG

janezd:
##### Issue

Fixes #7156.

1. On macOS, the native file dialog added a (correct) extension and then complained that, e.g. extension .tab is invalid and that the required extension is .tab. This may be due to changed behavior in Qt6 or macOS Sequoia.

2. Furthermore, macOS native dialog already asks whether to override the file (the flag `DontConfirmOverwrite` is ignored), hence the user was asked twice.

##### Description of changes

1. This seems to be fixed by

   a. keeping the `*` in filters and
   b. by not giving the extension to the initial file name.

   Point a won't affect Window because we already had * there. Point 2 can; we need to try and patch if necessary.
  
2. I removed the confirmation dialog for macOS, but kept it for Windows.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/data/owsave.py
+++ b/Orange/widgets/data/owsave.py
@@ -128,14 +128,13 @@ class OWSave(OWSaveBase):
 
     def initial_start_dir(self):
         if self.filename and os.path.exists(os.path.split(self.filename)[0]):
-            return self.filename
+            return os.path.splitext(self.filename)[0]
         else:
             data_name = getattr(self.data, 'name', '')
             if data_name:
                 if self.writer is None:
                     self.filter = self.default_filter()
                 assert self.writer is not None
-                data_name += self.writer.EXTENSIONS[0]
             return os.path.join(self.last_dir or _userhome, data_name)
 
     def valid_filters(self):
--- a/Orange/widgets/data/tests/test_owsave.py
+++ b/Orange/widgets/data/tests/test_owsave.py
@@ -73,7 +73,7 @@ class TestOWSave(OWSaveTestBase):
 
         with patch("os.path.exists", return_value=True):
             widget.filename = _w("/usr/foo/bar.csv")
-            self.assertEqual(widget.initial_start_dir(), widget.filename)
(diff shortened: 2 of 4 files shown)


MatejBevec:
Saving data (.tab), workflows (.ows) and reports (.pdf..) works as expected on Windows on this PR - no complaints about the extension, and only a single overwrite dialog when necessary.

janezd:
@MatejBevec , thanks! I tried to install Windows into a virtual machine in VirtualBox on my macOs, but doesn't seem to work on Arm. It is said to work with VmWare, but I gave up.

janezd:
@markotoplak , can you just try this on Linux, merge if OK, and then I guess we're ready for a release?

markotoplak:
On Linux it works in a very linuxy way: if I use a filename without an extension, it is NOT added, and I do NOT get any questions about it. I personally like this a lot.

The master branch exhibits the same behavior.

PR #7199, opened 2025-11-20
[ENH] Neighbors: Add an option to retain reference example(s)

### LINKED ISSUE #7054: Neighbors: bring back outputting reference+neighbors
**What's your use case?**
As per usual, my use case is text mining. When inputting text data, the output is a Table, which forces the user to add Concatenate then Corpus (to re-set text variable). This is cumbersome.

Bring back outputting reference+neighbors data, which retains the Corpus object. 🙏 

This 😒 
Image

janezd:
##### Issue

Fixes #7054.

@ajdapretnar, please confirm that this is what you meant, before I start writing tests (and add translation)

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -1348,7 +1348,7 @@ class Table(Sequence, Storage):
           appears in multiple dictionaries, the earlier are used.
 
         Args:
-            tables (Table): tables to be joined
+            tables (list of Table): tables to be joined
 
         Returns:
             table (Table)
--- a/Orange/widgets/data/owneighbors.py
+++ b/Orange/widgets/data/owneighbors.py
@@ -42,7 +42,7 @@ class OWNeighbors(OWWidget):
     class Info(OWWidget.Warning):
         removed_references = \
             Msg("Input data includes reference instance(s).\n"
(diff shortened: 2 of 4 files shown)


ajdapretnar:
Yes, this was the idea. We used to have that, but it was removed somewhere along the way. 🤷‍♀️ 
I am a bit unsure regarding treating reference as "neighbor". I understand the idea, the instance has the highest similarity to itself. However, when saying "neighbor", I don't consider myself my own neighbor. :) We can keep it the way it is, I don't object it. It is just a bit more technical approach that requires understanding distances.

ajdapretnar:
@janezd I thought I've already confirmed this. I like the solution. Once you finish the PR, I'm happy to merge.

janezd:
@ajdapretnar , let us put it this way: it's not that I'm an old confused professor. Not at all, I just work on too many different things at once. I am, therefore, a temporarily confused, but it's due to frequent context switches.

Yes, you confirmed it and I already wrote tests. I just haven't marked it as ready for review. I think it's done and probably can be merged.

PR #7204, opened 2025-11-22
Violin Plot: display time as time, not as a number

### LINKED ISSUE #7048: Violin Plot: display datetime/time as datetime/time and not as a number
**What's your use case?**
On the axis, Violin Plot displays datetime variables as numbers (i.e., seconds since t=0):

Image

Just like Box Plot, Violin plot should display datetime variables as datetime variables so that they can be interpreted straightforwardly.

Not that I know of

janezd:
##### Issue

Fixes #7048

##### Description of changes

Use `AxisItem` from the scatter plot util, and call `use_time` as necessary.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/visualize/owscatterplotgraph.py
+++ b/Orange/widgets/visualize/owscatterplotgraph.py
@@ -273,6 +273,9 @@ class AxisItem(AxisItem):
         self._use_time = enable
         self.enableAutoSIPrefix(not enable)
 
+    def is_time(self):
+        return self._use_time
+
     def tickValues(self, minVal, maxVal, size):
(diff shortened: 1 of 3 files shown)


Orange/widgets/visualize/owscatterplotgraph.py
```diff
@@ -273,6 +273,9 @@ def use_time(self, enable):
         self._use_time = enable
         self.enableAutoSIPrefix(not enable)
 
+    def is_time(self):
```
janezd: For easier testing.

PR #7205, opened 2025-11-22
Implement MAPE and SMAPE

### LINKED ISSUE #7041: Predictions and Test & Score: MAPE produces value where it should fail
**What's wrong?**
The formula for the Mean Absolute Percentage Error (see [here](https://en.wikipedia.org/wiki/Mean_absolute_percentage_error) or [here](https://medium.com/@tapassd94/smape-vs-mape-f2249cbd4ad5)) produces a zero division if the test data contains actual values that are zero.
Yet Predictions and Test & Score produce a (very large) value for MAPE if there are zeroes in the test data.

In the attached workflow, Formula is used to compute the Absolute Percentage Error for every record, of which MAPE is the mean over all rows. As it should, Formula fails if rows where the target equals zero are included, but at the same time Predictions and Test & Score produce very large values for MAPE.

[MAPE zero division.zip](https://github.com/user-attachments/files/19009292/MAPE.zero.division.zip)

Perhaps my idea to include MAPE wasn't a very good one, since it is a problematic measure. Perhaps consider adding sMAPE (Python code provided [here](https://medium.com/@tapassd94/smape-vs-mape-f2249cbd4ad5)) or even replacing MAPE by it.

janezd:
##### Issue

Fixes #7041.

If a function cannot compute a value (e.g. because of division by zero), it is an awesome idea to return an arbitrary value instead. At least this is a common practice in SciPy. For instance, when MAPE is infinite, it returns 1/epsilon instead, which is a meaningless large number.

##### Description of changes

I may not entirely subscribe to the above approach, though, so this PR contains a reimplementation of MAPE that returns inf in such case.

It also implements [SMAPE](https://en.wikipedia.org/wiki/Symmetric_mean_absolute_percentage_error), as suggested in #7041.

**To discuss:** I guess we should hide MAPE by default in predictions.

A workflow and data for which MAPE now returns inf: 

[test-mape.zip](https://github.com/user-attachments/files/23690742/test-mape.zip)

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/evaluation/scoring.py
+++ b/Orange/evaluation/scoring.py
@@ -23,7 +23,8 @@ from Orange.util import OrangeDeprecationWarning
 
 
 __all__ = ["CA", "Precision", "Recall", "F1", "PrecisionRecallFSupport", "AUC",
-           "MSE", "RMSE", "MAE", "MAPE", "R2", "LogLoss", "MatthewsCorrCoefficient"]
+           "MSE", "RMSE", "MAE", "MAPE", "SMAPE", "R2", "LogLoss",
+           "MatthewsCorrCoefficient"]
 
 
 class ScoreMetaType(WrapperMeta):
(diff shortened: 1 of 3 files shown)


JakaKokosar:
LGTM!

PR #7201, opened 2025-11-21
test_owparameterfitter: smaller test data, faster learners

markotoplak:
##### Issue

Quite often tests on Windows fail with.

```
======================================================================
FAIL: test_tooltip (Orange.widgets.evaluate.tests.test_owparameterfitter.TestOWParameterFitter.test_tooltip)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\unittest\mock.py", line 1378, in patched
    return func(*newargs, **newkeywargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\a\orange3\orange3\.tox\orange-released\Lib\site-packages\Orange\widgets\evaluate\tests\test_owparameterfitter.py", line 149, in test_tooltip
    self.wait_until_finished()
  File "D:\a\orange3\orange3\.tox\orange-released\Lib\site-packages\orangewidget\tests\base.py", line 529, in wait_until_finished
    self.assertTrue(
AssertionError: False is not true : Did not finish in the specified 5000ms timeout
```

##### Solution

This made tests a bit faster on my linux machine. Hoping it will also work better on Github Actions.

PR #7211, opened 2025-12-02
test_owparameterfitter: stop computation if results are not required

markotoplak:
##### Issue
#7201 did not solve all timeouts. This seems to: I re-run Github actions 4 times, and never got an issue in these tests.

##### Description of changes
Many tests don't actually require results of the fitting process, thus I stopped the fitting process explicitly. I think that this also stopped how many things were still running in parallel when multiple tests were executed. This decreased the runtime of test_owparameterfitter from 8s to 4s on my computer.

##### Includes
- [ ] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/evaluate/tests/test_owparameterfitter.py
+++ b/Orange/widgets/evaluate/tests/test_owparameterfitter.py
@@ -48,34 +48,38 @@ class TestOWParameterFitter(WidgetTest):
         self.widget.controls.maximum.setValue(6)
 
         self.send_signal(self.widget.Inputs.learner, self._pls)
+        self.widget.cancel()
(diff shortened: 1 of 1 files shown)


markotoplak:
> I think that this also stopped how many things were still running in parallel when multiple tests were executed.

I checked the `WidgetTest` class and this is indeed what is happening.  After a test (method) finishes, the widget object is still present and kicking (and perhaps computing in a thread). Widgets are only disposed of in `tearDownClass`. If widgets were, instead, deleted in `tearDown`, some of the `self.widget.cancel()` would be superfluous - and maybe we wouldn't eve get such timeouts in the first place.

@ales-erjavec, do you perhaps remember why `widgets` are not rather cleaned in `tearDown`?

PR #7224, opened 2025-12-28
ScatterplotGraph: Group points within dense cells

### LINKED ISSUE #7064: Geo Map: Randomize coloring by category
Coloring by category colors sequentially, namely one color first, then second, then third. This results in the overwhelm of a single color when observing Geo Map. Example below, where red seems to be the most prominent color, even though it is not the most frequent category in the data.

Proposal: color randomly so that color distribution is more evident. Or, even better, enable clustering on Geo Map that shows some color distribution (pie chart or mode).

Image

janezd:
##### Issue

Fixes #7064.

##### Description of changes

@ajdapretnar might get her new year's gift. :)

This doesn't check for actual overlapping of points. It divides the plotted area into 50-pixel grid and plots cells with > 10 points as bigger circles or pie charts (in case there are multiple colors).

The change is mostly constrained to the `paint` method and I'd prefer to keep it this way. The scatter plot graph is too complex to be poked. :)

Aggregation is opt-in and is disabled if:

- widget shows subset or selection,
- jittering is enabled,
- labels are shown.

https://github.com/user-attachments/assets/eab8ef44-c5a7-4388-90a2-61507e4e46c3

https://github.com/user-attachments/assets/aaeca847-561f-4309-91ca-86b0be59b284

##### Includes
- [X] Code changes
- [ ] Tests?! There's not much that can be tested, everything is in paint :|
- [ ] Documentation

### DIFF
--- a/i18n/si/msgs.jaml
+++ b/i18n/si/msgs.jaml
@@ -196,9 +196,9 @@ util.py:
         def `funcv`:
             unsafe: false
 version.py:
-    3.40.0: false
-    3.40.0.dev0+e7d811b: null
-    e7d811bac580723c707decd32c550a75874dc9c8: null
+    3.41.0: false
+    3.41.0.dev0+58e68e8: false
+    58e68e8bc9a6b560967ca49f37f89a3dd0d213b8: false
     .dev: false
 canvas/__main__.py:
     ORANGE_STATISTICS_API_URL: false
@@ -13640,6 +13640,12 @@ widgets/utils/plot/owplotgui.py:
         def `class_density_check_box`:
             class_density: false
             Show color regions: Pokaži barvna področja
+        def `aggregate_points_check_box`:
+            aggregate_dense_regions: false
+            Aggregate points in dense regions: Združi točke v gostih regijah
+            'Dense regions with many points are aggregated into ': "Goste regije z veliko točkami so združene v "
+            circles or pie charts,\n: kroge ali tortne diagrame,\n
+            unless data is jittered or labels, selection or subset is shown: razen če so točke tresene ali pa so prikazane oznake, izbori ali podmnožice.
         def `regression_line_check_box`:
             show_reg_line: false
             Show regression line: Pokaži regresijsko premico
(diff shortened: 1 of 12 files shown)


ajdapretnar:
I like it a lot for Geo Map! We can discuss the opacity of the circles (seem a bit too translucent to me), but otherwise a fantastic NYE gift!

i18n/si/msgs.jaml
```diff
@@ -13640,6 +13640,12 @@ widgets/utils/plot/owplotgui.py:
         def `class_density_check_box`:
             class_density: false
             Show color regions: Pokaži barvna področja
+        def `aggregate_points_check_box`:
+            aggregate_dense_regions: false
+            Aggregate points in dense regions: Združi točke v gostih regijah
+            'Dense regions with many points are aggregated into ': "Goste regije z veliko točkami so združene v "
+            circles or pie charts,\n: kroge ali tortne diagrame,\n
+            unless data is jittered or labels, selection or subset is shown: razen če so točke tresene ali pa so prikazane oznake, izbori ali podmnožice.
```
VesnaT: The original (eng.) is missing the dot. Just saying, I don't mind the difference.

VesnaT (commented):
The tooltips over the pies are incorrect. Other than that, it looks great.

janezd:
> The tooltips over the pies are incorrect. Other than that, it looks great.

Huh, this was not as trivial as it sounds.

~~I solved it by storing the indices of points in each cell grid and tied the tooltip to the grid, not the show points. As a result, a tooltip is shown if the mouse pointer is in the same cell, which might mean "close enough" to the shown piechart. If the piechart crosses the cell boundary (which should be rare!) and the pointer is on the wrong side, this may show a wrong tooltip. If this is a problem, I'd prefer fixing it by moving the piechart. Or even disable tooltips on aggregated points.~~

~~A proper solution - checking whether the cursor is actually on the object, is much more complicated. Piecharts are just painted, they are not real objects (e.g. proper QGraphicsItems). They are painted in transformed coordinates. In "real-world" coordinates, they are not even circles but ellipses.~~

I was stupid and possibly lazy. I've done it properly now.

PR #7226, opened 2025-12-29
Feature Statistics: Disable stretching of distribution column, set minimal width

### LINKED ISSUE #6983: Feature Statistics: cannot resize Distribution bar
**What's wrong?**

One cannot resize the Distribution bar. Admittedly, this is not the best place to inspect distribution, but the current size makes it useless to have.

Screenshot 2025-01-15 at 12 07 52

File - Feature Statistics

**What's your environment?**

- Operating system: OSX
- Orange version: 3.38.1

janezd:
##### Issue

Fixes #6983. I think. @ajdapretnar ?

##### Description of changes

Column distributions was set to stretch, i.e. occupy all extra width. Other columns had a default size (100).

After this, columns have a minimum size of 100, and the distribution column gets no special treatment. Sizes are reasonable. Downside is that they can't be manually reduced below 100.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owfeaturestatistics.py
+++ b/Orange/widgets/data/owfeaturestatistics.py
@@ -632,7 +632,7 @@ class FeatureStatisticsTableModel(AbstractSortTableModel):
 
 class FeatureStatisticsTableView(QTableView):
     HISTOGRAM_ASPECT_RATIO = (7, 3)
-    MINIMUM_HISTOGRAM_HEIGHT = 50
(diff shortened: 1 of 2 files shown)


ajdapretnar:
I like everything about this, except for the very first column with variable type, which IMO would never be bigger than minimum size. It takes unnecessary space now. I'd make an exception for this column being at absolute minimum, with the other columns as they are now.

PR #7231, opened 2025-12-30
[FIX] Create Class: Remove context settings in favour of schema-only non-context

### LINKED ISSUE #5499: Keeping it classy is hard
- [ ] **What's wrong?**

I'm trying to establish a class that pulls out the temperatures from my data set, but every time I try to delete a previous class I'm not using for this data set it overwrites my edits and it's very frustrating. I've had this happen in a couple of places, but this is the one that's making things harder right now.
[bug_class.zip](https://github.com/biolab/orange3/files/6694739/bug_class.zip)

Just open it up and delete the one that's labeled "delete" and watch the one I care about turn into the one that got deleted.

- Operating system: windows
- Orange version: 3.30.0.dev
- How you installed Orange: command line

### LINKED ISSUE #6961: New layout of “Create class” widget not reflected in docs
**What's wrong?**

This is just a cosmetic issue, but the “Create class” widget has a new layout which has not been updated in the docs yet, as shown in the attached screenshot.

Drag a “Create class” widget to the canvas, open the widget, press the “?” button, compare the layout of the actual widget with the docs.

**What's your environment?**

- Operating system: macOS 15.2
- Orange version: 3.38.0
- How you installed Orange: Installer at https://download.biolab.si/download/files/Orange3-3.38.0-Python3.10.11-x86_64.dmg
Screenshot 2024-12-20 at 11 36 57

janezd:
##### Issue

Fixes #5499, assuming that weird behaviour is caused by context settings. (Which it almost certainly is.)

##### Description of changes

As per title, I replaced context settings with schema-only (non-context) settings. The only remaining context setting is the variable.

It gets even better. Rules were already an "internal context setting" with the variable as a context. That is, if the user sets the rules for one variable, switches to another, sets some rules, and switches back (s)he will get the rules for the former variable.

After removing context settings, this works as hinting. Load zoo, set some rules for type. Load Iris (rules disappear), set different rules. Go back to zoo and you get the rules back. This hinting is based on variable name.

If this is merged and documentation is updated, close #6961, too.

##### Includes
- [X] Code changes
- [X] Tests

### DIFF
--- a/Orange/widgets/data/owcreateclass.py
+++ b/Orange/widgets/data/owcreateclass.py
@@ -14,7 +14,7 @@ from Orange.data.table import Table
 from Orange.statistics.util import bincount
 from Orange.preprocess.transformation import Transformation, Lookup
 from Orange.widgets import gui, widget
-from Orange.widgets.settings import DomainContextHandler, ContextSetting
+from Orange.widgets.settings import DomainContextHandler, ContextSetting, Setting
 from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.localization import pl
 from Orange.widgets.utils.widgetpreview import WidgetPreview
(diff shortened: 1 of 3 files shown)


markotoplak:
@janezd, I (think I) fixed settings migration. You forgot to migrate the attribute (which stays in the context and regular expression toggle.

I do not quite understand what you wanted to do with `update_dynamic_height`. Now it is only used at init (it has another unused parameter), and for me, what I see it doing, is to make the widget wider. I never missed width here. Or what was the reasoning behind it?

janezd:
`update_dynamic_height` was likely an attempt to reduce the height of the widget when the user removes a rule. It probably didn't work and I removed the call from `remove_row`, but forgot to remove the entire function and the call from `__init__`.

Now I removed the function and set the flag that will resize the widget automatically.

PR #7232, opened 2025-12-31
Discretize and Continuize: Unify design

### LINKED ISSUE #6984: Unify Discretize and Continuize
Discretize and Continuize have different wording for default. Judging by the outdated documentation, Continuize used to say "Default setting: leave". Discretize has "Default setting: keep", which is already inconsistent. However, Continuize now has "Preset: no change". It would be nice to have this streamlined.

Screenshot 2025-01-15 at 13 10 16
Screenshot 2025-01-15 at 13 10 19

janezd:
##### Issue

Resolves #6984.

##### Description of changes

I adopted a term "General preset" and consistently use it in both widges.

The Discretize widget had very large spacing between radio buttons; the height was made uniform by adjusting all heights to the largest height. Now the height of smaller items is adjusted to the mean  between minimal and maximal. This looks better, imho.

##### Includes
- [X] Code changes

### DIFF
--- a/Orange/widgets/data/owcontinuize.py
+++ b/Orange/widgets/data/owcontinuize.py
@@ -195,25 +195,16 @@ class ListViewSearch(listview.ListViewSearch):
         """
         A delegate that shows items (variables) with specific settings in bold
         """
-        def __init__(self, *args, **kwargs):
-            super().__init__(*args, **kwargs)
-            self._default_hints = False
-
         def initStyleOption(self, option, index):
             super().initStyleOption(option, index)
             hint = index.data(ContDomainModel.HintRole)
             option.font.setBold(hint is not None and hint[1])
 
-        def set_default_hints(self, show):
-            self._default_hints = show
-
         def displayText(self, value, _):
             if value is None:
                 return None
-            name, hint, nondefault = value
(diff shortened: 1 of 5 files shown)


ajdapretnar:
I like the choice of "General preset". However, if I might nitpick...
Discretize shows, i.e. `age (as preset)`. Continuize doesn't show anything, unless in some weird (irreproducible) instance when it shows `gender: preset` upon hover. Would it be possible to unify this, too? So both show the same thing. I propose `age: preset` (semicolon).
Oh, also, move `Use general preset` in Discretize to index 0, as is in Continuize.
But again, it's nitpicking, so happy to merge as is (unless we can reproduce the weird on hover issue).

janezd:
I followed your suggestions. I actually noticed all of these when making the PR. :)

- The idea behind showing individual settings on tooltip (in Continuize) was to reduce clutter and I thought to add it to Discretize, too, but was too lazy. Now I see it's unreliable and not really needed, so settings are now always shown in both widgets.
- I danced around colons and parentheses a lot. Now I did as you suggested.
- I was scared to change the order of radio buttons, but now I see widget is actually implemented well, and is resilient to changing the order.

Please squash the commits when merging.

PR #7233, opened 2025-12-31
Transpose: Rename output column

### LINKED ISSUE #7086: Transpose: address inadequate naming
Similar to what was reported in #7084 regarding Feature Statistics, the output of Transpose assigns the header Feature Name  to its first column while it doesn't necessarily hold just features, but also metas and targets. A quick fix would be to rename this column to Variable Name or Variable.

Actually, I propose some further improvements, since Transpose is particular useful for datasets like this one:

Image

The screenshot shows a set of users with ratings they gave to book titles. At some point, one might be interested in finding similar users based on distances between their rating profiles. But at some other point, in the same workflow (or another one), one might be interested in finding similar books based on users that gave them similar ratings. For the second investigation, Transpose can be applied. In that case, it would be nice to have an option to user-define the name of the first column. Also if UserID would have been a feature or target (and thus appear at the end of the rows), it would be nice if the user could select which of types of input columns to keep from metas, features and targets and/or categorical, numeric, datetime and text using check boxes.
Of course this could also be achieved using Select Columns before or Select Rows after Transpose, but in some cases this might require manually selecting many items or defining complex selection rules.

janezd:
##### Issue

Resolves #7086.

##### Description of changes

Default name is now "Column name", but user can change it.

To maintain compatibility with older workflows, settings migration sets the column name to "Feature name" -- as if the user manually changed it!

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/data/owtranspose.py
+++ b/Orange/widgets/data/owtranspose.py
@@ -82,7 +82,7 @@ class OWTranspose(OWWidget, ConcurrentWidgetMixin):
 
         # self.apply is changed later, pylint: disable=unnecessary-lambda
         box = gui.radioButtons(
-            self.controlArea, self, "feature_type", box="Feature names",
+            self.controlArea, self, "feature_type", box="Output column names",
             callback=self.commit.deferred)
 
         button = gui.appendRadioButton(box, "Generic")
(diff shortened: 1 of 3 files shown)


markotoplak:
That is an elegant settings migration!

markotoplak:
Also, while we are doing this... 

The first control box is called "Feature names". Should we rename to "Column names" or maybe "Output column names"?

Then, there is "From variable" which could also be to "From column". What do you think?

janezd:
Yes, it probably has to be a schema-only setting. (I asked you for this PR because you're an expert for noticing quirks related to settings.)

I changed feature and variable to column in GUI, as you suggested. I also added "output" and "original" to make clear what names we refer to.

PR #7250, opened 2026-03-03
Pin chardet<7

janezd:
##### Issue

**Note: #7250, if it works, is preferred over this PR.**

"Nightly wheels" tests fail with

```
  File "/home/runner/work/orange3/orange3/.tox/beta/lib/python3.14/site-packages/Orange/data/io_util.py", line 15, in <module>
    from chardet.universaldetector import UniversalDetector
ModuleNotFoundError: No module named 'chardet.universaldetector'
```

because `UniveralDetector was moved to `chardet.detector` in chardet 7.

##### Description of changes

Pin chardet to <7. Alternative fixes would be
- to try to import from the new location if the current import fails.
- to pin chardet to >=7.

### DIFF
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -1,7 +1,7 @@
 baycomp>=1.0.2
 bottleneck>=1.3.4
 # Encoding detection
-chardet>=3.0.2
+chardet>=3.0.2,<7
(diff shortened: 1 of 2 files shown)


janezd:
Closed in favor of #7252.

PR #7286, opened 2026-05-21
Scatterplotgraph: Implement manual repositioning of points

janezd:
##### Description of changes

- Implement moving of points (a single point or selection) by dragging them.

    The functionality is enabled if the widget (not graph!) implements a method `set_coordinates`.
- Implement this in MDS.
- Also, implement saving of coordinate hints in MDS.

I don't know how to write reasonable tests for this. Test on https://github.com/biolab/orange3-network/pull/293.

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/widgets/unsupervised/owmds.py
+++ b/Orange/widgets/unsupervised/owmds.py
@@ -520,6 +520,9 @@ class OWMDS(OWDataProjectionWidget, ConcurrentWidgetMixin):
             self.update_stress()
             self.graph.update_sizes()
 
+    def finish_dragging(self):
+        self.commit.deferred()
+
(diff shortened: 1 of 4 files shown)


VesnaT (commented):
On point(s) move the output should be updated.

janezd:
> On point(s) move the output should be updated.

I admire your ability to catch such things.

My first fix was to call commit on all drag events, which looked nice in a trivial workflow -- I dragged a point in the MDS and it instantly updated in the scatter plot. In reality, it could cause too many updates and redraws; now the new coordinates are committed when dragging is finished.

PR #7292, opened 2026-06-02
Change behavior of set class_var by RowInstance.

### LINKED ISSUE #7291: Assignment to one of the class columns modifies all class columns
**What's wrong?**

This is a behavior in the Orange python library.

When modifying a value in a class column/variable in a Table using `__setitem__`, all class columns will be set to that value. E.g. if `data` has two class columns at index 1 and 2, setting `data[0][1] = 42` will also set `data[0][2]`.

I presume this is a bug? (It certainly caused me a nasty one :))

The described behavior can be reproduced as follows:

```python
import numpy as np
from Orange.data import Domain, Table, ContinuousVariable, DiscreteVariable

foo = ContinuousVariable("Foo")
cont = ContinuousVariable("Cont Var")
disc = DiscreteVariable("Disc Var", values=("0", "1"))

domain = Domain([foo], [cont, disc], [])

n = 2
X = np.array([[1], [2]])
Y = np.array([
    [float("nan"), float("nan")],
    [float("nan"), float("nan")]
], dtype=float)

data = Table(domain, X, Y, None)

print(data)
row = data[1]
print("before:", row._y)
row[1] = 4
print("after:", row._y) # -> both class columns set to 4 for this row

# this then throws an error because 4 is an invalid value for the "Event" column:
print(data)
```

**What's your environment?**
- Operating system: Windows 11 Pro 10.0.26200
- Orange version: pypi package `Orange3 @ git+https://github.com/biolab/orange3@ca7307d6c820782afb5bfc923d3d980c56cf3b3d`
(but should be the same in the latest release)

krystofair:
Now one can change single cell on class_var table, but must to be remembered that indexes are relative to length of attributes (table.X). So to change index zero it must be exactly `len(table.X[row_index])`.

##### Issue

Fixes #7291

##### Includes
- [X] Code changes
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Orange/data/table.py
+++ b/Orange/data/table.py
@@ -150,9 +150,9 @@ class RowInstance(Instance):
                     self._x[key] = value
             else:
                 if self.sparse_y is not None:
-                    self.table._Y[self.row_index, key - len(self._x)] = value
+                    self._y[key - len(self._x)] = value
                 else:
-                    self.table._Y[self.row_index] = value
+                    self.table.Y[self.row_index, key - len(self._x)] = value
                     if self.table._Y.ndim == 1:  # if _y is not a view
                         self._y[0] = value
         else:


krystofair:
Okay that is impossible in this way. I am closing it.

Issue #6942, opened 2024-11-29
Change in the Neighbours widget


janezd:
@BlazZupan started a discussion about the Neighbours widgets. He basically, wanted an output Data with distances to *all* data instance, and overlooked that he can achieve the same effect by unchecking the checkbox that limits the number of neighbours.

Here are some notes:

**Neighbours**:

- Add missing distances (normalized Euclidean and Manhatta). (Note to self: an input with distances is not a solution because it would incur quadratic time complexity).
- Add an output Data with all instance and their distances to the reference, and **make it lazy**. The output is superfluous, but somehow consistent with other widgets that output subsets, and it's harmless if lazy.
- Perhaps: remove the checkbox, keep line edit. This is still a matter of discussion; removing the checkbox requires some code for backward compatibilitty.

We also discussed how to provide distances to multiple references. One option was an additional input to Distances of type Data Table that would represent reference examples), and an output with distances to all those instance. The problems with this solution are
- potentially many meta columns with distances
- Distances can currently be connected only to a few widgets, which is nice. With the additional output, it can be connected to almost everything, so the pop-up is populated with lots and lots of widgets.
- We want additional options, like a column that contains an id of the closest reference, or some kind of aggregation (min, max, avg) of distances to refrences.

We haven't decided anything, but it seems this calls for a new widget.

PR #7320, opened 2026-09-12
[ENH] Neighbors: unify available metrics with Distances

markotoplak:
##### Issue
**Neighbors** and **Distances** have a different set of available metrics.

I notices this when reviewing comments to one of our courses. There, students had to first use **Distances** with the normalized euclidean distance, and then get the closest elements in an easier way with **Neighbors**. But the problem was that the results did not match, as **Neighbors** does not support normalization. The particular problematic section is this one: https://notes.biolab.si/books/arisa/intro-ml#Distances

##### Description of changes
Neighbors now use the same MetricDefs as Distances (these were also moved into utils.py).

##### Includes
- [X] Code changes
- [X] Tests
- [ ] Documentation


janezd:
Just a thought.

This is related to https://github.com/biolab/orange3/issues/6942. See the first bullet. There I've made a comment to self, but there's another option: the Distances widget could output a "fitter" -- distances already have a fitter-model structure, though not exposed in widgets. The output of this widget could then be used in Neighbours, Outliers, Louvain(?), DBScan(?), and Network from Distances and Network of Neighbours in the Network add-on (as an alternative to the distance chosen in the combo).

On the other hand, we might prefer to replicate the change from this issues in other widgets. This is easier for the user to understand; "Distance" (as a method) would be even a stranger concept than a "Learner". Besides, few users would realize that they can connect Distances -> Outliers. With your change, we can add distances to `MetricDefs` and they would be available in all widgets -- without needing to change the widgets.

markotoplak:
Oh, we indeed had an issue on this. :) 

So the suggested feature would be kind of orthogonal to this PR, but I don't think we need a new output. The added column of distances could just get a `.compute_value` and then that applicably with "Apply Domain". This is how you can use, for example, the k-Means output to give you cluster assignments for new data. Not an ideal solution because of the Apply Domain.

Orange/widgets/data/owneighbors.py
```diff
@@ -117,18 +116,18 @@ def compute_distances(self):
...
+        metric_def = MetricDefs[self.metric_id]
+        metric = metric_def.metric
 
         # comparing only attributes, no metas and class-vars
         new_domain = Domain(self.data.domain.attributes)
         reference = self.reference.transform(new_domain)
         data = self.data.transform(new_domain)
 
-        all_data = Table.concatenate([reference, data], 0)
-        pp_all_data = Impute()(RemoveNaNColumns()(all_data))
```
janezd, line 129: I think you should keep this (or something similar). Try the changed widget on `Brown selected`, set distance to Mahalanobis.

I think the best practice would be to check `metric_def.metric.supports_missing`; impute if `False`. Otherwise, the metric probably does something better, so we should not impute.

markotoplak: Thanks for this. I wrongly assumed "impute=True" means this.

markotoplak:
@janezd, until the last commit I tried replicating previous behavior as closely as possible. I kept the previous imputation code - and did not replace it with the same imputation as **Distances** uses - because this code also properly worked with discrete variables (**Distances** has other safeguards). Let's forget for a moment that these don't work with most metrics.

In the last commit I then changed the code that was making data and reference compatible and doing imputation. I wanted to avoid processing both together because that is just weird. Now the imputation is only done on the data, and then the imputed domain is used to transform references. To me, this makes the code clearer. Imputation of the reference is still influenced by the (other) data, but I don't have a better idea. Feel free to disregard the last commit.

While doing this there are still some issues remaining. **Distances** is very thorough in data preprocessing and raising warnings and ideally that behavior should also be replicated here. Someday.

janezd:
I think that not using reference instances in construction of imputation makes more sense than the existing code.

Say that we select a reference example; we get the closest neighbor(s). Then we add another reference example (without removing the existing one). With the old behaviour, this could affect the imputation and change the set of nearest neighbours for the existing reference, too. This looks like transductive learning, but I don't think we want that fanciness here.

janezd:
Not to mention that it simplifies the code. :)

